PR #5, opened 2019-07-01
Add license

kainino0x:
@kangz

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,27 @@
+# GPU for the Web
+
+This repository is being used for work in the [W3C GPU for the Web Community
+Group](https://www.w3.org/community/gpu/), governed by the [W3C Community
+License Agreement (CLA)](http://www.w3.org/community/about/agreements/cla/). To
+make substantive contributions, you must join the CG.
+
+Contributions to the source code repository are subject to the terms of the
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,27 @@
+# GPU for the Web
+
+This repository is being used for work in the [W3C GPU for the Web Community
```
Kangz: Where did you get this template?

package.json
```diff
@@ -10,9 +10,9 @@
     "type": "git",
     "url": "git+https://github.com/kainino0x/cts-experiment.git"
   },
-  "author": "Kai Ninomiya <kainino@chromium.org>",
+  "author": "WebGPU CTS Contributors",
   "private": true,
-  "license": "W3C 3-clause BSD License",
```
Kangz: Why not W3C 3-clause BSD?

CONTRIBUTING.md
kainino0x: gpuweb/admin

package.json
kainino0x: `BSD-3-Clause` is a standard (SPDX) string, so using it here allows machine understanding of the field.

AFAICT (from gpuweb/admin), the license that was agreed on for gpuweb is actually the vanilla BSD-3-Clause, not the W3C 3-clause BSD License. I checked that the LICENSE in gpuweb/admin is the SPDX BSD-3-Clause exactly.

PR #7, opened 2019-07-03
Further simplify and begin documenting loading-related interfaces

kainino0x:


kainino0x:
Commits should be reviewable separately.

src/framework/loader.ts
```diff
@@ -1,32 +1,23 @@
...
-  readonly description: string;
-}
-
-export interface TestSuiteListing {
-  readonly suite: string;
-  readonly groups: Iterable<TestGroupDesc>;
-}
-
+// One of the following:
+// - A shell object describing a directory (from its README.txt).
```
Kangz: It's weird that the README would be there, can we use a sum type with README | TestSpecFile wherever this is needed?

PR #9, opened 2019-07-09
enable tslint rule to ensure some type annotations

kainino0x:

PR #8, opened 2019-07-08
Move test filters to classes, move loadFilter to a file, improve terminology

kainino0x:

PR #10, opened 2019-07-09
Start docs, and refactors found while writing docs

kainino0x:
Depends on #8; ignore those commits. (They may disappear if you merge #8 and #9.)

A lot of code moved (again), but there's not much meat to them so feel free to skim.

### DIFF
--- a/docs/terms.md
+++ b/docs/terms.md
@@ -1,8 +1,24 @@
+# Overview
+
+- _Suites_ contain multiple:
+  - _READMEs_
+  - _Test Spec Files_. Contains one:
+    - _Test Group_. Defines a _test fixture_ and contains multiple:
+      - _Tests_. Defines a _test function_ and contains multiple:
+        - _Test Cases_, each with the same test function but different _parameters_.
+- _Queries_ contain multiple:
+  - _Filters_ (positive or negative). Each filter may match one of:
+    - `S:p` In one suite `S`, all specs whose paths start with `p`.
+    - `S:s:t` In one spec `S:s`, all tests whose names start with `t`.
+    - `S:s:t~c` In one test `S:s:t`, all cases whose params are a superset of `c`.
+    - `S:s:t:c` In one test `S:s:t`, the single case whose params equal `c`.
+
 ## Suite
 
 A suite of tests.
-A single suite has a directory structure, and many _test spec files_.
-Each test spec file is identified by its path within the suite.
+A single suite has a directory structure, and many _test spec files_
+(`.spec.ts` files containing tests) and _READMEs_.
+Each member of a suite is identified by its path within the suite.
 
 **Example:** `src/suites/*/`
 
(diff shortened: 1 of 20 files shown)


docs/terms.md
```diff
@@ -0,0 +1,142 @@
...
+A single suite has a directory structure, and many _test spec files_.
+Each test spec file is identified by its path within the suite.
+
+**Example:** `src/suites/*/`
+
+### README
+
+**Example:** `src/suites/**/README.txt`
+
+Describes (in prose) the contents of a subdirectory in a suite.
```
Kangz: nit: Will they be surfaced somewhere in the runner UI?

docs/terms.md
```diff
@@ -0,0 +1,142 @@
...
+
+**Type:** After a README is loaded, it is stored as a `ReadmeFile`.
+
+## IDs
+
+### Test Spec ID
+
+Uniquely identifies a single test spec file.
+Comprised of suite name (`suite`) and test spec file path (`path`).
```
Kangz: and a test spec file path relative to the suite's root directory.

docs/terms.md
```diff
@@ -0,0 +1,142 @@
+## Suite
+
+A suite of tests.
+A single suite has a directory structure, and many _test spec files_.
+Each test spec file is identified by its path within the suite.
+
+**Example:** `src/suites/*/`
+
+### README
```
Kangz: Can we also explain what a test spec is here?

docs/terms.md
```diff
@@ -0,0 +1,142 @@
...
+### Filtered Test Spec
+
+A subset of the tests in a single test spec file.
+It is created at runtime, via a filter.
+
+It contains one "virtual" test group, which has type `RunCaseIterable`.
+
+## Test Group / Group
+
+A collection of test cases. There is one per test spec file.
```
Kangz: Is this necessarily true? Is there some use to having more than one group per spec?

docs/terms.md
```diff
@@ -0,0 +1,142 @@
+## Suite
```
Kangz: Maybe a comment on top describing the hierarchical structure of things with suite, test spec, test and params?

docs/terms.md
kainino0x: They are surfaced in the standalone runner (kind of poorly for now, but it's there). It was broken for a while but I fixed it recently.

docs/terms.md
kainino0x: It's sort of fundamentally true right now because groups are not identifiable within files (i.e. the path to a file uniquely identifies a group).

kainino0x: But there's a convention I'm using which is that if you want a hierarchy of tests within a test group, name them with slashes between the parts. I'm planning to actually make this nested in the standalone runner eventually.

PR #12, opened 2019-07-26
Spec compliance fixes, separate wpt build dir, and remove all tests with shaders

kainino0x:
These changes make it so the tests run on both Chrome (except one Chrome bug: unmap detach) and Safari (except for ~~two~~ three minor needed Safari changes: unmap detach, TRANSFER->COPY, loadOp->loadValue).

Commits can be reviewed separately.


src/suites/cts/command_buffer/render/render.spec.ts-disabled
```diff
@@ -0,0 +1,73 @@
+import { TestGroup } from '../../../../framework/index.js';
+import { GPUTest } from '../../gpu_test.js';
+
+export const g = new TestGroup(GPUTest);
```
austinEng: Should this be GPUShaderTest?

kainino0x: Technically yes, but I don't really intend to keep the GPUShaderTest class separate in the long run, so whenever I turn these back on I'll deal with it.

kainino0x: Actually, fixed, just in case anyone does try to use these between now and whenever I fix it.

PR #17, opened 2019-08-02
Examples

kainino0x:


src/framework/fixture.ts
```diff
@@ -43,11 +43,33 @@ export class Fixture {
...
   }
 
-  async shouldReject(p: Promise<unknown>, exceptionType?: string, msg?: string): Promise<void> {
+  protected async asyncExpectation(fn: () => Promise<void>): Promise<void> {
     this.numOutstandingAsyncExpectations++;
+    await fn();
+    this.numOutstandingAsyncExpectations--;
+  }
+
+  async shouldReject(p: Promise<unknown>, exceptionType?: string, msg?: string): Promise<void> {
```
austinEng: nit: If we don't care about the exceptionType, we'll have to call this as `shouldReject(p, undefined, msg)`
Thoughts on having the argument be a params dict `{ exceptionType?: string, msg?: string }` ?

src/suites/cts/examples.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+// - http://localhost:8080/?runnow=1&q=cts:examples:
+
+export const g = new TestGroup(GPUTest);
+
+g.test('basic', t => {
+  t.expect(true);
+  t.expect(true, 'true should be true');
+
+  t.shouldThrow(
+    () => {
```
austinEng: unit: Will the formatter let this function be one line?

src/suites/cts/gpu_test.ts
```diff
@@ -16,41 +16,43 @@ export class GPUTest extends Fixture {
...
-    }
+      if (size <= 256 && failedPixels > 0) {
+        const expHex = Array.from(exp)
+          .map(x => x.toString(16).padStart(2, '0'))
+          .join('');
+        const actHex = Array.from(actual)
+          .map(x => x.toString(16).padStart(2, '0'))
+          .join('');
+        this.rec.log('EXP: ' + expHex);
+        this.rec.log('ACT: ' + actHex);
```
austinEng: nit: I'd prefer the full words EXPECT and ACTUAL.

src/framework/fixture.ts
kainino0x: Decided to make the exception name required (and moved it to first position)

src/suites/cts/examples.spec.ts
kainino0x: Unfortunately not, it's very opinionated.

kainino0x:
Thanks for the review! Merging now so I can send out the email, but feel free to send comments on the last few changes.

PR #18, opened 2019-08-03
Use local grunt

greggman:
That way there are less conflicts with other projects

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,8 +20,7 @@ set up dependencies:
 cd cts/
 yarn install
 
-yarn global add grunt-cli  # install grunt globally
-grunt  # show available grunt commands
+npx grunt  # show available grunt commands
 ```
 
 ### Build
@@ -30,13 +29,13 @@ To build and run all pre-submit checks (including type and lint checks and
 unittests), use:
 
 ```sh
-grunt pre
+npx grunt pre
 ```
 
 To just build the project, use:
 
 ```sh
-grunt build
+npx grunt build
 ```
 
 ### Run


kainino0x:
IIUC, installing grunt globally is not supposed to conflict with other projects, because I think it's just a wrapper that calls into the grunt installed in the local node_modules.

That said, this change would be fine but I'd like to keep a note that it can be installed globally too.

(side note: at some point on one of my machines, `npx somecommand` would take about 5-10 seconds longer than `./node_modules/.bin/somecommand` due to some kind of superfluous package updating or something. Not happening anymore though, so I guess it's fine.)

greggman:
I guess it's just a pet peeve. Any project that requires me to install things globally seems out of date, left over from the 90s. There may be other commands named `grunt`. Given it's trivial not to have to install it globally I'd suggest it's better not to even suggest it. Some projects even go so far as to [put node itself in their package.json](https://www.npmjs.com/package/node). You still need one to run npm but it doesn't have to be the same one the project itself uses.

greggman:
Let me add this is one reason I was never able to run the WebGL harness

https://github.com/KhronosGroup/WebGL/tree/master/tools

Because it requires a ton of incompatible prerequisites. If it used everything locally all those instructions become simple and conflicts disappear.

It would be nice not to propagate the same issues in this repo. I get that grunt here might be minor but it's the precedent that starts it off.

kainino0x:
That's convincing enough for me.

Actually, the idea of putting node in the package.json sounds interesting. Do you know enough about that to recommend whether we should do it too?

PR #19, opened 2019-08-07
Add glslang and re-enable shader tests

kainino0x:
This isn't a very good build of glslang (it crashes on one basic thing, see TODO, and it's not very small). But works.


src/suites/cts/gpu_shader_test.ts
```diff
@@ -0,0 +1,31 @@
...
+      });
+    }
+  }
+
+  makeShaderModule(type: 'f' | 'v' | 'c', source: string): GPUShaderModule {
+    return this.device.createShaderModule({ code: this.compile(type, source) });
+  }
+
+  private compile(type: 'f' | 'v' | 'c', source: string): Uint32Array {
+    const t = type === 'f' ? 'fragment' : type === 'v' ? 'vertex' : 'compute';
```
austinEng: nit: If the compiler takes the full string names, let's just use the full name as the `type` argument for better readability.

src/suites/cts/gpu_shader_test.ts
```diff
@@ -0,0 +1,31 @@
...
+        });
+      });
+    }
+  }
+
+  makeShaderModule(type: 'f' | 'v' | 'c', source: string): GPUShaderModule {
+    return this.device.createShaderModule({ code: this.compile(type, source) });
+  }
+
+  private compile(type: 'f' | 'v' | 'c', source: string): Uint32Array {
```
austinEng: nit: Give `'f' | 'v' | 'c'` a name ?

src/suites/cts/command_buffer/render/rendering.spec.ts
```diff
@@ -18,14 +20,15 @@ g.test('fullscreen quad', async t => {
...
       const vec2 pos[3] = vec2[3](
           vec2(-1.f, -3.f), vec2(3.f, 1.f), vec2(-1.f, 1.f));
       gl_Position = vec4(pos[gl_VertexIndex], 0.f, 1.f);
     }
   `
   );
+  // TODO: Fix glslang build so '310 es' can be used here (crashes currently)
   const fragmentModule = t.makeShaderModule(
     'f',
     `#version 450
```
austinEng: It works with this? Or does the test just fail?

austinEng: (It works with `#version 450`?)

kainino0x: yup, it works with 450. Updated comment.

src/suites/cts/gpu_shader_test.ts
kainino0x: Done.

src/suites/cts/gpu_shader_test.ts
kainino0x: Done, figured out how to import the types from glslang-web and fixed all the anys too.

PR #21, opened 2019-08-15
Move GLSL compilation to build time

kainino0x:
This compiles GLSL using a build time macro(!) via babel-plugin-macros, instead of doing shader compilation online. Rationale:

1. It's too annoying to check my current non-size-optimized 900k binary into Chrome and WPT.
2. Even though we'll most likely need to be able to generate shaders programmatically later on, we can get away without runtime compilation for now.
3. We probably aren't going to use it in the long run, so hopefully by the time we need to replace it we'll know what the shading language is.


.babelrc
```diff
@@ -1,20 +1,23 @@
...
+  "plugins": [
+    "@babel/plugin-proposal-class-properties",
+    "@babel/plugin-syntax-dynamic-import",
+    "macros",
+    [
+      "module-resolver",
+      {
+        "root": ["./src"],
+        "alias": {
+          "@webgpu/shaderc": "./src/shaderc.js"
```
austinEng: I don't think we need shaderc anymore

src/tools/glsl.macro.js
```diff
@@ -0,0 +1,40 @@
+const { createMacro, MacroError } = require('babel-plugin-macros');
```
austinEng: nit: Can this be `import`s?

src/tools/glsl.macro.js
```diff
@@ -0,0 +1,40 @@
...
+const glslangModule = require('@webgpu/glslang');
+
+let glslang;
+function initializeGlslang() {
+  if (!glslang) {
+    glslang = glslangModule();
+  }
+}
+
+module.exports = createMacro(({ references, state, babel }) => {
```
austinEng: nit: Can this work with `export default`?

austinEng:
Another thought as I was reviewing the Chromium CL: Can we have the macro insert a comment with the source text of the compiled shader code?

src/tools/glsl.macro.js
kainino0x: Unfortunately it's not easy to get this code to run as an ES6 module. It's running in node as a Babel plugin and I can't figure out how to make it allow that.

austinEng: Ah yes, makes sense.

PR #25, opened 2019-09-09
Add validation tests

### LINKED ISSUE #24: Add validation tests
I've started adding some validation tests at https://github.com/gpuweb/cts/compare/validation
Note that those will need "push/pop Error scope" but for now, they use "uncapturederror".

FYI @Kangz @kainino0x @austinEng 

image

beaufortfrancois:
This PR is about adding some validation tests to the CTS.
Once "push/pop error scope" is implemented, I'll update `expectUncapturedError`.

R: @kainino0x @Kangz @austinEng


src/suites/cts/validation_test.ts
```diff
@@ -0,0 +1,233 @@
...
+  }
+
+  getErrorBuffer(): GPUBuffer {
+    return this.device.createBuffer({
+      size: 1024,
+      usage: 0xffff, // Invalid GPUBufferUsage
+    });
+  }
+
+  expectUncapturedError(commandBuffer: GPUCommandBuffer[] = []): Promise<void> {
```
Kangz: Our experience in Dawn is that is is very useful to know exactly when a validation error occurred. When updating to use `pushErrorScope` and `popErrorScope` I suggest making it so tests canbe written similar to this:

```
expectValidationError(() => {device.createBuffer(descriptor)});
```

A validation error must happen in the function passed to `expectValidationError` and no validation error must happen in the rest of the test (this can be checked by wrapping the whole test in an error scope).

src/suites/cts/validation_test.ts
```diff
@@ -0,0 +1,233 @@
...
+    this.sampler = this.device.createSampler();
+
+    this.sampledTexture = this.device.createTexture({
+      size: { width: 16, height: 16, depth: 1 },
+      format: 'rgba8unorm',
+      usage: GPUTextureUsage.SAMPLED,
+    });
+  }
+
+  getErrorBuffer(): GPUBuffer {
```
Kangz: Helpers like this for each object will be very useful. :+1:

src/suites/cts/validation_test.ts
```diff
@@ -0,0 +1,233 @@
...
+export class ValidationTest extends GPUTest {
+  uniformBuffer: GPUBuffer = undefined!;
+  storageBuffer: GPUBuffer = undefined!;
+  sampler: GPUSampler = undefined!;
+  sampledTexture: GPUTexture = undefined!;
+
+  async init(): Promise<void> {
+    await super.init();
+
+    this.uniformBuffer = this.device.createBuffer({
```
Kangz: These default values are somewhat arbitrary so I don't think they should be in the fixture for all of the validation tests. Instead I suggest adding them to the fixture for bind group layouts specifically.

src/suites/cts/validation_test.ts
```diff
@@ -0,0 +1,233 @@
...
+  expectUncapturedError(commandBuffer: GPUCommandBuffer[] = []): Promise<void> {
+    this.queue.submit(commandBuffer);
+    return this.asyncExpectation(async () => {
+      await new Promise(resolve => {
+        this.device.addEventListener('uncapturederror', resolve, { once: true });
+      });
+    });
+  }
+
+  async expectBindGroupValidationErrors(
```
Kangz: Why isn't this in the fixture for bind group layout tests?

Kangz: Ditto for other functions below.

src/suites/cts/validation.spec.ts
```diff
@@ -0,0 +1,673 @@
+export const description = `
+Validation tests.
```
Kangz: There are going to be a LOT of validation tests. I suggest splitting them per object tested, or even better per function tests. (in this case mostly createBindGroupLayout and createBindGroupLayout).

kainino0x: Yes, definitely need to split up this file. I'm not sure what directory structure is best: validation/.../createBindGroup or .../createBindGroup or .../createBindGroup/validation. But I'll deal with that later; pick whatever you want.

src/suites/cts/validation_test.ts
```diff
@@ -0,0 +1,233 @@
...
+    });
+    const pipeline = this.device.createComputePipeline({
+      computeStage: { module, entryPoint: 'main' },
+      layout: pipelineLayout,
+    });
+    const encoder = this.device.createCommandEncoder();
+    const computePass = encoder.beginComputePass();
+    computePass.setPipeline(pipeline);
+
+    //@ts-ignore TODO:Update @webgpu/types
```
kainino0x: Should be possible to use `as any` and use `// tslint:disable-next-line: no-any` instead

tslint.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "gts/tslint.json",
   "rules": {
+    "ban-ts-ignore": false,
```
kainino0x: shouldn't be needed with `as any`

src/suites/cts/validation.spec.ts
```diff
@@ -0,0 +1,673 @@
...
+  await t.expectUncapturedError();
+});
+
+g.test('validation/binding index too high in bind group', async t => {
+  const bindingIndexTooHigh = 16;
+
+  const bindGroupLayout = t.device.createBindGroupLayout({
+    bindings: [
+      {
+        binding: 0,
```
kainino0x: Since the layout has binding 0 and the bind group has binding 16, this would fail even if 16 was valid. Change this to bindingIndexTooHigh?

src/suites/cts/validation.spec.ts
```diff
@@ -0,0 +1,673 @@
...
+        },
+      },
+    ],
+    layout: bindGroupLayout,
+  });
+
+  await t.expectUncapturedError();
+});
+
+g.test('validation/storage buffer binding must contain exactly one storage buffer', async t => {
```
kainino0x: Can you add a TODO for me above this: `TODO(kainino0x): Look into whether it's possible to parameterize this set of tests.`

src/suites/cts/validation.spec.ts
```diff
@@ -0,0 +1,673 @@
...
+Validation tests.
+`;
+
+import { TestGroup } from '../../framework/index.js';
+
+import { ValidationTest } from './validation_test.js';
+
+export const g = new TestGroup(ValidationTest);
+
+g.test('validation/binding count mismatch', async t => {
```
kainino0x: no need for `validation/` at the beginning, since this file is already `validation.spec.ts`

src/suites/cts/validation.spec.ts
```diff
@@ -0,0 +1,673 @@
...
+      size: 256,
+    },
+    {
+      offset: 128,
+      size: 256,
+    },
+    {
+      offset: 255,
+      size: 256,
+    },
```
kainino0x: This one should be easy to parameterize, something like this: https://github.com/gpuweb/cts/blob/master/src/suites/cts/examples.spec.ts#L60

src/suites/cts/validation.spec.ts
```diff
@@ -0,0 +1,673 @@
...
+      offset: 0,
+      size: 256 * 5,
+    },
+    {
+      offset: 1024,
+      size: 1,
+    },
+    {
+      offset: 1,
+      size: undefined /* Whole size */,
```
kainino0x: According to the spec, this is supposed to be valid.

src/suites/cts/validation.spec.ts
```diff
@@ -0,0 +1,673 @@
...
+        visibility: GPUShaderStage.COMPUTE,
+        type: 'storage-buffer',
+      },
+    ],
+  });
+
+  t.device.createBindGroup({
+    bindings: [
+      {
+        binding: 0,
```
kainino0x: Same as above; use bindingIndexTooHigh here?

src/suites/cts/validation.spec.ts
```diff
@@ -0,0 +1,673 @@
...
+        },
+      },
+    ],
+    layout: bindGroupLayout,
+  });
+
+  await t.expectUncapturedError();
+});
+
+g.test('validation/dynamic set to true is not allowed for sampled texture', async t => {
```
kainino0x: should be parameterizable too

src/suites/cts/validation.spec.ts
```diff
@@ -0,0 +1,673 @@
...
+  const maxUniformBufferBindGroupLayout = t.device.createBindGroupLayout({
+    bindings: uniformBufferBindings,
+  });
+
+  const maxStorageBufferBindGroupLayout = t.device.createBindGroupLayout({
+    bindings: storageBufferBindings,
+  });
+
+  {
+    // Check dynamic uniform buffers excedd maximum in pipeline layout.
```
kainino0x: typo: exceed

src/suites/cts/validation.spec.ts
```diff
@@ -0,0 +1,673 @@
...
+  async t => {
+    const bindGroupLayout = t.device.createBindGroupLayout({
+      bindings: [
+        {
+          binding: 0,
+          visibility: GPUShaderStage.NONE,
+          type: 'storage-buffer',
+        },
+      ],
+    });
```
kainino0x: Shouldn't the bind group layout itself be invalid? I don't think you need to create a bind group to test this.

src/suites/cts/validation_test.ts
kainino0x: > this can be checked by wrapping the whole test in an error scope

(which we should do, for all tests)

src/suites/cts/validation_test.ts
kainino0x: Maybe these don't even need to be on the fixture - could probably just have standalone helper functions in the bind group test file that create new buffers of various types

src/suites/cts/validation_test.ts
beaufortfrancois: It's not as it is a TypeScript error. See https://github.com/Microsoft/vscode-tslint/issues/368#issuecomment-406175162
Hopefully with updated @webgpu/types, this won't be needed anymore.

tslint.json
beaufortfrancois: See https://github.com/gpuweb/cts/pull/25#discussion_r322641297

src/suites/cts/validation.spec.ts
beaufortfrancois: Done.

beaufortfrancois:
@Kangz @kainino0x I've reset this PR to include only bind group layout validation tests. Once we're good with the approach, I'll push other validation tests you can review one by one.
Hopefully https://github.com/gpuweb/cts/compare/validation looks better now.

src/suites/cts/validation/create_bind_group_layout.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+      {
+        binding: 1,
+        visibility: GPUShaderStage.COMPUTE,
+        type: 'storage-buffer',
+      },
+    ],
+  };
+
+  const badDescriptor = clone(goodDescriptor);
+  badDescriptor.bindings![1].binding = 0;
```
Kangz: What's that ! for?

src/suites/cts/validation/create_bind_group_layout.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+
+import { ValidationTest } from './validation_test.js';
+
+function clone(descriptor: GPUBindGroupLayoutDescriptor): GPUBindGroupLayoutDescriptor {
+  return JSON.parse(JSON.stringify(descriptor));
+}
+
+export const g = new TestGroup(ValidationTest);
+
+g.test('some binding index was specified more than once', async t => {
```
Kangz: This is a long name, which could make suppression strings super long, but also helps know what the test does. @kainino0x WDYT?

src/suites/cts/validation/create_bind_group_layout.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+  // Control case
+  t.device.createBindGroupLayout(goodDescriptor);
+
+  // Binding index 16 can't be specified.
+  await t.expectValidationError(() => {
+    t.device.createBindGroupLayout(badDescriptor);
+  });
+});
+
+g.test('Visibility of bindings cannot be None', async t => {
```
Kangz: Note that this is still under discussion in the spec repo.

src/suites/cts/validation_test.ts
kainino0x: AFAICT, `as any` should change it from a ts error to a tslint error, then you can suppress the latter. I do this in some other places too (just grep for `no-any`)

src/suites/cts/validation/create_bind_group_layout.spec.ts
kainino0x: it overrides the type of a nullable thing to be not nullable:
https://stackoverflow.com/q/42273853

src/suites/cts/validation/create_bind_group_layout.spec.ts
kainino0x: I think it's fine, we can shorten them if it becomes a pain

src/suites/cts/validation/create_bind_group_layout.spec.ts
kainino0x: OK to land the change though, we can just update the CTS once it's decided. Maybe leave a TODO.

src/suites/cts/validation/validation_test.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    fn();
+
+    // TODO: Remove queue submission when not needed anymore.
+    this.queue.submit([]);
+
+    const gpuValidationError = await this.device.popErrorScope();
+    if (!gpuValidationError) {
+      this.fail('Validation error was expected.');
+    }
```
kainino0x: Wrap this section in `return this.asyncExpectation(async () => { ... });` so the numOutstandingAsyncExpectations gets checked.

src/suites/cts/validation/create_bind_group_layout.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+  // Control case
+  t.device.createBindGroupLayout(goodDescriptor);
+
+  // Binding index 0 can't be specified twice.
+  await t.expectValidationError(() => {
+    t.device.createBindGroupLayout(badDescriptor);
+  });
+});
+
+g.test('some binding index exceeds the maximum value', async t => {
```
kainino0x: I think this may also change in spec, but we'll change the test when we get to it.

src/suites/cts/validation/validation_test.ts
```diff
@@ -0,0 +1,35 @@
+import { GPUTest } from '../gpu_test.js';
+
+export class ValidationTest extends GPUTest {
+  async init(): Promise<void> {
+    await super.init();
+
+    this.device.pushErrorScope('validation');
+  }
+
+  async finalize(): Promise<void> {
```
kainino0x: Can you move this override to gpu_test instead? It shouldn't apply just to validation tests.

src/suites/cts/validation/create_bind_group_layout.spec.ts
```diff
@@ -0,0 +1,87 @@
+export const description = `
+Create bind group layout validation tests.
```
kainino0x: I'd use "createBindGroupLayout" both here and in the filename, because that's the name of the function being tested.

src/suites/cts/validation/create_bind_group_layout.spec.ts
beaufortfrancois: Done.

src/suites/cts/validation/create_bind_group_layout.spec.ts
beaufortfrancois: Done.

src/suites/cts/validation/create_bind_group_layout.spec.ts
beaufortfrancois: Which issue should I link to for the TODO?

src/suites/cts/validation/validation_test.ts
beaufortfrancois: Done.

src/suites/cts/validation/validation_test.ts
beaufortfrancois: I've moved it to gpu_test and handled case where they were no error scopes to pop. Let me know if that looks good to you.

src/suites/cts/validation.spec.ts
beaufortfrancois: @kainino0x Good catch!
Should we even have a test then? Since we already have one for createBindGroupLayout that tests bindingIndexTooHigh.

beaufortfrancois:
@kainino0x @Kangz I've addressed your feedback, asked clarification and started adding more tests (`createBindGroup` and `createPipelineLayout`)

src/suites/cts/gpu_test.ts
```diff
@@ -14,6 +14,24 @@ export class GPUTest extends Fixture {
     this.queue = this.device.getQueue();
   }
 
+  async finalize(): Promise<void> {
```
Kangz: Why is this in GPUTest? It seems to go with the init of ValidationTest so it could go in a finalize there no?

beaufortfrancois: See https://github.com/gpuweb/cts/pull/25#discussion_r322960625

src/suites/cts/validation/createBindGroupLayout.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+      },
+      {
+        binding: 1,
+        visibility: GPUShaderStage.COMPUTE,
+        type: 'storage-buffer',
+      },
+    ],
+  };
+
+  const badDescriptor = clone(goodDescriptor);
```
Kangz: unit: I'd put the bad descriptor just after comment at [1]

src/suites/cts/validation/createBindGroupLayout.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+  badDescriptor.bindings![1].binding = 0;
+
+  // Control case
+  t.device.createBindGroupLayout(goodDescriptor);
+
+  // Binding index 0 can't be specified twice.
```
Kangz: [1]

src/suites/cts/validation/createBindGroupLayout.spec.ts
Kangz: Ditto for other tests.

src/suites/cts/validation/createBindGroupLayout.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+  await t.expectValidationError(() => {
+    t.device.createBindGroupLayout(badDescriptor);
+  });
+});
+
+g.test('some binding index exceeds the maximum value', async t => {
```
Kangz: This is a constraint of Dawn but isn't a validation error in Justin's BindGroupLayout PR.

src/suites/cts/validation/createBindGroup.spec.ts
```diff
@@ -0,0 +1,491 @@
...
+
+  getStorageTexture(): GPUTexture {
+    return this.device.createTexture({
+      size: { width: 16, height: 16, depth: 1 },
+      format: 'rgba8unorm',
+      usage: GPUTextureUsage.STORAGE,
+    });
+  }
+
+  async getErrorBuffer(): Promise<GPUBuffer> {
```
Kangz: nit: this will be useful for many tests and all error buffers are idempotent so it could go on ValidationTest.

Kangz:
Tests look very good and readable!

src/suites/cts/validation/setBindGroup.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+      bindGroupLayouts: [bindGroupLayout],
+    });
+
+    return { bindGroup, pipelineLayout };
+  }
+
+  testComputePass(dynamicOffsets: number[]): void {
+    const { bindGroup, pipelineLayout } = this.setup();
+
+    const module = this.device.createShaderModule({
```
Kangz: A compute pass with just a setBindGroup should still fail so the module isn't needed here (nor is the pipeline and dispatch call)

src/suites/cts/validation/setBindGroup.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+    computePass.dispatch(1, 1, 1);
+    computePass.endPass();
+
+    encoder.finish();
+  }
+
+  testRenderPass(dynamicOffsets: number[]): void {
+    const { bindGroup, pipelineLayout } = this.setup();
+
+    const vertexModule = this.device.createShaderModule({
```
Kangz: ditto

src/suites/cts/validation/setBindGroup.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+      usage: GPUBufferUsage.STORAGE,
+    });
+
+    const bindGroup = this.device.createBindGroup({
+      bindings: [
+        {
+          binding: 0,
+          resource: {
+            buffer: uniformBuffer,
+            size: BINDING_SIZE,
```
Kangz: It would be good to have one of these two at an offset so we know the computations are correctly using offset + dynamicOffset.

src/suites/cts/validation/setBindGroup.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+      t.testRenderPass(dynamicOffsets);
+    });
+  }
+}).params([
+  { dynamicOffsets: [256, 0], success: true }, // Dynamic offsets aligned
+  { dynamicOffsets: [], success: false }, // No dynamic offsets
+  { dynamicOffsets: [1, 2], success: false }, // Dynamic offsets not aligned
+  { dynamicOffsets: [1024, 0], success: false }, // Dynamic uniform buffer out of bounds
+  { dynamicOffsets: [0, 1024], success: false }, // Dynamic storage buffer out of bounds
+  { dynamicOffsets: [512, 0], success: false }, // Dynamic uniform buffer out of bounds because of binding size
```
Kangz: It would be nice to test with binding size 8 and binding size 9 to know the exact boundary.

Also we should have tests for overflows though I'm not sure how to pass UINT64_MAX through JS.

src/suites/cts/validation/setBindGroup.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+      t.testComputePass(dynamicOffsets);
+    });
+    // Dynamic offsets don't match expectations in render pass.
+    await t.expectValidationError(() => {
+      t.testRenderPass(dynamicOffsets);
+    });
+  }
+}).params([
+  { dynamicOffsets: [256, 0], success: true }, // Dynamic offsets aligned
+  { dynamicOffsets: [], success: false }, // No dynamic offsets
```
Kangz: Can we have a test with only one dynamic offset, and with 3?

Also a test with a bindgroup with no dynamic offsets and give a non-empty list.

src/suites/cts/validation/setBindGroup.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+}
+
+export const g = new TestGroup(F);
+
+g.test('dynamic offsets match expectations in pass encoder', async t => {
+  const { dynamicOffsets, success } = t.params;
+
+  if (success) {
+    // Control case
+    t.testComputePass(dynamicOffsets);
```
Kangz: Also render bundle?

src/suites/cts/validation/setBindGroup.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+setBindGroup validation tests.
+`;
+
+import { TestGroup } from '../../../framework/index.js';
+import GLSL from '../../../tools/glsl.macro.js';
+
+import { ValidationTest } from './validation_test.js';
+
+export class F extends ValidationTest {
+  setup(): { bindGroup: GPUBindGroup; pipelineLayout: GPUPipelineLayout } {
```
Kangz: SetBindGroup also needs a test where we pass an error bindgroup. And a test where we pass the wrong type?

src/suites/cts/validation/createPipelineLayout.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+import { TestGroup } from '../../../framework/index.js';
+
+import { ValidationTest } from './validation_test.js';
+
+function clone(descriptor: GPUBindGroupLayoutDescriptor): GPUBindGroupLayoutDescriptor {
+  return JSON.parse(JSON.stringify(descriptor));
+}
+
+export const g = new TestGroup(ValidationTest);
+
```
Kangz: Can we also have a test 4 BGLs being a success and 5BGLs being an error?

Kangz:
LGTM, can we land this one so other PRs can build on top? Also maybe this could only keep the createPipelineLayout file so the PR is more about the harness changes.

src/suites/cts/validation/setBindGroup.spec.ts
beaufortfrancois: Done.

src/suites/cts/validation/create_bind_group_layout.spec.ts
kainino0x: Not that important, could do a TODO without issue or just leave it out. We'll easily find the test when we need to fix it.

src/suites/cts/validation/setBindGroup.spec.ts
kainino0x: If you mean passing like a GPUBuffer instead of a GPUBindGroup, I think we should do all of the tests for incorrect types in one file later.

src/suites/cts/validation/createPipelineLayout.spec.ts
kainino0x: Added this one and several others as TODOs. @beaufortfrancois please feel free to take those on.

src/suites/cts/validation/createBindGroup.spec.ts
```diff
@@ -0,0 +1,491 @@
...
+    matchedResource = t.getSampledTexture().createView();
+  } else if (type === 'storage-texture') {
+    matchedResource = t.getStorageTexture();
+  } else {
+    throw new Error('Unexpected binding type');
+  }
+
+  async function* mismatchedResources(): AsyncIterable<GPUBindingResource> {
+    const errorBuffer = await t.getErrorBuffer();
+    yield { buffer: errorBuffer };
```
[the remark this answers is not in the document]
kainino0x: Well, unfortunately I had to remove it here, in order to split the test into more cases. Because we don't implement storage-texture yet, it was causing all of the tests to fail and losing test coverage on everything else.

kainino0x:
Ended up being more changes than expected. Feel free to take a look at the last 3 commits. Merging now.

src/suites/cts/validation/setBindGroup.spec.ts
kainino0x: I just realized I was wrong about this:
https://heycam.github.io/webidl/#abstract-opdef-converttoint

JS numbers larger than 2^53 can be passed in, it's just impossible to represent some of those numbers in JS.

I think that the spec should add `[EnforceRange]` here so that numbers above MAX_SAFE_INTEGER are not allowed. We should test those in the same section as all of the other IDL-enforced type checks, separately from validation (which I think should be only for the kind of error caught by error scopes).

PR #26, opened 2019-09-13
Add commandBuffer validation tests

beaufortfrancois:
This PR is about adding some commandBuffer validation tests.
It is based on https://github.com/gpuweb/cts/pull/25/

FYI @Kangz @austinEng

### DIFF
--- a/src/suites/cts/validation/commandBuffer.spec.ts
+++ b/src/suites/cts/validation/commandBuffer.spec.ts
@@ -1,0 +1,281 @@
+export const description = `
+commandBuffer validation tests.
+`;
+
+import { TestGroup } from '../../../framework/index.js';
+
+import { ValidationTest } from './validation_test.js';
+
+export class F extends ValidationTest {
+  beginPass(
+    type: string,
+    commandEncoder: GPUCommandEncoder
+  ): GPURenderPassEncoder | GPUComputePassEncoder {
+    if (type === 'render') {
+      return this.beginRenderPass(commandEncoder);
+    } else if (type === 'compute') {
+      return this.beginComputePass(commandEncoder);
+    } else {
+      throw new Error('Unexpected pass encoder type');
+    }
+  }
+
+  beginRenderPass(commandEncoder: GPUCommandEncoder): GPURenderPassEncoder {
+    const attachmentTexture = this.device.createTexture({
+      format: 'rgba8unorm',
+      size: { width: 16, height: 16, depth: 1 },
+      usage: GPUTextureUsage.OUTPUT_ATTACHMENT,
+    });
+
+    return commandEncoder.beginRenderPass({
+      colorAttachments: [
+        {
+          attachment: attachmentTexture.createView(),
+          loadValue: { r: 1.0, g: 0.0, b: 0.0, a: 1.0 },
+        },
+      ],
+    });
+  }
+
+  beginComputePass(commandEncoder: GPUCommandEncoder): GPUComputePassEncoder {
+    return commandEncoder.beginComputePass();
+  }
+}
+
+export const g = new TestGroup(F);
+
+g.test('basic', async t => {
+  const commandEncoder = t.device.createCommandEncoder();
+  commandEncoder.finish();
+});
+
+g.test('command buffer cannot be ended mid pass', async t => {
+  const { type } = t.params;
+
+  {
+    // Control case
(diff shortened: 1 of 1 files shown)


src/suites/cts/validation/commandBuffer.spec.ts
```diff
@@ -0,0 +1,281 @@
...
+    await t.expectValidationError(async () => {
+      const commandEncoder = t.device.createCommandEncoder();
+      const pass = t.beginPass(type, commandEncoder);
+      commandEncoder.finish();
+
+      await t.expectValidationError(() => {
+        pass.endPass();
+      });
+    });
+  }
```
kainino0x: I think it would be better to split these 3 into separate tests. The only annoying bit of this is they're separately parameterized, but fortunately in this case the parameters are very short. (and can be slightly shorter with `poptions('type', ['render', 'compute'])`). You can also store that params list off into a variable and reuse it, if you want.

src/suites/cts/validation/commandBuffer.spec.ts
```diff
@@ -0,0 +1,281 @@
...
+  {
+    // Control case
+    const commandEncoder = t.device.createCommandEncoder();
+    const pass = t.beginPass(type, commandEncoder);
+    pass.endPass();
+    commandEncoder.finish();
+  }
+  {
+    // Pass ended twice
+    await t.expectValidationError(() => {
```
kainino0x: nit: same, include only finish

src/suites/cts/validation/commandBuffer.spec.ts
```diff
@@ -0,0 +1,281 @@
...
+  {
+    // Control case
+    const commandEncoder = t.device.createCommandEncoder();
+    const pass = t.beginPass(type, commandEncoder);
+    pass.endPass();
+    commandEncoder.finish();
+  }
+  {
+    // Command buffer ended mid-pass.
+    await t.expectValidationError(() => {
```
kainino0x: nit: like in the tests below, move this so that the expectvalidationError only encloses the finish() call.

src/suites/cts/validation/commandBuffer.spec.ts
```diff
@@ -0,0 +1,281 @@
...
+    });
+  }
+  {
+    // Trying to use encoder pass after finish() should fail too
+    await t.expectValidationError(async () => {
+      const commandEncoder = t.device.createCommandEncoder();
+      const pass = t.beginPass(type, commandEncoder);
+      commandEncoder.finish();
+
+      await t.expectValidationError(() => {
```
kainino0x: Un-nest these 2 expectValidationError calls and scope them to only include the exact call that's supposed to error.

src/suites/cts/validation/commandBuffer.spec.ts
```diff
@@ -0,0 +1,281 @@
...
+  const errorBuffer = await t.getErrorBuffer();
+
+  // Failed finish() is called before encoding command
+  const commandEncoder = t.device.createCommandEncoder();
+  commandEncoder.copyBufferToBuffer(errorBuffer, 0, errorBuffer, 0, 16);
+
+  await t.expectValidationError(async () => {
+    commandEncoder.finish();
+
+    await t.expectValidationError(() => {
```
kainino0x: ditto, un-nest

src/suites/cts/validation/commandBuffer.spec.ts
beaufortfrancois: Done

src/suites/cts/validation/commandBuffer.spec.ts
beaufortfrancois: Done.

src/suites/cts/validation/commandBuffer.spec.ts
beaufortfrancois: Done.

src/suites/cts/validation/commandBuffer.spec.ts
beaufortfrancois: Done

src/suites/cts/validation/commandBuffer.spec.ts
beaufortfrancois: Done

beaufortfrancois:
@kainino0x I've addressed all your feedback. Thank you!

PR #27, opened 2019-09-13
Add indirect validation tests

beaufortfrancois:
This PR is about adding some compute/render indirect validation tests.
It is based on https://github.com/gpuweb/cts/pull/25/

Note that it requires https://github.com/kainino0x/-webgpu-types/issues/2 to be fixed. I'll update package.json file when it's done.

FYI @Kangz @austinEng

### DIFF
--- a/src/suites/cts/validation/dispatchIndirect.spec.ts
+++ b/src/suites/cts/validation/dispatchIndirect.spec.ts
@@ -3,61 +3,55 @@ dispatchIndirect validation tests.
 `;
 
 import { TestGroup } from '../../../framework/index.js';
-
-import { ValidationTest } from './validation_test.js';
 import GLSL from '../../../tools/glsl.macro.js';
 
-export class F extends ValidationTest {
-  async testIndirectOffset(data: number[], offset: number, success: boolean): Promise<void> {
-    const module = this.device.createShaderModule({
-      code: GLSL(
-        'compute',
-        `#version 450
-          void main() {
-          }
-        `
-      ),
-    });
+import { ValidationTest } from './validation_test.js';
 
-    const pipelineLayout = this.device.createPipelineLayout({
-      bindGroupLayouts: [],
-    });
-
-    const pipeline = this.device.createComputePipeline({
-      computeStage: { module, entryPoint: 'main' },
-      layout: pipelineLayout,
-    });
-
-    const [indirectBuffer, arrayBuffer] = await this.device.createBufferMappedAsync({
-      size: data.length * Uint32Array.BYTES_PER_ELEMENT,
-      usage: GPUBufferUsage.INDIRECT,
-    });
-    new Uint32Array(arrayBuffer).set(data);
-
-    const commandEncoder = this.device.createCommandEncoder();
-    const computePass = commandEncoder.beginComputePass();
-    computePass.setPipeline(pipeline);
-    computePass.dispatchIndirect(indirectBuffer, offset * Uint32Array.BYTES_PER_ELEMENT);
-    computePass.endPass();
-
-    if (success) {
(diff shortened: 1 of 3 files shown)


src/suites/cts/validation/dispatchIndirect.spec.ts
```diff
@@ -0,0 +1,69 @@
+export const description = `
+dispatchIndirect validation tests.
+`;
+
+import { TestGroup } from '../../../framework/index.js';
+
+import { ValidationTest } from './validation_test.js';
+import GLSL from '../../../tools/glsl.macro.js';
+
+export class F extends ValidationTest {
```
kainino0x: nit: doesn't need export

src/suites/cts/validation/drawIndirect.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  await t.testIndirectOffset(data, offset, false /* indexed */, success);
+}).params([
+  { data: [1, 2, 3, 4], offset: 0, success: true }, // In bounds
+  { data: [1, 2, 3, 4, 5, 6, 7], offset: 0, success: true }, // In bounds, bigger buffer
+  { data: [1, 2, 3, 4, 5, 6, 7, 8], offset: 4, success: true }, // In bounds, bigger buffer, positive offset
+  { data: [1, 2, 3], offset: 0, success: false }, // Out of bounds, buffer too small
+  { data: [1, 2, 3, 4], offset: 1, success: false }, // Out of bounds, index too big
+  { data: [1, 2, 3, 4], offset: 5, success: false }, // Out of bounds, past buffer
+  { data: [1, 2, 3, 4, 5, 6, 7], offset: Number.MAX_SAFE_INTEGER, success: false }, // In bounds,index + size of command overflows
```
kainino0x: nit: ~in~ **out of** bounds?
nit: space after comma
nit: it doesn't really overflow anything, probably better to say "is very large"

src/suites/cts/validation/drawIndirect.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+    }
+  }
+}
+
+export const g = new TestGroup(F);
+
+g.test('out of bounds indirect draw calls are caught early', async t => {
+  const { data, offset, success } = t.params;
+
+  await t.testIndirectOffset(data, offset, false /* indexed */, success);
```
kainino0x: Since this helper is only used in the 2 places, I think it would be better to do the following so we can keep the code local / avoid polluting the fixture.

- make indexed one of the members in the params lists
- inline the contents of the helper here

kainino0x: If you want, you can avoid including 'indexed' on every line by doing this:

```typescript
.params([
  ...pcombine([[{ indexed: false }], [
    // cases
  ]]),
  ...pcombine([[{ indexed: true }], [
    // cases
  ]]));
```

src/suites/cts/validation/dispatchIndirect.spec.ts
beaufortfrancois: Done.

src/suites/cts/validation/drawIndirect.spec.ts
beaufortfrancois: Done.

src/suites/cts/validation/drawIndirect.spec.ts
beaufortfrancois: I've inlined helper and used `pcombine`. Thanks~

beaufortfrancois:
@kainino0x I've addressed your feedback. Thanks!

PR #28, opened 2019-09-13
Improve exception logging

beaufortfrancois:
This nit allows us to learn more when an exception is thrown. For info, I've used this when running CTS in Safari.

image

### DIFF
--- a/src/framework/logger.ts
+++ b/src/framework/logger.ts
@@ -104,8 +104,6 @@ export class TestCaseRecorder {
 
   threw(e: Error): void {
     this.failed = true;
-    let m = 'EXCEPTION';
-    m += ' ' + e.name + ' ' + e.message + ' ' + getStackTrace(e);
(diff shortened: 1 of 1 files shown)


src/framework/logger.ts
```diff
@@ -105,7 +105,7 @@ export class TestCaseRecorder {
   threw(e: Error): void {
     this.failed = true;
     let m = 'EXCEPTION';
-    m += ' ' + getStackTrace(e);
+    m += ' ' + e.name + ' ' + e.message + ' ' + getStackTrace(e);
```
Kangz: unit: newline between the message and the stacktrace?

kainino0x: Done

kainino0x:
Unfortunately in chrome this is kind of redundant (stack already includes this info), but it's needed in firefox and safari. Updated the format it prints in.

PR #36, opened 2019-09-16
Add fences validation tests

beaufortfrancois:
This PR is about adding some validation tests for fences.

Note that this raised some issue in Chromium: https://bugs.chromium.org/p/chromium/issues/detail?id=1004368

FYI @austinEng @Kangz

### DIFF
--- a/src/suites/cts/validation/fences.spec.ts
+++ b/src/suites/cts/validation/fences.spec.ts
@@ -8,50 +8,36 @@ import { ValidationTest } from './validation_test.js';
 
 export const g = new TestGroup(ValidationTest);
 
-g.test('fence creation succeeds', t => {
-  t.queue.createFence();
-});
+// TODO: Remove if https://github.com/gpuweb/gpuweb/issues/377 is decided
+g.test('wait on a fence without signaling the value is invalid', async t => {
+  const fence = t.queue.createFence();
+  const promise = fence.onCompletion(2);
 
-g.test('fence starts at initial value', t => {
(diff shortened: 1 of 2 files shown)


kainino0x:
There are both validation and functionality tests in here, can we split them out? In particular, I think some of the basic cases are already tested in https://github.com/gpuweb/cts/blob/master/src/suites/cts/fences.spec.ts so let's make sure they're not duplicated.

beaufortfrancois:
Ooops. I didn't notice `fences.spec.ts` file.
I'll update this PR.

beaufortfrancois:
@kainino0x I've split functionality and validation tests. [LMKWYT](https://www.urbandictionary.com/define.php?term=lmkwyt)

src/suites/cts/validation/fences.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+  const fence = t.queue.createFence();
+
+  t.queue.signal(fence, 2);
+  await fence.onCompletion(2);
+
+  t.queue.signal(fence, 6);
+  await fence.onCompletion(6);
+});
+
+g.test('signal a fence on a different queue than it was created on is invalid', async t => {
```
kainino0x: nit: for these two tests, say "different device" instead of "different queue" because eventually there will be multiple queues on one device.

kainino0x: I'll just do this and land it.

PR #37, opened 2019-09-16
Add queue submit validation tests

beaufortfrancois:
This PR is about adding some validation tests for queue submit.

FYI @austinEng @Kangz

### DIFF
--- a/src/suites/cts/validation/queue_submit.spec.ts
+++ b/src/suites/cts/validation/queue_submit.spec.ts
@@ -9,34 +9,34 @@ import { ValidationTest } from './validation_test.js';
 export const g = new TestGroup(ValidationTest);
 
 g.test('submitting with a mapped buffer is disallowed', async t => {
-  // Create a map-write buffer
   const buffer = t.device.createBuffer({
     size: 4,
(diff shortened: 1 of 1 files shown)


src/suites/cts/validation/queue_submit.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+  const targetBuffer = t.device.createBuffer({
+    size: 4,
+    usage: GPUBufferUsage.COPY_DST,
+  });
+
+  const commandEncoder = t.device.createCommandEncoder();
+  commandEncoder.copyBufferToBuffer(buffer, 0, targetBuffer, 0, 4);
+
+  // Submitting when the buffer has never been mapped should succeed
+  const commandBuffer = commandEncoder.finish();
```
kainino0x: This command buffer gets reused 3 times, which is not supposed to be valid (though I think it's currently valid in chrome). Need e.g. a helper to re-create the command buffer.

beaufortfrancois: Done.

beaufortfrancois:
@kainino0x Feedback addressed.

PR #38, opened 2019-09-17
Add error scope validation tests

beaufortfrancois:
This PR is adding some error scope validation tests.

Note that I had to create a class that inherits from Fixture because push/pop error scopes were interfering with my tests.

FYI @Kangz @austinEng

### DIFF
--- a/src/suites/cts/validation/error_scope.spec.ts
+++ b/src/suites/cts/validation/error_scope.spec.ts
@@ -1,0 +1,106 @@
+export const description = `
+error scope validation tests.
+`;
+
+import { getGPU } from '../../../framework/gpu/implementation.js';
+import { Fixture, TestGroup } from '../../../framework/index.js';
+
+class F extends Fixture {
+  device: GPUDevice = undefined!;
+
+  async init(): Promise<void> {
+    super.init();
+    const gpu = getGPU();
+    const adapter = await gpu.requestAdapter();
+    this.device = await adapter.requestDevice();
+  }
+
+  createErrorBuffer(): void {
+    this.device.createBuffer({
+      size: 1024,
+      usage: 0xffff, // Invalid GPUBufferUsage
+    });
+  }
+
+  async expectUncapturedError(fn: Function): Promise<void> {
+    return this.asyncExpectation(async () => {
+      const promise = new Promise(resolve => {
+        this.device.addEventListener('uncapturederror', resolve, { once: true });
+      });
+
+      fn();
+
+      await promise;
+    });
+  }
+}
+
+export const g = new TestGroup(F);
+
+g.test('simple case where the error scope catches an error', async t => {
+  t.device.pushErrorScope('validation');
+
+  t.createErrorBuffer();
+
+  const error = await t.device.popErrorScope();
+  t.expect(error instanceof GPUValidationError);
+});
+
+g.test('errors bubble to the parent scope if not handled by the current scope', async t => {
+  t.device.pushErrorScope('validation');
+  t.device.pushErrorScope('out-of-memory');
+
+  t.createErrorBuffer();
+
+  {
+    const error = await t.device.popErrorScope();
+    t.expect(error === null);
+  }
+  {
+    const error = await t.device.popErrorScope();
+    t.expect(error instanceof GPUValidationError);
+  }
+});
+
+g.test('if an error scope matches an error it does not bubble to the parent scope', async t => {
+  t.device.pushErrorScope('out-of-memory');
+  t.device.pushErrorScope('validation');
+
+  t.createErrorBuffer();
+
+  {
+    const error = await t.device.popErrorScope();
+    t.expect(error instanceof GPUValidationError);
+  }
+  {
+    const error = await t.device.popErrorScope();
+    t.expect(error === null);
+  }
+});
+
+g.test('if no error scope handles an error it fires an uncapturederror event', async t => {
+  t.device.pushErrorScope('out-of-memory');
+
+  t.expectUncapturedError(() => {
+    t.createErrorBuffer();
+  });
+
+  const error = await t.device.popErrorScope();
+  t.expect(error === null);
+});
+
+g.test('push/popping error scopes must be balanced', async t => {
+  {
+    const promise = t.device.popErrorScope();
+    await t.shouldReject('OperationError', promise);
+  }
+  t.device.pushErrorScope('validation');
+  {
+    const error = await t.device.popErrorScope();
+    t.expect(error === null);
+  }
+  {
+    const promise = t.device.popErrorScope();
+    await t.shouldReject('OperationError', promise);
+  }
+});
--- a/src/suites/cts/validation/error_scope.spec.ts
+++ b/src/suites/cts/validation/error_scope.spec.ts
@@ -5,6 +5,14 @@ error scope validation tests.
 import { getGPU } from '../../../framework/gpu/implementation.js';
 import { Fixture, TestGroup } from '../../../framework/index.js';
 
+function rejectTimeout(ms: number, msg: string): Promise<void> {
+  return new Promise((resolve, reject) => {
+    setTimeout(() => {
+      reject(new Error(msg));
+    }, ms);
+  });
+}
+
 class F extends Fixture {
   device: GPUDevice = undefined!;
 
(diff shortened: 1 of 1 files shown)


src/suites/cts/validation/error_scope.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  createErrorBuffer(): void {
+    this.device.createBuffer({
+      size: 1024,
+      usage: 0xffff, // Invalid GPUBufferUsage
+    });
+  }
+
+  async expectUncapturedError(fn: Function): Promise<void> {
```
kainino0x: What happens if we expectUncapturedError but we don't get one? Does this cause a test timeout?

The WPT runner has the concept of test timeouts - our standalone runner would just never complete the test.

src/suites/cts/validation/error_scope.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+    t.expect(error === null);
+  }
+  {
+    const error = await t.device.popErrorScope();
+    t.expect(error instanceof GPUValidationError);
+  }
+});
+
+g.test('if an error scope matches an error it does not bubble to the parent scope', async t => {
+  t.device.pushErrorScope('out-of-memory');
```
kainino0x: An out-of-memory scope wouldn't catch this error anyway, even if it did bubble. We should also test two nested validation scopes.

src/suites/cts/validation/error_scope.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+  {
+    const error = await t.device.popErrorScope();
+    t.expect(error === null);
+  }
+});
+
+g.test('if no error scope handles an error it fires an uncapturederror event', async t => {
+  t.device.pushErrorScope('out-of-memory');
+
+  t.expectUncapturedError(() => {
```
kainino0x: This should require an await. And the `asyncExpectation` should be detecting the fact that we don't await it. Is it not catching this?

src/suites/cts/validation/error_scope.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+  t.device.pushErrorScope('validation');
+  {
+    const error = await t.device.popErrorScope();
+    t.expect(error === null);
+  }
+  {
+    const promise = t.device.popErrorScope();
+    await t.shouldReject('OperationError', promise);
+  }
+});
```
kainino0x: Should have a test that pushing a lot of error scopes (maybe 1000? 1,000,000?) works.

src/suites/cts/validation/error_scope.spec.ts
beaufortfrancois: Good catch! I've added a timeout handler.

src/suites/cts/validation/error_scope.spec.ts
beaufortfrancois: I've replaced 'out-of-memory' with 'validation' so that we test two nested validation scopes.

src/suites/cts/validation/error_scope.spec.ts
beaufortfrancois: It does require an await. Thanks!
`asyncExpectation` was not catching this error. How could it?

src/suites/cts/validation/error_scope.spec.ts
beaufortfrancois: I've updated this test to push 1000 error scopes.
1,000,000 was taking way too long to execute.

src/suites/cts/validation/error_scope.spec.ts
```diff
@@ -94,11 +112,15 @@ g.test('push/popping error scopes must be balanced', async t => {
...
   }
-  t.device.pushErrorScope('validation');
-  {
-    const error = await t.device.popErrorScope();
-    t.expect(error === null);
+
+  const promises = [];
+  for (let i = 0; i < 1000; i++) {
+    t.device.pushErrorScope('validation');
+    promises.push(t.device.popErrorScope());
```
kainino0x: This does 1000 sibling error scopes, can we add another test for 1000 nested error scopes?

src/suites/cts/validation/error_scope.spec.ts
kainino0x: :+1:

src/suites/cts/validation/error_scope.spec.ts
kainino0x: The purpose of asyncExpectation is that it tracks the numOutstandingAsyncExpectations, which doesn't decrement until after the asynchronous expectation function finishes. So if the test function (`async t => ...`) finishes before numOutstandingAsyncExpectations goes back down to 0, it's supposed to throw an error.

That said, Corentin asked that asyncExpectation change so that it just adds promises that the harness will wait on automatically at the end of the test, instead of asserting. That may be nicer.

src/suites/cts/validation/error_scope.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  async expectUncapturedError(fn: Function): Promise<void> {
+    // TODO: Make arbitrary timeout value a test runner variable
+    const TIMEOUT_IN_MS = 1000;
+
+    return this.asyncExpectation(async () => {
+      try {
+        const promise = new Promise((resolve, reject) => {
+          const timeoutId = window.setTimeout(() => {
```
kainino0x: nit: shouldn't use `window.`, since we may want this to work on workers too

src/suites/cts/validation/error_scope.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+          this.device.addEventListener('uncapturederror', onUncapturedError, { once: true });
+        });
+
+        fn();
+
+        await promise;
+      } catch (error) {
+        this.fail(error.message);
+      }
+    });
```
kainino0x: nice! Perhaps `Promise.race` would make this a little cleaner:

```typescript
function rejectTimeout(ms: number, msg: string) -> Promise<void> {
  return new Promise((resolve, reject) => {
    setTimeout(() => { reject(new Error(msg)); }, ms);
  });
}

...

  const promise = new Promise(resolve => {
    this.device.addEventListener('uncapturederror', resolve, { once: true });
  });
  await Promise.race([promise, rejectTimeout(TIMEOUT_IN_MS, 'Uncaptured error timeout occurred')]);
```

src/suites/cts/validation/error_scope.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    return this.asyncExpectation(async () => {
+      try {
+        const promise = new Promise((resolve, reject) => {
+          const timeoutId = window.setTimeout(() => {
+            reject(new Error('Uncaptured error timeout occured'));
+          }, TIMEOUT_IN_MS);
+
+          const onUncapturedError = () => {
+            window.clearTimeout(timeoutId);
```
kainino0x: nit: probably not strictly necessary, which helps below.

src/suites/cts/validation/error_scope.spec.ts
kainino0x: Actually, waiting on all async expectations at the end is not ideal for this particular case. We want to make sure that the uncapturederror is coming from the right place - so we want to make sure that we the uncapturederror expectation to pass or fail before doing any more work.

Technically, it would be best if we unregistered the uncapturederror handler if we reject on timeout, but omitting that can't cause false positives so it's OK.

src/suites/cts/validation/error_scope.spec.ts
beaufortfrancois: Done

src/suites/cts/validation/error_scope.spec.ts
beaufortfrancois: Noted.

src/suites/cts/validation/error_scope.spec.ts
beaufortfrancois: So much better! Thanks!

src/suites/cts/validation/error_scope.spec.ts
beaufortfrancois: Sure.

PR #42, opened 2019-09-19
Add setVertexBuffers validation tests

beaufortfrancois:
This PR adds setVertexBuffers validation tests.

~~It sits on top of https://github.com/gpuweb/cts/pull/43~~

FYI @Kangz @austinEng

### DIFF
--- a/src/suites/cts/validation/setVertexBuffers.spec.ts
+++ b/src/suites/cts/validation/setVertexBuffers.spec.ts
@@ -2,7 +2,7 @@ export const description = `
 setVertexBuffers validation tests.
 `;
 
-import { TestGroup } from '../../../framework/index.js';
+import { TestGroup, range } from '../../../framework/index.js';
 
 import { ValidationTest } from './validation_test.js';
 
@@ -29,29 +29,21 @@ class F extends ValidationTest {
         vertexBuffers: [
           {
             stride: 3 * 4,
-            attributeSet: [],
+            attributeSet: range(bufferCount, i => ({
+              format: 'float3',
+              shaderLocation: i,
+            })),
           },
         ],
       },
     };
-    for (let i = 0; i < bufferCount; i++) {
-      descriptor.vertexInput!.vertexBuffers![0].attributeSet.push({
-        format: 'float3',
-        shaderLocation: i,
-      });
-    }
     return this.device.createRenderPipeline(descriptor);
   }
 
   getVertexStage(bufferCount: number): GPUProgrammableStageDescriptor {
     const code = `
       #version 450
-      ${[...new Array(bufferCount)]
-        .map(
-          (_, i) => `
-      layout(location = ${i}) in vec3 a_position${i};`
-        )
-        .join('')}
+      ${range(bufferCount, i => `\nlayout(location = ${i}) in vec3 a_position${i};`).join('')}
       void main() {
         gl_Position = vec4(0.0, 0.0, 0.0, 1.0);
        }
@@ -134,7 +126,7 @@ g.test('vertex buffers inherit from previous pipeline', async t => {
   }
 });
 
-g.test('vertex buffers inherit between render passes', async t => {
+g.test('vertex buffers do not inherit between render passes', async t => {
   const pipeline1 = t.createRenderPipeline(1);
   const pipeline2 = t.createRenderPipeline(2);
 
--- a/src/suites/cts/validation/setVertexBuffers.spec.ts
+++ b/src/suites/cts/validation/setVertexBuffers.spec.ts
@@ -1,0 +1,184 @@
+export const description = `
+setVertexBuffers validation tests.
+`;
+
+import { TestGroup } from '../../../framework/index.js';
+
+import { ValidationTest } from './validation_test.js';
+
+class F extends ValidationTest {
+  async init(): Promise<void> {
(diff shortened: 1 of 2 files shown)


src/suites/cts/validation/setVertexBuffers.spec.ts
```diff
@@ -0,0 +1,184 @@
...
+setVertexBuffers validation tests.
+`;
+
+import { TestGroup } from '../../../framework/index.js';
+
+import { ValidationTest } from './validation_test.js';
+
+class F extends ValidationTest {
+  async init(): Promise<void> {
+    await Promise.all([super.init(), this.initGLSL()]);
```
kainino0x:
```suggestion
    return Promise.all([super.init(), this.initGLSL()]);
```

src/suites/cts/validation/setVertexBuffers.spec.ts
```diff
@@ -0,0 +1,184 @@
+export const description = `
+setVertexBuffers validation tests.
+`;
+
+import { TestGroup } from '../../../framework/index.js';
+
+import { ValidationTest } from './validation_test.js';
+
+class F extends ValidationTest {
+  async init(): Promise<void> {
```
kainino0x:
```suggestion
  init(): Promise<void> {
```

src/suites/cts/validation/setVertexBuffers.spec.ts
```diff
@@ -0,0 +1,184 @@
...
+      vertexStage: this.getVertexStage(bufferCount),
+      fragmentStage: this.getFragmentStage(),
+      layout: this.getPipelineLayout(),
+      primitiveTopology: 'triangle-list',
+      colorStates: [{ format: 'rgba8unorm' }],
+      vertexInput: {
+        vertexBuffers: [
+          {
+            stride: 3 * 4,
+            attributeSet: [],
```
kainino0x: if we had a helper in a common file somewhere:

```js
function range<T>(n: number, fn: (i: number): T) { return [...Array(n).keys()].map(fn); }
```

then we could use it naturally here:

```js
attributeSet: range(bufferCount, i => { shaderLocation: i, format: 'float3' }),
```

src/suites/cts/validation/setVertexBuffers.spec.ts
```diff
@@ -0,0 +1,184 @@
...
+        shaderLocation: i,
+      });
+    }
+    return this.device.createRenderPipeline(descriptor);
+  }
+
+  getVertexStage(bufferCount: number): GPUProgrammableStageDescriptor {
+    const code = `
+      #version 450
+      ${[...new Array(bufferCount)]
```
kainino0x: ... and then can use that helper here too.

src/suites/cts/validation/setVertexBuffers.spec.ts
```diff
@@ -0,0 +1,184 @@
...
+        .map(
+          (_, i) => `
```
kainino0x: I'd write the newline in this string as `\n` to make it fit on one line for readability.

src/suites/cts/validation/setVertexBuffers.spec.ts
```diff
@@ -0,0 +1,184 @@
...
+    renderPass.draw(3, 1, 0, 0);
+    renderPass.setPipeline(pipeline1);
+    renderPass.draw(3, 1, 0, 0);
+    renderPass.endPass();
+
+    commandEncoder.finish();
+  }
+});
+
+g.test('vertex buffers inherit between render passes', async t => {
```
kainino0x: should probably invert this description ('do not inherit')

src/suites/cts/validation/setVertexBuffers.spec.ts
beaufortfrancois: See below.

src/suites/cts/validation/setVertexBuffers.spec.ts
beaufortfrancois: This doesn't return `Promise<void>` anymore but `Promise<[void, void]>`. This is why I've used "await".

src/suites/cts/validation/setVertexBuffers.spec.ts
beaufortfrancois: Done

src/suites/cts/validation/setVertexBuffers.spec.ts
beaufortfrancois: Done.

src/suites/cts/validation/setVertexBuffers.spec.ts
beaufortfrancois: Done

src/suites/cts/validation/setVertexBuffers.spec.ts
beaufortfrancois: Note that I had to use `new Array` because `Calls to 'Array' are not allowed. tsstyle#array-constructor (ban)`

src/suites/cts/validation/setVertexBuffers.spec.ts
kainino0x: Oh, gotcha.

PR #44, opened 2019-09-19
Add vertexInput validation tests

beaufortfrancois:
This PR adds vertexInput validation tests.

~~It sits on top of https://github.com/gpuweb/cts/pull/43~~

Note that it currently crashes in Chrome. Fix is available at https://chromium-review.googlesource.com/c/chromium/src/+/1813025

FYI @Kangz @austinEng

### DIFF
--- a/src/suites/cts/validation/vertex_input.spec.ts
+++ b/src/suites/cts/validation/vertex_input.spec.ts
@@ -1,0 +1,619 @@
+export const description = `
+vertexInput validation tests.
+`;
+
+import { TestGroup } from '../../../framework/index.js';
+
+import { ValidationTest } from './validation_test.js';
+
+const MAX_VERTEX_ATTRIBUTES: number = 16;
+const MAX_VERTEX_BUFFER_END: number = 2048;
+const MAX_VERTEX_BUFFER_STRIDE: number = 2048;
+const MAX_VERTEX_BUFFERS: number = 16;
+
+const VERTEX_SHADER_CODE_WITH_NO_INPUT = `
+  #version 450
+  void main() {
+    gl_Position = vec4(0.0);
+  }
+`;
+
+class F extends ValidationTest {
+  async init(): Promise<void> {
+    await Promise.all([super.init(), this.initGLSL()]);
+  }
+
+  getDescriptor(
+    vertexInput: GPUVertexInputDescriptor,
+    vertexShaderCode: string
+  ): GPURenderPipelineDescriptor {
+    const descriptor: GPURenderPipelineDescriptor = {
+      vertexStage: this.getVertexStage(vertexShaderCode),
+      fragmentStage: this.getFragmentStage(),
+      layout: this.getPipelineLayout(),
+      primitiveTopology: 'triangle-list',
+      colorStates: [{ format: 'rgba8unorm' }],
+      vertexInput,
+    };
+    return descriptor;
+  }
+
+  getVertexStage(code: string): GPUProgrammableStageDescriptor {
+    return {
+      module: this.makeShaderModule('vertex', code),
+      entryPoint: 'main',
+    };
+  }
+
+  getFragmentStage(): GPUProgrammableStageDescriptor {
+    const code = `
+      #version 450
+      layout(location = 0) out vec4 fragColor;
+      void main() {
+        fragColor = vec4(0.0, 1.0, 0.0, 1.0);
+      }
+    `;
+    return {
+      module: this.makeShaderModule('fragment', code),
+      entryPoint: 'main',
+    };
+  }
+
+  getPipelineLayout(): GPUPipelineLayout {
+    return this.device.createPipelineLayout({ bindGroupLayouts: [] });
+  }
+}
+
+export const g = new TestGroup(F);
+
+g.test('an empty vertex input is valid', t => {
+  const vertexInput: GPUVertexInputDescriptor = {};
+  const descriptor = t.getDescriptor(vertexInput, VERTEX_SHADER_CODE_WITH_NO_INPUT);
+  t.device.createRenderPipeline(descriptor);
+});
+
+g.test('a null buffer is valid', t => {
+  {
+    // One null buffer is OK
+    const vertexInput: GPUVertexInputDescriptor = {
+      vertexBuffers: [
+        {
+          stride: 0,
+          attributeSet: [],
+        },
+      ],
+    };
+    const descriptor = t.getDescriptor(vertexInput, VERTEX_SHADER_CODE_WITH_NO_INPUT);
+    t.device.createRenderPipeline(descriptor);
+  }
+  {
+    //  One null buffer followed by a buffer is OK
+    const vertexInput: GPUVertexInputDescriptor = {
+      vertexBuffers: [
+        {
+          stride: 0,
+          attributeSet: [],
+        },
+        {
+          stride: 0,
+          attributeSet: [
+            {
+              shaderLocation: 0,
+              format: 'float',
+            },
+          ],
+        },
+      ],
+    };
+    const descriptor = t.getDescriptor(vertexInput, VERTEX_SHADER_CODE_WITH_NO_INPUT);
+    t.device.createRenderPipeline(descriptor);
+  }
+  {
+    //  One null buffer sitting between buffers is OK
+    const vertexInput: GPUVertexInputDescriptor = {
+      vertexBuffers: [
+        {
+          stride: 0,
+          attributeSet: [
+            {
+              shaderLocation: 0,
+              format: 'float',
+            },
+          ],
+        },
+        {
+          stride: 0,
+          attributeSet: [],
+        },
+        {
+          stride: 0,
+          attributeSet: [
+            {
+              shaderLocation: 1,
+              format: 'float',
+            },
+          ],
+        },
+      ],
+    };
+    const descriptor = t.getDescriptor(vertexInput, VERTEX_SHADER_CODE_WITH_NO_INPUT);
+    t.device.createRenderPipeline(descriptor);
+  }
+});
+
+g.test('pipeline vertex buffers are backed by attributes in vertex input', async t => {
+  const vertexInput: GPUVertexInputDescriptor = {
+    vertexBuffers: [
+      {
+        stride: 2 * Float32Array.BYTES_PER_ELEMENT,
+        attributeSet: [
+          {
+            shaderLocation: 0,
+            format: 'float',
+          },
+          {
+            shaderLocation: 1,
+            format: 'float',
+          },
+        ],
+      },
+    ],
+  };
+  {
+    // Control case: pipeline with one input per attribute
+    const code = `
+      #version 450
+      layout(location = 0) in vec4 a;
+      layout(location = 1) in vec4 b;
+      void main() {
+          gl_Position = vec4(0.0);
+      }
+    `;
+    const descriptor = t.getDescriptor(vertexInput, code);
+    t.device.createRenderPipeline(descriptor);
+  }
+  {
+    // Check it is valid for the pipeline to use a subset of the VertexInput
+    const code = `
+      #version 450
+      layout(location = 0) in vec4 a;
+      void main() {
+          gl_Position = vec4(0.0);
+      }
+    `;
+    const descriptor = t.getDescriptor(vertexInput, code);
+    t.device.createRenderPipeline(descriptor);
+  }
+  {
+    // Check for an error when the pipeline uses an attribute not in the vertex input
+    const code = `
+      #version 450
+      layout(location = 2) in vec4 a;
+      void main() {
+          gl_Position = vec4(0.0);
+      }
+    `;
+    const descriptor = t.getDescriptor(vertexInput, code);
+
+    await t.expectValidationError(() => {
+      t.device.createRenderPipeline(descriptor);
+    });
+  }
+});
+
+g.test('a stride of 0 is valid', t => {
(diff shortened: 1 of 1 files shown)


src/suites/cts/validation/vertex_input.spec.ts
```diff
@@ -0,0 +1,619 @@
...
+    ],
+  };
+  const descriptor = t.getDescriptor(vertexInput, VERTEX_SHADER_CODE_WITH_NO_INPUT);
+
+  await t.expectValidationError(() => {
+    t.device.createRenderPipeline(descriptor);
+  });
+});
+
+g.test('check for some potential underflow in the vertex input validation', async t => {
```
kainino0x: Can you clarify what this is trying to test? It looks like it's making an attribute (size 16) that's larger than the stride (4).

src/suites/cts/validation/vertex_input.spec.ts
```diff
@@ -0,0 +1,619 @@
...
+    ],
+  };
+  {
+    // Control case, setting max attribute offset for Float32 vertex format
+    const descriptor = t.getDescriptor(vertexInput, VERTEX_SHADER_CODE_WITH_NO_INPUT);
+    t.device.createRenderPipeline(descriptor);
+  }
+  {
+    // Test attribute offset out of bounds
+    vertexInput.vertexBuffers![0].attributeSet![0].offset = MAX_VERTEX_BUFFER_END - 1;
```
kainino0x: We also need a test that, for example, 3, 5, and 2045 *are* valid offsets. I don't think I saw one.

src/suites/cts/validation/vertex_input.spec.ts
```diff
@@ -0,0 +1,619 @@
...
+    ],
+  };
+  {
+    // Control case, setting correct stride and offset
+    const descriptor = t.getDescriptor(vertexInput, VERTEX_SHADER_CODE_WITH_NO_INPUT);
+    t.device.createRenderPipeline(descriptor);
+  }
+  {
+    // Test vertex attribute offset exceed vertex buffer stride range
+    vertexInput.vertexBuffers![0].stride = Float32Array.BYTES_PER_ELEMENT;
```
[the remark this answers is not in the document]
beaufortfrancois: I believe validation occurs when offset + elementSize > stride (4 + 4 > 4).
- offset is 4 (Float32Array.BYTES_PER_ELEMENT)
- elementSize is 4 (float format)
- stride is now set to 4 (Float32Array.BYTES_PER_ELEMENT)
So it shouldn't be valid.

src/suites/cts/validation/vertex_input.spec.ts
```diff
@@ -0,0 +1,619 @@
...
+    const descriptor = t.getDescriptor(vertexInput, VERTEX_SHADER_CODE_WITH_NO_INPUT);
+
+    await t.expectValidationError(() => {
+      t.device.createRenderPipeline(descriptor);
+    });
+  }
```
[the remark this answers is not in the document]
beaufortfrancois: Done

src/suites/cts/validation/vertex_input.spec.ts
```diff
@@ -0,0 +1,619 @@
...
+    ],
+  };
+  {
+    // Control case, setting max input stride
+    const descriptor = t.getDescriptor(vertexInput, VERTEX_SHADER_CODE_WITH_NO_INPUT);
+    t.device.createRenderPipeline(descriptor);
+  }
+  {
+    // Test input stride OOB
+    vertexInput.vertexBuffers![0].stride = MAX_VERTEX_BUFFER_STRIDE + 1;
```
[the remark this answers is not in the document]
beaufortfrancois: I assume you meant 2049 instead of 2048. I've updated stride to 2052 (2048+4).
Thanks for catching!

src/suites/cts/validation/vertex_input.spec.ts
```diff
@@ -0,0 +1,619 @@
...
+    ],
+  };
+  {
+    // Control case, setting input stride 4 bytes
+    const descriptor = t.getDescriptor(vertexInput, VERTEX_SHADER_CODE_WITH_NO_INPUT);
+    t.device.createRenderPipeline(descriptor);
+  }
+  {
+    // Test input stride not multiple of 4 bytes
+    vertexInput.vertexBuffers![0].stride = 2;
```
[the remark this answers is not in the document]
beaufortfrancois: I've used `uchar2` so that Chrome doesn't crash instead of `char2` as it's not yet supported.

src/suites/cts/validation/vertex_input.spec.ts
```diff
@@ -0,0 +1,619 @@
...
+    const descriptor = t.getDescriptor(vertexInput, VERTEX_SHADER_CODE_WITH_NO_INPUT);
+
+    await t.expectValidationError(() => {
+      t.device.createRenderPipeline(descriptor);
+    });
+  }
+});
+
+g.test('we cannot set an already set attribute', async t => {
```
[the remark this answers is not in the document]
beaufortfrancois: Done.

src/suites/cts/validation/vertex_input.spec.ts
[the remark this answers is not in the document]
beaufortfrancois: You're right. Thanks!

src/suites/cts/validation/vertex_input.spec.ts
beaufortfrancois: According to https://dawn-review.googlesource.com/c/dawn/+/7840,  it's trying to test that  no underflow is possible because the max vertex format size is smaller than 2048. @Kangz may know better

beaufortfrancois: After offline discussion with @Kangz, we decided to remove this test as it was not needed in the CTS.

src/suites/cts/validation/vertex_input.spec.ts
beaufortfrancois: Attribute offsets must be 4 byte aligned. This is why there were no tests for 3, 5, and 2045.

beaufortfrancois: I've added a test for "check multiple of 4 bytes constraint on offset" as it was missing

src/suites/cts/validation/vertex_input.spec.ts
```diff
@@ -0,0 +1,619 @@
...
+    };
+    const descriptor = t.getDescriptor(vertexInput, VERTEX_SHADER_CODE_WITH_NO_INPUT);
+
+    await t.expectValidationError(() => {
+      t.device.createRenderPipeline(descriptor);
+    });
+  }
+});
+
+g.test('check out of bounds condition on total number of vertex attributes', async t => {
```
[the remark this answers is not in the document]
beaufortfrancois: Done.

src/suites/cts/validation/vertex_input.spec.ts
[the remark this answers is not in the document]
beaufortfrancois: I've added one test for two attributes overlapping. Note that Chrome fails. Is this expected?
FYI @Kangz

beaufortfrancois:
@kainino0x I've addressed all your feedback but https://github.com/gpuweb/cts/pull/44#discussion_r331916857
Let me know what you think.

src/suites/cts/validation/vertex_input.spec.ts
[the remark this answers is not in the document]
kainino0x: I actually don't know whether it should be valid or invalid in WebGPU. Fine to leave the test in until we figure it out.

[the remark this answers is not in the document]
kainino0x: Oops, you're right. Got totally confused.

src/suites/cts/validation/vertex_input.spec.ts
```diff
@@ -359,6 +411,42 @@ g.test('check out of bounds condition on total number of vertex attributes', asy
...
+
+  {
+    // Control case, setting max vertex buffer number
+    const vertexInput: GPUVertexInputDescriptor = { vertexBuffers };
+    const descriptor = t.getDescriptor(vertexInput, VERTEX_SHADER_CODE_WITH_NO_INPUT);
+    t.device.createRenderPipeline(descriptor);
+  }
+  {
+    // Test vertex attribute number exceed the limit
+    vertexBuffers.push(getVertexBuffer(MAX_VERTEX_ATTRIBUTES));
```
kainino0x: Doesn't this also push the number of vertex _buffers_ to max+1? If the intent is to test the max attributes then the extra attribute needs to be on an existing buffer.

src/suites/cts/validation/vertex_input.spec.ts
```diff
@@ -564,9 +652,15 @@ g.test('check attribute offset out of bounds', async t => {
     const descriptor = t.getDescriptor(vertexInput, VERTEX_SHADER_CODE_WITH_NO_INPUT);
     t.device.createRenderPipeline(descriptor);
   }
+  {
+    // Control case, setting attribute offset to 4
+    vertexInput.vertexBuffers![0].attributeSet![0].offset = 4;
```
kainino0x: Another good control case would be the largest valid value: offset=MAX_VERTEX_BUFFER_END-4. Or if the format were float2, MAX_VERTEX_BUFFER_END-8.

src/suites/cts/validation/vertex_input.spec.ts
```diff
@@ -564,9 +652,15 @@ g.test('check attribute offset out of bounds', async t => {
...
+    const descriptor = t.getDescriptor(vertexInput, VERTEX_SHADER_CODE_WITH_NO_INPUT);
+    t.device.createRenderPipeline(descriptor);
+  }
   {
     // Test attribute offset out of bounds
-    vertexInput.vertexBuffers![0].attributeSet![0].offset = MAX_VERTEX_BUFFER_END - 1;
+    vertexInput.vertexBuffers![0].attributeSet![0].offset = MAX_VERTEX_BUFFER_END - 3;
```
kainino0x: Hmm, if offsets must be multiples of 4 as you said in the other comment, then this is triggering two errors (OOB, and multiple of 4). To avoid that you could do float2 and MAX_VERTEX_BUFFER_END-4.

src/suites/cts/validation/vertex_input.spec.ts
```diff
@@ -575,37 +669,47 @@ g.test('check attribute offset out of bounds', async t => {
...
-  await t.expectValidationError(() => {
+  {
+    // Control case, setting offset 4 bytes
+    const descriptor = t.getDescriptor(vertexInput, VERTEX_SHADER_CODE_WITH_NO_INPUT);
     t.device.createRenderPipeline(descriptor);
-  });
+  }
+  {
+    // Test offset not multiple of 4 bytes
+    vertexInput.vertexBuffers![0].attributeSet![0].offset = 2;
```
kainino0x: The size of a float is also 4, so it's conceivable that the offset would have to be a multiple of 4 for float, and a multiple of 2 for uchar2. Need to test with format=uchar2 and offset=2. Should that be invalid?

(If yes, that would seem unfortunate because it forces dead padding between values.)

src/suites/cts/validation/vertex_input.spec.ts
beaufortfrancois: You are absolutely right. I should have caught this.

src/suites/cts/validation/vertex_input.spec.ts
beaufortfrancois: This is actually the first control case, I've updated the "not that clear" comment.

src/suites/cts/validation/vertex_input.spec.ts
beaufortfrancois: Done.

src/suites/cts/validation/vertex_input.spec.ts
beaufortfrancois: I've set format to `uchar2` and it is actually valid.
This means I have to update https://dawn-review.googlesource.com/c/dawn/+/11902 to not consider format as well. 
I've updated test.

beaufortfrancois:
@kainino0x All your latest feedback has been addressed.

src/suites/cts/validation/vertex_input.spec.ts
beaufortfrancois: Argh! Even though https://developer.apple.com/documentation/metal/mtlvertexattributedescriptor/1515785-offset says "The offset value must be a multiple of 4 bytes.", it is not raising an error with offset set to 2 and format set to "uchar2".

@Kangz @austinEng What shall we do?

kainino0x: The documentation says so, so we should enforce it for now (or just leave it as is until we figure out the answer). Hopefully we can figure out the answer on the gpuweb thread.

src/suites/cts/validation/vertex_input.spec.ts
kainino0x: Oh yes, sorry I missed that.

src/suites/cts/validation/vertex_input.spec.ts
kainino0x: Found a good source from a glTF issue:
https://github.com/KhronosGroup/glTF-Validator/issues/10#issuecomment-269799846

Let's enforce the 4 byte alignments.

kainino0x: Done.

kainino0x:
Thank you for the revisions!

PR #50, opened 2019-09-25
Check uncapturedErrorEvent error in error scope test

beaufortfrancois:
This PR makes sure we test that the `uncapturederrorevent` error is a GPUValidationError in a error scope test.

FYI @Kangz @austinEng

### DIFF
--- a/src/suites/cts/validation/error_scope.spec.ts
+++ b/src/suites/cts/validation/error_scope.spec.ts
@@ -32,14 +32,20 @@ class F extends Fixture {
     this.device.getQueue().submit([]);
   }
 
-  async expectUncapturedError(fn: Function): Promise<void> {
+  async expectUncapturedError(fn: Function, callback: Function): Promise<void> {
     // TODO: Make arbitrary timeout value a test runner variable
     const TIMEOUT_IN_MS = 1000;
 
     return this.asyncExpectation(async () => {
       try {
         const promise = new Promise(resolve => {
-          this.device.addEventListener('uncapturederror', resolve, { once: true });
+          const eventListener = ((event: GPUUncapturedErrorEvent) => {
+            this.debug(`Got uncaptured error event with ${event.error}`);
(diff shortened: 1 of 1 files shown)


src/suites/cts/validation/error_scope.spec.ts
```diff
@@ -32,14 +32,20 @@ class F extends Fixture {
...
     // TODO: Make arbitrary timeout value a test runner variable
     const TIMEOUT_IN_MS = 1000;
 
     return this.asyncExpectation(async () => {
       try {
         const promise = new Promise(resolve => {
-          this.device.addEventListener('uncapturederror', resolve, { once: true });
+          const eventListener = ((event: GPUUncapturedErrorEvent) => {
+            this.debug(`Got uncaptured error event with ${event.error}`);
+            callback(event);
```
kainino0x: Instead of passing an extra callback, could we just resolve with the value of the event? In fact, despite the `: Promise<void>` at the top, that's probably what was already happening, because we passed `resolve` directly into addEventListener.

src/suites/cts/validation/error_scope.spec.ts
```diff
@@ -101,9 +107,14 @@ g.test('if an error scope matches an error it does not bubble to the parent scop
 g.test('if no error scope handles an error it fires an uncapturederror event', async t => {
   t.device.pushErrorScope('out-of-memory');
 
-  await t.expectUncapturedError(() => {
-    t.createErrorBuffer();
-  });
+  await t.expectUncapturedError(
```
kainino0x: Then this would be

```
const uncapturedErrorEvent = await t.expectUncapturedError(() => {
  t.createErrorBuffer();
});
t.expect(uncapturedErrorEvent.error instanceof GPUValidationError);
```

beaufortfrancois:
@kainino0x WDYT of this new patch?

PR #51, opened 2019-10-01
Protect expectUncapturedError with asyncExpectation

kainino0x:


src/framework/util/index.ts
```diff
@@ -8,6 +8,14 @@ export function now(): number {
   return perf.now();
 }
 
+export function rejectOnTimeout(ms: number, msg: string): Promise<GPUUncapturedErrorEvent> {
```
beaufortfrancois: This seems weird to have GPUUncapturedErrorEvent in until don't you think?

kainino0x: Oops, yes.

beaufortfrancois:
LGTM!

PR #58, opened 2019-10-08
Remove NONE usage in tests

beaufortfrancois:
Following WebGPU spec change at https://github.com/gpuweb/gpuweb/pull/461/files, this PR removes NONE usage in CTS tests.

### DIFF
--- a/src/suites/cts/validation/createBindGroupLayout.spec.ts
+++ b/src/suites/cts/validation/createBindGroupLayout.spec.ts
@@ -63,27 +63,18 @@ g.test('negative binding index', async t => {
   });
 });
 
-g.test('Visibility of bindings cannot be 0', async t => {
-  const goodDescriptor: GPUBindGroupLayoutDescriptor = {
+g.test('Visibility of bindings can be 0', async t => {
+  const descriptor: GPUBindGroupLayoutDescriptor = {
     bindings: [
       {
         binding: 0,
-        visibility: GPUShaderStage.COMPUTE,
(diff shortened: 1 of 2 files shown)


src/suites/cts/validation/createBindGroupLayout.spec.ts
```diff
@@ -63,30 +63,6 @@ g.test('negative binding index', async t => {
   });
 });
 
-// TODO: Update once https://github.com/gpuweb/gpuweb/issues/405 is decided.
-g.test('Visibility of bindings cannot be None', async t => {
```
Kangz: It can still be 0 though so I think we should keep the test.

beaufortfrancois: How can it be 0 it 0 is not a valid value?
See https://chromium-review.googlesource.com/c/chromium/src/+/1845762/1/third_party/blink/renderer/modules/webgpu/gpu_shader_stage.idl

Kangz: It is just a number, you can still pass 0. The only thing that changed is that we don't have a constant for 0 because it is understood that NONE is always 0 for bitflags.

I need to send the TL;DR for the F2F discussions but IIRC we decided that 0 was a valid thing to have, just not a useful constant to expose.

beaufortfrancois: Understood, I've renamed test then. Thanks!

PR #60, opened 2019-10-11
Fix generated trailing whitespace in GLSL macro

austinEng:


src/tools/glsl.macro.js
```diff
@@ -41,7 +41,7 @@ module.exports = createMacro(({ references, state, babel }) => {
       .substring(path.parent.start, path.parent.end)
       .trim()
       .split('\n')
-      .map(x => (x ? ' ' + x : x))
+      .map(x => x.trim().length === 0 ? x : ' ' + x)
```
kainino0x: Would still output spaces if there are spaces on a blank line. Just map it to `''` I guess.

kainino0x: Turns out the problem was due to a trailing CR, I think. Given that, this fix doesn't do anything (and this PR probably doesn't do anything either), but still strictly better I think.

PR #63, opened 2019-10-14
create setTimeout helper and add lint

kainino0x:
WPT doesn't allow setTimeout, so try to use its step_timeout when it is
available.


kainino0x:
Tested in standalone and in WPT.

PR #64, opened 2019-10-15
Use an async-mutex to run one test at a time in WPT runtime

kainino0x:
This is technically a workaround for a Chrome bug (in which I think every device is actually the same device).

However, this is an old TODO anyway - I never really wanted to run all the tests in parallel.

In the future I think we'll want to extend this to do something like "run up to 10 in parallel".

### DIFF
--- a/src/framework/util/async_mutex.ts
+++ b/src/framework/util/async_mutex.ts
@@ -1,0 +1,21 @@
+export class AsyncMutex {
+  private lastToRun: Promise<unknown> | undefined;
+
+  // Run an async function with a lock on this mutex.
+  // Waits until the mutex is available, locks it, runs the function, then releases it.
+  async with<T>(fn: () => Promise<T>): Promise<T> {
+    const p = (async () => {
+      // If the mutex is locked, wait for the last thing in the queue before running.
+      // (Everything in the queue runs in order, so this is after everything currently enqueued.)
+      if (this.lastToRun) {
+        await this.lastToRun;
+      }
+      return fn();
+    })();
+
+    // Push the newly-created Promise onto the queue.
+    this.lastToRun = p;
+    // And return so the caller can wait on the result.
+    return p;
+  }
+}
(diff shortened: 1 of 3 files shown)


src/framework/util/async_mutex.ts
```diff
@@ -0,0 +1,30 @@
+export class AsyncMutex {
+  private queue: Array<Promise<unknown>> = [];
+
+  // Run an async function with a lock on this mutex.
+  // Waits until the mutex is available, locks it, runs the function, then releases it.
+  async with<T>(fn: () => Promise<T>): Promise<T> {
```
austinEng: This looks correct, but is the queue really necessary? I think the AsyncMutex only needs to remember the last promise it generated, `await` it, and then replace the last promise.

kainino0x: Good call, let me try to make that work.

kainino0x: Much better, thanks!

src/framework/util/async_mutex.ts
```diff
@@ -0,0 +1,21 @@
+export class AsyncMutex {
+  private lastToRun: Promise<unknown> | undefined;
+
+  // Run an async function with a lock on this mutex.
+  // Waits until the mutex is available, locks it, runs the function, then releases it.
+  async with<T>(fn: () => Promise<T>): Promise<T> {
+    const p = (async () => {
+      // If the mutex is locked, wait for the last thing in the queue before running.
```
austinEng: nit: Now the queue is sort of implicitly defined by a chain of promises. These comments may be a bit confusing.

src/framework/util/async_mutex.ts
```diff
@@ -0,0 +1,21 @@
...
+      // (Everything in the queue runs in order, so this is after everything currently enqueued.)
+      if (this.lastToRun) {
+        await this.lastToRun;
+      }
+      return fn();
+    })();
+
+    // Push the newly-created Promise onto the queue.
```
austinEng: This comment too

src/framework/util/async_mutex.ts
kainino0x: Done

PR #66, opened 2019-10-18
Add "private" params, prefixed with underscore

kainino0x:
This cleans up the test case IDs because they no longer have non-identifying "success"/"result" keys in them.

### DIFF
--- a/src/suites/cts/examples.spec.ts
+++ b/src/suites/cts/examples.spec.ts
@@ -55,13 +55,21 @@ g.test('basic/async', async t => {
   );
 });
 
+// A test can be parameterized with a simple array of objects.
+//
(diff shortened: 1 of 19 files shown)


kainino0x:
@beaufortfrancois The 2 commits can be reviewed separately.

src/suites/cts/examples.spec.ts
```diff
@@ -55,11 +55,16 @@ g.test('basic/async', async t => {
   );
 });
 
+// A test can be parameterized with a simple array of objects.
+//
+// Parameters can be public (x, y) which means they're part of the case name.
+// They can also be private by starting with an underscore (_result), which passes
+// them into the test but does not make them part of the case name.
```
beaufortfrancois: Nit: Add the name of the test case.

PR #67, opened 2019-10-19
Texture format table

kainino0x:
First 2 commits are #66; can ignore them.

The other 3 commits can each be reviewed separately.

PTAL!

### DIFF
--- a/src/suites/cts/format_info.ts
+++ b/src/suites/cts/format_info.ts
@@ -1,0 +1,64 @@
+import { poptions } from '../../framework/index.js';
+
+export interface TextureFormatInfo {
+  renderable: boolean;
+  // Add fields as needed
+}
+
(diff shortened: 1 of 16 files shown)


src/suites/cts/format_info.ts
```diff
@@ -0,0 +1,55 @@
...
+}
+
+interface TextureFormats {
+  [k: string]: TextureFormatInfo;
+}
+
+// prettier-ignore
+export const textureFormatInfo: TextureFormats = {
+  // Try to keep these manually-formatted in a readable grid.
+  'r8unorm':                { renderable:  true },
```
beaufortfrancois: Removing comments such as `// 8-bit formats` make it hard to know why this list is not ordered alphabetically.
How do you feel about keeping those?

PR #68, opened 2019-10-22
make pcombine take variadic args instead of an array

kainino0x:
Just a tiny readability/writability improvement.


kainino0x:
PTAL, but I'm going to land now.

PR #70, opened 2019-10-23
Add eventualAsyncExpectations, and use it for most async expectations

kainino0x:
eventualAsyncExpectations are awaited during finalize, but don't need to
block the rest of the test. More importantly, you don't have to remember
to use "await" every single time in every test.

The old thing, now called immediateAsyncExpectation, still exists but is
only needed in one place: expectUncapturedError.


kainino0x:
@beaufortfrancois PTAL
@Kangz FYI

beaufortfrancois:
Pretty cool indeed! So much better and faster. Thank you!

PR #73, opened 2019-10-24
standalone runtime: make tree structure, fix selections, add links

kainino0x:
- selecting test names no longer also selects the black triangle
- on chrome/firefox, clicking the name of a test/case selects it
- added links to test and case lines to open just that test/case
- convert flat structure into tree structure, splitting on slashes
  - make run buttons work recursively
  - make link buttons work for any subtree
  - do css to make tree structure nice
- prevent page reload if there are any test results

### DIFF
--- a/index.html
+++ b/index.html
@@ -3,6 +3,9 @@
     <title></title>
     <script src="third_party/jquery/jquery-3.3.1.min.js"></script>
     <style>
+      * {
+        box-sizing: border-box;
+      }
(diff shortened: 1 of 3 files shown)


kainino0x:
Code is gross but better than before, it seems to work and we'll just fix bugs if they come up. Landing without review since this doesn't affect tests.

cc/TBR @beaufortfrancois

PR #79, opened 2019-10-25
Support running tests on a worker

kainino0x:
Tested locally in standalone and WPT.


kainino0x:
PTAL. Can optionally be reviewed one commit at a time.

src/tools/crawl.ts
```diff
@@ -4,10 +4,11 @@
 
 // tslint:disable: no-console
 
-import fg from 'fast-glob';
 import * as fs from 'fs';
 import * as path from 'path';
 
+const fg = require('fast-glob');
```
austinEng: nit: was this supposed to change?

kainino0x: Yeah, I did this as a hack to make TS happy after turning off esModuleInterop, which isn't otherwise necessary.

PR #83, opened 2019-10-31
Add t.skip()

kainino0x:
- Use it in initGLSL
- Use it in canvas element test
- Refactor logging state machine
- Add logging test and strengthen other logging state tests (fail > warn > skip > pass)

### DIFF
--- a/src/framework/logger.ts
+++ b/src/framework/logger.ts
@@ -56,16 +56,15 @@ export class TestSpecRecorder {
 }
 
 enum PassState {
-  Pass = 0,
-  Skip = 1,
-  Warn = 2,
-  Fail = 3,
+  pass = 0,
+  skip = 1,
+  warn = 2,
+  fail = 3,
 }
-const stateName: Status[] = ['pass', 'skip', 'warn', 'fail'];
 
 export class TestCaseRecorder {
   private result: LiveTestCaseResult;
-  private state = PassState.Pass;
+  private state = PassState.pass;
   private startTime = -1;
   private logs: string[] = [];
   private debugging = false;
@@ -77,7 +76,7 @@ export class TestCaseRecorder {
   start(debug: boolean = false): void {
     this.startTime = now();
     this.logs = [];
-    this.state = PassState.Pass;
+    this.state = PassState.pass;
     this.debugging = debug;
   }
 
@@ -88,7 +87,7 @@ export class TestCaseRecorder {
     const endTime = now();
     // Round to next microsecond to avoid storing useless .xxxx00000000000002 in results.
     this.result.timems = Math.ceil((endTime - this.startTime) * 1000) / 1000;
-    this.result.status = stateName[this.state];
+    this.result.status = PassState[this.state] as Status;
 
     this.result.logs = this.logs;
     this.debugging = false;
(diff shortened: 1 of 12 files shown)


src/framework/test_group.ts
```diff
@@ -138,13 +138,16 @@ class RunCaseSpecific<F extends Fixture> implements RunCase {
 
     try {
       const inst = new this.fixture(rec, this.params || {});
+      // Note: init() may throw SkipTestCase, in which case finalize() is not run.
```
austinEng: Hmm.. but why not have `finalize()` in a `finally` block?

I guess right now there's no need for it and we can add it in the future if necessary.

src/runtime/wpt.ts
```diff
@@ -45,6 +45,7 @@ declare function async_test(f: (this: WptTestObject) => Promise<void>, name: str
           }
 
           this.step(() => {
+            // Unfortunately, it seems not possible to surface any logs for warn/skip.
```
austinEng: We still get them in standalone, right?

In wpt, could we console.warn them? Maybe behind a verbose flag?

src/suites/cts/validation/vertex_input.spec.ts
```diff
@@ -24,7 +24,9 @@ function clone(descriptor: GPUVertexInputDescriptor): GPUVertexInputDescriptor {
 
 class F extends ValidationTest {
   async init(): Promise<void> {
-    await Promise.all([super.init(), this.initGLSL()]);
+    // Give initGLSL the opportunity to throw SkipTestCase before super.init() pushes error scopes.
```
austinEng: Is this first because `finalize()` is not run if the test is skipped and we don't want to leave error scopes open?

If we called `finalize()` in a `finally` block as mentioned above, then we avoid this problem.

src/framework/test_group.ts
kainino0x: I was thinking it doesn't make sense to finalize if init fails, but maybe it does. The only thing that would happen right now (I think) is that the popErrorScopes would fail.

src/runtime/wpt.ts
kainino0x: Yeah, standalone handles all these. Good idea to console.warn.

src/framework/logger.ts
```diff
@@ -54,10 +55,17 @@ export class TestSpecRecorder {
   }
 }
 
+enum PassState {
+  Pass = 0,
+  Skip = 1,
+  Warn = 2,
+  Fail = 3,
+}
+const stateName: Status[] = ['pass', 'skip', 'warn', 'fail'];
```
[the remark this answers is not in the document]
kainino0x: Made the PassState members lowercase instead. Thanks for the pointer!

src/runtime/wpt.ts
kainino0x: I remembered WPT lints against console logging. Maybe we should just leave this as is, since people should hopefully be debugging on the standalone version anyway.

kainino0x: (which also has the typescript source so source maps work)

austinEng: sgtm

kainino0x:
PTAL @ second commit.

kainino0x:
thanks!

PR #87, opened 2019-11-05
fix getStackTrace (for the 6th time)

kainino0x:


kainino0x:
Decided to do a more involved change to the error logging, which I'll push tomorrow, so I repurposed this PR for just the one isolated fix.

PR #88, opened 2019-11-05
Add "print stack" button to standalone

kainino0x:
- Logs error objects instead of strings, and convert to strings in toJSON.
  - Refactor logger to only take error objects.
- Add a button to print that stack to the console so you can click source links.
- Capture stacks before async work so they're useful.
- Update uses of t.warn() et al to use t.rec.warn() et al.
- Remove t.log() and t.rec.log() in favor of t.debug().
- Improve logging of multi-line messages in cmdline runner.

chrome:
screenshot chrome
safari:
screenshot safari

### DIFF
--- a/src/framework/fixture.ts
+++ b/src/framework/fixture.ts
@@ -22,11 +22,7 @@ export class Fixture {
   async init(): Promise<void> {}
 
   debug(msg: string): void {
-    this.rec.debug(msg);
-  }
-
-  log(msg: string): void {
-    this.rec.log(msg);
+    this.rec.debug(new Error(msg));
   }
 
   skip(msg: string): never {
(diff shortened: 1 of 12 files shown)


src/framework/fixture.ts
```diff
@@ -58,33 +54,38 @@ export class Fixture {
...
-  protected eventualAsyncExpectation<T>(fn: () => Promise<T>): Promise<T> {
-    const promise = fn();
+  protected eventualAsyncExpectation<T>(fn: (niceStack: Error) => Promise<T>): Promise<T> {
+    const promise = fn(new Error());
     this.eventualExpectations.push(promise);
     return promise;
   }
 
-  private expectErrorValue(expectedName: string, ex: unknown, m: string): void {
+  private expectErrorValue(expectedName: string, ex: unknown, niceStack: Error): void {
```
beaufortfrancois: Is `niceStack` because we can't/don't want to use `stack`?

kainino0x: No, just because it's more descriptive. I.e. it's the "nice stack" to show alongside this message instead of the "real" stack (because it's captured before any awaits).

kainino0x: (I'll revise after landing if you think another name would be better.)

PR #91, opened 2019-11-11
Add cts/command_buffer/compute/sampled_texture_clear test

natashalee:
Add testing for compute pass copied from dawn repo test TextureZeroInitTest::ComputePassSampledTextureClear

### DIFF
--- a/src/suites/cts/resource_init/sampled_texture_clear.spec.ts
+++ b/src/suites/cts/resource_init/sampled_texture_clear.spec.ts
@@ -1,0 +1,75 @@
+export const description = `
+computePass test that sampled texture is cleared`;
+
+import { TestGroup } from '../../../framework/index.js';
+import GLSL from '../../../tools/glsl.macro.js';
+import { GPUTest } from '../gpu_test.js';
+
+export const g = new TestGroup(GPUTest);
+
+g.test('compute pass test that sampled texture is cleared', async t => {
+  const texture = t.device.createTexture({
+    size: { width: 256, height: 256, depth: 1 },
+    format: 'r8unorm',
+    usage: GPUTextureUsage.SAMPLED,
+  });
+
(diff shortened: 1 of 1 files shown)


src/suites/cts/command_buffer/compute/sampled_texture_clear.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+
+import { TestGroup } from '../../../../framework/index.js';
+import GLSL from '../../../../tools/glsl.macro.js';
+import { GPUTest } from '../../gpu_test.js';
+
+export const g = new TestGroup(GPUTest);
+
+g.test('compute pass test that sampled texture is cleared', async t => {
+  const texture = t.device.createTexture({
+    size: { width: 1, height: 1, depth: 1 },
```
kainino0x: Without too much trouble, I think we could make this more robust by testing a larger texture. You can just do a similar copy from texture into buffer, but with a larger texture, larger dispatch, and larger buffer, then check the whole buffer at the end. I'd make it pretty large, say 256x256.

src/suites/cts/command_buffer/compute/sampled_texture_clear.spec.ts
```diff
@@ -0,0 +1,75 @@
+export const description = `
```
kainino0x: nit: I think we'll end up having a number of tests for resource initialization. How about making a new directory just for that? e.g. `src/suites/cts/resource_init/`, `sampled_texture.spec.ts` or `compute/sampled_texture.spec.ts` depending on whether you think we'll probably want to split files based on pipeline type in the long run.

natashalee:
Thanks kai, I pushed the updates!

kainino0x:
Oops, sorry I missed it (no notif on pushes to a PR). thank you!

PR #94, opened 2019-11-18
Change .getQueue() to .defaultQueue

romandev:
Spec changes:
https://github.com/gpuweb/gpuweb/pull/490

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "@types/jquery": "^3.3.30",
     "@types/node": "^12.0.8",
     "@webgpu/glslang": "^0.0.8",
-    "@webgpu/types": "0.0.17",
+    "@webgpu/types": "0.0.18",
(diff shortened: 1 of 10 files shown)


Kangz:
LGTM but it doesn't pass the CI because we didn't push the webgpu-types update to npm yet. @kainino0x or @beaufortfrancois I think you both have accesses to do that, can you push a new version of webgpu-types?

kainino0x:
If François can't do it, please update the package.json to point directly
at the GitHub repo instead of npn, which I'm pretty sure is possible. I'll
publish to npm when I'm back from vacation.

beaufortfrancois:
I don't have access to https://www.npmjs.com/package/@webgpu/types sadly.

kainino0x:
Types updated and rolled. Thanks for the patch!

PR #95, opened 2019-12-12
Add test case for copyImageBitmapToTexture

shaoboyan:


shaoboyan:
@kainino0x @austinEng @Kangz  PTAL if you have time.
This the cts for copyImageBitmapToTexture. And it will cover https://chromium-review.googlesource.com/c/chromium/src/+/1962612 this implementation.

src/suites/cts/command_buffer/copies.spec.ts
```diff
@@ -109,3 +109,67 @@ g.test('b2t2t2b', async t => {
...
+
+  const testBuffer = t.device.createBuffer({
+    size: imageBitmap.width * imageBitmap.height * 4,
+    usage: GPUBufferUsage.COPY_SRC | GPUBufferUsage.COPY_DST,
+  });
+
+  const encoder = t.device.createCommandEncoder();
+  encoder.copyTextureToBuffer(
+    { texture: dst, mipLevel: 0, origin: { x: 0, y: 0, z: 0 } },
+    { buffer: testBuffer, rowPitch: imageBitmap.width * 4, imageHeight: imageBitmap.height },
```
austinEng: We should explicitly compute the rowPitch with the 256-byte alignment. If this works, then I think we're getting lucky the the image width is already a multiple of 256.

The size of `testBuffer` would become `rowPitch * imageBitmap.height`

src/suites/cts/command_buffer/copies.spec.ts
```diff
@@ -109,3 +109,67 @@ g.test('b2t2t2b', async t => {
 
   t.expectContents(dst, data);
 });
+
+g.test('imageBitmapToTexture', async t => {
+  const img = document.createElement('img');
+  img.src = '../../assets/img/test.png';
+  await img.decode();
```
kainino0x: is this necessary? Shouldn't it happen implicitly during createImageBitmap, which is also async?

src/suites/cts/command_buffer/copies.spec.ts
```diff
@@ -109,3 +109,67 @@ g.test('b2t2t2b', async t => {
 
   t.expectContents(dst, data);
 });
+
+g.test('imageBitmapToTexture', async t => {
```
kainino0x: This is not a command_buffer test anymore because copyImageBitmapToTexture was moved to the queue.
Could you move these new tests into a new file, say, `src/suites/cts/copyImageBitmapToTexture.spec.ts` for now.

kainino0x: Let's be more specific about what this test tests: `'from image element'` (imageBitmapToTexture is implicit if moved to copyImageBitmapToTexture.spec.ts).

src/suites/cts/command_buffer/copies.spec.ts
kainino0x: yeah, source image is 512x512.

src/suites/cts/command_buffer/copies.spec.ts
```diff
@@ -109,3 +109,67 @@ g.test('b2t2t2b', async t => {
...
+  const img = document.createElement('img');
+  img.src = '../../assets/img/test.png';
```
kainino0x: `test.png` is pretty large. Can we choose a smaller one? E.g. make a simpler image with solid colors (which will compress better), and/or take one from here: 
https://github.com/KhronosGroup/WebGL/tree/master/sdk/tests/resources

Alternatively, we could just test only ImageBitmaps created from a `<canvas>` with generated contents, so we wouldn't need the PNG at all. This would make it easy to test lots (maybe 10s) of different image sizes, which would be really great.

src/suites/cts/command_buffer/copies.spec.ts
```diff
@@ -109,3 +109,67 @@ g.test('b2t2t2b', async t => {
...
+      width: img.width,
+      height: img.height,
+      depth: 1,
+    },
+    format: 'rgba8uint',
+    usage: GPUTextureUsage.COPY_DST | GPUTextureUsage.COPY_SRC,
+  });
+
+  // TODO(shaobo.yan@intel.com) : remove as any after @webgpu/types updating
+  (t.device.defaultQueue as any).copyImageBitmapToTexture(
```
kainino0x: This is available now, can you update `package.json` to point to `@webgpu/types` 0.0.19?

shaoboyan: Sure. Previously I think the roll version patch should be a owner's right :). I'll do it.

src/suites/cts/command_buffer/copies.spec.ts
shaoboyan: Yes, it happens to be a multiple of 256 so I directly use it. But yes, I can add a logic to compute the rowPitch to make it more general

src/suites/cts/command_buffer/copies.spec.ts
shaoboyan: Sry I'm not very farmiliar with Promise-similar thing. But for createImageBitmap, I need to get image's width and height. And I think if I don't want to implement this part in a callback. It should be await for the image been decoded.

src/suites/cts/command_buffer/copies.spec.ts
shaoboyan: Sure. I classify it into the copy buffer. But you're right, under commadnBuffer test suite is not proper.Will do it.

src/suites/cts/command_buffer/copies.spec.ts
shaoboyan: The only problem to test ImageBitmap from <canvas> is it won't cover uploading CPU data to WebGPU texture path (Pls correct me if I'm wrong, I think using canvas which has drawImage call means it is a textrue backed ImageBitmap). 
Happy to choose a smaller image for testing.

src/suites/cts/command_buffer/copies.spec.ts
shaoboyan: Added a helper funtion for rowPtich

src/suites/cts/command_buffer/copies.spec.ts
shaoboyan: Done.

shaoboyan: Done.

src/suites/cts/command_buffer/copies.spec.ts
shaoboyan: Done.

src/suites/cts/command_buffer/copies.spec.ts
kainino0x: Oh, I understand, thanks for the explanation.
I don't think it's actually necessary to pass the width/height into createImageBitmap. Try this: `createImageBitmap(img)`, then the `img.decode()` shouldn't be needed anymore (please double check though).

src/suites/cts/command_buffer/copies.spec.ts
kainino0x: That's true, yeah. If you are specifically trying to avoid texture-backed ImageBitmaps, you should be able to go through [ImageData](https://developer.mozilla.org/en-US/docs/Web/API/ImageData/ImageData): `createImageBitmap(new ImageData(uint8clampedarray, width))`

src/suites/cts/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+g.test('from image element', async t => {
+  const img = document.createElement('img');
+  img.src = '../../assets/img/test.png';
+  await img.decode();
+
+  const imageBitmap = await createImageBitmap(img, 0, 0, img.width, img.height);
+  const dst = t.device.createTexture({
+    size: {
+      width: img.width,
+      height: img.height,
```
kainino0x: ... then use `imageBitmap.width` and `.height` here instead of `img.`

src/suites/cts/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+      depth: 1,
+    },
+    format: 'rgba8uint',
+    usage: GPUTextureUsage.COPY_DST | GPUTextureUsage.COPY_SRC,
+  });
+
+  // TODO(shaobo.yan@intel.com) : remove as any after @webgpu/types updating
+  //(t.device.defaultQueue as any).copyImageBitmapToTexture(
```
kainino0x: outdated comment

src/suites/cts/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+  const testBuffer = t.device.createBuffer({
+    size: rowPitchValue * imageBitmap.height,
+    usage: GPUBufferUsage.COPY_SRC | GPUBufferUsage.COPY_DST,
+  });
+
+  const encoder = t.device.createCommandEncoder();
+
+  encoder.copyTextureToBuffer(
+    { texture: dst, mipLevel: 0, origin: { x: 0, y: 0, z: 0 } },
+    { buffer: testBuffer, rowPitch: rowPitchValue, imageHeight: imageBitmap.height },
```
kainino0x: Do we allow an imageHeight of 0? There's no reason to actually provide an imageHeight here since it's not an array or 3d texture.

Opened https://github.com/gpuweb/gpuweb/pull/517

src/suites/cts/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+import { GPUTest } from './gpu_test.js';
+
+export const g = new TestGroup(GPUTest);
+
+function calculateRowPitchForImageBitmap(width: number): number {
+  if (width % 256 === 0) {
+    return 4 * width;
+  }
+  return 4 * (width + 256 - (width % 256));
+}
```
kainino0x: IIRC the pitch needs to be a multiple of 256 bytes, not 256 pixels (correct me if I'm wrong).

src/suites/cts/copyImageBitmapToTexture.spec.ts
shaoboyan: Agree with you. I feel a little bit confusing when I write this copy.

src/suites/cts/copyImageBitmapToTexture.spec.ts
shaoboyan: You're right. Done.

src/suites/cts/copyImageBitmapToTexture.spec.ts
shaoboyan: Done.

src/suites/cts/copyImageBitmapToTexture.spec.ts
shaoboyan: Done.

src/suites/cts/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+  }
+  return bytesPerRow + 256 - (bytesPerRow % 256);
+}
+
+g.test('from image element', async t => {
+  const img = document.createElement('img');
+  img.src = '../../assets/img/test.png';
+  await img.decode();
+
+  const imageBitmap = await createImageBitmap(img, 0, 0, img.width, img.height);
```
kainino0x: If you remove the extra parameters here, I think the img.decode shouldn't be needed.

```suggestion
  const imageBitmap = await createImageBitmap(img);
```

src/suites/cts/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+import { GPUTest } from './gpu_test.js';
+
+export const g = new TestGroup(GPUTest);
+
+function calculateRowPitchForImageBitmap(width: number, bytesPerPixel: number): number {
+  const bytesPerRow = width * bytesPerPixel;
+  if (bytesPerRow % 256 === 0) {
+    return bytesPerRow;
+  }
+  return bytesPerRow + 256 - (bytesPerRow % 256);
```
kainino0x: If bytesPerRow is 256, this returns 512, but it should return 256.
Could use `Math.ceil(bytesPerRow / 256) * 256)` or `(((bytesPerRow - 1) >> 8 + 1) << 8`

src/suites/cts/command_buffer/copies.spec.ts
kainino0x: Have you had a chance to evaluate whether this approach (ImageData from CPU) would work?

src/suites/cts/copyImageBitmapToTexture.spec.ts
kainino0x: If imageBitmap:0 is allowed in chrome, let's use it just to simplify.

src/suites/cts/copyImageBitmapToTexture.spec.ts
shaoboyan: That seems great!

src/suites/cts/copyImageBitmapToTexture.spec.ts
shaoboyan: No, if bytesPerRow is 256. it will return in the if block. so it is still 256. But I think your suggestion is better.

kainino0x: Oops, you're right. I missed that.

src/suites/cts/command_buffer/copies.spec.ts
shaoboyan: Sry, it seems I missing all your comments when I push the wrong branch.

src/suites/cts/command_buffer/copies.spec.ts
shaoboyan: Sry, it seems I missing all your comments when I push the wrong branch. Will try it today. Thanks for your suggestion.

src/suites/cts/copyImageBitmapToTexture.spec.ts
shaoboyan: Ack.

src/suites/cts/command_buffer/copies.spec.ts
shaoboyan: Tried ImageData from CPU and it works. So remove the image in latest update. Thanks for your guidance.

src/suites/cts/copyImageBitmapToTexture.spec.ts
shaoboyan: Done

src/suites/cts/copyImageBitmapToTexture.spec.ts
shaoboyan: Done

src/suites/cts/copyImageBitmapToTexture.spec.ts
shaoboyan: Done

shaoboyan:
@kainino0x  PTAL the skeleton. If it makes sense, I'll add multiple size imageBitmap to complete the coverage.

src/suites/cts/gpu_test.ts
```diff
@@ -114,6 +114,44 @@ export class GPUTest extends Fixture {
...
   }
 
+  checkCopyImageBitmapResult(
+    src: GPUBuffer,
+    expected: ArrayBufferView,
+    width: number,
+    height: number,
+    rowPitch: number,
+    bytesPerPixel: number
+  ): void {
```
kainino0x: This function duplicates some of the code from expectContents. Could they be refactored to share some code?

Sorry I don't have a specific recommendation on how to do this, I don't have time to dive into it right now.

kainino0x: This helper name implies it's specific to copyImageBitmapToTexture tests. Can you either rename it (if it has more general utility), or move it to a standalone helper function in copyImageBitmapToTexture.spec.ts?

src/suites/cts/gpu_test.ts
```diff
@@ -153,4 +191,48 @@ export class GPUTest extends Fixture {
...
   }
+
+  checkBufferWithRowPitch(
+    actual: Uint8Array,
+    exp: Uint8Array,
+    width: number,
+    height: number,
+    rowPitch: number,
+    bytesPerPixel: number
+  ): string | undefined {
```
kainino0x: same here, can we deduplicate some of this code with checkBuffer?

src/suites/cts/gpu_test.ts
```diff
@@ -153,4 +191,48 @@ export class GPUTest extends Fixture {
...
+    const size = exp.byteLength;
+    const lines = [];
+    let failedPixels = 0;
+    for (let i = 0; i < height; ++i) {
+      for (let j = 0; j < width * bytesPerPixel; ++j) {
+        const indexExp = j + width * i * bytesPerPixel;
```
kainino0x: nit: give "width * bytesPerPixel" a name and use it on both lines

src/suites/cts/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+  const bytesPerPixel = 4;
+  const imagePixels = new Uint8ClampedArray(bytesPerPixel * width * height);
+  for (let i = 0; i < width * height * bytesPerPixel; ++i) {
+    imagePixels[i] = i % 256;
+  }
+
+  const imageData = new ImageData(imagePixels, width, height);
+  const imageBitmap = await createImageBitmap(imageData);
+
+  //const imageBitmap = await createImageBitmap(img, 0, 0, img.width, img.height);
```
kainino0x: nit: stray comment

src/suites/cts/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+  imageCanvas.height = imageBitmap.height;
+
+  const imageCanvasContext = imageCanvas.getContext('2d');
+  t.expect(imageCanvasContext != null);
+  if (!imageCanvasContext) {
+    return;
+  }
+
+  imageCanvasContext.drawImage(imageBitmap, 0, 0, imageBitmap.width, imageBitmap.height);
+  const data = imageCanvasContext.getImageData(0, 0, imageBitmap.width, imageBitmap.height).data;
```
kainino0x: can this be removed and just use `imagePixels` as the reference data?

src/suites/cts/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+copy imageBitmap To texture tests.
+`;
+
+import { TestGroup } from '../../framework/index.js';
+
+import { GPUTest } from './gpu_test.js';
+
+export const g = new TestGroup(GPUTest);
+
+function calculateRowPitchForImageBitmap(width: number, bytesPerPixel: number): number {
```
kainino0x: nit: this function isn't really specific to ImageBitmaps, maybe rename to "calculateRowPitch"

src/suites/cts/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+  const bytesPerRow = width * bytesPerPixel;
+  return (((bytesPerRow - 1) >> 8) + 1) << 8;
```
kainino0x: nit: should comment what this does (rounds up to a multiple of 256)

src/suites/cts/gpu_test.ts
shaoboyan: Sure, and refactor makes sense. I'll address this.

shaoboyan: For current status of CTS, I think copyImageBitmapToTexture is the only one need this. I prefer to move it into copyImageBitmapToTexture.spec.ts

src/suites/cts/copyImageBitmapToTexture.spec.ts
shaoboyan: I tired using imagePixels directly but failed. And refer to https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/imagebitmap/image_bitmap.cc?type=cs&g=0&l=752 I think createImageBitmap won't keep the imageData as what it is. WDYT?

kainino0x: imagePixels itself should definitely not change in value. But you're right, the contents of the ImageBitmap itself might be different (due to premultiplication, and maybe other conversions). It might work if you modify the data generation to avoid invalid values (e.g. just make alpha always 255). 

Otherwise, how about just factoring this out into a helper imageBitmapToData?

shaoboyan: Make alpha always 255 seems good. And consider the follwing GPU copy path, a helper function seems great. I'm considering the refactor of expectContents by adding default values but don't think it is good so I'll continue on this.
BTW, today our team has some affairs so maybe I cannot update it today. Sry about this.

src/suites/cts/gpu_test.ts
shaoboyan: Did some simple refactor to reuse some common code.  A further deep refactor still under consideration.

shaoboyan: Done. Add a helper class

src/suites/cts/gpu_test.ts
shaoboyan: Done.

src/suites/cts/gpu_test.ts
shaoboyan: Done.

src/suites/cts/copyImageBitmapToTexture.spec.ts
shaoboyan: Done.

src/suites/cts/copyImageBitmapToTexture.spec.ts
shaoboyan: Double check with alpha always 255. It turns out we can check the copy result with imagePixels directly. But still abstract the  drawImage method to cover more cases.

src/suites/cts/copyImageBitmapToTexture.spec.ts
shaoboyan: Done.

src/suites/cts/copyImageBitmapToTexture.spec.ts
shaoboyan: Done.

src/suites/cts/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+  // Using drawImage to extract imageBitmap content.
+  imageBitmapToData(imageBitmap: ImageBitmap): ArrayBufferView | undefined {
+    const imageCanvas = document.createElement('canvas');
+    imageCanvas.width = imageBitmap.width;
+    imageCanvas.height = imageBitmap.height;
+
+    const imageCanvasContext = imageCanvas.getContext('2d');
+    this.expect(imageCanvasContext != null);
+    if (!imageCanvasContext) {
+      return undefined;
```
kainino0x: `throw new Error(...)` instead of `return undefined` so that the return type of the function doesn't have to have `| undefined` in it.

src/suites/cts/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,168 @@
+export const description = `
+copy imageBitmap To texture tests.
+`;
+
+import { TestGroup } from '../../framework/index.js';
+
+import { GPUTest } from './gpu_test.js';
+
+class F extends GPUTest {
+  calculateRowPitch(width: number, bytesPerPixel: number): number {
```
kainino0x: nit: can be a standalone function or static method (does not use `this`).

src/suites/cts/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+  t.device.defaultQueue.submit([encoder.finish()]);
+
+  t.checkCopyImageBitmapResult(
+    testBuffer,
+    data,
+    imageBitmap.width,
+    imageBitmap.height,
+    bytesPerPixel
+  );
+});
```
kainino0x: Let's go ahead and add a `.params(...)` so we can test many different widths and heights, so we can verify that the test code is working correctly for other cases.

src/suites/cts/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+  const width = 4;
+  const height = 4;
+  const bytesPerPixel = 4;
+  const imagePixels = new Uint8ClampedArray(bytesPerPixel * width * height);
+  for (let i = 0; i < width * height * bytesPerPixel; ++i) {
+    if (i % 4 === 3) {
+      imagePixels[i] = 255;
+      continue;
+    }
+    imagePixels[i] = i % 256;
```
kainino0x: nit:
```ts
// Generate an image with alpha always 255
imagePixels[i] = (i % 4 === 3) ? 255 : (i % 256);
```

src/suites/cts/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+      width: imageBitmap.width,
+      height: imageBitmap.height,
+      depth: 1,
+    }
+  );
+
+  const data = t.imageBitmapToData(imageBitmap);
+  if (data === undefined) {
+    return;
+  }
```
kainino0x: won't be needed anymore

src/suites/cts/copyImageBitmapToTexture.spec.ts
kainino0x: Oh, and the 'expect' won't be needed either. An exception will cause the test to fail.

src/suites/cts/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+          failedPixels++;
+          lines.push(`at [${indexExp}], expected ${exp[indexExp]}, got ${actual[indexActual]}`);
+        }
+      }
+      if (failedPixels > 4) {
+        lines.push('... and more');
+        break;
+      }
+    }
+    return this.handleErrorMessage(actual, exp, size, failedPixels, lines);
```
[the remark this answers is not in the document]
shaoboyan: Yes, currently it prints the actual with all padding 0s

src/suites/cts/copyImageBitmapToTexture.spec.ts
shaoboyan: Done.

src/suites/cts/copyImageBitmapToTexture.spec.ts
shaoboyan: Done.

src/suites/cts/copyImageBitmapToTexture.spec.ts
shaoboyan: Done.

src/suites/cts/copyImageBitmapToTexture.spec.ts
shaoboyan: Done.

src/suites/cts/copyImageBitmapToTexture.spec.ts
shaoboyan: Done.

shaoboyan:
BTW, the status of copyImageBitmapToTexture is I'm trying to add a blink unittest. (But it seems not that straight forward to create objects like GPUDevice)

src/suites/cts/gpu_test.ts
```diff
@@ -87,14 +87,9 @@ export class GPUTest extends Fixture {
     return this.device.createShaderModule({ code });
   }
 
-  // TODO: add an expectContents for textures, which logs data: uris on failure
-
-  expectContents(src: GPUBuffer, expected: ArrayBufferView): void {
-    const exp = new Uint8Array(expected.buffer, expected.byteOffset, expected.byteLength);
-
-    const size = expected.buffer.byteLength;
+  createAndMapBufferForReadingContents(src: GPUBuffer, size: number): GPUBuffer {
```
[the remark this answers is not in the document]
shaoboyan: Done.

src/suites/cts/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+    {
+      width: imageBitmap.width,
+      height: imageBitmap.height,
+      depth: 1,
+    }
+  );
+
+  const data = t.imageBitmapToData(imageBitmap);
+
+  // The texture format is r8uint, so the bytes per pixel is 4.
```
kainino0x:
```suggestion
  // The texture format is rgba8uint, so the bytes per pixel is 4.
```

src/suites/cts/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+copy imageBitmap To texture tests.
+`;
+
+import { TestGroup, pcombine, poptions } from '../../framework/index.js';
+
+import { GPUTest } from './gpu_test.js';
+
+function calculateRowPitch(width: number, bytesPerPixel: number): number {
+  const bytesPerRow = width * bytesPerPixel;
+  // Rounds up to a multiple of 256 according to WebGPU requests.
```
kainino0x:
```suggestion
  // Rounds up to a multiple of 256 according to WebGPU requirements.
```

src/suites/cts/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+
+  t.checkCopyImageBitmapResult(
+    testBuffer,
+    data,
+    imageBitmap.width,
+    imageBitmap.height,
+    bytesPerPixel
+  );
+}).params(
+  pcombine(poptions('width', [4, 15]), poptions('height', [1, 2]), poptions('bytesPerPixel', [4]))
```
kainino0x: nit: bytesPerPixel can't actually take any values other than 4 right now, so I'm just going to remove it from the params.

kainino0x:
Did those things, tested against a build of Chrome with your patch.

PR #96, opened 2019-12-13
roll types

shaoboyan:
Roll @webgpu/types from 0.018 to 0.019


shaoboyan:
@kainino0x @austinEng  PTAL.

PR #97, opened 2019-12-16
Add a fence test with time limit

szatkus:
Hi,

I implemented fences today and I discovered that empty `onCompletion` could pass the CTS.

I added a new test case which catches that problem.

Regards

### DIFF
--- a/src/framework/util/index.ts
+++ b/src/framework/util/index.ts
@@ -18,6 +18,10 @@ export function rejectOnTimeout(ms: number, msg: string): Promise<never> {
   });
 }
 
+export function raceWithRejectOnTimeout<T>(p: Promise<T>, ms: number, msg: string): Promise<T> {
+  return Promise.race([p, rejectOnTimeout(ms, msg)]);
+}
+
 export function objectEquals(x: unknown, y: unknown): boolean {
   if (typeof x !== 'object' || typeof y !== 'object') return x === y;
   if (x === null || y === null) return x === y;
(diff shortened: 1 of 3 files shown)


szatkus:
`Calls to 'setTimeout' are not allowed. WPT (Web Platform Tests) doesn't allow setTimeout. (ban)`

Oh.

kainino0x:
Thanks for the contribution! If you don't mind me asking, what WebGPU implementation are you working on? 🙂 

We have a helper to fix the timeout issue: `src/framework/util/timeout.ts`.
But we also have a helper just for tests like this: `rejectOnTimeout` from `src/framework/util/index.ts`. You should be able to use that one.

szatkus:
My side project: https://github.com/szatkus/kagepu
Basically WebGPU implemented using existing technologies. Quite fun and challenging :)

Thank you for you help. Looks like tests are passing, but Travis still doesn't let the PR. I don't see any meaningful error messages :/

szatkus:
gts was the problem. I forgot about semicolons :/

Funnily enough, when I ran gts on Windows (hey, it's good enough for personal projects) it butchered the whole project by replacing EVERY LF with CRLF. I needed to use my Linux laptop :)

kainino0x:
Oh boy. We absolutely want to support development on windows. There must be a way to configure gts to always use LF.

kainino0x:
> My side project: https://github.com/szatkus/kagepu
> Basically WebGPU implemented using existing technologies. Quite fun and challenging :)

Very cool, thank you for sharing! Sounds like a huge project - good luck!

After looking at this again, I decided to factor out a helper to wrap rejectOnTimeout and pushed it to your branch. LGTM - thanks again for the contribution!

kainino0x:
Hm, tested in chrome and I realized this doesn't quite work because it asks for onCompletion before signalling, which isn't valid. Let me fix this up and also add a test for that.

PR #105, opened 2020-01-07
Use new constants to eliminate !s; minor refactors, and (over)generalize a bunch of tests

kainino0x:


src/suites/cts/validation/createBindGroup.spec.ts
```diff
@@ -2,94 +2,36 @@ export const description = `
...
 g.test('binding count mismatch', async t => {
   const bindGroupLayout = t.device.createBindGroupLayout({
-    bindings: [
-      {
-        binding: 0,
-        visibility: GPUShaderStage.COMPUTE,
-        type: 'storage-buffer',
-      },
-    ],
+    bindings: [{ binding: 0, visibility: GPUShaderStage.COMPUTE, type: 'storage-buffer' }],
```
austinEng: Can `'storage-buffer'` be `C.BindingType.StorageBuffer` ?

src/suites/cts/format_info.ts
```diff
@@ -62,4 +55,22 @@ export const textureFormatInfo: TextureFormats = {
...
-export const textureFormatParams = Array.from(poptions('format', textureFormats));
+
+export const bindingTypeInfo: {
+  [k in GPUBindingType]: {
+    type: 'buffer' | 'texture' | 'sampler';
+    maxDynamicCount: number;
+    // Add fields as needed
+  };
+} = /* prettier-ignore */ {
+  'uniform-buffer':          { type: 'buffer',  maxDynamicCount: 8 },
```
austinEng: Can this be like `[C.BindingType.UniformBuffer]: { ... }` ?

src/suites/cts/validation/validation_test.ts
```diff
@@ -1,16 +1,103 @@
...
+        return { buffer: this.getErrorBuffer() };
+      } else if (objectType === 'sampler') {
+        return this.getErrorSampler();
+      } else if (objectType === 'texture') {
+        return this.getErrorTextureView();
+      } else {
+        unreachable('unknown binding resource type');
+      }
+    } else {
+      if (bindingType === 'uniform-buffer') {
```
austinEng: Can these use the const enums?

src/suites/cts/validation/validation_test.ts
```diff
@@ -1,16 +1,103 @@
+import { unreachable } from '../../../framework/index.js';
+import { bindingTypeInfo } from '../format_info.js';
 import { GPUTest } from '../gpu_test.js';
 
+export enum BindingResourceType {
+  'uniform-buffer' = 'uniform-buffer',
```
austinEng: Maybe not worth it, but can this be `'uniform-buffer' = C.BindingType.UniformBuffer` ?'

Although it looks like we only use this as a type annotation and to do Object.keys(BindingResourceType).

Would it work to use the `C.BindingType` type annotation everywhere, and then have `export const BindingResourceTypes = [C.BindingType.UniformBuffer, ... ]` ?

src/suites/cts/validation/createBindGroup.spec.ts
kainino0x: Could, but I don't think it's helpful. These keys are already tightly typed by `[k in GPUBindingType]` so typos aren't possible. In fact, that forces us to be complete too (no missing variants). Also the original string values are shorter and I think easier to read.

src/suites/cts/format_info.ts
kainino0x: Could, but I don't think it's helpful. These keys are already tightly typed by `[k in GPUBindingType]` so typos aren't possible. In fact, that forces us to be complete too (no missing variants). Also the original string values are shorter and I think easier to read.

src/suites/cts/validation/validation_test.ts
kainino0x: Yes, and have to if I do the next suggestion

src/suites/cts/validation/validation_test.ts
kainino0x: Good idea, will tighten the type from `string`.

kainino0x: hm, actually this is kind of annoying: BindingResourceType is still not a subtype of GPUBindingType if you do that, so can't index `bindingTypeInfo` with it.

src/suites/cts/validation/validation_test.ts
kainino0x: I guess I won't do that one, so I won't do this one for the same reason. These are already typo guarded:

```
This condition will always return 'false' since the types 'BindingResourceType.storage-texture' and '"storage-bexture"' have no overlap.ts(2367)
```

src/suites/cts/validation/createBindGroup.spec.ts
kainino0x: yes, though it doesn't make any difference in this context (inference can determine this is a GPUBindingType so it's already type/typo-checked) and only really serves to make it longer

kainino0x:
> Find it a bit unfortunate that we have to copy the `constants.js` in this way :-/

Better solution would be great, but I don't have any good ideas on how unless we take on a bundler just for out-of-src/ dependencies.

kainino0x:
I'll merge now, we can talk about whether I'll make any of these changes as followups

PR #111, opened 2020-01-11
Assorted fixes for WPT, workers

kainino0x:
- `?q=cts:` fix fixes the default generated out-wpt/cts.html (i.e. the one exported to WPT)
    Tested in a local roll into actual WPT with `localhost:8000/tools/runner/index.html`.
- improve grunt wpt
- fix incorrect type assertion of a transferred object, and manually fix its prototype chain
- rename .worker.js to -worker.js so WPT won't try to run it as a standalone worker test
- copy constants.js into out-wpt/


kainino0x:
TBR= @austinEng 
These are fairly uninteresting changes, landing now to see if I can run a chromium dry run

PR #115, opened 2020-01-18
Don't expect storage buffers to work in vertex stage

### LINKED ISSUE #113: createPipelineLayout tests expect storage buffers to work in vertex stage
I think I made a mistake when generalizing these tests to cover more cases.

- `cts:validation/createPipelineLayout:dynamic_offsets_are_only_allowed_on_buffers~`
- `cts:validation/createPipelineLayout:number_of_bind_group_layouts_exceeds_the_maximum_value~`

kainino0x:
Fixes #113

### DIFF
--- a/src/suites/cts/validation/createPipelineLayout.spec.ts
+++ b/src/suites/cts/validation/createPipelineLayout.spec.ts
@@ -3,12 +3,7 @@ createPipelineLayout validation tests.
 `;
 
 import { TestGroup, pbool, pcombine, poptions } from '../../../framework/index.js';
-import {
-  kBindingTypeInfo,
-  kBindingTypes,
-  kShaderStageCombinations,
-  kShaderStages,
-} from '../capability_info.js';
+import { kBindingTypeInfo, kBindingTypes, kShaderStageCombinations } from '../capability_info.js';
 
 import { ValidationTest } from './validation_test.js';
 
--- a/src/suites/cts/validation/createPipelineLayout.spec.ts
+++ b/src/suites/cts/validation/createPipelineLayout.spec.ts
@@ -94,10 +94,8 @@ g.test('visibility and dynamic offsets', t => {
 );
 
 g.test('number of bind group layouts exceeds the maximum value', async t => {
-  const visibility: GPUShaderStageFlags = t.params.visibility;
-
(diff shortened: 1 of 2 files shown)


src/suites/cts/validation/createPipelineLayout.spec.ts
```diff
@@ -61,39 +61,38 @@ g.test('number of dynamic buffers exceeds the maximum value', async t => {
...
   )
 );
 
 g.test('number of bind group layouts exceeds the maximum value', async t => {
-  const { visibility, type } = t.params;
+  const type: GPUBindingType = t.params.type;
 
   const bindGroupLayoutDescriptor: GPUBindGroupLayoutDescriptor = {
-    bindings: [{ binding: 0, visibility, type }],
+    bindings: [{ binding: 0, visibility: C.ShaderStage.Fragment, type }],
```
kvark: might add a comment that we'll still need to test the read-only storage in VS stage once it's in

kvark: also, why discriminate against compute stage? :)

austinEng: It seems that most of the binding count limits with all permutations of usage and visibility happens in createBindGroupLayout.spec.ts `max resources per stage/in pipeline layout` and `max resources per stage/in bind group layout`.

So if the purpose of the test is `number of bind group layouts exceeds the maximum value`, then I question whether we need any `bindings:` and can instead just make a pipelineLayout with 4, and then 4 + 1 empty bind group layouts.

kainino0x:
TODO: `kBindingTypeInfo['readonly-storage-buffer'].validStages` should allow vertex shader

src/suites/cts/validation/createPipelineLayout.spec.ts
kainino0x: yes, and in fact I guess it doesn't really make sense to test multiple binding types here. Maybe the best thing is to test multiple values for `visibility` and just one for the binding type.

kainino0x: Also, IIRC the change to this test is unrelated to the title of the PR, sorry (rushed this PR out at end of day)

kainino0x:
PTAL

PR #116, opened 2020-01-20
Remove document usage in copyImageBitmapToTexture tests

shaoboyan:
Use raw data as expectation value to remove dependent of document
element. This change will enable this test on worker thread.


shaoboyan:
@kainino0x  sry for backing to this so late. PTAL if you have time.

kainino0x:
No problem! I'll review after the US holiday.

shaoboyan:
Got it and agree :)

Issue #117, opened 2020-02-02
Node integration


maierfelix:
Hey,

I'd like to use this for the node bindings - @kainino0x could you take a look into how the cts could be integrated? Need some initial steps to get started

Thanks

kainino0x:
Hey Felix, unfortunately this can't be a priority for me so it'll be at least 2 weeks before I can consider it. I think it will be pretty quick for you to do it - and easier, since you already have builds.

All you need to do is get the getGPU() function to return an object that implements requestAdapter:
https://github.com/gpuweb/cts/blob/master/src/framework/gpu/implementation.ts

I used to have support in there for my proof-of-concept node bindings (that had like 5 entry points implemented). You can look at how it was done then as an example:
https://github.com/gpuweb/cts/blob/26b96dac48eeaee39ccb62c905b34f59cfd74129/src/framework/gpu/implementation.ts

maierfelix:
Thanks for the useful information, I will try

dcerisano:
Hey guys, I am working on integrating Kai's `CTS` and Felix's `node-webgpu`.
I see that the official webgpu spec is imported into `CTS` from the `@webgpu `package.

Seems that `@webgpu` is currently designed for chromium and so does not provide for the stand-alone (addon) requirement for a window, since chromium provides one internally.

Felix extended the addon API by providing a WebGPUWindow.
He also changed the GPU.requestAdapter API by passing a GPUWindow there instead of GPURequestAdpaterOptions as specified by @webgpu.

So I think we need to add GPUWindow to the @webgpu spec and also add WebGPUWindow as in GPURequestAdapterOptions. That would be one way of satisfying the requirement for both chromium and stand-alone modes in the official @webgpu API.

Basically just need a way to specify the window context if none exists. Again in chromium of course one always exists.

Could also just overload the requestAdapter function to do the window creation internally, I suppose.

My test branch where I am working on this is [here](https://github.com/dcerisano/cts/tree/test-branch), a clone of Kai's experimental tree he gave above.

dcerisano:
Hey guys, I am working on integrating Kai's CTS and Felix's node-webgpu while Felix is completing his super awesome node-webgpu RT demos.

I see that the official webgpu spec is imported into CTS from the @webgpu package. Seems that @webgpu is currently designed mostly for browser support and so does not provide for the stand-alone (eg. addon) requirement to set a window, since browsers provide one internally.

[...]

PR #122, opened 2020-02-19
Use only one GPUDevice to run tests

kainino0x:
Creates a "DevicePool" which has one device, lazily initialized, and
hands it out to exclusive borrowers to ensure there's never more than
one test running.

Later this will probably be extended to have some small number of
devices (like 8) and be able to hand out more than one at a time.

cc #117


kainino0x:
TBR @austinEng

kainino0x:
That's a good point; I did these out of order. At this point, though, I think I'll just wait to fix this until we're actually working on running things in parallel.

PR #124, opened 2020-02-27
Add WGSL folder and test cases

sarahM0:


Kangz:
Would there be a way to have markers in the file so that two test cases can be extracted from it? One of the case would be valid, and the other would be invalid, with minimal differences between the two. This way we know the invalid cases are still invalid because they test the correct thing, not because we forgot to update them for a syntax change.

sarahM0:
> Would there be a way to have markers in the file so that two test cases can be extracted from it? One of the case would be valid, and the other would be invalid, with minimal differences between the two. This way we know the invalid cases are still invalid because they test the correct thing, not because we forgot to update them for a syntax change.

So for example one validation rule says the only acceptable import is GLSL.std.450, so a test file should look like this?

_PASS_MARKER_BEGIN
import "GLSL.std.450" as glsl;
PASS_MARKER_END

FAIL_MARKER_BEGIN
import "GLSL.std.330" as glsl;
FAIL_MARKER_END_

Sounds like a good idea. I'll go over all the rules to make sure writing a test file with markers is possible.

dj2:
With the validation tests, that information is encoded in the test name. The `v-0001` maps back to a validation error in the WGSL spec. So, you should be able to determine why a test fails based on the name and see if that's the cause of the failure in your test.

dj2:
I'd prefer to avoid the pass/fail markers as those are going to make the tests more brittle.

Kangz:
> With the validation tests, that information is encoded in the test name. The `v-0001` maps back to a validation error in the WGSL spec. So, you should be able to determine why a test fails based on the name and see if that's the cause of the failure in your test.

Sorry I wasn't clear. I am worried that we will forget to update the failure tests and that they will be failing for reasons unrelated to what they try to test. How do we prevent that? (being careful about it isn't a very satisfying solution imho).

sarahM0:
> > With the validation tests, that information is encoded in the test name. The `v-0001` maps back to a validation error in the WGSL spec. So, you should be able to determine why a test fails based on the name and see if that's the cause of the failure in your test.
> 
> Sorry I wasn't clear. I am worried that we will forget to update the failure tests and that they will be failing for reasons unrelated to what they try to test. How do we prevent that? (being careful about it isn't a very satisfying solution imho).

Can we add the expected error message as a comment to each fail file so the test script can do a diff between the encountered validation and what was expected (ie. the comment).

dj2:
The error messages can differ per implementation, but the ID should, hopefully be stable.  The plan for tint was to have the test runner verify that the `v-yyyy` value output matches the prefix of the test case name.

@Kangz did you have a syntax in mind for the markers you were thinking of?

Kangz:
Ah, we talked about error IDs in the group for the API but the discussion stalled. I don't think we should rely on having them in the future.

As for syntax, I was thinking about C preprocessor like macros like `#if error/success` `#endif`. It doesn't need to be a full preprocessor, just matching the line with one of these three tokens. No strong opinions though.

dj2:
Heh, WGSL _already_ has IDs for all validation messages. They're in the spec now, I'm strongly in favour of keeping them as it makes it a lot easier to track if you've covered all the validation cases.

sarahM0:
I put this PR on hold until we resolve issue #125

sarahM0:
Ready for review.

src/suites/cts/wgsl/validation/continue-outside-for.fail.wgsl
```diff
@@ -0,0 +1,8 @@
+# v0010 - This fails because the continue is used outside of a for block.
```
kainino0x:
```suggestion
# v-0010 - This fails because the continue is used outside of a for block.
```

src/suites/cts/wgsl/validation/import-non-glsl.fail.wgsl
```diff
@@ -0,0 +1,10 @@
+# v0001 - This test fails as only "GLSL.std.450" is permitted in as import
```
kainino0x:
```suggestion
# v-0001 - This test fails as only "GLSL.std.450" is permitted in as import
```

src/suites/cts/wgsl/validation/no-enty-point-declared.fail.wgsl
```diff
@@ -0,0 +1,6 @@
+# v0003 - This fails because there must be at least one function with a vertex,
```
kainino0x:
```suggestion
# v-0003 - This fails because there must be at least one function with a vertex,
```

src/suites/cts/wgsl/validation/duplicate-entry-point.fail.wgsl
```diff
@@ -0,0 +1,13 @@
+# v-0020 -  Duplicate entry point <name, stage> pair.
```
kainino0x:
```suggestion
# v-0020 - Duplicate entry point <name, stage> pair.
```

PR #129, opened 2020-03-09
CTS of CopyImageBitmap to cover canvas path

shaoboyan:
This patch added a test path to cover CopyImageBitmap from canvas source which is mainly the GPU copy path.

### DIFF
--- a/src/suites/cts/copyImageBitmapToTexture.spec.ts
+++ b/src/suites/cts/copyImageBitmapToTexture.spec.ts
@@ -73,7 +73,7 @@ class F extends GPUTest {
     return failedPixels > 0 ? lines.join('\n') : undefined;
   }
 
-  DoTestAndCheckResult(
+  doTestAndCheckResult(
     imageBitmapCopyView: GPUImageBitmapCopyView,
     dstTextureCopyView: GPUTextureCopyView,
     copySize: GPUExtent3D,
@@ -139,22 +139,11 @@ g.test('from ImageData', async t => {
     format: 'rgba8unorm',
     usage: GPUTextureUsage.COPY_DST | GPUTextureUsage.COPY_SRC,
   });
-  t.DoTestAndCheckResult(
-    {
-      imageBitmap,
-      origin: {
-        x: 0,
-        y: 0,
-      },
-    },
-    {
-      texture: dst,
-    },
-    {
-      width: imageBitmap.width,
-      height: imageBitmap.height,
-      depth: 1,
-    },
+
+  t.doTestAndCheckResult(
+    { imageBitmap, origin: { x: 0, y: 0 } },
+    { texture: dst },
+    { width: imageBitmap.width, height: imageBitmap.height, depth: 1 },
     bytesPerPixel,
     imagePixels
   );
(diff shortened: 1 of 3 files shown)


shaoboyan:
@kainino0x in this cts I need to add a package offscreenCanvas, and it seems this will break the CI.  How can I handle this?

kainino0x:
Run `yarn install`, which will change the `yarn.lock` file, then check that in too.

src/suites/cts/copyImageBitmapToTexture.spec.ts
```diff
@@ -146,3 +146,97 @@ g.test('from ImageData', async t => {
...
+  // Generate original data.
+  const imagePixels = new Uint8ClampedArray(bytesPerPixel * width * height);
+  for (let i = 0; i < width * height * bytesPerPixel; ++i) {
+    imagePixels[i] = i % 256;
+  }
+
+  const imageData = new ImageData(imagePixels, width, height);
+
+  // CTS works on workder thread sometime where doucment
+  // element is not available. In such situation, offscreenCanvas
```
kainino0x:
```suggestion
  // CTS works on worker threads sometimes, where document
  // is not available. In such situations, OffscreenCanvas
```

src/suites/cts/copyImageBitmapToTexture.spec.ts
```diff
@@ -146,3 +146,97 @@ g.test('from ImageData', async t => {
...
+  // can be the candidate. But not all browsers support creating
+  // 2d context on offscreenCanvas. In this situation, the case will
+  // be skipped.
```
kainino0x: I believe Safari doesn't support OffscreenCanvas at all, so you'll have to check for `typeof OffscreenCanvas === 'undefined'` too.

(This will fail on Safari anyway because they do not have ImageBitmap. But we should still skip the test if OffscreenCanvas isn't supported.)

src/suites/cts/copyImageBitmapToTexture.spec.ts
```diff
@@ -146,3 +146,97 @@ g.test('from ImageData', async t => {
...
+  }
+  const imageCanvasContext = imageCanvas.getContext('2d');
+  if (imageCanvasContext === null) {
+    t.skip('OffscreenCanvas "2d" context not available');
+    return;
+  }
+
+  imageCanvasContext.putImageData(imageData, 0, 0);
+
+  const imageBitmap = await createImageBitmap(imageCanvas);
```
kainino0x: A lot of the code below this point is very similar or the same as the previous test. Can we factor it out?

shaoboyan: factor it out.

src/suites/cts/copyImageBitmapToTexture.spec.ts
shaoboyan: Done.

src/suites/cts/copyImageBitmapToTexture.spec.ts
shaoboyan: Oops... Done.

PR #131, opened 2020-03-11
Fix call to devicePool.release

kainino0x:
sigh, can't get this right.

TBR @austinEng


kainino0x:
> LGTM; sorry I probably should have noticed this the first time as well.

The bug is caused by an existing place where I bypass the type checking (this.device is initialized to `undefined!`) so it was very hard to notice. I should actually improve that.

PR #132, opened 2020-03-11
Fix a race in init of GLSL tests, plus strengthen assert

kainino0x:


src/suites/cts/validation/createRenderPipeline.spec.ts
```diff
@@ -10,7 +10,8 @@ import { ValidationTest } from './validation_test.js';
 
 class F extends ValidationTest {
   async init(): Promise<void> {
-    await Promise.all([super.init(), this.initGLSL()]);
```
kainino0x: These were causing flakes when running without glslang: `Promise.all` is racy if any of the input promises rejects. `initGLSL` rejects when glslang is not available, so `F.init` would return before `super.init()` was finished.

kainino0x:
TBR @austinEng. Still debugging my way through other flakes.

src/suites/cts/validation/createRenderPipeline.spec.ts
kainino0x: Somehow these particular flakes might be new since the last roll. They were blocking my roll on both Windows and Linux (not Mac for some reason) but I don't see them when looking at other logs on the bots.

Fortunately these flakes were easily reproducible locally (on Mac!) unlike the rest of our flakes....

kainino0x: Of the three I fixed here, this one was the only one that was already significantly [flaky](https://analysis.chromium.org/p/chromium/flake-portal/flakes/occurrences?key=ag9zfmZpbmRpdC1mb3ItbWVyQAsSBUZsYWtlIjVjaHJvbWl1bUBibGlua193ZWJfdGVzdHNAd3B0X2ludGVybmFsL3dlYmdwdS9jdHMuaHRtbAw&show_all_occurrences=1).

austinEng: Ah, awesome glad this helps a bunch

PR #135, opened 2020-03-12
Allow overriding of the path to glslang

kainino0x:
This potential change allows browsers to specify a path to a file outside WPT (which can
e.g. be downloaded as a dependency instead of checked into Git). This
would allow testing of runtime-generated shaders through WPT.


kainino0x:
Closing to re-open on a branch

PR #136, opened 2020-03-12
Replace textureDimension with viewDimension

romandev:
The member for the texture-view-dimension is of type
GPUTextureViewDimension, but the name is textureDimension, which is
confusing because there's also a type GPUTextureDimension[1][2].

[1] https://github.com/gpuweb/gpuweb/issues/587
[2] https://github.com/gpuweb/gpuweb/pull/589


kainino0x:
This will require a fix in https://github.com/gpuweb/types first. There might be other things that need updating there too, I'm not sure.

romandev:
FYI, I've just uploaded PR: https://github.com/gpuweb/types/pull/13

kainino0x:
I'm going to do the full types roll as a separate PR. Pulled your commit in there, thanks for taking care of that part!

PR #137, opened 2020-03-18
Roll @webgpu/types to latest, and update tests

kainino0x:
Also update the capability tables to include the new binding types and the correct restrictions on writable storage buffers and textures.

Haven't run these tests yet because we have yet to implement these changes. But soon.

### DIFF
--- a/src/suites/cts/capability_info.ts
+++ b/src/suites/cts/capability_info.ts
@@ -80,8 +80,8 @@ export const kPerStageBindingLimits: {
   'storage-texture': 4,
 };
 
-const kStagesVFC = C.ShaderStage.Vertex | C.ShaderStage.Fragment | C.ShaderStage.Compute;
-const kStagesC = C.ShaderStage.Compute;
+const kStagesAll = C.ShaderStage.Vertex | C.ShaderStage.Fragment | C.ShaderStage.Compute;
+const kStagesCompute = C.ShaderStage.Compute;
 export const kBindingTypeInfo: {
   [k in GPUBindingType]: {
     type: 'buffer' | 'texture' | 'sampler';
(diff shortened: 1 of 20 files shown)


src/suites/cts/capability_info.ts
```diff
@@ -80,8 +80,8 @@ export const kPerStageBindingLimits: {
   'storage-texture': 4,
 };
 
-const kStagesAll = C.ShaderStage.Vertex | C.ShaderStage.Fragment | C.ShaderStage.Compute;
-const kStagesNonVertex = C.ShaderStage.Fragment | C.ShaderStage.Compute;
+const kStagesVFC = C.ShaderStage.Vertex | C.ShaderStage.Fragment | C.ShaderStage.Compute;
```
austinEng: nit: Would prefer kStagesAll and kStageCompute. The letters get hard to parse and overloaded since we import the constants as `C`

src/suites/cts/validation/render_pass_descriptor.spec.ts
```diff
@@ -29,9 +29,8 @@ class F extends ValidationTest {
     } = options;
 
     return this.device.createTexture({
-      size: { width, height, depth: 1 },
+      size: { width, height, depth: arrayLayerCount },
```
austinEng: side thought: Maybe we can have separate dictionaries for 3D vs 2D Array vs 1D Array, etc. 
Using: `size: [x, y, z]` is "easy mode" where there's no extra validation constraints
But using `size: { width: w, height: h, depth: d }` or `size: { width: w, height: h, layers: count }` or `size: { width: w, layers: count }` will incur extra validation that you used `3d` or `2d` or `1d`

That probably means the webidl would need to take `object` though..

kainino0x: I think the validation rules on the two should be the same. The array syntax should just be a shorthand for the object syntax. So either we allow:
`{width}` or `[width]` or `{width, height}` or `[width, height]` or `{width, height, depth}` or `[width, height, depth]`
or we allow only:
`{width, height, depth}` or `[width, height, depth]`

And then any validation becomes orthogonal.
We would have to have no defaults on height/depth, instead of defaulting to 0 like GPUOrigin3DDict.

Objects with optional fields always come with the hazard that any typos in the names of those fields are silently ignored. I still want strict rules for all our objects (no unknown fields allowed), but it would be a mess to implement.

Personally I think the object versions aren't useful and we should have only the array versions, but other folks didn't agree.

PR #138, opened 2020-03-18
Remove a stray setSubData call

kainino0x:


kainino0x:
Missed this because the types allow setSubData and I haven't been able to find a way to mark it deprecated. We'll probably have to change this again soon anyway. Merging TBR, tested on Chrome.

austinEng:
LGTM

PR #139, opened 2020-03-24
Add initial robust access tests.

Kangz:
PTAL! I'm not super fluent in Typescript, so I'm sure there will be a lot of things to fix!


src/suites/cts/robust_access.spec.ts
```diff
@@ -0,0 +1,364 @@
...
+  const pass = encoder.beginComputePass();
+  pass.setPipeline(pipeline);
+  pass.setBindGroup(0, testGroup);
+  pass.setBindGroup(1, group);
+  pass.dispatch(1);
+  pass.endPass();
+
+  t.queue.submit([encoder.finish()]);
+
+  await t.expectContents(resultBuffer, new Uint32Array([0]));
```
kainino0x: expectContents isn't async.
```suggestion
  t.expectContents(resultBuffer, new Uint32Array([0]));
```

src/suites/cts/robust_access.spec.ts
```diff
@@ -0,0 +1,364 @@
...
+const kUintMax = 4294967295;
+const kIntMax = 2147483647;
+
+// A small utility to test shaders:
+//  - it wraps the source into a small harness that checks the runTest() function returns 0.
+//  - it runs the shader with the testBindings set as bindgroup 0.
+//
+// The shader also has access to a uniform value that's equal to 0u to avoid constant propagation
+// in the shader compiler.
+async function runShaderTest(
```
kainino0x:
```suggestion
function runShaderTest(
```
don't think async is needed

src/suites/cts/robust_access.spec.ts
```diff
@@ -0,0 +1,364 @@
...
+    // Check with max uint
+    `${kUintMax} * one`,
+    `-1 * ${kUintMax} * one`,
+
+    // Check with max int
+    `${kIntMax} * one`,
+    `-1 * ${kIntMax} * one`,
+  ];
+
+  const testSource = [];
```
kainino0x: optional nit: building this as a string would be fine, no strong need to build it as an array and join at the end.

src/suites/cts/robust_access.spec.ts
```diff
@@ -0,0 +1,364 @@
...
+
+g.test('bufferMemory', async t => {
+  await t.initGLSL();
+
+  const type = typeParams[t.params.type];
+  const baseType = type.baseType;
+
+  const indicesToTest = [
+    // Check exact bounds
+    `-1 * one`,
```
kainino0x: nit: could move the `* one` out from here to the top of the indicesToTest loop (`const index = '(' + indexValue + ') * one';`)

Also, should we test with constants as well (i.e. omitting the `* one`)?

src/suites/cts/robust_access.spec.ts
```diff
@@ -0,0 +1,364 @@
...
+  // Build the test function that will do the tests.
+  testSource.push(`
+    uint runTest() {
+  `);
+
+  for (const index of indicesToTest) {
+    if (t.params.access === 'read') {
+      testSource.push(`
+          if(data[${index}] != ${type.zero}) {
+            return __LINE__;
```
kainino0x: nit: Might be slightly confusing because the __LINE__ doesn't really map to anything interesting since the GLSL is generated. How about 0x0BAD or something just to make it ctrl-f-able?

src/suites/cts/robust_access.spec.ts
Kangz: Isn't building a string with a bunch of concatenations a quadratic operations?

src/suites/cts/robust_access.spec.ts
Kangz: Did the test with constants but GLSLang yells about the OOB access, so I think we can't do it for now.

Kangz:
Note that the tests don't pass on the machine I tried, because SPIRV-Cross doesn't support atomics on vector elements (the MSL compiler complains about taking the address of a vector element).

More importantly the test found something that might require relaxing slightly the robust access rules. For the matNx3 tests, the last columns writes to the 4 floats instead of just 3, even if the clamping is done correctly in the MSL code:

```
static inline __attribute__((always_inline))
uint runTest(device TestData& v_15, constant Constants& v_21)
{
    v_15.data[uint(clamp(int(0u * v_21.one), 0, 1))] = float3(0.0);
    v_15.data[uint(clamp(int(1u * v_21.one), 0, 1))] = float3(0.0);
    v_15.data[uint(clamp(int(4294967295u * v_21.one), 0, 1))] = float3(0.0);
    v_15.data[uint(clamp(int(2u * v_21.one), 0, 1))] = float3(0.0);
    v_15.data[uint(clamp(int(4293967296u * v_21.one), 0, 1))] = float3(0.0);
    v_15.data[uint(clamp(int(1000000u * v_21.one), 0, 1))] = float3(0.0);
    v_15.data[uint(clamp(int(4294967295u * v_21.one), 0, 1))] = float3(0.0);
    v_15.data[uint(clamp(int(1u * v_21.one), 0, 1))] = float3(0.0);
    v_15.data[uint(clamp(int(2147483647u * v_21.one), 0, 1))] = float3(0.0);
    v_15.data[uint(clamp(int(2147483649u * v_21.one), 0, 1))] = float3(0.0);
    return 0u;
}
```

 I think the compiler might see consecutive memory accesses and optimizing them to a single 16 byte memory access. @dneto0 WDYT?

src/suites/cts/robust_access.spec.ts
kainino0x: I don't think it matters here, but my understanding is that JS engines have optimizations specifically for string-builder patterns like that.

src/suites/cts/robust_access.spec.ts
kainino0x: Ack

src/suites/cts/robust_access.spec.ts
Kangz: This one is supposed to help debug which test failed. I'd like to make a nicer shader test harness at some point that reports errors with a string.

Kangz:
Addressed all comments but we can't merge until I rebase on top of the types changes, and that requires changes in Blink to land first.

src/suites/cts/robust_access.spec.ts
kainino0x: Makes sense

kainino0x:
Rebased/squashed; closing this and reopening against the new branch

PR #140, opened 2020-03-30
Texture zero init tests

austinEng:
Hey @kainino0x, I've got this to the point of "it works" and submitting this PR now to get some early feedback. Still needs a lot of cleanup so please don't spend too much time on smaller details.

Overall:
 - I have a giant nested-nested[...]-nested for-loop generator function to produce the params which filters out invalid combinations.
- Creation vs StoreOp Uninitialize parameter: Determines whether to just create a texture and check it, or to create it, initialize it with some value we should never see, do a StoreOp.Clear, and then check it.
- "Canary" values are used on some subset of a texture's subresources that we're not testing zero-init behavior for. Zero init of a texture's uninitialized subresources should not affect the already-initialized ones.
- ReadMethod parameter: Determines how to read data back. Sampling or Copy or Storage, etc...
  - Depth/StencilTest: Self-explanatory. Uses an "equal" compare function.
  - Sample: Generate a shader to sample the texture, and write out both the sampled and expected shader-interpreted values into buffers (right now I'm only checking the red component. TODO others).
  - Copy: Self explanatory. But how do you determine the "correct" value? Before the test runs, I render to a dummy texture and read back the result to determine the bytes a particular value is supposed to map to. I manually compute the byte format for non-renderable textures.. Probably need to worry about endianness though :-/ 
   - Storage: Not done yet.


austinEng:
I could probably remove the complexity of doing a render to figure out the bytes for a particular format if I always set / readback the canary values with a copyToBufferToTexture / copyTextureToBuffer and not cared what the actual contents were so long as what was originally set and readback are the same.

src/suites/cts/textures/zero_init.spec.ts
```diff
@@ -0,0 +1,1244 @@
...
+            stencilLoadValue: C.LoadOp.Load,
+          }
+        }).endPass();
+      }
+    }
+    this.device.defaultQueue.submit([commandEncoder.finish()]);
+  }
+}
+
+function* generateParams(): Generator<TestParams> {
```
kainino0x: didn't review this in depth but it looks very reasonable. It got me thinking about whether we could possibly express this in a more functional way (more in the style of combine and friends). How about:

```js
new Params()
  .combine(
    options('format', kTextureFormats), //
    options('aspect', kTextureAspects))
  .unless(p => p.aspect === 'depth-only' && !kTextureFormatInfo[p.format].depth)
  .unless(p => p.aspect === 'stencil-only' && !kTextureFormatInfo[p.format].stencil)
  .combine(
    kCreationSizes, //
    options('mipLevelCount', kMipLevelCounts),
    options('sampleCount', kSampleCounts))
  .unless(p => p.sampleCount > 1 && p.mipLevelCount > 1)
  .combine(
    options('uninitializeMethod', kUninitializeMethods)), //
    options('readMethod', kReadMethods))
  .unless(p => {
    const usage = getRequiredTextureUsage(sampleCount, uninitializeMethod, readMethod);
    ...
  })
  ...
```

WDYT? Though I know you tried to make nice types for something similar and it didn't seem to work out :(

Also probably we should have helpers for things like `formatSupportsAspect` and `formatSupportsUsage`.

src/suites/cts/textures/zero_init.spec.ts
```diff
@@ -0,0 +1,1244 @@
...
+      default:
+        assert(false);
+    }
+  }
+
+  initializeTexture(texture: GPUTexture,
+                    state: InitializedState,
+                    subresourceRange: SubresourceRange) {
+    if (this.params.sampleCount > 1) {
+      // Copies to multisampled textures unspecified.
```
kainino0x:
```suggestion
      // Copies to multisampled textures are not allowed.
```

src/suites/cts/textures/zero_init.spec.ts
```diff
@@ -0,0 +1,1244 @@
...
+      assert(kTextureFormatInfo[this.params.format].renderable);
+      this.initializeWithStoreOp(state, texture, this.params.format, this.params.aspect, subresourceRange);
+      return;
+    }
+
+    switch (this.params.dimension) {
+      case '1d':
+      case '3d':
+        // TODO: https://github.com/gpuweb/gpuweb/issues/69
+        // Copies with 1D and 3D textures are unspecified
```
kainino0x:
```suggestion
        // Copies with 1D and 3D textures are not yet specified
```
(although we'll have them soon)

src/suites/cts/textures/zero_init.spec.ts
```diff
@@ -0,0 +1,1244 @@
...
+    await this.initGLSL();
+    this.gpuDataRepresentation = await this.getGPUDataRepresentation();
+  }
+
+  get params(): TestParams {
+    return super.params as TestParams;
+  }
+
+  get textureWidth(): number {
+    return 1 << this.params.mipLevelCount;
```
kainino0x: possibly we should be able to test NPOT textures

src/suites/cts/textures/zero_init.spec.ts
```diff
@@ -0,0 +1,1244 @@
...
+      sampleCount,
+    });
+    return this.stencilTestPipelineCache[key];
+  }
+
+  checkContents(texture: GPUTexture,
+                state: InitializedState,
+                subresourceRange: SubresourceRange) {
+    switch (this.params.readMethod) {
+      case ReadMethod.CopyToBuffer: {
```
kainino0x: Maybe separate the body of each case into a method

src/suites/cts/textures/zero_init.spec.ts
```diff
@@ -0,0 +1,1244 @@
...
+              format: this.params.format,
+              usage: C.TextureUsage.CopyDst | C.TextureUsage.CopySrc,
+            });
+
+            const commandEncoder = this.device.createCommandEncoder();
+            commandEncoder.copyTextureToTexture(
+              { texture, mipLevel: level, arrayLayer: layer },
+              { texture: dst, mipLevel: 0, arrayLayer: 0 },
+              { width, height, depth: 1 },
+            );
```
kainino0x: Didn't actually read this code, but after this point shouldn't this be able to share code with the CopyToBuffer case?

src/suites/cts/textures/zero_init.spec.ts
```diff
@@ -0,0 +1,1244 @@
...
+            attachment: texture.createView(viewDescriptor),
+            depthStoreOp: C.StoreOp.Store,
+            depthLoadValue: initializedStateAsDepth(state),
+            stencilStoreOp: C.StoreOp.Store,
+            stencilLoadValue: initializedStateAsStencil(state),
+          },
+        }).endPass();
+      }
+    }
+    this.device.defaultQueue.submit([commandEncoder.finish()]);
```
kainino0x: I think we also have `this.queue`

src/suites/cts/gpu_test.ts
```diff
@@ -179,15 +191,18 @@ export class GPUTest extends Fixture {
...
+
     const size = exp.byteLength;
     if (actual.byteLength !== size) {
       return 'size mismatch';
     }
     const lines = [];
     let failedPixels = 0;
     for (let i = 0; i < size; ++i) {
-      if (actual[i] !== exp[i]) {
+      if (Math.abs(actual[i] - exp[i]) > tolerance) {
```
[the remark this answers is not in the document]
Kangz: Maybe checkBuffer should take a comparison function that return Ok or Err(some debug string)?

src/suites/cts/capability_info.ts
```diff
@@ -6,60 +6,82 @@ export const kTextureFormatInfo: {
...
+})();
+
+export const kTextureAspects = (() => {
+  const kTextureAspectsObject: { [k in GPUTextureAspect]: GPUTextureAspect } = {
+    [C.TextureAspect.All]: C.TextureAspect.All,
+    [C.TextureAspect.DepthOnly]: C.TextureAspect.DepthOnly,
+    [C.TextureAspect.StencilOnly]: C.TextureAspect.StencilOnly,
+  };
+  return Object.values(kTextureAspectsObject);
+})();
```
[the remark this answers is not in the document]
austinEng: `[k in GPUTextureAspect]: {...}` is type safety that all of them are specified.
saying `export const kTextureAspects = [C.TextureAspect.All, ... ]` could miss enums if they get added.

src/suites/cts/gpu_test.ts
```diff
@@ -157,12 +163,18 @@ export class GPUTest extends Fixture {
...
 
-  expectContents(src: GPUBuffer, expected: ArrayBufferView): void {
-    const exp = new Uint8Array(expected.buffer, expected.byteOffset, expected.byteLength);
+  expectContents(src: GPUBuffer, expected: (
+    Int8Array | Int16Array | Int32Array |
+    Uint8Array | Uint16Array | Uint32Array |
+    Float32Array | Float64Array
+  )): void {
+    // @ts-ignore
+    const exp = new expected.constructor(expected.buffer, expected.byteOffset, expected.length);
```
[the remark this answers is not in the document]
austinEng: I found this really helpful so that failures are interpreted in terms of type of TypedArray you're using. That way I don't have to interpret multiple hex failures into float when looking at errors.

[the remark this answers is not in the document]
kainino0x: I just mean I think this is now the same as:
```suggestion
    const exp = expected;
```

src/suites/cts/gpu_test.ts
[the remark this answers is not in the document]
austinEng: Yea.. ULPs or something would be nice. For now I have the tolerance as a number or a function which maps index=>number. For formats like rgb10a2unorm I couldn't have a single tolerance.

src/suites/cts/textures/zero_init.spec.ts
austinEng: Would be quite nice, though even for this test, that method could get quite complicated since there's so many different combinations that might not be supported.

src/suites/cts/textures/zero_init.spec.ts
austinEng: Whoops didn't do this yet.

austinEng:
Updated this @kainino0x; PTAL again! Probably sufficient to just review the first commit for now which is a bunch of utilities plus just sampling texture init.

src/webgpu/util/texture/layout.ts
```diff
@@ -0,0 +1,153 @@
...
+    assert(rowsPerImage >= size[1]);
+  }
+
+  const outputTexelData = new Uint8Array(outputBuffer);
+  for (let slice = 0; slice < size[2]; ++slice) {
+    for (let row = 0; row < size[1]; ++row) {
+      for (let col = 0; col < size[0]; ++col) {
+        const byteOffset =
+          slice * rowsPerImage * bytesPerRow + row * bytesPerRow + col * inputTexelData.byteLength;
+        outputTexelData.set(new Uint8Array(inputTexelData), byteOffset);
```
kainino0x, line 80: nit: preallocate the Uint8Array wrapper instead of setting it every time

src/webgpu/util/texture/layout.ts
```diff
@@ -0,0 +1,153 @@
...
+    minBytesPerRow,
+    bytesPerRow,
+    rowsPerImage,
+  };
+}
+
+export function fillTextureDataRows(
+  outputBuffer: ArrayBuffer,
+  size: GPUExtent3D,
+  inputTexelData: ArrayBuffer,
```
kainino0x, line 51: nit: `texelValue` to go with that

src/webgpu/util/texture/layout.ts
```diff
@@ -0,0 +1,153 @@
...
+  return {
+    bytesPerTexel,
+    byteLength: bytesPerRow * rowsPerImage,
+    minBytesPerRow,
+    bytesPerRow,
+    rowsPerImage,
+  };
+}
+
+export function fillTextureDataRows(
```
kainino0x, line 48: nit: didn't realize at first this was filling every texel with the same data (instead of every row), maybe `fillTextureDataWithTexelValue`

src/webgpu/util/texture/texelData.ts
```diff
@@ -0,0 +1,359 @@
...
+            },
+          }
+        );
+      }
+
+      default:
+        break;
+    }
+
+    const m = format.match(/([rgba]+)(\d+)(snorm|unorm|uint|sint|float)(-srgb)?/);
```
kainino0x, line 149: Perhaps we should put all this metadata in the capability_info instead of trying to compute it?

src/webgpu/util/texture/texelData.ts
```diff
@@ -0,0 +1,359 @@
...
+          }
+        } else {
+          // Packed representations are all 32-bit and use Uint as the data type.
+          switch (this.totalBitLength()) {
+            case 32: {
+              const view = new DataView(data);
+              const currentValue = view.getUint32(0, true);
+
+              let mask = 0xffffffff;
+              mask = (mask >> bitOffset) << bitOffset;
```
kainino0x, line 313: I think `>>` here will sign extend.

src/webgpu/util/texture/texelData.ts
```diff
@@ -0,0 +1,359 @@
...
+              new DataView(data, byteOffset, byteLength).setUint16(0, value, true);
+              break;
+            case 4:
+              new DataView(data, byteOffset, byteLength).setUint32(0, value, true);
+              break;
+            default:
+              unreachable();
+          }
+        } else {
+          // Packed representations are all 32-bit and use Uint as the data type.
```
kainino0x, line 306: I don't think the layout of depth24plus formats are observable at all. If this branch isn't hit, maybe we should unreachable() for now.

austinEng: This is hit for rg10b11float and rgb10a2unorm

src/webgpu/gpu_test.ts
```diff
@@ -157,13 +177,13 @@ export class GPUTest extends Fixture {
...
 
-  expectContents(src: GPUBuffer, expected: ArrayBufferView): void {
-    const exp = new Uint8Array(expected.buffer, expected.byteOffset, expected.byteLength);
+  expectContents(src: GPUBuffer, expected: TypedArrayBufferView): void {
     const dst = this.createCopyForMapRead(src, expected.buffer.byteLength);
 
     this.eventualAsyncExpectation(async niceStack => {
-      const actual = new Uint8Array(await dst.mapReadAsync());
-      const check = this.checkBuffer(actual, exp);
+      const constructor = expected.constructor as TypedArrayBufferViewConstructor;
```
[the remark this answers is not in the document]
austinEng: Get an error like:
Type 'Function' is missing the following properties from type 'Float64ArrayConstructor': BYTES_PER_ELEMENT, of, from

typeof .constructor is just "Function"

src/webgpu/capability_info.ts
```diff
@@ -6,60 +6,86 @@ export const kTextureFormatInfo: {
...
+  'rgba32uint':             { renderable:  true, color:  true, depth: false, stencil: false, storage:  true, bytes: 16 },
+  'rgba32sint':             { renderable:  true, color:  true, depth: false, stencil: false, storage:  true, bytes: 16 },
+  'rgba32float':            { renderable:  true, color:  true, depth: false, stencil: false, storage:  true, bytes: 16 },
   // Depth/stencil formats
-  'depth32float':           { renderable:  true, color: false },
-  'depth24plus':            { renderable:  true, color: false },
-  'depth24plus-stencil8':   { renderable:  true, color: false },
+  'depth32float':           { renderable:  true, color: false, depth:  true, stencil: false, storage: false, bytes:  4 },
+  'depth24plus':            { renderable:  true, color: false, depth:  true, stencil: false, storage: false, bytes:  4 },
+  'depth24plus-stencil8':   { renderable:  true, color: false, depth:  true, stencil:  true, storage: false, bytes:  4 },
```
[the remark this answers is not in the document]
austinEng: Now `bytesPerBlock?`

src/webgpu/capability_info.ts
```diff
@@ -6,60 +6,86 @@ export const kTextureFormatInfo: {
...
 };
 export const kTextureFormats = Object.keys(kTextureFormatInfo) as GPUTextureFormat[];
 
+export function isPackedFormat(format: GPUTextureFormat): boolean {
+  return format === 'depth24plus' || format === 'depth24plus-stencil8';
+}
```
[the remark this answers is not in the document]
austinEng: Added `copyable: boolean`

src/webgpu/capability_info.ts
```diff
@@ -6,60 +6,86 @@ export const kTextureFormatInfo: {
...
-  'depth32float':           { renderable:  true, color: false },
-  'depth24plus':            { renderable:  true, color: false },
-  'depth24plus-stencil8':   { renderable:  true, color: false },
+  'depth32float':           { renderable:  true, color: false, depth:  true, stencil: false, storage: false, bytes:  4 },
+  'depth24plus':            { renderable:  true, color: false, depth:  true, stencil: false, storage: false, bytes:  4 },
+  'depth24plus-stencil8':   { renderable:  true, color: false, depth:  true, stencil:  true, storage: false, bytes:  4 },
 };
 export const kTextureFormats = Object.keys(kTextureFormatInfo) as GPUTextureFormat[];
 
+export function isPackedFormat(format: GPUTextureFormat): boolean {
```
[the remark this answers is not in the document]
austinEng: Added `copyable: boolean` to the capability table

src/webgpu/capability_info.ts
```diff
@@ -6,60 +6,86 @@ export const kTextureFormatInfo: {
...
+}
+
+export const kTextureDimensions = (() => {
+  const kTextureDimensionsObject: { [k in GPUTextureDimension]: GPUTextureDimension } = {
+    [C.TextureDimension.E1d]: C.TextureDimension.E1d,
+    [C.TextureDimension.E2d]: C.TextureDimension.E2d,
+    [C.TextureDimension.E3d]: C.TextureDimension.E3d,
+  };
+  return Object.values(kTextureDimensionsObject);
+})();
```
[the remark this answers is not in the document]
austinEng: Done

src/webgpu/util/conversion.ts
```diff
@@ -0,0 +1,75 @@
+import { assert } from '../../common/framework/index.js';
+
+export function floatAsNormalizedInteger(float: number, bits: number, signed: boolean): number {
+  if (signed) {
+    assert(float >= -1 && float <= 1);
+    const min = -Math.pow(2, bits - 1);
+    const max = Math.pow(2, bits - 1) - 1;
```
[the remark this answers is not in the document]
austinEng: Whoops you're right; done.

src/webgpu/util/conversion.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  // >> to remove fraction, & to remove sign.
+  const exp = (bits >> 23) & 0xff;
+
+  // Convert to the new biased exponent.
+  const newExp = exp - (127 - bias);
+
+  // Mask only the fraction, and discard the lower bits.
+  const newFraction = (bits & 0x7fffff) >> fractionBitsToDiscard;
+  return (sign << (exponentBits + fractionBits)) | (newExp << fractionBits) | newFraction;
```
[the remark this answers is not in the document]
austinEng: Done

src/webgpu/capability_info.ts
```diff
@@ -6,60 +6,86 @@ export const kTextureFormatInfo: {
   [k in GPUTextureFormat]: {
     renderable: boolean;
     color: boolean;
+    depth: boolean;
+    stencil: boolean;
+    storage: boolean;
+    bytes: number;
```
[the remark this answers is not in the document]
austinEng: Done

src/webgpu/util/texture/layout.ts
```diff
@@ -0,0 +1,153 @@
...
+  minBytesPerRow: number;
+  bytesPerRow: number;
+  rowsPerImage: number;
+} {
+  if (!Array.isArray(size)) {
+    size = [size.width, size.height, size.depth];
+  }
+
+  const bytesPerTexel = kTextureFormatInfo[format].bytes;
+  const minBytesPerRow = (size[0] >> mipLevel) * bytesPerTexel;
```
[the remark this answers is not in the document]
austinEng: Done

src/webgpu/util/texture/layout.ts
```diff
@@ -0,0 +1,153 @@
...
+export const kBytesPerRowAlignment = 256;
+
+export function getTextureCopyLayout(
+  format: GPUTextureFormat,
+  size: GPUExtent3D,
+  mipLevel: number = 0,
+  bytesPerRow?: number,
+  rowsPerImage?: number
+): {
+  bytesPerTexel: number;
```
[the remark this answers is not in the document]
austinEng: Done

src/webgpu/util/texture/layout.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  if (rowsPerImage !== undefined) {
+    assert(rowsPerImage >= size[1]);
+  } else {
+    rowsPerImage = size[1];
+  }
+
+  return {
+    bytesPerTexel,
+    byteLength: bytesPerRow * rowsPerImage,
```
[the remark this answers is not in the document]
austinEng: Done (I think)

src/webgpu/util/texture/layout.ts
```diff
@@ -0,0 +1,153 @@
+import { C, assert } from '../../../common/framework/index.js';
+import { kTextureFormatInfo } from '../../capability_info.js';
+
+export const kBytesPerRowAlignment = 256;
+
+export function getTextureCopyLayout(
+  format: GPUTextureFormat,
+  size: GPUExtent3D,
```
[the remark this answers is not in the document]
austinEng: Done

src/webgpu/util/texture/layout.ts
austinEng: Done

src/webgpu/util/texture/layout.ts
austinEng: Done

src/webgpu/util/texture/layout.ts
austinEng: Done

src/webgpu/util/texture/texelData.ts
austinEng: Maybe? But it would all be duplicate information, and that table is getting huge. Seems odd to have:

`'r32uint': { components: ['r'], bits: 32, type: 'uint' }`
^ This representation also doesn't work for all the formats (like rg11b10float) so it would probably be more complex or have even more fields

src/webgpu/util/texture/texelData.ts
austinEng: Done

austinEng:
@kainino0x separated the shader tests out; I'll make separate PRs to `glsl-dependent` after this part is in master.

src/webgpu/util/texture/texelData.ts
kainino0x: Right.

src/webgpu/util/texture/texelData.ts
kainino0x: What about about a new table that's just a hardcoded version of kRepresentationCache? Seems less complex than this logic.

src/webgpu/capability_info.ts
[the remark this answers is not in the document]
kainino0x: I'll think about niceifying it some other time. I also realized the types exported by this file are insufficiently const-y so that's going to take some fixing.

src/webgpu/util/texture/layout.ts
```diff
@@ -0,0 +1,194 @@
...
+      break;
+    case '3d':
+      mipSize = [size[0] >> mipLevel, size[1] >> mipLevel, size[2] >> mipLevel];
+      break;
+    default:
+      unreachable();
+  }
+
+  const { blockWidth, blockHeight, bytesPerBlock } = kTextureFormatInfo[format];
+  assert(!!bytesPerBlock && !!blockWidth && !!blockHeight);
```
kainino0x, line 39: nit: `> 0`
although I don't think this assert is really necessary.

src/webgpu/util/texture/layout.ts
```diff
@@ -0,0 +1,194 @@
...
+    bytesPerRow = alignedMinBytesPerRow;
+  }
+
+  if (rowsPerImage !== undefined) {
+    assert(rowsPerImage >= size[1]);
+  } else {
+    rowsPerImage = size[1];
+  }
+
+  const bytesPerSlice = (bytesPerRow * rowsPerImage) / blockWidth;
```
kainino0x, line 56: nit: `bytesPerRow * (rowsPerImage / blockHeight)`?
nit: assert rowsPerImage is aligned to blockHeight

src/webgpu/util/texture/layout.ts
```diff
@@ -0,0 +1,194 @@
...
+  const sliceSize =
+    bytesPerRow * (size[1] / blockHeight - 1) + bytesPerBlock * (size[0] / blockWidth);
+  const byteLength = bytesPerSlice * (size[2] - 1) + sliceSize;
```
kainino0x, line 59: should these `size`s be `align(mipSize, block size)`?

src/webgpu/util/texture/layout.ts
```diff
@@ -0,0 +1,194 @@
...
+export function fillTextureDataWithTexelValue(
+  format: GPUTextureFormat,
+  outputBuffer: ArrayBuffer,
+  size: [number, number, number],
+  texelValue: ArrayBuffer,
+  bytesPerRow?: number,
+  rowsPerImage?: number
+): void {
+  const { blockWidth, blockHeight, bytesPerBlock } = kTextureFormatInfo[format];
+  assert(!!bytesPerBlock && !!blockWidth && !!blockHeight);
```
kainino0x, line 79: same

src/webgpu/util/texture/layout.ts
```diff
@@ -0,0 +1,194 @@
...
+  assert(bytesPerBlock === texelValue.byteLength);
+
+  const minBytesPerRow = (size[0] / blockWidth) * bytesPerBlock;
```
kainino0x, line 82: nit: assert size[0] is aligned to blockWidth

src/webgpu/util/texture/layout.ts
```diff
@@ -0,0 +1,194 @@
...
+      mipSize = [size[0] >> mipLevel, size[1] >> mipLevel, size[2] >> mipLevel];
+      break;
+    default:
+      unreachable();
+  }
+
+  const { blockWidth, blockHeight, bytesPerBlock } = kTextureFormatInfo[format];
+  assert(!!bytesPerBlock && !!blockWidth && !!blockHeight);
+
+  const minBytesPerRow = (mipSize[0] / blockWidth) * bytesPerBlock;
```
kainino0x, line 41: nit: assert mipSize[0] is aligned to blockWidth (or maybe add a new helper which does both the assert and the divide together)

src/webgpu/util/texture/layout.ts
```diff
@@ -0,0 +1,194 @@
...
+  if (rowsPerImage !== undefined) {
+    assert(rowsPerImage >= size[1]);
+  } else {
+    rowsPerImage = size[1];
+  }
+
+  const bytesPerSlice = (bytesPerRow * rowsPerImage) / blockWidth;
+  const sliceSize =
+    bytesPerRow * (size[1] / blockHeight - 1) + bytesPerBlock * (size[0] / blockWidth);
+  const outputByteLength = bytesPerSlice * (size[2] - 1) + sliceSize;
```
kainino0x, line 100: possible to deduplicate?

src/webgpu/util/texture/layout.ts
```diff
@@ -0,0 +1,194 @@
...
+  buffer.unmap();
+
+  return {
+    buffer,
+    bytesPerRow,
+    rowsPerImage,
+  };
+}
+
+export function createTextureReadbackBuffer(
```
kainino0x, line 155: what's the cpuData? Method name doesn't make it clear. Is it the expected data?

src/webgpu/api/operation/resource_init/texture_zero_init_test.ts
```diff
@@ -0,0 +1,674 @@
...
+                    readMethod === ReadMethod.CopyToBuffer ||
+                    readMethod === ReadMethod.CopyToTexture
+                  ) {
+                    // It doesn't make sense to copy from a packed depth format.
+                    // This is not specified yet, but it will probably be disallowed as the bits may
+                    // be vendor-specific.
+                    if (format === 'depth24plus') {
+                      continue;
+                    }
+                    if (format === 'depth24plus-stencil8' && aspect === 'depth-only') {
```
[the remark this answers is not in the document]
austinEng: Hm, I think I'll remove this for now at least until it's investigated further

src/webgpu/util/texture/layout.ts
austinEng: They're `number | undefined` so it's both a type assertion and >0 check.

austinEng:
fyi: in Chrome on Mac, some of the copy tests fail, but I think it's all Dawn issues related to depth/stencil textures
 - store op clear doesn't work with subresources. Dawn has some TODOs about that
 - "The entire subresource must be copied when using a depth/stencil texture or when samples are greater than 1"
    - this is a [D3D12 restriction](https://docs.microsoft.com/en-us/windows/win32/api/d3d12/nf-d3d12-id3d12graphicscommandlist-copytextureregion), but these tests are indeed copying the whole subresource. Dawn's validation is looking at the whole texture size, not the mip level
 - 'Texture aspect must be 'all'' - Dawn simply doesn't support anything that's not all yet

src/webgpu/util/texture/layout.ts
austinEng: Hm.. I may be missing some rounding as well. I looked for what we've decided on block-compressed copies and found https://github.com/gpuweb/gpuweb/issues/363 which doesn't look resolved.

like TBD on whether you have a small mip like 15x15, you actually have to round up and copy a 16x16 block

src/webgpu/util/texture/texelData.ts
austinEng: Separate table is a good idea; did that in the latest update

src/webgpu/util/texture/layout.ts
austinEng: I forget why I had this function, but I never use it, even in later patches, so it's removed

src/webgpu/util/texture/layout.ts
austinEng: done

src/webgpu/util/texture/layout.ts
austinEng: done, though may change based on compressed texture copy rules

src/webgpu/util/texture/layout.ts
austinEng: done, though may change based on compressed texture copy rules

src/webgpu/util/texture/layout.ts
austinEng: done, though may change based on compressed texture copy rules

src/webgpu/api/operation/resource_init/texture_zero_init_test.ts
```diff
@@ -0,0 +1,674 @@
...
+                    if (
+                      readMethod === ReadMethod.CopyToBuffer ||
+                      readMethod === ReadMethod.CopyToTexture
+                    ) {
+                      switch (format) {
+                        case 'depth32float':
+                          break;
+                        case 'depth24plus':
+                        case 'depth24plus-stencil8':
+                          // Copies with packed depth/stencil formats currently not yet specified.
```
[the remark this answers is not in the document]
austinEng: done

src/webgpu/api/operation/resource_init/texture_zero_init_test.ts
```diff
@@ -0,0 +1,674 @@
...
+                    // be vendor-specific.
+                    if (format === 'depth24plus') {
+                      continue;
+                    }
+                    if (format === 'depth24plus-stencil8' && aspect === 'depth-only') {
+                      continue;
+                    }
+
+                    // Copies to multisampled textures currently not yet specified
+                    if (sampleCount > 1) {
```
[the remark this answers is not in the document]
austinEng: done

src/webgpu/api/operation/resource_init/texture_zero_init_test.ts
```diff
@@ -0,0 +1,674 @@
...
+      // Copies to multisampled textures not yet specified.
+      // Use a storeOp for now.
+      assert(kTextureFormatInfo[this.params.format].renderable);
+      this.initializeWithStoreOp(state, texture, subresourceRange);
+    } else {
+      this.initializeWithCopy(texture, state, subresourceRange);
+    }
+
+    // Make sure it actually initialized the subresources.
+    this.checkContents(texture, state, subresourceRange);
```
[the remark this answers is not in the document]
austinEng: done

austinEng:
Okay, @kainino0x I think I responded to all feedback. Github is kinda hard for big reviews

kainino0x:
FYI: this and all the other recent changes are now merged to `glsl-dependent`.

PR #141, opened 2020-03-31
[glsl-dependent] Allow overriding of the path to glslang

kainino0x:
This allows browsers to specify a path to a file outside WPT (which can
e.g. be downloaded as a dependency instead of checked into Git). This
will allow testing of runtime-generated shaders through WPT.

### DIFF
--- a/src/common/glslang.ts
+++ b/src/common/glslang.ts
@@ -1,0 +1,44 @@
+import { SkipTestCase } from './framework/fixture.js';
+import { assert } from './framework/util/util.js';
+import { getGlslangPath } from './glslang_path.js';
+
+type glslang = typeof import('@webgpu/glslang/dist/web-devel/glslang');
+type Glslang = import('@webgpu/glslang/dist/web-devel/glslang').Glslang;
+type ShaderStage = import('@webgpu/glslang/dist/web-devel/glslang').ShaderStage;
+type SpirvVersion = import('@webgpu/glslang/dist/web-devel/glslang').SpirvVersion;
+
+let glslangAttempted: boolean = false;
+let glslangInstance: Glslang | undefined;
+
+export async function initGLSL(): Promise<void> {
+  if (glslangAttempted) {
+    if (!glslangInstance) {
+      throw new SkipTestCase('glslang is not available');
+    }
+  } else {
+    glslangAttempted = true;
+    const glslangPath = getGlslangPath() || '../glslang.js';
+    let glslangModule: () => Promise<Glslang>;
+    try {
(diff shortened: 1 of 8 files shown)


src/framework/glslang_path.ts
```diff
@@ -0,0 +1,12 @@
+import { assert } from './util/index.js';
+
+let glslangPath: string | undefined;
+
+export function getGlslangPath(): string | undefined {
+  return glslangPath;
+}
+
+export function setGlslangPath(path: string): void {
+  assert(path.startsWith('/'), 'glslang path must be absolute');
```
austinEng: Will this work on Windows?

src/framework/glslang.ts
```diff
@@ -0,0 +1,44 @@
...
+    const glslangPath = getGlslangPath() || '../glslang.js';
+    let glslangModule: () => Promise<Glslang>;
+    try {
+      glslangModule = ((await import(glslangPath)) as glslang).default;
+    } catch (ex) {
+      throw new SkipTestCase('glslang is not available');
+    }
+
+    const glslang = await glslangModule();
+    glslangInstance = glslang;
```
austinEng: Could just assign directly to `glslangInstance`

templates/cts.html
```diff
@@ -45,4 +45,10 @@
 </style>
 
 <textarea id=results></textarea>
-<script type=module src=/webgpu/runtime/wpt.js></script>
+<script type=module>
+  import { setGlslangPath } from '/webgpu/framework/glslang_path.js';
```
austinEng: Huh my comment disappeared.

Unfamiliar with <script type="module">. Does setGlslangPath become global? Or is this an example and a browser using it would either replace this block entirely, or add another one?

kainino0x: Imports are like `const`, iirc they don't escape the <script>.

The browser's fork of cts.html would insert new code inside this block.

austinEng: Thanks for the clarification, makes sense.

src/framework/glslang_path.ts
kainino0x: Yes, on web it's an absolute http request path, and on node it's never needed because glslang is in a known location.

templates/cts.html
kainino0x: https://chromium-review.googlesource.com/c/chromium/src/+/2100018/1/third_party/blink/web_tests/wpt_internal/webgpu/cts.html

kainino0x:
Will rebase and merge after the glsl-dependent branch is up to date with the changes I want to make before the branch point

PR #142, opened 2020-03-31
Adjust the directory structure outside the tests for navigability

kainino0x:
- Moves `src/{framework,runtime,tools,constants.ts}` to `src/common/*`
- Moves `templates` to `src/common/templates`
- Moves `src/suites/*` up to `src/*`
- Renames the `cts:` suite to `webgpu:`
- Unhides some files from VS Code
- Updates outdated `package.json`

Restructure of the webgpu suite to match the CTS guidelines will come next.


kainino0x:
@austinEng PTAL. This should just be renames, so feel free to skim; if there are any critical places where I missed a rename in the code, it's not really possible to see in code review anyway, and we'll discover it soon. (But I did test cmdline, standalone, and wpt, so I think it's mostly correct.)

PR #149, opened 2020-04-08
Remove all GLSL dependence from the master branch

kainino0x:
All of this now lives on the glsl-dependent branch. It will migrate back as we convert these tests to WGSL.


package.json
```diff
@@ -26,7 +26,6 @@
     "@types/jquery": "^3.3.30",
     "@types/node": "^13.1.2",
     "@types/offscreencanvas": "^2019.6.1",
-    "@webgpu/glslang": "^0.0.15",
```
austinEng, line 29: (huh I couldn't comment on the lower line)
nit: I don't think we need `babel-plugin-macros` anymore. Same comment in #150 as well

PR #150, opened 2020-04-08
[glsl-dependent] Allow dual-language tests, make glslang required, remove GLSL() macro

kainino0x:


kainino0x:
whoops, I misread and thought we were still using it for the header comments.

PR #154, opened 2020-04-09
Add files missing from out-wpt/

kainino0x:

PR #155, opened 2020-04-09
Fix GLSL tests

kainino0x:
Somehow these changes, which are necessary to actually run the GLSL tests, got lost after the point when I tested them.

Also, there was a conflict where the actual `glslang.js` was getting overwritten with the built version of src/common/glslang.ts.

Tested in standalone and WPT

Reverts #154


kainino0x:
TBR: @austinEng 

landing this so I can make sure the complete roll process works

PR #158, opened 2020-04-10
[glsl-dependent] Remove setGlslangPath and just put glslang.js in a known location

kainino0x:
This fixes tests on workers, for which the overridden glslang path was not taking effect.

Reverts #141


kainino0x:
Now that we have two branches, the version of the CTS in WPT will be master. We'll have a full copy of the glsl-dependent branch in wpt_internal/, where we can put extra files without any auto-exporting.

kainino0x:
Travis build passed but didn't report back here somehow:
https://travis-ci.org/github/gpuweb/cts/builds/673240172

Merging TBR so I can continue testing this in Chromium.

austinEng:
Ah okay, sounds good then. Didn't realize we'd have both there.

PR #159, opened 2020-04-14
Add lint and compile watcher. Move from TSLint to ESLint.

austinEng:
This commit adds a "grunt dev" task which signficiantly improves
productivity by watching the src/ directory and recompiling and
relinting files on save.

It also moves the linting from TSLint to ESLint. TSLint is being
deprecated, and ESLint is the recommended path forward. It also
has nicer integration with prettier for code formatting. gts, the
Typescript styles we extend, is also in the process of migrating,
so this commit updates to using gts@next which includes the ESLint
config.


.eslintrc.json
```diff
@@ -0,0 +1,42 @@
...
+        "memberVariableDeclaration": false
+      }
+    ],
+    "@typescript-eslint/no-inferrable-types": "off",
+
+    "sort-imports": "warn",
+    "no-useless-escape": "warn",
+    "no-empty": "warn",
+    "@typescript-eslint/no-unused-vars": "warn",
+    "no-inner-declarations": "warn"
```
austinEng, line 31: @kainino0x what do you think? These several rules I have as "warn" right now because they're a bit different from what we had in TSLint.

kainino0x: I'd like to keep master warning-free if we can, but I'm okay with doing lint autofixes across the codebase if they're not too huge

kainino0x: sort-imports and no-useless-escape should definitely be okay to autofix

Gruntfile.js
```diff
@@ -43,16 +43,26 @@ module.exports = function (grunt) {
           'src/',
         ],
       },
-      'gts-check': {
+      lint: {
         cmd: 'node',
-        args: ['node_modules/gts/build/src/cli', 'check'],
+        args: ['node_modules/eslint/bin/eslint', 'src/**/*.ts'],
```
kainino0x, line 48: Are all of the gts checks now encompassed by eslint? (not sure what the others were, maybe just prettier)

.eslintrc.json
austinEng: I don't think any are actually autofixable, but here's what would change:

`no-useless-escape`: `\/` --> `/`

`sort-imports`: this one means sorting the import names, not the import filepaths which is what TSLint did. However, multi-imports take precedence. so:
```js
import { X, b, c } from './foo' // multi first
import { A } from './a' // then capital
import { d } from './d' // then lowercase
```
Most of the `@typescript-eslint/no-unused-vars` errors are from `g.test('foo', t => ...)` where `t` isn't used in the body. Most of these are unittests or examples, so I think we can /* eslint-disable-next-line */
 
`no-empty`: This is happening on our empty catch statements. We could use `allowEmptyCatch`, but there's only a couple right now that are very intentional. I think we should suppress this one at the site.

`no-inner-declarations`: functions need to be declared at the root scope.  The couple places we have to do this could be arrow functions instead which sounds good to me.

Gruntfile.js
austinEng: Yea, eslint runs the gts ones and it extends

"eslint:recommended",
    "plugin:node/recommended",
    "prettier"

tsconfig.json
```diff
@@ -3,22 +3,18 @@
   "compilerOptions": {
     "lib": ["dom", "es2016"],
     "module": "esnext",
-
     /* Output options */
     "noEmit": true,
-
+    "allowUmdGlobalAccess": true,
```
austinEng, line 8: Whoops, don't think this was needed. (was part of me trying to get global typescripts to be seen by eslint)

src/common/framework/params_utils.ts
```diff
@@ -37,7 +37,7 @@ export function paramsEquals(x: ParamSpec | null, y: ParamSpec | null): boolean
   }
 
   for (const xk of Object.keys(x)) {
-    if (x[xk] !== undefined && !y.hasOwnProperty(xk)) {
+    if (x[xk] !== undefined && !(xk in y)) {
```
austinEng, line 40: because [`no-prototype-builtins`](https://eslint.org/docs/rules/no-prototype-builtins)

package.json
```diff
@@ -5,6 +5,9 @@
   "scripts": {
     "test": "grunt pre"
   },
+  "engines": {
+    "node": ">=11.14.0"
```
austinEng, line 9: because some of the Node APIs we use require at least this version

.eslintrc.json
```diff
@@ -0,0 +1,42 @@
...
+        "message": "WPT (Web Platform Tests) doesn't allow setTimeout."
+      }
+    ],
+    "linebreak-style": ["error", "unix"],
+    "no-console": "error",
+    "@typescript-eslint/typedef": [
+      "error",
+      {
+        "arrowParameter": false,
+        "memberVariableDeclaration": false
```
austinEng, line 22: This is to match what the current code does, but it overrides some of the ESLint recommended rules. They recommend being more strict with types everywhere, even things like:
class A {
  foo = 0; // needs type annotation
}

.eslintrc.json
kainino0x: Those all sound good to me. I'm surprised they don't have autofixers, especially sort-imports. It's going to be a pain if we have to manually fix our import sorting on every patch.

src/common/framework/collect_garbage.ts
```diff
@@ -34,7 +34,7 @@ export function attemptGarbageCollection(): void {
   let i: number;
   function gcRec(n: number): void {
     if (n < 1) return;
-    // tslint:disable-next-line: no-any
+    /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
     let temp: any = { i: 'ab' + i + i / 100000 };
```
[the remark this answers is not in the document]
kainino0x: I don't know, I think it's just supposed to help produce garbage. I copied this verbatim from webgl-test-utils.

src/common/tools/.eslintrc.json
```diff
@@ -0,0 +1,8 @@
+{
+  "rules": {
+    "no-console": "off",
+    "no-process-exit": "off",
+    "node/no-unpublished-require": "off",
```
kainino0x, line 5: Maybe someday we should actually move these (actually, most) devDependencies into dependencies. It seems silly but there's a real possibility a project might add the cts as a dependency (e.g. webgpu for node). We would need to test it though because I have no idea what will happen if you actually try to do this.

tsconfig.json
```diff
@@ -3,22 +3,18 @@
...
-
     /* Module Options */
     "moduleResolution": "node",
-    "esModuleInterop": false,
+    "esModuleInterop": false
   },
-  "include": [
-    "src/**/*.ts",
-  ],
+  "include": ["src/**/*.ts", "node_modules/@webgpu/types"]
```
kainino0x, line 19: Is this new include necessary? Would expect it to have no effect.

austinEng: not needed.

.eslintrc.json
austinEng: Okay, will do in a follow-up

kainino0x:
Nice!

PR #161, opened 2020-04-17
Use ParamsBuilder for texture_zero_init_tests

austinEng:


src/common/framework/params.ts
```diff
@@ -33,7 +33,7 @@ export function params(): ParamsBuilder<{}> {
   return new ParamsBuilder();
 }
 
-class ParamsBuilder<A extends {}> implements ParamSpecIterable {
+export class ParamsBuilder<A extends {}> implements ParamSpecIterable {
```
austinEng, line 36: Compiles and lints fine without this, but VSCode complains?

> Return type of public static method from exported class has or is using name 'ParamsBuilder' from external module "src/common/framework/params" but cannot be named.ts (4050)

kainino0x: I didn't see this in linting. Getting vscode to sync exactly with the linting rules is tricky. If it doesn't affect linting I'd slightly prefer to not export this (but it doesn't really matter).

austinEng: Okay, removed the export.

kainino0x: I have two extensions: Prettier and ESLint (removed tslint). Together they seem to work (Prettier is needed otherwise ESLint screws up indentation) and I don't see that warning

src/webgpu/api/operation/resource_init/texture_zero_init_test.ts
```diff
@@ -458,139 +459,72 @@ export abstract class TextureZeroInitTest extends GPUTest {
...
+          ({ readMethod, sampleCount }) =>
+            // We can only read from multisampled textures by sampling.
+            sampleCount > 1 &&
+            (readMethod === ReadMethod.CopyToBuffer || readMethod === ReadMethod.CopyToTexture)
+        )
+        .expand(function* () {
+          for (const { dimension, sliceCount } of kCreationSizes) {
+            yield { dimension, sliceCount };
+          }
+        })
```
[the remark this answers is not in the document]
austinEng: Done

PR #164, opened 2020-04-20
Add present_to_canvas reftest

shaoboyan:
This test case clear whole canvas as green and using wpt framework to do
pixel compare to ensure webgpu swapchain works well


src/webgpu/web-platform/reftests/present_to_canvas.ts
```diff
@@ -0,0 +1,45 @@
...
+        storeOp: 'store',
+      },
+    ],
+  });
+  pass.endPass();
+  t.device.defaultQueue.submit([encoder.finish()]);
+  await t.finalize();
+}
+PresentToCanvasTest().then(() => {
+  takeScreenshotDelayed(50);
```
shaoboyan, line 44: This is a function in wpt utils, I don't know how to handle it

shaoboyan:
@kainino0x  I find that typescript doesn't support multiple inherit and default implementation of interface. So I cannot abstract all codes from GPUTest.  And in the test, it relies on a wpt utils, I don't know how to handle it either. Do you have suggestions on this?

src/webgpu/web-platform/reftests/present_to_canvas.ts
kainino0x: You can just make typescript happy by doing something like:

```typescript
declare function takeScreenshotDelayed(ms: number): void;
```
at the top.
https://github.com/gpuweb/cts/blob/master/src/common/framework/util/timeout.ts#L1

src/webgpu/web-platform/reftests/gpu_ref_test.ts
```diff
@@ -0,0 +1,49 @@
+import { DevicePool } from '../../../common/framework/gpu/implementation.js';
+import { assert, unreachable } from '../../../common/framework/util/util.js';
+
+const devicePool = new DevicePool();
```
kainino0x, line 4: We don't need to use the DevicePool for this test. It will always run independently of other tests (not on the same page load), and so won't need to share Devices with anything else. You can just create a device the usual way.

src/webgpu/web-platform/reftests/present_to_canvas.ts
```diff
@@ -0,0 +1,45 @@
+import { GPURefTest } from './gpu_ref_test.js';
+
+async function PresentToCanvasTest(): Promise<void> {
+  const t = new GPURefTest();
+  await t.init();
+  if (typeof document === 'undefined') {
+    // Skip if there is no document (Workers, Node)
+    console.log('DOM is not available to create canvas element');
+    return;
+  }
```
kainino0x, line 10: Not needed because this file is only ever included from the html file (so the DOM is always available)

src/webgpu/web-platform/reftests/gpu_ref_test.ts
```diff
@@ -0,0 +1,49 @@
...
+    const queue = device.defaultQueue;
+    this.objects = { device, queue };
+
+    try {
+      await device.popErrorScope();
+      unreachable('There was an error scope on the stack at the beginning of the test');
+    } catch (ex) { }
+
+    device.pushErrorScope('out-of-memory');
+    device.pushErrorScope('validation');
```
kainino0x, line 32: I don't think it's even very important to check for errors during this test. We'll have plenty of other tests where we can check errors (including tests which do about the same thing as this, but without the reftesting).

src/webgpu/web-platform/reftests/present_to_canvas.ts
kainino0x: Also, wouldn't we want to do this for all ref tests? Maybe find a way to include this in `gpu_ref_test.ts`

src/webgpu/web-platform/reftests/gpu_ref_test.ts
kainino0x: Actually don't even need to use getGPU as we can assume these are running in the browser (and not Node or something).

kainino0x:
oops, didn't mean to push to your branch (rolled it back so there is no new change).

I opened a PR against your branch with the changes I had in mind: https://github.com/shaoboyan/cts/pull/1
but feel free to change as needed

shaoboyan:
@kainino0x Thanks for your help on simplify and make how to do reftest clear. I use your code to generate js and verified in wpt. With some updating in html file, they works pretty fine. So I update the patch based on your change. PTAL again if you have time!

Kangz:
It would be nice to have different colors for the 4 corners of the canvas, so that we know the orientation is correct. Unfortunately this would likely require shaders so maybe we can make it a TODO and revisit once more of WGSL works?

kainino0x:
How about a second test that uses copyBufferToTexture to put, say, 2x2 pixels into the canvas? (Can be a followup)

shaoboyan:
@kainino0x @Kangz Yes. a rect with four different colors is a good test base(WebGL use it to test orientation I think) And Yes, I think we need a plan to describe all the tests we want(I remember Kai sent out a link about a template to desribe test and their current states on slack but I fail to find it :( )
So how about landing this patch after we collect and write down these TODOs as the first step for the reftest(and present-to-canvas) ?

kainino0x:
I set up the file copy in the last commit (PTAL, if you'd like). I'd like to merge but something is wrong with Travis...

kainino0x:
FYI I started playing with copyBufferToTexture in a reftest (was easy to just modify your code). It's only half done right now (and really broken), but I'm happy to finish it up if you'd like.

shaoboyan:
@kainino0x  Sure. And happy you have it :)

PR #165, opened 2020-04-20
Test uninitialized depth/stencil attachments are zero

austinEng:
- Multi-subresource tests fail on Chrome Canary because Dawn doesn't correctly bind depth/stencil texture views yet.
- Needs #163 to update types first

### DIFF
--- a/src/webgpu/api/operation/resource_init/depth_stencil_attachment_clear.spec.ts
+++ b/src/webgpu/api/operation/resource_init/depth_stencil_attachment_clear.spec.ts
@@ -20,18 +20,16 @@ import {
 } from './texture_zero_init_test.js';
 
 class DepthStencilAttachmentClearTest extends TextureZeroInitTest {
-  private depthTestPipelineCache: Map<string, GPURenderPipeline> = new Map();
+  // Construct a pipeline which will render a single triangle with depth
+  // equal to |initializeStateAsDepth(state)|. The depth compare function
+  // is set to "equal" so the fragment shader will only write 1.0 to the
+  // R8Unorm output if the depth buffer contains exactly the expected value.
   private getDepthTestReadbackPipeline(
     state: InitializedState,
     format: GPUTextureFormat,
     sampleCount: number
   ): GPURenderPipeline {
-    const key = [state, format, sampleCount].join('_');
-    if (this.depthTestPipelineCache.has(key)) {
-      return this.depthTestPipelineCache.get(key)!;
-    }
-
-    const pipeline = this.device.createRenderPipeline({
+    return this.device.createRenderPipeline({
       vertexStage: {
         entryPoint: 'main',
         module: this.makeShaderModule('vertex', {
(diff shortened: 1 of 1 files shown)


src/webgpu/api/operation/resource_init/depth_stencil_attachment_clear.spec.ts
```diff
@@ -0,0 +1,242 @@
...
+import {
+  InitializedState,
+  ReadMethod,
+  TextureZeroInitTest,
+  initializedStateAsDepth,
+  initializedStateAsStencil,
+} from './texture_zero_init_test.js';
+
+class DepthStencilAttachmentClearTest extends TextureZeroInitTest {
+  private depthTestPipelineCache: Map<string, GPURenderPipeline> = new Map();
```
kainino0x, line 23: I didn't think about this before, but does this cache do anything? The whole fixture class gets constructed once for each test case, and anyway we can't rely on the device being the same between test cases.

We could cache the generated SPIR-V across cases. But since we're already not doing it, it doesn't seem extremely necessary.

src/webgpu/api/operation/resource_init/depth_stencil_attachment_clear.spec.ts
```diff
@@ -0,0 +1,242 @@
...
+        }),
+      },
+      colorStates: [
+        {
+          format: 'r8unorm',
+        },
+      ],
+      depthStencilState: {
+        format,
+        depthCompare: 'equal',
```
kainino0x, line 68: nit: comment to briefly explain how the test works, here and in main() above; same for stencil

src/webgpu/api/operation/resource_init/depth_stencil_attachment_clear.spec.ts
```diff
@@ -0,0 +1,242 @@
...
+
+      this.queue.submit([commandEncoder.finish()]);
+
+      const arrayBuffer = new ArrayBuffer(byteLength);
+      fillTextureDataWithTexelValue(expectedTexelData, 'r8unorm', '2d', arrayBuffer, [
+        width,
+        height,
+        1,
+      ]);
+      this.expectContents(buffer, new Uint8Array(arrayBuffer));
```
kainino0x, line 233: A lot of this looks like maybe it could be a general expectSingleColorTextureContents helper. Could you consider whether that could be abstracted in a followup if it would be used elsewhere?

src/webgpu/api/operation/resource_init/depth_stencil_attachment_clear.spec.ts
```diff
@@ -0,0 +1,242 @@
...
+          pass.setPipeline(
+            this.getDepthTestReadbackPipeline(state, this.params.format, this.params.sampleCount)
+          );
+          break;
+
+        case ReadMethod.StencilTest:
+          pass.setPipeline(
+            this.getStencilTestReadbackPipeline(this.params.format, this.params.sampleCount)
+          );
+          pass.setStencilReference(initializedStateAsStencil(state));
```
kainino0x, line 194: add a comment here too, since this is critical to the test

austinEng: Done

src/webgpu/api/operation/resource_init/depth_stencil_attachment_clear.spec.ts
austinEng: Done

src/webgpu/api/operation/resource_init/depth_stencil_attachment_clear.spec.ts
austinEng: We can probably do that. Probably on master first..

src/webgpu/api/operation/resource_init/depth_stencil_attachment_clear.spec.ts
austinEng: Ah welp. Removed the caching then. I'll remove it from the other test in another PR.

PR #166, opened 2020-04-21
Fix storage-texture and comparison-sampler tests

kainino0x:
Handles:
- storageTextureFormat required for storage texture bindings in BGL
- comparison sampler objects not tested
- default sampler is not a valid comparison sampler

PR #167, opened 2020-04-21
Clean up capability info

kainino0x:
This is a first attempt at cleaning up the capability info for correctness and precision.

Ignore the first commit, which is #166.

### DIFF
--- a/src/webgpu/api/validation/createBindGroupLayout.spec.ts
+++ b/src/webgpu/api/validation/createBindGroupLayout.spec.ts
@@ -160,8 +160,7 @@ g.test('max resources per stage/in bind group layout')
         binding: i,
         visibility: maxedVisibility,
         type: maxedType,
-        storageTextureFormat:
-          maxedTypeInfo.resource === 'storage-textureview' ? 'rgba8unorm' : undefined,
+        storageTextureFormat: maxedTypeInfo.resource === 'storageTex' ? 'rgba8unorm' : undefined,
       });
     }
 
@@ -175,8 +174,7 @@ g.test('max resources per stage/in bind group layout')
       binding: maxedCount,
       visibility: extraVisibility,
       type: extraType,
-      storageTextureFormat:
-        extraTypeInfo.resource === 'storage-textureview' ? 'rgba8unorm' : undefined,
+      storageTextureFormat: extraTypeInfo.resource === 'storageTex' ? 'rgba8unorm' : undefined,
     });
 
     const shouldError = maxedCount >= kMaxBindingsPerBindGroup;
@@ -203,8 +201,7 @@ g.test('max resources per stage/in pipeline layout')
         binding: i,
         visibility: maxedVisibility,
         type: maxedType,
-        storageTextureFormat:
-          maxedTypeInfo.resource === 'storage-textureview' ? 'rgba8unorm' : undefined,
+        storageTextureFormat: maxedTypeInfo.resource === 'storageTex' ? 'rgba8unorm' : undefined,
       });
     }
 
@@ -219,8 +216,7 @@ g.test('max resources per stage/in pipeline layout')
           binding: 0,
           visibility: extraVisibility,
           type: extraType,
-          storageTextureFormat:
-            extraTypeInfo.resource === 'storage-textureview' ? 'rgba8unorm' : undefined,
+          storageTextureFormat: extraTypeInfo.resource === 'storageTex' ? 'rgba8unorm' : undefined,
         },
       ],
     });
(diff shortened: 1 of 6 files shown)


kainino0x:
@austinEng, PTAL, but I'm looking for suggestions if you have any (it's kind of ugly right now).

src/webgpu/capability_info.ts
```diff
@@ -100,41 +119,146 @@ export type PerStageBindingLimitType =
...
+} = /* prettier-ignore */ {
+  'sampled-texture':           { usage: C.TextureUsage.Sampled, ...kBindingKind.sampledTex,  ...kValidStagesAll,     },
+  'writeonly-storage-texture': { usage: C.TextureUsage.Storage, ...kBindingKind.storageTex,  ...kValidStagesCompute, },
+  'readonly-storage-texture':  { usage: C.TextureUsage.Storage, ...kBindingKind.storageTex,  ...kValidStagesAll,     },
+};
+export const kTextureBindingTypes = keysOf(kTextureBindingTypeInfo);
+
+// All binding types (merged from above)
+
+interface BindingTypeInfo {
```
austinEng, line 243: nit: Include this before other type declarations that use it in an intersection type `& BindingTypeInfo`

kainino0x:
> `kTextureBindingTypeInfo` and friends so that it's easy to grab just the texture binding types in tests.

It's more than just that; some of the fields like `usage: TextureUsage` are only accessible via `kTextureBindingTypeInfo` (they're hidden when merged into `kBindingTypeInfo`).

> ```ts
>   getBindingInfo(bindingType: GPUBindingType, ...features: BindingInfoFeature[]) : FancyStuffToInferResultType {
> ```

I think code that calls this would end up doing:
```ts
  const { a, b, c } = getBindingInfo(type, ['a', 'b', 'c']);
```
which is more verbose/redundant than just extracting the fields you need.

---

In the last commit I shuffled things slightly to make it more readable. WDYT?

src/webgpu/api/validation/createBindGroupLayout.spec.ts
```diff
@@ -93,7 +94,8 @@ g.test('dynamic set to true is allowed only for buffers')
   .params(poptions('type', kBindingTypes))
   .fn(async t => {
     const { type } = t.params;
-    const success = kBindingTypeInfo[type].type === 'buffer';
+    const success =
+      kPerPipelineBindingLimits[kBindingTypeInfo[type].perPipelineBindingLimitType].maxDynamic > 0;
```
austinEng, line 98: Would we be able to remove this two-level indirection? Or perhaps expose the binding limits on the BindingTypeInfo object so that you don't have to know to import another table to get them and instead do `kBindingTypeInfo[type].pipelineBindingLimits.maxDynamic`.

Perhaps we can have `kBindingTypeInfo = expand(kBindingTypeInfo)` where
```typescript
function expand(info: BindingTypeInfo): ExpandedThing {
  return {
      ...info,
      pipelineBindingLimits: kPerPipelineBindingLimits[info.perPipelineBindingLimitType]
  };
}
```

src/webgpu/capability_info.ts
```diff
@@ -65,76 +73,199 @@ export const kTextureFormatInfo: {
...
+  | 'compareSamp'
+  | 'sampledTex'
+  | 'storageTex';
+type ValidBindableResource =
+  | 'uniformBuf'
+  | 'storageBuf'
+  | 'plainSamp'
+  | 'compareSamp'
+  | 'sampledTex'
+  | 'storageTex';
```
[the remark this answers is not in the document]
kainino0x: Oh, yes. For some reason I couldn't think of a way to do that.

src/webgpu/api/validation/createBindGroupLayout.spec.ts
kainino0x: I was thinking about this but I'm not sure it makes sense. maxDynamic is not a property of the binding type because it is shared across multiple binding types. However "supportsDynamic" could be a property of the binding type...

austinEng: maybe just behind a function on the object, so it's not directly part of it? I think it'll be a bit nicer to use than looking up in another table. Even this would be nice:

```typescript
return {
      ...info,
      getPipelineBindingLimits: () => kPerPipelineBindingLimits[info.perPipelineBindingLimitType]
};
```

kainino0x: Eh, probably just removing the indirection would be ok. Probably not that unclear.

kainino0x:
Rebased and revised, how does the last commit look?

kainino0x:
> personally not a fan of "Class.class" but LGTM

Not sure how to name this and retain clarity. Open to suggestions (or PRs if you want to take a shot at any of this at any point) but I think it's pretty okay now.

PR #170, opened 2020-04-21
Don't allow multisampled 2d-array

kainino0x:
This is not supported on Metal until macOS 10.14+.

https://crbug.com/dawn/243
https://developer.apple.com/documentation/metal/mtltexturetype/mtltexturetype2dmultisamplearray?language=objc

~Blocked on #167~


kainino0x:
> Also, maybe we should encode what is a valid/invalid texture in some shared place so that here, we can use it to enumerate all valid/invalid textures. And then other tests (like the lazy zero it) can use it to filter out invalid textures.

Not sure how to do this right now so added a TODO in both files.

PR #171, opened 2020-04-23
Add more complex canvas reftest

kainino0x:
~dependent on #164~

### DIFF
--- a/src/webgpu/web-platform/reftests/canvas_complex.ts
+++ b/src/webgpu/web-platform/reftests/canvas_complex.ts
@@ -1,0 +1,78 @@
+import { unreachable, assert } from '../../../common/framework/util/util.js';
+
+import { runRefTest } from './gpu_ref_test.js';
+
+// <canvas> element from html page
+declare const cvs: HTMLCanvasElement;
+
+const formats = new URLSearchParams(window.location.search).getAll('format');
+assert(formats.length === 1);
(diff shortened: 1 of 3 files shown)


kainino0x:
@shaoboyan PTAL!

shaoboyan:
@kainino0x  LGTM(And it seems that this will cover the orientation too).  Is this tested under wpt?

src/webgpu/web-platform/reftests/canvas_complex.ts
```diff
@@ -0,0 +1,78 @@
...
+    usage: GPUTextureUsage.COPY_DST,
+  });
+
+  const rows = 2;
+  const bytesPerRow = 256;
+  const [buffer, mapping] = t.device.createBufferMapped({
+    size: rows * bytesPerRow,
+    usage: GPUBufferUsage.COPY_SRC,
+  });
+  switch (format) {
```
shaoboyan, line 35: A quick(silly) question, No rgba8unorm?

kainino0x: That's right. Initially I did test it, but actually I decided to disallow it in Chrome, as it's not supposed to be supported by default. (bgra8unorm and rgba16float were the two formats we tentatively agreed on for core.)

kainino0x:
Thanks for the review!

PR #172, opened 2020-04-23
Fix Travis build

kainino0x:


.travis.yml
```diff
@@ -2,8 +2,6 @@ language: node_js
 node_js:
   - 'node'
 
-before_install: yarn global add grunt-cli
```
kainino0x, line 5: this is not part of the fix, it just wasn't necessary anymore

Issue #173, opened 2020-04-23
expectSingleColorTextureContents


kainino0x:
> A lot of this looks like maybe it could be a general expectSingleColorTextureContents helper. Could you consider whether that could be abstracted in a followup if it would be used elsewhere?

_Originally posted by @kainino0x in https://github.com/gpuweb/cts/pull/165/files_

PR #175, opened 2020-04-24
Use expectSingleColor helper in glsl-dependent tests

austinEng:
This function checks that the entire subresource region of some texture is entirely filled by a single color.
For #173

### DIFF
--- a/src/webgpu/api/operation/command_buffer/render/storeop.spec.ts
+++ b/src/webgpu/api/operation/command_buffer/render/storeop.spec.ts
@@ -9,7 +9,7 @@ export const g = new TestGroup(GPUTest);
 
 g.test('storeOp controls whether 1x1 drawn quad is stored')
   .params([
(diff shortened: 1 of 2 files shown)


austinEng:
@kainino0x PTAL, I'll split the two commits and make another PR for the first half to master after review

src/webgpu/gpu_test.ts
```diff
@@ -225,4 +233,79 @@ export class GPUTest extends Fixture {
...
+    format: GPUTextureFormat,
+    size: [number, number, number],
+    dimension: GPUTextureDimension,
+    options: ExpectSingleColorOptions,
+    exp: PerTexelComponent<number>
+  ): void;
+  expectSingleColor(
+    src: GPUTexture,
+    format: GPUTextureFormat,
+    size: [number, number, number],
```
kainino0x, line 261: getTextureCopyLayout uses this as the base (level 0) size, but copyTextureToBuffer uses it as the non-base size

PR #176, opened 2020-04-24
Introduce expectSingleColor helper function

austinEng:
This function checks that the entire subresource region of some texture is entirely filled by a single color.


kainino0x:
And merged into `glsl-dependent`

PR #177, opened 2020-04-29
Handle device loss and popErrorScope hangs in test finalization

kainino0x:


src/webgpu/gpu_test.ts
```diff
@@ -29,15 +35,35 @@ type TypedArrayBufferViewConstructor =
...
+  device: GPUDevice;
+  lostReason?: string;
+}
+
+const kPopErrorScopeTimeoutMS = 5000;
+
 class DevicePool {
-  device: GPUDevice | undefined = undefined;
-  state: 'free' | 'acquired' | 'uninitialized' | 'failed' = 'uninitialized';
+  holder?: DeviceHolder = undefined; // undefined means "uninitialized"
```
austinEng, line 46: As well as "lost"

src/webgpu/gpu_test.ts
```diff
@@ -29,15 +35,35 @@ type TypedArrayBufferViewConstructor =
...
+        device: await adapter.requestDevice(),
+        lostReason: undefined,
+      };
+
+      holder.device.lost.then(ev => {
+        holder.lostReason = ev.message;
+      });
+
+      this.holder = holder;
+      this.state = 'free';
```
austinEng, line 66: IIUC, the holder is set to `undefined` when we discover on test finalization that the device has been lost, or we think there may have been an unrecoverable timeout. That means we should no longer try to use the device and instead make a new one on the next test run.

Comments in release() are helpful; I think it would help to add some comments about the device lifecycle here as well.

kainino0x:
big new diff, but should hopefully be a lot less confusing and error prone now. PTAL

kainino0x:
hm, forgot to test it. Need to fix a state issue still.

kainino0x:
Okay, fixed.

src/webgpu/gpu_test.ts
```diff
@@ -29,39 +34,136 @@ type TypedArrayBufferViewConstructor =
...
-    this.state = 'free';
+    if (gpuValidationError !== null) {
+      assert(gpuValidationError instanceof GPUValidationError);
+      throw new TestFailedButDeviceReusable(
+        `Unexpected validation error occurred: ${gpuValidationError.message}`
+      );
+    }
+    if (gpuOutOfMemoryError !== null) {
+      assert(gpuOutOfMemoryError instanceof GPUOutOfMemoryError);
+      throw new TestFailedButDeviceReusable('Unexpected out-of-memory error occurred');
```
austinEng, line 165: If we had an unexpected OOM, I wonder if it's a good idea to reuse the device as it may cause OOM as well in other unrelated tests.

For some classes of leaks where the Device is holding onto unreachable resources and keeping things alive, it would be better to discard that device. Maybe on unexpected OOMs we can `window.gc` before the next test run. That's outside of the scope of this PR though.

src/webgpu/gpu_test.ts
```diff
@@ -29,39 +34,136 @@ type TypedArrayBufferViewConstructor =
...
+    if (this.holder === undefined) {
+      try {
+        this.holder = await DevicePool.makeHolder();
+      } catch (ex) {
+        this.failed = true;
+        throw ex;
+      }
+    } else {
+      assert(!this.holder.acquired, 'Device was in use on DevicePool.acquire');
+    }
```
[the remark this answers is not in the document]
kainino0x: Good call.

src/webgpu/gpu_test.ts
kainino0x: Good point; I'll change this error type.

We have attemptGarbageCollection so it would be easy to do that. I'll do it in a followup.

PR #178, opened 2020-05-02
Attempt GC on unexpected GPU OOM

kainino0x:
Also makes attemptGarbageCollection async just in case yielding to the event loop helps the GC run.
(It's okay if a caller forgets to await; it otherwise has no effect.)


src/webgpu/gpu_test.ts
```diff
@@ -162,6 +163,10 @@ class DevicePool {
     }
     if (gpuOutOfMemoryError !== null) {
       assert(gpuOutOfMemoryError instanceof GPUOutOfMemoryError);
+
+      // Try to clean up, in case there are stray GPU resources in need of collection.
+      await attemptGarbageCollection();
```
austinEng, line 168: nit: a future improvement may be to do this _after_ setting `this.holder` to be undefined in `release(GPUDevice)` so that the GPUDevice is unreachable. More complexity involved with bubbling that signal out and hard to know how useful it will be, so a TODO is also okay.

kainino0x: Not so hard to do, and if we don't do it now we may never do it. Take a look at the first new commit (second one is just a move)

PR #183, opened 2020-05-15
Don't use variants in reftests

kainino0x:
Turns out WPT doesn't support test variants in reftests.

**Recommend reviewing [without whitespace](https://github.com/gpuweb/cts/pull/183/files?diff=split&w=1)**


kainino0x:
@shaoboyan PTAL

kainino0x:
Oops, accidentally uploaded this twice. Closing in favor of #185 (already merged).

PR #186, opened 2020-05-18
only checkForDuplicateCases during crawl(); slightly speed up comparePublicParamsPaths

kainino0x:
Avoids doing checkForDuplicateCases when just importing a `.spec.js` file.

Significantly speeds up page load time for standalone and wpt.

### DIFF
--- a/src/common/framework/test_group.ts
+++ b/src/common/framework/test_group.ts
@@ -5,6 +5,7 @@ import {
   CaseParamsIterable,
   extractPublicParams,
   publicParamsEquals,
+  paramKeyIsPublic,
 } from './params_utils.js';
 import { kPathSeparator } from './query/separators.js';
 import { stringifySingleParam } from './query/stringify_params.js';
@@ -34,6 +35,7 @@ export function makeTestGroup<F extends Fixture>(fixture: FixtureClass<F>): Test
 // Interface for running tests
 export interface RunCaseIterable {
   iterate(): Iterable<RunCase>;
+  checkForDuplicateCases(): void;
 }
 export function makeTestGroupForUnitTesting<F extends Fixture>(
   fixture: FixtureClass<F>
@@ -84,6 +86,12 @@ class TestGroup<F extends Fixture> implements RunCaseIterable, TestGroupBuilder<
     this.tests.push(test);
     return test;
   }
+
+  checkForDuplicateCases(): void {
+    for (const test of this.tests) {
+      test.checkForDuplicateCases();
+    }
+  }
 }
 
 interface TestBuilderWithName<F extends Fixture, P extends {}> extends TestBuilderWithParams<F, P> {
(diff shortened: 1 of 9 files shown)


src/common/framework/test_group.ts
```diff
@@ -109,27 +117,31 @@ class TestBuilder<F extends Fixture, P extends {}> {
...
+      for (const [k, v] of Object.entries(testcase as CaseParams)) {
+        if (paramKeyIsPublic(k)) {
+          stringifySingleParam(k, v); // To check for invalid params values
+        }
       }
 
       assert(
-        !seen.some(x => publicParamsEquals(x, publicParams)),
-        'Duplicate test case params: ' + JSON.stringify(publicParams)
+        !seen.some(x => publicParamsEquals(x, testcase)),
```
kainino0x, line 135: Come to think of it, I don't know why this isn't a Set.

kainino0x: Done

src/common/framework/test_group.ts
```diff
@@ -109,27 +111,23 @@ class TestBuilder<F extends Fixture, P extends {}> {
...
-      assert(
-        !seen.some(x => publicParamsEquals(x, publicParams)),
-        'Duplicate test case params: ' + JSON.stringify(publicParams)
-      );
-      seen.push(publicParams);
+    const seen = new Set<string>();
+    for (const testcase of this.cases) {
+      // stringifyPublicParams also checks for invalid params values
+      const testcaseString = stringifyPublicParams(testcase).join(kParamSeparator);
+      assert(!seen.has(testcaseString));
```
austinEng: nit: lost the assert message ?

src/unittests/test_group.spec.ts
```diff
@@ -96,30 +96,32 @@ g.test('duplicate_test_name').fn(t => {
...
   t.shouldThrow('Error', () => {
-    g.test('abc')
-      .params([
-        { a: 1, _b: 1 }, //
-        { a: 1, _b: 2 },
-      ])
-      .fn(() => {
-        //
-      });
+    g.checkForDuplicateCases();
```
austinEng, line 124: nit: And add a test that g.checkForDuplicateCases doesn't throw for unparameterized tests

src/common/framework/test_group.ts
austinEng: nvm you fixed it

src/unittests/test_group.spec.ts
kainino0x: bleh, in testing I remembered why it was supposed to be n^2: if params are `{a:1,b:2}, {b:2,a:1}`

austinEng: Oh, I missed that you had switched it to a Set and back. It can't be so bad anyway, there should be a someone reasonable-ish number of params.

kainino0x: Yeah, it's fine since now that it's only at build time. It might become a problem eventually, but that can be dealt with then.

kainino0x:
Okay, finished revising. Could you take another quick look?

PR #187, opened 2020-05-18
standalone runner: implement collapsing, show relative names

kainino0x:
DOM is not lazy-loaded, but at least every case is not displayed by default.

The default state is everything expanded, except for:
- top level of a file (`s:f:*`)
- top level of a test (`s:f:t:*`)

Followup: should collapse logs for each case, as well.

### DIFF
--- a/src/common/framework/tree.ts
+++ b/src/common/framework/tree.ts
@@ -9,6 +9,7 @@ import {
   TestQueryMultiFile,
   TestQueryMultiTest,
 } from './query/query.js';
+import { kBigSeparator, kWildcard, kPathSeparator, kParamSeparator } from './query/separators.js';
 import { stringifySingleParam } from './query/stringify_params.js';
 import { RunCase, RunFn } from './test_group.js';
 import { assert } from './util/util.js';
@@ -38,6 +39,8 @@ import { assert } from './util/util.js';
 //     about expectation granularity.
 
 export interface TestSubtree<T extends TestQuery = TestQuery> {
+  /** For display in standalone runner. */
+  readonly relativeName: string;
   readonly query: T;
   readonly children: Map<string, TestTreeNode>;
   readonly collapsible: boolean;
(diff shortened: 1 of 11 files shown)


kainino0x:
I may revise this based on my experience actually using the standalone runner to debug failures.

austinEng:
Did you still want me to look at this now?

kainino0x:
Not yet, I'll let you know when I get to it.

kainino0x:
Okay, PTAL. No need for in depth review, especially of the DOM and CSS stuff. I took out dissolveSingleChildNodes so there's not much else, it's very assorted.

src/common/framework/tree.ts
```diff
@@ -282,7 +286,11 @@ function addSubtreeForTestPath(
         tree.query.filePathParts,
         subqueryTest
```
austinEng, line 287: I noticed here that we're passing a reference to `subqueryTest` into different `TestQueryMultiTest` objects, and appending to the array as we go. That doesn't seem like what we want?

Nevermind, I see the constructor makes a copy. Maybe we should make a copy before constructing otherwise callsites like this are a bit confusing.

This one could be:
```typescript
for (const [i, part] of test.entries()) {
  new TestQueryMultiTest(..., test.slice(0, i + 1));
}
```

And at [1] we would just pass `test` because at that point I think the arrays are the same.

src/common/framework/tree.ts
```diff
@@ -293,7 +301,12 @@ function addSubtreeForTestPath(
       subqueryTest,
       {}
     );
-    return { query, collapsible: isCollapsible(query) };
+    return {
+      relativeName: subqueryTest[subqueryTest.length - 1] + kBigSeparator + kWildcard,
```
austinEng, line 305: unit: assert non-zero length

src/common/framework/tree.ts
```diff
@@ -261,7 +260,12 @@ function addSubtreeForFilePath(
   // This goes from that -> suite:a,b:*
   const subtree = getOrInsertSubtree('', tree, () => {
     const query = new TestQueryMultiTest(tree.query.suite, tree.query.filePathParts, []);
-    return { query, description, collapsible: checkCollapsible(query) };
+    return {
+      relativeName: file[file.length - 1] + kBigSeparator + kWildcard,
```
austinEng, line 264: unit: assert non-zero length

src/common/framework/tree.ts
```diff
@@ -293,7 +301,12 @@ function addSubtreeForTestPath(
       subqueryTest,
```
austinEng, line 301: [1]

src/common/framework/tree.ts
kainino0x: It is confusing, but it's also less error prone. I did it this way so I could be sure I wasn't missing any places that needed copies.

Doing the copy in the constructor makes the TestQuery classes immutable by construction, which reduces cognitive load working with them.

PR #188, opened 2020-05-18
Reduce logging verbosity

kainino0x:
Excess verbosity was making it difficult to navigate both the standalone runner and the logs from our WPT bots.

### DIFF
--- a/src/common/framework/logging/log_message.ts
+++ b/src/common/framework/logging/log_message.ts
@@ -1,8 +1,8 @@
 import { extractImportantStackTrace } from '../util/stack.js';
 
 export class LogMessageWithStack extends Error {
-  printStack: boolean = true;
-  timesSeen: number = 1;
+  private stackHidden: boolean = false;
+  private timesSeen: number = 1;
 
   constructor(name: string, ex: Error) {
     super(ex.message);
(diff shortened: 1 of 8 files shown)


src/common/framework/logging/log_message.ts
```diff
@@ -1,21 +1,27 @@
 import { extractImportantStackTrace } from '../util/stack.js';
 
 export class LogMessageWithStack extends Error {
-  constructor(name: string, ex: Error, includeStack: boolean = true) {
+  printStack: boolean = true;
+  timesSeen: number = 1;
```
austinEng, line 5: Either comment that these members can and should be set externally, or maybe add some setters for them. Seems like we only ever increment `timesSeen` so we could have a method to do only that.

src/common/framework/logging/test_case_recorder.ts
```diff
@@ -48,35 +63,63 @@ export class TestCaseRecorder {
...
-    this.logs.push(new LogMessageWithStack('EXCEPTION', ex));
+    this.logImpl(LogSeverity.ThrewException, new LogMessageWithStack('EXCEPTION', ex));
   }
 
-  private setState(state: PassState): void {
-    this.state = Math.max(this.state, state);
+  private logImpl(level: LogSeverity, logMessage: LogMessageWithStack): void {
+    // Deduplicate errors with the exact same stack
+    if (logMessage.stack) {
+      const seen = this.seenStacks.get(logMessage.stack);
```
austinEng, line 98: "seen" makes it seem like `seenStacks` is a map of stack to boolean. Perhaps rename it to "seenLogMessage"

PR #191, opened 2020-05-19
Index format tests

haoxli:
Add test cases for testing drawIndexed, IndexFormat, Primitive restart


Kangz:
This seems to be mostly a port of the Dawn's end2end tests. They are ok for there, but definitely not comprehensive enough. Please take a look at the [CTS guidelines](https://github.com/gpuweb/gpuweb/wiki/WebGPU-CTS-guidelines) for an explanation of what the CTS tests should look like.

haoxli:
There are too many combinations for the check points, I will submit separately, so close this one.

PR #193, opened 2020-05-20
standalone: dissolveLevelBoundaries

kainino0x:
Reduces node count from 6460 to 4171, making things much more navigable.

Dependent on #187


kainino0x:
PTAL (second commit)

src/common/framework/tree.ts
```diff
@@ -377,3 +398,44 @@ function insertLeaf(parent: TestSubtree, query: TestQuerySingleCase, t: RunCase)
...
+
+/** Generate a readable relative name for a case (used in standalone). */
+function readableNameForCase(query: TestQuerySingleCase): string {
+  let name: string;
+  const paramsKeys = Object.keys(query.params);
+  if (paramsKeys.length === 0) {
+    name = query.testPathParts[query.testPathParts.length - 1] + kBigSeparator;
+  } else {
+    const lastKey = paramsKeys[paramsKeys.length - 1];
+    name = stringifySingleParam(lastKey, query.params[lastKey]);
```
austinEng, line 438: Why do we only show a single param? Is it not possible to a single test with a single leaf node with multiple params?

kainino0x: It's really subtle, but it works out because we only dissolve one level boundary (the one between multi-case and single-case), and not every single-child subtree (like I was doing before).
image

kainino0x:
latest: 
image

kainino0x:
The sibling 'x=0' and 'y=0' are weird, but ultimately I think it doesn't matter.

kainino0x:
image

austinEng:
> The sibling 'x=0' and 'y=0' are weird (they're not actually siblings), but ultimately I think it doesn't matter.

Does that test run twice then? or is it like `x=0;y=undefined` (thought we explicitly printed undefined though)

kainino0x:
No, it only runs each case (`x=0`, `x=0;y=0`) once. It's just a matter of those being collapsed down to looking like siblings when they weren't before.

kainino0x:
The actual types are:
image

austinEng:
Okay, I see now. I thought it was splitting a single param {x=0;y=0} up, but the single param is what `case_depth_2_in_single_child_test` is doing.

PR #196, opened 2020-05-28
Replace usages of constants.ts with `as const`

kainino0x:
I discovered `as const` today. It's really useful.

- [ ] Probably needs more work on `glsl-dependent` after landing

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -37,7 +37,7 @@ module.exports = function (grunt) {
       },
       lint: {
         cmd: 'node',
-        args: ['node_modules/eslint/bin/eslint', 'src/**/*.ts'],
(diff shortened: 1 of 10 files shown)


Kangz:
I'm not sure I understand what this does, is it because previously typechecking didn't work in some cases unless you used named constants?

kainino0x:
yup, e.g.
`{ a: 'somestring' }` is type `{ a: string }`
but
`{ a: 'somestring' as const }` is type `{ a: 'somestring' }`
and
`{ a: 'somestring' } as const` is type `{ readonly a: 'somestring' }`

kainino0x:
Given this feature exists, the string constants in constants.ts might even be unnecessary. WDYT @sebavan ?

sebavan:
Should try it at least, :-)

PR #198, opened 2020-05-29
[glsl-dependent] Replace some usages of constants.ts with `as const`

kainino0x:
I had to fix up some of the remaining usages with e.g. `'float' as GPUTextureFormat` (edit: needed if the field gets reassigned later). I previously avoided that as it's not typo safe, so I am sort of considering add a helper function to safely assert the type `is<GPUTextureFormat>('float')`. Unfortunately TS doesn't have a way to do it without creating a function call (https://github.com/microsoft/TypeScript/issues/7481).

Removes more runtime dependence on constants.ts.

### DIFF
--- a/src/webgpu/api/operation/command_buffer/render/storeop.spec.ts
+++ b/src/webgpu/api/operation/command_buffer/render/storeop.spec.ts
@@ -1,7 +1,6 @@
 export const description = `
 renderPass store op test that drawn quad is either stored or cleared based on storeop`;
 
-import * as C from '../../../../../common/constants.js';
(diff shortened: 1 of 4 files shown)


austinEng:
Sorry, not completely sure what you mean. Why is it that sometimes you use 'as const' and sometimes 'as GPUVertexFormat' ?

kainino0x:
```ts
const desc = { a: 'float' as const };
desc.a = 'float2';
```
isn't valid since `desc` gets type `{ a: 'float' }`.

kainino0x:
thanks!

PR #202, opened 2020-06-01
Fence cross device signal tests

### LINKED ISSUE #201: Different device Queue and Fence tests send validation error to the wrong receiver
In the following tests:

webgpu:api,validation,fences:signal_a_fence_on_a_different_device_than_it_was_created_on_is_invalid:
webgpu:api,validation,fences:signal_a_fence_on_a_different_device_does_not_update_fence_signaled_value:

A fence is signaled on a queue from a different device than the fence. The test expects a validation error on the fence's device. This is wrong, and the error should be sent to the "receiver" object, the queue.

austinEng:
Fixes #201. The test originally checked for the validation error
on the wrong device.

### DIFF
--- a/src/webgpu/api/validation/fences.spec.ts
+++ b/src/webgpu/api/validation/fences.spec.ts
@@ -56,28 +56,32 @@ g.test('increasing_fence_value_by_more_than_1_succeeds').fn(async t => {
 });
 
 g.test('signal_a_fence_on_a_different_device_than_it_was_created_on_is_invalid').fn(async t => {
-  const fence = t.queue.createFence();
-
   const anotherDevice = await t.device.adapter.requestDevice();
-  const anotherQueue = anotherDevice.defaultQueue;
+  const fence = anotherDevice.defaultQueue.createFence();
 
   t.expectValidationError(() => {
-    anotherQueue.signal(fence, 2);
+    t.queue.signal(fence, 2);
   });
 });
 
 g.test('signal_a_fence_on_a_different_device_does_not_update_fence_signaled_value').fn(async t => {
-  const fence = t.queue.createFence({ initialValue: 1 });
-
   const anotherDevice = await t.device.adapter.requestDevice();
-  const anotherQueue = anotherDevice.defaultQueue;
+  const fence = anotherDevice.defaultQueue.createFence({ initialValue: 1 });
 
   t.expectValidationError(() => {
-    anotherQueue.signal(fence, 2);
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/fences.spec.ts
```diff
@@ -56,28 +56,32 @@ g.test('increasing_fence_value_by_more_than_1_succeeds').fn(async t => {
...
 
   t.expectValidationError(() => {
-    anotherQueue.signal(fence, 2);
+    t.queue.signal(fence, 2);
   });
 
   t.expect(fence.getCompletedValue() === 1);
 
-  t.queue.signal(fence, 2);
+  anotherDevice.pushErrorScope('validation');
```
kainino0x, line 77: Can we factor expectValidationErrorOnDevice out from ValidationTest and use it here?

austinEng: This is expecting no errors. We can have `expectNoErrorsOnDevice`?

austinEng: I guess what we'd really want is to make a new `DevicePool` and use the existing `beginErrorScopes` and `endErrorScopes`

kainino0x: expectValidationError takes an optional `shouldError: boolean = true` which we can keep for this helper. (Perhaps someday we should invert that, though, so it's `expectValidationResult(shouldSucceed: boolean, fn: Function)`.)

kainino0x: hm, a DevicePool or something like it could work here. We'd want a helper to return a new device so that we can make sure to check for errors and clean up at the end of tests that use it. Maybe we can factor DevicePool into two parts, a pool and an RAII-style device wrapper? (Where the "destructor" part of the wrapper is handled manually in GPUTest, because JS.)

austinEng: mm okay we'll have to update the handling of that argument though to actually make a new scope on the device if it's not `this.device`. Right now it lets DevicePool catch it.

```ts
    // If no error is expected, we let the scope surrounding the test catch it.
    if (shouldError === false) {
      fn();
      return;
    }
```

kainino0x: Ah right, good point.

kainino0x: Let's not block this on that, then.

PR #203, opened 2020-06-02
Add missing storageTextureFormat to BGL descriptor

### LINKED ISSUE #200: createPipelineLayout:visibility_and_dynamic_offsets: produces invalid BGL descriptor
webgpu:api,validation,createPipelineLayout:visibility_and_dynamic_offsets:

iterates through combinations of bindings types, visibility, and dynamic true/false.
For storage textures it is missing the `storageTextureFormat` in the `GPUBindGroupLayoutDescriptor`

It is probably sufficient to just put `r32uint` here.

austinEng:
Fixes #200. createPipelineLayout:visibility_and_dynamic_offsets
tests were incorrectly missing the storageTextureFormat for storage
texture binding types.

### DIFF
--- a/src/webgpu/api/validation/createPipelineLayout.spec.ts
+++ b/src/webgpu/api/validation/createPipelineLayout.spec.ts
@@ -85,8 +85,12 @@ g.test('visibility_and_dynamic_offsets')
     const { type, hasDynamicOffset, visibility } = t.params;
     const info = kBindingTypeInfo[type as GPUBindingType];
 
+    const storageTextureFormat =
+      type === 'readonly-storage-texture' || type === 'writeonly-storage-texture'
+        ? ('r32uint' as const)
+        : undefined;
     const descriptor = {
-      entries: [{ binding: 0, visibility, type, hasDynamicOffset }],
+      entries: [{ binding: 0, visibility, type, hasDynamicOffset, storageTextureFormat }],
     };
 
     const supportsDynamicOffset = kBindingTypeInfo[type].perPipelineLimitClass.maxDynamic > 0;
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/createPipelineLayout.spec.ts
```diff
@@ -85,8 +85,12 @@ g.test('visibility_and_dynamic_offsets')
     const { type, hasDynamicOffset, visibility } = t.params;
     const info = kBindingTypeInfo[type as GPUBindingType];
 
+    const storageTextureFormat =
+      type === 'readonly-storage-texture' || type === 'writeonly-storage-texture'
+        ? ('r32uint' as const)
+        : undefined;
```
austinEng, line 91: Should it be valid to always specify this, even for non-storage texture formats?

src/webgpu/api/validation/createPipelineLayout.spec.ts
```diff
@@ -85,8 +85,12 @@ g.test('visibility_and_dynamic_offsets')
     const { type, hasDynamicOffset, visibility } = t.params;
     const info = kBindingTypeInfo[type as GPUBindingType];
 
+    const storageTextureFormat =
+      type === 'readonly-storage-texture' || type === 'writeonly-storage-texture'
```
kainino0x, line 89: nit: can be `info.resource === 'storageBuf'`

src/webgpu/api/validation/createPipelineLayout.spec.ts
kainino0x: I'm not sure. Currently the spec allows it. I'm iffy on whether we should, though, because it allows storageTextureFormat to be completely ignored (we don't check it at all, beyond bindings, if it's not a storage texture) so it could have any random texture format in it. I think we should keep the test like this.

src/webgpu/api/validation/createPipelineLayout.spec.ts
austinEng: Done

src/webgpu/api/validation/createPipelineLayout.spec.ts
kainino0x: Filed https://github.com/gpuweb/gpuweb/issues/829

PR #205, opened 2020-06-03
Expect GPUOutOfMemoryError in map_oom tests

### LINKED ISSUE #199: webgpu:api,operation,buffers,map_oom: only checks for a JS exception
This test checks for a JS exception, expecting that mapping an enormous range would cause a RangeError due to failure to allocate an ArrayBuffer that large. On some platforms, this allocation actually succeeds, and then the test fails with an OOM error when the implementation attempts to map such a large GPU buffer.

I suggest this test be updated to allow a GPU OOM error if there is no JS exception.

austinEng:
Fixes #199. Also adds a helper to asynchronously expect errors.

### DIFF
--- a/src/webgpu/api/operation/buffers/map_oom.spec.ts
+++ b/src/webgpu/api/operation/buffers/map_oom.spec.ts
@@ -3,27 +3,31 @@ export const description = '';
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
 import { GPUTest } from '../../../gpu_test.js';
 
-function getBufferDesc(): GPUBufferDescriptor {
+function getBufferDesc(usage: GPUBufferUsageFlags): GPUBufferDescriptor {
   return {
     size: Number.MAX_SAFE_INTEGER,
-    usage: GPUBufferUsage.MAP_WRITE,
+    usage,
   };
 }
 
 export const g = makeTestGroup(GPUTest);
 
 g.test('mapWriteAsync').fn(async t => {
-  const buffer = t.device.createBuffer(getBufferDesc());
+  const buffer = t.expectError('out-of-memory', () => {
+    return t.device.createBuffer(getBufferDesc(GPUBufferUsage.MAP_WRITE));
+  });
   t.shouldReject('RangeError', buffer.mapWriteAsync());
 });
 
 g.test('mapReadAsync').fn(async t => {
-  const buffer = t.device.createBuffer(getBufferDesc());
+  const buffer = t.expectError('out-of-memory', () => {
+    return t.device.createBuffer(getBufferDesc(GPUBufferUsage.MAP_READ));
+  });
   t.shouldReject('RangeError', buffer.mapReadAsync());
 });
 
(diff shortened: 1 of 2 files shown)


austinEng:
I believe this will probably fail on Windows NVIDIA in Chrome Canary (because it doesn't actually OOM) somehow, though I haven't been able to repro locally on my Windows NVIDIA workstation. Probably a Dawn/Chrome bug though.

austinEng:
I realized I should make this against `master`. I'll rebase.

src/webgpu/gpu_test.ts
```diff
@@ -226,4 +226,40 @@ got [${failedByteActualValues.join(', ')}]`;
     fillTextureDataWithTexelValue(expectedTexelData, format, dimension, arrayBuffer, size, layout);
     this.expectContents(buffer, new Uint8Array(arrayBuffer));
   }
+
+  expectError<R>(filter: GPUErrorFilter, fn: () => R): R {
```
kainino0x, line 230: nit: perhaps `expectGPUError`?

src/webgpu/api/operation/buffers/map_oom.spec.ts
```diff
@@ -3,27 +3,31 @@ export const description = '';
...
 }
 
 export const g = makeTestGroup(GPUTest);
 
 g.test('mapWriteAsync').fn(async t => {
-  const buffer = t.device.createBuffer(getBufferDesc());
+  const buffer = t.expectError('out-of-memory', () => {
+    return t.device.createBuffer(getBufferDesc(GPUBufferUsage.MAP_WRITE));
+  });
   t.shouldReject('RangeError', buffer.mapWriteAsync());
```
kainino0x, line 19: I think these should be DOMExceptions (OperationErrors) now.

src/webgpu/api/operation/buffers/map_oom.spec.ts
```diff
@@ -3,27 +3,31 @@ export const description = '';
...
 g.test('mapReadAsync').fn(async t => {
-  const buffer = t.device.createBuffer(getBufferDesc());
+  const buffer = t.expectError('out-of-memory', () => {
+    return t.device.createBuffer(getBufferDesc(GPUBufferUsage.MAP_READ));
+  });
   t.shouldReject('RangeError', buffer.mapReadAsync());
 });
 
 g.test('createBufferMapped').fn(async t => {
   t.shouldThrow('RangeError', () => {
```
kainino0x, line 30: (should still be RangeError probably)

kainino0x:
> I realized I should make this against `master`. I'll rebase.

Huh. I didn't realize you could change which branch a PR is against.

src/webgpu/gpu_test.ts
austinEng: Done

austinEng:
> Huh. I didn't realize you could change which branch a PR is against.

I didn't either; maybe it's a new Github thing.

PR #208, opened 2020-06-09
Lint for imports from wrong directories

kainino0x:


.eslint-resolver.js
```diff
@@ -0,0 +1,22 @@
+const path = require('path');
+const resolve = require('resolve')
+
+exports.interfaceVersion = 2
+
+const dotJS = /\.js$/;
+exports.resolve = function (source, file, config) {
+  if (resolve.isCore(source)) return { found: true, path: null }
+
+  source = source.replace(dotJS, '.ts');
```
austinEng, line 10:
```suggestion
  source = source.replace(/\.js$/, '.ts');
```
nit: seems like we could just inline the `dotJS` string here

.eslint-resolver.js
```diff
@@ -0,0 +1,22 @@
+const path = require('path');
+const resolve = require('resolve')
+
```
austinEng, line 3: nit: could use a link to the documentation for the interface this file implements
```suggestion

// Implements the following resolver spec:
// https://github.com/benmosher/eslint-plugin-import/blob/master/resolvers/README.md
```

.eslint-resolver.js
```diff
@@ -0,0 +1,22 @@
...
+exports.interfaceVersion = 2
+
+const dotJS = /\.js$/;
```
austinEng, line 6:
```suggestion
```

PR #209, opened 2020-06-09
Switch to npm instead of yarn, and update deps

kainino0x:
Using npm slightly reduces unnecessary dependencies of the project.


kainino0x:
My understanding is that npm has solved all of the major issues that yarn originally solved. Performance is fine for me: second `rm -r node_modules && time npm install` takes 20.4s, second `rm -r node_modules && time yarn` takes 20.2s. (First `npm install` took 73s, I think filling its cache.)

kainino0x:
Point is just to avoid developers having to install yarn on their system, since npm is packaged with node.

PR #210, opened 2020-06-09
Don't export LICENSE to WPT export, add AUTHORS.md

kainino0x:
Our license is the same as WPT's license. Excluding this license from
the export reduces the apparent complexity of WPT's licensing (by
removing a redundant extra license file).

Update CONTRIBUTING.md to explain this.

PR #212, opened 2020-06-09
Use Babel instead of copy to generate out-wpt/

kainino0x:
This lets us use different Babel options for out-wpt/.

Also tweak Grunt targets.

(dependent on #210 #211)


Gruntfile.js
```diff
@@ -34,6 +38,24 @@ module.exports = function (grunt) {
           'src/',
         ],
       },
+      'build-out-wpt': {
+        cmd: 'node',
+        args: [
+          'node_modules/@babel/cli/bin/babel',
+          '--source-maps=false',
```
kainino0x, line 45: note: source maps are now disabled for the wpt export (which will clean up a lot of useless console warnings). Alternatives: could embed the source inline as a giant base64 blob, or could even include the source files in the export.

austinEng: --source-maps inline would be nice if it doesn't cause any issues

Gruntfile.js
```diff
@@ -35,6 +38,24 @@ module.exports = function (grunt) {
...
+          '--extensions=.ts',
+          '--delete-dir-on-start',
+          '--out-dir=out-wpt/',
+          '--only=src/common/framework/',
+          '--only=src/common/runtime/helper/',
+          '--only=src/common/runtime/wpt.ts',
+          '--only=src/webgpu/',
+          '--ignore=src/*/listing.ts',
```
austinEng, line 53: should this be ** ?
```suggestion
          '--ignore=src/**/listing.ts',
```

kainino0x: No, we only have listings at the root of each suite. This needs to not match `src/common/framework/listing.ts` although that's incredibly unclear. I'll change this to just explicitly list the listing file since there's only one relevant one here.

Gruntfile.js
kainino0x: It should work fine but the file size is about 4x (webgpu/ 172kB -> 687kB). For that I'm not sure whether it's worth it.

I also just (re?)discovered that using the retainLines option will fix the annoyance where comments were always being printed at the end of the last line, so I guess I'll turn that on.

austinEng: I'm okay either way. Our compiled sources are pretty readable anyway.

kainino0x: Okay, I'll leave them disabled for now.

PR #220, opened 2020-06-22
Replace optional chaining operator with equivalent syntax

kunalmohan:
Optional chaining has recently been introduced and not all browsers support it yet. It would be good if we could avoid using that. I am working on implementing WebGPU in [Servo](https://github.com/servo/servo), which doesn't support optional chaining yet.

### DIFF
--- a/src/webgpu/gpu_test.ts
+++ b/src/webgpu/gpu_test.ts
@@ -235,7 +235,11 @@ got [${failedByteActualValues.join(', ')}]`;
 
     const commandEncoder = this.device.createCommandEncoder();
     commandEncoder.copyTextureToBuffer(
-      { texture: src, mipLevel: layout?.mipLevel, arrayLayer: slice },
+      {
+        texture: src,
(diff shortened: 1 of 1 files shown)


kainino0x:
Oh also, this change needs to be made against the `master` branch rather than the `glsl-dependent` branch.

kainino0x:
BTW I'm not 100% sure how to do this. I recently removed @babel/preset-env (intentionally as it made the code generation a lot cleaner). But it seems like it's needed if you want to do this configuration (which I didn't realize).

Ideally we could turn @babel/preset-env back on, but configure it so that it outputs the latest version supported by the major browsers, with hand-picked features disabled (like optional chaining).

kunalmohan:
I was going going for https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining. But it's under experimental plugins. Will that be okay?

kainino0x:
We didn't need that plugin because optional chaining in core Babel now. But if that lets you set things up to generate the necessary code with minimal other changes to the generated code, then that would be perfect!

PR #221, opened 2020-06-22
Use babel plugin to prevent optional chaining in generated code

kunalmohan:
Follow up from #220 

Please do not merge this yet. I'll test this locally to make sure we don't need any more changes.
(Sorry! probably should have done this before opening this PR)

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -4,6 +4,9 @@ module.exports = function (api) {
     presets: ['@babel/preset-typescript'],
     plugins: [
       '@babel/plugin-proposal-class-properties',
+      // Transforms optional chaining operator to equivalent ternary operator syntax.
+      // We need this to run tests on browsers that don't support optional chaining yet.
+      '@babel/plugin-proposal-optional-chaining',
       'const-enum',
       [
         'add-header-comment',
--- a/package-lock.json
+++ b/package-lock.json
@@ -240,6 +240,33 @@
         "@babel/helper-plugin-utils": "^7.10.1"
       }
     },
+    "@babel/plugin-proposal-optional-chaining": {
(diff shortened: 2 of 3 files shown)


kainino0x:
Looks great so far! Tested locally to make sure it has minimal impact on the generated code (and it does) :)

babel.config.js
```diff
@@ -4,6 +4,7 @@ module.exports = function (api) {
     presets: ['@babel/preset-typescript'],
     plugins: [
       '@babel/plugin-proposal-class-properties',
+      '@babel/plugin-proposal-optional-chaining',
```
kainino0x, line 7: Please add a comment about why we use this plugin even though it's supported in core.

kunalmohan:
I have added the comment and tested it locally. I am able to run the tests now, so these changes are enough.

kunalmohan:
@kainino0x how soon will these changes be reflected in the live tests and wpt?

kainino0x:
Normally they are automatically pushed to the live tests, I thought I set it up correctly for the main branch rename but I guess not. Thanks for catching that, I'll check on it.

WPT is rolled manually (currently only done through Chromium) so it won't happen until next time we get to it. Do you need it?

kunalmohan:
Updating the live tests should be enough at the moment.
I'll ping you once when we start using the wpt (if it hasn't been updated by then).
Thank you!

kainino0x:
Sounds good!

Live version fixed and rolled out: https://gpuweb.github.io/cts/out/common/framework/version.js

kunalmohan:
@kainino0x Now will be a good time to update cts in WPT :)

PR #223, opened 2020-06-30
Add Culling Test

natashalee:
This test logic was copied from the dawn_end2end_tests CullingTests. It's possible to add more formats to test, but I wasn't sure which ones were suitable to test with. PTAL and let me know how I can cover more cases.

Thanks!

### DIFF
--- a/src/webgpu/api/operation/render_pipeline/culling_tests.spec.ts
+++ b/src/webgpu/api/operation/render_pipeline/culling_tests.spec.ts
@@ -1,0 +1,162 @@
+export const description = `Test culling and rasterizaion state.
+
+Test coverage:
+Test all culling combinations of GPUFrontFace and GPUCullMode show the correct output.
+
+Use 2 triangles with different winding orders:
+
+- Test that the counter-clock wise triangle has correct output for:
+  - All FrontFaces (ccw, cw)
+  - All CullModes (none, front, back)
+  - All depth stencil attachment types (none, depth24plus, depth32float, depth24plus-stencil8)
+  - Some primitive topologies (triangle-list, TODO: triangle-strip)
+
+- Test that the clock wise triangle has correct output for:
+  - All FrontFaces (ccw, cw)
+  - All CullModes (none, front, back)
+  - All depth stencil attachment types (none, depth24plus, depth32float, depth24plus-stencil8)
+  - Some primitive topologies (triangle-list, TODO: triangle-strip)
+`;
+
+import { poptions, params } from '../../../../common/framework/params_builder.js';
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { GPUTest } from '../../../gpu_test.js';
+
+function faceIsCulled(face: 'cw' | 'ccw', frontFace: GPUFrontFace, cullMode: GPUCullMode): boolean {
+  return cullMode !== 'none' && (frontFace === face) === (cullMode === 'front');
+}
+
+function faceColor(face: 'cw' | 'ccw', frontFace: GPUFrontFace, cullMode: GPUCullMode): Uint8Array {
+  // front facing color is green, non front facing is red, background is blue
+  const isCulled = faceIsCulled(face, frontFace, cullMode);
+  if (!isCulled && face === frontFace) {
+    return new Uint8Array([0x00, 0xff, 0x00, 0xff]);
+  } else if (isCulled) {
+    return new Uint8Array([0x00, 0x00, 0xff, 0xff]);
+  } else {
+    return new Uint8Array([0xff, 0x00, 0x00, 0xff]);
+  }
+}
+
+export const g = makeTestGroup(GPUTest);
+
+g.test('culling')
+  .params(
+    params()
+      .combine(poptions('frontFace', ['ccw', 'cw'] as const))
+      .combine(poptions('cullMode', ['none', 'front', 'back'] as const))
+      .combine(
+        poptions('depthStencilFormat', [
+          null,
+          'depth24plus',
+          'depth32float',
+          'depth24plus-stencil8',
+        ] as const)
+      )
+      // TODO: test triangle-strip as well
+      .combine(poptions('primitiveTopology', ['triangle-list'] as const))
+  )
+  .fn(t => {
+    const size = 4;
+    const format = 'rgba8unorm';
+
+    const texture = t.device.createTexture({
+      size: { width: size, height: size, depth: 1 },
+      format,
+      usage: GPUTextureUsage.OUTPUT_ATTACHMENT | GPUTextureUsage.COPY_SRC,
+    });
+
+    const depthTexture = t.params.depthStencilFormat
+      ? t.device.createTexture({
+          size: { width: size, height: size, depth: 1 },
+          format: t.params.depthStencilFormat,
+          usage: GPUTextureUsage.OUTPUT_ATTACHMENT,
+        })
+      : null;
+
+    const encoder = t.device.createCommandEncoder();
+    const pass = encoder.beginRenderPass({
+      colorAttachments: [
+        {
+          attachment: texture.createView(),
+          loadValue: { r: 0.0, g: 0.0, b: 1.0, a: 1.0 },
+        },
+      ],
+      depthStencilAttachment: depthTexture
+        ? {
+            attachment: depthTexture.createView(),
+            depthLoadValue: 1.0,
+            depthStoreOp: 'store',
+            stencilLoadValue: 0,
+            stencilStoreOp: 'store',
+          }
+        : undefined,
+    });
+
+    // Draw two triangles with different winding orders:
+    // 1. The top-left one is counterclockwise (CCW)
+    // 2. The bottom-right one is clockwise (CW)
+    const vertexModule = t.makeShaderModule('vertex', {
+      glsl: `#version 450
+            const vec2 pos[6] = vec2[6](vec2(-1.0f,  1.0f),
+                                        vec2(-1.0f,  0.0f),
+                                        vec2( 0.0f,  1.0f),
+                                        vec2( 0.0f, -1.0f),
+                                        vec2( 1.0f,  0.0f),
+                                        vec2( 1.0f, -1.0f));
+            void main() {
+                gl_Position = vec4(pos[gl_VertexIndex], 0.0, 1.0);
+            }`,
+    });
+
+    const fragmentModule = t.makeShaderModule('fragment', {
+      glsl: `#version 450
+      layout(location = 0) out vec4 fragColor;
+      void main() {
+        if (gl_FrontFacing) {
+          fragColor = vec4(0.0, 1.0, 0.0, 1.0);
+        } else {
+          fragColor = vec4(1.0, 0.0, 0.0, 1.0);
+        }
+      }`,
+    });
+
+    pass.setPipeline(
+      t.device.createRenderPipeline({
+        vertexStage: { module: vertexModule, entryPoint: 'main' },
+        fragmentStage: { module: fragmentModule, entryPoint: 'main' },
+        primitiveTopology: t.params.primitiveTopology,
+        rasterizationState: {
+          frontFace: t.params.frontFace,
+          cullMode: t.params.cullMode,
+        },
+        colorStates: [{ format }],
+        depthStencilState: depthTexture
+          ? { format: t.params.depthStencilFormat as GPUTextureFormat }
+          : undefined,
+      })
+    );
+
+    pass.draw(6, 1, 0, 0);
+    pass.endPass();
+
+    t.device.defaultQueue.submit([encoder.finish()]);
+
+    // front facing color is green, non front facing is red, background is blue
+    const kCCWTriangleTopLeftColor = faceColor('ccw', t.params.frontFace, t.params.cullMode);
+    t.expectSinglePixelIn2DTexture(
+      texture,
+      format,
+      { x: 0, y: 0 },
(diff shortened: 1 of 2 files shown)


natashalee:
@kainino0x @austinEng not sure why I am not able to add reviewers, but could you take a look for me? Thank you!

src/webgpu/api/operation/command_buffer/render/culling_tests.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+      ? new Uint8Array([0x00, 0x00, 0xff, 0xff])
+      : new Uint8Array([0x03, 0x03, 0x00, 0xff]);
+    this.expectContents(bottomRightBuffer, kCWTriangleBottomRightColor);
+  }
+}
+
+export const g = makeTestGroup(CullingTest);
+
+g.test('CullNoneWhenCCWIsFrontFace')
+  .params([{ format: 'rgba8unorm' as GPUTextureFormat }])
```
kainino0x, line 127: I think most  different formats are unlikely to have much of an effect in these tests. We should leave it out if we only test one variation, though we should probably have some more combinatorics here - or at least TODOs to add them. Here are some ideas (that would go into a test plan if we had one):

test culling is performed and correctly written into all attachments, with following parameters:
- x= actual face orientation [cw, ccw] (we have this above)
- x= all possible frontFace (have this below)
- x= all possible cullMode (have this below)
- x= DS attachment = [none, depth, stencil, depth+stencil]
- x= color attachments = [0, 1, >1]

I haven't thought of other combinatorics that are likely to interact with culling, but there might be more. We want to be as thorough as is reasonable.

src/webgpu/api/operation/command_buffer/render/culling_tests.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+                                        vec2( 1.0f,  0.0f),
+                                        vec2( 1.0f, -1.0f));
+            void main() {
+                gl_Position = vec4(pos[gl_VertexIndex], 0.0, 1.0);
+            }`,
+    });
+
+    // gl_FragCoord of pixel(x, y) in framebuffer coordinate is (x + 0.5, y + 0.5). And we use
+    // RGBA8 format for the back buffer. So (gl_FragCoord.xy - vec2(0.5)) / 255 in shader code
+    // will make the pixel's R and G channels exactly equal to the pixel's x and y coordinates.
```
kainino0x, line 30: Does the FragCoord matter? - can we just always render green or something instead?

src/webgpu/api/operation/command_buffer/render/culling_tests.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+      rasterizationState: { frontFace, cullMode },
+      colorStates: [
+        {
+          format,
+          alphaBlend: { dstFactor: 'zero', srcFactor: 'one', operation: 'add' },
+          colorBlend: { dstFactor: 'zero', srcFactor: 'one', operation: 'add' },
+          writeMask: GPUColorWrite.ALL,
+        },
+      ],
+      sampleCount: 1,
```
kainino0x, line 52: These are all defaults, don't need to specify them
```suggestion
      colorStates: [ { format } ],
```

src/webgpu/api/operation/command_buffer/render/culling_tests.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+  ): void {
+    const size = 4;
+    const texture = this.create2DTextureForTest(format, size);
+    const encoder = this.device.createCommandEncoder();
+    const pass = encoder.beginRenderPass({
+      colorAttachments: [
+        {
+          attachment: texture.createView(),
+          loadValue: { r: 0.0, g: 0.0, b: 1.0, a: 1.0 },
+          storeOp: 'store',
```
kainino0x, line 82: 'store' should be the default.
```suggestion
```

src/webgpu/api/operation/command_buffer/render/culling_tests.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+        },
+      ],
+    });
+    pass.setPipeline(this.createPipelineForTest(frontFace, cullMode, format));
+    pass.draw(6, 1, 0, 0);
+    pass.endPass();
+
+    const topLeftBuffer = this.device.createBuffer({
```
kainino0x, line 90: Let's make a helper function (can be in gpu_test.ts) to expect the color of a single pixel in a texture.

It might be easy to implement this on top of the existing expectSingleColor, but I'm not sure whether it can check the contents of a specified subrectangle of a texture (seems like not right now...)

Eventually we're also going to need something to check the values of depth/stencil (which probably will be done by sampling the texture in a shader?) but this can wait a bit. (Leave comments for any TODOs we think of, though.)

src/webgpu/api/operation/command_buffer/render/culling_tests.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+          alphaBlend: { dstFactor: 'zero', srcFactor: 'one', operation: 'add' },
+          colorBlend: { dstFactor: 'zero', srcFactor: 'one', operation: 'add' },
+          writeMask: GPUColorWrite.ALL,
+        },
+      ],
+      sampleCount: 1,
+    });
+  }
+
+  create2DTextureForTest(format: GPUTextureFormat, size: number) {
```
kainino0x, line 56: This is only used once, so I think it should just be inlined.

src/webgpu/api/operation/command_buffer/render/culling_tests.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+    return this.device.createTexture({
+      size: { width: size, height: size, depth: 1 },
+      format,
+      mipLevelCount: 1,
+      sampleCount: 1,
+      dimension: '2d',
```
kainino0x, line 62: These are defaults and can be omitted

```suggestion
```

src/webgpu/api/operation/command_buffer/render/culling_tests.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+  .params([{ format: 'rgba8unorm' as GPUTextureFormat }])
+  .fn(t => {
+    t.run('cw', 'front', false, true, t.params.format);
+  });
+
+g.test('CullBackFaceWhenCWIsFrontFace')
+  .params([{ format: 'rgba8unorm' as GPUTextureFormat }])
+  .fn(t => {
+    t.run('cw', 'back', true, false, t.params.format);
+  });
```
kainino0x, line 160: We can parameterize over these different cases by doing something like this (totally untested):

```typescript
// use this in t.run() to compute the expectations:
function faceIsCulled(face: 'cw' | 'ccw', frontFace: GPUFrontFace, cullMode: GPUCullMode): boolean {
  return cullMode !== 'none' && ((frontFace === face) === (cullMode === 'front'));
}

g.test('cull_state')
  .params(
    params()
      .combine(poptions('frontFace', ['ccw', 'cw']))
      .combine(poptions('cullMode', ['none', 'front', 'back']))
  )
  .fn(t => {
    t.run(t.params.frontFace, t.params.cullMode);
  });
```

alternatively:

```typescript
g.test('cull_state')
  .params([
    { cullMode: 'none', frontFace: 'cw', _expectCCWCulled: false, _expectCWCulled: false },
    // ...
  ] as const)
  .fn(t => {
    t.run(t.params.frontFace, t.params.cullMode, t.params._expectCCWCulled, t.params._expectCWCulled);
  });
```

Note with either change, there would now be only one usage of t.run(), so if you don't end up with more, then you can just move its contents inline here.

kainino0x:
oh, one more nit, I'd call this a render pipeline test rather than command buffer test, so maybe `src/webgpu/api/operation/render_pipeline/culling.spec.ts`.

src/webgpu/api/operation/command_buffer/render/culling_tests.spec.ts
Kangz: It would be nice to test with triangle list vs. triangle strip too because there's some complexity in determining the front face of the triangles in a strip. We should also have a similar test (or even the same one) for the gl_FrontFace builtin.

src/webgpu/api/operation/render_pipeline/culling_tests.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    const kCWTriangleBottomRightColor = faceIsCulled('cw', t.params.frontFace, t.params.cullMode)
+      ? new Uint8Array([0x00, 0x00, 0xff, 0xff])
+      : new Uint8Array([0xff, 0x00, 0x00, 0xff]);
+    t.expectSinglePixelIn2DTexture(
+      texture,
+      format,
+      { x: size - 1, y: size - 1 },
+      { exp: kCWTriangleBottomRightColor }
+    );
```
kainino0x, line 133: TODO: check the contents of the depth and stencil outputs

src/webgpu/api/operation/render_pipeline/culling_tests.spec.ts
```diff
@@ -0,0 +1,134 @@
+export const description = 'Test culling and rasterizaion state.';
```
Kangz, line 1: Can you add the test plan in the description similarly to https://github.com/gpuweb/cts/pull/223/files ?

src/webgpu/api/operation/render_pipeline/culling_tests.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+        poptions('depthStencilFormat', [
+          null,
+          'depth24plus',
+          'depth32float',
+          'depth24plus-stencil8',
+        ] as GPUTextureFormat[])
+      )
+      .combine(poptions('depthBias', [0, 5]))
+      .combine(poptions('depthBiasSlopeScale', [0, 5]))
+      .combine(poptions('depthBiasClamp', [0, 5]))
```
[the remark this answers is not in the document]
Kangz: i think these parameters are independent from the culling and should be tested separately.

src/webgpu/api/operation/render_pipeline/culling_tests.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+      void main() {
+        fragColor = vec4(1.0, 0.0, 0.0, 1.0);
+      }`,
+    });
+
+    pass.setPipeline(
+      t.device.createRenderPipeline({
+        vertexStage: { module: vertexModule, entryPoint: 'main' },
+        fragmentStage: { module: fragmentModule, entryPoint: 'main' },
+        // TODO: test triangle-strip as well
```
Kangz, line 95: Can the primitive topology be a parameter as well, even if strips are left as TODO?

src/webgpu/api/operation/render_pipeline/culling_tests.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+        primitiveTopology: 'triangle-list',
+        rasterizationState: {
+          frontFace: t.params.frontFace,
+          cullMode: t.params.cullMode,
+          depthBias: t.params.depthBias,
```
Kangz, line 100: All the depthBias stuff can be left out.

src/webgpu/api/operation/render_pipeline/culling_tests.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+      void main() {
+        fragColor = vec4(1.0, 0.0, 0.0, 1.0);
+      }`,
+    });
+
+    pass.setPipeline(
+      t.device.createRenderPipeline({
+        vertexStage: { module: vertexModule, entryPoint: 'main' },
+        fragmentStage: { module: fragmentModule, entryPoint: 'main' },
+        // TODO: test triangle-strip as well
```
kainino0x, line 111: nit: move this up to the `.combine` line
```suggestion
```

src/webgpu/api/operation/render_pipeline/culling_tests.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+      .combine(poptions('cullMode', ['none', 'front', 'back'] as const))
+      .combine(
+        poptions('depthStencilFormat', [
+          null,
+          'depth24plus',
+          'depth32float',
+          'depth24plus-stencil8',
+        ] as const)
+      )
+      .combine(poptions('primitiveTopology', ['triangle-list'] as const))
```
kainino0x, line 44:
```suggestion
      // TODO: test triangle-strip as well
      .combine(poptions('primitiveTopology', ['triangle-list'] as const))
```

src/webgpu/api/operation/render_pipeline/culling_tests.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+            void main() {
+                gl_Position = vec4(pos[gl_VertexIndex], 0.0, 1.0);
+            }`,
+    });
+
+    const fragmentModule = t.makeShaderModule('fragment', {
+      glsl: `#version 450
+      layout(location = 0) out vec4 fragColor;
+      void main() {
+        fragColor = vec4(1.0, 0.0, 0.0, 1.0);
```
kainino0x, line 103: > We should also have a similar test (or even the same one) for the gl_FrontFace builtin.

We should be able to do this easily now, just make this output red or green depending on the gl_FrontFace value.

kainino0x:
Thanks!

PR #224, opened 2020-07-01
Add a test for storage buffer and readonly storage buffer

Richard-Yunchao:
The first patch to webgpu cts project. PTAL. Thank you!

We have already had a similar test in Dawn validation test.

### DIFF
--- a/src/webgpu/api/validation/createBindGroupLayout.spec.ts
+++ b/src/webgpu/api/validation/createBindGroupLayout.spec.ts
@@ -2,13 +2,14 @@ export const description = `
 createBindGroupLayout validation tests.
 `;
 
-import { poptions, params } from '../../../common/framework/params_builder.js';
+import { pbool, poptions, params } from '../../../common/framework/params_builder.js';
 import { makeTestGroup } from '../../../common/framework/test_group.js';
 import {
   kBindingTypeInfo,
   kBindingTypes,
   kMaxBindingsPerBindGroup,
   kShaderStages,
+  kShaderStageCombinations,
 } from '../../capability_info.js';
 
 import { ValidationTest } from './validation_test.js';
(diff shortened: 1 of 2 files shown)


src/webgpu/api/validation/createBindGroupLayout.spec.ts
```diff
@@ -45,6 +45,32 @@ g.test('visibility_of_bindings_can_be_0').fn(async t => {
   });
 });
```
kainino0x, line 46: suggestion below will probably obsolete this test

src/webgpu/api/validation/createBindGroupLayout.spec.ts
```diff
@@ -45,6 +45,32 @@ g.test('visibility_of_bindings_can_be_0').fn(async t => {
...
 
+g.test('storage_buffer_binding_support').fn(async t => {
```
kainino0x, line 48: Let's try to parameterize this, something like the following:

```typescript
g.test('buffer_binding_valid_stages')
    params()
      .combine(poptions('type', kBufferBindingTypes))
      .combine(poptions('visibility', kShaderStageCombinations))
  )
  .fn(async t => {
    const { binding, visibility, type } = t.params;
    const typeInfo = kBufferBindingTypeInfo[type];

    const visibleOnlyToValidStages = (visibility & typeInfo.validStages) === visibility;

    const shouldError = !visibleOnlyToValidStages;
      t.device.createBindGroupLayout({
        entries: [{ binding: 0, visibility, type }],
      });
    }, shouldError);
  });
```

kainino0x: Actually, while we're at it, we should make this test all binding types, not just buffer binding types, because nothing about this test is specific to buffer binding types.

That said, I just realized this has overlap with the test createPipelineLayout:visibility_and_dynamic_offsets, which, despite being in createPipelineLayout.spec.ts I'm pretty sure is actually testing createBindGroupLayout validation. We should just move it and tweak it slightly to look more like what I just wrote above, instead of using what I wrote above.

kainino0x: Note that visibility_and_dynamic_offsets is actually written incorrectly right now (oops) - it wraps both createPipelineLayout and createBindGroupLayout in expectValidationError, so it's not correctly checking which of those two functions the error is in. One of the tweaks will be to fix that.

Richard-Yunchao: @kainino0x , thanks for your review! You are right, visibility_and_dynamic_offsets actually tests createBindGroupLayout, not createPipelineLayout. And it can cover my previous code. So I just moved that test to createBindGroupLayout and removed my previous code. :). PTAL.

src/webgpu/api/validation/createBindGroupLayout.spec.ts
kainino0x: Can you double check that this is obsolete and if so remove it?

Richard-Yunchao:
Yeah. It is redundant. I removed that test. @kainino0x

kainino0x:
Thanks!

Issue #225, opened 2020-07-02
Implement RenderPass StoreOp Tests


bjjones:
Coverage needed for testing render pass store operations

**/api/validation/**
- [ ] Test that when `depthReadOnly` is `true`, `depthStoreOp` must be `store`
- [ ] Test that when `stencilReadOnly` is `true`, `stencilStoreOp` must be `store`

**/api/operation/**
- [ ] Test that a render pass with colorAttachment `storeOp` set to `clear` has correct output for:
    - [ ] All color renderable formats
    - [ ] All color attachments, from 0 to `MAX_COLOR_ATTACHMENTS`
 
- [ ] Test that a render pass with colorAttachment `storeOp` set to `store` has correct output for:
    - [ ] All color renderable formats
    - [ ] All color attachments, from 0 to `MAX_COLOR_ATTACHMENTS`

- [ ] Test that a render pass with `depthStoreOp` set to `clear` has correct output for:
    - [ ] All depth/stencil formats
- [ ] Test that a render pass with `depthStoreOp` set to `store` has correct output for:
    - [ ] All depth/stencil formats
- [ ] Test that a render pass with `stencilStoreOp` set to `clear` has correct output for:
    - [ ] All depth/stencil formats with stencil component.
- [ ] Test that a render pass with `stencilStoreOp` set to `store` has correct output for:
    - [ ] All depth/stencil formats with stencil component.

**/idl/**
- [ ] Test that a color attachment's `storeOp` is `store` by default
- [ ] Test that depth/stencil attachment's `depthStoreOp` is `store` by default
- [ ] Test that depth/stencil attachment's `stencilStoreOp` is `store` by default

bjjones:
It was mentioned I should write a test plan for the tests I was assigned - so here it is. Feel free to edit the initial comment if I'm missing something.

Kangz:
Thanks for putting this up!

For validation I think we might want to validate that if the `depthStoreOp` or `depthReadOnly` is set, then the attachment must have a depth aspect (and such for all depth-stencil formats). Same thing for `stencilStoreOp` and `stencilReadOnly`.

For operation we might want to test rendering of a bunch of values, but maybe it's not part of the storeOp tests, but for general pixel output?

Also we should test all combinations of depth/stencil ops together for all formats since they are two planes of the same texture. IDK if the checks for default values should be in IDL or operation though. @kainino0x, do you have an opinion?

[...]

PR #227, opened 2020-07-06
Add Initial Coverage TODOs for Render Pass StoreOP

bjjones:
From the feedback in #225.

### DIFF
--- a/src/webgpu/api/operation/render_pass/storeOp.spec.ts
+++ b/src/webgpu/api/operation/render_pass/storeOp.spec.ts
@@ -1,0 +1,23 @@
+export const description = `API Operation Tests for RenderPass StoreOp.
+
+  Test Coverage Needed:
+
+  - Test that a render pass has correct output for combinations of:
+    - All color attachments from '0' to 'MAX_COLOR_ATTACHMENTS' with combinations of:
+      - storeOp set to {'clear', 'store', 'undefined}
+      - All color renderable formats
+      - mip level set to {'0', mip > '0'}
+      - array layer set to {'0', layer > '1'} for 2D textures
+      - depth slice set to {'0', slice > '0'} for 3D textures
+    - With and without a depthStencilAttachment that has the combinations of:
+      - depthStoreOp set to {'clear', 'store', 'undefined'}
+      - stencilStoreOp set to {'clear', 'store', 'undefined'}
+      - All depth/stencil formats
+      - mip level set to {'0', mip > '0'}
+      - array layer set to {'0', layer > '1'} for 2D textures
+      - depth slice set to {'0', slice > '0'} for 3D textures`;
+
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { GPUTest } from '../../../gpu_test.js';
+
+export const g = makeTestGroup(GPUTest);
--- a/src/webgpu/api/validation/render_pass/storeOp.spec.ts
+++ b/src/webgpu/api/validation/render_pass/storeOp.spec.ts
@@ -1,0 +1,13 @@
+export const description = `API Validation Tests for RenderPass StoreOp.
+
+  Test Coverage Needed:
+
+  - Test that when depthReadOnly is true, depthStoreOp must be 'store'
+
+  - Test that when stencilReadOnly is true, stencilStoreOp must be 'store'`;
+
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+
+import { ValidationTest } from './../validation_test.js';
+
+export const g = makeTestGroup(ValidationTest);


src/webgpu/api/operation/render_pass/storeOp.spec.ts
```diff
@@ -0,0 +1,34 @@
+export const description = `API Operation Tests for RenderPass StoreOp.
```
austinEng, line 1: nit: we should use a different variable since this `description` is put into the HTML page when load the tests.

https://github.com/gpuweb/cts/blob/40e337a38784ad72fa5c7b9afd1b9c358a9e0f1a/src/common/runtime/standalone.ts#L186-L191

Not sure if Kai wants this to be an exported variable or just a comment, but I think the variable is nice for potential future tooling.

src/webgpu/api/validation/render_pass/storeOp.spec.ts
```diff
@@ -0,0 +1,16 @@
...
+
+  Test Coverage Needed:
+
+  - Test that when depthReadOnly is true, depthStoreOp must be "store".
+
+  - Test that when stencilReadOnly is true, stencilStoreOp must be "store".
+
+  - Test that when depthStoreOp or depthReadOnly are set, the attachment must have a depth aspect.
+
+  - Test that when stencilStoreOp or stencilReadOnly are set, the attachment must have a stencil aspect.`;
```
kainino0x, line 11: depthStoreOp and stencilStoreOp are actually required, which I guess means they're ignored if the format doesn't have the necessary aspect.

Similarly, depthReadOnly and stencilReadOnly have defaults, so in a sense they're always "set". If the store ops are ignored I'd ignore these too for consistency.

cc @kvark

src/webgpu/idl/render_pass/storeOp.spec.ts
```diff
@@ -0,0 +1,14 @@
...
+  - Test that a color attachment's storeOp is store by default.
+
+  - Test that depth/stencil attachment's depthStoreOp is store by default.
+
+  - Test that depth/stencil attachment's stencilStoreOp is store by default.`;
+
+import { makeTestGroup } from '../../../common/framework/test_group.js';
+import { IDLTest } from '../idl_test.js';
+
+export const g = makeTestGroup(IDLTest);
```
kainino0x, line 14: FYI there's an issue right now where IDLTests are not GPUTests so you can't actually test creating anything. I was meaning to fix this, but anyone can fix it anytime.

That said, the IDL tests are less important (they're a lot less nuanced than the validation tests, and already have a standardized language backing them so the implementations should be mostly consistent by default), so okay to put these tests off. I was also considering whether we could partially automate these to some extent, so probably at some point someone should sit down and focus on just the IDL tests and figure out the right way to do them.

src/webgpu/api/operation/render_pass/storeOp.spec.ts
kainino0x: I was thinking about this earlier; maybe we should just leave it in the description. We could just set up the standalone runner so it hides the full description by default. Having it in the description means the standalone runner can basically be a test plan viewer with no changes.

src/webgpu/api/operation/render_pass/storeOp.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  - Test that a render pass with color attachment storeOp set to 'clear' has correct output for:
+      - All color renderable formats
+      - All color attachments from '0' to 'MAX_COLOR_ATTACHMENTS'
+      - Zero/Non-zero Mip.
+
+  - Test that a render pass with color attachment storeOp set to 'store' has correct output for:
+      - All color renderable formats
+      - All color attachments from '0' to 'MAX_COLOR_ATTACHMENTS'
+      - Zero/Non-zero Mip.
```
kainino0x, line 13: optional: maybe for readability, could dedup parts of this. Also can use the conventions I'm drafting (linked from test plan I'm drafting and haven't published: https://hackmd.io/@kainino0x/H1MwoqqAU)

```
  - Test that a render pass with color attachment storeOp has correct output for combinations of:
      - storeOp set to {'clear', 'store'}
      - All color renderable formats
      - All color attachments from '0' to 'MAX_COLOR_ATTACHMENTS'
      - Zero/Non-zero Mip.
```

and similar below

kainino0x: I already forgot where I wrote this, but yeah, let's test the default values in operation tests. We can just add "undefined" to the list of options to test which should make most default values very easy to test.

kainino0x: Added in guidelines: https://hackmd.io/@kainino0x/SJtUks9CU

src/webgpu/idl/render_pass/storeOp.spec.ts
```diff
@@ -0,0 +1,14 @@
+export const description = `IDL Tests for RenderPass StoreOp.
+
+  Test Coverage Needed:
+
+  - Test that a color attachment's storeOp is 'store' by default.
+
+  - Test that depthReadOnly is 'false' by default.
+
+  - Test that stencilReadOnly is 'false' by default.`;
```
kainino0x, line 9: Will move to operation tests with https://github.com/gpuweb/cts/pull/227#discussion_r450854447

bjjones: Done. Removed the entire/idl/ file.

src/webgpu/api/operation/render_pass/storeOp.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  - Test that a render pass with depthStoreOp set to clear has correct output for:
```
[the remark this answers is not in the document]
bjjones: Done.

src/webgpu/api/operation/render_pass/storeOp.spec.ts
```diff
@@ -0,0 +1,34 @@
+export const description = `API Operation Tests for RenderPass StoreOp.
+
+  Test Coverage Needed:
+
+  - Test that a render pass with color attachment storeOp set to 'clear' has correct output for:
+      - All color renderable formats
+      - All color attachments from '0' to 'MAX_COLOR_ATTACHMENTS'
+      - Zero/Non-zero Mip.
```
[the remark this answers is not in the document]
bjjones: Done.

src/webgpu/idl/render_pass/storeOp.spec.ts
bjjones: IDL tests were moved into /operation/, so not going to worry about this.

src/webgpu/api/operation/render_pass/storeOp.spec.ts
bjjones: Kept it as description. Resolving.

src/webgpu/api/operation/render_pass/storeOp.spec.ts
bjjones: Done.

src/webgpu/api/validation/render_pass/storeOp.spec.ts
bjjones: Removed these.

src/webgpu/api/operation/render_pass/storeOp.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+      - depth slice set to {'0', slice > '0'}
+
+  - Test that a render pass with a depthStencilAttachment has correct output for combinations of:
+      - depthStoreOp set to {'clear', 'store', 'undefined'}
+      - stencilStoreOp set to {'clear', 'store', 'undefined'}
+      - All depth/stencil formats
+      - mip level set to {'0', mip > '0'}
+      - array layer set to {'0', layer > '0'}
+      - depth slice set to {'0', slice > '0'}`;
+
```
austinEng, line 20: also test that using color attachments and depthStencil attachments together work correctly,

and test using color attachments with different store ops as they should clear/store independently

bjjones: Done. I think this has now might be best done in a single test with a giant param list.

src/webgpu/api/operation/render_pass/storeOp.spec.ts
```diff
@@ -0,0 +1,23 @@
+export const description = `API Operation Tests for RenderPass StoreOp.
+
+  Test Coverage Needed:
+
+  - Test that a render pass has correct output for combinations of:
+    - All color attachments from '0' to 'MAX_COLOR_ATTACHMENTS' with combinations of:
+      - storeOp set to {'clear', 'store', 'undefined}
+      - All color renderable formats
+      - mip level set to {'0', mip > '0'}
+      - array layer set to {'1', layer > '1'}
```
Kangz, line 10: nit: array layer set to 0 and >1

src/webgpu/api/operation/render_pass/storeOp.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+      - depth slice set to {'0', slice > '0'}
+    - With and without a depthStencilAttachment that has the combinations of:
+      - depthStoreOp set to {'clear', 'store', 'undefined'}
+      - stencilStoreOp set to {'clear', 'store', 'undefined'}
+      - All depth/stencil formats
+      - mip level set to {'0', mip > '0'}
+      - array layer set to {'0', layer > '0'}
```
Kangz, line 17: ditto

Kangz:
Looks pretty good!

src/webgpu/api/operation/render_pass/storeOp.spec.ts
bjjones: >  array layer set to 0 and >1 for 2D textures

Do you mean array layer set to 0 and **>0** for 2D textures? where array layer means `GPUTextureView.baseArrayLayer = {0, > 0}` and `GPUTextureView.arrayLayerCount = 1`

Kangz: This was for the GPUTextureView baseArrayLyer. Of course if it is >0, then the texture needs to have depth > 1 (arrayLayerCount was folded in depth recently)

bjjones: Okay, done.

bjjones:
Can I merge now? Not sure if I need approval from all, or 2 is good enough.

kainino0x:
2 is fine, and LGTM; sorry for the delay

PR #228, opened 2020-07-06
Add initial coverage for robust access vertex tests

iraiter2:
Planning to write these tests in glsl first, but should be easy to convert to wgsl since the shaders will be built with templates.

### DIFF
--- a/src/webgpu/shader/execution/robust_access_vertex.spec.ts
+++ b/src/webgpu/shader/execution/robust_access_vertex.spec.ts
@@ -29,9 +29,20 @@ instanceCount / firstInstance are tested.
 
 The tests will include multiple attributes per vertex buffer.
 
+The vertex buffers will be filled by repeating a few chosen values until the end of the buffer.
+
 The test will run a render pipeline which verifies the following:
-1) All vertex attribute values are in-bounds or zero (buffers filled with a few random values)
-2) All gl_VertexIndex values are within the index buffer or 0`;
+1) All vertex attribute values occur in the buffer or are zero
+2) All gl_VertexIndex values are within the index buffer or 0
+
+TODO:
+
+A suppression may be needed for d3d12 on tests that have non-zero baseVertex, since d3d12 counts
+from 0 instead of from baseVertex (will fail check for gl_VertexIndex).
+
+Vertex buffer contents could be randomized to prevent the case where a previous test creates
+a similar buffer to ours and the OOB-read seems valid. This should be deterministic, which adds
+more complexity that we may not need.`;
 
 import { makeTestGroup } from '../../../common/framework/test_group.js';
 import { GPUTest } from '../../gpu_test.js';
--- a/src/webgpu/shader/execution/robust_access_vertex.spec.ts
+++ b/src/webgpu/shader/execution/robust_access_vertex.spec.ts
@@ -1,0 +1,40 @@
+export const description = `
+Test vertex attributes behave correctly (no crash / data leak) when accessed out of bounds
+`;
+
+/*
+Test coverage:
+
+The following will be parameterized (all combinations tested):
+
+1) Draw call indexed? (false / true)
+  - Run the draw call using an index buffer
+
+2) Draw call indirect? (false / true)
+  - Run the draw call using an indirect buffer
+
+3) Draw call parameter (vertexCount, firstVertex, indexCount, firstIndex, baseVertex, instanceCount,
+  firstInstance)
+  - The parameter which will go out of bounds. Filtered depending on if the draw call is indexed.
+
+4) Attribute type (float, vec2, vec3, vec4, mat2, mat3, mat4)
+  - The input attribute type in the vertex shader
+
+5) Error scale (1, 100, 100000000)
+  - Offset to add to the correct draw call parameter
+
+6) Additional vertex buffer (false / true)
+  - Tests that no OOB occurs if another vertex buffer is bound
+
+The tests will also have another vertex buffer bound for an instanced attribute, to make sure
+instanceCount / firstInstance are tested.
+
+The test will run a render pipeline which verifies the following:
+1) All vertex attribute values are in-bounds (equal to zero, since buffers will be zero'd)
+2) All gl_VertexIndex values are within the index buffer or 0
+*/
+
+import { makeTestGroup } from '../../../common/framework/test_group.js';
+import { GPUTest } from '../../gpu_test.js';
+
+export const g = makeTestGroup(GPUTest);
--- a/src/webgpu/shader/execution/robust_access_vertex.spec.ts
+++ b/src/webgpu/shader/execution/robust_access_vertex.spec.ts
@@ -1,8 +1,6 @@
 export const description = `
 Test vertex attributes behave correctly (no crash / data leak) when accessed out of bounds
-`;
 
-/*
 Test coverage:
 
 The following will be parameterized (all combinations tested):
@@ -20,19 +18,20 @@ The following will be parameterized (all combinations tested):
 4) Attribute type (float, vec2, vec3, vec4, mat2, mat3, mat4)
   - The input attribute type in the vertex shader
 
-5) Error scale (1, 100, 100000000)
+5) Error scale (1, 4, 16, 64, 256, 1000000)
   - Offset to add to the correct draw call parameter
 
-6) Additional vertex buffer (false / true)
-  - Tests that no OOB occurs if another vertex buffer is bound
+6) Additional vertex buffers (+2, +4)
+  - Tests that no OOB occurs if more vertex buffers are used
 
 The tests will also have another vertex buffer bound for an instanced attribute, to make sure
 instanceCount / firstInstance are tested.
 
+The tests will include multiple attributes per vertex buffer.
+
 The test will run a render pipeline which verifies the following:
 1) All vertex attribute values are in-bounds (equal to zero, since buffers will be zero'd)
-2) All gl_VertexIndex values are within the index buffer or 0
-*/
+2) All gl_VertexIndex values are within the index buffer or 0`;
 
 import { makeTestGroup } from '../../../common/framework/test_group.js';
 import { GPUTest } from '../../gpu_test.js';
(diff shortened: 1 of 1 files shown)


src/webgpu/shader/execution/robust_access_vertex.spec.ts
```diff
@@ -0,0 +1,40 @@
+export const description = `
+Test vertex attributes behave correctly (no crash / data leak) when accessed out of bounds
+`;
+
+/*
```
kainino0x, line 5: I think let's start out putting the test plan inside the description string like in #227. We can change it later if we decide it's unwieldy. Format doesn't need to be consistent (but could be consistent with how I've been writing, if you think that's a good format https://hackmd.io/@kainino0x/H1MwoqqAU)

src/webgpu/shader/execution/robust_access_vertex.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+
+1) Draw call indexed? (false / true)
+  - Run the draw call using an index buffer
+
+2) Draw call indirect? (false / true)
+  - Run the draw call using an indirect buffer
+
+3) Draw call parameter (vertexCount, firstVertex, indexCount, firstIndex, baseVertex, instanceCount,
+  firstInstance)
+  - The parameter which will go out of bounds. Filtered depending on if the draw call is indexed.
```
kainino0x, line 18: tip: perhaps nice to write this as

```typescript
.combine(pbool('indexed'))
.expand(function* (p) {
  if (p.indexed) {
    yield* ['indexCount', 'firstIndex', 'baseVertex'];
  } else {
    yield* ['vertexCount', 'firstVertex'];
  }
  yield* ['instanceCount', 'firstInstance'];
})
```

src/webgpu/shader/execution/robust_access_vertex.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+
+4) Attribute type (float, vec2, vec3, vec4, mat2, mat3, mat4)
+  - The input attribute type in the vertex shader
+
+5) Error scale (1, 100, 100000000)
+  - Offset to add to the correct draw call parameter
```
Kangz, line 24: Can we have a couple more granular steps just to make sure one of them successfully crashes the GPU / reads OOB if the implementation doesn't implement RBA correctly.

src/webgpu/shader/execution/robust_access_vertex.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+
+6) Additional vertex buffer (false / true)
+  - Tests that no OOB occurs if another vertex buffer is bound
+
+The tests will also have another vertex buffer bound for an instanced attribute, to make sure
+instanceCount / firstInstance are tested.
+
+The test will run a render pipeline which verifies the following:
+1) All vertex attribute values are in-bounds (equal to zero, since buffers will be zero'd)
```
Kangz, line 33: Can the test check that values sampled in bounds have the correct value (non zero, they should be pretty specific values, maybe randomly generated? @kainino0x maybe we should have repeatable random generators?), and that the value sampled OOB has one of the allowed values?

src/webgpu/shader/execution/robust_access_vertex.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+  firstInstance)
+  - The parameter which will go out of bounds. Filtered depending on if the draw call is indexed.
+
+4) Attribute type (float, vec2, vec3, vec4, mat2, mat3, mat4)
+  - The input attribute type in the vertex shader
+
+5) Error scale (1, 100, 100000000)
+  - Offset to add to the correct draw call parameter
+
+6) Additional vertex buffer (false / true)
```
Kangz, line 26: Maybe we could try varying the number of vertex buffers, or the one that goes OOB to make sure programmable vertex pulling is implemented correctly for multiple vertex buffers?

src/webgpu/shader/execution/robust_access_vertex.spec.ts
kainino0x: We probably should have some simple repeatable random generators, @iraiter2 feel free to add some in a common place (probably a new file in framework/)

austinEng: Can't values be equal to zero, or some other value inside the buffer?

src/webgpu/shader/execution/robust_access_vertex.spec.ts
austinEng: Also consider one or multiple vertex attributes per vertex buffer since if the implementation does vertex pulling it'll need to load more than one attribute from a single buffer.

src/webgpu/shader/execution/robust_access_vertex.spec.ts
iraiter2: Nice. My first attempt was using two enums, something like NonIndexedDrawCallParameter and IndexedDrawCallParameter, and then doing:
```typescript
.combine(pbool('indexed'))
.combine(poptions('drawCallTestParameter', kDrawCallParameters))
.filter(({ indexed, drawCallTestParameter }) =>
    indexed && drawCallTestParameter in IndexedDrawCallParameter ||
    !indexed && drawCallTestParameter in NonIndexedDrawCallParameter)
```

Do you prefer one or the other?

src/webgpu/shader/execution/robust_access_vertex.spec.ts
iraiter2: I'm thinking if the values can be either zero or a value in the buffer, filling the buffer with random info doesn't detect anything more? With a buffer full of zeros, if the value we see is anything non-zero, we can be sure it didn't come from inside the buffer. Is there a case where we would detect an invalid RBA implementation if we filled the buffer with data compared to if it was full of zeros?

src/webgpu/shader/execution/robust_access_vertex.spec.ts
iraiter2: Moved it in, seems like the hackmd link is down?

src/webgpu/shader/execution/robust_access_vertex.spec.ts
iraiter2: Added some more, do you think it should be more/less granular than this new list?

src/webgpu/shader/execution/robust_access_vertex.spec.ts
iraiter2: Added some text to say we'll check varied number of vertex buffers (+2 / +4 more buffers) and that the test will bind multiple attributes to a buffer.

src/webgpu/shader/execution/robust_access_vertex.spec.ts
Kangz: Oh, I was thinking of larger steps, like 1, 4, 10^2, 10^4, 10^6 10^9 WDYT?

src/webgpu/shader/execution/robust_access_vertex.spec.ts
Kangz: Ah, I forgot about the exact RBA rules and that it allows things like % the index. I still think it's useful to have non-zero data in the vertex buffer (maybe just 1, 2, 3, 4, 5, ...) for two reasons:

 - We can check that in-bounds access return the correct value, which means the test doesn't always sees zeroes because of a bug, and passes because of that.
 - By looking at the values, we can learn about how RBA is implemented on different hardware. (and maybe constrain it to be only clamping in the future?)

src/webgpu/shader/execution/robust_access_vertex.spec.ts
kainino0x: Either is fine, whatever works well.
Realized my suggestion was not right because it doesn't have a key, but you can fix that.

Maybe this would be even simpler:
```ts
.combine(pbool('indexed'))
.expand(p => poptions('drawCallTestParameter', p.indexed ?
  ['indexCount', 'instanceCount', 'firstIndex', 'baseVertex', 'firstInstance'] :
  ['vertexCount', 'instanceCount', 'firstVertex', 'firstInstance']))
```

Or depending on how you're actually writing the test:

```ts
function makeSequence(n) { return [... Array(n).keys()]; }

...
.combine(pbool('indexed'))
.expand(p => poptions('drawCallTestParameter', makeSequence(p.indexed ? 5 : 4))

...

const args = p.indexed ? [1, 1, 1, 1, 1] : [1, 1, 1, 1];
args[p.drawCallTestParameter] = 100;
```

or if you want more explicitness/control over the args:

```ts
.combine(pbool('indexed'))
.expand(p => poptions('drawCallArgs', p.indexed ? [
  [100,1,1,1,1],
  [1,100,1,1,1],
  ...
] : [
  [100,1,1,1],
  [1,100,1,1],
  ...
])
```

src/webgpu/shader/execution/robust_access_vertex.spec.ts
kainino0x: Sorry, link moved and I guess it doesn't redirect: https://hackmd.io/@webgpu/H1MwoqqAU

src/webgpu/shader/execution/robust_access_vertex.spec.ts
iraiter2: Sounds good

src/webgpu/shader/execution/robust_access_vertex.spec.ts
iraiter2: Sounds good. Maybe it would be best to have just a few random values throughout, so that a real OOB read isn't accidentally marked valid?

src/webgpu/shader/execution/robust_access_vertex.spec.ts
iraiter2: Thank you very much! Will try these.

src/webgpu/shader/execution/robust_access_vertex.spec.ts
iraiter2: Talking about it we think a few values repeated for now would work, and randomness can be added if we think conflicting buffers could exist from previous tests. But, the random values should be repeatable so that may be more complicated.

iraiter2:
@kainino0x How's this look so far? Should we edit anything or merge it?

PR #233, opened 2020-07-13
Add a test for minBufferBindingSize

Richard-Yunchao:


Richard-Yunchao:
PTAL. 
@kainino0x and @Kangz and @austinEng , how can I add minBufferBindingSize into node_modules/@webgpu/types/dist/index.d.ts?

austinEng:
Please make a PR here https://github.com/gpuweb/types, and then Kai or I have to publish it to NPM

Richard-Yunchao:
> Please make a PR here https://github.com/gpuweb/types, and then Kai or I have to publish it to NPM

Thank you Austin. 

Now that the change has been made in that project. How could I re-run the CI? BTW, please take a look at this pr when you have time. If this one is OK, I will add tests for a few other parameters which must be undefined or false under some situations.

kainino0x:
you'll have to update the `@webgpu/types` dependency in package.json to get the new types. I just published 0.0.26 so you can update to that version.

kainino0x:
I think I messed up that release, use 0.0.27 instead.

Richard-Yunchao:
The CI failed again. Looking into the failure info, I suppose that we need to update package-lock.json too? But I tried at my local machine, it didn't work (I only changed the version from 0.0.24 to 0.0.27 for `@webgpu/types` in package-lock.json).

austinEng:
You should update `package.json` and then run `npm install` again. That will automatically update `package-lock.json`

kainino0x:
oh yes, sorry, forgot about that step.

src/webgpu/api/validation/createBindGroupLayout.spec.ts
```diff
@@ -64,6 +64,33 @@ g.test('visibility_and_dynamic_offsets')
...
+  .fn(async t => {
+    const { type, minBufferBindingSize } = t.params;
+
+    let success = false;
+    if (minBufferBindingSize > 0) {
+      if (
+        type === 'uniform-buffer' ||
+        type === 'storage-buffer' ||
+        type === 'readonly-storage-buffer'
+      ) {
```
kainino0x, line 82: optional nit:
```suggestion
      if (type in kBufferBindingTypeInfo) {
```

src/webgpu/api/validation/createBindGroupLayout.spec.ts
```diff
@@ -64,6 +64,33 @@ g.test('visibility_and_dynamic_offsets')
     }, !success);
   });
 
+g.test('mininum_buffer_binding_size')
+  .params(
+    params()
+      .combine(poptions('type', kBindingTypes))
+      .combine(poptions('minBufferBindingSize', [0, 4]))
```
kainino0x, line 71: I think we should test `undefined` here in order to test the default value of this option.
```suggestion
      .combine(poptions('minBufferBindingSize', [undefined, 0, 4]))
```

Most of the time I think these will go in operation tests, but in this case it only really affects validation.

kainino0x:
Thanks!

Richard-Yunchao:
Thanks for your review, @kainino0x and @austinEng !

PR #235, opened 2020-07-13
Update node_module/@webgpu/types to 0.0.27

Richard-Yunchao:
GPUTextureCopyView.arrayLayer was merged into .origin.z in
webgpu/types, this patch revised corresponding tests.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -57,7 +57,7 @@
         "debug": {
           "version": "4.1.1",
           "resolved": "https://registry.npmjs.org/debug/-/debug-4.1.1.tgz",
-          "integrity": "sha1-O3ImAlUQnGtYnO4FDx1RYTlmR5E=",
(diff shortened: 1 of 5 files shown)


Richard-Yunchao:
There are a lot of changes in package-lock.json when I run `npm install` to update node_module/webgpu/types project for CTS, so I submit this separate PR. PTAL. @kainino0x , @austinEng . Thank you.

kainino0x:
Thanks for taking care of these test fixes. It looks like the reason your package.lock changed so much is the hashes reverted from sha512 to sha1, which I guess is probably due to an older npm. I'll pull this and do another npm update to see if I can keep it on sha512.

Richard-Yunchao:
Oops. I just updated my system to Ubuntu 20.04 last month. It's weird that npm is not new enough. Thanks for your revision, Kai

PR #239, opened 2020-07-15
Add validation for index access

haoxli:
Validate index access out of bounds in drawIndexed and drawIndexedIndirect.

### DIFF
--- a/src/webgpu/api/validation/encoding/cmds/index_access.spec.ts
+++ b/src/webgpu/api/validation/encoding/cmds/index_access.spec.ts
@@ -1,0 +1,167 @@
+export const description = `
+indexed draws validation tests.
+`;
+
+import { makeTestGroup } from '../../../../../common/framework/test_group.js';
+
+import { ValidationTest } from './../../validation_test.js';
+
+class F extends ValidationTest {
+  createVertexAndIndexBuffer(): GPUBuffer[] {
+    /* prettier-ignore */
+    const vertexArray = new Float32Array([
+      // float4 position
+      1.0, 1.0, 0.0, 1.0,
+      1.0, -1.0, 0.0, 1.0,
+      -1.0, 1.0, 0.0, 1.0,
+      - 1.0, -1.0, 0.0, 1.0
+    ]);
+
+    const indexArray = new Uint32Array([0, 1, 2, 3, 1, 2]);
+
+    const [vertexBuffer, vetexMapping] = this.device.createBufferMapped({
+      size: vertexArray.byteLength,
+      usage: GPUBufferUsage.VERTEX,
+    });
+    new Float32Array(vetexMapping).set(vertexArray);
+    vertexBuffer.unmap();
+
+    const [indexBuffer, indexMapping] = this.device.createBufferMapped({
+      size: indexArray.byteLength,
+      usage: GPUBufferUsage.INDEX,
+    });
+    new Uint32Array(indexMapping).set(indexArray);
+    indexBuffer.unmap();
+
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/encoding/cmds/index_access.spec.ts
```diff
@@ -0,0 +1,167 @@
...
+    pass.drawIndexedIndirect(indirectBuffer, indirectOffset);
+    pass.endPass();
+
+    this.device.defaultQueue.submit([encoder.finish()]);
+  }
+}
+
+export const g = makeTestGroup(F);
+
+g.test('draw_indexed_index_access_out_of_bounds').fn(t => {
```
kainino0x, line 145: nit: Test name is slightly redundant with the filename. Maybe just `out_of_bounds`.

kainino0x, line 145: Can be parameterized to make it easier to add additional cases, perhaps something like this:

```ts
g.test('out_of_bounds')
  .params(
    params()
      .combine(pbool('indirect'))
      .combine([
        { indexCount: 7, firstIndex: 0 },
        { indexCount: 10000, firstIndex: 0 },
        { indexCount: 6, firstIndex: 1 },
        { indexCount: 6, firstIndex: 10000 },
      ] as const)
      .combine(poptions('instanceCount', [1, 10000]))
  )
  .fn(t => {
    const { indirect, indexCount, firstIndex, instanceCount } = t.params;
    if (indirect) {
      t.drawIndexedIndirect(new Uint32Array([indexCount, instanceCount, firstIndex, 0, 0), 0);
    } else {
      t.drawIndexed(indexCount, instanceCount, firstIndex, 0, 0);
    }
  });
```

src/webgpu/api/validation/encoding/cmds/index_access.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+
+  beginRenderPass(encoder: GPUCommandEncoder) {
+    const colorAttachment = this.device.createTexture({
+      format: 'rgba8unorm',
+      size: { width: 1, height: 1, depth: 1 },
+      usage: GPUTextureUsage.OUTPUT_ATTACHMENT,
+    });
+
+    return encoder.beginRenderPass({
+      colorAttachments: [
```
kainino0x, line 61: (sorry I missed this before)
nit: I'm not sure if any attachments are actually needed at all - I don't remember. Would this test be valid without them?

haoxli: The colorAttachments is required in GPURenderPassDescriptor. It will throw error if no attachment is set.

kainino0x: Thank you for confirming!

kainino0x:
LGTM, merging now but @iraiter2 still feel free to take a look if you have a chance.

Quadmium:
LGTM! Reading through it seems we are testing that OOB calls don't raise a validation error, which I think makes sense. An implementation could pass these tests but fail mine if robust buffer access isn't enabled, but still validates correctly, so I think it makes sense that these are validation tests.

iraiter2:
LGTM! I think the difference from my tests is that these check that no validation error happens if we send an OOB draw call, which makes sense for RBA. Also, I think if an implementation doesn't enable RBA it could pass these tests but fail mine, so putting them as validation tests makes sense (since it would be validating correctly but not executing correctly).

kainino0x:
Yup, agreed. Thanks!

PR #241, opened 2020-07-15
RenderPass StoreOp Validation Tests

bjjones:
Implements Validation Tests For StoreOp

### DIFF
--- a/src/webgpu/api/validation/render_pass/storeOp.spec.ts
+++ b/src/webgpu/api/validation/render_pass/storeOp.spec.ts
@@ -1,13 +1,73 @@
-export const description = `API Validation Tests for RenderPass StoreOp.
+export const description = `
+API Validation Tests for RenderPass StoreOp.
 
-  Test Coverage Needed:
+Test Coverage:
+  - Tests that when depthReadOnly is true, depthStoreOp must be 'store'.
+    - When depthReadOnly is true and depthStoreOp is 'clear', an error should be generated.
 
-  - Test that when depthReadOnly is true, depthStoreOp must be 'store'
+  - Tests that when stencilReadOnly is true, stencilStoreOp must be 'store'.
+    - When stencilReadOnly is true and stencilStoreOp is 'clear', an error should be generated.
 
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/render_pass/storeOp.spec.ts
```diff
@@ -1,13 +1,156 @@
...
+API Validation Tests for RenderPass StoreOp.
+`;
 
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
 
 import { ValidationTest } from './../validation_test.js';
 
 export const g = makeTestGroup(ValidationTest);
+
+g.test('depthStoreOp_must_be_store_when_depthReadOnly_is_true').fn(async t => {
```
kainino0x, line 11: Rather than having both the control case and failure case expanded out like this, we can use parameterization to condense it. In addition, we can combine all of these tests into one.

Opened a PR with these suggestions here:
https://github.com/bjjones/cts/pull/1
(untested)
and I think it covers the comments I have below too.

src/webgpu/api/validation/render_pass/storeOp.spec.ts
```diff
@@ -1,13 +1,156 @@
-export const description = `API Validation Tests for RenderPass StoreOp.
-
-  Test Coverage Needed:
-
-  - Test that when depthReadOnly is true, depthStoreOp must be 'store'
-
-  - Test that when stencilReadOnly is true, stencilStoreOp must be 'store'`;
+export const description = `
+API Validation Tests for RenderPass StoreOp.
+`;
```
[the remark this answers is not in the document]
bjjones: I added it back and made it more detailed - lmk if you want something different.

src/webgpu/api/validation/render_pass/storeOp.spec.ts
bjjones: Using your code now, which resolves everything else.

bjjones:
> @austinEng PTAL since I contributed major changes to this.

FYI - I tested the change locally and test results were as expected.

PR #238, opened 2020-07-14
Add tests for bindGroupLayout's variables

Richard-Yunchao:
This change adds tests for 'viewDimension', 'textureComponentType',
'multisampled', and 'storageTextureFormat' in bindGroupLayout.

PR #242, opened 2020-07-16
In params().combine() types, flatten unions of interfaces before merging them

kainino0x:
Bug: https://github.com/gpuweb/cts/pull/238#issuecomment-659086834

Changes the type of
```ts
const p = params()
  .combine([{ a: 1 }])
  .combine([
    ...poptions('b', [1]), //
    ...poptions('c', [1]),
  ])
```
from `ParamsBuilder<{ a: number }>`
to `ParamsBuilder<{ a: number, b: number | undefined, c: number | undefined }>`.

### DIFF
--- a/src/common/framework/params_builder.ts
+++ b/src/common/framework/params_builder.ts
@@ -1,14 +1,97 @@
 import { CaseParams, CaseParamsIterable, publicParamsEquals } from './params_utils.js';
 import { assert } from './util/util.js';
 
-// https://stackoverflow.com/a/56375136
+/** Forces a type to resolve its type definitions, to make it readable/debuggable. */
+export type ResolveType<T> = T extends object
+  ? T extends infer O
+    ? { [K in keyof O]: ResolveType<O[K]> }
+    : never
+  : T;
+
+/**
+ * Computes the intersection of a set of types, given the union of those types.
+ *
+ * From: https://stackoverflow.com/a/56375136
+ */
 /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
 type UnionToIntersection<U> = (U extends any ? (k: U) => void : never) extends (k: infer I) => void
   ? I
   : never;
+
+/** Conditionally chooses between two types depending on whether T is a union. */
 type CheckForUnion<T, TErr, TOk> = [T] extends [UnionToIntersection<T>] ? TOk : TErr;
 
+/** Conditionally chooses a type (or void) depending on whether T is a string. */
 type CheckForStringLiteralType<T, TOk> = string extends T ? void : CheckForUnion<T, void, TOk>;
+
+type KeyOfNeverable<T> = T extends never ? never : keyof T;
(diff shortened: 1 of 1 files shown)


kainino0x:
Opening tentatively, might want to figure out whether this is actually correct before merging though

kainino0x:
@austinEng I can't quite get this working the way I want, but PTAL

austinEng:
err whoops I just saw this

kainino0x:
No problem, it changed completely

austinEng:
to make sure I'm understanding, this is a different goal than:
```typescript
const p = params()
  .combine([{ a: 1 }])
  .combine(poptions('b', [1, 2]))
  .combine(poptions('c', [3, 4]))
  ])
```

which would yield `{ a: 1, b: 1, c: 3}, { a: 1, b: 1, c: 4}, { a: 1, b: 2, c: 3}, { a: 1, b: 2, c: 4}`

```typescript
const p = params()
  .combine([{ a: 1 }])
  .combine([...poptions('b', [1, 2]), ...poptions('c', [3, 4])])
  ])
```

looks like `{ a: 1, b: 1}, { a: 1, b: 2}, { a: 1, c: 3}, { a: 1, c: 4}`

src/common/framework/params_builder.ts
```diff
@@ -1,15 +1,98 @@
...
+ *
+ * And also works for `{ a: number } | { b: string }` which maps to the same.
+ */
+type FlattenUnionOfInterfaces<T> = {
+  [K in AllKeysFromUnion<T>]: KeyOfOr<
+    T,
+    // If T always has K, just take T[K] (union of C[K] for each component C of T):
+    K,
+    // Otherwise, take the union of C[K] for each component C of T, PLUS undefined:
+    undefined | KeyOfOr<UnionToIntersection<T>, K, void>
```
austinEng, line 47: Is the following correct?
 - K is only a keyof the union T if it is present in all of the union's subtypes,
 - if we're in the default case, it means K was absent in at least one of T's subtypes
 - so UnionToIntersection<T>[K] will be never?

I believe these are the cases you describe in your comment "Unexpected test results - hopefully okay to ignore these"

austinEng: Okay nvm I see it's only a problem if in UnionToIntersection the types conflict -- that's when you'll get never. but if they're optional (undefined) then it's ok.

kainino0x: Yup. `keyof ({a: number} | {b: string})` is `never` but `keyof ({a: number} & {b: string})` is `'a' | 'b'` I think. for some reason.

kainino0x:
> to make sure I'm understanding

👍

kainino0x:
Types can recurse, so if there was a way to decompose (i.e. `reduce()` over) union types, I think you could do that instead of variadic. But I don't think you can do a pattern match to decompose a union into `First | ...Rest`.

PR #243, opened 2020-07-17
Implement RenderPass StoreOp Operation Tests

bjjones:
Initially I expected this to be one big test - however I found that when I combined all the discussed parameters, it noticeably impacted the site load time. I've split it into 4 separate tests that I think cover all the scenarios discussed effectively.

### DIFF
--- a/src/webgpu/api/operation/render_pass/storeOp.spec.ts
+++ b/src/webgpu/api/operation/render_pass/storeOp.spec.ts
@@ -1,23 +1,326 @@
 export const description = `API Operation Tests for RenderPass StoreOp.
 
-  Test Coverage Needed:
+  Test Coverage:
 
-  - Test that a render pass has correct output for combinations of:
-    - All color attachments from '0' to 'MAX_COLOR_ATTACHMENTS' with combinations of:
-      - storeOp set to {'clear', 'store', 'undefined}
-      - All color renderable formats
+  - Tests that color and depth-stencil store operations {'clear', 'store'} work correctly for a
+    render pass with both a color attachment and depth-stencil attachment.
+      TODO: use depth24plus-stencil8
+
+  - Tests that store operations {'clear', 'store'} work correctly for a render pass with multiple
+    color attachments.
+      TODO: test with more interesting loadOp values
+
+  - Tests that store operations {'clear', 'store'} work correctly for a render pass with a color
+    attachment for:
+      - All renderable color formats
       - mip level set to {'0', mip > '0'}
       - array layer set to {'0', layer > '1'} for 2D textures
-      - depth slice set to {'0', slice > '0'} for 3D textures
-    - With and without a depthStencilAttachment that has the combinations of:
-      - depthStoreOp set to {'clear', 'store', 'undefined'}
-      - stencilStoreOp set to {'clear', 'store', 'undefined'}
-      - All depth/stencil formats
+      TODO: depth slice set to {'0', slice > '0'} for 3D textures
+
+  - Tests that store operations {'clear', 'store'} work correctly for a render pass with a
+    depth-stencil attachment for:
+      - All renderable depth-stencil formats
       - mip level set to {'0', mip > '0'}
       - array layer set to {'0', layer > '1'} for 2D textures
-      - depth slice set to {'0', slice > '0'} for 3D textures`;
+      TODO: test depth24plus and depth24plus-stencil8 formats
+      TODO: test that depth and stencil aspects are set seperately
+      TODO: depth slice set to {'0', slice > '0'} for 3D textures
+      TODO: test with more interesting loadOp values`;
 
+import { params, poptions } from '../../../../common/framework/params_builder.js';
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { kTextureFormatInfo, kTextureFormats } from '../../../capability_info.js';
 import { GPUTest } from '../../../gpu_test.js';
+import { PerTexelComponent } from '../../../util/texture/texelData.js';
+
+// Test with a zero and non-zero mip.
+const kMipLevel: number[] = [0, 1];
+const kMipLevelCount = 2;
+
+// Test with different numbers of color attachments.
+type NumColorAttachments = 1 | 2 | 3 | 4;
+const kNumColorAttachments: NumColorAttachments[] = [1, 2, 3, 4];
+
+// Test with a zero and non-zero array layer.
+const kArrayLayers: number[] = [0, 1];
+
+const kStoreOps: GPUStoreOp[] = ['clear', 'store'];
+
+const kHeight = 2;
+const kWidth = 2;
 
 export const g = makeTestGroup(GPUTest);
+
+// Tests a render pass with both a color and depth stencil attachment to ensure store operations are
+// set independently.
+g.test('render_pass_store_op,color_attachment_with_depth_stencil_attachment')
+  .params(
+    params()
+      .combine(poptions('colorStoreOperation', kStoreOps))
+      .combine(poptions('depthStencilStoreOperation', kStoreOps))
+  )
+  .fn(t => {
+    // Create a basic color attachment.
+    const kColorFormat: GPUTextureFormat = 'rgba8unorm';
+    const colorAttachment = t.device.createTexture({
+      format: kColorFormat,
+      size: { width: kWidth, height: kHeight, depth: 1 },
+      usage: GPUTextureUsage.COPY_SRC | GPUTextureUsage.OUTPUT_ATTACHMENT,
+    });
+
+    const colorAttachmentView = colorAttachment.createView();
+
+    // Create a basic depth/stencil attachment.
+    const kDepthStencilFormat: GPUTextureFormat = 'depth32float';
+    const depthStencilAttachment = t.device.createTexture({
+      format: kDepthStencilFormat,
+      size: { width: kWidth, height: kHeight, depth: 1 },
+      usage: GPUTextureUsage.COPY_SRC | GPUTextureUsage.OUTPUT_ATTACHMENT,
+    });
+
+    // Color load operation will clear to {1.0, 1.0, 1.0, 1.0}.
+    // Depth & stencil load operations will clear to 1.0.
+    // Store operations are determined by test the params.
+    const encoder = t.device.createCommandEncoder();
+    const pass = encoder.beginRenderPass({
+      colorAttachments: [
+        {
+          attachment: colorAttachmentView,
+          loadValue: { r: 1.0, g: 1.0, b: 1.0, a: 1.0 },
+          storeOp: t.params.colorStoreOperation,
+        },
+      ],
+      depthStencilAttachment: {
+        attachment: depthStencilAttachment.createView(),
+        depthLoadValue: 1.0,
+        depthStoreOp: t.params.depthStencilStoreOperation,
+        stencilLoadValue: 1.0,
+        stencilStoreOp: t.params.depthStencilStoreOperation,
+      },
+    });
+    pass.endPass();
+
+    t.device.defaultQueue.submit([encoder.finish()]);
+
+    // Check that the correct store operation occurred.
+    let expectedColorValue: PerTexelComponent<number> = {};
+    if (t.params.colorStoreOperation === 'clear') {
+      // If colorStoreOp was clear, the texture should now contain {0.0, 0.0, 0.0, 0.0}.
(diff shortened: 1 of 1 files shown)


kainino0x:
> Initially I expected this to be one big test - however I found that when I combined all the discussed parameters, it noticeably impacted the site load time. I've split it into 4 separate tests that I think cover all the scenarios discussed effectively.

We really need to do something to improve the load time of the standalone test page, but I haven't gotten around to it. If we lazily loaded these then it wouldn't cause a slowdown.

That said we do still need to keep the combinatorial explosion under control. So I think it's okay to reduce the combinations here.

src/webgpu/api/operation/render_pass/storeOp.spec.ts
```diff
@@ -1,23 +1,352 @@
...
+      format: t.params.depthStencilFormat,
+      size: { width: kWidth, height: kHeight, depth: t.params.arrayLayer + 1 },
+      mipLevelCount: t.params.mipLevel + 1,
+      usage: GPUTextureUsage.COPY_SRC | GPUTextureUsage.OUTPUT_ATTACHMENT,
+    });
+
+    const depthStencilViewDesc: GPUTextureViewDescriptor = {
+      format: t.params.depthStencilFormat,
+      dimension: '2d',
+      aspect: 'all',
```
kainino0x, line 312: We should probably test rendering into views with subsets of the aspects to make sure those also behave properly.

src/webgpu/api/operation/render_pass/storeOp.spec.ts
```diff
@@ -1,23 +1,352 @@
...
+    const depthStencilAttachment = t.device.createTexture({
+      format: t.params.depthStencilFormat,
+      size: { width: kWidth, height: kHeight, depth: t.params.arrayLayer + 1 },
+      mipLevelCount: t.params.mipLevel + 1,
+      usage: GPUTextureUsage.COPY_SRC | GPUTextureUsage.OUTPUT_ATTACHMENT,
+    });
+
+    const depthStencilViewDesc: GPUTextureViewDescriptor = {
+      format: t.params.depthStencilFormat,
+      dimension: '2d',
```
kainino0x, line 311: nit: shouldn't need to specify these

src/webgpu/api/operation/render_pass/storeOp.spec.ts
```diff
@@ -1,23 +1,352 @@
...
+      });
+    }
+  });
+
+// Tests that render pass depth stencil store operations work correctly for all renderable color
+// formats, mip levels and array layers.
+g.test('render_pass_store_op,depth_stencil_attachment_only')
+  .params(
+    params()
+      .combine(poptions('depthStencilFormat', kTextureFormats))
```
kainino0x, line 289: I'm thinking perhaps we should split kTextureFormats/FormatInfo into two tables that get merged. That way we don't have to do filtering when we really just want one or the other. You're welcome to do that, otherwise I'll do the refactor sometime.

src/webgpu/api/operation/render_pass/storeOp.spec.ts
```diff
@@ -1,23 +1,352 @@
...
+      // Filter out color and non-renderable formats.
+      .unless(
+        ({ depthStencilFormat }) =>
+          kTextureFormatInfo[depthStencilFormat].color ||
+          !kTextureFormatInfo[depthStencilFormat].renderable ||
+          !kTextureFormatInfo[depthStencilFormat].copyable
```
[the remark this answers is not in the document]
austinEng: Hmm I think it would probably be worth generalizing `expectSingleColor` to do this.
We can always copy from the stencil aspect though -- currently unimplemented in Dawn.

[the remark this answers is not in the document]
austinEng: Actually we could probably go beyond `expectSingleColor` and make a full `expectContents` helper. For the stencil we would copy. For depth32 we would copy, and for depth24 we would draw a bunch of points at the desired depth and then do depth test equals.

Maybe that's more than we need though.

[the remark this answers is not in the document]
bjjones: I can write another test. Making expectSingleColor more generic to accept depth24 formats seems ideal. Do I need to be the one to do this? - and if so, I guess it needs to be done with WGSL?

src/webgpu/api/operation/render_pass/storeOp.spec.ts
bjjones: I agree - although Dawn doesn't support anything other than "all" at the moment, which will make it difficult for me to test my test.

kainino0x: sure, let's have a TODO for this

src/webgpu/api/operation/render_pass/storeOp.spec.ts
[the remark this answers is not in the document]
kainino0x: Could be done with GLSL in the glsl-dependent branch, but not urgent. Let's leave a TODO for this.

src/webgpu/api/operation/render_pass/storeOp.spec.ts
```diff
@@ -1,23 +1,352 @@
...
+      .combine(poptions('colorStoreOperation', kStoreOps))
+      .combine(poptions('depthStencilStoreOperation', kStoreOps))
+  )
+  .fn(t => {
+    // Create a basic color attachment.
+    const kColorFormat: GPUTextureFormat = 'rgba8unorm';
+    const colorAttachment = t.device.createTexture({
+      format: kColorFormat,
+      size: { width: kWidth, height: kHeight, depth: 1 },
+      mipLevelCount: 1,
```
Kangz, line 67: nit: that's the default value and can be skipped

src/webgpu/api/operation/render_pass/storeOp.spec.ts
```diff
@@ -1,23 +1,352 @@
...
-  Test Coverage Needed:
+  Test Coverage:
 
-  - Test that a render pass has correct output for combinations of:
-    - All color attachments from '0' to 'MAX_COLOR_ATTACHMENTS' with combinations of:
-      - storeOp set to {'clear', 'store', 'undefined}
-      - All color renderable formats
+  - Tests that color and depth-stencil store operations {'clear', 'store'} work correctly for a
+    render pass with both a color attachment and depth-stencil attachment.
+      TODO: Use depth24plus-stencil8
```
Kangz, line 7: nit: TODO also check that depthStoreOp and stencilStoreOp are separate.

src/webgpu/api/operation/render_pass/storeOp.spec.ts
```diff
@@ -1,23 +1,352 @@
...
+      .combine(poptions('storeOperation', kStoreOps))
+      .combine(poptions('mipLevel', kMipLevel))
+      .combine(poptions('arrayLayer', kArrayLayers))
+  )
+  .fn(t => {
+    const kColorFormat: GPUTextureFormat = 'rgba8unorm';
+    const colorAttachment = t.device.createTexture({
+      format: kColorFormat,
+      size: { width: kWidth, height: kHeight, depth: t.params.arrayLayer + 1 },
+      mipLevelCount: t.params.mipLevel + 1,
```
Kangz, line 159: nit: Maybe have a fixed mip level count so we don't always test rendering to the last one.

src/webgpu/api/operation/render_pass/storeOp.spec.ts
```diff
@@ -1,23 +1,352 @@
...
+      usage: GPUTextureUsage.COPY_SRC | GPUTextureUsage.OUTPUT_ATTACHMENT,
+    });
+
+    const colorViewDesc: GPUTextureViewDescriptor = {
+      format: kColorFormat,
+      dimension: '2d',
+      aspect: 'all',
```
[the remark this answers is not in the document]
Kangz: More specifically, "format", "dimension" and "aspect" can be removed.

src/webgpu/api/operation/render_pass/storeOp.spec.ts
```diff
@@ -1,23 +1,352 @@
...
+      baseArrayLayer: t.params.arrayLayer,
+      baseMipLevel: t.params.mipLevel,
+      arrayLayerCount: 1,
+      mipLevelCount: 1,
+    };
+
+    const colorAttachmentView = colorAttachment.createView(colorViewDesc);
+
+    // Color load operation will clear to {1.0, 0.0, 0.0, 1.0}.
```
Kangz, line 175: Can we have all non-zeros? It would help make sure all components are correctly written.

src/webgpu/api/operation/render_pass/storeOp.spec.ts
```diff
@@ -1,23 +1,352 @@
...
+  .fn(t => {
+    const kColorFormat: GPUTextureFormat = 'rgba8unorm';
+    const colorAttachments: GPUTexture[] = [];
+
+    for (let i = 0; i < t.params.colorAttachments; i++) {
+      colorAttachments.push(
+        t.device.createTexture({
+          format: kColorFormat,
+          size: { width: kWidth, height: kHeight, depth: 1 },
+          mipLevelCount: 1,
```
Kangz, line 225: ditto

src/webgpu/api/operation/render_pass/storeOp.spec.ts
```diff
@@ -1,23 +1,352 @@
...
+    const storeOps: GPUStoreOp[] = [];
+    for (let i = 0; i < t.params.colorAttachments; i++) {
+      if (i % 2 === 0) {
+        storeOps.push(t.params.storeOperation1);
+      } else {
+        storeOps.push(t.params.storeOperation2);
+      }
+    }
+
+    // Color load operation will clear to {1.0, 0.0, 0.0, 1.0}
```
Kangz, line 249: ditto

src/webgpu/api/operation/render_pass/storeOp.spec.ts
```diff
@@ -1,23 +1,352 @@
...
+      )
+      .combine(poptions('storeOperation', kStoreOps))
+      .combine(poptions('mipLevel', kMipLevel))
+      .combine(poptions('arrayLayer', kArrayLayers))
+  )
+  .fn(t => {
+    const depthStencilAttachment = t.device.createTexture({
+      format: t.params.depthStencilFormat,
+      size: { width: kWidth, height: kHeight, depth: t.params.arrayLayer + 1 },
+      mipLevelCount: t.params.mipLevel + 1,
```
Kangz, line 305: ditto

bjjones:
Sorry this is staying open so long. I’m needing to take some time off. I’ll have the next version on Monday.

Kangz:
Oh sorry about the ping then, no rush.

kainino0x:
No worries, just wanted to make sure we weren't blocking *your* work!

Kangz:
Something that came up in https://dawn-review.googlesource.com/c/dawn/+/25643/3..4 is that we should test "interesting" values for clearing each format. Things like UINT32MAX, +-INFINITY, Nan, +0, -0, INT32MIN (depending on the format of course). It could be done as a follow-up and the list of "interesting values" per format put in a common place so that it can be reused between tests.

src/webgpu/api/operation/render_pass/storeOp.spec.ts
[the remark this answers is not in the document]
bjjones: TODO added.

src/webgpu/api/operation/render_pass/storeOp.spec.ts
bjjones: TODO added.

src/webgpu/api/operation/render_pass/storeOp.spec.ts
bjjones: Added.

austinEng:
> Something that came up in https://dawn-review.googlesource.com/c/dawn/+/25643/3..4 is that we should test "interesting" values for clearing each format. Things like UINT32MAX, +-INFINITY, Nan, +0, -0, INT32MIN (depending on the format of course). It could be done as a follow-up and the list of "interesting values" per format put in a common place so that it can be reused between tests.

slightly related: I remember running into validation layer errors when using floating point clear values to clear normalized textures outside the [0, 1] range (similar reason as to why we can't copy values outside 0-1 to depth textures). We probably need to add validation rules about this.

PR #245, opened 2020-07-22
Update to use new mapping primitives everywhere

kainino0x:
No new tests for the new mapping primitives (in particular the offset/size arguments) yet, but expanded the map_oom tests a little bit.


src/webgpu/api/operation/buffers/map.spec.ts
```diff
@@ -16,19 +16,22 @@ g.test('mapWriteAsync')
...
   });
 
-g.test('mapReadAsync')
+g.test('mapAsync,read')
   .params(poptions('size', [12, 512 * 1024]))
   .fn(async t => {
     const { size } = t.params;
 
-    const [buffer, init] = t.device.createBufferMapped({
+    const buffer = t.device.createBuffer({
```
Kangz, line 29: Shouldn't this test us `mapAsync(GPUMapMode.READ)` ?

src/webgpu/api/operation/buffers/map_oom.spec.ts
```diff
@@ -2,32 +2,61 @@ export const description = '';
...
-    t.device.createBufferMapped(getBufferDesc(GPUBufferUsage.COPY_SRC));
+g.test('mappedAtCreation')
+  .params(
+    params()
+      .combine(pbool('oom')) //
+      .combine(poptions('usage', kBufferUsages))
+  )
+  .fn(async t => {
+    const { oom, usage } = t.params;
+    const size = oom ? Number.MAX_SAFE_INTEGER - 3 : 16;
```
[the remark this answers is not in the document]
kainino0x: I don't think createBuffer has size alignment constraints? But made this consistently a multiple of 8 to avoid testing that here.

src/webgpu/api/operation/buffers/map_oom.spec.ts
```diff
@@ -2,32 +2,90 @@ export const description = '';
...
+      f();
+    }
   });
-  t.shouldReject('OperationError', buffer.mapReadAsync());
-});
 
-g.test('createBufferMapped').fn(async t => {
-  t.shouldThrow('RangeError', () => {
-    t.device.createBufferMapped(getBufferDesc(GPUBufferUsage.COPY_SRC));
+g.test('mappedAtCreation,smaller_getMappedRange')
```
kainino0x, line 76: PTAL at this in particular - is this the expected behavior? (Doing this definitely requires lazy allocating, unless we expect to crash instead.)

src/webgpu/api/operation/buffers/map_oom.spec.ts
```diff
@@ -2,32 +2,90 @@ export const description = '';
...
+    const buffer = t.expectGPUError(
+      'out-of-memory',
+      () =>
+        t.device.createBuffer({
+          size,
+          usage: write ? GPUBufferUsage.MAP_WRITE : GPUBufferUsage.MAP_READ,
+        }),
+      oom
+    );
+    const promise = t.expectGPUError(
```
kainino0x, line 34: This too - is it correct to expect an out-of-memory error in the scope in addition to the promise rejection? (I think we said yes somewhere?)

src/webgpu/api/operation/buffers/map_oom.spec.ts
austinEng: I think this makes sense.

src/webgpu/api/operation/buffers/mapping_test.ts
```diff
@@ -1,18 +1,24 @@
...
+    offset: number,
+    mappedContents: ArrayBuffer,
+    size: number
+  ): void {
     this.checkMapWriteZeroed(mappedContents, size);
 
     const mappedView = new Uint32Array(mappedContents);
     const expected = new Uint32Array(new ArrayBuffer(size));
-    this.expect(mappedView.byteLength === size);
+    assert(mappedView.byteLength === size);
```
austinEng, line 15: Why does this need to be an assert instead of an expect? If the length is wrong, is the rest of the error log too noisy?

src/webgpu/api/operation/buffers/map.spec.ts
```diff
@@ -1,57 +1,87 @@
...
+  { size: 12, range: [0, 0] },
+  { size: 512 * 1024, range: [] },
+];
+
+g.test('mapAsync,write')
+  .params(kCases)
   .fn(async t => {
-    const { size } = t.params;
+    const { size, range } = t.params;
+    const rangeSize = range[1] !== undefined ? range[1] : size;
```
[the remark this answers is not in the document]
kainino0x: whoops, I meant to add those cases.

src/webgpu/api/operation/buffers/map_oom.spec.ts
kainino0x: Okay, so both `validation` error and rejection?

src/webgpu/api/operation/buffers/mapping_test.ts
kainino0x: If the length is wrong it doesn't make sense to check the contents because the loop could go too long for `expected[i]`.

kainino0x: (Could just let it hit an exception, or use min(mappedView.length, expected.length), though)

src/webgpu/api/operation/buffers/map_oom.spec.ts
austinEng: Yup

src/webgpu/api/operation/buffers/mapping_test.ts
austinEng: ok yea, seems fine to `assert`

PR #247, opened 2020-07-23
Represent undefined with a magic value in JSON

kainino0x:
Needed for testing in #245.

Allows representing `null` and `undefined` separately inside param values. Previously did this only for the toplevel value (which was needed for parameterized testing of a bunch of apis that take undefined), but this is more robust.

### DIFF
--- a/src/common/framework/query/json_param_value.ts
+++ b/src/common/framework/query/json_param_value.ts
@@ -1,0 +1,18 @@
+import { ParamArgument } from '../params_utils.js';
+import { assert } from '../util/util.js';
+
+// JSON can't represent `undefined` and by default stores it as `null`.
+// Instead, store `undefined` as this magic string value in JSON.
+const jsUndefinedMagicValue = '✗undefined';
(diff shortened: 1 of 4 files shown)


src/common/framework/query/json_param_value.ts
```diff
@@ -0,0 +1,18 @@
+import { ParamArgument } from '../params_utils.js';
+import { assert } from '../util/util.js';
+
+// JSON can't represent `undefined` and by default stores it as `null`.
+// Instead, store `undefined` as this magic string value in JSON.
+const jsUndefinedMagicValue = '✗undefined';
```
kainino0x, line 6: I almost made it `𝓊𝓃𝒹𝑒𝒻𝒾𝓃𝑒𝒹` but it was too annoying for encode_selectively 😁

austinEng: heh, nice :) I didn't even know such a thing could be possible

PR #248, opened 2020-07-25
add ParamsBuilder.combineArgs to work around a type inference issue

kainino0x:
TypeScript overly-conservatively infers the type in this situation:

```ts
const x = [{ a: 0 }];
const y = [{}, ...x];  // has type {}[]
```

which was a problem for `.combine([{}, ...x])`. `.combineArgs({}, ...x)` works properly.

But we can work around it with `.combine([{ _: 0 }, ...x])`, so just closing this and keeping it for posterity.

### DIFF
--- a/src/common/framework/params_builder.ts
+++ b/src/common/framework/params_builder.ts
@@ -48,6 +48,11 @@ type FlattenUnionOfInterfaces<T> = {
   >;
 };
 
+/**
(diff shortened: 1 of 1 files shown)


kainino0x:
We can work around it with `.combine([{ _: 0 }, ...x])`, so just closing this and keeping it for posterity.

PR #249, opened 2020-07-25
Fixes about hasDynamicOffset and multisampled in BGL

Richard-Yunchao:
https://github.com/gpuweb/gpuweb/pull/949

hasDynamicOffset and multisampled should be undefined in BGL if the
binding types are incorrect. Previously we allow them to be false.

This change moves hasDynamicOffset tests out of visiblity test
and combine it with other optional parameters in BGL.

It also added a validation rule that if the binding type is storage
texture, the viewDimension can't be 'cube' or 'cube-array'.

### DIFF
--- a/src/webgpu/api/validation/createBindGroupLayout.spec.ts
+++ b/src/webgpu/api/validation/createBindGroupLayout.spec.ts
@@ -46,26 +46,20 @@ g.test('some_binding_index_was_specified_more_than_once').fn(async t => {
   });
 });
 
-g.test('visibility_and_dynamic_offsets')
+g.test('visibility')
   .params(
     params()
       .combine(poptions('type', kBindingTypes))
-      .combine(pbool('hasDynamicOffset'))
       .combine(poptions('visibility', kShaderStageCombinations))
   )
   .fn(async t => {
-    const { type, hasDynamicOffset, visibility } = t.params;
-    const info = kBindingTypeInfo[type as GPUBindingType];
+    const { type, visibility } = t.params;
 
-    const supportsDynamicOffset = kBindingTypeInfo[type].perPipelineLimitClass.maxDynamic > 0;
-    let success = true;
-    if (!supportsDynamicOffset && hasDynamicOffset) success = false;
-    if ((visibility & ~info.validStages) !== 0) success = false;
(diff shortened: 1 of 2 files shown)


Richard-Yunchao:
This should be the last change about BGL. PTAL. After this patch and another patch about BGL spec revision are merged, CTS and spec will exactly match each other.

src/webgpu/api/validation/createBindGroupLayout.spec.ts
```diff
@@ -100,14 +96,21 @@ g.test('bindingTypeSpecific_optional_members')
...
     let success = true;
     if (!(type in kBufferBindingTypeInfo)) {
+      if (hasDynamicOffset !== undefined) success = false;
       if (minBufferBindingSize !== undefined) success = false;
     }
     if (!(type in kTextureBindingTypeInfo)) {
       if (viewDimension !== undefined) success = false;
+      if (
+        kBindingTypeInfo[type].resource === 'storageTex' &&
+        (viewDimension === 'cube' || viewDimension === 'cube-array')
```
kainino0x, line 106: Put this inside the `if (kBindingTypeInfo[type].resource !== 'storageTex')` on line 115 instead?

kainino0x, line 106: Let's encode this info inside kTextureViewDimensionInfo in capability_info.ts (as a `storage: boolean` member)

src/webgpu/api/validation/createBindGroupLayout.spec.ts
```diff
@@ -46,26 +46,20 @@ g.test('some_binding_index_was_specified_more_than_once').fn(async t => {
   });
 });
 
-g.test('visibility_and_dynamic_offsets')
+g.test('visibility')
```
kainino0x, line 49: nice separation!

src/webgpu/api/validation/createBindGroupLayout.spec.ts
Richard-Yunchao: Done.

Richard-Yunchao: Done

PR #252, opened 2020-07-30
Add texture subresource usage validation tests in render

Richard-Yunchao:
This change added read and write bindings upon the same texture
subresource, and upon different subresources (different mip levels,
different array layers, different aspects) of the same texture.

### DIFF
--- a/src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
+++ b/src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
@@ -1,0 +1,179 @@
+export const description = `
+Texture Usages Validation Tests in Render Pass.
+
+Test Coverage:
+ - Tests that read and write usages upon the same texture subresource, or different subresources
+   of the same texture. Different subresources of the same texture includes different mip levels,
+   different array layers, and different aspects.
+   - When read and write usages are binding to the same texture subresource, an error should be
+     generated. Otherwise, no error should be generated.
+`;
+
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { ValidationTest } from '../validation_test.js';
+
+const READ_BASE_LEVEL = 3;
+const READ_BASE_LAYER = 0;
+
+class F extends ValidationTest {
+  createTexture(
+    options: {
+      width?: number;
+      height?: number;
+      arrayLayerCount?: number;
+      mipLevelCount?: number;
+      sampleCount?: number;
+      format?: GPUTextureFormat;
+    } = {}
+  ): GPUTexture {
+    const {
+      width = 32,
+      height = 32,
+      arrayLayerCount = 1,
+      mipLevelCount = 1,
+      sampleCount = 1,
+      format = 'rgba8unorm',
+    } = options;
+
+    return this.device.createTexture({
+      size: { width, height, depth: arrayLayerCount },
+      mipLevelCount,
+      sampleCount,
+      dimension: '2d',
+      format,
+      usage: GPUTextureUsage.OUTPUT_ATTACHMENT | GPUTextureUsage.SAMPLED,
+    });
+  }
+
+  getDescriptor(
+    options: {
+      format?: GPUTextureFormat;
+      dimension?: GPUTextureViewDimension;
+      aspect?: GPUTextureAspect;
+      baseMipLevel?: number;
+      mipLevelCount?: number;
+      baseArrayLayer?: number;
+      arrayLayerCount?: number;
+    } = {}
+  ): GPUTextureViewDescriptor {
+    const {
+      format = 'rgba8unorm',
+      dimension = '2d',
+      aspect = 'all',
+      baseMipLevel = 0,
+      mipLevelCount = 1,
+      baseArrayLayer = 0,
+      arrayLayerCount = 1,
+    } = options;
+    return {
+      format,
+      dimension,
+      aspect,
+      baseMipLevel,
+      mipLevelCount,
+      baseArrayLayer,
+      arrayLayerCount,
+    };
+  }
+}
+
+export const g = makeTestGroup(F);
+
+g.test('readwrite_on_different_types_of_subresources')
+  .params([
+    // read and write usages are binding to the same texture subresource.
+    {
+      format: 'rgba8unorm' as const,
+      writeBaseLevel: READ_BASE_LEVEL,
+      writeBaseLayer: READ_BASE_LAYER,
+      writeAspect: 'all' as const,
+      success: false,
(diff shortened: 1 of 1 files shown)


Richard-Yunchao:
PTAL. If this one looks good, I will add the rest of tests for texture subresource usages validation in render pass (tests for compute will be after that).

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
```diff
@@ -0,0 +1,179 @@
...
+   different array layers, and different aspects.
+   - When read and write usages are binding to the same texture subresource, an error should be
+     generated. Otherwise, no error should be generated.
+`;
+
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { ValidationTest } from '../validation_test.js';
+
+const READ_BASE_LEVEL = 3;
+const READ_BASE_LAYER = 0;
```
kainino0x, line 16: nit: I'd move these down so they're closer to where they're used

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
```diff
@@ -0,0 +1,179 @@
...
+      writeBaseLayer: READ_BASE_LAYER,
+      writeAspect: 'stencil-only' as const,
+      success: true,
+    },
+  ])
+  .fn(async t => {
+    const { format, writeBaseLevel, writeBaseLayer, writeAspect, success } = t.params;
+
+    const texture = t.createTexture({ arrayLayerCount: 2, mipLevelCount: 6, format });
+    const readAspect = format === 'depth24plus-stencil8' ? 'depth-only' : 'all';
```
kainino0x, line 124: nit: Pass this in through .params() as a private parameter (`_readAspect`) so you don't have to encode this switch in the test code. You can have it default to 'all' so it's only specified for the one test case that needs it.

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
```diff
@@ -0,0 +1,179 @@
...
+    },
+
+    // read and write usages are binding to different aspects of the same texture.
+    {
+      format: 'depth24plus-stencil8' as const,
+      writeBaseLevel: READ_BASE_LEVEL,
+      writeBaseLayer: READ_BASE_LAYER,
+      writeAspect: 'stencil-only' as const,
+      success: true,
+    },
```
kainino0x, line 118: Need to more thoroughly test the depth/stencil cases:
- DS format, both 'all'
- DS format, both 'depth'
- DS format, both 'stencil'
- DS format, read 'stencil', write 'depth'
- DS format, read 'depth', write 'stencil'
- DS format, read 'all', write 'depth'
- DS format, read 'all', write 'stencil'
- DS format, read 'stencil', write 'all'
- DS format, read 'depth', write 'all'
- S format, both 'all'
- S format, both 'stencil'
- S format, read 'stencil', write 'all'
- S format, read 'all', write 'stencil'
- D format, both 'all'
- D format, both 'stencil'
- D format, read 'stencil', write 'all'
- D format, read 'all', write 'stencil'

i.e. something like:

```ts
params()
  .combine(poptions('format', [ds, s, d]))
  .combine(poptions('readAspect', ['all', 'depth-only', 'stencil-only']))
  .combine(poptions('writeAspect', ['all', 'depth-only', 'stencil-only']))
  .unless(format doesn't have depth and either aspect is 'depth-only')
  .unless(format doesn't have stencil and either aspect is 'stencil-only')
```

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
```diff
@@ -0,0 +1,179 @@
...
+      size: { width, height, depth: arrayLayerCount },
+      mipLevelCount,
+      sampleCount,
+      dimension: '2d',
+      format,
+      usage: GPUTextureUsage.OUTPUT_ATTACHMENT | GPUTextureUsage.SAMPLED,
+    });
+  }
+
+  getDescriptor(
```
kainino0x, line 48: This seems like unnecessary boilerplate; does this really simplify vs having it inline? We've been adding things like this to a number of tests but it just means we have a lot of slightly different copies of this code all over the place.

createTexture also seems possibly unnecessary (but less so), so maybe consider whether it makes sense to inline it too.

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
Richard-Yunchao: I assume "S format" you mentioned means stencil-only format. It looks like there is no stencil-only format. The format we can support for depth/stencil are     
    "depth32float",
    "depth24plus",
    "depth24plus-stencil8", 
I suppose the first two are depth-only. The last one is depth-stencil. So there is no stencil-only. WDYT? @kainino0x

kainino0x: right, sorry, I forgot - we decided to add one in gpuweb/gpuweb#306 but it's not done yet. Since we will add it, let's make sure to set up the test so it can handle adding an additional format to the list. Either add TODO to test it once it's in, or parameterize it with a const array defined in capability_info that lists all the D/S/DS formats.

OK to make it a different test as you see fit.

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
Richard-Yunchao: Not needed after separating tests.

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
Richard-Yunchao: Done

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
Richard-Yunchao: Done. Added a separate test for depth/stencil, add added a TODO for stencil-only format

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
Richard-Yunchao: Done. I removed getDescriptor for createView(). createTexture() might be used a lot in this test, so I didn't remove it and we can reuse the parameters in upcoming tests.

Richard-Yunchao:
Thanks! @kainino0x

PR #253, opened 2020-07-31
Add test for different read write usages combinations

Richard-Yunchao:
Texture read usages include sampled-texture and readonly storage
texture. Texture write usages include writeonly storage texture
and render target. Different usages upon the same subresource
should follow this rule:
- multiple read usages upon the same subresource is valid.
- multiple write usages, or read/write usages upon the same
  subresource is invalid, with one exception that multiple
  writeonly storage texture usages are valid.

---
This change is [Reviewable](https://reviewable.io/reviews/gpuweb/cts/253)

### DIFF
--- a/src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
+++ b/src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
@@ -3,14 +3,20 @@ Texture Usages Validation Tests in Render Pass.
 
 Test Coverage:
 
-  - Test the combination of different pairs of usages upon the same texture subresource or different
-    subresources of the same texture. Different subresources of the same texture includes different
-    mip levels, different array layers, and different aspects.
-    - When read-write or write-write usages are binding to the same texture subresource, an error
-      should be generated. Otherwise, no error should be generated. One exception is race condition
-      upon two writeonly-storage-texture usages, which is valid.
+  - For each combination of two texture usages:
+    - For various subresource ranges (different mip levels or array layers) that overlap a given
+      subresources or not for color formats:
+      - Check that an error is generated when read-write or write-write usages are binding to the
+        same texture subresource. Otherwise, no error should be generated. One exception is race
+        condition upon two writeonly-storage-texture usages, which is valid.
 
-  - Test different shader stages.
+  - For each combination of two texture usages:
+    - For various aspects (all, depth-only, stencil-only) that overlap a given subresources or not
+      for depth/stencil formats:
+      - Check that an error is generated when read-write or write-write usages are binding to the
+        same aspect. Otherwise, no error should be generated.
+
+  - Test combinations of two shader stages:
     - Texture usages in bindings with invisible shader stages should be tracked. Invisible shader
       stages include shader stage with visibility none and compute shader stage in render pass.
 `;
@@ -312,10 +318,10 @@ g.test('subresources_and_binding_types_combination_for_aspect')
     pass.setBindGroup(0, bindGroup);
     pass.endPass();
 
-    const resourceSuccess =
+    const disjointAspects =
       (aspect0 === 'depth-only' && aspect1 === 'stencil-only') ||
       (aspect0 === 'stencil-only' && aspect1 === 'depth-only');
-    const success = resourceSuccess || _usageSuccess;
+    const success = disjointAspects || _usageSuccess;
 
     t.expectValidationError(() => {
       encoder.finish();
--- a/src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
+++ b/src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
@@ -232,3 +232,94 @@ g.test('shader_stages_and_visibility')
       encoder.finish();
     });
   });
+
+g.test('binding_types_combination')
+  .params([
+    { type0: 'sampled-texture' as const, type1: 'sampled-texture' as const, _success: true },
+    {
+      type0: 'sampled-texture' as const,
+      type1: 'readonly-storage-texture' as const,
+      _success: true,
+    },
+    {
+      type0: 'sampled-texture' as const,
+      type1: 'writeonly-storage-texture' as const,
+      _success: false,
+    },
+    { type0: 'sampled-texture' as const, type1: 'render-target' as const, _success: false },
+    {
+      type0: 'readonly-storage-texture' as const,
+      type1: 'readonly-storage-texture' as const,
+      _success: true,
+    },
+    {
+      type0: 'readonly-storage-texture' as const,
+      type1: 'writeonly-storage-texture' as const,
+      _success: false,
+    },
+    {
+      type0: 'readonly-storage-texture' as const,
+      type1: 'render-target' as const,
+      _success: false,
+    },
+    {
+      type0: 'writeonly-storage-texture' as const,
+      type1: 'writeonly-storage-texture' as const,
+      _success: true,
+    },
+    {
+      type0: 'writeonly-storage-texture' as const,
+      type1: 'render-target' as const,
+      _success: false,
+    },
+  ])
+  .fn(async t => {
+    const { type0, type1, _success } = t.params;
+
+    const view = t
+      .createTexture({
+        usage:
+          GPUTextureUsage.SAMPLED | GPUTextureUsage.STORAGE | GPUTextureUsage.OUTPUT_ATTACHMENT,
+      })
+      .createView();
+
+    const bglEntries: GPUBindGroupLayoutEntry[] = [
+      {
+        binding: 0,
+        visibility: GPUShaderStage.FRAGMENT,
+        type: type0,
+        storageTextureFormat: type0 === 'sampled-texture' ? undefined : 'rgba8unorm',
+      },
+    ];
+    const bgEntries: GPUBindGroupEntry[] = [{ binding: 0, resource: view }];
+    if (type1 !== 'render-target') {
+      bglEntries.push({
+        binding: 1,
+        visibility: GPUShaderStage.FRAGMENT,
+        type: type1,
+        storageTextureFormat: type1 === 'sampled-texture' ? undefined : 'rgba8unorm',
+      });
+      bgEntries.push({ binding: 1, resource: view });
+    }
+    const bindGroup = t.device.createBindGroup({
+      entries: bgEntries,
+      layout: t.device.createBindGroupLayout({ entries: bglEntries }),
+    });
+
+    const encoder = t.device.createCommandEncoder();
+    const pass = encoder.beginRenderPass({
+      colorAttachments: [
+        {
+          attachment: type1 === 'render-target' ? view : t.createTexture().createView(),
+          loadValue: { r: 0.0, g: 1.0, b: 0.0, a: 1.0 },
+          storeOp: 'store',
+        },
+      ],
+    });
+    pass.setBindGroup(0, bindGroup);
+    pass.endPass();
+
+    t.expectValidationError(() => {
+      encoder.finish();
+    }, !_success);
+  });
--- a/src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
+++ b/src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
@@ -105,52 +105,52 @@ g.test('subresources_and_binding_types_combination_for_color')
       ])
       .combine([
         {
-          type0: 'sampled-texture' as const,
-          type1: 'sampled-texture' as const,
+          type0: 'sampled-texture',
+          type1: 'sampled-texture',
           _usageSuccess: true,
         },
         {
-          type0: 'sampled-texture' as const,
-          type1: 'readonly-storage-texture' as const,
+          type0: 'sampled-texture',
+          type1: 'readonly-storage-texture',
           _usageSuccess: true,
         },
         {
-          type0: 'sampled-texture' as const,
-          type1: 'writeonly-storage-texture' as const,
+          type0: 'sampled-texture',
+          type1: 'writeonly-storage-texture',
           _usageSuccess: false,
         },
         {
-          type0: 'sampled-texture' as const,
-          type1: 'render-target' as const,
+          type0: 'sampled-texture',
+          type1: 'render-target',
           _usageSuccess: false,
         },
         {
-          type0: 'readonly-storage-texture' as const,
-          type1: 'readonly-storage-texture' as const,
+          type0: 'readonly-storage-texture',
+          type1: 'readonly-storage-texture',
           _usageSuccess: true,
         },
         {
-          type0: 'readonly-storage-texture' as const,
-          type1: 'writeonly-storage-texture' as const,
+          type0: 'readonly-storage-texture',
+          type1: 'writeonly-storage-texture',
           _usageSuccess: false,
         },
         {
-          type0: 'readonly-storage-texture' as const,
-          type1: 'render-target' as const,
+          type0: 'readonly-storage-texture',
+          type1: 'render-target',
           _usageSuccess: false,
         },
         // Race condition upon multiple writable storage texture is valid.
         {
-          type0: 'writeonly-storage-texture' as const,
-          type1: 'writeonly-storage-texture' as const,
+          type0: 'writeonly-storage-texture',
+          type1: 'writeonly-storage-texture',
           _usageSuccess: true,
         },
         {
-          type0: 'writeonly-storage-texture' as const,
-          type1: 'render-target' as const,
+          type0: 'writeonly-storage-texture',
+          type1: 'render-target',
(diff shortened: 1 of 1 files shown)


Richard-Yunchao:
PTAL. Thank you!

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
```diff
@@ -173,3 +175,97 @@ g.test('readwrite_upon_aspects')
       encoder.finish();
     }, !success);
   });
+
+g.test('binding_types_combination')
```
Kangz, line 179: Maybe this should be folded in the test above and do the following combinations:

 - For each pair of usages:
 - For each combination of subresources (same, different mip, different layer, single subresource+entire conflicting layer, single subresource + entire conflicting mip)

Then the test body would get metadata on each usage (readonly vs. not) and decide whether success should be true.

Also we should be ready to factor out some of the logic to common helper.

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
```diff
@@ -173,3 +175,97 @@ g.test('readwrite_upon_aspects')
...
+
+    const bglEntries: GPUBindGroupLayoutEntry[] = [
+      {
+        binding: 0,
+        visibility: GPUShaderStage.FRAGMENT,
+        type: type0,
+        // Currently setting two 'render-target' usages in the same render pass is impossible.
+        // We don't need to test that. So type0 is either 'sampled-texture' or storage usages.
+        // It can't be 'render-target'.
+        storageTextureFormat: type0 === 'sampled-texture' ? undefined : 'rgba8unorm',
```
Kangz, line 237: This is an example where setting information on the bind group layout entry would have been useful even if it isn't used by this `GPUBindingType`.

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
Richard-Yunchao: Corentin, My original proposal about the test plan was something like this: a super big combinations with every possible factors: different pairs of read/write bindings like the new added test, different combinations of subresources like the tests we have already added, different shader stages and visibility (another new patch), different sync scope (inside a draw calls, different draw calls but inside a render pass, different passes), etc. And combine all of these factor's together. That original test plan was really all-encompassing like your suggestion here and went even further. But it seems that is not necessary and the logic might be complex and hard to understand. So Kai and I simplified the combinations.   

However, single subresource + entire conflicting layers, single subresource + entire conflicting mip might be good to being added into the tests. And maybe we can add a few more like single subresource + entire resource (all layers, all mip levels), a single layer + entire resource, a single mip chain + entire resource. You see, it's not easy to draw a line and say "that's enough". 

WDYT, @kainino0x and @austinEng?

kainino0x: I do think it's likely easier to read and write with the tests split this way. But it depends on whether we think it's necessary to test the combinations that we're missing here. I am having a hard time figuring out exactly what we're missing / deciding whether we need it.

The main thing we need to ensure is that the subresource usage tracking is working properly. That means, in each place that a subresource usage has to be tracked, we need to make sure (1) it is tracked (2) the subresource "address" (level, layer, aspect) and usage (at least, whether it's R, W, or RW) is passed into the tracking correctly.

Perhaps we don't need to cover the full combination of (different ways to read and/or write a subresource) x (different ways to address disjoint/non-disjoint ranges of subresources). But it would be good to test at least (different ways to write a subresource) x (different ways to address different subresources). And by the time you reach that it might not be hard to test more.

There's also the issue of depth/stencil render targets, and aspects. But looking at the code you have already written now, it's really not that different from the others.

I think it's possible to get complete testing here while maintaining a lot of the straightforwardness that we have now. The hard parts will probably be (1) expressing the combinatorics (maybe can be expressed as a sum of one combination for color + one combination for depth/stencil) and (2) keeping the success expectation computation simple (I think this can be done with good helper tables and methods, e.g. a helper to compute whether two subresource ranges overlap, a helper to compute whether two usages conflict, and some data in tables to tell us what binding types read, write, or read-write).

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
```diff
@@ -173,3 +175,97 @@ g.test('readwrite_upon_aspects')
...
+      .createView();
+
+    const bglEntries: GPUBindGroupLayoutEntry[] = [
+      {
+        binding: 0,
+        visibility: GPUShaderStage.FRAGMENT,
+        type: type0,
+        // Currently setting two 'render-target' usages in the same render pass is impossible.
+        // We don't need to test that. So type0 is either 'sampled-texture' or storage usages.
+        // It can't be 'render-target'.
```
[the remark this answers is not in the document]
Richard-Yunchao: The reason that I said "setting two render-target usages in the same render pass is impossible" right now is that we haven't supported multiple render targets. But I revisited the spec, it looks like we have multiple render targets: required sequence<GPURenderPassColorAttachmentDescriptor> colorAttachments in GPURenderPassDescriptor. So, Yeah, it should be tested.

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
Richard-Yunchao: As Corentin and Kai suggestion, I added a combination of subresources for color (same, different mip levels, different array layers, single subresource + entire mip chain, single subresource + entire array layer) and pairs of binding types, and another combination of subresources for  aspects and pairs of binding types. Note that for aspect combination, because depth/stencil texture formats can't be used as storage bindings, so the pairs of binding types are only sampled-texture and render-target.

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
Richard-Yunchao: What are you talking about, it seems that it is related to another issue? :)

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
```diff
@@ -173,3 +175,97 @@ g.test('readwrite_upon_aspects')
       encoder.finish();
     }, !success);
   });
+
+g.test('binding_types_combination')
+  .params([
+    { type0: 'sampled-texture' as const, type1: 'sampled-texture' as const, _success: true },
```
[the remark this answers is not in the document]

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
```diff
@@ -173,3 +175,97 @@ g.test('readwrite_upon_aspects')
       encoder.finish();
     }, !success);
   });
+
+g.test('binding_types_combination')
+  .params([
```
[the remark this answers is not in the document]
Richard-Yunchao: Hmm. It might be not important right now. In addition, the first type can be "render-target" too and I suppose the expected test result should be a failure. I will add that test later (I have already added a TODO)

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
```diff
@@ -173,3 +175,97 @@ g.test('readwrite_upon_aspects')
...
+    },
+    {
+      type0: 'readonly-storage-texture' as const,
+      type1: 'render-target' as const,
+      _success: false,
+    },
+    {
+      type0: 'writeonly-storage-texture' as const,
+      type1: 'writeonly-storage-texture' as const,
+      _success: true,
```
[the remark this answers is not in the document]

Richard-Yunchao:
Sorry for the late response, coz I took a few ALs since last Friday (I expected it should be 1-day AL, but...) for some personal issues. PTAL when you have time. Thank you. @kainino0x , @Kangz and @austinEng.

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
kainino0x: oops, sorry for the lack of context. Part of the discussion: https://github.com/gpuweb/gpuweb/issues/829, https://github.com/gpuweb/gpuweb/pull/949#issuecomment-664365066
I think there was more elsewhere, but I don't remember where (maybe in a meeting).

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
```diff
@@ -23,6 +23,7 @@ class TextureUsageTracking extends ValidationTest {
       mipLevelCount?: number;
```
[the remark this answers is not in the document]
Richard-Yunchao: There are quite a lot usages of createTexture. And I expect there will be more calls in the upcoming tests. And how to create texture with these parameters are not important for most test cases. So I'd like to use a helper.

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
Richard-Yunchao: Thank you, Kai. This is useful in CTS because it is easier for us to iterate over each binding type. But I suppose we may rarely use it in real application because we know exactly the particular binding type in most cases.

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
```diff
@@ -115,42 +230,65 @@ g.test('readwrite_upon_subresources')
...
       encoder.finish();
-    }, !_success);
+    }, !success);
   });
 
-g.test('readwrite_upon_aspects')
+g.test('subresources_and_binding_types_combination_for_aspect')
   .params(
     params()
       .combine(poptions('format', ['depth32float', 'depth24plus', 'depth24plus-stencil8'] as const))
```
Kangz, line 242: Maybe this shouldn't be a hardcoded list of formats?

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
```diff
@@ -160,17 +298,25 @@ g.test('readwrite_upon_aspects')
...
+              depthStoreOp: 'clear',
+              depthLoadValue: 'load',
+              stencilStoreOp: 'clear',
+              stencilLoadValue: 'load',
+            },
     });
     pass.setBindGroup(0, bindGroup);
     pass.endPass();
 
+    const resourceSuccess =
```
Kangz, line 315: nit: disjointAspects

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
```diff
@@ -2,11 +2,17 @@ export const description = `
 Texture Usages Validation Tests in Render Pass.
 
 Test Coverage:
- - Tests that read and write usages upon the same texture subresource, or different subresources
-   of the same texture. Different subresources of the same texture includes different mip levels,
-   different array layers, and different aspects.
-   - When read and write usages are binding to the same texture subresource, an error should be
-     generated. Otherwise, no error should be generated.
+
+  - Test the combination of different pairs of usages upon the same texture subresource or different
```
Kangz, line 6: Maybe use less prose for the test plan:

```
 - For each combination of two texture usages:
   - For various subresource ranges that overlap a given subresources or not (detail of what we test):
      - Check that an error is generated ...

 Something about the depth stencil test with aspects?

 Tests for shader stages. (and maybe have a test for invisible shader stages in the compute pass?)
```

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
```diff
@@ -49,64 +55,173 @@ class TextureUsageTracking extends ValidationTest {
...
+const TOTAL_LAYERS = 2;
 
-    const texture = t.createTexture({ arrayLayerCount: 2, mipLevelCount: 6 });
+g.test('subresources_and_binding_types_combination_for_color')
+  .params(
+    params()
+      .combine([
+        // Two texture usages are binding to the same texture subresource.
+        {
+          baseLevel: BASE_LEVEL,
```
Kangz, line 69: @kainino0x I'm not sure how we want to do things in the CTS, maybe we could compute the success value in the test directly? Also maybe the list of ranges to test could be stored as a helper list and reused in several places? Finally it might be useful to test multiple ranges for both view0 and view1. (this comment doesn't need to be addressed in this PR)

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
Richard-Yunchao: I didn't see a small and simple change for this. Maybe we can split all texture formats (kTextureFormatInfo) into color formats (kColorTextureFormatInfo) and depth/stencil formats (kDepthStencilTextureFormatInfo) in capability_info.ts? @Kangz and @kainino0x , WDYT?

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
Richard-Yunchao: Done

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
Richard-Yunchao: Done. Testing for compute will come later, maybe in a separate test file.

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
Richard-Yunchao: If we limit the computation to fulfill the requirement of this test, say only two views. It is not difficult. But if we need a common function to trace the resource overlap for various situations, it might be very complicated (like the impl in Dawn project). 

BTW, do you think it is necessary to test more cases about resource conflict, it is not difficult to add them, for example:
1) same layer: a few mip levels overlap or not with another few mip levels: layer = 0: view0_mip = [0, 1, 2], view1_mip = [3, 4] / view1_mip = [2, 3]. The current test we have only test view0 with one single subresource. 
2) same level: a few array layers overlap or not with another few layers
3) a few mip levels within the same layer overlap or not with a few array layers within the same level 
4) a few subresources (multiple levels * multiple layers) overlap or not with another few subresource (multiple levels * multiple layers).   

If you think it is necessary for cover resource subresource overlap, I can add it (maybe in another pr?). 

Looks like subresource overlap doesn't interact a lot with read-write usage conflict and their combinations might be unnecessary. But doing more tests in CTS is not a bad thing, per my understanding.

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
kainino0x: I've been wanting to do this for a while and with compressed formats it became more pressing. Now being done in #263!

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
```diff
@@ -49,64 +61,173 @@ class TextureUsageTracking extends ValidationTest {
...
+          type0: 'sampled-texture',
+          type1: 'render-target',
+          _usageSuccess: false,
+        },
+        {
+          type0: 'readonly-storage-texture',
+          type1: 'readonly-storage-texture',
+          _usageSuccess: true,
+        },
+        {
```
Kangz, line 144: Based on the rules in #994 this will actually be allowed, which I think makes sense.

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
Kangz: Yeah it could be worth having a test that's more detailed in how the various resource ranges overlap, and this test can stay mostly as is.

Richard-Yunchao:
Can we merge this now? @Kangz and @kainino0x ? Using kDepthStencilTextureFormatInfo and adding more texture subresource overlap tests can be a follow-up.

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
Richard-Yunchao: Thanks!

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
kainino0x: > I'm not sure how we want to do things in the CTS, maybe we could compute the success value in the test directly?

We have a lot of tests that go one way or the other. Usually we compute the success value in the test if the parameterizations are generated; otherwise it's generally not that hard to write them manually.

For some validation tests I've contemplated suggesting that we write a full TS reimplementation of the full validation rules from the spec, so you can throw arbitrary test cases at it and compare them with the browser implementation, but I'm never sure whether it's a good idea.

Agree it makes sense to have more detailed tests on resource range overlaps be separate from these.

PR #254, opened 2020-07-31
Add subresource usage validation tests for different shader stages

Richard-Yunchao:
Texture subresource usage validation should be done on all shader
stages, including stages that are not visible in current pass.
For example, stage 0 and stage compute in render pass.

### DIFF
--- a/src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
+++ b/src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
@@ -11,11 +11,7 @@ Test Coverage:
 
 import { poptions, params } from '../../../../common/framework/params_builder.js';
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
-import {
-  kTextureFormatInfo,
-  kShaderStages,
-  kShaderStageCombinations,
-} from '../../../capability_info.js';
+import { kTextureFormatInfo, kShaderStages } from '../../../capability_info.js';
 import { ValidationTest } from '../validation_test.js';
 
 class TextureUsageTracking extends ValidationTest {
@@ -183,7 +179,7 @@ g.test('readwrite_upon_aspects')
 g.test('shader_stages_and_visibility')
   .params(
     params()
-      .combine(poptions('readVisibility', kShaderStageCombinations))
+      .combine(poptions('readVisibility', [0, ...kShaderStages]))
       .combine(poptions('writeVisibility', [0, ...kShaderStages]))
   )
   .fn(async t => {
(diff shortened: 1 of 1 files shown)


Richard-Yunchao:
PTAL. Thank you. Do you prefer to combine this one and https://github.com/gpuweb/cts/pull/253 and review them altogether? a combined larger one vs a few separated medium-size patches. I'd like to know your preference.

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
```diff
@@ -173,3 +179,59 @@ g.test('readwrite_upon_aspects')
...
+          attachment: writeHasVertexStage ? view : t.createTexture().createView(),
+          loadValue: { r: 0.0, g: 1.0, b: 0.0, a: 1.0 },
+          storeOp: 'store',
+        },
+      ],
+    });
+    pass.setBindGroup(0, bindGroup);
+    pass.endPass();
+
+    // Texture usages in bindings with invisible shader stages should be tracked.
```
austinEng, line 233: nit: this test isn't super clear about what is meant be "invisible shader stages". Maybe we should say "with visibility none" ?

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
```diff
@@ -173,3 +179,59 @@ g.test('readwrite_upon_aspects')
...
+      .combine(poptions('readVisibility', kShaderStageCombinations))
+      .combine(poptions('writeVisibility', [0, ...kShaderStages]))
+  )
+  .fn(async t => {
+    const { readVisibility, writeVisibility } = t.params;
+
+    // writeonly-storage-texture binding type is not supported in vertex stage. So, this test
+    // uses writeonly-storage-texture binding as writable binding upon the same subresource if
+    // vertex stage is not included. Otherwise, it uses output attachment instead.
+    const writeHasVertexStage = writeVisibility & GPUShaderStage.VERTEX;
```
austinEng, line 195: Can we just have a boolean then?

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
Richard-Yunchao: Yeah. I explained a little bit in the comment about invisible shader stages.

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
```diff
@@ -173,3 +179,60 @@ g.test('readwrite_upon_aspects')
       encoder.finish();
     }, !success);
   });
+
+g.test('shader_stages_and_visibility')
+  .params(
+    params()
+      .combine(poptions('readVisibility', kShaderStageCombinations))
```
kainino0x, line 186: optional nit: Perhaps no need to test all 8x4 combinations, and just [0, kShaderStages] would be fine here too.

Richard-Yunchao: Agreed. Done!

PR #255, opened 2020-08-03
Copy between linear data and texture validation tests

ponitka:
This is still work in progress with some some details missing, but it would be great to have some feedback on the general structure of those tests. It covers everything for writeTexture, most of the tests are being run on copyB2T and copyT2B too.

@kainino0x : Can you take a look, please?

PR #256, opened 2020-08-07
Add validation tests for CopyBufferToBuffer

Jiawei-Shao:
This patch adds all the validation tests about CopyBufferToBuffer().
1. Both the source and destination buffer must be valid.
2. The source buffer must have COPY_SRC usage and the destination
   buffer must have COPY_DST usage.
3. The copySize must be a multiple of 4.
4. Both the source and destiantion offset are multiple of 4.
5. Both (srcOffset + copySize) and (dstOffset + copySize) must not
   overflow a buffer.
6. The source buffer and destination buffer cannot be the same one.


Jiawei-Shao:
@kainino0x PTAL, thanks!

src/webgpu/api/validation/copyBufferToBuffer.spec.ts
```diff
@@ -0,0 +1,263 @@
...
+    { srcOffset: 0, dstOffset: 0, isSuccess: true },
+    { srcOffset: 2, dstOffset: 0, isSuccess: false },
+    { srcOffset: 4, dstOffset: 0, isSuccess: true },
+    { srcOffset: 5, dstOffset: 0, isSuccess: false },
+    { srcOffset: 8, dstOffset: 0, isSuccess: true },
+    { srcOffset: 0, dstOffset: 2, isSuccess: false },
+    { srcOffset: 0, dstOffset: 4, isSuccess: true },
+    { srcOffset: 0, dstOffset: 5, isSuccess: false },
+    { srcOffset: 0, dstOffset: 8, isSuccess: true },
+    { srcOffset: 4, dstOffset: 8, isSuccess: true },
```
kainino0x, line 149: nit:
```suggestion
    { srcOffset: 4, dstOffset: 4, isSuccess: true },
```
just to make sure dstOffset allows the smallest allowed alignment

src/webgpu/api/validation/copyBufferToBuffer.spec.ts
```diff
@@ -0,0 +1,263 @@
...
+      isSuccess,
+    });
+  });
+
+g.test('copy_overflow')
+  .params([
+    { srcOffset: 16, dstOffset: 0, copySize: MAX_ALIGNED_SAFE_INTEGER },
+    { srcOffset: 0, dstOffset: 16, copySize: MAX_ALIGNED_SAFE_INTEGER },
+    { srcOffset: MAX_ALIGNED_SAFE_INTEGER, dstOffset: 0, copySize: 16 },
+    { srcOffset: 0, dstOffset: MAX_ALIGNED_SAFE_INTEGER, copySize: 16 },
```
kainino0x, line 181: Test the base case (0, 0), and cases where MAX_ALIGNED_SAFE_INTEGER might get added to itself.

```suggestion
    { srcOffset: 0, dstOffset: 0, copySize: MAX_ALIGNED_SAFE_INTEGER },
    { srcOffset: 16, dstOffset: 0, copySize: MAX_ALIGNED_SAFE_INTEGER },
    { srcOffset: 0, dstOffset: 16, copySize: MAX_ALIGNED_SAFE_INTEGER },
    { srcOffset: MAX_ALIGNED_SAFE_INTEGER, dstOffset: 0, copySize: 16 },
    { srcOffset: 0, dstOffset: MAX_ALIGNED_SAFE_INTEGER, copySize: 16 },
    { srcOffset: MAX_ALIGNED_SAFE_INTEGER, dstOffset: 0, copySize: MAX_ALIGNED_SAFE_INTEGER },
    { srcOffset: 0, dstOffset: MAX_ALIGNED_SAFE_INTEGER, copySize: MAX_ALIGNED_SAFE_INTEGER },
```

src/webgpu/api/validation/copyBufferToBuffer.spec.ts
```diff
@@ -0,0 +1,263 @@
...
+      srcOffset,
+      dstBuffer,
+      dstOffset,
+      copySize,
+      isSuccess: false,
+    });
+  });
+
+g.test('copy_out_of_bounds')
+  .params([
```
kainino0x, line 207: Would be best to have a base case which does succeed, to be more sure the rest of the test cases are testing what they intend to test.
```suggestion
  .params([
    { srcOffset: 0, dstOffset: 0, copySize: 32, _success: true },
```

src/webgpu/api/validation/copyBufferToBuffer.spec.ts
```diff
@@ -0,0 +1,263 @@
...
+    { srcOffset: 0, dstOffset: 0, copySize: 36 },
+    { srcOffset: 36, dstOffset: 0, copySize: 4 },
+    { srcOffset: 0, dstOffset: 36, copySize: 4 },
+    { srcOffset: 36, dstOffset: 0, copySize: 0 },
+    { srcOffset: 0, dstOffset: 36, copySize: 0 },
+    { srcOffset: 20, dstOffset: 0, copySize: 16 },
+    { srcOffset: 0, dstOffset: 20, copySize: 16 },
+  ] as const)
+  .fn(async t => {
+    const { srcOffset, dstOffset, copySize } = t.params;
```
kainino0x, line 217:
```suggestion
    const { srcOffset, dstOffset, copySize, _success = false } = t.params;
```

src/webgpu/api/validation/copyBufferToBuffer.spec.ts
```diff
@@ -0,0 +1,263 @@
...
+      usage: GPUBufferUsage.COPY_DST,
+    });
+
+    t.TestCopyBufferToBuffer({
+      srcBuffer,
+      srcOffset,
+      dstBuffer,
+      dstOffset,
+      copySize,
+      isSuccess: false,
```
kainino0x, line 235:
```suggestion
      isSuccess: _success,
```

src/webgpu/api/validation/copyBufferToBuffer.spec.ts
```diff
@@ -0,0 +1,263 @@
...
+      srcOffset,
+      dstBuffer,
+      dstOffset,
+      copySize,
+      isSuccess,
+    });
+  });
+
+g.test('copy_overflow')
+  .params([
```
kainino0x, line 177: Depending on whether IDL allows it, we should also test larger numbers here (like kFirstNumberBefore2_64) - see thread: https://github.com/gpuweb/cts/pull/255#discussion_r467330003

src/webgpu/api/validation/copyBufferToBuffer.spec.ts
```diff
@@ -0,0 +1,263 @@
+export const description = `
+copyBufferToBuffer tests.
```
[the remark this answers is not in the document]
Jiawei-Shao: Done

src/webgpu/api/validation/copyBufferToBuffer.spec.ts
Jiawei-Shao: Done

src/webgpu/api/validation/copyBufferToBuffer.spec.ts
Jiawei-Shao: Sorry I am not quite clear about this. Do you mean we should use 2 ** 31 in this test like what we are doing in the test  "offset_plus_required_bytes_in_copy_overflow" ?

src/webgpu/api/validation/copyBufferToBuffer.spec.ts
```diff
@@ -0,0 +1,287 @@
...
+  const bufferSize = 16;
+
+  const validBuffer = t.device.createBuffer({
+    size: bufferSize,
+    usage: GPUBufferUsage.COPY_SRC | GPUBufferUsage.COPY_DST,
+  });
+
+  const errorBuffer = t.getErrorBuffer();
+
+  const srcOffset = 0;
```
Kangz, line 63: IMHO these constants don't help with readability because when looking at a `TestCopyBufferToBuffer` call we need to look up in the function to know what the parameters are. Maybe they could be inlined in the `TestCopyBufferToBuffer` calls instead. Ditto for other tests.

src/webgpu/api/validation/copyBufferToBuffer.spec.ts
```diff
@@ -0,0 +1,287 @@
...
+    dstBuffer: errorBuffer,
+    dstOffset,
+    copySize,
+    isSuccess: false,
+  });
+});
+
+g.test('buffer_usage')
+  .params([
+    { srcUsage: GPUBufferUsage.COPY_SRC, dstUsage: GPUBufferUsage.COPY_DST, _isSuccess: true },
```
Kangz, line 88: nit: maybe we could have parametrization overall all buffer usages?

src/webgpu/api/validation/copyBufferToBuffer.spec.ts
```diff
@@ -0,0 +1,287 @@
...
+      srcBuffer,
+      srcOffset,
+      dstBuffer,
+      dstOffset,
+      copySize,
+      isSuccess,
+    });
+  });
+
+g.test('copy_size')
```
Kangz, line 119: nit: copy_size_alignment?

src/webgpu/api/validation/copyBufferToBuffer.spec.ts
```diff
@@ -0,0 +1,287 @@
...
+      srcBuffer,
+      srcOffset,
+      dstBuffer,
+      dstOffset,
+      copySize,
+      isSuccess,
+    });
+  });
+
+g.test('copy_offset')
```
Kangz, line 153: nit: copy_offset_alignment?

src/webgpu/api/validation/copyBufferToBuffer.spec.ts
```diff
@@ -0,0 +1,287 @@
...
+      dstBuffer,
+      dstOffset,
+      copySize,
+      isSuccess: false,
+    });
+  });
+
+g.test('copy_out_of_bounds')
+  .params([
+    { srcOffset: 0, dstOffset: 0, copySize: 32, _isSuccess: true },
```
Kangz, line 231: Also it could be helpful to have other cases that succeeed, where the offset isn't 0, to test exactly on both side of the limit?

src/webgpu/api/validation/copyBufferToBuffer.spec.ts
kainino0x: I mean, numbers larger than MAX_ALIGNED_SAFE_INTEGER (which is ~`2**53`). You can copy a definition of kFirstNumberBefore2_64 from that comment and use it here. This lets us have cases where offset+size > `2**64`.

kainino0x: Okay, @ponitka found that that's actually not valid (see that comment thread). So closing this thread.

PR #258, opened 2020-08-11
Add rendering tests for primitive topology.

bbernhar:
This patch repeatedly draws a primitive using a given topology. To verify rasterization, the framebuffer is checked if the test points are shaded or not.

### DIFF
--- a/src/webgpu/api/operation/render_pipeline/primitive_topology.spec.ts
+++ b/src/webgpu/api/operation/render_pipeline/primitive_topology.spec.ts
@@ -1,0 +1,395 @@
+export const description = `Test primitive topology rendering.
+
+Draw a primitive using 6 vertices with each topology and check if the pixel is covered.
+
+Vertex sequence and coordinates are the same for each topology:
+  - Vertex buffer = [v1, v2, v3, v4, v5, v6]
+  - Topology = [point-list, line-list, line-strip, triangle-list, triangle-strip]
+
+Test locations are framebuffer coordinates:
+  - Pixel value { valid: green, invalid: black, format: 'rgba8unorm'}
+  - Test point is valid if the pixel value equals the covered pixel value at the test location.
+  - Primitive restart occurs for strips (line-strip and triangle-strip) between [v3, v4].
+
+  Topology: point-list         Valid test location(s)           Invalid test location(s)
+
+       v2    v4     v6         Every vertex.                    Line-strip locations.
+                                                                Triangle-list locations.
+                                                                Triangle-strip locations.
+
+   v1     v3     v5
+
+  Topology: line-list (3 lines)
+
+       v2    v4     v6         Center of three line segments:   Line-strip locations.
+      *      *      *          {v1,V2}, {v3,v4}, and {v4,v5}.   Triangle-list locations.
+     *      *      *                                            Triangle-strip locations.
+    *      *      *
+   v1     v3     v5
+
+  Topology: line-strip (5 lines)
+
+       v2    v4     v6
+       **    **     *
+      *  *  *  *   *           Line-list locations              Triangle-list locations.
+     *    **     **          + Center of two line segments:     Triangle-strip locations.
+    v1    v3     v5            {v2,v3} and {v4,v5}.
+                                                                With primitive restart:
+                                                                Line segment {v3, v4}.
+
+  Topology: triangle-list (2 triangles)
+
+      v2       v4    v6
+      **        ******         Center of two triangle(s):       Triangle-strip locations.
+     ****        ****          {v1,v2,v3} and {v4,v5,v6}.
+    ******        **
+   v1     v3      v5
+
+  Topology: triangle-strip (4 triangles)
+
+      v2        v4      v6
+      ** ****** ** ******      Triangle-list locations          None.
+     **** **** **** ****     + Center of two triangle(s):
+    ****** ** ****** **        {v2,v3,v4} and {v3,v4,v5}.       With primitive restart:
+   v1       v3        v5                                        Triangle {v2, v3, v4}
+                                                                and {v3, v4, v5}.
+`;
+
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { GPUTest } from '../../../gpu_test.js';
+
+const kRTSize: number = 56;
+const kColorFormat = 'rgba8unorm';
+const kValidPixelColor = new Uint8Array([0x00, 0xff, 0x00, 0xff]); // green
+const kInvalidPixelColor = new Uint8Array([0x00, 0x00, 0x00, 0x00]); // black
+
+class Point2D {
+  x: number;
+  y: number;
+  z: number;
+  w: number;
+
+  constructor(x: number, y: number) {
+    this.x = x;
+    this.y = y;
+    this.z = 0;
+    this.w = 1;
+  }
+
+  toNDC(): Point2D {
+    // NDC coordinate space is y-up, so we negate the y mapping.
+    // To ensure the resulting vertex in NDC will be placed at the center of the pixel, we
+    // must offset by the pixel coordinates or 0.5.
+    return new Point2D((2 * (this.x + 0.5)) / kRTSize - 1, (-2 * (this.y + 0.5)) / kRTSize + 1);
+  }
+
+  static getMidpoint(a: Point2D, b: Point2D) {
+    return new Point2D((a.x + b.x) / 2, (a.y + b.y) / 2);
+  }
+
+  static getCentroid(a: Point2D, b: Point2D, c: Point2D) {
+    return new Point2D((a.x + b.x + c.x) / 3, (a.y + b.y + c.y) / 3);
+  }
+}
+
+interface TestLocation {
+  location: Point2D;
+  color: Uint8Array;
+}
+
+const VertexLocations = [
+  new Point2D(8, 24), // v1
+  new Point2D(16, 8), // v2
+  new Point2D(24, 24), // v3
+  new Point2D(32, 8), // v4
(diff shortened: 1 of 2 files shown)


src/webgpu/api/operation/render_pipeline/primitive_topology.spec.ts
```diff
@@ -0,0 +1,376 @@
...
+  - Vertex buffer = [v1, v2, v3, v4, v5, v6]
+  - Topology = [point-list, line-list, line-strip, triangle-list, triangle-strip]
+
+Test locations are framebuffer coordinates:
+  - Pixel value { valid: green, invalid: black, format: 'rgba8unorm'}
+  - Test point is valid if the pixel value equals the covered pixel value at the test location.
+
+  Topology: point-list    Test points(s)
+
+  v2      v3      v6      Every vertex.
```
Kangz, line 15: Maybe we should also have points that we check haven't been drawn (the center of the triangles, and middle of lines). Same thing for other cases.

Kangz: And it seems the test already does it, but the test plan should call it out.

src/webgpu/api/operation/render_pipeline/primitive_topology.spec.ts
```diff
@@ -0,0 +1,376 @@
...
+
+  v2       v4      v6
+  |        |       |      Center of each vertical line segment.
+  |        |       |
+  |        |       |
+  v1       v3      v5
+
+  Topology: line-strip (5 lines)
+
+  v2 ----- v3      v6
```
Kangz, line 31: Maybe we should test that the primitive restart correctly breaks the strip in two. Ditto for triangles.

src/webgpu/api/operation/render_pipeline/primitive_topology.spec.ts
```diff
@@ -0,0 +1,376 @@
...
+  | x           xxx |
+  v1       v4 ---- v5
+
+  Topology: triangle-strip (4 triangles)
+
+  v2 ----- v3       v6
+  | xxx xxx x     x |       Center of two triangle(s): {v1,v2,v3} and {v4,v5,v6}.
+  | xx   xx xx   xx |     + Center of two triangle(s): {v2,v3,v4} and {v3,v4,v5}.
+  | x     x xxx xxx |
+  v1        v4 ---- v5
```
kainino0x, line 51: The triangles v1-v2-v3 and v2-v3-v4 overlap (though it's hard to show that in the diagram).

Couldn't it happen that the centroid of one is within the other, so testing the centroids doesn't guarantee that the triangle was rasterized? (It depends on the exact triangle shapes, but I didn't look closely enough at them to be sure. Would be ideal if this problem were avoided by construction, e.g. avoiding overlapping triangles.)

src/webgpu/api/operation/render_pipeline/primitive_topology.spec.ts
```diff
@@ -0,0 +1,376 @@
...
+    renderPass.draw(6, 1, 0, 0);
+    renderPass.endPass();
+
+    this.device.defaultQueue.submit([encoder.finish()]);
+
+    // Compute test locations for valid and invalid pixels.
+    // If the primitive covers the pixel, the color value will be |kValidPixelColor|.
+    // Otherwise, a non-covered pixel will be |kInvalidPixelColor|.
+    let testLocations: TestLocation[] = [];
+    switch (this.params.primitiveTopology) {
```
kainino0x, line 301: nit: It's unfortunate to have so much of the test embedded into this switch statement. Perhaps instead of this with the parameterized single test, we could have one test for each primitive topology, and each test could just embed the contents of these switch cases, and pass it into `t.run(primitiveTopology, testLocations)`. WDYT?

src/webgpu/api/operation/render_pipeline/primitive_topology.spec.ts
```diff
@@ -0,0 +1,376 @@
...
+      vertexCoords[i * 4 + 3] = point.w;
+    }
+    return vertexCoords;
+  }
+
+  get params(): TestParams {
+    return super.params as TestParams;
+  }
+
+  static getPointTestLocations(expectedColor: Uint8Array): TestLocation[] {
```
kainino0x, line 146: nit: these could become standalone functions (rather than static methods) just to minimize the amount of code inside the fixture class (would make the fixture more readable).

src/webgpu/api/operation/render_pipeline/primitive_topology.spec.ts
bbernhar: Done.

src/webgpu/api/operation/render_pipeline/primitive_topology.spec.ts
bbernhar: Done.

src/webgpu/api/operation/render_pipeline/primitive_topology.spec.ts
bbernhar: Good point, staggered the points.

src/webgpu/api/operation/render_pipeline/primitive_topology.spec.ts
bbernhar: Done.

bbernhar:
Thanks @kainino0x and @Kangz, did another pass. PTAL!

src/webgpu/api/operation/render_pipeline/primitive_topology.spec.ts
```diff
@@ -0,0 +1,421 @@
...
+
+    renderPass.setPipeline(
+      this.device.createRenderPipeline({
+        vertexStage: { module: vertexModule, entryPoint: 'main' },
+        fragmentStage: { module: fragmentModule, entryPoint: 'main' },
+        primitiveTopology,
+        rasterizationState: {
+          frontFace: 'ccw',
+          cullMode: 'none',
+        },
```
kainino0x, line 297: These are defaults; rasterizationState can be omitted entirely
```suggestion
```

src/webgpu/api/operation/render_pipeline/primitive_topology.spec.ts
```diff
@@ -0,0 +1,421 @@
...
+
+  Topology: triangle-list (2 triangles)
+
+      v2       v4    v6
+      **        ******         Center of two triangle(s):       Triangle-strip locations.
+     ****        ****          {v1,v2,v3} and {v4,v5,v6}.
+    ******        **
+   v1     v3      v5
+
+  Topology: triangle-strip (5 triangles)
```
kainino0x, line 48: nit: 4 triangles

src/webgpu/api/operation/render_pipeline/primitive_topology.spec.ts
```diff
@@ -0,0 +1,421 @@
...
+      size: vertexArray.byteLength,
+      usage: GPUBufferUsage.VERTEX,
+    });
+    const vertexMapping = vertexBuffer.getMappedRange();
+    new Float32Array(vertexMapping).set(vertexArray);
+    vertexBuffer.unmap();
+    return vertexBuffer;
+  }
+
+  static generateVertexBuffer(vertexLocations: Point2D[]): Float32Array {
```
Kangz, line 225: nit: can be a freestanding function to avoid having to type the class name.

src/webgpu/api/operation/render_pipeline/primitive_topology.spec.ts
```diff
@@ -0,0 +1,421 @@
...
+    },
+  ];
+}
+
+class PrimitiveTopologyTest extends GPUTest {
+  makeAttachmentTexture(): GPUTexture {
+    return this.device.createTexture({
+      format: kColorFormat,
+      size: { width: kRTSize, height: kRTSize, depth: 1 },
+      sampleCount: 1,
```
Kangz, line 208: nit: this is the default value and not needed.

src/webgpu/api/operation/render_pipeline/primitive_topology.spec.ts
bbernhar: Done

src/webgpu/api/operation/render_pipeline/primitive_topology.spec.ts
bbernhar: Done

src/webgpu/api/operation/render_pipeline/primitive_topology.spec.ts
bbernhar: Done

src/webgpu/api/operation/render_pipeline/primitive_topology.spec.ts
bbernhar: Done

src/webgpu/api/operation/render_pipeline/primitive_topology.spec.ts
```diff
@@ -0,0 +1,421 @@
...
+      const point = vertexLocations[i].toNDC();
+      vertexCoords[i * 4 + 0] = point.x;
+      vertexCoords[i * 4 + 1] = point.y;
+      vertexCoords[i * 4 + 2] = point.z;
+      vertexCoords[i * 4 + 3] = point.w;
+    }
+    return vertexCoords;
+  }
+
+  makeIndexBuffer(indexArray: Uint32Array): GPUBuffer {
```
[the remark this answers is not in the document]
bbernhar: Done.

bbernhar:
Thanks @Kangz  and @kainino0x!

PR #260, opened 2020-08-13
Add more tests for texture resource usage tracking in render

Richard-Yunchao:
The new tests are:
  * texture usages by replaced bindings
  * texture usages in bundle

### DIFF
--- a/src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
+++ b/src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
@@ -28,12 +28,14 @@ Test Coverage:
     - Texture usages in bundle should be tracked if that bundle is executed in the current scope.
 `;
 
-import { poptions, params } from '../../../../common/framework/params_builder.js';
+import { pbool, poptions, params } from '../../../../common/framework/params_builder.js';
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
 import {
-  kShaderStages,
   kDepthStencilFormats,
   kDepthStencilFormatInfo,
+  kTextureBindingTypes,
+  kTextureBindingTypeInfo,
+  kShaderStages,
 } from '../../../capability_info.js';
 import { ValidationTest } from '../validation_test.js';
 
@@ -421,6 +423,8 @@ g.test('shader_stages_and_visibility')
     });
   });
 
+// We should track the texture usages in bindings which are replaced by another setBindGroup()
+// call site upon the same index in the same render pass.
 g.test('replaced_binding')
   .params(poptions('bindingType', kTextureBindingTypes))
   .fn(async t => {
@@ -428,11 +432,12 @@ g.test('replaced_binding')
     const info = kTextureBindingTypeInfo[bindingType];
     const bindingTexFormat = info.resource === 'storageTex' ? 'rgba8unorm' : undefined;
 
-    const view0 = t.createTexture().createView();
-    const view1 = t
+    const sampledView = t.createTexture().createView();
+    const sampledStorageView = t
       .createTexture({ usage: GPUTextureUsage.STORAGE | GPUTextureUsage.SAMPLED })
       .createView();
 
+    // Create bindGroup0. It has two bindings. These two bindings use different views/subresources.
     const bglEntries0: GPUBindGroupLayoutEntry[] = [
       { binding: 0, visibility: GPUShaderStage.FRAGMENT, type: 'sampled-texture' },
       {
@@ -443,15 +448,17 @@ g.test('replaced_binding')
       },
     ];
     const bgEntries0: GPUBindGroupEntry[] = [
-      { binding: 0, resource: view0 },
-      { binding: 1, resource: view1 },
+      { binding: 0, resource: sampledView },
+      { binding: 1, resource: sampledStorageView },
     ];
     const bindGroup0 = t.device.createBindGroup({
       entries: bgEntries0,
       layout: t.device.createBindGroupLayout({ entries: bglEntries0 }),
     });
 
-    const bindGroup1 = t.createBindGroup(0, view1, 'sampled-texture', undefined);
+    // Create bindGroup1. It has one binding, which use the same view/subresoure of a binding in
+    // bindGroup0. So it may or may not conflicts with that binding in bindGroup0.
+    const bindGroup1 = t.createBindGroup(0, sampledStorageView, 'sampled-texture', undefined);
 
     const encoder = t.device.createCommandEncoder();
     const pass = encoder.beginRenderPass({
@@ -463,31 +470,40 @@ g.test('replaced_binding')
         },
       ],
     });
+
+    // Set bindGroup0 and bindGroup1. bindGroup0 is replaced by bindGroup1 in the current pass.
+    // But bindings in bindGroup0 should be tracked too.
     pass.setBindGroup(0, bindGroup0);
     pass.setBindGroup(0, bindGroup1);
     pass.endPass();
 
-    // We should track the texture usages in bindings which are replaced by another setBindGroup()
-    // call site upon the same index in the same render pass.
     const success = bindingType === 'writeonly-storage-texture' ? false : true;
     t.expectValidationError(() => {
       encoder.finish();
     }, !success);
   });
 
-g.test('binding_in_bundle')
+g.test('bindings_in_bundle')
   .params(
     params()
-      .combine(poptions('typeInBundle', kTextureBindingTypes))
-      .combine(poptions('typeInPass', ['sampled-texture', 'render-target'] as const))
+      .combine(pbool('binding0InBundle'))
+      .combine(pbool('binding1InBundle'))
+      .combine(poptions('type0', ['render-target', ...kTextureBindingTypes] as const))
+      .combine(poptions('type1', ['render-target', ...kTextureBindingTypes] as const))
+      .unless(
+        ({ binding0InBundle, binding1InBundle, type0, type1 }) =>
+          // We can't set 'render-target' in bundle, so we need to exclude it from bundle.
+          // In addition, if both bindings are non-bundle, there is no need to test it because
+          // we have far more comprehensive test cases for that situation in this file.
+          (binding0InBundle && type0 === 'render-target') ||
+          (binding1InBundle && type1 === 'render-target') ||
+          (!binding0InBundle && !binding1InBundle)
+      )
   )
   .fn(async t => {
-    const { typeInBundle, typeInPass } = t.params;
-    const info = kTextureBindingTypeInfo[typeInBundle];
-    const bindingTexFormat = info.resource === 'storageTex' ? 'rgba8unorm' : undefined;
+    const { binding0InBundle, binding1InBundle, type0, type1 } = t.params;
 
-    // Two bindings are attached to the same texture view. The first binding in bindGroup0 is set
-    // in bundle, the second binding is set in render pass via bindGroup1 or color attachment.
+    // Two bindings are attached to the same texture view.
     const view = t
       .createTexture({
         usage:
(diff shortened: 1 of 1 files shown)


Richard-Yunchao:
For bundle, do you think that we need to test texture usage conflict between bundles (no texture usages in non-bundle at all)? I think it is not necessary from the perspective of tracking: if we are tracking usages in bundle(s), we can find out the usage conflict, no matter the conflict happens between bundle and non-bundle, or between bundles themselves. But if you think it is necessary, I can add more tests.

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
```diff
@@ -232,3 +237,133 @@ g.test('shader_stages_and_visibility')
       encoder.finish();
     });
   });
+
+g.test('replaced_binding')
```
Kangz, line 241: Can we have these tests added to the test plan above?

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
```diff
@@ -232,3 +237,133 @@ g.test('shader_stages_and_visibility')
...
+
+    // We should track the texture usages in bindings which are replaced by another setBindGroupi()
+    // call site upon the same index in the same render pass.
+    const success = bindingType === 'writeonly-storage-texture' ? false : true;
+    t.expectValidationError(() => {
+      encoder.finish();
+    }, !success);
+  });
+
+g.test('binding_in_bundle')
```
Kangz, line 300: ditto

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
```diff
@@ -232,3 +237,133 @@ g.test('shader_stages_and_visibility')
...
+    params()
+      .combine(poptions('typeInBundle', kTextureBindingTypes))
+      .combine(poptions('typeInPass', ['sampled-texture', 'render-target'] as const))
+  )
+  .fn(async t => {
+    const { typeInBundle, typeInPass } = t.params;
+    const info = kTextureBindingTypeInfo[typeInBundle];
+    const bindingTexFormat = info.resource === 'storageTex' ? 'rgba8unorm' : undefined;
+
+    const view = t
```
Kangz, line 311: u-nit: try to have some block comments to make following the code slightly easier.

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
```diff
@@ -232,3 +237,133 @@ g.test('shader_stages_and_visibility')
...
+      .createTexture({
+        usage:
+          GPUTextureUsage.OUTPUT_ATTACHMENT | GPUTextureUsage.STORAGE | GPUTextureUsage.SAMPLED,
+      })
+      .createView();
+
+    const bindGroup0 = t.device.createBindGroup({
```
Kangz, line 318: This is done in multiple tests, maybe we could have a helpers that's createBindGroupForTextureUsage ?

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
```diff
@@ -232,3 +237,133 @@ g.test('shader_stages_and_visibility')
...
+    const success = bindingType === 'writeonly-storage-texture' ? false : true;
+    t.expectValidationError(() => {
+      encoder.finish();
+    }, !success);
+  });
+
+g.test('binding_in_bundle')
+  .params(
+    params()
+      .combine(poptions('typeInBundle', kTextureBindingTypes))
```
Kangz, line 303: Maybe this could be an additional parametrization of the `readwrite_upon_subresources` test that for each of the views says whether it is used in a bundle or not?

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
Richard-Yunchao: Done

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
Richard-Yunchao: Done

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
Richard-Yunchao: Done

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
Richard-Yunchao: readwrite_upon_subresources will be change in another pull request. I am considering that keep this test (a check for resource usage between bundle and render pass) but add a similar parameter as you suggested for "bundle" (pure bundle, no resource usage in render. The only exception is render-target's attachment), "render", "compute". Then most tests in this file will be expanded to 3 times as it is now. I am not sure whether that expansion are too many though. WDYT?

Kangz: Don't we need to test that bundles properly merge their subresource tracking with the main pass (including conflicts between two different bundles)? Doing it via another parametrization of `readwrite_upon_subresources` adds a x4 multiplier but gives us great coverage of that feature.

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
Kangz: Sorry what I meant is have comments every 5-15 lines to explain what's happening next. It helps skim code very quickly.

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
Richard-Yunchao: Yeah. we can add that and get a lot more coverage. BTW, bundle is only introduced in render pass, so it should add a x3 multiplier: render pass with bundle, render pass without bundle, compute pass as I said. But you mentioned x4 multiplier. What's the 4th multiplier? compute pass with bundle doesn't exist per my understanding. Or you think we need to track one subresource in render pass, another in bundle for almost all test and make it be the 4th? (It is what this test is doing, but this test doesn't cover a lot combination, considering that we have already had a lot of combination somewhere. I just make sure resource in bundle is being tracked).

austinEng: I think the four are:

setBindGroup, executeBundles
setBindGroup, setBindGroup
executeBundles, executeBundles
executeBundles, setBindGroup

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
```diff
@@ -232,3 +251,113 @@ g.test('shader_stages_and_visibility')
...
+    const bgEntries0: GPUBindGroupEntry[] = [
+      { binding: 0, resource: view0 },
+      { binding: 1, resource: view1 },
+    ];
+    const bindGroup0 = t.device.createBindGroup({
+      entries: bgEntries0,
+      layout: t.device.createBindGroupLayout({ entries: bglEntries0 }),
+    });
+
+    const bindGroup1 = t.createBindGroup(0, view1, 'sampled-texture', undefined);
```
austinEng, line 285: For better readability, could we name these as their type? like `sampledView`, `sampledStorageView`, `sampledAndParameterizedEntries`, `sampledAndSampledStorageEntries` etc.. 

This makes it easier to tell what each bind group contains, and what bindings are replaced when we call setBindGroup

kainino0x:
Sorry for the review delay; will get to it Monday

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
Kangz: Although like @Richard-Yunchao said the `executeBundles,  setBindGroup` and `setBindGroup, executeBundles` are similar so they could be deduplicated for only a x3 multiplier.

kainino0x: One of the typeInPass options is render-target, which can't be inside a setBindGroup or an executeBundles. Unless we want to generalize it heavily I don't think this is the right test to test conflicts between bindgroups and bundles.

Kangz: That's ok if typeInPass is render-target for a bundle texture we can skip the test. But we definitely need to test interactions between a texture being used as an attachment and sampled in a bundle in that render pass.

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
Richard-Yunchao: Done.

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
Richard-Yunchao: I revisit the test carefully and I personally think that it is not necessary to add "bindingInBundleOrNot" parameter to all tests in this file, because factors like subresources' combination or aspects or binding types are somehow irrelevant to bundle or not. And your suggestion seems to expand x3 or x4 in this test only. So I expanded this test to x3: 
setBindGroup in pass directly, executeBundle,
executeBundle, setBindGroup in pass directly,
executeBundle, excueteBundle. 
I didn't add non-bundle combination (setBindGroup in pass directly + setBindGroup in pass directly) because it was covered in many other tests.

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
Richard-Yunchao: Good suggestion. I used sampledView and sampledStorageView. For bg/bgl/bgEntries/bglEntries, I still use 0 or 1 to keep it short and consistent among bg/bgl/bgEntries/bglEntries.

Richard-Yunchao:
@kainino0x , I somehow significantly changed the tests to validate bindings for bundles. Please take another look.

kainino0x:
I'll take a look now, but to make reviewing easier, two organizational requests: (1) please mark comment threads as "Resolved" when they're done, unless they need re-review/discussion, (2) avoid force pushes and rebases between re-reviews if possible (merge commits are okay) because it prevents GitHub from being able to show changes since my last review (unlike Gerrit).
Thanks :)

kainino0x:
Though I realized (2) probably wouldn't have made much of a difference in this case since the test changed quite a bit.

Richard-Yunchao:
> I'll take a look now, but to make reviewing easier, two organizational requests: (1) please mark comment threads as "Resolved" when they're done, unless they need re-review/discussion, (2) avoid force pushes and rebases between re-reviews if possible (merge commits are okay) because it prevents GitHub from being able to show changes since my last review (unlike Gerrit).
> Thanks :)

Thanks for your suggestion, @kainino0x . If you feel anything that can improve the review process, please let know. The user experience of different review tools differs a lot. 1) is easy, I will do it in future. For 2), what's the best practice if I know that there is a conflict and I do need to rebase the code?

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
```diff
@@ -392,3 +422,155 @@ g.test('shader_stages_and_visibility')
...
+
+    // Two bindings are attached to the same texture view.
+    const view = t
+      .createTexture({
+        usage:
+          GPUTextureUsage.OUTPUT_ATTACHMENT | GPUTextureUsage.STORAGE | GPUTextureUsage.SAMPLED,
+      })
+      .createView();
+
+    const bindGroups: GPUBindGroup[] = [];
```
kainino0x, line 514: uNit:
```suggestion
    const bindGroups: (GPUBindGroup | undefined)[] = [];
```

kainino0x:
> 2), what's the best practice if I know that there is a conflict and I do need to rebase the code?

My suggestion is just avoid `git rebase` and `git push -f` after there's been a review, and instead use `git merge` if a conflict needs to be resolved with the main branch. Otherwise merge commits aren't needed either. That way I can review individual commits and skip over the merge commits.

austinEng:
Strong +1 on these workflow suggestions. I try to review and follow PRs as well, but it becomes quite difficult doing so and looking at the diff. Kai, we should put this in a Contributing.md file as well.

Richard-Yunchao:
> > 2), what's the best practice if I know that there is a conflict and I do need to rebase the code?
> 
> My suggestion is just avoid `git rebase` and `git push -f` after there's been a review, and instead use `git merge` if a conflict needs to be resolved with the main branch. Otherwise merge commits aren't needed either. That way I can review individual commits and skip over the merge commits.

OK. I will use `git merge` in future.

Richard-Yunchao:
> Strong +1 on these workflow suggestions. I try to review and follow PRs as well, but it becomes quite difficult doing so and looking at the diff. Kai, we should put this in a Contributing.md file as well.
Yeah, good suggestion. Looks like this is not a problem when we use gerrit.

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
Richard-Yunchao: Hmm. I tried this. But if failed when I run "npx grunt pre". The error reported that when we call setBindGroup(i, bindGroups[i]); type 'GPUBindGroup | undefined' is not assignable to parameter of type 'GPUBindGroup'.

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
```diff
@@ -392,3 +422,155 @@ g.test('shader_stages_and_visibility')
...
+    if (type0 !== 'render-target') {
+      const binding0TexFormat = type0 === 'sampled-texture' ? undefined : 'rgba8unorm';
+      bindGroups[0] = t.createBindGroup(0, view, type0, binding0TexFormat);
+    }
+    if (type1 !== 'render-target') {
+      const binding1TexFormat = type1 === 'sampled-texture' ? undefined : 'rgba8unorm';
+      bindGroups[1] = t.createBindGroup(1, view, type1, binding1TexFormat);
+    }
```
Richard-Yunchao, line 522: I also tried to use an array and a for loop for this code snippet above (You see, it repeated twice). It turned out that "npx grunt pre" will fail because "render-target" is not assignable to binding type at somewhere. Similar to the error if I do the change for code at line 545 as you suggested.

kainino0x: This isn't the error I get if I make the change I suggested, but I do get a different error that's not a real problem.

Let's just ignore this, it's fine as is.

Richard-Yunchao:
Can we merge this pr? @kainino0x . All feedback has been resolved.

kainino0x:
Right, sorry :)

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
Kangz: Sorry for the really slow reviews, I'm having a hard time staying on top of my inbox these days. It is important to test bundles in combination right now we have very imperfect coverage of [PassResourceUsageTracker::AddTextureUsage](https://source.chromium.org/chromium/chromium/src/+/master:third_party/dawn/src/dawn_native/PassResourceUsageTracker.cpp;drc=1562cab3f1eda927938f8f4a5a91991fefde66d3;l=67).

Usage tracking is a place where implementations are likely to do fancy optimizations and we need to provide them with great coverage. This is why I'm strongly suggesting that we do:

 - combination testing of render-pass render-pass interaction, render-bundle render-pass interaction in both direction, and render-bundle interactions between different render bundles
 - more varied shapes of subresources intersections for both resources

Richard-Yunchao: Corentin, sounds like you want to expand the tests to 3 times for many other test cases (say `subresources_and_binding_types_combinations_for_color`) in this file, just as I suggested before. I thought about this carefully but right now I think it might be unnecessary. For example, subresource types combination and overlap measuring (different mip/array/plane slices, etc) is orthogonal to binding types combination (read-read, read-write, write-write usages). Because the implementation about whether a bunch of subresources overlap with another bunch of subresources is independent from the implementation about whether a binding type conflicts with another binding type. So I don't think the further combination of subresource types combination and binding types combinations make a lot of sense. So when we test subresource types/overlap, we should exhaust the subresource types/overlaps combination, with a few typical binding types combination. We don't need to iterate and exhaust every single binding types combinations for subresource types/overlap tests. It is the same when we test binding type combinations. But we did combination of these two orthogonal factors, which definitely improved the coverage, even though it might not make a lot of sense. 

`Bindings in bundle or not` is another factor, which is orthogonal to subresource types (They are set in BindGroup, we don't know they will be in bundle or not at that time) and binding types (they are set in BGL, we don't know they will be in bundle or not either), per my understanding. So I don't think that it makes a lot of sense to combine `bindings in bundle or not` with other factors (we can introduce a few typical cases of other factors, though). But it we exhaust the combination, it will definitely improves the coverage blindly. 

And we have other factors like visibility, sync scope (per draw/dispatch, or per pass, or out of pass), unused bindings, etc. Ideally we can combine all these factors and exhaust every possible combinations and make the tests cover every possible case, with the cost of huge complexity and terrible readability. So It's a balance, even though CTS is meant to cover every single corner (but it might not necessary to combine orthogonal/irrelevant factors). 

Well, if you think it is necessary, I will still try to combine `bindings in bundle or not` as a new parameter and expand other tests and see the complexity and readability after I added `compute` as a new parameter for tests to cover validation in compute.

Kangz: This comment is not yet addressed.

Richard-Yunchao: Sorry for late response, Corentin. It is addressed at https://github.com/gpuweb/cts/pull/281. PTAL.

PR #263, opened 2020-08-14
Update types and split texture tables to accommodate compressed formats

kainino0x:


kainino0x:
@austinEng PTAL
@ponitka @Richard-Yunchao FYI

src/webgpu/api/operation/resource_init/texture_zero_init_test.ts
```diff
@@ -2,7 +2,13 @@ import { TestCaseRecorder } from '../../../../common/framework/logging/test_case
 import { params, poptions, pbool } from '../../../../common/framework/params_builder.js';
 import { CaseParams } from '../../../../common/framework/params_utils.js';
 import { assert, unreachable } from '../../../../common/framework/util/util.js';
-import { kTextureAspects, kTextureFormatInfo, kTextureFormats } from '../../../capability_info.js';
+import {
+  kTextureAspects,
+  CoreTextureFormat,
```
austinEng, line 7: nit: sort alphabetically or put the lowercase ones together?

src/webgpu/util/texture/texelData.ts
```diff
@@ -126,7 +126,7 @@ const repeatComponents = (
 };
 
 const kRepresentationInfo: {
-  [k in GPUTextureFormat]: {
+  [k in CoreTextureFormat]: {
```
austinEng, line 129: hm.. maybe "Core" isn't the right separation, since if we have other extension formats that aren't compressed, this part should work. Maybe Uncompressed/Compressed ? Or maybe Block ?

kainino0x: I was thinking this would be the list of formats that don't require any extensions, but I guess in practice any tests for the core formats ought to be updated to support non-core formats too. Part of the problem is I still haven't figured out how to write tests that require WebGPU extensions to be enabled, and as soon as we add non-core formats to this list a bunch of tests would need to be updated to handle extensions (which can't be done yet). So not sure what to do here.

austinEng: Well, I imagine there will be some way to do it though, and we could call this something that's not "Core" and figure out the mechanism when we need to

Maybe something like

```typescript
g.test('asdf')
   // a special pcombine where they're all pbool for on/off. And at runtime we'll filter things out
  .extensions('texture_compression_bc', ...)

  .params(
    params()
      .combine(kCases)

      // maybe also have a runtime filter? so the static test list generator will always assume everything could be
      // present, but then we remove some at runtime
      .extensionFilter(p => { ... })
   )

  .fn(async t => {
      t.extensions['texture_compression_bc'] // => true/false for enabled/disabled
  });
```

PR #264, opened 2020-08-14
[glsl-dependent] Remove usages of old mapping

kainino0x:


kainino0x:
np, it was only like one working hour

PR #265, opened 2020-08-14
Improve shader robust access tests with more memory types.

Kangz:
PTAL!

### DIFF
--- a/src/webgpu/shader/execution/robust_access.spec.ts
+++ b/src/webgpu/shader/execution/robust_access.spec.ts
@@ -1,5 +1,15 @@
 export const description = `
-Tests to check array clamping in shaders is correctly implemented including vector / matrix indexing
+Tests to check datatype clamping in shaders is correctly implemented including vector / matrix indexing
+
+- For each shader stage (TODO):
+  - For various memory spaces (storage, uniform, global, function local, shared memory, TODO(in, out))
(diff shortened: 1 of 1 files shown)


kainino0x:
Sorry for the review delay; will get to it Monday

src/webgpu/shader/execution/robust_access.spec.ts
```diff
@@ -295,74 +382,80 @@ g.test('bufferMemory')
...
+      ${globalSource}
+
+      uint runTest() {
+        ${testFunctionSource}
+        return 0;
+      }`;
+
+    // Run it.
+    if (usesBuffer) {
+      // Create a buffer that contains zeroes in the allowed access area, and 42s everywhere else.
```
kainino0x, line 434: Not new, but aren't we supposed to allow reads from anywhere in the buffer to allow Vulkan's robust access behavior?

Kangz: I'm almost 100% sure that we decided that the array accesses would be clamped.

kainino0x: ugh, I can't keep anything straight anymore. We've gone through too many proposals on every single little decision in the API. I think you're right.

PR #270, opened 2020-08-18
Support JSON objects in param values

kainino0x:
This was mostly supported before, but parsed incorrectly in query strings.

### DIFF
--- a/src/common/framework/query/parseQuery.ts
+++ b/src/common/framework/query/parseQuery.ts
@@ -30,9 +30,29 @@ function parseQueryImpl(s: string): TestQuery {
   // Undo encodeURIComponentSelectively
   s = decodeURIComponent(s);
 
-  // bigParts are: suite, group, test, params (note kBigSeparator could appear in params)
-  const [suite, fileString, testString, paramsString] = s.split(kBigSeparator, 4);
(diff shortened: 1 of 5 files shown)


kainino0x:
cc @ponitka

src/common/framework/query/parseQuery.ts
```diff
@@ -30,9 +30,29 @@ function parseQueryImpl(s: string): TestQuery {
   // Undo encodeURIComponentSelectively
   s = decodeURIComponent(s);
 
-  // bigParts are: suite, group, test, params (note kBigSeparator could appear in params)
-  const [suite, fileString, testString, paramsString] = s.split(kBigSeparator, 4);
-  assert(fileString !== undefined, `filter string must have at least one ${kBigSeparator}`);
+  // bigParts are: suite, file, test, params (note kBigSeparator could appear in params)
```
austinEng, line 33: mm.. for simplicity, whenever we reconsider the commas, maybe we can also reconsider this separator as well. even just doubling it to "::" would be fine

kainino0x: It's fine because we do a limited number of splits, so it doesn't matter if it appears in params. But was considering making a change like that anyway (not sure if it's worth it)

PR #274, opened 2020-08-21
Add more texture usage validation tests: unused bindings

Richard-Yunchao:
The tests have a simple no-op wgsl shader, but it leads to device lost when I verify the tests in Chromium (I pulled the code and built it a few days ago, not very old). Either the wgsl is incorrect, or the wgsl code snippet I wrote can't be supported by Chromium so far. I guess the wgsl code itself is incorrect. For example, the bindings definition in wgsl. I am not familiar with wgsl and I can't find a good example about wgsl for render. Can you take a look?@kainino0x , @Kangz , @austinEng ?

### DIFF
--- a/src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
+++ b/src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
@@ -579,26 +579,30 @@ g.test('bindings_in_bundle')
   });
 
 g.test('unused_bindings_in_pipeline')
-  .params(params().combine(pbool('useBindGroup0')).combine(pbool('useBindGroup1')))
+  .params(
+    params()
+      .combine(pbool('useBindGroup0'))
+      .combine(pbool('useBindGroup1'))
+      .combine(pbool('callSetPipeline'))
+      .combine(pbool('callDraw'))
+  )
   .fn(async t => {
-    const { useBindGroup0, useBindGroup1 } = t.params;
+    const { useBindGroup0, useBindGroup1, callSetPipeline, callDraw } = t.params;
     const view = t.createTexture({ usage: GPUTextureUsage.STORAGE }).createView();
     const bindGroup0 = t.createBindGroup(0, view, 'readonly-storage-texture', 'rgba8unorm');
     const bindGroup1 = t.createBindGroup(0, view, 'writeonly-storage-texture', 'rgba8unorm');
 
     const wgslVertex = `
       fn main() -> void {
-      return;
+        return;
       }
 
       entry_point vertex = main;
     `;
-    const binding0: string = useBindGroup0 ? '[[set 0, binding 0]] var<image> image0' : '';
-    const binding1: string = useBindGroup1 ? '[[set 1, binding 0]] var<image> image1' : '';
-    const wgslFragment =
-      binding0 +
-      binding1 +
-      `
+    // TODO: revisit the shader code once 'image' can be supported in wgsl.
+    const wgslFragment = `
+      ${useBindGroup0 ? '[[set 0, binding 0]] var<image> image0;' : ''}
+      ${useBindGroup1 ? '[[set 1, binding 0]] var<image> image1;' : ''}
       fn main() -> void {
         return;
       }
@@ -632,10 +636,10 @@ g.test('unused_bindings_in_pipeline')
         },
       ],
     });
-    pass.setPipeline(pipeline);
     pass.setBindGroup(0, bindGroup0);
     pass.setBindGroup(1, bindGroup1);
-    pass.draw(3, 1, 0, 0);
+    if (callSetPipeline) pass.setPipeline(pipeline);
+    if (callDraw) pass.draw(3, 1, 0, 0);
     pass.endPass();
 
     t.expectValidationError(() => {
--- a/src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
+++ b/src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
@@ -588,7 +588,9 @@ g.test('unused_bindings_in_pipeline')
 
     const wgslVertex = `
       fn main() -> void {
+      return;
       }
+
       entry_point vertex = main;
     `;
     const binding0: string = useBindGroup0 ? '[[set 0, binding 0]] var<image> image0' : '';
@@ -598,7 +600,9 @@ g.test('unused_bindings_in_pipeline')
       binding1 +
       `
       fn main() -> void {
+        return;
       }
+
       entry_point fragment = main;
     `;
     const pipeline = t.device.createRenderPipeline({
--- a/src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
+++ b/src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
@@ -26,6 +26,9 @@ Test Coverage:
 
   - Test texture usages in bundle:
     - Texture usages in bundle should be tracked if that bundle is executed in the current scope.
+
+  - Test texture usages with unused bindings:
+    - Texture usages should be tracked even its bindings is not used in pipeline.
 `;
 
 import { pbool, poptions, params } from '../../../../common/framework/params_builder.js';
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
```diff
@@ -574,3 +577,64 @@ g.test('bindings_in_bundle')
...
+  .params(params().combine(pbool('useBindGroup0')).combine(pbool('useBindGroup1')))
+  .fn(async t => {
+    const { useBindGroup0, useBindGroup1 } = t.params;
+    const view = t.createTexture({ usage: GPUTextureUsage.STORAGE }).createView();
+    const bindGroup0 = t.createBindGroup(0, view, 'readonly-storage-texture', 'rgba8unorm');
+    const bindGroup1 = t.createBindGroup(0, view, 'writeonly-storage-texture', 'rgba8unorm');
+
+    const wgslVertex = `
+      fn main() -> void {
+      }
```
dj2, line 591: This requires a 'return' statement at the end.

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
```diff
@@ -574,3 +577,64 @@ g.test('bindings_in_bundle')
...
+      entry_point vertex = main;
+    `;
+    const binding0: string = useBindGroup0 ? '[[set 0, binding 0]] var<image> image0' : '';
+    const binding1: string = useBindGroup1 ? '[[set 1, binding 0]] var<image> image1' : '';
+    const wgslFragment =
+      binding0 +
+      binding1 +
+      `
+      fn main() -> void {
+      }
```
dj2, line 601: Missing 'return' statement.

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
Richard-Yunchao: Done.

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
Richard-Yunchao: Done.

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
```diff
@@ -574,3 +577,68 @@ g.test('bindings_in_bundle')
...
+  .params(params().combine(pbool('useBindGroup0')).combine(pbool('useBindGroup1')))
+  .fn(async t => {
+    const { useBindGroup0, useBindGroup1 } = t.params;
+    const view = t.createTexture({ usage: GPUTextureUsage.STORAGE }).createView();
+    const bindGroup0 = t.createBindGroup(0, view, 'readonly-storage-texture', 'rgba8unorm');
+    const bindGroup1 = t.createBindGroup(0, view, 'writeonly-storage-texture', 'rgba8unorm');
+
+    const wgslVertex = `
+      fn main() -> void {
+      return;
```
kainino0x, line 591:
```suggestion
        return;
```

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
```diff
@@ -574,3 +577,68 @@ g.test('bindings_in_bundle')
...
+      }
+
+      entry_point vertex = main;
+    `;
+    const binding0: string = useBindGroup0 ? '[[set 0, binding 0]] var<image> image0' : '';
+    const binding1: string = useBindGroup1 ? '[[set 1, binding 0]] var<image> image1' : '';
```
kainino0x, line 597: Missing terminal semicolons

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
```diff
@@ -574,3 +577,68 @@ g.test('bindings_in_bundle')
...
+    const wgslFragment =
+      binding0 +
+      binding1 +
+      `
+      fn main() -> void {
+        return;
+      }
+
+      entry_point fragment = main;
+    `;
```
kainino0x, line 607: optional nit: I think this is *slightly* more readable if written more like a preprocessor?:
```suggestion
    const wgslFragment = `
      ${useBindGroup0 ? '[[set 0, binding 0]] var<image> image0;' : ''}
      ${useBindGroup1 ? '[[set 1, binding 0]] var<image> image1;' : ''}
      
      fn main() -> void { return; }
      
      entry_point fragment = main;
    `;
```

Likely we should soon consider having some very lightweight preprocessing helper utils, but I don't know what they should look like yet.

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
```diff
@@ -574,3 +577,68 @@ g.test('bindings_in_bundle')
...
+          attachment: t.createTexture().createView(),
+          loadValue: { r: 0.0, g: 1.0, b: 0.0, a: 1.0 },
+          storeOp: 'store',
+        },
+      ],
+    });
+    pass.setPipeline(pipeline);
+    pass.setBindGroup(0, bindGroup0);
+    pass.setBindGroup(1, bindGroup1);
+    pass.draw(3, 1, 0, 0);
```
kainino0x, line 638: Technically this test should fail even if neither `setPipeline` nor `draw` is set at all. Since we're trying to tickle errors in validation how about testing a few additional permutations here, like:
- setPipeline, setBindGroups, draw
- setPipeline, setBindGroups
- setBindGroups, setPipeline
- setPipeline, setBindGroups, setPipeline

x= setBindGroups {[0, 1], [1, 0]}

params could be:
- setPipelineBefore = {true,false}
- setBindGroupsOrder = {[0, 1], [1, 0]}
- setPipelineAfter = {true, false}
- draw = {true, false}

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
Richard-Yunchao: Done. We can revisit this after preprocessing helper and 'image' can be supported in WGSL.

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
Richard-Yunchao: Per my understanding, setting Pipeline before setting bind groups or after setting bind groups should be tested in pipeline (or bind group) related tests. And I remember we do have those tests in Dawn validation tests in pipeline (or bindgroups?) related tests, which manifest that the sequence doesn't change anything. So I suggest we ignore the sequence here. But adding setPipeline and draw as bool parameters to the tests is a really good suggestion, in addition to adding bindings or not in shader code. Right now, the tests can manifest that we should validate resources if they are set via state setting API (like setBindGroup) regardless of whether the shader, or pipeline, or draw call depends on it.

Richard-Yunchao:
> Looks good; suggested a small parameterization. Do you want to wait until we've been able to run this before landing? Or should we just go ahead with it?

I'd like to go ahead and land, and revisit if we need to do some small changes to the shader code after we can verify it in browser. I have already added a TODO. Is it OK? @kainino0x

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
Kangz: It can't hurt to have it here too, since like @kainino0x said the goal is to tickle the resource tracking validation and we're combining with other interesting tickly combinations.

kainino0x: Yes, and these are fast tests so there's no real harm in there being possible duplication. Especially since it's very easy to express these tests here. You can even add cases like setBindGroupsOrder [0, 1, 0].

kainino0x:
Sounds good since the TODO is there.

I'd like to do the small generalization suggested above. It should be a very quick modification to the test and could improve our coverage a lot, I think.

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
Richard-Yunchao: No problem. I have done this change. I am not sure how to do  "setBindGroupsOrder [0, 1, 0]" or "setBindGroupsOrder = {[0, 1], [1, 0]}". But I suppose that you want to reverse the bind groups indices. In addition, setPipeline can be added into the middle of two setBindGroups, which make sense for the index order of bind groups too.

Richard-Yunchao:
> Sounds good since the TODO is there.
> 
> I'd like to do the small generalization suggested above. It should be a very quick modification to the test and could improve our coverage a lot, I think.

Make sense. Please take a look at the latest patch set when you have time, @kainino0x

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
kainino0x: With the addition you made to setPipeline between the two calls, you're right setBindGroupsOrder no longer makes sense. Your latest change LGTM.

kainino0x (approved):

PR #276, opened 2020-08-24
Fix setScissorRect() validation

kunalmohan:
According to the spec-
- x+width is less than or equal to this.[[attachment_size]].width.
- y+height is less than or equal to this.[[attachment_size]].height.


kainino0x:
Looks like this appeared here: https://github.com/gpuweb/gpuweb/pull/850
and I don't actually remember which behavior is supposed to be the correct one. Looks like right now the viewport can be anything but the scissor has to be inside the attachment?

@toji: what did you base that rule on?
cc @kvark in case you remember the details.

kainino0x:
(AFAIK this test was originally based on Dawn's behavior - Francois wrote it against Chrome - and I don't remember if I specifically reviewed this test expectation at the time.)

toji:
Scissor validation came from this comment thread: https://github.com/gpuweb/gpuweb/pull/850#discussion_r436922968

kainino0x:
Aha, found https://github.com/gpuweb/gpuweb/issues/373

PR #278, opened 2020-08-28
Add Resolve Operation Test

bjjones:
This tests most of the necessary combinations. Unfortunately I don't have a convenient way to make the shader compatible with all renderable formats.

### DIFF
--- a/src/webgpu/api/operation/render_pass/resolve.spec.ts
+++ b/src/webgpu/api/operation/render_pass/resolve.spec.ts
@@ -1,0 +1,197 @@
+export const description = `API Operation Tests for RenderPass StoreOp.
+
+Tests a render pass with a resolveTarget resolves correctly for many combinations of:
+  - number of color attachments, some with and some without a resolveTarget
+  - renderPass storeOp set to {‘store’, ‘clear’}
+  - resolveTarget mip level set to {‘0’, base mip > ‘0’}
+  - resolveTarget base array layer set to {‘0’, base layer > '0'} for 2D textures
+  TODO: test all renderable color formats
+  TODO: test that any not-resolved attachments are rendered to correctly.
+`;
+
+import { params, poptions } from '../../../../common/framework/params_builder.js';
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { GPUTest } from '../../../gpu_test.js';
+
+const kSlotsToResolve = [
+  [0, 2],
+  [1, 3],
+  [0, 1, 2, 3],
+];
+
+const kSize = 4;
+const kFormat: GPUTextureFormat = 'rgba8unorm';
+
+export const g = makeTestGroup(GPUTest);
+
+g.test('render_pass_resolve')
+  .params(
+    params()
+      .combine(poptions('numColorAttachments', [2, 4] as const))
+      .combine(poptions('slotsToResolve', kSlotsToResolve))
+      .combine(poptions('storeOperation', ['clear', 'store'] as const))
+      .combine(poptions('resolveTargetBaseMipLevel', [0, 1] as const))
+      .combine(poptions('resolveTargetBaseArrayLayer', [0, 1] as const))
+  )
+  .fn(t => {
+    // These shaders will draw a white triangle into a texture. After draw, the top left
+    // half of the texture will be white, and the bottom right half will be unchanged. When this
+    // texture is resolved, there will be two distinct colors in each portion of the texture, as
+    // well as a line between the portions that contain the midpoint color due to the multisample
+    // resolve.
+    const vertexModule = t.makeShaderModule('vertex', {
+      glsl: `
+      #version 450
+      const vec2 pos[3] = vec2[3](vec2(-1.0f, -1.0f), vec2(-1.0f, 1.0), vec2(1.0, 1.0));
+        void main() {
+          gl_Position = vec4(pos[gl_VertexIndex], 0.0f, 1.0f);
+        }
+      `,
+    });
+
+    const fragmentModule = t.makeShaderModule('fragment', {
+      glsl: `
+      #version 450
+        layout(location = 0) out vec4 fragColor0;
+        layout(location = 1) out vec4 fragColor1;
+        layout(location = 2) out vec4 fragColor2;
+        layout(location = 3) out vec4 fragColor3;
+        void main() {
+          fragColor0 = vec4(1.0, 1.0, 1.0, 1.0);
+          fragColor1 = vec4(1.0, 1.0, 1.0, 1.0);
+          fragColor2 = vec4(1.0, 1.0, 1.0, 1.0);
+          fragColor3 = vec4(1.0, 1.0, 1.0, 1.0);
+        }
+      `,
+    });
+
+    const colorStateDescriptors: GPUColorStateDescriptor[] = [];
+    for (let i = 0; i < t.params.numColorAttachments; i++) {
+      colorStateDescriptors.push({ format: kFormat });
+    }
+
+    const pipeline = t.device.createRenderPipeline({
+      vertexStage: { module: vertexModule, entryPoint: 'main' },
+      fragmentStage: { module: fragmentModule, entryPoint: 'main' },
+      primitiveTopology: 'triangle-list',
+      rasterizationState: {
+        frontFace: 'ccw',
+      },
+      colorStates: colorStateDescriptors,
+      vertexState: {
+        indexFormat: 'uint32',
+        vertexBuffers: [],
+      },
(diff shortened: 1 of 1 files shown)


src/webgpu/api/operation/render_pass/resolve.spec.ts
```diff
@@ -0,0 +1,204 @@
...
+    const pl = t.device.createPipelineLayout({ bindGroupLayouts: [] });
+    const colorStateDescriptors: GPUColorStateDescriptor[] = [];
+    for (let i = 0; i < t.params.numColorAttachments; i++) {
+      colorStateDescriptors.push({ format: kFormat });
+    }
+
+    const pipeline = t.device.createRenderPipeline({
+      vertexStage: { module: vertexModule, entryPoint: 'main' },
+      fragmentStage: { module: fragmentModule, entryPoint: 'main' },
+      layout: pl,
```
austinEng, line 86: nit: the layout member should be optional and could be omitted

src/webgpu/api/operation/render_pass/resolve.spec.ts
```diff
@@ -0,0 +1,204 @@
...
+          })
+        );
+
+        // Clear to black for the load operation. After the draw, the top left half of the attachment
+        // will be white and the bottom right half will be black.
+        renderPassColorAttachmentDescriptors.push({
+          attachment: colorAttachments[i].createView(),
+          loadValue: { r: 0.0, g: 0.0, b: 0.0, a: 0.0 },
+          storeOp: t.params.storeOperation,
+          resolveTarget: resolveTargets[numResolveTargets].createView({
```
austinEng, line 142: nit: slightly more clear we're using the resolve target we just made if we make make the texture and assign it, like `const resolveTarget = `, pass that resolveTarget here, and then push it into the list after. I don't think we need `numResolveTargets` as a variable because later, we can just iterate the whole list.

src/webgpu/api/operation/render_pass/resolve.spec.ts
```diff
@@ -0,0 +1,204 @@
...
+      t.expectSinglePixelIn2DTexture(
+        resolveTargets[i],
+        kFormat,
+        { x: kSize - 1, y: 0 },
+        {
+          exp: new Uint8Array([0x7f, 0x7f, 0x7f, 0x7f]),
+          slice: t.params.resolveTargetBaseArrayLayer,
+          layout: { mipLevel: t.params.resolveTargetBaseMipLevel },
+        }
+      );
```
austinEng, line 202: nit: I know we have `storeOp.spec.ts` but I think it might be valuable here to check a pixel in the non-resolved attachment to make sure that one was correctly written/cleared. Particularly if there was a resolve target and storeOp were `clear`, I could see there being bugs where the implementation wrongly does a store.

Kangz (approved):
No additional comments

src/webgpu/api/operation/render_pass/resolve.spec.ts
bjjones: Done.

src/webgpu/api/operation/render_pass/resolve.spec.ts
bjjones: Done.

src/webgpu/api/operation/render_pass/resolve.spec.ts
bjjones: I agree it would be nice to test - but I'm not seeing a non-complex way to do this.
1. We can't copy from a multisampled texture.
2. We can't render to both multisampled and single-sampled textures in the same pass.

So to check this for correctness, I guess we could render multisampled without resolve, then do a subsequent renderpass that samples from the multisampled texture and renders it into a single sampled texture. WDYT?

austinEng: Another option is to do a texel fetch in a compute shader, which might be more complex than we want for this test. I'd be okay with a TODO for now, and when we implement better helpers to check texture contents we can improve. @kainino0x WDYT?

kainino0x: I do think we should test this. Let's leave a TODO.

We're starting to run into more situations where we need to be able to test values that can't be copied. Someone should own an item to implement this as completely as possible (I guess it can be me) before we accumulate too much tech debt.

kainino0x: (I'm guessing such a helper would need to be glsl-dependent for now, unfortunately.)

src/webgpu/api/operation/render_pass/resolve.spec.ts
```diff
@@ -0,0 +1,206 @@
...
+  - resolveTarget base array layer set to {‘0’, base layer > '0'} for 2D textures
+  TODO: test all renderable color formats
+`;
+
+import { params, poptions } from '../../../../common/framework/params_builder.js';
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { GPUTest } from '../../../gpu_test.js';
+
+// Test with a zero and non-zero base mip level.
+const kBaseMipLevel: number[] = [0, 1];
```
kainino0x, line 16: nit: Not needed as this is the default inferred type (and the type isn't needed for readability since it's clear).

src/webgpu/api/operation/render_pass/resolve.spec.ts
```diff
@@ -0,0 +1,206 @@
...
+
+// Test with different numbers of color attachments.
+type NumColorAttachments = 2 | 4;
+const kNumColorAttachments: NumColorAttachments[] = [2, 4];
```
kainino0x, line 20: nit: Doesn't seem useful to constrain the type to 2|4 here.

src/webgpu/api/operation/render_pass/resolve.spec.ts
kainino0x, line 16: I think it would be better to inline these lists of options in the `poptions` calls, to put them closer to where they're used. Unless you intend to use these more than once and really want them to stay in sync.

src/webgpu/api/operation/render_pass/resolve.spec.ts
```diff
@@ -0,0 +1,206 @@
...
+const kSize = 4;
+const kFormat: GPUTextureFormat = 'rgba8unorm';
+export const g = makeTestGroup(GPUTest);
+
+g.test('render_pass_resolve')
+  .params(
+    params()
+      .combine(poptions('numColorAttachments', kNumColorAttachments))
+      .combine(poptions('slotsToResolve', kSlotsToResolve))
+      .combine(poptions('storeOperation', kStoreOps))
```
kainino0x, line 41: With inline values, the trick to get the types to infer properly here is:
```suggestion
      .combine(poptions('storeOperation', ['clear', 'store'] as const))
```

src/webgpu/api/operation/render_pass/resolve.spec.ts
```diff
@@ -0,0 +1,206 @@
...
+      },
+      sampleCount: 4,
+    });
+
+    const resolveTargets: GPUTexture[] = [];
+    const renderPassColorAttachmentDescriptors: GPURenderPassColorAttachmentDescriptor[] = [];
+
+    // The resolve target must be the same size as the color attachment. If we're resolving to mip
+    // level 1, the resolve target base mip level should be 2x the color attachment size.
+    const kResolveTargetSize: number = t.params.resolveTargetBaseMipLevel === 1 ? kSize * 2 : kSize;
```
kainino0x, line 102: This can be computed more simply as:
```suggestion
    const kResolveTargetSize = kSize << t.params.resolveTargetBaseMipLevel;
```

bjjones: Done.

src/webgpu/api/operation/render_pass/resolve.spec.ts
bjjones: Done.

src/webgpu/api/operation/render_pass/resolve.spec.ts
bjjones: Done.

src/webgpu/api/operation/render_pass/resolve.spec.ts
bjjones: Done.

src/webgpu/api/operation/render_pass/resolve.spec.ts
bjjones: Added TODO at the top.

PR #280, opened 2020-09-04
Add a very lightweight preprocessor using JS template strings

kainino0x:
Example:
```typescript
    const shader = pp`
${pp._if(expr)}
  const x: ${type} = ${value};
${pp._elif(expr)}
${pp.__if(expr)}
...
${pp.__else}
...
${pp.__endif}
${pp._endif}`;
```

---

**OLD:**

The preprocessor supports only %if/elif/else/endif, relying on the surrounding JS environment for everything else you would do in a preprocessor (defines and expression evaluation). Boolean values for %if/elif are passed into the preprocessor via template substitutions (`${}`) meaning they can have arbitrary JS expressions (and are typechecked to be boolean).

It requires proper indentation of preprocessor directives (`%if %%if %%endif %endif`) for readability.

`%` is used because `#` is the comment delimiter in WGSL.

Example:
```
const source = pp`
a
%if ${somecondition} #comment
b
%else
c
%%if ${anothercondition} #comment
d
%%elif ${yetanothercondition}
e
%%else # more comments
f
%%endif
g
%endif #comment
h
`;
```


src/common/framework/preprocessor.ts
```diff
@@ -0,0 +1,101 @@
...
+
+export function pp(parts: TemplateStringsArray, ...predicates: readonly boolean[]): string {
+  let result = '';
+
+  const stateStack: State[] = [State.Passing];
+
+  for (let partIndex = 0; partIndex < parts.length; ++partIndex) {
+    const lines = parts[partIndex].split('\n');
+
+    checkDirectiveComment(partIndex, lines[0]);
```
dneto0, line 23: I don't understand the motivation having this requirement:

The first line has to be empty, or only spaces, or only spaces followed by a #-initiated comment.

Maybe I'm not reading that correctly?

dneto0 (commented):
This looks really nice.

src/common/framework/preprocessor.ts
kainino0x: That's correct, and I admit it's kind of silly. I initially did it mainly because it simplified the implementation: the first line of the first `part`*s isn't treated differently from the first line of other `part`s.

\* where `part` is e.g. `"# c1\nA\n#if "` or `" # c2\nB\n#endif"` as parsed by JS in:
```js
pp` # c1
A
#if ${true} # c2
B
#endif`
```

The line with `# c1` can be treated like it's a directive-tail comment just like `# c2`.

Also because it forces a style where there's no code on the first line of an expression that looks like:

```js
    const x = pp`line 1 <- code here wouldn't line up with the rest of the code
line 2 <- lines up
line 3
`;
```

Though maybe it would be nice to allow a style like this. I could easily use the number of spaces at the beginning of the second line to decide the indentation of the directives
```
    const y = pp`
      line 2
      line 3
      %if <- this has spaces before it, so currently treated like plain text, not as a directive
    `;
```

kainino0x:
I wrote an alternate implementation. Example with the old version:

```ts
    const shader = pp`
%if ${expr}
  const x: ${type} = ${value};
%elif ${expr}
  more stuff
%endif`;
```
- Enforces nesting syntax (%if %%if %%endif %endif)
- Prettier/easier to read IMO
- Line based
- Could easily gain functionality for having indentation inside the string match the JS outside the string
- Can't be used inline; requires e.g. `const ${condition ? 'x' : 'y'}: ${type} = ${value};`

Example with the new version:

```ts
    const shader = pp`
${pp.if(expr)}
  const x: ${type} = ${value};
${pp.elif(expr)}
  more stuff
${pp.endif}`;
```
- Simpler implementation, much easier to reason about (no parsing, no comment handling; simplifies dealing with substitutions (type/value above))
- Stronger types (`pp.if` et al have a unique type which can't be confused with string/number)
- Emits more stray newlines (before and after the `${pp.x}`)
- Non-line-based, can be used inline: `const ${pp.if(true)}x${pp.else}y${pp.endif}: ${type} = ${value};`

kainino0x:
@dneto0, maybe @austinEng @kvark WDYT?

kvark:
Dumb question, this is a preprocessor for... WGSL?
In this case, could this be avoided by using the specialization constants that we agreed to be present in WGSL?

kainino0x:
It's intended for WGSL in this project, yes. It's orthogonal to injecting constants (which can already be done with JS's `${}` template syntax anyway, without this preprocessor) or even injecting identifier or type names (which can't be done with specialization constants, but can be done with templates).

The most runtime-generated shader we have right now is: https://github.com/gpuweb/cts/blob/glsl-dependent/src/webgpu/shader/execution/robust_access_vertex.spec.ts#L359 though admittedly it doesn't benefit from this preprocessor. That said you can imagine this being used for anything that a GLSL/C preprocessor would normally be used for.

kainino0x:
Alright, I've cleaned up v2 and made it better, and replaced v1. I'd like to land this if there aren't objections; I expect it to find use over time as we write WGSL tests. If not, we'll remove or replace it.

@austinEng could you review?

src/common/framework/preprocessor.ts
```diff
@@ -0,0 +1,130 @@
...
+          : this.predicate
+          ? State.Passing
+          : State.Seeking,
+    });
+  }
+}
+
+class ElseIf extends If {
+  applyTo(stack: StateStack) {
+    const { allowsFollowingElse, state: siblingState } = stack.pop()!;
```
austinEng, line 58: length assertion?

src/common/framework/preprocessor.ts
```diff
@@ -0,0 +1,130 @@
...
+      stack.push({ allowsFollowingElse: true, state: State.Skipping });
+    } else {
+      super.applyTo(stack);
+    }
+  }
+}
+
+class Else extends Directive {
+  applyTo(stack: StateStack) {
+    const { allowsFollowingElse, state: siblingState } = stack.pop()!;
```
austinEng, line 71: length assertion?

austinEng (approved):
looks simple enough. Two overall thoughts:
 - do we really need $, $$, $$$, etc. ? Couldn't we just count that you don't push/pop the stack too much?
 - Although the $, $$, $$$, is nice for spacing/"indentation". Maybe don't use the $ token though because it looks messy next to the ${ ... }

kainino0x:
fyi @Kangz since you're working on porting shaders. Just in case this turns out to be useful.

Kangz:
Thanks. I think most of the tests are either so simple they don't need the pp, or so complicate (robust access) that they cannot use it.

PR #282, opened 2020-09-08
Add operation tests of CopyBufferToBuffer

Jiawei-Shao:


src/webgpu/api/operation/command_buffer/copies.spec.ts
```diff
@@ -1,33 +1,131 @@
...
+  .params(
+    params()
+      .combine(poptions('srcOffset', [0, 4, 8, 16]))
+      .combine(poptions('dstOffset', [0, 4, 8, 16]))
+      .combine(poptions('copySize', [0, 4, 8, 16]))
+  )
+  .fn(async t => {
+    const { srcOffset, dstOffset, copySize } = t.params;
+
+    const kSrcBufferSizes = [srcOffset + copySize, srcOffset + copySize + 8];
```
Kangz, line 38: Maybe these should be in the parametrization so it's clear we iterate over them as well.

Jiawei-Shao: I just don't want to test invalid parameters here. For example, srcBufferSize == 8 is not legal for the cases that "srcOffset + copySize > 8". Computing buffer sizes in this way can ensure the buffer size is legal in all the tests. What do you think?

Kangz: Maybe there's a way to put the computations in the params list? But if that's impractical let's ignore this comment.

kainino0x: You can do something like this in the .combine()-chain:

```
      .expand(p => poptions('srcBufferSizes', [p.srcOffset + p.copySize, p.srcOffset + p.copySize + 8]))
      .expand(p => poptions('dstBufferSizes', [p.dstOffset + p.copySize, p.dstOffset + p.copySize + 8]))
```

Jiawei-Shao: Done. Thanks @kainino0x  for your suggestions!

Kangz (approved):

kainino0x (approved):
LGTM!

PR #283, opened 2020-09-08
Add texture usage validation scope test

Richard-Yunchao:
Resource usages validation should be done per each render pass.
Inside a pass, resource usage conflict should always validated.
But resource usage conflict between passes is not a problem.

### DIFF
--- a/src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
+++ b/src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
@@ -845,49 +845,28 @@ g.test('validation_scope')
   )
   .fn(async t => {
     const { scope } = t.params;
+
+    // Create two bind groups. Resource usages conflict between these two bind groups. But resource
+    // usage inside each bind group doesn't conflict.
     const view = t
       .createTexture({ usage: GPUTextureUsage.STORAGE | GPUTextureUsage.SAMPLED })
       .createView();
     const bindGroup0 = t.createBindGroup(0, view, 'sampled-texture', undefined);
     const bindGroup1 = t.createBindGroup(0, view, 'writeonly-storage-texture', 'rgba8unorm');
 
-    // Create pipeline. Note that unused bindings should be tracked too.
-    const wgslVertex = `
-      fn main() -> void {
-        return;
-      }
-
-      entry_point vertex = main;
-    `;
-    const wgslFragment = `
-      fn main() -> void {
-        return;
-      }
-
-      entry_point fragment = main;
-    `;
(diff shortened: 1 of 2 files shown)


src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
```diff
@@ -818,3 +821,76 @@ g.test('unused_bindings_in_pipeline')
       encoder.finish();
     });
   });
+
+g.test('validation_scope')
+  .params(
+    poptions('scope', ['no-draw', 'same-draw', 'different-draws', 'different-passes'] as const)
```
kainino0x, line 827: IMO this test would be much easier to read as 4 separate tests (`'validation_scope,no_draw`, etc.) with all of the common code in a helper function. It's hard to see exactly how the different `scope` values result in different function call orders.

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
```diff
@@ -818,3 +821,76 @@ g.test('unused_bindings_in_pipeline')
...
+      },
+      fragmentStage: {
+        module: t.device.createShaderModule({
+          code: wgslFragment,
+        }),
+        entryPoint: 'main',
+      },
+      primitiveTopology: 'triangle-list',
+      colorStates: [{ format: 'rgba8unorm' }],
+    });
```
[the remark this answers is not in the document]
Richard-Yunchao: Done. I meant to do this but forgot it.

src/webgpu/api/validation/resource_usages/textureUsageInRender.spec.ts
Richard-Yunchao: It looks like the code is too simple to construct a helper function. Because the common part is just a few lines to create a texture, two bind groups, a render pipeline, a command encoder, a render pass, and we need to return almost all of these objects because they are used later. BTW, can we return an array of different objects? 

I added some comments. It might help to understand the code.

kainino0x: You can have the helper return `{ pipeline: GPURenderPipeline, bindGroup0: GPUBindGroup, bindGroup1: GPUBindGroup }` and destructure it with `const { pipeline, bindGroup0, bindGroup1 } = helper();`. Is that what you had in mind?

Richard-Yunchao: Yeah. It works!

kainino0x (approved):
LGTM, thanks for the refactor!

PR #285, opened 2020-09-10
Validation tests for copyImageBitmapToTexture.

shaoboyan:
Add validation tests for copyImageBitmapToTexture. Left TODOs to
cover more imageBitmap source and more texture dimensions.

### DIFF
--- a/src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
+++ b/src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
@@ -1,0 +1,409 @@
+export const description = `
+copyImageBitmapToTexture Validation Tests in Queue.
+
+Test Coverage:
+- For source.imagBitmap:
+  - imageBitmap generated from ImageData:
+    - Check that an error is generated when imageBitmap is closed. Otherwise, no error should be
+      generated.
+
+- For destination.texture:
+  - For 2d destination textures:
+    - Check that an error is generated when texture is in destroyed state. Otherwise, no error should
+      be generated.
+    - Check that an error is generated when texture is created without usage COPY_DST. Otherwise,
+      no error should be generated.
+    - Check that an error is generated when sample count is not 1. Otherwise, no error should be
+      generated.
+    - Check that an error is generated when mipLevel is too large. Otherwise, no error should be
+      generated.
+    - Check that an error is generated when texture format is not valid. Otherwise, no error should
+      be generated.
+
+- For copySize:
+  - Check that an error is generated when copySize has 0 in width, height or depth. Otherwise,
+    no error should be generated.
+  - Check that an error is generated when destination.texture.origin + copySize is too large.
+    Otherwise, no error should be generated.
+
+TODO: more test to cover source.imageBitmap generated from different source.
+TODO: 1d and 3d texture
+`;
+
+import { poptions, params } from '../../../../common/framework/params_builder.js';
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { unreachable } from '../../../../common/framework/util/util.js';
+import { kAllTextureFormats, kTextureUsages } from '../../../capability_info.js';
+import { ValidationTest } from '../validation_test.js';
+
+const DEFAULT_BYTES_PER_PIXEL = 4;
+const DEFAULT_WIDTH = 4;
+const DEFAULT_HEIGHT = 4;
+const DEFAULT_DEPTH = 4;
+const DEFAULT_MIPLEVEL_COUNT = 1;
+
+const kValidTextureFormatsForCopyIB2T = [
+  'rgba8unorm',
+  'rgba8unorm-srgb',
+  'bgra8unorm',
+  'bgra8unorm-srgb',
+  'rgb10a2unorm',
+  'rgba16float',
+  'rgba32float',
+  'rg8unorm',
+  'rg16float',
+];
+
+interface WithDimension {
+  dimension: GPUTextureDimension;
+}
+
+// Helper function to generate correct texture size.
+function copySizeTestForDstTextureSize({ dimension }: WithDimension) {
+  switch (dimension) {
+    case '1d':
+      return poptions('dstTextureSize', [
+        { width: DEFAULT_WIDTH, height: 1, depth: 1 },
+        { width: DEFAULT_WIDTH + 1, height: 1, depth: 1 },
+      ]);
+    case '2d':
+      return poptions('dstTextureSize', [
+        { width: DEFAULT_WIDTH, height: DEFAULT_HEIGHT, depth: 1 },
+        { width: DEFAULT_WIDTH + 1, height: DEFAULT_HEIGHT + 1, depth: 1 },
+      ]);
+    case '3d':
+      return poptions('dstTextureSize', [
+        { width: DEFAULT_WIDTH, height: DEFAULT_HEIGHT, depth: DEFAULT_DEPTH },
+        { width: DEFAULT_WIDTH + 1, height: DEFAULT_HEIGHT + 1, depth: DEFAULT_DEPTH + 1 },
+      ]);
+  }
+}
+
+// Generate source origin value[valid, invalid] which are GPUOrigin2D type.
+function generateSrcOriginValue() {
+  const value = [];
+  // Values for each dimensions are valid or too large.
+  for (let i = 0; i <= DEFAULT_WIDTH; i += DEFAULT_WIDTH) {
+    for (let j = 0; j <= DEFAULT_HEIGHT; j += DEFAULT_HEIGHT) {
+      value.push({ x: i, y: j });
+    }
+  }
+
+  return value;
+}
+
+// Generate dst origin value[valid, invalid] which are GPUOrigin3D type
+function generateDstOriginValue() {
+  const value = [];
+  // Values for each dimensions are valid or too large.
+  for (let i = 0; i <= DEFAULT_WIDTH; i += DEFAULT_WIDTH) {
+    for (let j = 0; j <= DEFAULT_HEIGHT; j += DEFAULT_HEIGHT) {
+      for (let k = 0; k <= DEFAULT_DEPTH; k += DEFAULT_DEPTH) {
+        value.push({ x: i, y: j, z: k });
+      }
+    }
+  }
+
+  return value;
+}
+
+// Generate copySizes[valid, invalid], which are GPUExtent3D type
+function generateCopySizeValue() {
+  const value = [];
+  for (let i = 0; i <= DEFAULT_WIDTH; i += DEFAULT_WIDTH) {
+    for (let j = 0; j <= DEFAULT_HEIGHT; j += DEFAULT_HEIGHT) {
+      for (let k = 0; k <= DEFAULT_DEPTH; k += DEFAULT_DEPTH) {
+        value.push({ width: i, height: j, depth: k });
+      }
+    }
+  }
+  return value;
+}
+
+class CopyImageBitmapToTextureTest extends ValidationTest {
+  private getImageData(width: number, height: number): ImageData {
+    const pixelSize = DEFAULT_BYTES_PER_PIXEL * width * height;
+    const imagePixels = new Uint8ClampedArray(pixelSize);
+    for (let i = 0; i < pixelSize; ++i) {
+      imagePixels[i] = i;
+    }
+    return new ImageData(imagePixels, width, height);
+  }
+
+  async getImageBitmapCopyView(
+    width: number,
+    height: number,
+    origin: GPUOrigin2D,
+    closed: boolean
+  ): Promise<GPUImageBitmapCopyView> {
+    const imageBitmap = await createImageBitmap(this.getImageData(width, height));
+    if (closed) {
+      imageBitmap.close();
+    }
+    return { imageBitmap, origin };
+  }
+
+  // Get default source copy view with Default width, height and correct origin.
+  async getDefaultImageBitmapCopyView(): Promise<GPUImageBitmapCopyView> {
+    return this.getImageBitmapCopyView(DEFAULT_WIDTH, DEFAULT_HEIGHT, { x: 0, y: 0 }, false);
+  }
+
+  getTextureCopyView(
+    size: GPUExtent3D,
+    sampleCount: number,
+    dimension: GPUTextureDimension,
+    format: GPUTextureFormat,
+    usage: GPUTextureUsageFlags,
+    mipLevel: number,
+    origin: GPUOrigin3D,
+    destroyed: boolean
+  ): GPUTextureCopyView {
+    // In generated tests, there are some pairs will trigger validation errors here, eg. sampleCount
+    // 4 and BC formats. Filter these validation and let copyImageBitmapToTexture to catch them.
+    this.device.pushErrorScope('validation');
+    const dstTexture = this.device.createTexture({
+      size,
+      sampleCount,
+      dimension,
+      mipLevelCount: DEFAULT_MIPLEVEL_COUNT,
+      format,
+      usage,
+    });
+
+    if (destroyed) {
+      dstTexture.destroy();
+    }
+    this.device.popErrorScope();
+
+    return { texture: dstTexture, mipLevel, origin };
+  }
+
+  // Generate dst texture copy view with default width, height.
+  getTextureCopyViewWithDefaultSize(
+    sampleCount: number,
+    dimension: GPUTextureDimension,
+    format: GPUTextureFormat,
+    usage: GPUTextureUsageFlags,
+    mipLevel: number,
+    origin: GPUOrigin3D,
+    destroyed: boolean
+  ): GPUTextureCopyView {
+    let size: GPUExtent3D = { width: 0, height: 0, depth: 0 };
+    switch (dimension) {
+      case '1d':
+        size = { width: DEFAULT_WIDTH, height: 1, depth: 1 };
+        break;
+      case '2d':
+        size = { width: DEFAULT_WIDTH, height: DEFAULT_HEIGHT, depth: 1 };
+        break;
+      case '3d':
+        size = { width: DEFAULT_WIDTH, height: DEFAULT_HEIGHT, depth: DEFAULT_DEPTH };
+        break;
+      default:
+        unreachable('unknown texture dimension');
+    }
+
+    return this.getTextureCopyView(
+      size,
+      sampleCount,
+      dimension,
+      format,
+      usage,
+      mipLevel,
+      origin,
+      destroyed
+    );
+  }
+
+  // Generate dst texture copy view with default width, height and correct configs.
+  getDefaultTextureCopyView(dimension: GPUTextureDimension): GPUTextureCopyView {
+    return this.getTextureCopyViewWithDefaultSize(
+      1,
+      dimension,
+      'bgra8unorm',
+      GPUTextureUsage.COPY_DST,
+      0,
+      { x: 0, y: 0, z: 0 },
+      false
+    );
+  }
+
+  testRun(
+    imageBitmapCopyView: GPUImageBitmapCopyView,
+    textureCopyView: GPUTextureCopyView,
+    copySize: GPUExtent3D,
+    success: boolean,
+    exceptionName?: string
+  ): void {
+    // CopyImageBitmapToTexture will generate two types of errors. One is exceptions generated out
+    // of Dawn and the other is validation errors from Dawn.
+    if (exceptionName && exceptionName.length !== 0) {
+      this.shouldThrow(exceptionName, () => {
+        this.device.defaultQueue.copyImageBitmapToTexture(
+          imageBitmapCopyView,
+          textureCopyView,
+          copySize
+        );
+      });
+    } else {
+      this.expectValidationError(() => {
+        this.device.defaultQueue.copyImageBitmapToTexture(
+          imageBitmapCopyView,
+          textureCopyView,
+          copySize
+        );
+      }, !success);
+    }
+  }
+}
+
+export const g = makeTestGroup(CopyImageBitmapToTextureTest);
+
+g.test('copyImageBitmapToTexture_source_imageBitmap_copy_view')
+  .params(poptions('closed', [true, false]))
+  .fn(async t => {
+    const { closed } = t.params;
+    const src = await t.getImageBitmapCopyView(
+      DEFAULT_WIDTH,
+      DEFAULT_HEIGHT,
+      { x: 0, y: 0 },
+      closed
+    );
+    const dst = t.getDefaultTextureCopyView('2d');
+    const copySize = { width: 1, height: 1, depth: 1 };
+    const exceptionName = closed ? 'InvalidStateError' : '';
+
+    t.testRun(src, dst, copySize, !closed, exceptionName);
+  });
+
+g.test('copyImageBitmapToTexture_destination_texture_copy_view')
+  .params(
+    params()
+      .combine(poptions('destroyed', [true, false]))
+      .combine(poptions('dimension', ['2d'] as const))
+      .combine(poptions('usage', kTextureUsages))
+      .combine(poptions('sampleCount', [1, 4]))
+      .combine(poptions('mipLevel', [0, 2]))
+      .combine(poptions('format', kAllTextureFormats))
+  )
+  .fn(async t => {
+    const { destroyed, dimension, usage, sampleCount, mipLevel, format } = t.params;
+    const src = await t.getDefaultImageBitmapCopyView();
+    const dst = t.getTextureCopyViewWithDefaultSize(
+      sampleCount,
+      dimension,
+      format,
+      usage,
+      mipLevel,
+      { x: 0, y: 0, z: 0 },
+      destroyed
+    );
+    const copySize = { width: 1, height: 1, depth: 1 };
+
+    let success = true;
+    let exceptionName = '';
+
+    if (destroyed) {
+      success = false;
+    }
+
+    if (!(usage & GPUTextureUsage.COPY_DST)) {
+      // Usage must have GPUTextureUsage.COPY_DST
+      success = false;
+    }
+
+    if (sampleCount > 1) {
+      // Dst texture cannot be multisampled due to imageBitmap.
+      success = false;
+    }
+
+    if (mipLevel > 1) {
+      // MipLevel value cannot exceed dst texture max mipLevel, which is 1 in tests.
+      success = false;
+    }
+
+    if (!kValidTextureFormatsForCopyIB2T.includes(format)) {
+      // Dst texture format needs to be supported in copyImageBitmapToTexture.
+      success = false;
+      exceptionName = 'TypeError';
+    }
+
+    t.testRun(src, dst, copySize, success, exceptionName);
+  });
+
+g.test('copyImageBitmapToTexture_copy_size')
+  .params(
+    params()
+      .combine(poptions('dimension', ['2d'] as const))
+      .expand(copySizeTestForDstTextureSize)
+      .combine(
+        poptions('srcImageBitmapSize', [
+          { width: DEFAULT_WIDTH, height: DEFAULT_HEIGHT }, // case: src size <= dst size
+          { width: DEFAULT_WIDTH + 1, height: DEFAULT_HEIGHT + 1 }, // case: src size >= dst size
+        ])
+      )
+      .combine(poptions('srcOriginValue', generateSrcOriginValue()))
+      .combine(poptions('dstOriginValue', generateDstOriginValue()))
+      .combine(poptions('copySize', generateCopySizeValue()))
+  )
+  .fn(async t => {
+    const {
+      dimension,
+      dstTextureSize,
+      srcImageBitmapSize,
+      srcOriginValue,
+      dstOriginValue,
+      copySize,
+    } = t.params;
+
+    if (dimension !== '2d') {
+      t.skip('Skip test case for now because no 1d, 3d texture supported');
+    }
+
+    const src = await t.getImageBitmapCopyView(
+      srcImageBitmapSize.width,
+      srcImageBitmapSize.height,
+      srcOriginValue,
+      false
+    );
+
+    const dst = t.getTextureCopyView(
+      dstTextureSize,
+      1,
+      dimension,
+      'rgba8unorm',
+      GPUTextureUsage.COPY_DST,
+      0,
+      dstOriginValue,
+      false
+    );
+
+    let success = true;
+    let exceptionName = '';
+
+    if (
+      srcImageBitmapSize.width <= srcOriginValue.x ||
+      srcImageBitmapSize.width - srcOriginValue.x < copySize.width ||
+      srcImageBitmapSize.height <= srcOriginValue.y ||
+      srcImageBitmapSize.height - srcOriginValue.y < copySize.height ||
+      copySize.width === 0 ||
+      copySize.height === 0 ||
+      copySize.depth === 0
+    ) {
+      success = false;
+      exceptionName = 'RangeError';
+    }
+
+    if (
+      dstTextureSize.width <= dstOriginValue.x ||
+      dstTextureSize.width - dstOriginValue.x < copySize.width ||
+      dstTextureSize.height <= dstOriginValue.y ||
+      dstTextureSize.height - dstOriginValue.y < copySize.height ||
+      dstTextureSize.depth <= dstOriginValue.z ||
+      dstTextureSize.depth - dstOriginValue.z < copySize.depth
+    ) {
+      success = false;
+    }
+
+    t.testRun(src, dst, copySize, success, exceptionName);
+  });
--- a/src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
+++ b/src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
@@ -22,7 +22,7 @@ Test Plan:
 TODO: 1d, 3d texture and 2d array textures.
 `;
 
-import { poptions, params } from '../../../../common/framework/params_builder.js';
+import { poptions, params, pbool } from '../../../../common/framework/params_builder.js';
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
 import { kAllTextureFormats, kTextureUsages } from '../../../capability_info.js';
 import { ValidationTest } from '../validation_test.js';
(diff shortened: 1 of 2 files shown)


shaoboyan:
@kainino0x  PTAL, thanks!

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+      generated.
+    - Check that an error is generated when mipLevel is too large. Otherwise, no error should be
+      generated.
+    - Check that an error is generated when texture format is not valid. Otherwise, no error should
+      be generated.
+
+- For copySize:
+  - Check that an error is generated when copySize has 0 in width, height or depth. Otherwise,
+    no error should be generated.
+  - Check that an error is generated when destination.texture.origin + copySize is too large.
```
Kangz, line 26: Maybe test this for multiple mip levels, with and without an offset.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+    Otherwise, no error should be generated.
+
+TODO: more test to cover source.imageBitmap generated from different source.
```
Kangz, line 29: Shouldn't that be in execution tests? Or maybe a single test that checks it is valid for all kind of sources.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+  'rg8unorm',
+  'rg16float',
+];
+
+interface WithDimension {
+  dimension: GPUTextureDimension;
+}
+
+// Helper function to generate correct texture size.
+function copySizeTestForDstTextureSize({ dimension }: WithDimension) {
```
Kangz, line 62: Why not just GPUTextureDimension?

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+
+Test Coverage:
+- For source.imagBitmap:
+  - imageBitmap generated from ImageData:
+    - Check that an error is generated when imageBitmap is closed. Otherwise, no error should be
+      generated.
+
+- For destination.texture:
+  - For 2d destination textures:
+    - Check that an error is generated when texture is in destroyed state. Otherwise, no error should
```
Kangz, line 12: nit: also test for an error texture.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+}
+
+export const g = makeTestGroup(CopyImageBitmapToTextureTest);
+
+g.test('copyImageBitmapToTexture_source_imageBitmap_copy_view')
+  .params(poptions('closed', [true, false]))
+  .fn(async t => {
+    const { closed } = t.params;
+    const src = await t.getImageBitmapCopyView(
+      DEFAULT_WIDTH,
```
Kangz, line 267: nit: can't this just be 1, 1 for this case?

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+    }
+    return { imageBitmap, origin };
+  }
+
+  // Get default source copy view with Default width, height and correct origin.
+  async getDefaultImageBitmapCopyView(): Promise<GPUImageBitmapCopyView> {
+    return this.getImageBitmapCopyView(DEFAULT_WIDTH, DEFAULT_HEIGHT, { x: 0, y: 0 }, false);
+  }
+
+  getTextureCopyView(
```
Kangz, line 151: I'm not sure about all these chained helpers, they don't seem to save so much typing in the tests (especially if taking advantage of the default values in descriptors) and make it harder to know in the test what values are used.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+      dimension,
+      'bgra8unorm',
+      GPUTextureUsage.COPY_DST,
+      0,
+      { x: 0, y: 0, z: 0 },
+      false
+    );
+  }
+
+  testRun(
```
Kangz, line 231: nit: runTest or doTest.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+      // Usage must have GPUTextureUsage.COPY_DST
+      success = false;
+    }
+
+    if (sampleCount > 1) {
+      // Dst texture cannot be multisampled due to imageBitmap.
+      success = false;
+    }
+
+    if (mipLevel > 1) {
```
Kangz, line 320: Maybe have a bit more miplevels, so that the validation check cannot be != 0 but has to be < texture.mipLevelCount.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+
+    t.testRun(src, dst, copySize, !closed, exceptionName);
+  });
+
+g.test('copyImageBitmapToTexture_destination_texture_copy_view')
+  .params(
+    params()
+      .combine(poptions('destroyed', [true, false]))
+      .combine(poptions('dimension', ['2d'] as const))
+      .combine(poptions('usage', kTextureUsages))
```
Kangz, line 284: nit: some of these tests could actually be split. Right now it does a lot of combination for almost 0 increased coaverage.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+    }
+
+    t.testRun(src, dst, copySize, success, exceptionName);
+  });
+
+g.test('copyImageBitmapToTexture_copy_size')
+  .params(
+    params()
+      .combine(poptions('dimension', ['2d'] as const))
+      .expand(copySizeTestForDstTextureSize)
```
Kangz, line 338: We should also test at a miplevel, to check that the miplevel size is correctly taken into account.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
shaoboyan: You're right, I'll remove the chain helpers.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+          copySize
+        );
+      }, !success);
+    }
+  }
+}
+
+export const g = makeTestGroup(CopyImageBitmapToTextureTest);
+
+g.test('copyImageBitmapToTexture_source_imageBitmap_copy_view')
```
kainino0x, line 262: nit: copyImageBitmapToTexture is redundant here, because it's already in the filename. Same for the rest of the tests.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+    const exceptionName = closed ? 'InvalidStateError' : '';
+
+    t.testRun(src, dst, copySize, !closed, exceptionName);
+  });
+
+g.test('copyImageBitmapToTexture_destination_texture_copy_view')
+  .params(
+    params()
+      .combine(poptions('destroyed', [true, false]))
+      .combine(poptions('dimension', ['2d'] as const))
```
kainino0x, line 283: Not useful to make this a param unless you expect to expand it soon. Can just hardcode `'2d'` in the test.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+      exceptionName = 'TypeError';
+    }
+
+    t.testRun(src, dst, copySize, success, exceptionName);
+  });
+
+g.test('copyImageBitmapToTexture_copy_size')
+  .params(
+    params()
+      .combine(poptions('dimension', ['2d'] as const))
```
kainino0x, line 337: same

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
kainino0x: Agree, most of these rules are very separate. For example, implementations wouldn't be likely to catch a destroyed texture, and catch a multisampled texture, but not catch a destroyed multisampled texture.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+      format,
+      usage,
+      mipLevel,
+      { x: 0, y: 0, z: 0 },
+      destroyed
+    );
+    const copySize = { width: 1, height: 1, depth: 1 };
+
+    let success = true;
+    let exceptionName = '';
```
kainino0x, line 304: Can be `let exceptionName: string | undefined;`

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+    const {
+      dimension,
+      dstTextureSize,
+      srcImageBitmapSize,
+      srcOriginValue,
+      dstOriginValue,
+      copySize,
+    } = t.params;
+
+    if (dimension !== '2d') {
```
kainino0x, line 359: Would become unneeded.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+      be generated.
+
+- For copySize:
+  - Check that an error is generated when copySize has 0 in width, height or depth. Otherwise,
+    no error should be generated.
+  - Check that an error is generated when destination.texture.origin + copySize is too large.
+    Otherwise, no error should be generated.
+
+TODO: more test to cover source.imageBitmap generated from different source.
+TODO: 1d and 3d texture
```
kainino0x, line 30: I had a thought reading these tests. The usefulness of copyImageBitmapToTexture for 1d and 3d textures is very limited. I think we should consider disallowing them in the WebGPU spec for MVP, with possible addition later on. ImageBitmap interaction is one of the more complicated parts of an implementation (since it interacts with other parts of the Web platform) so it would be best to avoid extra complexity with minimal benefit. Would you file a gpuweb issue about this?

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+      dimension,
+      'rgba8unorm',
+      GPUTextureUsage.COPY_DST,
+      0,
+      dstOriginValue,
+      false
+    );
+
+    let success = true;
+    let exceptionName = '';
```
kainino0x, line 382: `let exceptionName: string | undefined;`

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+
+    if (
+      srcImageBitmapSize.width <= srcOriginValue.x ||
+      srcImageBitmapSize.width - srcOriginValue.x < copySize.width ||
+      srcImageBitmapSize.height <= srcOriginValue.y ||
+      srcImageBitmapSize.height - srcOriginValue.y < copySize.height ||
```
kainino0x, line 388: Note that these operations operate on `doubles`. So I think it would be valid to write:
```suggestion
      srcOriginValue.x + copySize.width > srcImageBitmapSize.width ||
      srcOriginValue.y + copySize.height > srcImageBitmapSize.height ||
```
which is much easier to read.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+      exceptionName = 'RangeError';
+    }
+
+    if (
+      dstTextureSize.width <= dstOriginValue.x ||
+      dstTextureSize.width - dstOriginValue.x < copySize.width ||
+      dstTextureSize.height <= dstOriginValue.y ||
+      dstTextureSize.height - dstOriginValue.y < copySize.height ||
+      dstTextureSize.depth <= dstOriginValue.z ||
+      dstTextureSize.depth - dstOriginValue.z < copySize.depth
```
kainino0x, line 403: Same here

kainino0x (commented):
Overall test coverage looking good; a few more suggestions.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
shaoboyan: Agree. I'll file one.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
shaoboyan: Yes, and sry for that. Your points is correct because we don't know the wrong param is caught by corresponding validation rule in such a big combination. Working on address this.

kainino0x: I don't think that's an issue, as I think that even with the big combination, there will be one test that tests ONLY each validation rule. But it would reduce the runtime and reduce the big list of validation computations in the test function (which would be easier to understand in isolation, one per test).

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
shaoboyan: Reshape these tests and take mipLevel into consideration into multiple subset tests.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
shaoboyan: (Similar interface used in new commit)
The function is used in .expand template, which need this interface.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
shaoboyan: Done.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
shaoboyan: Done.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
shaoboyan: Done.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
shaoboyan: Done.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
shaoboyan: Ack. And reshape the tests.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
shaoboyan: Done.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,409 @@
+export const description = `
+copyImageBitmapToTexture Validation Tests in Queue.
+
+Test Coverage:
+- For source.imagBitmap:
```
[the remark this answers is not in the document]
shaoboyan: Done.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+- For destination.texture:
+  - For 2d destination textures:
+    - Check that an error is generated when texture is in destroyed state. Otherwise, no error should
+      be generated.
+    - Check that an error is generated when texture is created without usage COPY_DST. Otherwise,
+      no error should be generated.
+    - Check that an error is generated when sample count is not 1. Otherwise, no error should be
+      generated.
+    - Check that an error is generated when mipLevel is too large. Otherwise, no error should be
+      generated.
```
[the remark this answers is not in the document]
shaoboyan: Considering this patch is a bit large, Left a TODO for 2D array.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+TODO: 1d and 3d texture
+`;
+
+import { poptions, params } from '../../../../common/framework/params_builder.js';
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { unreachable } from '../../../../common/framework/util/util.js';
+import { kAllTextureFormats, kTextureUsages } from '../../../capability_info.js';
+import { ValidationTest } from '../validation_test.js';
+
+const DEFAULT_BYTES_PER_PIXEL = 4;
```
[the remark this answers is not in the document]
shaoboyan: Done.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+    case '3d':
+      return poptions('dstTextureSize', [
+        { width: DEFAULT_WIDTH, height: DEFAULT_HEIGHT, depth: DEFAULT_DEPTH },
+        { width: DEFAULT_WIDTH + 1, height: DEFAULT_HEIGHT + 1, depth: DEFAULT_DEPTH + 1 },
+      ]);
+  }
+}
+
+// Generate source origin value[valid, invalid] which are GPUOrigin2D type.
+function generateSrcOriginValue() {
```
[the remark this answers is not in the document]
shaoboyan: Choose to seperate this test into several small test sets to reduce unnecessary tests.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+      { x: 0, y: 0, z: 0 },
+      false
+    );
+  }
+
+  testRun(
+    imageBitmapCopyView: GPUImageBitmapCopyView,
+    textureCopyView: GPUTextureCopyView,
+    copySize: GPUExtent3D,
+    success: boolean,
```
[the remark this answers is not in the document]
shaoboyan: Done.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+    exceptionName?: string
+  ): void {
+    // CopyImageBitmapToTexture will generate two types of errors. One is exceptions generated out
+    // of Dawn and the other is validation errors from Dawn.
+    if (exceptionName && exceptionName.length !== 0) {
```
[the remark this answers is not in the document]
shaoboyan: Done.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+
+  testRun(
+    imageBitmapCopyView: GPUImageBitmapCopyView,
+    textureCopyView: GPUTextureCopyView,
+    copySize: GPUExtent3D,
+    success: boolean,
+    exceptionName?: string
+  ): void {
+    // CopyImageBitmapToTexture will generate two types of errors. One is exceptions generated out
+    // of Dawn and the other is validation errors from Dawn.
```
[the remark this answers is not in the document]
shaoboyan: Done.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
shaoboyan: Done.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
shaoboyan: Done.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
shaoboyan: Done.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
shaoboyan: Done.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
shaoboyan: Done.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
shaoboyan: Softly ping @kainino0x and cc @Kangz

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
shaoboyan: Done.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
shaoboyan: Done

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
shaoboyan: Done.

shaoboyan:
@kainino0x @Kangz  Reshape lots of tests in the original patch and take mipLevel into consideration in multiple subset tests.
PTAL again!

Kangz:
@shaoboyan can you marked conversations as resolved (instead of Done like in Gerrit)?

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,480 @@
...
+
+Test Coverage:
+- For source.imageBitmap:
+  - imageBitmap generated from ImageData:
+    - Check that an error is generated when imageBitmap is closed. Otherwise, no error should be
+      generated.
+
+- For destination.texture:
+  - For 2d destination textures:
+    - Check that an error is generated when texture is in destroyed state. Otherwise, no error should
```
Kangz, line 12: nit: no need to say "Otherwise, no error should be generated." It's kind of obvious from the context.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,480 @@
...
+      no error should be generated.
+    - Check that an error is generated when sample count is not 1. Otherwise, no error should be
+      generated.
+    - Check that an error is generated when mipLevel is too large. Otherwise, no error should be
+      generated.
+    - Check that an error is generated when texture format is not valid. Otherwise, no error should
+      be generated.
+
+- For copySize:
+  - Check that an error is generated when copySize has 0 in width, height or depth. Otherwise,
```
Kangz: As discussed in the Chromium review, this likely shouldn't be an error (and we should check that it doesn't produce an error)

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,480 @@
...
+      size: { width: kDefaultWidth, height: kDefaultHeight, depth: kDefaultDepth },
+      mipLevelCount: kDefaultMipLevelCount,
+      format: 'bgra8unorm',
+      usage: GPUTextureUsage.COPY_DST,
+    });
+
+    let success = true;
+    let exceptionName: string | undefined;
+
+    const dstMipMapSize = getMipMapSize(kDefaultWidth, kDefaultHeight, mipLevel);
```
Kangz, line 346: nit: computeMipMapSize

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,480 @@
...
+    t.runTest(
+      { imageBitmap },
+      { texture: dstTexture },
+      { width: 1, height: 1, depth: 1 },
+      success,
+      success ? '' : 'TypeError'
+    );
+  });
+
+g.test('copy_view_origin')
```
Kangz, line 320: This and `copy_size_OOB ` seem to test the kinda same thing, but it isn't clear what the separation is.

Instead can we have a test for source OOB and a test for destination OOB? The src OOB test would parametrize on copySize and srcOriginValue while the dst OOB test would parametrize on copySize, dstOriginValue and mipLevel.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,480 @@
...
+  }
+
+  return poptions('copySize', [justFitCopySize]);
+}
+
+class CopyImageBitmapToTextureTest extends ValidationTest {
+  getImageData(width: number, height: number): ImageData {
+    const pixelSize = kDefaultBytesPerPixel * width * height;
+    const imagePixels = new Uint8ClampedArray(pixelSize);
+    for (let i = 0; i < pixelSize; ++i) {
```
Kangz, line 135: nit: the content doesn't matter.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,480 @@
...
+  return poptions('dstOriginValue', [
+    { x: 0, y: 0, z: 0 },
+    { x: origin.mipWidth, y: 0, z: 0 },
+    { x: 0, y: origin.mipHeight, z: 0 },
+    { x: 0, y: 0, z: kDefaultDepth },
+  ]);
+}
+
+// Helper function to generate copySize to take mipLevel, origin into consideration
+function generateCopySize({
```
Kangz, line 88: This could generate a copySizes just for a single target texture level, with the rework of the OOB tests suggested below.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -0,0 +1,480 @@
...
+  dstTextureSize: GPUExtent3DDict;
+  srcOriginValue: GPUOrigin2DDict;
+  dstOriginValue: GPUOrigin3DDict;
+}
+
+// Helper function to generate dst origin value to take mipLevel into consideration.
+function generateDstOriginValue({ mipLevel }: WithMipLevel) {
+  const origin = getMipMapSize(kDefaultWidth, kDefaultHeight, mipLevel);
+
+  return poptions('dstOriginValue', [
```
Kangz, line 79: Maybe also test origins just before the mipLevel's edge?

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
shaoboyan: Left this one for double checking whether it is resolved.

shaoboyan:
@Kangz @kainino0x  CTS updated. PTAL !

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
Kangz: Either way seems fine, but I'd rather have a polished version of 2D copyImageBitmapToTexture for the MVP than something that does 1D / 3D but isn't as nice. Also the 0 copy path investigation is probably more important than support for 1D and 3D in copyIB2T.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -65,76 +57,104 @@ interface WithMipLevel {
...
-    width:
-      dstMipMapSize.mipWidth < srcImageSize.width ? dstMipMapSize.mipWidth : srcImageSize.width,
-    height:
-      dstMipMapSize.mipHeight < srcImageSize.height ? dstMipMapSize.mipHeight : srcImageSize.height,
-    depth: kDefaultDepth,
+// Helper function to generate copySize for dst OOB test
+function generateCopySizeForDstOOB({ mipLevel, dstOriginValue }: WithDstOriginMipLevel) {
+  const dstMipMapSize = computeMipMapSize(kDefaultWidth, kDefaultHeight, mipLevel);
+  const origin = {
+    x: dstOriginValue.x ? dstOriginValue.x : 0,
```
Kangz, line 114: I'm not sure I understand why this is needed, does it happen that `dstOriginValue.x` is undefined?

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -207,7 +227,7 @@ g.test('destination_texture_state')
     t.runTest(
       { imageBitmap },
       { texture: dstTexture },
-      { width: 1, height: 1, depth: 1 },
+      { width: 0, height: 0, depth: 0 },
```
Kangz, line 230: Maybe we could have a poptions for this that gives both the noop copy and actual copy cases?

Kangz (commented):
Thanks for the updates, it looks much better!

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
shaoboyan: Yes. It will report the error that x can be undefined.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
shaoboyan: Ack.

shaoboyan:
@kainino0x @Kangz Update to resolve some comments. PTAL, thanks!

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
Kangz: @kainino0x any idea what could be done in Typescript to avoid this issue? If not LGTM

Kangz (approved):

kainino0x (approved):
Looking great! Just a bunch of nits

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -252,158 +183,286 @@ class CopyImageBitmapToTextureTest extends ValidationTest {
...
+      { imageBitmap },
+      { texture: dstTexture },
+      noopCopy ? { width: 0, height: 0, depth: 0 } : { width: 1, height: 1, depth: 1 },
+      !closed,
+      closed ? 'InvalidStateError' : ''
+    );
   });
 
-g.test('copyImageBitmapToTexture_destination_texture_copy_view')
+g.test('destination_texture_state')
```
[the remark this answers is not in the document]
shaoboyan: Comment addressed but pls double check.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -54,190 +46,129 @@ const kValidTextureFormatsForCopyIB2T = [
...
+  // OOB origin fails even with noop copy.
+  if (origin.x > kDefaultWidth || origin.y > kDefaultHeight) {
+    return poptions('copySize', [{ width: 0, height: 0, depth: 0 }]);
   }
 
-  return value;
+  const justFitCopySize = {
+    width: kDefaultWidth - origin.x,
+    height: kDefaultHeight - origin.y,
+    depth: 1,
```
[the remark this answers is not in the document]
shaoboyan: I'm sry, not quite understand this comment. I tried to remove the depth and it report an error.

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
shaoboyan: Thanks! Works perfectly!

shaoboyan:
@kainino0x  Thanks for reviewing.
Landed most comments but pls take another look!

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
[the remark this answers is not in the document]
kainino0x: The suggestion is to write a helper function with a signature like that, and put it in the common ValidationTest fixture so other tests can share it.

src/webgpu/api/validation/validation_test.ts
```diff
@@ -3,6 +3,30 @@ import { BindableResource } from '../../capability_info.js';
...
+  ): GPUTexture {
+    let texture: GPUTexture;
+    switch (state) {
+      case 'valid':
+        texture = this.device.createTexture(descriptor);
+        break;
+      case 'invalid':
+        this.device.pushErrorScope('validation');
+        texture = this.device.createTexture(descriptor);
+        this.device.popErrorScope();
```
kainino0x, line 18: This case should just return getErrorTexture().

src/webgpu/api/validation/queue/copyImageBitmapToTexture.spec.ts
```diff
@@ -210,68 +201,50 @@ g.test('source_imageBitmap_state')
...
   )
   .fn(async t => {
-    const { isErrDstTexture, noopCopy } = t.params;
+    const { state, copySize } = t.params;
     const imageBitmap = await createImageBitmap(t.getImageData(1, 1));
-    const correctTexture = t.device.createTexture({
-      size: { width: 1, height: 1, depth: 1 },
+    const descriptor: GPUTextureDescriptor = {
+      size:
+        state === 'invalid' ? { width: 0, height: 0, depth: 0 } : { width: 1, height: 1, depth: 1 },
```
kainino0x, line 226: With the suggested change inside createTextureWithState, you can change this back to just `size: { width: 1, height: 1, depth: 1 }`.

kainino0x (approved):
LGTM with a small change!

src/webgpu/api/validation/validation_test.ts
shaoboyan: Done. And change a bit more for this helper function (make descriptor as optional and have a default descriptor in it). Pls take a look at whether this change is Ok, thanks!

kainino0x (approved):

PR #286, opened 2020-09-11
Add debug groups and makers test

egalli:
Adding tests for debug groups and markers.

FYI, it looks like Dawn is missing support for _GPUCommandEncoder.push/popDebugGroup_ and _insertDebugMarker_. You'll have to skip "non-pass" tests.

### DIFF
--- a/src/webgpu/api/validation/encoding/cmds/debug.spec.ts
+++ b/src/webgpu/api/validation/encoding/cmds/debug.spec.ts
@@ -22,7 +22,7 @@ const kEncoderTypes = ['non-pass', 'compute pass', 'render pass', 'render bundle
 type EncoderType = typeof kEncoderTypes[number];
 
 class F extends ValidationTest {
-  #commandEncoder: GPUCommandEncoder | undefined = undefined;
+  private commandEncoder: GPUCommandEncoder | undefined = undefined;
 
   makeAttachmentTexture(): GPUTexture {
     return this.device.createTexture({
@@ -33,18 +33,18 @@ class F extends ValidationTest {
   }
 
   createEncoder(encoderType: EncoderType): Encoder {
-    assert(this.#commandEncoder === undefined);
+    assert(this.commandEncoder === undefined);
     switch (encoderType) {
       case 'non-pass':
         return this.device.createCommandEncoder({});
       case 'render bundle':
         return this.device.createRenderBundleEncoder({ colorFormats: ['rgba8unorm'] });
       case 'compute pass':
-        this.#commandEncoder = this.device.createCommandEncoder({});
-        return this.#commandEncoder.beginComputePass({});
+        this.commandEncoder = this.device.createCommandEncoder({});
+        return this.commandEncoder.beginComputePass({});
       case 'render pass':
-        this.#commandEncoder = this.device.createCommandEncoder({});
-        return this.#commandEncoder.beginRenderPass({
+        this.commandEncoder = this.device.createCommandEncoder({});
+        return this.commandEncoder.beginRenderPass({
           colorAttachments: [
             {
               attachment: this.makeAttachmentTexture().createView(),
@@ -80,10 +80,10 @@ class F extends ValidationTest {
       }
       case 'compute pass':
       case 'render pass': {
-        assert(this.#commandEncoder !== undefined);
+        assert(this.commandEncoder !== undefined);
         (encoder as GPUComputePassEncoder | GPURenderPassEncoder).endPass();
-        commandBuffer = this.#commandEncoder?.finish();
-        this.#commandEncoder = undefined;
+        commandBuffer = this.commandEncoder?.finish();
+        this.commandEncoder = undefined;
         break;
       }
     }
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/encoding/cmds/debug.spec.ts
```diff
@@ -0,0 +1,143 @@
...
+
+g.test('debug_marker_empty_string')
+  .params(poptions('encoderType', kEncoderTypes))
+  .fn(t => {
+    const encoder = t.createEncoder(t.params.encoderType);
+    encoder.insertDebugMarker('');
+    t.finishEncoder(encoder, t.params.encoderType);
+  });
+
+g.test('debug_marker_non_empty_string')
```
kainino0x, line 137: tiny nit: These two tests could be one, with a `.params(poptions('label', ['', 'marker']))`

Similarly, could expand `debug_group_empty_string` in the same way (even though I know that case is already hit by the previous test).

src/webgpu/api/validation/encoding/cmds/debug.spec.ts
```diff
@@ -0,0 +1,143 @@
...
+          colorAttachments: [
+            {
+              attachment: this.makeAttachmentTexture().createView(),
+              loadValue: { r: 1.0, g: 0.0, b: 0.0, a: 1.0 },
+            },
+          ],
+        });
+        pass.executeBundles([bundle]);
+        pass.endPass();
+        commandBuffer = commandEncoder.finish({});
```
kainino0x, line 79: The abstraction you've built here over renderPass and renderBundle is nice and looks very useful. I think it could be useful for other tests as well (we might even have some already that try to do a similar abstraction). Perhaps as a class that creates either a pass or a bundle, exposes it as a GPURenderEncoderBase, and has a finish() method similar to this which returns a GPUCommandBuffer. Perhaps with extended versions for GPUProgrammablePassEncoder (same + compute pass) and `GPUCommandEncoder | GPUProgrammablePassEncoder` (same as what you have here).

Would you be interested in looking at this as a follow-up patch? (Otherwise I'll put it on my infinite todo list.)

src/webgpu/api/validation/encoding/cmds/debug.spec.ts
```diff
@@ -0,0 +1,143 @@
...
+    - Test inserting a debug marker with a non-empty string.
+`;
+
+import { poptions, params } from '../../../../../common/framework/params_builder.js';
+import { makeTestGroup } from '../../../../../common/framework/test_group.js';
+import { assert } from '../../../../../common/framework/util/util.js';
+
+import { ValidationTest } from './../../validation_test.js';
+
+type Encoders = GPUCommandEncoder | GPUProgrammablePassEncoder;
```
kainino0x, line 21: nit: Encoders -> Encoder

src/webgpu/api/validation/encoding/cmds/debug.spec.ts
```diff
@@ -0,0 +1,143 @@
...
+const kEncoderTypes = ['non-pass', 'compute pass', 'render pass', 'render bundle'] as const;
+type EncoderTypes = typeof kEncoderTypes[number];
```
kainino0x, line 23: Huh, I don't think it's occurred to me to do this before. Nice!

nit: EncoderTypes -> EncoderType

src/webgpu/api/validation/encoding/cmds/debug.spec.ts
```diff
@@ -0,0 +1,143 @@
...
+
+class F extends ValidationTest {
+  #commandEncoder: GPUCommandEncoder | undefined = undefined;
```
kainino0x, line 26: Hm, what is this syntax? I'm unfamiliar with it and having a hard time looking it up.

kainino0x (approved):

src/webgpu/api/validation/encoding/cmds/debug.spec.ts
egalli: Sure, I can take a look at it.

src/webgpu/api/validation/encoding/cmds/debug.spec.ts
egalli: It is ECMAScript's private fields syntax. Unlike typescript's private keyword, the fields aren't accessible outside of typescript. I can switch it to private to match the rest of the code base.

reference:  https://devblogs.microsoft.com/typescript/announcing-typescript-3-8/#ecmascript-private-fields

kainino0x: Cool! Fine with me to use the new syntax, but we should look at the generated output and make sure it works on various browsers. (If it outputs the new syntax verbatim, then we need to make sure at least the latest dev versions of Firefox and Safari support it. Otherwise, we should set up Babel to translate to older syntax.)

egalli: I checked the generated output. It looks like the typescript compiler emulates private fields using a WeakMap. WeakMap has pretty good cross-browser support, although it is probably slower than a private field.

kainino0x: Hm, it's too bad that the emulation is a bit complex. I think we should just keep using typescript's private keyword. There isn't much value in true-private in this project since it's not a library, so its types are never exposed outside the project (and the project is all typescript).

src/webgpu/api/validation/encoding/cmds/debug.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+    const shouldError = t.params.popCount !== t.params.pushCount;
+    t.expectValidationError(() => {
+      for (let i = 0; i < t.params.pushCount; ++i) {
+        encoder.pushDebugGroup(`${i}`);
+      }
+      for (let i = 0; i < t.params.popCount; ++i) {
+        encoder.popDebugGroup();
+      }
+      t.finishEncoder(encoder, t.params.encoderType);
+    }, shouldError);
```
Kangz, line 116: nit: the expectValidationError should be just for the finishEncoder

kainino0x: oops, I missed this, thanks for catching it

kainino0x:
Thanks for the revisions!

PR #287, opened 2020-09-16
Add validation tests for CopyTextureToTexture - Part II

Jiawei-Shao:
This patch adds the second part of the validation tests on
CopyTextureToTexture.

Not that the following usages are not tested in WebGPU CTS with this
patch:
- copy with compressed formats (need to create a device with extension)
- copy with aspects (not included in SPEC yet)
- copy with 1D/3D textures

### DIFF
--- a/src/webgpu/api/validation/copyTextureToTexture.spec.ts
+++ b/src/webgpu/api/validation/copyTextureToTexture.spec.ts
@@ -340,7 +340,10 @@ g.test('depth_stencil_copy_restrictions')
     const copyHeight =
       Math.min(srcSizeAtLevel.height, dstSizeAtLevel.height) + copyBoxOffsets.height - copyOrigin.y;
 
+    // Depth/stencil copies must copy whole subresources.
     const isSuccess =
+      copyOrigin.x === 0 &&
+      copyOrigin.y === 0 &&
       copyWidth === srcSizeAtLevel.width &&
       copyHeight === srcSizeAtLevel.height &&
       copyWidth === dstSizeAtLevel.width &&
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/copyTextureToTexture.spec.ts
```diff
@@ -210,3 +236,222 @@ g.test('multisampled_copy_restrictions')
...
+      )
+      .combine(
+        poptions('dstCopyOrigin', [
+          { x: 0, y: 0, z: 0 },
+          { x: 1, y: 0, z: 0 },
+          { x: 0, y: 1, z: 0 },
+          { x: 1, y: 1, z: 0 },
+        ])
+      )
+      .combine(poptions('srcTextureSize', [64, 32]))
```
Kangz, line 292: I just realized that we do a lot of tests with the same size in X and Y, we should probably make rectangular texture so we check we correctly test width with X and height with Y.

src/webgpu/api/validation/copyTextureToTexture.spec.ts
```diff
@@ -210,3 +236,222 @@ g.test('multisampled_copy_restrictions')
...
+      kTextureSize,
+      isSuccess
+    );
+  });
+
+g.test('depth_stencil_copy_restrictions')
+  .params(
+    params()
+      .combine(poptions('format', kDepthStencilFormats))
+      .combine(
```
Kangz, line 276: The parametrization seems a bit much maybe. What we want to test can be split for src and dst (or just have a parameter that inverts them). And for example for src, we want to test for a couple different levels with:

 - a full size copy for the miplevel
 - copies that are like fullsize but with just an offset of 1 in top, bottom, left and right.

So maybe we could have the following:

```
  poptions('copyRectangleOffsets' [
    {x:0, y:0, width:0, height:0},
    {x:1, y:0, width:0, height:0},
    {x:0, y:1, width:0, height:0},
    {x:0, y:0, width:-1, height:0},
    {x:0, y:0, width:0, height:-1},
  ])
```

src/webgpu/api/validation/copyTextureToTexture.spec.ts
```diff
@@ -210,3 +236,222 @@ g.test('multisampled_copy_restrictions')
...
+      { width: copyWidth, height: copyHeight, depth: 1 },
+      isSuccess
+    );
+  });
+
+g.test('copy_ranges')
+  .params(
+    params()
+      .combine(
+        poptions('srcCopyOrigin', [
```
Kangz, line 355: Maybe the same trick with offset could be used so that we test exactly at the correct boundary.

Kangz (approved):

kainino0x (commented):
Just one additional comment

src/webgpu/api/validation/copyTextureToTexture.spec.ts
```diff
@@ -210,3 +236,222 @@ g.test('multisampled_copy_restrictions')
       isSuccess
     );
   });
+
+g.test('uncompressed_texture_format_equality')
+  .params(
+    params()
+      .combine(poptions('srcFormat', kUncompressedTextureFormats))
+      .combine(poptions('dstFormat', kUncompressedTextureFormats))
```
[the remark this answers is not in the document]
Jiawei-Shao: I need to create a device with "textureCompressionBC" enabled to test BC formats, which will add more complexity, so I decide to add all the tests related to BC formats in one PR. What do you think?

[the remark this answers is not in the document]
kainino0x: Oh yes, good point, and we haven't figured out how to do extension tests yet.

src/webgpu/api/validation/copyTextureToTexture.spec.ts
Jiawei-Shao: Done

src/webgpu/api/validation/copyTextureToTexture.spec.ts
Jiawei-Shao: I'm not sure I have fully understood your comments here. PTAL, thanks!

src/webgpu/api/validation/copyTextureToTexture.spec.ts
Jiawei-Shao: I'm not sure I have fully understood your comments here. PTAL, thanks!

src/webgpu/api/validation/copyTextureToTexture.spec.ts
```diff
@@ -210,3 +236,248 @@ g.test('multisampled_copy_restrictions')
...
+      { width: copyWidth, height: copyHeight, depth: 1 },
+      isSuccess
+    );
+  });
+
+g.test('copy_ranges')
+  .params(
+    params()
+      .combine(
+        poptions('copyRectangleOffsets', [
```
Kangz, line 365: nit: copyBoxOffets?

src/webgpu/api/validation/copyTextureToTexture.spec.ts
Kangz: The offsets stuff looks good. Let's not worry about the src vs. dst part.

src/webgpu/api/validation/copyTextureToTexture.spec.ts
```diff
@@ -210,3 +236,248 @@ g.test('multisampled_copy_restrictions')
...
+    );
+    t.TestCopyTextureToTexture(
+      { texture: srcTexture, origin: copyOrigin, mipLevel: srcCopyLevel },
+      { texture: dstTexture, origin: { x: 0, y: 0, z: 0 }, mipLevel: dstCopyLevel },
+      { width: copyWidth, height: copyHeight, depth: 1 },
+      isSuccess
+    );
+  });
+
+g.test('copy_ranges')
```
Kangz, line 361: It would also be good to have a test that checks that for low miplevels the width / height is clamped to one (instead of the naive >> that will produce a 0). Maybe we could have a helper function that takes a GPUExtent3D and compute the logical size at a mipmap level, and we'd use it in several of the tests here.

src/webgpu/api/validation/copyTextureToTexture.spec.ts
```diff
@@ -210,3 +236,248 @@ g.test('multisampled_copy_restrictions')
...
+      mipLevelCount: kMipLevelCount,
+      usage: GPUTextureUsage.COPY_DST,
+    });
+
+    const srcTextureWidthAtLevel = kTextureWidth >> srcCopyLevel;
+    const srcTextureHeightAtLevel = kTextureHeight >> srcCopyLevel;
+    const dstTextureWidthAtLevel = kTextureWidth >> dstCopyLevel;
+    const dstTextureHeightAtLevel = kTextureHeight >> dstCopyLevel;
+
+    const copyOrigin = { x: copyRectangleOffsets.x, y: copyRectangleOffsets.y, z: 0 };
```
Kangz, line 412: nit: what about copyBoxOffsets.z?

src/webgpu/api/validation/copyTextureToTexture.spec.ts
```diff
@@ -210,3 +236,248 @@ g.test('multisampled_copy_restrictions')
...
+      copyRectangleOffsets.height -
+      copyOrigin.y;
+
+    {
+      const isSuccess =
+        copyWidth <= srcTextureWidthAtLevel &&
+        copyHeight <= srcTextureHeightAtLevel &&
+        copyOrigin.x + copyWidth <= dstTextureWidthAtLevel &&
+        copyOrigin.y + copyHeight <= dstTextureHeightAtLevel &&
+        copyOrigin.z + copyRectangleOffsets.depth <= kArrayLayerCount;
```
Kangz, line 429: nit: the semantic of copyBoxOffsets.depth seems a bit different from width and height, maybe we could make it the same?

src/webgpu/api/validation/copyTextureToTexture.spec.ts
```diff
@@ -210,3 +236,248 @@ g.test('multisampled_copy_restrictions')
...
+      isSuccess
+    );
+  });
+
+g.test('copy_ranges')
+  .params(
+    params()
+      .combine(
+        poptions('copyRectangleOffsets', [
+          { x: 0, y: 0, z: 0, width: 0, height: 0, depth: 1 },
```
Kangz, line 366: Aren't we missing testing of exactly corner cases? I'd assume for X we'd do the following:

```
          // full layer copy
          { x: 0, y: 0, z: 0, width: 0, height: 0, depth: 1 },
          // copy too big because x + width is larger than the width
          { x: 1, y: 0, z: 0, width: 0, height: 0, depth: 1 },
          { x: 0, y: 0, z: 0, width: 1, height: 0, depth: 1 },
//....
```

src/webgpu/api/validation/copyTextureToTexture.spec.ts
Jiawei-Shao: Done

src/webgpu/api/validation/copyTextureToTexture.spec.ts
Jiawei-Shao: Done

src/webgpu/api/validation/copyTextureToTexture.spec.ts
Jiawei-Shao: Done

src/webgpu/api/validation/copyTextureToTexture.spec.ts
Jiawei-Shao: Done

src/webgpu/api/validation/copyTextureToTexture.spec.ts
Jiawei-Shao: Done

src/webgpu/api/validation/copyTextureToTexture.spec.ts
kainino0x: src vs dst is already done by just testing both directions in the test function, which lgtm

src/webgpu/api/validation/copyTextureToTexture.spec.ts
kainino0x: Please mark comments as resolved on GitHub when done (and no need to comment "Done")

src/webgpu/api/validation/copyTextureToTexture.spec.ts
```diff
@@ -274,157 +281,311 @@ g.test('depth_stencil_copy_restrictions')
...
-          { width: 8, height: 4, depth: 2 },
-          { width: 16, height: 8, depth: 1 },
-          { width: 16, height: 8, depth: 2 },
-          { width: 16, height: 16, depth: 0 },
-        ])
-      )
-      .combine(poptions('srcCopyLevel', [0, 1]))
-      .combine(poptions('dstCopyLevel', [0, 1]))
-  )
+  .params([
```
kainino0x, line 363: This big list is ok, but consider:
```ts
g.test()
  .params([
    ...params()
      .combine(poptions('copyBoxOffsets', [
        { x: 0, y: 0, z: 0, width: 0, height: 0, depth: -2 },
        // ...
      ]))
      .combine(poptions('srcCopyLevel', [0, 1]))
      .combine(poptions('dstCopyLevel', [0, 1])),
    ...params()
      .combine(poptions('copyBoxOffsets', [
        { x: 0, y: 0, z: 0, width: 1, height: 0, depth: -2 },
        // ...
      ]))
      .combine([{ srcCopyLevel: 0, dstCopyLevel: 0 }])
  ] as const)
```

(Let me know if TypeScript has issues with doing this, I may need to fix something. You may also need more `as const` throughout.)

src/webgpu/api/validation/copyTextureToTexture.spec.ts
```diff
@@ -274,157 +281,311 @@ g.test('depth_stencil_copy_restrictions')
...
+    const dstSizeAtLevel = t.GetPhysicalSubresourceSize(dstTextureSize, dstCopyLevel);
+
+    const copyOrigin = { x: copyBoxOffsets.x, y: copyBoxOffsets.y, z: 0 };
+
+    const copyWidth =
+      Math.min(srcSizeAtLevel.width, dstSizeAtLevel.width) + copyBoxOffsets.width - copyOrigin.x;
+    const copyHeight =
+      Math.min(srcSizeAtLevel.height, dstSizeAtLevel.height) + copyBoxOffsets.height - copyOrigin.y;
+
     const isSuccess =
```
kainino0x, line 343: Took me a bit to figure out why these are the conditions. Suggest:
```suggestion
    // Depth/stencil copies must copy whole subresources.
    const isSuccess =
```

kainino0x, line 343: And perhaps include the `copyOrigin.x === 0 && copyOrigin.y === 0` even if it's redundant.

kainino0x (approved):

Kangz:
Still LGTM, I'll let @kainino0x merge this.

kainino0x (approved):
Simpler combination for copy_ranges LGTM

PR #288, opened 2020-09-16
Resource usage validation in compute

Richard-Yunchao:
PTAL. Thank you!

### DIFF
--- a/src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
+++ b/src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
@@ -133,10 +133,8 @@ class TextureUsageTracking extends ValidationTest {
     bindGroup0: GPUBindGroup;
     bindGroup1: GPUBindGroup;
     encoder: GPUCommandEncoder;
-    renderPass: GPURenderPassEncoder;
-    computePass: GPUComputePassEncoder;
-    renderPipeline: GPURenderPipeline;
-    computePipeline: GPUComputePipeline;
+    pass: GPURenderPassEncoder | GPUComputePassEncoder;
+    pipeline: GPURenderPipeline | GPUComputePipeline;
   } {
     // Create two bind groups. Resource usages conflict between these two bind groups. But resource
     // usage inside each bind group doesn't conflict.
@@ -154,46 +152,38 @@ class TextureUsageTracking extends ValidationTest {
     );
 
     const encoder = this.device.createCommandEncoder();
-    const renderPass = this.beginSimpleRenderPass(encoder, this.createTexture().createView());
-    const computePass = encoder.beginComputePass();
+    const pass = compute
+      ? encoder.beginComputePass()
+      : this.beginSimpleRenderPass(encoder, this.createTexture().createView());
 
     // Create pipeline. Note that bindings unused in pipeline should be validated too.
-    const renderPipeline = this.createNoOpRenderPipeline();
-    const computePipeline = this.createNoOpComputePipeline();
+    const pipeline = compute ? this.createNoOpComputePipeline() : this.createNoOpRenderPipeline();
     return {
       bindGroup0,
       bindGroup1,
       encoder,
-      renderPass,
-      computePass,
-      renderPipeline,
-      computePipeline,
+      pass,
+      pipeline,
     };
   }
 
   setPipeline(
-    renderPass: GPURenderPassEncoder,
-    computePass: GPUComputePassEncoder,
-    renderPipeline: GPURenderPipeline,
-    computePipeline: GPUComputePipeline,
+    pass: GPURenderPassEncoder | GPUComputePassEncoder,
+    pipeline: GPURenderPipeline | GPUComputePipeline,
     compute: boolean
   ) {
     if (compute) {
-      computePass.setPipeline(computePipeline);
+      (pass as GPUComputePassEncoder).setPipeline(pipeline as GPUComputePipeline);
     } else {
-      renderPass.setPipeline(renderPipeline);
+      (pass as GPURenderPassEncoder).setPipeline(pipeline as GPURenderPipeline);
     }
   }
 
-  issueDrawOrDispatch(
-    renderPass: GPURenderPassEncoder,
-    computePass: GPUComputePassEncoder,
-    compute: boolean
-  ) {
+  issueDrawOrDispatch(pass: GPURenderPassEncoder | GPUComputePassEncoder, compute: boolean) {
     if (compute) {
-      computePass.dispatch(1);
+      (pass as GPUComputePassEncoder).dispatch(1);
     } else {
-      renderPass.draw(3, 1, 0, 0);
+      (pass as GPURenderPassEncoder).draw(3, 1, 0, 0);
     }
   }
 }
@@ -401,29 +391,29 @@ g.test('subresources_and_binding_types_combination_for_color')
       ] as const)
       // Every color attachment can use only one single subresource.
       .unless(
-        ({ type0, levelCount0, layerCount0, type1, levelCount1, layerCount1 }) =>
-          (type0 === 'render-target' && (levelCount0 !== 1 || layerCount0 !== 1)) ||
-          (type1 === 'render-target' && (levelCount1 !== 1 || layerCount1 !== 1))
+        p =>
+          (p.type0 === 'render-target' && (p.levelCount0 !== 1 || p.layerCount0 !== 1)) ||
+          (p.type1 === 'render-target' && (p.levelCount1 !== 1 || p.layerCount1 !== 1))
       )
       // All color attachments' size should be the same.
       .unless(
-        ({ type0, type1, baseLevel1 }) =>
-          type0 === 'render-target' && type1 === 'render-target' && baseLevel1 !== BASE_LEVEL
+        p =>
+          p.type0 === 'render-target' && p.type1 === 'render-target' && p.baseLevel1 !== BASE_LEVEL
       )
       .unless(
-        ({ binding0InBundle, binding1InBundle, type0, type1 }) =>
+        p =>
           // We can't set 'render-target' in bundle, so we need to exclude it from bundle.
-          (binding0InBundle && type0 === 'render-target') ||
-          (binding1InBundle && type1 === 'render-target')
+          (p.binding0InBundle && p.type0 === 'render-target') ||
+          (p.binding1InBundle && p.type1 === 'render-target')
       )
       .unless(
-        ({ compute, binding0InBundle, binding1InBundle, type0, type1 }) =>
+        p =>
           // We can't set 'render-target' or bundle in compute.
-          compute &&
-          (binding0InBundle ||
-            binding1InBundle ||
-            type0 === 'render-target' ||
-            type1 === 'render-target')
+          p.compute &&
+          (p.binding0InBundle ||
+            p.binding1InBundle ||
+            p.type0 === 'render-target' ||
+            p.type1 === 'render-target')
       )
   )
   .fn(async t => {
@@ -487,12 +477,12 @@ g.test('subresources_and_binding_types_combination_for_color')
       });
       pass.endPass();
     } else {
-      const computePass = encoder.beginComputePass();
-      const renderPass = t.beginSimpleRenderPass(
-        encoder,
-        type1 === 'render-target' ? view1 : t.createTexture().createView()
-      );
-      const pass = compute ? computePass : renderPass;
+      const pass = compute
+        ? encoder.beginComputePass()
+        : t.beginSimpleRenderPass(
+            encoder,
+            type1 === 'render-target' ? view1 : t.createTexture().createView()
+          );
 
       // Create bind groups. Set bind groups in pass directly or set bind groups in bundle.
       const storageTextureFormat0 = type0 === 'sampled-texture' ? undefined : 'rgba8unorm';
@@ -505,7 +495,8 @@ g.test('subresources_and_binding_types_combination_for_color')
         compute
       );
       if (binding0InBundle) {
-        t.createAndExecuteBundle(0, bindGroup0, renderPass);
+        assert(pass instanceof GPURenderPassEncoder);
+        t.createAndExecuteBundle(0, bindGroup0, pass as GPURenderPassEncoder);
       } else {
         pass.setBindGroup(0, bindGroup0);
       }
@@ -520,7 +511,8 @@ g.test('subresources_and_binding_types_combination_for_color')
           compute
         );
         if (binding1InBundle) {
-          t.createAndExecuteBundle(1, bindGroup1, renderPass);
+          assert(pass instanceof GPURenderPassEncoder);
+          t.createAndExecuteBundle(1, bindGroup1, pass as GPURenderPassEncoder);
         } else {
           pass.setBindGroup(1, bindGroup1);
         }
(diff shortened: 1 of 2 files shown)


Richard-Yunchao:
Do you know why the checks failed? @kainino0x. This change can pass at my local side via "npx grunt pre" with latest code.

kainino0x:
It looks like there was a TLS issue when installing Node in the builder container. Hopefully that's an external issue and will go away on its own.

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
```diff
@@ -844,7 +989,8 @@ g.test('unused_bindings_in_pipeline')
     });
 
     const encoder = t.device.createCommandEncoder();
-    const pass = encoder.beginRenderPass({
+    const computePass = encoder.beginComputePass();
+    const renderPass = encoder.beginRenderPass({
```
kainino0x, line 993: same

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
```diff
@@ -797,17 +899,40 @@ g.test('bindings_in_bundle')
...
     const view = t.createTexture({ usage: GPUTextureUsage.STORAGE }).createView();
-    const bindGroup0 = t.createBindGroup(0, view, 'readonly-storage-texture', '2d', 'rgba8unorm');
-    const bindGroup1 = t.createBindGroup(0, view, 'writeonly-storage-texture', '2d', 'rgba8unorm');
+    const bindGroup0 = t.createBindGroup(
+      0,
+      view,
+      'readonly-storage-texture',
+      '2d',
+      'rgba8unorm',
+      false
```
kainino0x, line 926: Needs to be `compute`?

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
```diff
@@ -853,71 +999,135 @@ g.test('unused_bindings_in_pipeline')
...
+      renderPipeline,
+      computePipeline,
+    } = t.testValidationScope(compute);
+    const pass = compute ? computePass : renderPass;
+    t.setPipeline(renderPass, computePass, renderPipeline, computePipeline, compute);
+    pass.setBindGroup(0, bindGroup0);
+    pass.endPass();
+
+    const renderPass1 = t.beginSimpleRenderPass(encoder, t.createTexture().createView());
+    const computePass1 = encoder.beginComputePass();
```
kainino0x, line 1125: same

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
```diff
@@ -124,33 +127,74 @@ class TextureUsageTracking extends ValidationTest {
...
       '2d',
-      'rgba8unorm'
+      'rgba8unorm',
+      compute
     );
 
     const encoder = this.device.createCommandEncoder();
-    const pass = this.beginSimpleRenderPass(encoder, this.createTexture().createView());
+    const renderPass = this.beginSimpleRenderPass(encoder, this.createTexture().createView());
+    const computePass = encoder.beginComputePass();
```
kainino0x, line 158: You can't beginComputePass and then beginRenderPass on the same encoder; both require that the encoder be in a non-pass state beforehand.

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
```diff
@@ -124,33 +127,74 @@ class TextureUsageTracking extends ValidationTest {
...
+  ): {
     bindGroup0: GPUBindGroup;
     bindGroup1: GPUBindGroup;
     encoder: GPUCommandEncoder;
-    pass: GPURenderPassEncoder;
-    pipeline: GPURenderPipeline;
+    renderPass: GPURenderPassEncoder;
+    computePass: GPUComputePassEncoder;
+    renderPipeline: GPURenderPipeline;
+    computePipeline: GPUComputePipeline;
```
kainino0x, line 139: Can this possibly be made to return:
```suggestion
    pass: GPURenderPassEncoder | GPUComputePassEncoder;
    pipeline: GPURenderPipeline | GPUComputePipeline;
```
instead?

Richard-Yunchao: Some APIs are exposed in GPUProgrammablePassEncoder, for example, setBindGroup. And endPass() exists in both GPURenderPassEncoder and GPUComputePassEncoder. These APIs can be shared between GPURenderPassEncoder and GPUComputePassEncoder. We can reuse the code for both render and compute easily. However, some APIs belong to a specific encoder, for example, setPipeline (both encoders have setPipeline, but these two setPipeline() take different parameters), draw or dispatch, etc. So I need to return renderPass/computePass, renderPipeline/computePipeline. Otherwise, the tests fail when I run "grunt pre" because of type mismatch. For example, it complained that "draw call requires a GPURenderPassEncoder, while you provide GPURenderPassEncoder | GPUComputePassEncoder."

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
Richard-Yunchao: Oops. Looks like we can't share code between render pass and compute pass. Code snippet outside pass, say constructing bind groups, can be shared. I should have run the tests in Chromium.

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
kainino0x: You can `if (pass instanceof GPURenderPassEncoder)` or `assert(pass instanceof GPURenderPassEncoder);` if needed. You can also use unsafe type casts (`pass as GPURenderPassEncoder`) if you can't get the type system to do what you want.

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
kainino0x: I think code can be shared between render and compute pass. You have another case where you do `const pass = compute ? beginComputePass(...) : beginSimpleRenderPass(...)` and that's fine.

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
```diff
@@ -370,9 +416,19 @@ g.test('subresources_and_binding_types_combination_for_color')
           (binding0InBundle && type0 === 'render-target') ||
           (binding1InBundle && type1 === 'render-target')
       )
+      .unless(
+        ({ compute, binding0InBundle, binding1InBundle, type0, type1 }) =>
```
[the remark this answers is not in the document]
Richard-Yunchao: Done.

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
```diff
@@ -556,7 +637,8 @@ g.test('subresources_and_binding_types_combination_for_aspect')
     // Color attachment's size should match depth/stencil attachment's size. Note that if
     // type1 !== 'render-target' then there's no depthStencilAttachment to match anyway.
     const size = SIZE >> baseLevel;
-    const pass = encoder.beginRenderPass({
+    const computePass = encoder.beginComputePass();
+    const renderPass = encoder.beginRenderPass({
```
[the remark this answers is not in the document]
Richard-Yunchao: Done.

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
```diff
@@ -431,24 +487,40 @@ g.test('subresources_and_binding_types_combination_for_color')
       });
       pass.endPass();
     } else {
-      const pass = t.beginSimpleRenderPass(
+      const computePass = encoder.beginComputePass();
+      const renderPass = t.beginSimpleRenderPass(
```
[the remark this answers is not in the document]
Richard-Yunchao: Done.

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
Richard-Yunchao: Done.

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
Richard-Yunchao: Oops. My bad. I must be sleepy at that time... Done.

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
Richard-Yunchao: Done

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
Richard-Yunchao: Agreed! We can use unsafe type cast as you suggested below to share code between render pass and compute pass. Done.

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
Richard-Yunchao: Yeah. unsafe type cast is a good solution that can resolve the problem of sharing code between render pass and compute pass I met here.

Richard-Yunchao:
> It looks like there was a TLS issue when installing Node in the builder container. Hopefully that's an external issue and will go away on its own.

Hmm, you are right. The bot passed this time. Please take another look when you have time. Thank you!

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
```diff
@@ -797,17 +893,40 @@ g.test('bindings_in_bundle')
 g.test('unused_bindings_in_pipeline')
   .params(
     params()
+      // Test compute pass if 'compute' is true. Test render pass otherwise.
```
Kangz, line 896: I'm not sure that it helps to have that comment copy-pasted everywhere. Maybe an explanation at the top of the file would be more helpful.

Kangz (approved):

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
```diff
@@ -85,15 +87,16 @@ class TextureUsageTracking extends ValidationTest {
     view: GPUTextureView,
     bindingType: GPUBindingType,
     dimension: GPUTextureViewDimension,
-    bindingTexFormat: GPUTextureFormat | undefined
+    bindingTexFormat: GPUTextureFormat | undefined,
+    compute: boolean
```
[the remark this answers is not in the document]
Richard-Yunchao: I thought about this as well when I wrote the tests. We do have tests for visibility, which will test the whole visibility (vertex + fragment + compute). But for other tests that are not intended to test visibility, I think it might be a little better to use the visibility it really depends on.

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
```diff
@@ -826,98 +945,145 @@ g.test('unused_bindings_in_pipeline')
...
-        module: t.device.createShaderModule({
-          code: wgslFragment,
-        }),
-        entryPoint: 'main',
-      },
-      primitiveTopology: 'triangle-list',
-      colorStates: [{ format: 'rgba8unorm' }],
-    });
+
+    // TODO: revisit the shader code once 'image' can be supported in wgsl.
```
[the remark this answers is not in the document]
Richard-Yunchao: That's true, if we use getBindGroupLayout when we create bind group, then the unused binding problems seems gone. But we can still specify bind groups we want directly. So I think we need to cover that situation.

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
Richard-Yunchao: Done.

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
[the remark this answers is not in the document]
Kangz: It just requires piping the compute boolean, I think it would be slightly more readable if we didn't have to do it.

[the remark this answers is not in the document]
Richard-Yunchao: What do you mean "it just requires piping the compute boolean", sounds like the tiny change you are asking is to pipe the compute boolean. Do you still think that it is slightly better to remove "compute" parameter and always use COMPUTE | FRAGMENT for visibility for all tests? I will do that if you still think so. It's not a big deal, anyway.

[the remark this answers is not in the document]
Kangz: Readability of code is a lot of small things that add up. If we don't need this boolean, then it's marginally better readability to remove it.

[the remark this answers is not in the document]
Richard-Yunchao: Agreed. I hesitated to add 'compute' parameter too: It could make the code more verbose but I thought it might be good to use the visibility the tests really depends on. It's a trade-off but I chose a different one. It's done right now.

Richard-Yunchao:
Can we merge this pr?

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
```diff
@@ -431,24 +475,28 @@ g.test('subresources_and_binding_types_combination_for_color')
...
+            type1 === 'render-target' ? view1 : t.createTexture().createView()
+          );
 
       // Create bind groups. Set bind groups in pass directly or set bind groups in bundle.
       const storageTextureFormat0 = type0 === 'sampled-texture' ? undefined : 'rgba8unorm';
       const bindGroup0 = t.createBindGroup(0, view0, type0, dimension0, storageTextureFormat0);
       if (binding0InBundle) {
-        t.createAndExecuteBundle(0, bindGroup0, pass);
+        assert(pass instanceof GPURenderPassEncoder);
+        t.createAndExecuteBundle(0, bindGroup0, pass as GPURenderPassEncoder);
```
kainino0x, line 490: nit: `as` shouldn't be needed because `assert` already constrains the type.

```suggestion
        t.createAndExecuteBundle(0, bindGroup0, pass);
```

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
```diff
@@ -431,24 +475,28 @@ g.test('subresources_and_binding_types_combination_for_color')
...
       } else {
         pass.setBindGroup(0, bindGroup0);
       }
       if (type1 !== 'render-target') {
         const storageTextureFormat1 = type1 === 'sampled-texture' ? undefined : 'rgba8unorm';
         const bindGroup1 = t.createBindGroup(1, view1, type1, dimension1, storageTextureFormat1);
         if (binding1InBundle) {
-          t.createAndExecuteBundle(1, bindGroup1, pass);
+          assert(pass instanceof GPURenderPassEncoder);
+          t.createAndExecuteBundle(1, bindGroup1, pass as GPURenderPassEncoder);
```
kainino0x, line 499:
```suggestion
          t.createAndExecuteBundle(1, bindGroup1, pass);
```

kainino0x (approved):
Sorry for the slow review. Got burned out on reviews end of last week.

Richard-Yunchao:
> Sorry for the slow review. Got burned out on reviews end of last week.

No worries. Thanks for your review and your minor changes I missed. BTW, I plan to write tests for memory synchronization for buffer, which is my task on the latest sprint. I need to write shader code to write into storage buffer and then read from that buffer in order to verify the data is correct. So, I wonder that whether wgsl support what I am going to do for tests. I think it might requires that:
  * The definitions about some features are clear enough in wgsl. For example, readonly and writable storage buffer, writing into color buffer and rendering some stuff according data in storage buffer, and reading back, etc.
  * It is not a must but it is better to have if Chromium can support these features in tint. Then we can verify the tests in Chromium. 
Do you know the status of wgsl. Alternatively, I can either wait for a while or use glsl.

kainino0x:
If you just need storage buffers and color buffer I think we can do it. ComputeBoids uses storage:
https://github.com/austinEng/webgpu-samples/blob/d47ceca60910d5ccd2652eafef39eb5f1fcc0eb8/src/examples/computeBoids.ts#L366

However I think(?) readonly storage is to be discussed (soon).

kainino0x:
For shader tests, FYI in case it is helpful: https://github.com/gpuweb/cts/pull/280

Richard-Yunchao:
Thanks for the information, @kainino0x

PR #289, opened 2020-09-18
Significantly speed up build by optimizing n^2 code

kainino0x:
Replaces n^2 comparisons with a `Set<string>`.

### DIFF
--- a/src/common/framework/util/util.ts
+++ b/src/common/framework/util/util.ts
@@ -48,6 +48,14 @@ export function raceWithRejectOnTimeout<T>(p: Promise<T>, ms: number, msg: strin
   return Promise.race([p, rejectOnTimeout(ms, msg)]);
 }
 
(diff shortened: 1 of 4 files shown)


src/common/framework/util/util.ts
```diff
@@ -48,6 +48,14 @@ export function raceWithRejectOnTimeout<T>(p: Promise<T>, ms: number, msg: strin
   return Promise.race([p, rejectOnTimeout(ms, msg)]);
 }
 
+export function sortObjectByKey(v: { [k: string]: unknown }): object {
```
austinEng, line 51: nit: return type  could be more specific
```suggestion
export function sortObjectByKey(v: { [k: string]: unknown }): { [k: string]: unknown } {
```

austinEng (approved):

PR #291, opened 2020-09-22
Test replaced bindings for texture usage validation in compute

Richard-Yunchao:
Replaced bindings are always validated in render pass because removing replaced bindings in Chromium/Dawn implementation is difficult (but it is doable, I think). 

This change added tests for `replaced bindings in compute pass`, I think the validation rule should be: 
  * replaced bindings within the same dispatch (or no dispatch call at all in a compute pass, which is only for test, but not useful in real world app) should be validated, just as what we did for replaced bindings in render. 
  * replaced bindings across dispatch calls (but within the same compute pass) should not be validated, because resource usages validation is done per each dispatch call for compute. So replaced bindings across dispatch calls should be ignored. 

I am not sure whether it is easy to add the validation code for replaced bindings across dispatch calls in Dawn. Per may understanding, bindings are inherited across dispatch calls, so it may require us to inherit and then remove replaced bindings, which might be difficult in current Dawn implementation. I mean, I am worried about the implementation in Dawn.

PTAL. Feel free to add Dzmitry to discuss and review if needed.

### DIFF
--- a/src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
+++ b/src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
@@ -185,6 +185,10 @@ class TextureUsageTracking extends ValidationTest {
       (pass as GPURenderPassEncoder).draw(3, 1, 0, 0);
     }
   }
+
+  isWritableBinding(bindingType: GPUBindingType | 'render-target'): boolean {
+    return bindingType === 'writeonly-storage-texture' ? true : false;
+  }
 }
 
 export const g = makeTestGroup(TextureUsageTracking);
@@ -789,7 +793,7 @@ g.test('replaced_binding')
     pass.setBindGroup(0, bindGroup1);
     pass.endPass();
 
-    let success = bindingType === 'writeonly-storage-texture' ? false : true;
+    let success = !t.isWritableBinding(bindingType);
     // Replaced bindings across dispatch calls should not be validated in compute pass.
     if (compute && callDrawOrDispatch) success = true;
 
@@ -871,9 +875,7 @@ g.test('bindings_in_bundle')
       success = true;
     }
 
-    if (type0 === 'writeonly-storage-texture' && type1 === 'writeonly-storage-texture') {
-      success = true;
-    }
+    if (t.isWritableBinding(type0) && t.isWritableBinding(type1)) success = true;
 
     // Resource usages in bundle should be validated.
     t.expectValidationError(() => {
--- a/src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
+++ b/src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
@@ -789,6 +789,8 @@ g.test('replaced_binding')
     pass.setBindGroup(0, bindGroup1);
     pass.endPass();
 
+    // TODO: If the Compatible Usage List (https://gpuweb.github.io/gpuweb/#compatible-usage-list)
+    // gets programmatically defined in capability_info, use it here, instead of this logic, for clarity.
     let success = bindingType === 'writeonly-storage-texture' ? false : true;
     // Replaced bindings should not be validated in compute pass, because validation only occurs
     // inside dispatch() which only looks at the current resource usages.
--- a/src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
+++ b/src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
@@ -18,19 +18,20 @@ Test Coverage:
         same aspect. Otherwise, no error should be generated.
 
   - Test combinations of two shader stages:
-    - Texture usages in bindings with invisible shader stages should be tracked. Invisible shader
+    - Texture usages in bindings with invisible shader stages should be validated. Invisible shader
       stages include shader stage with visibility none, compute shader stage in render pass, and
       vertex/fragment shader stage in compute pass.
 
   - Tests replaced bindings:
     - Texture usages via bindings replaced by another setBindGroup() upon the same bindGroup index
-      in current scope should be tracked.
+      in render pass should be validated. However, replaced bindings crorss dispatch calls should
+      not be validated.
 
   - Test texture usages in bundle:
-    - Texture usages in bundle should be tracked if that bundle is executed in the current scope.
+    - Texture usages in bundle should be validated if that bundle is executed in the current scope.
 
   - Test texture usages with unused bindings:
-    - Texture usages should be tracked even its bindings is not used in pipeline.
+    - Texture usages should be validated even its bindings is not used in pipeline.
 
   - Test texture usages validation scope:
     - Texture usages should be validated per each render pass. And they should be validated per each
@@ -721,7 +722,7 @@ g.test('shader_stages_and_visibility')
     pass.setBindGroup(0, bindGroup);
     pass.endPass();
 
-    // Texture usages in bindings with invisible shader stages should be tracked. Invisible shader
+    // Texture usages in bindings with invisible shader stages should be validated. Invisible shader
     // stages include shader stage with visibility none, compute shader stage in render pass, and
     // vertex/fragment shader stage in compute pass.
     t.expectValidationError(() => {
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
```diff
@@ -767,15 +774,25 @@ g.test('replaced_binding')
...
+    if (callDrawOrDispatch) {
+      const pipeline = compute ? t.createNoOpComputePipeline() : t.createNoOpRenderPipeline();
+      t.setPipeline(pass, pipeline, compute);
+      t.issueDrawOrDispatch(pass, compute);
+    }
     pass.setBindGroup(0, bindGroup1);
     pass.endPass();
 
-    const success = bindingType === 'writeonly-storage-texture' ? false : true;
+    let success = bindingType === 'writeonly-storage-texture' ? false : true;
```
Kangz, line 792: Maybe have a isWritableUsage helper to make this statement more clear, maybe there's other uses possible in the file as well.

Kangz (approved):

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
Richard-Yunchao: Done. I named the helper as isWritableBinding because 'render-target' is also writable usage but it is not writable binding for texture.

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
```diff
@@ -184,6 +185,10 @@ class TextureUsageTracking extends ValidationTest {
       (pass as GPURenderPassEncoder).draw(3, 1, 0, 0);
     }
   }
+
+  isWritableBinding(bindingType: GPUBindingType | 'render-target'): boolean {
```
Kangz, line 189: @kainino0x any place where we can put such helpers?

Kangz (approved):

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
kainino0x: It should be a new field in capability_info. In this case, a column in kTextureBindingTypeInfo.

kainino0x (commented):
> * replaced bindings within the same dispatch (or no dispatch call at all in a compute pass, which is only for test, but not useful in real world app) should be validated, just as what we did for replaced bindings in render.
> * replaced bindings across dispatch calls (but within the same compute pass) should not be validated, because resource usages validation is done per each dispatch call for compute. So replaced bindings across dispatch calls should be ignored.

I'm confused about this, I don't think this is what we discussed on the spec. Replaced bindings would never count in validation because validation only occurs inside dispatch() which only looks at the current state of things. Unlike render passes where validation is for the entire pass, instead of a single dispatch.

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
Richard-Yunchao: Hmm, we can add a new field in kTextureBindingTypeInfo. But the new field will be set true for only one single binding type and it might be used in only one single test file (this one only?). So adding a new field doesn't make too much sense compared to a helper function here or a simple statement like `bindingType === "writeonly-storage-texture"`. In addition, "render-target" is a resource usage but it is not a binding type, so statement like `kTextureBindingTypeInfo[bindingType].writable` will fail when we run `grunt pre`. For these reasons, maybe we should keep the code as what it is right now? @kainino0x , @Kangz .

Richard-Yunchao:
> > * replaced bindings within the same dispatch (or no dispatch call at all in a compute pass, which is only for test, but not useful in real world app) should be validated, just as what we did for replaced bindings in render.
> > * replaced bindings across dispatch calls (but within the same compute pass) should not be validated, because resource usages validation is done per each dispatch call for compute. So replaced bindings across dispatch calls should be ignored.
> 
> I'm confused about this, I don't think this is what we discussed on the spec. Replaced bindings would never count in validation because validation only occurs inside dispatch() which only looks at the current state of things. Unlike render passes where usage scope validation is for the entire pass, not for a single draw.

Oops. I misunderstood your discussions about replaced bindings in compute. The code has been revised accordingly, PTAL.

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
Kangz: Having the writable set to true for a single binding type is not a problem, especially if we consider that in the future we'll have `storage-texture` that's also writable. However I'm not sure what to do for `render-target`. Maybe we could have a sum type that contains the binding type | render-target? I don't know enough Typescript to be sure.

Richard-Yunchao: Any thoughts about this? @kainino0x . If we don't have a good solution for 'render-target', maybe we should proceed and merge this pr? I think the remain concern is really trivial: one line code like "bindingType === 'writeonly-storage-texture'" or a function like isWritableBinding is not hard to understand, and it doesn't hurt the readability a lot, per my understanding.

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
```diff
@@ -854,12 +876,9 @@ g.test('bindings_in_bundle')
       success = true;
     }
 
-    if (type0 === 'writeonly-storage-texture' && type1 === 'writeonly-storage-texture') {
-      success = true;
-    }
+    if (t.isWritableBinding(type0) && t.isWritableBinding(type1)) success = true;
```
kainino0x, line 879: I thought the ONLY write-write conflict allowed was between two writeonly-storage-textures (not between writeonly-storage-texture and arbitrary other write usages. (Though it probably also would be allowed between readwrite-storage-texture and writeonly-storage-texture in the future.)

We may eventually want to try to find a way to describe the [compatible usage list](https://gpuweb.github.io/gpuweb/#compatible-usage-list) in capability_info, but that's not necessary now (and may never be necessary).

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
kainino0x: If I'm understanding correctly, this function is no longer needed as explained below

Richard-Yunchao: Sounds like I don't need to do any follow-ups on this pr and it can be merged?

kainino0x: "IsWritableBinding" is misleading. The logic should just be inlined like it was before.

It could be replaced with a  function saying "bindingsConflict(a, b)" but we don't want to write that fully general function right now so we shouldn't.

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
Richard-Yunchao: Ack. We have already had tests for combinations for binding types + render-target, it is kind of like a compatible usage list.

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
Richard-Yunchao: Done.

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
```diff
@@ -767,15 +774,26 @@ g.test('replaced_binding')
...
+    if (callDrawOrDispatch) {
+      const pipeline = compute ? t.createNoOpComputePipeline() : t.createNoOpRenderPipeline();
+      t.setPipeline(pass, pipeline, compute);
+      t.issueDrawOrDispatch(pass, compute);
+    }
     pass.setBindGroup(0, bindGroup1);
     pass.endPass();
 
-    const success = bindingType === 'writeonly-storage-texture' ? false : true;
+    let success = bindingType === 'writeonly-storage-texture' ? false : true;
```
kainino0x, line 792:
```suggestion
    // TODO: If the Compatible Usage List (https://gpuweb.github.io/gpuweb/#compatible-usage-list)
    // gets programmatically defined in capability_info, use it here, instead of this logic, for clarity.
    let success = bindingType === 'writeonly-storage-texture' ? false : true;
```

kainino0x (approved):

Richard-Yunchao:
Thanks for adding a comment, @kainino0x

PR #292, opened 2020-09-23
Test BC formats in CopyTextureToTexture.texture_format_equality tests

Jiawei-Shao:


src/webgpu/api/validation/copyTextureToTexture.spec.ts
```diff
@@ -244,24 +283,35 @@ g.test('multisampled_copy_restrictions')
...
-      .combine(poptions('srcFormat', kUncompressedTextureFormats))
-      .combine(poptions('dstFormat', kUncompressedTextureFormats))
+      .combine(poptions('srcFormat', kAllTextureFormats))
+      .combine(poptions('dstFormat', kAllTextureFormats))
   )
   .fn(async t => {
     const { srcFormat, dstFormat } = t.params;
 
+    let deviceForTest: GPUDevice | null;
+    if (t.kBCFormats.includes(srcFormat) || t.kBCFormats.includes(dstFormat)) {
```
Jiawei-Shao, line 296: The members in kUncompressedTextureFormats are in "CompressedTextureFormat" type, while the members in kAllTextureFormats are in "GPUTextureFormat" type, so here I can only use my own-defined kBCFormats to use the function "includes".

kainino0x: I think you can use `kAllTextureFormatInfo[srcFormat].extension === 'texture-compression-bc'`

src/common/framework/gpu/device_pool.ts
```diff
@@ -35,6 +35,31 @@ export class DevicePool {
...
   }
 
+  async acquireWithDeviceDescriptor(
+    devicedescriptor: GPUDeviceDescriptor
+  ): Promise<GPUDevice | null> {
+    assert(!this.failed, 'WebGPU device previously failed to initialize; not retrying');
+
+    if (this.holder === undefined) {
+      try {
+        this.holder = await DevicePool.makeHolderWithDeviceDescriptor(devicedescriptor);
```
kainino0x, line 45: This logic doesn't exactly make sense, because the devicedescriptor will be ignored if there was ever a device created before. For this to be robust, the device descriptor should simply be supplied in the DevicePool constructor.

Ultimately I think we want to key the pool off of the device creation options (i.e. have a map of descriptor -> pool). But I haven't figured out how to implement that exactly. I was thinking about getting to this soon so since you have started writing extension tests I might look at it today.

kainino0x (commented):
overall lgtm but I'm going to try to get to making better extension support, which might simplify this PR.

src/common/framework/gpu/device_pool.ts
Jiawei-Shao: Thanks for pointing out this issue! I've tried to improve it a bit more. PTAL, thanks!

kainino0x (approved):
This looks like it will work. However, I found time to implement this more completely: #294 - sorry for the conflict.
Could you please try making the same kUncompressedTextureFormats -> kAllTextureFormats change on top of that change, to verify that it works for this test and works properly? Then perhaps we can just get it landed and use that mechanism. Any feedback on the mechanism is also welcome.

Code would be something like:
```ts
const extension = kAllTextureFormatInfo[format].extension;
if (extension) {
  await t.asyncReinitDeviceWithDescriptor({ extensions: [extension] });
}
```

Jiawei-Shao:
Hi @kainino0x, I just tried changing  kUncompressedTextureFormats to kAllTextureFormats on top of #294 and it works well. Thanks for fixing this issue! I'll close this PR and add all the tests related to BC formats in another PR after #294 is landed.

PR #293, opened 2020-09-23
Moving createEncoder to ValidationTest

egalli:
Follow-up on https://github.com/gpuweb/cts/pull/286#discussion_r487357285

@kainino0x is this what you were thinking?

### DIFF
--- a/src/webgpu/api/validation/validation_test.ts
+++ b/src/webgpu/api/validation/validation_test.ts
@@ -2,6 +2,14 @@ import { unreachable } from '../../../common/framework/util/util.js';
 import { BindableResource } from '../../capability_info.js';
 import { GPUTest } from '../../gpu_test.js';
 
+type Encoder = GPUCommandEncoder | GPUProgrammablePassEncoder | GPURenderBundleEncoder;
+export const kEncoderTypes = ['non-pass', 'compute pass', 'render pass', 'render bundle'] as const;
+type EncoderType = typeof kEncoderTypes[number];
+
+interface FinishableEncoder {
+  finishEncoder(): GPUCommandBuffer;
+}
+
 export class ValidationTest extends GPUTest {
   getStorageBuffer(): GPUBuffer {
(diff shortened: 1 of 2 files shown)


src/webgpu/api/validation/validation_test.ts
```diff
@@ -148,6 +156,77 @@ export class ValidationTest extends GPUTest {
...
+  createEncoder(encoderType: 'render pass'): GPURenderPassEncoder & FinishableEncoder;
+  createEncoder(encoderType: 'compute pass'): GPUComputePassEncoder & FinishableEncoder;
+  createEncoder(encoderType: 'render bundle'): GPURenderBundleEncoder & FinishableEncoder;
+  createEncoder(
+    encoderType: 'render pass' | 'render bundle'
+  ): (GPURenderPassEncoder | GPURenderBundleEncoder) & FinishableEncoder;
+  createEncoder(
+    encoderType: 'compute pass' | 'render pass' | 'render bundle'
+  ): GPUProgrammablePassEncoder & FinishableEncoder;
+  createEncoder(encoderType: EncoderType): Encoder & FinishableEncoder;
```
kainino0x, line 169: Nice type preservation with these overloads!

kainino0x (approved):
Yup, this is exactly the abstraction I had in mind!

The only difference with what I had in mind was I was thinking of returning something like `class CommandBufferMaker<E extends Encoder> { readonly encoder: E; finish(): GPUCommandBuffer; }`. It would be a little more verbose (`maker.encoder.pushDebugGroup()`) but maybe a little clearer especially to contributors who are less familiar with TypeScript. Either is okay with me though; WDYT?

egalli:
Sounds good to me. I was never a huge fan of modifying native types/casting through unknown. Regarding the extra verbosity, you can always destructure the type. 😄

kainino0x (approved):
> Regarding the extra verbosity, you can always destructure the type. 😄

Very true, good call!

PR #294, opened 2020-09-24
Enable WebGPU tests with device descriptors

kainino0x:
Add GPUTest.asyncReinitDeviceWithDescriptor. Calling this will
reinitialize `this.device` with a device with the provided descriptor
(ensuring that `this.device` has not actually been used already).

### DIFF
--- a/src/common/framework/gpu/device_pool.ts
+++ b/src/common/framework/gpu/device_pool.ts
@@ -1,56 +1,61 @@
-import { assert, raceWithRejectOnTimeout, unreachable, assertReject } from '../util/util.js';
+import { SkipTestCase } from '../fixture.js';
+import {
+  assert,
+  raceWithRejectOnTimeout,
+  unreachable,
+  assertReject,
+  objectEquals,
+} from '../util/util.js';
 
 import { getGPU } from './implementation.js';
 
-interface DeviceHolder {
-  acquired: boolean; // whether the device is currently in use by a test
-  device: GPUDevice;
-  lostReason?: string; // initially undefined; becomes set when the device is lost
+export interface DeviceProvider {
+  acquire(): GPUDevice;
 }
 
 class TestFailedButDeviceReusable extends Error {}
 export class TestOOMedShouldAttemptGC extends Error {}
 
-const kPopErrorScopeTimeoutMS = 5000;
-
 export class DevicePool {
-  failed: boolean = false; // if device init failed once, never try again
-  holder?: DeviceHolder = undefined; // undefined if "uninitialized" (not yet initialized, or lost)
+  /** Device with no descriptor. */
+  private defaultHolder: DeviceHolder | 'uninitialized' | 'failed' = 'uninitialized';
+  /** Devices with descriptors. */
+  private nonDefaultHolders = new DescriptorToDeviceMap();
 
-  async acquire(): Promise<GPUDevice> {
-    assert(!this.failed, 'WebGPU device previously failed to initialize; not retrying');
-
-    if (this.holder === undefined) {
+  /** Request a device from the pool. */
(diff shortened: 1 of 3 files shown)


Jiawei-Shao (approved):
LGTM!

src/common/framework/gpu/device_pool.ts
```diff
@@ -1,56 +1,61 @@
...
 class TestFailedButDeviceReusable extends Error {}
 export class TestOOMedShouldAttemptGC extends Error {}
 
-const kPopErrorScopeTimeoutMS = 5000;
-
 export class DevicePool {
-  failed: boolean = false; // if device init failed once, never try again
-  holder?: DeviceHolder = undefined; // undefined if "uninitialized" (not yet initialized, or lost)
+  /** Device with no descriptor. */
+  private defaultHolder: DeviceHolder | 'uninitialized' | 'failed' = 'uninitialized';
```
Kangz, line 21: why `"failed" = "uninitialized"`?

src/common/framework/gpu/device_pool.ts
```diff
@@ -64,66 +69,183 @@ export class DevicePool {
...
+   * inserts it, and returns it.
+   *
+   * Throws SkipTestCase if devices with this descriptor are unsupported.
+   */
+  async getOrInsert(
+    key: GPUDeviceDescriptor,
+    create: () => Promise<DeviceHolder>
+  ): Promise<DeviceHolder> {
+    // Never retry unsupported configurations.
+    for (const desc of this.unsupported) {
```
Kangz, line 122: Eeek O(n) is a bit sad, but IDK if there's a better way to do it in JS.

Kangz (approved):
Reinit is a bit weird, maybe there could be a `.deviceDescriptor()` along with `.params()` that takes the params and can choose a non-default device descriptor?

src/common/framework/gpu/device_pool.ts
kainino0x: the type is `DeviceHolder | 'uninitialized' | 'failed'`, the initial value is `'uninitialized'`

src/common/framework/gpu/device_pool.ts
kainino0x: It can be done like I did #289 recently. It doesn't have to be perfect anyway, it just is intended to reduce the number of redundant failed device requests. (It already isn't perfect because it doesn't know the difference between default values and unspecified values.)

kainino0x:
> Reinit is a bit weird, maybe there could be a `.deviceDescriptor()` along with `.params()` that takes the params and can choose a non-default device descriptor?

I thought about it but I think it adds unnecessary added complexity to writing tests. Some logic (e.g. looking up info in the texture tables) would end up being duplicated between the two. Also right now TestGroupBuilders don't know anything specific to WebGPU so more likely it would be some special private param value like `_deviceDescriptor`, which is even less nice.

kainino0x:
Landed to unblock Jiawei; opened #295

src/common/framework/gpu/device_pool.ts
Kangz: Gotcha

Kangz:
> > Reinit is a bit weird, maybe there could be a `.deviceDescriptor()` along with `.params()` that takes the params and can choose a non-default device descriptor?
> 
> I thought about it but I think it adds unnecessary added complexity to writing tests. Some logic (e.g. looking up info in the texture tables) would end up being duplicated between the two. Also right now TestGroupBuilders don't know anything specific to WebGPU so more likely it would be some special private param value like `_deviceDescriptor`, which is even less nice.

IDK, maybe GPUTest could have an overload of params() that returns something that knows how to `.deviceDescriptor`. I agree that it makes the discovery of necessary extensions happen in two places, but right now it looks like it uses double the devices. I'm not sure what's the best solution.

PR #295, opened 2020-09-25
Optimize device pool lookup

kainino0x:
Replace O(n^2) with stringification+Map.

Follow-up to #294


src/common/framework/gpu/device_pool.ts
```diff
@@ -140,30 +130,54 @@ class DescriptorToDeviceMap {
...
+    maxDynamicStorageBuffersPerPipelineLayout: 4,
+    maxSampledTexturesPerShaderStage: 16,
+    maxSamplersPerShaderStage: 16,
+    maxStorageBuffersPerShaderStage: 4,
+    maxStorageTexturesPerShaderStage: 4,
+    maxUniformBuffersPerShaderStage: 12,
+    maxUniformBufferBindingSize: 16384,
+    ...desc.limits,
+  };
+  return JSON.stringify({ extensions, limits });
```
kainino0x, line 178: todo: this could miss members in the device descriptor, if members are added. Would be hard to debug because it would make the cache misbehave. Should  probably do something like `{ extensions: undefined, limits: undefined, ...desc, ...{ extensions, limits } }`

src/common/framework/gpu/device_pool.ts
```diff
@@ -114,56 +102,85 @@ class DescriptorToDeviceMap {
...
+    const [descriptor, key] = canonicalizeDescriptor(uncanonicalizedDescriptor);
     // Never retry unsupported configurations.
-    for (const desc of this.unsupported) {
-      if (objectEquals(key, desc)) {
-        throw new SkipTestCase(`GPUDeviceDescriptor previously failed: ${JSON.stringify(key)}`);
-      }
+    if (this.unsupported.has(key)) {
+      throw new SkipTestCase(
+        `GPUDeviceDescriptor previously failed: ${JSON.stringify(descriptor)}`
+      );
```
austinEng, line 111: what about tests that intentionally fail creation? we don't want those tests to get skipped

src/common/framework/gpu/device_pool.ts
```diff
@@ -114,56 +102,85 @@ class DescriptorToDeviceMap {
...
 
     // Search for an existing device with the same descriptor.
-    for (const item of this.items) {
-      if (objectEquals(key, item.key)) {
-        // Move the item to the end of the set (most recently used).
-        this.items.delete(item);
-        this.items.add(item);
-        return item.value;
+    {
+      const value = this.items.get(key);
```
austinEng, line 116: nit: `items` is a very general name

austinEng (commented):
high level q: why are we optimizing device lookup - is it because device discovery in Chrome is particularly slow right now? (i.e. much slower than the round trip).

overall, think it makes sense to avoid the round trip if we can. just wondering if there's another motivating reason for it.

src/common/framework/gpu/device_pool.ts
kainino0x: Tests that intentionally fail creation should call createDevice directly instead of going through this mechanism (which only exists to reuse existing valid devices)

kainino0x:
It's for this comment thread:
https://github.com/gpuweb/cts/pull/294#discussion_r494206905
but it's also not really very important.

However this also improves things to canonicalize the adapter descriptor so that devices get reused more often.

src/common/framework/gpu/device_pool.ts
kainino0x: This mechanism is exposed as `t.selectDeviceOrSkipTestCase()` so the behavior (and the fact that it's not suited for such tests) should be clear.

austinEng: Ah okay, sounds good.

src/common/framework/gpu/device_pool.ts
```diff
@@ -114,56 +102,85 @@ class DescriptorToDeviceMap {
...
+  const limitsCanonicalized: CanonicalLimits = {
+    maxBindGroups: lim.maxBindGroups ?? 4,
+    maxDynamicUniformBuffersPerPipelineLayout: lim.maxDynamicUniformBuffersPerPipelineLayout ?? 8,
+    maxDynamicStorageBuffersPerPipelineLayout: lim.maxDynamicStorageBuffersPerPipelineLayout ?? 4,
+    maxSampledTexturesPerShaderStage: lim.maxSampledTexturesPerShaderStage ?? 16,
+    maxSamplersPerShaderStage: lim.maxSamplersPerShaderStage ?? 16,
+    maxStorageBuffersPerShaderStage: lim.maxStorageBuffersPerShaderStage ?? 4,
+    maxStorageTexturesPerShaderStage: lim.maxStorageTexturesPerShaderStage ?? 4,
+    maxUniformBuffersPerShaderStage: lim.maxUniformBuffersPerShaderStage ?? 12,
+    maxUniformBufferBindingSize: lim.maxUniformBufferBindingSize ?? 16384,
```
austinEng, line 174: maybe these base limits can come from `constants.ts`

austinEng (approved):

PR #296, opened 2020-09-27
Add tests about compressed texture formats in CopyT2T validation tests

Jiawei-Shao:


src/webgpu/api/validation/copyTextureToTexture.spec.ts
```diff
@@ -244,14 +259,34 @@ g.test('multisampled_copy_restrictions')
...
+      .combine(poptions('srcFormat', kAllTextureFormats))
+      .combine(poptions('dstFormat', kAllTextureFormats))
   )
   .fn(async t => {
     const { srcFormat, dstFormat } = t.params;
+    const extensions: Array<GPUExtensionName> = [];
+
+    const srcFormatExtension = kAllTextureFormatInfo[srcFormat].extension;
+    if (srcFormatExtension !== undefined) {
+      if (!(await t.IsExtensionSupported(srcFormatExtension))) {
```
Kangz, line 274: It is unfortunate that an await is required here, maybe we could just set it on the descriptor, and see if asyncReinitDeviceWithDescriptor succeeds.

Kangz (approved):

src/webgpu/api/validation/copyTextureToTexture.spec.ts
Jiawei-Shao: Hi @kainino0x , sorry that I am not sure if this usage is correct or not. Could you take a look?

kainino0x (approved):
LGTM except for the extension setup (sorry I didn't explain/document this properly)!

PR #297, opened 2020-09-28
Add .vscode/ to gitignore

Jiawei-Shao:


Kangz (approved):

kainino0x:
LGTM. I've been thinking about taking out some of the more opinionated options from the .vscode file that is in here, too. And cleaning up the tasks.json. Do you think that would be a good idea?

PR #298, opened 2020-09-29
Rename asyncReinitDeviceWithDescriptor -> selectDeviceOrSkipTestCase

kainino0x:


kainino0x:
> > > Reinit is a bit weird, maybe there could be a `.deviceDescriptor()` along with `.params()` that takes the params and can choose a non-default device descriptor?
> > 
> > 
> > I thought about it but I think it adds unnecessary added complexity to writing tests. Some logic (e.g. looking up info in the texture tables) would end up being duplicated between the two. Also right now TestGroupBuilders don't know anything specific to WebGPU so more likely it would be some special private param value like `_deviceDescriptor`, which is even less nice.
> 
> IDK, maybe GPUTest could have an overload of params() that returns something that knows how to `.deviceDescriptor`. I agree that it makes the discovery of necessary extensions happen in two places, but right now it looks like it uses double the devices. I'm not sure what's the best solution.

I think the model is the right one; the runtime checks make sure you do the right thing. This is a rename (plus documentation) to make it sound like the lightweight operation that it is.

Jiawei-Shao (approved):
LGTM!

Kangz (approved):

PR #300, opened 2020-09-30
Fix a bug for texture usage validation in compute

Richard-Yunchao:
Resource usage validation scope in compute is defined by dispatch calls.
If dispatch command is not called in a compute pass, we don't need to
do resource usage validation for that pass. As a result, it should success
in this situation.

This change also sets compute pipeline and calls dispatch command for
compute pass if the test is not testing validation scope.

### DIFF
--- a/src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
+++ b/src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
@@ -1017,7 +1017,7 @@ g.test('validation_scope,no_draw_or_dispatch')
     // we don't need to do resource usage validation and no validation error to be reported.
     t.expectValidationError(() => {
       encoder.finish();
-    }, compute);
+    }, !compute);
   });
 
 g.test('validation_scope,same_draw_or_dispatch')
--- a/src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
+++ b/src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
@@ -185,6 +185,12 @@ class TextureUsageTracking extends ValidationTest {
       (pass as GPURenderPassEncoder).draw(3, 1, 0, 0);
     }
   }
+
+  setComputePipelineAndCallDispatch(pass: GPUComputePassEncoder) {
+    const pipeline = this.createNoOpComputePipeline();
+    pass.setPipeline(pipeline);
+    pass.dispatch(1);
+  }
 }
 
 export const g = makeTestGroup(TextureUsageTracking);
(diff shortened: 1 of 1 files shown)


Richard-Yunchao:
@kainino0x , we discussed this at https://github.com/gpuweb/gpuweb/pull/1074

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
```diff
@@ -1000,9 +1013,11 @@ g.test('validation_scope,no_draw_or_dispatch')
     pass.setBindGroup(1, bindGroup1);
     pass.endPass();
 
+    // Resource usage validation scope is defined by dispatch calls. If dispatch is not called,
+    // we don't need to do resource usage validation and no validation error to be reported.
     t.expectValidationError(() => {
       encoder.finish();
-    });
+    }, compute);
```
kainino0x, line 1020: Isn't this inverted?
```suggestion
    }, !compute);
```

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
```diff
@@ -984,9 +993,13 @@ g.test('unused_bindings_in_pipeline')
     if (callDrawOrDispatch) t.issueDrawOrDispatch(pass, compute);
     pass.endPass();
 
+    // Resource usage validation scope is defined by dispatch calls. If dispatch is not called,
+    // we don't need to do resource usage validation and no validation error to be reported.
+    const success = compute && !callDrawOrDispatch;
```
kainino0x, line 998: Shouldn't this be:
```suggestion
    const success = compute || !callDrawOrDispatch;
```

kainino0x (commented):
Thanks for catching these, and sorry I didn't catch it in review!

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
Richard-Yunchao: I think this should be a "&&", because the validation scope in render pass is the pass, regardless we call draw command. However, the validation scope for compute is each dispatch call. If we don't call dispatch, no validation scope at all (so it will success).

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
Richard-Yunchao: True. Success equals to compute.

Kangz (approved):

src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
kainino0x: Oh yes, sorry. I have a hard time with boolean expressions often.

kainino0x (approved):

PR #302, opened 2020-10-02
Memory sync test for buffer

Richard-Yunchao:
This change add memory synchronization tests for buffer in compute.
Each test read the data from a buffer, add 1, and write back to the
buffer and iterate twice in different validation scopes: different
dispatch calls, different passes, or different command buffers. Then
verify the data in buffer.

### DIFF
--- a/src/webgpu/api/operation/memory_sync/buffer.spec.ts
+++ b/src/webgpu/api/operation/memory_sync/buffer.spec.ts
@@ -193,13 +193,13 @@ class BufferSyncTest extends GPUTest {
   }
 
   // Write buffer via mapping and writing
-  writeByMapWrite(buffer: GPUBuffer, value: number) {
+  async writeByMapWrite(buffer: GPUBuffer, value: number) {
     const data = new Uint32Array(SIZE / 4);
     for (let i = 0; i < SIZE / 4; ++i) {
       data[i] = value;
     }
 
-    buffer.mapAsync(GPUMapMode.WRITE);
+    await buffer.mapAsync(GPUMapMode.WRITE);
     new Uint8Array(buffer.getMappedRange()).set(data);
     buffer.unmap();
   }
@@ -243,7 +243,7 @@ class BufferSyncTest extends GPUTest {
     return encoder.finish();
   }
 
-  createQueueSubmitsAndIssueWriteOp(
+  async createQueueSubmitsAndIssueWriteOp(
     writeOp: string,
     bundle: boolean,
     buffer: GPUBuffer,
--- a/src/webgpu/api/operation/memory_sync/buffer.spec.ts
+++ b/src/webgpu/api/operation/memory_sync/buffer.spec.ts
@@ -1,0 +1,257 @@
+export const description = `
+Memory Synchronization Tests for Buffer.
+
+- Test write-after-write on a single buffer. Create one single buffer and initiate it to 0.
+Write a number (say 1) into the buffer via render pass, compute pass, or copy. Write another
+number (say 2) into the same buffer via render pass, compute pass, or copy.
+  - x= 1st write type: {storage buffer in {render, compute}, T2B, B2B}
+  - x= 2nd write type: {storage buffer in {render, compute}, T2B, B2B}
+  - for each write, if render, x= {bundle, non-bundle}
+  - if pass type is the same, x= {single pass, separate passes} (note: render has loose guarantees)
+  - if not single pass, x= writes in {same cmdbuf, separate cmdbufs, separate queues}
+`;
+
+import { pbool, poptions, params } from '../../../../common/framework/params_builder.js';
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { assert } from '../../../../common/framework/util/util.js';
+import { GPUTest } from '../../../gpu_test.js';
+
+const SIZE = 4;
+class BufferSyncTest extends GPUTest {
+  // Create a buffer, and initiate it to zero.
+  createBuffer(): GPUBuffer {
+    const data = new Uint32Array(SIZE / 4);
+    for (let i = 0; i < SIZE / 4; ++i) {
+      data[i] = 0;
+    }
+    const buffer = this.device.createBuffer({
+      mappedAtCreation: true,
+      size: SIZE,
+      usage: GPUBufferUsage.COPY_SRC | GPUBufferUsage.COPY_DST | GPUBufferUsage.STORAGE,
+    });
+    new Uint8Array(buffer.getMappedRange()).set(data);
+    buffer.unmap();
+    return buffer;
+  }
+
+  createBindGroup(
+    pipeline: GPURenderPipeline | GPUComputePipeline,
+    buffer: GPUBuffer
+  ): GPUBindGroup {
+    return this.device.createBindGroup({
+      layout: pipeline.getBindGroupLayout(0),
+      entries: [{ binding: 0, resource: { buffer } }],
+    });
+  }
+
+  // Create a compute pipeline and write data into storage buffer.
+  // If the pipeline is used in the first write, then write 1 into the buffer.
+  // Otherwise, write 2 into the buffer.
+  createComputePipeline(isFirstWrite: boolean): GPUComputePipeline {
+    const value: string = isFirstWrite ? '1' : '2';
+    const wgslCompute = `
+      type Data = [[block]] struct {
+        [[offset 0]] a : i32;
+      };
+
+      [[binding 0, set 0]] var<storage_buffer> data : Data;
+      fn main() -> void {
+	data.a = ${value};
+        return;
+      }
+
+      entry_point compute = main;
+    `;
+
+    return this.device.createComputePipeline({
+      computeStage: {
+        module: this.device.createShaderModule({
+          code: wgslCompute,
+        }),
+        entryPoint: 'main',
+      },
+    });
+  }
+
+  // Create a render pipeline and write data into storage buffer at fragment stage.
+  // If the pipeline is used in the first write, then write 1 into the buffer.
+  // Otherwise, write 2 into the buffer.
+  createRenderPipeline(isFirstWrite: boolean): GPURenderPipeline {
+    const wgslVertex = `
+    [[builtin position]] var<out> Position : vec4<f32>;
+      fn vert_main() -> void {
+        Position = vec4<f32>(0.5, 0.5, 0.0, 1.0);
+        return;
+      }
+
+      entry_point vertex = vert_main;
+    `;
+
+    const value: string = isFirstWrite ? '1' : '2';
+    const wgslFragment = `
+    [[location 0]] var<out> outColor : vec4<f32>;
+      type Data = [[block]] struct {
+        [[offset 0]] a : i32;
+      };
+
+      [[binding 0, set 0]] var<storage_buffer> data : Data;
+      fn frag_main() -> void {
+	data.a = ${value};
+	outColor = vec4<f32>(1.0, 0.0, 0.0, 1.0);
+        return;
+      }
+
+      entry_point fragment = frag_main;
+    `;
+
+    return this.device.createRenderPipeline({
+      vertexStage: {
+        module: this.device.createShaderModule({
+          code: wgslVertex,
+        }),
+        entryPoint: 'vert_main',
+      },
+      fragmentStage: {
+        module: this.device.createShaderModule({
+          code: wgslFragment,
+        }),
+        entryPoint: 'frag_main',
+      },
+      primitiveTopology: 'point-list',
+      colorStates: [{ format: 'rgba8unorm' }],
+    });
+  }
+
+  beginSimpleRenderPass(encoder: GPUCommandEncoder): GPURenderPassEncoder {
+    const view = this.device
+      .createTexture({
+        size: { width: 4, height: 4, depth: 1 },
+        format: 'rgba8unorm',
+        usage: GPUTextureUsage.OUTPUT_ATTACHMENT,
+      })
+      .createView();
+    return encoder.beginRenderPass({
+      colorAttachments: [
+        {
+          attachment: view,
+          loadValue: { r: 0.0, g: 1.0, b: 0.0, a: 1.0 },
+          storeOp: 'store',
+        },
+      ],
+    });
+  }
+
+  // Create render pass or compute pass, set pipeline and bind groups and write data into storage
+  // buffer. If the pass is render pass, it may use bundle. Note that we write 1 into the buffer
+  // if the pass is used in the first write, and we write 2 into the buffer otherwise.
+  createPass(
+    buffer: GPUBuffer,
+    bundle: boolean,
+    isRenderPass: boolean,
+    encoder: GPUCommandEncoder,
+    isFirstWrite: boolean
+  ) {
+    assert(!bundle || isRenderPass);
+    if (isRenderPass) {
+      const pipeline = this.createRenderPipeline(isFirstWrite);
+      const bindGroup = this.createBindGroup(pipeline, buffer);
+      const pass = this.beginSimpleRenderPass(encoder);
+      if (bundle) {
+        const bundleEncoder = this.device.createRenderBundleEncoder({
+          colorFormats: ['rgba8unorm'],
+        });
+        bundleEncoder.setBindGroup(0, bindGroup);
+        bundleEncoder.setPipeline(pipeline);
+        bundleEncoder.draw(1, 1, 0, 0);
+        const bundle = bundleEncoder.finish();
+        pass.executeBundles([bundle]);
+      } else {
+        pass.setPipeline(pipeline);
+        pass.setBindGroup(0, bindGroup);
+        pass.draw(1, 1, 0, 0);
+      }
+      pass.endPass();
+    } else {
+      const pipeline = this.createComputePipeline(isFirstWrite);
+      const bindGroup = this.createBindGroup(pipeline, buffer);
+      const pass = encoder.beginComputePass();
+      pass.setPipeline(pipeline);
+      pass.setBindGroup(0, bindGroup);
+      pass.dispatch(1);
+      pass.endPass();
+    }
+  }
+
+  createCommandBuffer(
+    buffer: GPUBuffer,
+    bundle: boolean,
+    isRenderPass: boolean,
+    isFirstWrite: boolean
+  ): GPUCommandBuffer {
+    const encoder = this.device.createCommandEncoder();
+    this.createPass(buffer, bundle, isRenderPass, encoder, isFirstWrite);
+    return encoder.finish();
+  }
+
+  verifyData(buffer: GPUBuffer, expectedValue: number) {
+    const bufferData = new Uint32Array(1);
+    bufferData[0] = expectedValue;
+    this.expectContents(buffer, bufferData);
+  }
+}
+
+export const g = makeTestGroup(BufferSyncTest);
+
+// Test write-after-write between passes. The first write will write 1 into a storage buffer.
+// The second write will write 2 into the same storage buffer. So, expected data in buffer is 2.
+// The two writes can be in render pass or compute pass, and it can be in the same command buffer,
+// or separate command buffers, or separate queues. If the pass is render pass, it may use bundle.
+g.test('write_after_write,passes')
+  .params(
+    params()
+      // TODO (yunchao.he@intel.com): add multi-queue.
+      // Then the pbool 'sameCommandBuffer' can be a poption which consits of 3 options:
+      // same command buffer, separate command buffers, separate queues.
+      .combine(pbool('sameCommandBuffer'))
+      .combine(pbool('firstWriteInBundle'))
+      .combine(pbool('secondWriteInBundle'))
+      .combine(poptions('firstWrite', ['render', 'compute'] as const))
+      .combine(poptions('secondWrite', ['render', 'compute'] as const))
+      .unless(
+        p =>
+          (p.firstWriteInBundle && p.firstWrite !== 'render') ||
+          (p.secondWriteInBundle && p.secondWrite !== 'render')
+      )
+  )
+  .fn(async t => {
+    const {
+      sameCommandBuffer,
+      firstWriteInBundle,
+      secondWriteInBundle,
+      firstWrite,
+      secondWrite,
+    } = t.params;
+    const buffer = t.createBuffer();
+    if (sameCommandBuffer) {
+      const encoder = t.device.createCommandEncoder();
+      t.createPass(buffer, firstWriteInBundle, firstWrite === 'render', encoder, true);
+      t.createPass(buffer, secondWriteInBundle, secondWrite === 'render', encoder, false);
+      t.device.defaultQueue.submit([encoder.finish()]);
+    } else {
+      const command_buffers: GPUCommandBuffer[] = [];
+      command_buffers.push(
+        t.createCommandBuffer(buffer, firstWriteInBundle, firstWrite === 'render', true)
+      );
+      command_buffers.push(
+        t.createCommandBuffer(buffer, secondWriteInBundle, secondWrite === 'render', false)
+      );
+      t.device.defaultQueue.submit(command_buffers);
+    }
+    t.verifyData(buffer, 2);
+  });
+
+// TODO (yunchao.he@intel.com):
+// g.test('write_after_write,draw_or_dispatch_in_same_pass')
+// g.test('write_after_write,copy_to_pass')
+// g.test('write_after_write,pass_to_copy')
+// g.test('write_after_write,copy_to_copy')
--- a/src/webgpu/api/operation/memory_sync/buffer.spec.ts
+++ b/src/webgpu/api/operation/memory_sync/buffer.spec.ts
@@ -19,7 +19,9 @@ import { GPUTest } from '../../../gpu_test.js';
 const SIZE = 4;
 class BufferSyncTest extends GPUTest {
   // Create a buffer, and initiate it to a specified value for all elements.
-  createBufferWithValue(initValue: number): GPUBuffer {
+  async createBufferWithValue(initValue: number): Promise<GPUBuffer> {
+    const fence = this.queue.createFence();
+    this.queue.signal(fence, 1);
     const data = new Uint32Array(SIZE / 4);
     for (let i = 0; i < SIZE / 4; ++i) {
       data[i] = initValue;
@@ -31,6 +33,7 @@ class BufferSyncTest extends GPUTest {
     });
     new Uint8Array(buffer.getMappedRange()).set(data);
     buffer.unmap();
+    await fence.onCompletion(1);
     return buffer;
   }
 
@@ -157,14 +160,14 @@ class BufferSyncTest extends GPUTest {
   }
 
   // Write bufer via BuferToBuffer copy.
-  writeByB2BCopy(buffer: GPUBuffer, value: number, encoder: GPUCommandEncoder) {
-    const tmpBuffer = this.createBufferWithValue(value);
+  async writeByB2BCopy(buffer: GPUBuffer, value: number, encoder: GPUCommandEncoder) {
+    const tmpBuffer = await this.createBufferWithValue(value);
     encoder.copyBufferToBuffer(tmpBuffer, 0, buffer, 0, SIZE);
   }
 
   // Write buffer via TextureToBuffer copy.
-  writeByT2BCopy(buffer: GPUBuffer, value: number, encoder: GPUCommandEncoder) {
-    const tmpBuffer = this.createBufferWithValue(value);
+  async writeByT2BCopy(buffer: GPUBuffer, value: number, encoder: GPUCommandEncoder) {
+    const tmpBuffer = await this.createBufferWithValue(value);
     const tmpTexture = this.device.createTexture({
       size: { width: 1, height: 1, depth: 1 },
       format: 'rgba8uint',
@@ -193,7 +196,7 @@ class BufferSyncTest extends GPUTest {
   }
 
   // Issue write operation via render pass, compute pass, copy, etc.
-  issueWriteOp(
+  async issueWriteOp(
     writeOp: string,
     bundle: boolean,
     buffer: GPUBuffer,
@@ -209,10 +212,10 @@ class BufferSyncTest extends GPUTest {
         this.writeByComputePass(buffer, value, encoder);
         break;
       case 'b2bCopy':
-        this.writeByB2BCopy(buffer, value, encoder);
+        await this.writeByB2BCopy(buffer, value, encoder);
         break;
       case 't2bCopy':
-        this.writeByT2BCopy(buffer, value, encoder);
+        await this.writeByT2BCopy(buffer, value, encoder);
         break;
       default:
         assert(true);
@@ -220,18 +223,18 @@ class BufferSyncTest extends GPUTest {
     }
   }
 
-  createCommandBufferAndIssueWriteOp(
+  async createCommandBufferAndIssueWriteOp(
     writeOp: string,
     bundle: boolean,
     buffer: GPUBuffer,
     value: number
-  ): GPUCommandBuffer {
+  ): Promise<GPUCommandBuffer> {
     const encoder = this.device.createCommandEncoder();
-    this.issueWriteOp(writeOp, bundle, buffer, value, encoder);
+    await this.issueWriteOp(writeOp, bundle, buffer, value, encoder);
     return encoder.finish();
   }
 
-  createQueueSubmitsAndIssueWriteOp(
+  async createQueueSubmitsAndIssueWriteOp(
     writeOp: string,
     bundle: boolean,
     buffer: GPUBuffer,
@@ -241,7 +244,7 @@ class BufferSyncTest extends GPUTest {
       this.writeByWriteBuffer(buffer, value);
     } else {
       const encoder = this.device.createCommandEncoder();
-      this.issueWriteOp(writeOp, bundle, buffer, value, encoder);
+      await this.issueWriteOp(writeOp, bundle, buffer, value, encoder);
       this.device.defaultQueue.submit([encoder.finish()]);
     }
   }
@@ -306,7 +309,7 @@ g.test('write_after_write')
       secondWriteInBundle,
     } = t.params;
 
-    const buffer = t.createBufferWithValue(0);
+    const buffer = await t.createBufferWithValue(0);
 
     const writeInBundle = [firstWriteInBundle, secondWriteInBundle];
     const writeOp = [firstWriteOp, secondWriteOp];
@@ -314,7 +317,7 @@ g.test('write_after_write')
       case 'sameCmdbuf': {
         const encoder = t.device.createCommandEncoder();
         for (let i = 0; i < 2; i++) {
-          t.issueWriteOp(writeOp[i], writeInBundle[i], buffer, i + 1, encoder);
+          await t.issueWriteOp(writeOp[i], writeInBundle[i], buffer, i + 1, encoder);
         }
         t.device.defaultQueue.submit([encoder.finish()]);
         break;
@@ -323,7 +326,7 @@ g.test('write_after_write')
         const command_buffers: GPUCommandBuffer[] = [];
         for (let i = 0; i < 2; i++) {
           command_buffers.push(
-            t.createCommandBufferAndIssueWriteOp(writeOp[i], writeInBundle[i], buffer, i + 1)
+            await t.createCommandBufferAndIssueWriteOp(writeOp[i], writeInBundle[i], buffer, i + 1)
           );
         }
         t.device.defaultQueue.submit(command_buffers);
@@ -331,7 +334,7 @@ g.test('write_after_write')
       }
       case 'separateSubmits': {
         for (let i = 0; i < 2; i++) {
-          t.createQueueSubmitsAndIssueWriteOp(writeOp[i], writeInBundle[i], buffer, i + 1);
+          await t.createQueueSubmitsAndIssueWriteOp(writeOp[i], writeInBundle[i], buffer, i + 1);
         }
         break;
       }
--- a/src/webgpu/api/operation/memory_sync/buffer.spec.ts
+++ b/src/webgpu/api/operation/memory_sync/buffer.spec.ts
@@ -4,11 +4,11 @@ Memory Synchronization Tests for Buffer.
 - Test write-after-write on a single buffer. Create one single buffer and initiate it to 0.
 Write a number (say 1) into the buffer via render pass, compute pass, or copy. Write another
 number (say 2) into the same buffer via render pass, compute pass, or copy.
-  - x= 1st write type: {storage buffer in {render, compute}, T2B, B2B}
-  - x= 2nd write type: {storage buffer in {render, compute}, T2B, B2B}
+  - x= 1st write type: {storage buffer in {render, compute}, T2B, B2B, WriteBuffer, MapWrite}
+  - x= 2nd write type: {storage buffer in {render, compute}, T2B, B2B, WriteBuffer, MapWrite}
   - for each write, if render, x= {bundle, non-bundle}
   - if pass type is the same, x= {single pass, separate passes} (note: render has loose guarantees)
-  - if not single pass, x= writes in {same cmdbuf, separate cmdbufs, separate queues}
+  - if not single pass, x= writes in {same cmdbuf, separate cmdbufs, separate submits, separate queues}
 `;
 
 import { pbool, poptions, params } from '../../../../common/framework/params_builder.js';
@@ -18,11 +18,11 @@ import { GPUTest } from '../../../gpu_test.js';
 
 const SIZE = 4;
 class BufferSyncTest extends GPUTest {
-  // Create a buffer, and initiate it to zero.
-  createBuffer(): GPUBuffer {
+  // Create a buffer, and initiate it to a specified value for all elements.
+  createBufferWithValue(initValue: number): GPUBuffer {
     const data = new Uint32Array(SIZE / 4);
     for (let i = 0; i < SIZE / 4; ++i) {
-      data[i] = 0;
+      data[i] = initValue;
     }
     const buffer = this.device.createBuffer({
       mappedAtCreation: true,
@@ -48,16 +48,14 @@ class BufferSyncTest extends GPUTest {
   createStorageWriteComputePipeline(value: number): GPUComputePipeline {
     const wgslCompute = `
       type Data = [[block]] struct {
-        [[offset 0]] a : i32;
+        [[offset(0)]] a : i32;
       };
 
-      [[binding 0, set 0]] var<storage_buffer> data : Data;
-      fn main() -> void {
-	data.a = ${value};
+      [[binding(0), set(0)]] var<storage_buffer> data : Data;
+      [[stage(compute)]] fn main() -> void {
+        data.a = ${value};
         return;
       }
-
-      entry_point compute = main;
     `;
 
     return this.device.createComputePipeline({
@@ -73,29 +71,25 @@ class BufferSyncTest extends GPUTest {
   // Create a render pipeline and write given data into storage buffer at fragment stage.
   createStorageWriteRenderPipeline(value: number): GPURenderPipeline {
     const wgslVertex = `
-    [[builtin position]] var<out> Position : vec4<f32>;
-      fn vert_main() -> void {
+      [[builtin(position)]] var<out> Position : vec4<f32>;
+      [[stage(vertex)]] fn vert_main() -> void {
         Position = vec4<f32>(0.5, 0.5, 0.0, 1.0);
         return;
       }
-
-      entry_point vertex = vert_main;
     `;
 
     const wgslFragment = `
-    [[location 0]] var<out> outColor : vec4<f32>;
+      [[location(0)]] var<out> outColor : vec4<f32>;
       type Data = [[block]] struct {
-        [[offset 0]] a : i32;
+        [[offset(0)]] a : i32;
       };
 
-      [[binding 0, set 0]] var<storage_buffer> data : Data;
-      fn frag_main() -> void {
-	data.a = ${value};
-	outColor = vec4<f32>(1.0, 0.0, 0.0, 1.0);
+      [[binding(0), set(0)]] var<storage_buffer> data : Data;
+      [[stage(fragment)]] fn frag_main() -> void {
+        data.a = ${value};
+        outColor = vec4<f32>(1.0, 0.0, 0.0, 1.0);
         return;
       }
-
-      entry_point fragment = frag_main;
     `;
 
     return this.device.createRenderPipeline({
@@ -134,29 +128,24 @@ class BufferSyncTest extends GPUTest {
     });
   }
 
-  // Write storage buffer via draw call in render pass. Use bundle if needed.
+  // Write buffer via draw call in render pass. Use bundle if needed.
   writeByRenderPass(bundle: boolean, buffer: GPUBuffer, value: number, encoder: GPUCommandEncoder) {
     const pipeline = this.createStorageWriteRenderPipeline(value);
     const bindGroup = this.createBindGroup(pipeline, buffer);
+
     const pass = this.beginSimpleRenderPass(encoder);
-    if (bundle) {
-      const bundleEncoder = this.device.createRenderBundleEncoder({
-        colorFormats: ['rgba8unorm'],
-      });
-      bundleEncoder.setBindGroup(0, bindGroup);
-      bundleEncoder.setPipeline(pipeline);
-      bundleEncoder.draw(1, 1, 0, 0);
-      const bundles = bundleEncoder.finish();
-      pass.executeBundles([bundles]);
-    } else {
-      pass.setPipeline(pipeline);
-      pass.setBindGroup(0, bindGroup);
-      pass.draw(1, 1, 0, 0);
-    }
+    const renderer = bundle
+      ? this.device.createRenderBundleEncoder({ colorFormats: ['rgba8unorm'] })
+      : pass;
+    renderer.setBindGroup(0, bindGroup);
+    renderer.setPipeline(pipeline);
+    renderer.draw(1, 1, 0, 0);
+
+    if (bundle) pass.executeBundles([(renderer as GPURenderBundleEncoder).finish()]);
     pass.endPass();
   }
 
-  // Write storage buffer via dispatch call in compute pass.
+  // Write buffer via dispatch call in compute pass.
   writeByComputePass(buffer: GPUBuffer, value: number, encoder: GPUCommandEncoder) {
     const pipeline = this.createStorageWriteComputePipeline(value);
     const bindGroup = this.createBindGroup(pipeline, buffer);
@@ -167,6 +156,54 @@ class BufferSyncTest extends GPUTest {
     pass.endPass();
   }
 
+  // Write bufer via BuferToBuffer copy.
+  writeByB2BCopy(buffer: GPUBuffer, value: number, encoder: GPUCommandEncoder) {
+    const tmpBuffer = this.createBufferWithValue(value);
+    encoder.copyBufferToBuffer(tmpBuffer, 0, buffer, 0, SIZE);
+  }
+
+  // Write buffer via TextureToBuffer copy.
+  writeByT2BCopy(buffer: GPUBuffer, value: number, encoder: GPUCommandEncoder) {
+    const tmpBuffer = this.createBufferWithValue(value);
+    const tmpTexture = this.device.createTexture({
+      size: { width: 1, height: 1, depth: 1 },
+      format: 'rgba8uint',
+      usage: GPUTextureUsage.COPY_SRC | GPUTextureUsage.COPY_DST,
+    });
+    encoder.copyBufferToTexture(
+      { buffer: tmpBuffer, bytesPerRow: 256 },
+      { texture: tmpTexture, mipLevel: 0, origin: { x: 0, y: 0, z: 0 } },
+      { width: 1, height: 1, depth: 1 }
+    );
+    encoder.copyTextureToBuffer(
+      { texture: tmpTexture, mipLevel: 0, origin: { x: 0, y: 0, z: 0 } },
+      { buffer, bytesPerRow: 256 },
+      { width: 1, height: 1, depth: 1 }
+    );
+  }
+
+  // Write buffer via writeBuffer API on queue
+  writeByWriteBuffer(buffer: GPUBuffer, value: number) {
+    const data = new Uint32Array(SIZE / 4);
+    for (let i = 0; i < SIZE / 4; ++i) {
+      data[i] = value;
+    }
+
+    this.device.defaultQueue.writeBuffer(buffer, 0, data);
+  }
+
+  // Write buffer via mapping and writing
+  writeByMapWrite(buffer: GPUBuffer, value: number) {
+    const data = new Uint32Array(SIZE / 4);
+    for (let i = 0; i < SIZE / 4; ++i) {
+      data[i] = value;
+    }
+
+    buffer.mapAsync(GPUMapMode.WRITE);
+    new Uint8Array(buffer.getMappedRange()).set(data);
+    buffer.unmap();
+  }
+
   // Issue write operation via render pass, compute pass, copy, etc.
   issueWriteOp(
     writeOp: string,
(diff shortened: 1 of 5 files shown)


Richard-Yunchao:
The first pr for memory sync tests for buffer. I verified the tests in Chromium because currently we have little wgsl code in CTS. It works, which indicate that we can support storage buffer in wgsl(tint) in Chromium. Please take a look at the latest test plan too (We may need to add more binding types, and/or read-after-write, write-after-write, read-before-write combinations). Comments are welcome.

Richard-Yunchao:
Please don't review this pr. I have revised it at my local side. I will upload the new one soon.

kainino0x:
Thanks for the heads up. FYI I just reviewed your test plan and I think it looks good, if we can manage to write the code for it. I changed one small thing and added a suggestion for the R-W and W-R tests.

Richard-Yunchao:
Thanks for reviewing the test plan. @kainino0x . This is the first test for write-after-write test. PTAL. Note that I may separate write-after-write into a few tests (for example, copy_and_pass_to pass, and copy_and_pass_to_copy, considering the complexity and the latter will change the destination buffer. Or 4 categories: {copy, pass} x {copy, pass}). The current test is pass to pass. Comments for how to organize the upcoming PRs are also welcome.

kainino0x:
Sorry I still haven't been able to get to this. There is too much to do. :)

Richard-Yunchao:
> Sorry I still haven't been able to get to this. There is too much to do. :)

No worries. I suppose the test plan is OK. So I am writing tests for the others at my local side.

src/webgpu/api/operation/memory_sync/buffer.spec.ts
```diff
@@ -0,0 +1,257 @@
...
+          storeOp: 'store',
+        },
+      ],
+    });
+  }
+
+  // Create render pass or compute pass, set pipeline and bind groups and write data into storage
+  // buffer. If the pass is render pass, it may use bundle. Note that we write 1 into the buffer
+  // if the pass is used in the first write, and we write 2 into the buffer otherwise.
+  createPass(
```
Kangz, line 147: The chain of ifs here, in createCommandBuffer and in the test is a bit difficult to understand. What about the following: we could have a "context" structure that can contain all 4 types of encoders. Each of the usage helper methods would call it with "ensureComputeEncoder()" similarly to what we do in Dawn with the Metal EncodingContext. This context structure would have the logic to end encoders or start new ones if needed.

Then the top level of the test can become:

```
context = new Context(device, queue);
value = 1;
operationType[firstOperation].performWrite(context, value);

// Either as a bunch of ifs, or as a helper that hides that logic in the context
if (differentBundle) {
    context.InvalidateBundle();
}
if (differentPass) {
    context.InvalidatePass();
}

value ++;
operationType[firstOperation].performWrite(context, value);
```

Turns out that with this structure the read tests become exactly the same thing!

src/webgpu/api/operation/memory_sync/buffer.spec.ts
```diff
@@ -0,0 +1,257 @@
...
+  verifyData(buffer: GPUBuffer, expectedValue: number) {
+    const bufferData = new Uint32Array(1);
+    bufferData[0] = expectedValue;
+    this.expectContents(buffer, bufferData);
+  }
+}
+
+export const g = makeTestGroup(BufferSyncTest);
+
+// Test write-after-write between passes. The first write will write 1 into a storage buffer.
```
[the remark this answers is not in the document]
kainino0x: nit: it's `g.test().desc()` which is per-test (g.desc would be per-file which is the same as the existing `description`)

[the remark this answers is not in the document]
Richard-Yunchao: Sounds like we don't need to change for the description part.

src/webgpu/api/operation/memory_sync/buffer.spec.ts
```diff
@@ -0,0 +1,257 @@
...
+    params()
+      // TODO (yunchao.he@intel.com): add multi-queue.
+      // Then the pbool 'sameCommandBuffer' can be a poption which consits of 3 options:
+      // same command buffer, separate command buffers, separate queues.
+      .combine(pbool('sameCommandBuffer'))
+      .combine(pbool('firstWriteInBundle'))
+      .combine(pbool('secondWriteInBundle'))
+      .combine(poptions('firstWrite', ['render', 'compute'] as const))
+      .combine(poptions('secondWrite', ['render', 'compute'] as const))
+      .unless(
```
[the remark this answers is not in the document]
Richard-Yunchao: Sure. But I would add other write operations like copy, writeBuffer, etc., in the next pull request. Can you explain "allowed scope" in details, I am glad to do that if it is not too complex. I think the current logic via "unless" is not complex either.

src/webgpu/api/operation/memory_sync/buffer.spec.ts
```diff
@@ -0,0 +1,257 @@
...
+    return this.device.createBindGroup({
+      layout: pipeline.getBindGroupLayout(0),
+      entries: [{ binding: 0, resource: { buffer } }],
+    });
+  }
+
+  // Create a compute pipeline and write data into storage buffer.
+  // If the pipeline is used in the first write, then write 1 into the buffer.
+  // Otherwise, write 2 into the buffer.
+  createComputePipeline(isFirstWrite: boolean): GPUComputePipeline {
```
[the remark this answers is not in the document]
Richard-Yunchao: Done. I used an exact value as a parameter.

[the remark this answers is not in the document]
Richard-Yunchao: Done. I changed the name from createComputePipeline to createStorageWriteComputePipeline. The latter name is more specific for the test.

src/webgpu/api/operation/memory_sync/buffer.spec.ts
```diff
@@ -0,0 +1,257 @@
...
+        size: { width: 4, height: 4, depth: 1 },
+        format: 'rgba8unorm',
+        usage: GPUTextureUsage.OUTPUT_ATTACHMENT,
+      })
+      .createView();
+    return encoder.beginRenderPass({
+      colorAttachments: [
+        {
+          attachment: view,
+          loadValue: { r: 0.0, g: 1.0, b: 0.0, a: 1.0 },
```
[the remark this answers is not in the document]
Richard-Yunchao: I tried to remove these two ops. I turned out that we can remove store op, but load value is required. I checked the WebGPU spec, which states that `loadValue` is requires (just as `attachment` is required).

src/webgpu/api/operation/memory_sync/buffer.spec.ts
Richard-Yunchao: Thanks for your suggestion, Corentin. I actually wrote tests for write operations like copy, etc as you mentioned above. It turned out that we can use helper functions here. I updated code to be aligned with the pull request followed. And the new one is easy to extend to more write operations like t2b copy, b2b copy, writeBuffer. Now that we have the helper functions, it is much easier to understand and it is extensible too. Can you take a look at this solution which is already there?

src/webgpu/api/operation/memory_sync/buffer.spec.ts
```diff
@@ -0,0 +1,261 @@
...
+    encoder: GPUCommandEncoder
+  ) {
+    const isRenderPass = writeOp === 'render';
+    assert(!bundle || isRenderPass);
+    if (isRenderPass) {
+      this.writeByRenderPass(bundle, buffer, value, encoder);
+    } else {
+      assert(writeOp === 'compute');
+      this.writeByComputePass(buffer, value, encoder);
+    }
```
Richard-Yunchao, line 185: We can add a few more if-else to cover other wrietOps like t2b copy, b2b copy, writeBuffer, etc.

src/webgpu/api/operation/memory_sync/buffer.spec.ts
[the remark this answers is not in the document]
Kangz: What I meant is that the content of this comment can be put in `g.test().desc()` so that it appears in the harness (and Kai pointed out I mistakenly said `g.desc()`). This is a new feature of `g.test()` that isn't used in any test but that AFAIK Kai would like to move towards having used eveywhere.

src/webgpu/api/operation/memory_sync/buffer.spec.ts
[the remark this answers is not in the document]
Kangz: Each of the read / write operations could be associated with the encoder they need to be used on (or no encoder). So for example `storage-write-compute` would be associated with `compute-encoder` while `map-write` would be associated with `none`. Then `compute-encoder`, `none` and friends can be used to do an `unless()` statement to see if `secondWriteInBundle`, `samePass` or `sameCommandBuffer` make sense.

src/webgpu/api/operation/memory_sync/buffer.spec.ts
Kangz: How do you handle the fact that some of them happen on passes, some on the command encoder, and some at the device-level? I worry there are going to be a lot if if-else that bobble information up and down the stack to know whether to finish passes, command buffers etc.

Basically the current approach seems to work fine for two types of operations that are both on passes but extensibility to the full test seems like it will result in very complicated and branchy code.

src/webgpu/api/operation/memory_sync/buffer.spec.ts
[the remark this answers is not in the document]
Richard-Yunchao: @Kangz , It looks like MapWrite (you mentioned above) upon a buffer can happen at any scopes? For example, inside a pass, on a command buffer but outside any pass, etc.

src/webgpu/api/operation/memory_sync/buffer.spec.ts
```diff
@@ -0,0 +1,261 @@
...
+  // Create a compute pipeline and write given data into storage buffer.
+  createStorageWriteComputePipeline(value: number): GPUComputePipeline {
+    const wgslCompute = `
+      type Data = [[block]] struct {
+        [[offset 0]] a : i32;
+      };
+
+      [[binding 0, set 0]] var<storage_buffer> data : Data;
+      fn main() -> void {
+	data.a = ${value};
```
kainino0x, line 56:
```suggestion
	  data.a = ${value};
```

Similar issues below.

src/webgpu/api/operation/memory_sync/buffer.spec.ts
```diff
@@ -0,0 +1,261 @@
...
+        return;
+      }
+
+      entry_point compute = main;
```
kainino0x, line 60: Please update to use the new syntax (`[[stage(compute)]]` decoration on `fn main` instead of this)

src/webgpu/api/operation/memory_sync/buffer.spec.ts
```diff
@@ -0,0 +1,261 @@
...
+      layout: pipeline.getBindGroupLayout(0),
+      entries: [{ binding: 0, resource: { buffer } }],
+    });
+  }
+
+  // Create a compute pipeline and write given data into storage buffer.
+  createStorageWriteComputePipeline(value: number): GPUComputePipeline {
+    const wgslCompute = `
+      type Data = [[block]] struct {
+        [[offset 0]] a : i32;
```
kainino0x, line 51: Please update to the new syntax for attributes, e.g.:
```suggestion
        [[offset(0)]] a : i32;
```

src/webgpu/api/operation/memory_sync/buffer.spec.ts
```diff
@@ -0,0 +1,261 @@
...
+      ],
+    });
+  }
+
+  // Write storage buffer via draw call in render pass. Use bundle if needed.
+  writeByRenderPass(bundle: boolean, buffer: GPUBuffer, value: number, encoder: GPUCommandEncoder) {
+    const pipeline = this.createStorageWriteRenderPipeline(value);
+    const bindGroup = this.createBindGroup(pipeline, buffer);
+    const pass = this.beginSimpleRenderPass(encoder);
+    if (bundle) {
```
kainino0x, line 142: Is it possible to use the helper from https://github.com/gpuweb/cts/pull/293?

src/webgpu/api/operation/memory_sync/buffer.spec.ts
kainino0x: Maybe a helper class abstracting the encoder types and operations would work here?

src/webgpu/api/operation/memory_sync/buffer.spec.ts
```diff
@@ -0,0 +1,261 @@
...
+// The second write will write 2 into the same storage buffer. So, expected data in buffer is 2.
+// The two writes can be in the same command buffer, or separate command buffers, or separate
+// queues. Each write operation can be done via render, compute, or copy, etc. If the write
+// operation is done by a render pass, it may use bundle.
+g.test('write_after_write,passes')
+  .params(
+    params()
+      // TODO (yunchao.he@intel.com): add multi-queue.
+      // Then the pbool 'sameCommandBuffer' can be a poption which consits of 3 options:
+      // same command buffer, separate command buffers, separate queues.
```
kainino0x, line 218: Should probably test separate submit()s as well (this can be done now).

src/webgpu/api/operation/memory_sync/buffer.spec.ts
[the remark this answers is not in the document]
kainino0x: No, mapping occurs outside of submit()s, so it can only be map-before-submit or submit-before-map.

kainino0x (commented):
Sorry it's taken me so long to look at this. Added a few comments. @Kangz could you continue to work with Yunchao on the open comments?

Overall I would like to make this cover the complete matrix of pairs of possible reads/writes. Helpers that can do functionally-equivalent reads and writes in every possible way would be complicated but would hopefully make the tests easy to write (and review). It would probably be a helper class, because in a way these tests are stateful (what type of encoder is current) and it should encapsulate that.

src/webgpu/api/operation/memory_sync/buffer.spec.ts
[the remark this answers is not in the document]
Richard-Yunchao: @kainino0x , just to verify. It looks like WriteBuffer is outside of submit()s, and MapWrite is not? Actually I am using mapwrite at line 32 - 34, although that code snippet is not necessary. Is that code snippet mapwrite?

src/webgpu/api/operation/memory_sync/buffer.spec.ts
Richard-Yunchao: Maybe this new syntax is supported in spec, but It is not supported in Chromium/Dawn yet. I updated my local code of Chromium and built the new code and tried for the new syntax again. It turned out that both the suggested code and my previous wgsl code fail. So I can't verify the tests in Chromium now. Do you know the recent changes (2 or 3 weeks, I think) in wgsl in Chromium/Dawn?

src/webgpu/api/operation/memory_sync/buffer.spec.ts
Richard-Yunchao: Looks like this new syntax is not supported in Chromium/Dawn yet either.

src/webgpu/api/operation/memory_sync/buffer.spec.ts
Richard-Yunchao: The suggested change works in latest Chromium/Dawn. I looked into the error reported by tint and revised [[builtin location]] to [[builtin(location)]], then tests passed! And the previous syntax is not working in latest Chromium/Dawn.

src/webgpu/api/operation/memory_sync/buffer.spec.ts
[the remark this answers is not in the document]
Richard-Yunchao: Done.

src/webgpu/api/operation/memory_sync/buffer.spec.ts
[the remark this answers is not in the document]
Richard-Yunchao: Added all ways to write into a buffer Corentin suggested. According to your suggestion, both writeBuffer and mapWrite are outside of a queue submit.

src/webgpu/api/operation/memory_sync/buffer.spec.ts
Richard-Yunchao: I use different levels of wrapper functions instead, it's c/c++-style though. But the logic is easy to understand too.

src/webgpu/api/operation/memory_sync/buffer.spec.ts
```diff
@@ -0,0 +1,261 @@
...
+import { GPUTest } from '../../../gpu_test.js';
+
+const SIZE = 4;
+class BufferSyncTest extends GPUTest {
+  // Create a buffer, and initiate it to zero.
+  createBuffer(): GPUBuffer {
+    const data = new Uint32Array(SIZE / 4);
+    for (let i = 0; i < SIZE / 4; ++i) {
+      data[i] = 0;
+    }
```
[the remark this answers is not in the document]
Richard-Yunchao: Ack. But I need to use this function to assign any given value (not only zero) to a buffer.

src/webgpu/api/operation/memory_sync/buffer.spec.ts
```diff
@@ -0,0 +1,261 @@
...
+`;
+
+import { pbool, poptions, params } from '../../../../common/framework/params_builder.js';
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { assert } from '../../../../common/framework/util/util.js';
+import { GPUTest } from '../../../gpu_test.js';
+
+const SIZE = 4;
+class BufferSyncTest extends GPUTest {
+  // Create a buffer, and initiate it to zero.
```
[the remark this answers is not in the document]
Richard-Yunchao: Ack. But I need to use this function to assign any given value (not only zero) to a buffer.

src/webgpu/api/operation/memory_sync/buffer.spec.ts
Richard-Yunchao: Done

src/webgpu/api/operation/memory_sync/buffer.spec.ts
Richard-Yunchao: Done.

src/webgpu/api/operation/memory_sync/buffer.spec.ts
Richard-Yunchao: True, actually I added tests for separate submits in my subsequent pr at my local side. I merged them into this one, including other write operations like copy, etc.(in addition to render and compute), and separate submits (in addition to one single cmdbuf, and separate cmdbufs).

src/webgpu/api/operation/memory_sync/buffer.spec.ts
```diff
@@ -0,0 +1,367 @@
...
+  }
+
+  // Write buffer via mapping and writing
+  writeByMapWrite(buffer: GPUBuffer, value: number) {
+    const data = new Uint32Array(SIZE / 4);
+    for (let i = 0; i < SIZE / 4; ++i) {
+      data[i] = value;
+    }
+
+    buffer.mapAsync(GPUMapMode.WRITE);
```
Richard-Yunchao, line 202: mapSync is async, looks like we need an `await` or stuff like that? But I can't use `await` here because this function is not async. Any idea about this?

Richard-Yunchao:
Thanks for your review, @Kangz and @kainino0x . It is a little bit late to address all comments because the changes are relatively significant. PTAL.

src/webgpu/api/operation/memory_sync/buffer.spec.ts
Richard-Yunchao: Hmm, I need to test different `scopes`, so I don't need to use `finish` in that function, which always suppose that the test need a separate command encoder (it is not true for this test). So it will become unused variable if I use that function.

src/webgpu/api/operation/memory_sync/buffer.spec.ts
kainino0x: Why can't you make writeByMapWrite async (and so on up the call stack)?

Richard-Yunchao: Done. But the tests still fail and complained that "The buffer must have the MapWrite usage" when I run the tests in Chromium. anything wrong in tests?

kainino0x: The buffer is not created with the MAP_WRITE bit. `mappedAtCreation` buffers cannot be remapped after creation unless they have that bit.

Richard-Yunchao: If I understand correctly, MAP_WRITE usage can be compatible with COPY_SRC only. This means that MapWrite operation is impossible to be included into write_after_write test? Because another write operation is either a storage buffer or a COPY_DST buffer, which is not compatible with MAP_WRITE usage.

src/webgpu/api/operation/memory_sync/buffer.spec.ts
Richard-Yunchao: Done

src/webgpu/api/operation/memory_sync/buffer.spec.ts
Richard-Yunchao: removed mapWrite writeOperation related tests for write-after-write tests. We can add it when we test read-after-write and read-before-write.

kainino0x: Right. Sorry I didn't catch that. We may not want do the map synchronization tests in this file, since it would only be tested in conjunction with copies.

That said, if it were tightly abstracted along with the rest of the ways of writing to and reading from buffers, it would probably be easiest to do it here (would just have to filter out the test cases that wouldn't be valid).

kainino0x (commented):
Latest changes LGTM. I'm sorry I haven't had time to review the whole thing in a while, I am not currently certain about whether this is implementing the direction we want to take or not.

Maybe tomorrow we can chat on slack (maybe in the morning with @Kangz too?) about what the current state of these tests (this PR + the later ones) is? And also talk about fully general read-op/write-op abstractions that should be generally useful in the test suite.

Richard-Yunchao:
> Latest changes LGTM. I'm sorry I haven't had time to review the whole thing in a while, I am not currently certain about whether this is implementing the direction we want to take or not.
> 
> Maybe tomorrow we can chat on slack (maybe in the morning with @Kangz too?) about what the current state of these tests (this PR + the later ones) is? And also talk about fully general read-op/write-op abstractions that should be generally useful in the test suite.

No problem for me. We can discuss via slack tomorrow morning if needed. What about the time around 9am or 10am? 

My rough idea is:
* Put all helper functions, for example, createBuffer, createBindGroup, writeByRenderPass, etc., in a separate file. And write tests in write-after-write-in-buffer.spec.ts, read-after-write-in-buffer.spec.ts, and read-before-write-in-buffer.spec.ts. 
* In each test, it might has a couple of comprehensive tests of combined tests to cover most cases, while a few separated tests to cover some special cases. Take write-after-write as an example, the comprehensive test is the current one we are reviewing (write-after-write), but we need to test two writes in the same render or compute pass. The latter tests are not easy to combined into the current one, and the result of render pass (two writes in the same render pass) should have loose guarantees.

kainino0x:
Thanks for the summary, that will definitely be helpful to discuss.

src/webgpu/api/operation/memory_sync/buffer.spec.ts
Richard-Yunchao: Yeah. I intended to do so, too. MapWrite can be tested in read-after-write or read-before-write. And it might be in a separate tests (not combined with others) because it can be tested with copy only for the sake of buffer usage compatibility.

Richard-Yunchao:
@kainino0x , I added fence for buffer initialization, Please take another look.

kainino0x:
Latest change looks good. Did you have a chance to work on the plan document we talked about for these tests? (If so, I haven't seen it, sorry!)

Richard-Yunchao:
> Latest change looks good. Did you have a chance to work on the plan document we talked about for these tests? (If so, I haven't seen it, sorry!)

Yeah. I am working on that. Will send out it soon.

Richard-Yunchao:
@kainino0x , Is this change ready for being merged? What do I need to do if it is not ready?

kainino0x:
> @kainino0x , Is this change ready for being merged? What do I need to do if it is not ready?

I can't rereview this in detail tonight but I don't think it matches the structure in the plan doc. I don't think it will be easier for me to review the changes than to just rereview the tests from scratch once those changes have been made, so I think I'd rather update this PR to match at least the high level structure (file/test/parameterization structure, test descriptions, helpers), rather than landing it and then moving a bunch of code around.

Richard-Yunchao:
> > @kainino0x , Is this change ready for being merged? What do I need to do if it is not ready?
> 
> I can't rereview this in detail tonight but I don't think it matches the structure in the plan doc. I don't think it will be easier for me to review the changes than to just rereview the tests from scratch once those changes have been made, so I think I'd rather update this PR to match at least the high level structure (file/test/parameterization structure, test descriptions, helpers), rather than landing it and then moving a bunch of code around.

No problem. I separated the helper functions into a standalone file. I think the vast majority (if not all) of the helper functions can be shared among tests. 

I noticed your latest comment, are you saying that the helper functions should not be put in a dedicated class for buffer sync test (I named the class as BufferSyncTest though)?  

The bot failed, but it looks like it is not caused by the test code according to logs?

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
```diff
@@ -0,0 +1,242 @@
+import { assert } from '../../../../../common/framework/util/util.js';
+import { GPUTest } from '../../../../gpu_test.js';
+
+const SIZE = 4;
+export class BufferSyncTest extends GPUTest {
+  // Create a buffer, and initiate it to a specified value for all elements.
+  async createBufferWithValue(initValue: number): Promise<GPUBuffer> {
+    const fence = this.queue.createFence();
+    this.queue.signal(fence, 1);
```
Kangz, line 9: The fence is signaled before the buffer is unmapped so if we want to make sure any transfer operations are finished, we should do it after the unmap call.

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
```diff
@@ -0,0 +1,242 @@
...
+          code: wgslCompute,
+        }),
+        entryPoint: 'main',
+      },
+    });
+  }
+
+  // Create a render pipeline and write given data into storage buffer at fragment stage.
+  createStorageWriteRenderPipeline(value: number): GPURenderPipeline {
+    const wgslVertex = `
```
Kangz, line 61: nit: I think both entrypoints could be in the same module probably?

Unless it triggers a validation error, which would be because we don't properly reflect WGSL bindings yet.

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
```diff
@@ -0,0 +1,242 @@
...
+      },
+      primitiveTopology: 'point-list',
+      colorStates: [{ format: 'rgba8unorm' }],
+    });
+  }
+
+  beginSimpleRenderPass(encoder: GPUCommandEncoder): GPURenderPassEncoder {
+    const view = this.device
+      .createTexture({
+        size: { width: 4, height: 4, depth: 1 },
```
Kangz, line 104: nit: the texture can be 1x1 (and depth omitted)

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
```diff
@@ -0,0 +1,242 @@
...
+    const tmpBuffer = await this.createBufferWithValue(value);
+    encoder.copyBufferToBuffer(tmpBuffer, 0, buffer, 0, SIZE);
+  }
+
+  // Write buffer via TextureToBuffer copy.
+  async writeByT2BCopy(buffer: GPUBuffer, value: number, encoder: GPUCommandEncoder) {
+    const tmpBuffer = await this.createBufferWithValue(value);
+    const tmpTexture = this.device.createTexture({
+      size: { width: 1, height: 1, depth: 1 },
+      format: 'rgba8uint',
```
Kangz, line 158: u-nit: I think r32uint would make things marginally clearer.

src/webgpu/api/operation/memory_sync/buffer/write_after_write_in_buffer.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+    The second write will write 2 into the same storage buffer. So, expected data in buffer is 2.
+    The two writes can be in the same command buffer, or separate command buffers, or separate
+    submits, or separate queues. Each write operation can be done via render, compute, copy,
+    writeBuffer, map write, etc. If the write operation is done by a render pass, it may use bundle.`
+  )
+  .params(
+    params()
+      // TODO (yunchao.he@intel.com): add multi-queue.
+      .combine(
+        poptions('writeScopes', ['sameCmdbuf', 'separateCmdbufs', 'separateSubmits'] as const)
```
Kangz, line 34: I think we're missing samePass, especially for compute passes?

Kangz (commented):
Looks much better. I think the idea about a context with various encoders could still prove useful in the future but I don't think we need to block on it.

kainino0x:
> I noticed your latest comment, are you saying that the helper functions should not be put in a dedicated class for buffer sync test (I named the class as BufferSyncTest though)?

Yes, I don't think they need to be part of a fixture; they can just be standalone helper functions that take the GPUDevice as an argument.

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
```diff
@@ -0,0 +1,242 @@
...
+      usage: GPUTextureUsage.COPY_SRC | GPUTextureUsage.COPY_DST,
+    });
+    encoder.copyBufferToTexture(
+      { buffer: tmpBuffer, bytesPerRow: 256 },
+      { texture: tmpTexture, mipLevel: 0, origin: { x: 0, y: 0, z: 0 } },
+      { width: 1, height: 1, depth: 1 }
+    );
```
kainino0x, line 165: The initialization of the source texture should ideally be separated from the test copyTextureToBuffer call by a fence, like is done for the buffer initialization. Instead of `await createBufferWithValue`+`createTexture`+`copyBufferToTexture`, consider `createTexture`+`writeTexture`+`signal`+`await onCompletion`.

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
```diff
@@ -0,0 +1,242 @@
...
+        {
+          attachment: view,
+          loadValue: { r: 0.0, g: 1.0, b: 0.0, a: 1.0 },
+        },
+      ],
+    });
+  }
+
+  // Write buffer via draw call in render pass. Use bundle if needed.
+  writeByRenderPass(bundle: boolean, buffer: GPUBuffer, value: number, encoder: GPUCommandEncoder) {
```
kainino0x, line 120: nit: encodeWriteAsStorageBufferInRenderPass

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
```diff
@@ -0,0 +1,242 @@
...
+    renderer.setBindGroup(0, bindGroup);
+    renderer.setPipeline(pipeline);
+    renderer.draw(1, 1, 0, 0);
+
+    if (bundle) pass.executeBundles([(renderer as GPURenderBundleEncoder).finish()]);
+    pass.endPass();
+  }
+
+  // Write buffer via dispatch call in compute pass.
+  writeByComputePass(buffer: GPUBuffer, value: number, encoder: GPUCommandEncoder) {
```
kainino0x, line 137: nit: encodeWriteAsStorageBufferInComputePass

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
kainino0x, line 120: uNit: I think the order of args would be more naturally (encoder, buffer, value, inBundle).

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
```diff
@@ -0,0 +1,242 @@
...
+    const bindGroup = this.createBindGroup(pipeline, buffer);
+    const pass = encoder.beginComputePass();
+    pass.setPipeline(pipeline);
+    pass.setBindGroup(0, bindGroup);
+    pass.dispatch(1);
+    pass.endPass();
+  }
+
+  // Write bufer via BuferToBuffer copy.
+  async writeByB2BCopy(buffer: GPUBuffer, value: number, encoder: GPUCommandEncoder) {
```
kainino0x, line 148: nit: encodeWriteByB2BCopy, and so on below.

It's really important that these encode writes, not that they actually write. In particular, for the async ones,f they actually wrote immediately, then the async tmpBuffer initialization would happen inline instead of ahead of time. In addition to the naming change, maybe add a comment clarifying this.

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
```diff
@@ -0,0 +1,242 @@
...
+      { width: 1, height: 1, depth: 1 }
+    );
+  }
+
+  // Write buffer via writeBuffer API on queue
+  writeByWriteBuffer(buffer: GPUBuffer, value: number) {
+    const data = new Uint32Array(SIZE / 4);
+    for (let i = 0; i < SIZE / 4; ++i) {
+      data[i] = value;
+    }
```
kainino0x, line 178: nit: use `.fill`

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
```diff
@@ -0,0 +1,242 @@
...
+
+    this.device.defaultQueue.writeBuffer(buffer, 0, data);
+  }
+
+  // Issue write operation via render pass, compute pass, copy, etc.
+  async issueWriteOp(
+    writeOp: string,
```
kainino0x, line 185:
```suggestion
    writeOp: 'render' | 'compute' | 'b2bCopy' | 't2bCopy',
```
and same below

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
```diff
@@ -0,0 +1,242 @@
...
+    bundle: boolean,
+    buffer: GPUBuffer,
+    value: number,
+    encoder: GPUCommandEncoder
+  ) {
+    assert(!bundle || writeOp === 'render');
```
kainino0x, line 191: IMO, make `bundle` one of the possible writeOps ('render-via-bundle') instead of asserting it's only used with writeOp 'render'.

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
```diff
@@ -0,0 +1,242 @@
...
+    const data = new Uint32Array(SIZE / 4);
+    for (let i = 0; i < SIZE / 4; ++i) {
+      data[i] = value;
+    }
+
+    this.device.defaultQueue.writeBuffer(buffer, 0, data);
+  }
+
+  // Issue write operation via render pass, compute pass, copy, etc.
+  async issueWriteOp(
```
kainino0x, line 184: nit: encodeWriteOp

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
```diff
@@ -0,0 +1,242 @@
...
+      case 't2bCopy':
+        await this.writeByT2BCopy(buffer, value, encoder);
+        break;
+      default:
+        assert(true);
+        break;
+    }
+  }
+
+  async createCommandBufferAndIssueWriteOp(
```
kainino0x, line 211: createCommandBufferWithWriteOp

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
```diff
@@ -0,0 +1,242 @@
...
+    bundle: boolean,
+    buffer: GPUBuffer,
+    value: number
+  ): Promise<GPUCommandBuffer> {
+    const encoder = this.device.createCommandEncoder();
+    await this.issueWriteOp(writeOp, bundle, buffer, value, encoder);
+    return encoder.finish();
+  }
+
+  async createQueueSubmitsAndIssueWriteOp(
```
kainino0x, line 222: submitWriteOp

src/webgpu/api/operation/memory_sync/buffer/write_after_write_in_buffer.spec.ts
kainino0x: Plan was to make this a separate test. There's a TODO for it below which can become an `unimplemented`.

src/webgpu/api/operation/memory_sync/buffer/write_after_write_in_buffer.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+        ] as const)
+      )
+      .combine(pbool('firstWriteInBundle'))
+      .combine(pbool('secondWriteInBundle'))
+      .unless(
+        p =>
+          (p.firstWriteInBundle && p.firstWriteOp !== 'render') ||
+          (p.secondWriteInBundle && p.secondWriteOp !== 'render') ||
+          ((p.firstWriteOp === 'writeBuffer' || p.secondWriteOp === 'writeBuffer') &&
+            p.writeScopes !== 'separateSubmits')
```
kainino0x, line 61: pbool+pbool+unless will go away if 'render-via-bundle' is added as one of the writeOps as in the plan doc.

src/webgpu/api/operation/memory_sync/buffer/write_after_write_in_buffer.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+      secondWriteOp,
+      firstWriteInBundle,
+      secondWriteInBundle,
+    } = t.params;
+
+    const buffer = await t.createBufferWithValue(0);
+
+    const writeInBundle = [firstWriteInBundle, secondWriteInBundle];
+    const writeOp = [firstWriteOp, secondWriteOp];
+    switch (writeScopes) {
```
kainino0x, line 77: There's essentially no code in common between these writeScopes modes. Let's make them separate tests. You could store the params builder into a variable and reuse it between the tests, though I don't think it's very necessary in this case.

src/webgpu/api/operation/memory_sync/buffer/write_after_write_in_buffer.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+        break;
+      }
+      case 'separateSubmits': {
+        for (let i = 0; i < 2; i++) {
+          await t.createQueueSubmitsAndIssueWriteOp(writeOp[i], writeInBundle[i], buffer, i + 1);
+        }
+        break;
+      }
+      default:
+        assert(true);
```
kainino0x, line 103: I think you meant `assert(false)` though `unreachable()` would also work. This code will disappear with the suggestion above though.

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
```diff
@@ -0,0 +1,242 @@
...
+        this.writeByComputePass(buffer, value, encoder);
+        break;
+      case 'b2bCopy':
+        await this.writeByB2BCopy(buffer, value, encoder);
+        break;
+      case 't2bCopy':
+        await this.writeByT2BCopy(buffer, value, encoder);
+        break;
+      default:
+        assert(true);
```
kainino0x, line 206: assert(false)

src/webgpu/api/operation/memory_sync/buffer/write_after_write_in_buffer.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+        ] as const)
+      )
+      .combine(
+        poptions('secondWriteOp', [
+          'render',
+          'compute',
+          'b2bCopy',
+          't2bCopy',
+          'writeBuffer',
+        ] as const)
```
kainino0x, line 52: nit: Factor out this duplicated list and give it a name (e.g. kAllWriteOps). That constant can go in the helper file.

kainino0x (commented):
Many small nits but looking good overall!

src/webgpu/api/operation/memory_sync/buffer/write_after_write_in_buffer.spec.ts
```diff
@@ -0,0 +1,113 @@
+export const description = `
```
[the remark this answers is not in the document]
Richard-Yunchao: Ack. I renamed to ww.spec.ts, and added rw_and_wr.spec.ts (but it is empty).

src/webgpu/api/operation/memory_sync/buffer/write_after_write_in_buffer.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+      }
+      default:
+        assert(true);
+        break;
+    }
+
+    t.verifyData(buffer, 2);
+  });
+
+// TODO (yunchao.he@intel.com):
```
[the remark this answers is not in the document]
Richard-Yunchao: Done.

src/webgpu/api/operation/memory_sync/buffer/write_after_write_in_buffer.spec.ts
```diff
@@ -0,0 +1,113 @@
+export const description = `
+Memory Synchronization Tests for Buffer.
+
+- Test write-after-write on a single buffer. Create one single buffer and initiate it to 0.
```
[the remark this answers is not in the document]
Richard-Yunchao: Done.

src/webgpu/api/operation/memory_sync/buffer/write_after_write_in_buffer.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+import { makeTestGroup } from '../../../../../common/framework/test_group.js';
+import { assert } from '../../../../../common/framework/util/util.js';
+
+import { BufferSyncTest } from './buffer_sync_test.js';
+
+export const g = makeTestGroup(BufferSyncTest);
+
+g.test('write_after_write')
+  .desc(
+    `Test write-after-write operations. The first write will write 1 into a storage buffer.
```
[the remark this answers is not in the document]
Richard-Yunchao: Ack. I revised it.

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
Richard-Yunchao: Done.

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
Richard-Yunchao: I tried, but it failed. I did notice that we use one single module for vertex code and fragment code.

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
Richard-Yunchao: Done.

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
Richard-Yunchao: I tried but It failed

src/webgpu/api/operation/memory_sync/buffer/write_after_write_in_buffer.spec.ts
Richard-Yunchao: Ack. Add a few separate test as `unimplemented`

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
```diff
@@ -0,0 +1,242 @@
+import { assert } from '../../../../../common/framework/util/util.js';
+import { GPUTest } from '../../../../gpu_test.js';
+
+const SIZE = 4;
```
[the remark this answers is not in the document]
Richard-Yunchao: Done. Thanks for your suggestion, Kai!

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
```diff
@@ -0,0 +1,242 @@
...
+export class BufferSyncTest extends GPUTest {
+  // Create a buffer, and initiate it to a specified value for all elements.
+  async createBufferWithValue(initValue: number): Promise<GPUBuffer> {
+    const fence = this.queue.createFence();
+    this.queue.signal(fence, 1);
+    const data = new Uint32Array(SIZE / 4);
+    for (let i = 0; i < SIZE / 4; ++i) {
+      data[i] = initValue;
+    }
```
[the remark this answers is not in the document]
Richard-Yunchao: It seems to me that it will fail if I write data directly into the mapped memory. But `.fill` works.

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
Richard-Yunchao: Done.

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
Richard-Yunchao: Done.

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
Richard-Yunchao: Done.

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
Richard-Yunchao: Done.

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
Richard-Yunchao: Done. I added a comment as you suggested, but I am not 100% sure whether it is appropriate, though.

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
Richard-Yunchao: Done.

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
Richard-Yunchao: We need to change the `writeOp` for the entire call stack. Seems that it is not necessary. And `writeOp`s are slightly different (w/ or w/o `write-buffer`). So I didn't change it. I think it is not a big deal.

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
Richard-Yunchao: Ack. Good suggestion!

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
Richard-Yunchao: Done.

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
Richard-Yunchao: Done.

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
Richard-Yunchao: Done. Sorry for inappropriate names...

src/webgpu/api/operation/memory_sync/buffer/write_after_write_in_buffer.spec.ts
Richard-Yunchao: Ack. Done.

src/webgpu/api/operation/memory_sync/buffer/write_after_write_in_buffer.spec.ts
Richard-Yunchao: Done. Yeah, it is not necessary to store params into a variable because there are only two params. They are not complicated as what they were.

src/webgpu/api/operation/memory_sync/buffer/write_after_write_in_buffer.spec.ts
Richard-Yunchao: Ack. Done.

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
Richard-Yunchao: Done.

src/webgpu/api/operation/memory_sync/buffer/write_after_write_in_buffer.spec.ts
Richard-Yunchao: Good suggestion! Done.

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
```diff
@@ -0,0 +1,253 @@
+import { assert } from '../../../../../common/framework/util/util.js';
+import { GPUTest } from '../../../../gpu_test.js';
+
+const kSize = 4;
+
+export class BufferSyncTest extends GPUTest {
```
Richard-Yunchao, line 6: @kainino0x , you mentioned that it is not necessary to put helper functions into a fixture. Do you mean that we don't need to define `class BufferSyncTest`? In that case, how can I include the helper functions into ww.spec.ts?

Richard-Yunchao:
Thanks for your detailed review, Corentin and Kai. I forgot to use `git merge` but rebased the code in the middle in order to fix a build error after I updated my local main branch. Sorry for that. It might be inconvenient for you to review.

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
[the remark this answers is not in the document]
kainino0x: I don't understand why it would fail. It sounds like a bug? If so, we need to capture it as a regression test so we can fix it.

src/webgpu/api/operation/memory_sync/buffer/write_after_write_in_buffer.spec.ts
kainino0x: OK. It shouldn't fail to build. If we want to reuse it in the next patch for rw/wr we can move it.

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
kainino0x: Ack

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
kainino0x: Wording looks good, but actually please include it in the comment on each of these functions, and change the comment style to a `/** doc comment */` so it will be picked up by IDEs.

Also nit: bufer -> buffer

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
kainino0x: Why?

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
kainino0x: How did it fail? Is it due to incompleteness in Tint/Dawn?

src/webgpu/api/operation/memory_sync/buffer/write_after_write_in_buffer.spec.ts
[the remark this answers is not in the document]
kainino0x: I don't see the new file, do you need to `git add` it?

kainino0x (approved):
np, in this case I just reviewed the two new commits assuming there's nothing new before that.

kainino0x:
I cleaned up the "resolved" comment state so every comment that isn't marked as resolved still needs attention.

src/webgpu/api/operation/memory_sync/buffer/write_after_write_in_buffer.spec.ts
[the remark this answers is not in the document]
Richard-Yunchao: Oops... Done.

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
Richard-Yunchao: If failed because my local Chromium was too old. It can success now. Done!

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
Richard-Yunchao: My local Chromium was old. Done!

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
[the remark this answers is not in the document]
Richard-Yunchao: It works. Done!

src/webgpu/api/operation/memory_sync/buffer/write_after_write_in_buffer.spec.ts
Richard-Yunchao: It works! Done.

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
Richard-Yunchao: Done.

src/webgpu/api/operation/memory_sync/buffer/ww.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+import { poptions, params } from '../../../../../common/framework/params_builder.js';
+import { makeTestGroup } from '../../../../../common/framework/test_group.js';
+
+import { BufferSyncTest } from './buffer_sync_test.js';
+
+export const g = makeTestGroup(BufferSyncTest);
+
+const kAllWriteOps = ['render', 'render-via-bundle', 'compute', 'b2b-copy', 't2b-copy'];
+
+g.test('write_after_write,same_cmdbuf')
```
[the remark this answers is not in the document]
Richard-Yunchao: Ack. Done!

Richard-Yunchao:
All comments were resolved. PTAL @kainino0x

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
```diff
@@ -0,0 +1,258 @@
...
+    const pipeline = this.createStorageWriteComputePipeline(value);
+    const bindGroup = this.createBindGroup(pipeline, buffer);
+    const pass = encoder.beginComputePass();
+    pass.setPipeline(pipeline);
+    pass.setBindGroup(0, bindGroup);
+    pass.dispatch(1);
+    pass.endPass();
+  }
+
+  // Write bufer via BuferToBuffer copy.
```
kainino0x, line 179:
```suggestion
  /** Write buffer via BufferToBuffer copy. */
```

src/webgpu/api/operation/memory_sync/buffer/rw_and_wr.spec.ts
```diff
@@ -0,0 +1,20 @@
+export const description = `
+Memory Synchronization Tests for Buffer: read before write and read after write.
+
+- Create a single buffer and initialize it to 0, wait on the fence to ensure the data is initialized.
+Write a number (say 1) into the buffer via render pass, compute pass, copy or writeBuffer.
+Read the data and use it in render, compute, or copy.
+Wait on another fence, then call expectContents to verify the written buffer.
+This is a read-after write test but if the write and read operations are reverted, it will be a read-before-write test.
```
kainino0x, line 8:
```suggestion
This is a read-after write test but if the write and read operations are reversed, it will be a read-before-write test.
```

kainino0x (approved):

kainino0x:
LGTM

PR #303, opened 2020-10-05
Update WGSL tests

sarahM0:
This CL:
Cleans up WGSL validation tests to follow entry point spec changes.
Adds cts tests for switch statements.
Fixes other minor syntax issues, eg. adding parenthesis for location, offset, ...


src/webgpu/shader/validation/wgsl/duplicate-var-name-within-func.fail.wgsl
```diff
@@ -1,13 +1,11 @@
-# v-0014 - This fails because of the duplicated `a` variable.
+# v-0014 - This passes because variable `a` is redeclared.
```
kainino0x, line 1: I know we agreed on lexical scoping, but did we agree to allow shadowing? @dj2

src/webgpu/shader/validation/wgsl/switch-must-have-exactly-one-default-clause.fail.wgsl
```diff
@@ -0,0 +1,11 @@
+# v-0008: line 6: switch statement must have exactly one default clause
```
kainino0x, line 1: I keep losing track of this one - is this what we agreed? I thought it was okay to not have a default clause (the semantics should be pretty obvious I think). Not trying to change this, just want to make sure it's the right thing.

kainino0x (approved):

src/webgpu/shader/validation/wgsl/duplicate-entry-point.fail.wgsl
```diff
@@ -1,12 +1,11 @@
 # v-0020 - Duplicate entry point <name, stage> pair.
```
dj2, line 1: This isn't quite true any more. You can't have a duplicate entry point name at all anymore, regardless of the stage. So, we should add a second test with `stage(vertex)` and `stage(fragment)` with the same name to verify it fails.

src/webgpu/shader/validation/wgsl/duplicate-var-name-within-func.fail.wgsl
dj2: As far as I remember, shadowing is _not_ allowed.

src/webgpu/shader/validation/wgsl/switch-case-selector-must-have-the-same-type-as-the-selector-expression-2.fail.wgsl
```diff
@@ -0,0 +1,12 @@
+# v-0026: line 7: the case selector values must have the same type as the selector expression
+
+[[stage(vertex)]]
+fn main() -> void {
+  var a: i32 = -2;
```
dj2, line 5: Can we add a passing case for switch statements?

src/webgpu/shader/validation/wgsl/switch-must-have-exactly-one-default-clause-2.fail.wgsl
```diff
@@ -0,0 +1,13 @@
+# v-0008: line 6: switch statement must have exactly one default clause
```
dj2, line 1: nit: s/exactly one/at most one/

src/webgpu/shader/validation/wgsl/switch-must-have-exactly-one-default-clause.fail.wgsl
dj2: I believe it's `at most one` although the spec isn't updated yet. @dneto0 to verify my memory.

src/webgpu/shader/validation/wgsl/duplicate-var-name-within-func.fail.wgsl
kainino0x: sg, in that case we should test both the old test (siblings, passes) and new test (nested, fails)

src/webgpu/shader/validation/wgsl/switch-must-have-exactly-one-default-clause-2.fail.wgsl
dj2: I'm wrong, @dneto0 pointed out must have is correct.

src/webgpu/shader/validation/wgsl/switch-must-have-exactly-one-default-clause.fail.wgsl
dj2: I'm wrong, it's must have currently.

sarahM0:
Sorry for the delay, I totally forgot about this PR.

dj2 (approved):

PR #308, opened 2020-10-16
Add API validation tests for compute pass

egalli:
Code is missing the upper limit on dispatchIndirect. We are currently working on an investigation to find the upper limit.

### DIFF
--- a/src/webgpu/api/validation/encoding/cmds/compute_pass.spec.ts
+++ b/src/webgpu/api/validation/encoding/cmds/compute_pass.spec.ts
@@ -46,9 +46,9 @@ export const g = makeTestGroup(F);
 
 g.test('set_pipeline')
   .desc(
-    `- Tests using compute pipelines
-  - An error should be generated when using an 'invalid' pipeline`
-  )
+    `
+setPipeline should generate an error iff using an 'invalid' pipeline.
+`)
   .params(poptions('state', ['valid', 'invalid'] as const))
   .fn(t => {
     const pipeline = t.createComputePipeline(t.params.state);
@@ -61,9 +61,12 @@ g.test('set_pipeline')
 
 g.test('dispatch_sizes')
   .desc(
-    `- For both 'direct' and 'indirect' dispatch
-  - Tests using the following workgroup sizes: {[0, 0, 0], [1, 1, 1]}
-  - An error should occur when the number exceeds <fill number here>`
+    `
+Test 'direct' and 'indirect' dispatch with various sizes.
+  - workgroup sizes:
+    - valid, {[0, 0, 0], [1, 1, 1]}
+    - invalid,  <fill number here>
+`
   )
   .params(
     params()
@@ -72,8 +75,9 @@ g.test('dispatch_sizes')
         poptions('workSizes', [
           [0, 0, 0],
           [1, 1, 1],
-          // TODO: Add tests for workSizes right under and above upper limit once the limit has been decided.
-        ] as [number, number, number][])
+          // TODO: Add tests for workSizes right under and above upper limit once the limit has
+          //  been decided.
+        ] as const)
       )
   )
   .fn(t => {
@@ -92,15 +96,15 @@ g.test('dispatch_sizes')
 const kBufferData = new Uint32Array(6).fill(1);
 g.test('indirect_dispatch_buffer')
   .desc(
-    `- For 'indirect' dispatch
-  - Tests that indirect buffers:
-    - 'invalid', 'destroyed' generate an error
-    - 'valid' buffers do not generate an error
-  - Tests that, for a buffer with 6 elements, indirect offsets:
-    - 0, 'sizeof(uint32)', and '3 * sizeof(uint32)' do not generate an error
-    - An error should be generate by the following:
-      - 1: non-multiple of 4
-      - '4 * sizeof(uint32): z-components outside of buffer`
+    `
+Test dispatchIndirect validation by submitting various dispatches with a no-op pipeline and an
+indirectBuffer with 6 elements.
+- indirectBuffer: {'valid', 'invalid', 'destroyed'}
+- indirectOffset:
+  - valid, within the buffer: {beginning, middle, end} of the buffer
+  - invalid, non-multiple of 4
+  - invalid, the last element is outside the buffer
+`
   )
   .params(
     params()
@@ -111,7 +115,8 @@ g.test('indirect_dispatch_buffer')
           Uint32Array.BYTES_PER_ELEMENT, // valid for 'valid' buffers
           kBufferData.byteLength - 3 * Uint32Array.BYTES_PER_ELEMENT, // valid for 'valid' buffers
           1, // invalid, non-multiple of 4 offset
-          kBufferData.byteLength - Uint32Array.BYTES_PER_ELEMENT, // invalid, last element outside buffer
+          // invalid, last element outside buffer
+          kBufferData.byteLength - Uint32Array.BYTES_PER_ELEMENT,
         ])
       )
   )
--- a/src/webgpu/api/validation/encoding/cmds/compute_pass.spec.ts
+++ b/src/webgpu/api/validation/encoding/cmds/compute_pass.spec.ts
@@ -1,21 +1,5 @@
 export const description = `
 API validation test for compute pass
-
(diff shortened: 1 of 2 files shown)


src/webgpu/api/validation/encoding/cmds/compute_pass.spec.ts
```diff
@@ -0,0 +1,135 @@
+export const description = `
+API validation test for compute pass
+
+Test Coverage:
+  - Tests using compute pipelines
```
Kangz, line 5: This can now be split up into the various tests using `g.test().desc()`!

src/webgpu/api/validation/encoding/cmds/compute_pass.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+
+import { ValidationTest } from './../../validation_test.js';
+
+class F extends ValidationTest {
+  createComputePipeline(state: 'valid' | 'invalid'): GPUComputePipeline {
+    if (state === 'valid') {
+      return this.createNoOpComputePipeline();
+    }
+
+    this.device.pushErrorScope('validation');
```
Kangz, line 32: Maybe we should use `ValidationTest.createErrorComputePipeline` or add it if it  doesn't exist.

src/webgpu/api/validation/encoding/cmds/compute_pass.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+
+g.test('dispatch_sizes')
+  .params(
+    params()
+      .combine(poptions('dispatchType', ['direct', 'indirect'] as const))
+      .combine(
+        poptions('workSizes', [
+          [0, 0, 0],
+          [1, 1, 1],
+          // TODO: Add a test for a large workSize when the upper limit has been decided.
```
Kangz, line 93: nit: and TODO sizes right under the limit.

src/webgpu/api/validation/encoding/cmds/compute_pass.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+  )
+  .fn(t => {
+    const pipeline = t.createNoOpComputePipeline();
+    const [x, y, z] = t.params.workSizes;
+    const { encoder, finish } = t.createEncoder('compute pass');
+    encoder.setPipeline(pipeline);
+    if (t.params.dispatchType === 'direct') {
+      encoder.dispatch(x, y, z);
+    } else if (t.params.dispatchType === 'indirect') {
+      encoder.dispatchIndirect(t.createIndirectBuffer('valid', new Uint32Array([x, y, z])), 0);
```
Kangz, line 105: note: I think a valid thing to do for invalid `dispatchIndirect` is to just drop them on the floor, so we'll have to figure out how to measure that when we add invalid cases.

src/webgpu/api/validation/encoding/cmds/compute_pass.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+    const { state, offset } = t.params;
+    const pipeline = t.createNoOpComputePipeline();
+    const buffer = t.createIndirectBuffer(state, kBufferData);
+    const { encoder, finish } = t.createEncoder('compute pass');
+    encoder.setPipeline(pipeline);
+    t.expectValidationError(() => {
+      encoder.dispatchIndirect(buffer, offset);
+      t.queue.submit([finish()]);
+    }, state !== 'valid' || offset % 4 !== 0 || offset + 3 * Uint32Array.BYTES_PER_ELEMENT > kBufferData.byteLength);
+  });
```
Kangz, line 135: @kainino0x do we need to test that `dispatch` requires a valid pipelines with the correct bingroups or will it be in a different test?

Kangz (approved):

Kangz (approved):

src/webgpu/api/validation/encoding/cmds/compute_pass.spec.ts
kainino0x: pipeline-bindgroup matching tests are going to be pretty involved and probably can be shared between compute and render, so probably a different file.

src/webgpu/api/validation/encoding/cmds/compute_pass.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+    const { encoder, finish } = t.createEncoder('compute pass');
+    encoder.setPipeline(pipeline);
+    t.expectValidationError(() => {
+      finish();
+    }, t.params.state === 'invalid');
+  });
+
+g.test('dispatch_sizes')
+  .desc(
+    `- For both 'direct' and 'indirect' dispatch
```
kainino0x, line 64: nit: Put a newline after the ` to make this readable in the source
```suggestion
    `
- For both 'direct' and 'indirect' dispatch
```

src/webgpu/api/validation/encoding/cmds/compute_pass.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+  .desc(
+    `- For 'indirect' dispatch
+  - Tests that indirect buffers:
+    - 'invalid', 'destroyed' generate an error
+    - 'valid' buffers do not generate an error
+  - Tests that, for a buffer with 6 elements, indirect offsets:
+    - 0, 'sizeof(uint32)', and '3 * sizeof(uint32)' do not generate an error
+    - An error should be generate by the following:
+      - 1: non-multiple of 4
+      - '4 * sizeof(uint32): z-components outside of buffer`
```
kainino0x, line 103: Here's this plan formatted the way I had in mind - note it's a little lighter on case details since they're right below. WDYT? (I'm hoping to make all of the plans more uniform in the future.)

```suggestion
    `
Test dispatchIndirect validation by submitting various dispatches with a no-op
pipeline and an indirectBuffer with 6 elements.
- indirectBuffer: {'valid', 'invalid', 'destroyed'}
- indirectOffset:
  - valid, within the buffer: {beginning, middle, end} of the buffer
  - invalid, non-multiple of 4
  - invalid, the last element is outside the buffer
```

kainino0x (approved):
LGTM with a few comments

src/webgpu/api/validation/encoding/cmds/compute_pass.spec.ts
egalli: Looks good to me. Looks like you added a line break at 80 col. Was that intentional? If so, should we enforce 80 cols with ESLint?

BTW, I used the demo as an example for the "style". https://github.com/gpuweb/cts/blob/5ab9cd41d491881e3a40575a3841c27fc0ea4a8a/src/demo/a/b/c.spec.ts#L10-L14
We should probably make that the demo match the correct styling.

kainino0x: > Looks good to me. Looks like you added a line break at 80 col. Was that intentional? If so, should we enforce 80 cols with ESLint?

I added it in an arbitrary place (can't see the columns since I was editing in GitHub), but we do enforce a column limit (I think it's 100 cols). The problem is this is inside a string so idk if ESLint can wrap it for us (I guess it would insert `\\\n`).

> We should probably make that the demo match the correct styling.

Oh, very excellent point. I didn't think very deeply about the formatting when I wrote that (was just trying to make sure the text showed up at all on /standalone/). Would you mind making that tweak, and make sure there's an example in `examples.spec.ts` too? (can be the same)

egalli: I just noticed `printWidth: 100` on prettier.config.js. It doesn't seem to work on some lines, 
https://github.com/gpuweb/cts/blob/c66e2bd6f5f60b45c5acdca197110910f4e13f33/src/webgpu/api/validation/encoding/cmds/compute_pass.spec.ts#L132
It seems like prettier is forcing that statement to be in a single line. Adding `max-len` to eslint just makes it error out with no way to fix it.

Also, I'll update the examples.

kainino0x: That's weird, I don't think I have run into cases where prettier fails like that. Thanks for looking into this.

kainino0x (approved):

PR #309, opened 2020-10-19
Loop to exhaust eventualExpectations

kainino0x:
If the async code inside an eventualExpectation adds another
eventualExpectation, this makes it so it will be awaited at the end of
the test.

~Draft because I haven't tested this code yet.~


kainino0x:
PTAL

austinEng (approved):
Looks good; possible to add a unit test for the behavior?

kainino0x:
Good call, done.

PR #312, opened 2020-10-20
[DRAFT] Change build to generate AMD module definitions instead of ES6

kainino0x:
A possible step toward doing the transpilation at runtime in the standalone and wpt runtimes.

Issue: #314


kainino0x:
Not ready to land until/unless we manage to get it working with runtime-Babel.
FYI I am not going to try to make that work right now, but we would have to figure out how to hook RequireJS (or find an alternate implementation that we can hook) in the right way.

kainino0x:
Seems like if we ever do this in WPT we can do it with a service worker. In any case, closing in favor of #386 for now.

PR #313, opened 2020-10-20
Remove unneeded @babel/plugin-proposal-class-properties

kainino0x:
This changes the generated code to use class properties instead of
polyfilling them. This cleans up the generated code for a lot of files.

According to MDN, these are supported starting from Chrome 72,
Firefox 69, and Safari 14.


austinEng (approved):
great!

PR #315, opened 2020-10-20
Fix exception type for getMappedRange in map_oom

Kangz:
PTAL!

### DIFF
--- a/src/webgpu/api/operation/buffers/map_oom.spec.ts
+++ b/src/webgpu/api/operation/buffers/map_oom.spec.ts
@@ -68,7 +68,7 @@ g.test('mappedAtCreation')
     const f = () => buffer.getMappedRange(0, size);
 
     if (oom) {
-      t.shouldThrow('RangeError', f);
+      t.shouldThrow('OperationError', f);
     } else {
       f();
     }


kainino0x (approved):
I think this looks good but I think @austinEng should review.

austinEng:
Based on https://github.com/gpuweb/gpuweb/issues/872, I think it should probably stay `RangeError` to match how `new ArrayBuffer` fails.

kainino0x:
Right, that makes sense. MAX_ALIGNED_SAFE_INTEGER is way too large for anything so it makes sense it would match. I'm not sure if we figured out what to do on real cpu-OOM but we can't easily test that here.

austinEng:
I think we said the buffer is an error, you get an async OOM, and we make dummy data so that getMappedRange works (unless it is RangeError)

kainino0x:
Actually, in this test case the buffer itself is invalid (due to OOM), so probably the algorithm should early-out before it "tries" to create an ArrayBuffer? (It's not well defined in the spec yet.) The same thing would happen if the buffer was smaller than the getMappedRange range.

kainino0x:
[race condition above]

Oh yes, I forgot about the buffer OOM being async and having to provide dummy memory for mappedAtCreation. Yeah. RangeError still makes sense to me then.

Kangz:
Gotcha, closing this PR.

PR #316, opened 2020-10-20
Update WGSL to new struct syntax.

dj2:
This CL updates the WGSL CTS tests to use the `struct IDENT` form
instead of the original `type IDENT = struct` form.


kainino0x (approved):
Thanks!

PR #318, opened 2020-10-23
Remove non-test dependence on WebGPU ambient constants

### LINKED ISSUE #317: Make /standalone/ run without WebGPU support
Currently /standalone/ will crash if the WebGPU constants (`GPUBufferUsage.*`, `GPUShaderStage.*`, etc.) aren't defined. Now that the test plan is becoming embedded in the CTS, it is annoying that you can't load it in a release browser. Shouldn't be too hard to fix, it's basically a revert of a change I made a while ago.

kainino0x:
Fixes #317


austinEng (approved):
lgtm

PR #319, opened 2020-10-23
Allow marking tests as explicitly unimplemented

kainino0x:
This allows for test structure to be committed into the CTS before the
tests themselves are implemented.

Also hide SKIP error messages when not in "debugging" mode.

### DIFF
--- a/src/common/framework/test_group.ts
+++ b/src/common/framework/test_group.ts
@@ -1,4 +1,4 @@
-import { Fixture } from './fixture.js';
+import { Fixture, SkipTestCase } from './fixture.js';
 import { TestCaseRecorder } from './logging/test_case_recorder.js';
 import { CaseParams, CaseParamsIterable, extractPublicParams } from './params_utils.js';
 import { kPathSeparator } from './query/separators.js';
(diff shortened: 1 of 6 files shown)


src/common/framework/test_group.ts
```diff
@@ -120,10 +123,27 @@ class TestBuilder<F extends Fixture, P extends {}> {
...
+      throw new SkipTestCase('test unimplemented');
+    };
+  }
+
+  validate(): void {
+    const testPathString = this.testPath.join(kPathSeparator);
+    assert(this.testFn !== undefined, () => {
+      let s = `Test is missing .fn(): ${testPathString}`;
+      if (this.testConstructorStack.stack) {
+        s += `\n-> test created at:\n${this.testConstructorStack.stack}`;
```
austinEng, line 142: at least in Chrome, i think this will say: 
> -> test created at:
> Error
> ___ at ...
> ___ at ...

I remember you may have some utilities elsewhere that separate the "Error" message and the stack ?

austinEng (approved):
sorry missed this one

src/common/framework/test_group.ts
kainino0x: yes, in chrome it will. I think I used to have utilities to do that but they just stripped the first line from the message, and that doesn't actually work (message could have a newline in it, some engines don't put a line at the top of .stack, etc.)

PR #320, opened 2020-10-23
Create sampler validation test

natashalee:
Same cases listed on cts test plan MD, just added NAN should fail for lodMinClamp and lodMaxClamp.


src/webgpu/api/validation/createSampler.spec.ts
```diff
@@ -0,0 +1,19 @@
+export const description = `
+createSampler validation tests.
+
+Test Plan: 
+
+* lodMinClamp:
```
kainino0x, line 6: I just added a new `.unimplemented()` so you can do e.g. the following:

```ts
g.test('lodMinClamp')
.desc('Test various values for lodMinClamp')
.unimplemented();
```

Could you update this to that format?

You can even pre-write the `.params()` if you want.

src/webgpu/api/validation/createSampler.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+  * < 0
```
kainino0x, line 7: Should test several valid values, too, and large values. So, e.g. `-1`, `-Number.MIN_VALUE`, `0`, `0.5`, `1`, `10`, `Number.MAX_VALUE`. (Not 100% sure the CTS will be happy with all these values, let me know if there are problems.)

src/webgpu/api/validation/createSampler.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+  * = NAN
```
kainino0x, line 8: We actually don't need to test NaN (or infinities)here because it's [invalid](https://heycam.github.io/webidl/#es-float) according to WebIDL. It would end up going in the idl/ tests later on.

src/webgpu/api/validation/createSampler.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+
+* lodMaxClamp:
+  * > MAX_SAFE_INTEGER
+  * = NAN
+  * =,> FLT_MAX
+  * < lodMinClamp
```
kainino0x, line 14: Will need to test this for several values of lodMinClamp. I would recommend combining these two tests into one, where the cases are combinations of lodMinClamp and lodMaxClamp, e.g.

```ts
.params([
  { lodMinClamp: -1 },
  { lodMinClamp: 0 },
  /*...*/
  { lodMaxClamp: Number.MAX_VALUE },
  { lodMaxClamp: 0 },
  /*...*/
  { lodMinClamp: 2, lodMaxClamp: 2 },
  { lodMinClamp: 2.01, lodMaxClamp: 2 },
])
```

src/webgpu/api/validation/createSampler.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+
+* maxAnisotropy
+  * = 0
+  * = 65535
```
kainino0x, line 18: note: we don't need to add extreme value tests here, because I think the upstream spec will change to make those IDL errors (by adding EnforceRange). Filed https://github.com/gpuweb/gpuweb/issues/1187

src/webgpu/api/validation/createSampler.spec.ts
natashalee: It shouldn't be passing the test for -Number.MIN_VALUE right? could be int overflow, but createsampler is passing on  
-Number.MIN_VALUE right now

kainino0x: oops, in the spec lodMinClamp is `float` not `double` so both Number.MIN_VALUE and Number.MAX_VALUE are non-representable values. I think MIN_VALUE will round to 0 and MAX_VALUE will become Infinity or NaN which is an IDL error?

kainino0x: You can just come up with some smallish and largeish numbers instead like `1e-30` and `1e30`

PR #321, opened 2020-10-30
standalone: Lazily-generate DOM subtrees

### LINKED ISSUE #304: Lazy-load test cases on /standalone/ page
The big nested DOM tree on the /standalone/ runner page needs to lazily load subtrees instead of just loading everything at once. The only level that really needs to be lazily loaded is the "case" subtrees; it should be okay to greedily load everything up to the "test" (g.test()) because every "test" is handwritten, while the list of cases is generated by the ParamsBuilder so it has huge explosions of cases. In other words, the subtrees underneath DOM nodes corresponding to TestQueryLevel 1 and 2 should be greedily loaded, but 3 and 4 should be lazily loaded.

The slightly tricky part is in what it means to "lazily load" those subtrees. The way the actual test loading works there's currently no way to lazily load the underlying data (the TestTree object). However I'm pretty sure that's not very slow right now, so hopefully we can keep that for now, and all that actually needs to be lazy is the creation of the DOM elements which occurs in makeSubtreeHTML (makeSubtreeChildrenHTML).

kainino0x:
Significantly improves the performance of loading /standalone/. Code is still messy but at least it still seems to work :)

Before: ~110s
After: ~2s

Note that due to the way the code is currently written, actually running tests causes the DOM for those test cases to get generated, which means running the whole CTS in standalone is difficult (it was better before because once the page loaded, tests ran quickly).

Can be reviewed in separate commits.

Partially fixes #304

### DIFF
--- a/src/common/runtime/standalone.ts
+++ b/src/common/runtime/standalone.ts
@@ -33,17 +33,19 @@ type RunSubtree = () => Promise<void>;
 
 function makeTreeNodeHTML(
   tree: TestTreeNode,
+  parentElement: HTMLElement,
   parentLevel: TestQueryLevel
-): [HTMLElement, RunSubtree] {
+): RunSubtree {
+  const div = $('<div>').appendTo(parentElement)[0];
   if ('children' in tree) {
-    return makeSubtreeHTML(tree, parentLevel);
+    return makeSubtreeHTML(tree, div, parentLevel);
   } else {
-    return makeCaseHTML(tree);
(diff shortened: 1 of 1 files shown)


kainino0x:
marked as draft; Fixing a bug with the lastQueryLevelToExpand

kainino0x:
Fixed.

egalli (approved):
LGTM

kainino0x:
thanks!

PR #322, opened 2020-11-02
standalone: Replace results JSON textarea with copy button

kainino0x:
In Chromium, this textarea causes a massive relayout whenever it gets
updated (e.g. 40s on my machine). Remove it, and replace it with a button
that copies the JSON to the clipboard (fast).

### DIFF
--- a/standalone/index.html
+++ b/standalone/index.html
@@ -212,9 +212,7 @@
   </head>
   <body>
     <h1>WebGPU Conformance Test Suite</h1>
-    <p>
-      <input type="checkbox" id="expandall" /><label for="expandall">Expand All (slow!)</label>
-    </p>
+    <input type="checkbox" id="expandall" /><label for="expandall">Expand All</label>
 
     <div id="info"></div>
     <div id="resultsVis"></div>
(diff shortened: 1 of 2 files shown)


kainino0x:
@egalli PTAL!

standalone/index.html
```diff
@@ -216,7 +216,10 @@ <h1>WebGPU Conformance Test Suite</h1>
 
     <div id="info"></div>
     <div id="resultsVis"></div>
-    <textarea id="resultsJSON" spellcheck="false"></textarea>
+
+    <p>
+      <input type=button id=copyResultsJSON value="Copy results as JSON">
```
austinEng, line 221: nit: quotes on the html attributes?

austinEng (approved):

standalone/index.html
kainino0x: Elsewhere in this project I have been not using quotes (src/common/templates/cts.html) but sg, let's just keep this one internally consistent for now.

egalli (approved):
LGTM

PR #323, opened 2020-11-02
standalone: Don't generate DOM when test finishes

### LINKED ISSUE #304: Lazy-load test cases on /standalone/ page

kainino0x:
Previously the test status and logs would get inserted into the DOM immediately when run. This just stores off the result so that it can be rendered into the DOM on-demand.

Also fixes "expand all" button. Previously this just searched the entire page for checkboxes and checked all of them. With lazy DOM generation, this didn't work anymore. Now returns a function from GenerateSubtreeHTML which sets the subtree as "checked", recursively.

Fixes #304

### DIFF
--- a/src/common/runtime/standalone.ts
+++ b/src/common/runtime/standalone.ts
@@ -2,6 +2,7 @@
 
 import { DefaultTestFileLoader } from '../framework/file_loader.js';
 import { Logger } from '../framework/logging/logger.js';
+import { LiveTestCaseResult } from '../framework/logging/result.js';
 import { parseQuery } from '../framework/query/parseQuery.js';
 import { TestQueryLevel } from '../framework/query/query.js';
 import { TestTreeNode, TestSubtree, TestTreeLeaf } from '../framework/tree.js';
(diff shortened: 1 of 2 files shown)


kainino0x:
@egalli PTAL. Sorry this PR is in a big chunk; had to twist around some more of the structure to make this work. A super detailed review isn't needed, but please look for any glaring issues (including suggestions on making the code clearer).

egalli (approved):
LGTM

PR #324, opened 2020-11-03
standalone: Show only the full query string for a test, and aligned

kainino0x:
Remove the "relative" paths in the test tree and promote the full query
string. With the "relative" paths I find that it's way too hard to see
"where" you are in the tree when browsing.

Aligns all of the full query strings with one another (unless the screen
is too small), so the tree structure is easy to see.


kainino0x:
image

austinEng (approved):

PR #327, opened 2020-11-05
add create sampler validation test

natashalee:
My bad, I made a mistake with git trying to avoid rebasing and using merge (I've gotten too used to that workflow), hopefully I fixed it (again).

### DIFF
--- a/src/webgpu/api/validation/createSampler.spec.ts
+++ b/src/webgpu/api/validation/createSampler.spec.ts
@@ -17,27 +17,10 @@ g.test('lodMinAndMaxClamp')
       .combine(poptions('lodMaxClamp', [-4e-30, -1, 0, 0.5, 1, 10, 4e30]))
   )
   .fn(async t => {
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/createSampler.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+          lodMinClamp: t.params.lodMinClamp,
+          lodMaxClamp: t.params.lodMaxClamp,
+        });
+      });
+    } else {
+      t.device.createSampler({
+        lodMinClamp: t.params.lodMinClamp,
+        lodMaxClamp: t.params.lodMaxClamp,
+      });
+    }
```
kainino0x, line 42: expectValidationError can take a boolean for whether to expect a validation error or not. Can you compute that boolean using similar logic, and then just have one createSampler call?

kainino0x (approved):
Thanks!

PR #328, opened 2020-11-05
wpt: Don't return promise from async_test

kainino0x:
async_test now asserts that its test function doesn't return anything,
to detect whether you actually intended to use promise_test.
We don't need to use promise_test, so we just don't return the promise.


kainino0x:
Landing TBR because this is the same change previously made and reviewed here:
https://chromium-review.googlesource.com/c/chromium/src/+/2436654

PR #329, opened 2020-11-06
Roll webgpu-types.

Kangz:
This required adding some texture formats, including rgb9e5ufloat that
doesn't have a correct formatInfo for now.

PTAL! I'll let the rgb9e5 as a TODO because I have no idea what's going on in this part of the code.

### DIFF
--- a/src/webgpu/util/texture/texelData.ts
+++ b/src/webgpu/util/texture/texelData.ts
@@ -112,6 +112,13 @@ const float32 = {
   bitLength: 32,
 };
 
+const componentUnimplemented = {
+  write: () => {
+    unreachable('TexelComponentInfo not implemented for this texture format');
+  },
+  bitLength: 0,
+};
+
 const repeatComponents = (
   componentOrder: TexelComponent[],
   perComponentInfo: SingleComponentInfo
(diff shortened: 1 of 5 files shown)


austinEng (approved):
yea... we'll have to customize serialization for rgb9e5.. :(

austinEng:
> yea... we'll have to customize serialization for rgb9e5.. :(

Don't think you actually put a TODO there though - can you do that / file an issue ?

Kangz:
I got confused with VSCode and it didn't save my buffer >_>

src/webgpu/util/texture/texelData.ts
```diff
@@ -173,7 +173,9 @@ const kRepresentationInfo: {
   'rgba32float':            { ...repeatComponents(kRGBA, float32), sRGB: false },
 
   'rgb10a2unorm':           { componentOrder: kRGBA, componentInfo: { R: unorm10, G: unorm10, B: unorm10, A: unorm2 }, sRGB: false },
-  'rg11b10float':           { componentOrder: kRGB, componentInfo: { R: float11, G: float11, B: float10 }, sRGB: false },
+  'rg11b10ufloat':          { componentOrder: kRGB, componentInfo: { R: float11, G: float11, B: float10 }, sRGB: false },
+  // TODO the e5 is shared between all components, figure out how to write it.
+  'rgb9e5ufloat':           { componentOrder: kRGB, componentInfo: { R: float11, G: float11, B: float10 }, sRGB: false },
```
kainino0x, line 178: I'm going to try to change this to throw exceptions if you try to use this, instead of putting the wrong data.

kainino0x (approved):
Not sure why the lockfileVersion got incremented and the file got rewritten, but it might present compatibility issues with older npm. I'll roll it back and do the minimal update (which is what my `npm install` does).

PR #330, opened 2020-11-06
[wgsl] Update tests to match spec update: Relaxing void return requirement

sarahM0:
Update wgsl validation tests to match this spec update:
https://github.com/gpuweb/gpuweb/pull/1179

### DIFF
--- a/src/webgpu/shader/validation/wgsl/function-return-missing-return-empty-body.fail.wgsl
+++ b/src/webgpu/shader/validation/wgsl/function-return-missing-return-empty-body.fail.wgsl
@@ -2,10 +2,8 @@
 # function.
 
 fn func() -> i32 {
-  var a : i32 = 0;
 }
 
 [[stage(vertex)]]
 fn main() -> void {
-  func();
 }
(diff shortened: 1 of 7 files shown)


src/webgpu/shader/validation/wgsl/function-return-missing-return-empty-body.fail.wgsl
```diff
@@ -0,0 +1,9 @@
+# v-0002 - This program fails due to the missing return at the end of the
+# function.
```
kainino0x, line 2: This test case seems almost identical to the one right below it.

kainino0x (commented):
LGTM but I'll let Dan approve

src/webgpu/shader/validation/wgsl/function-return-missing-return-empty-body.fail.wgsl
sarahM0: In this one main() is empty and func() is not called anywhere.

kainino0x:
Oops, for some reason I thought Dan was already on the PR. @dj2 PTAL

src/webgpu/shader/validation/wgsl/duplicate-shader-entry-point.fail.wgsl
```diff
@@ -0,0 +1,12 @@
+# v-0020 - Duplicate entry point
```
dj2, line 1: This should be allowed I think

dj2 (approved):

PR #331, opened 2020-11-09
Add CI on Actions

mehmetoguzderin:
Changes in this pull request enable GitHub Actions to act for CI on the push and pull request (target) events. Instead of having less conditional steps, there are two actions: better readability and taking advantage of pull request target events better.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -1,0 +1,61 @@
+name: Pull Request CI
+
+on:
+  pull_request_target:
+    branches: [main]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2.3.1
+        with:
+          persist-credentials: false
+      - run: |
+          git fetch
+          git checkout ${{ github.event.pull_request.head.sha }}
+      - uses: actions/setup-node@v2-beta
+        with:
+          node-version: "15.x"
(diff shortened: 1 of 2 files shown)


kainino0x (approved):
I didn't realize you were going to create PR previews for this repo as well; all we had before was publish-on-merge. Very cool, thanks! It will be useful, especially because I am moving test plan development into this repo, which makes the "standalone" runner also serve as a document viewer for the test plan.

kainino0x:
Looks like I will have to figure out *a little bit* of this GitHub Actions stuff after all. Can you help me figure out how to get this running, and in particular figure out how to replace the old `continuous-integration/travis-ci/pr` check with the Action? (Is another Action other than the ones in this PR needed to enable that?)

kainino0x:
Also seems to have encountered an error on https://github.com/gpuweb/cts/pull/333/checks?check_run_id=1377808887

mehmetoguzderin:
@kainino0x One sad fact about `pull_request_target`: although it is a clean trigger for building PRs (and it should replace the old `continuous-integration/travis-ci/pr`, I built by taking it as reference [assuming replacing it means having the Actions as a check on PRs]), it will only propagate to new PRs (though it doesn't take the changes in the PR into consideration, making the execution still safe for secret key purposes). Sorry about the problematic execution BTW, explained in the fix PR.

PR #333, opened 2020-11-10
Simplify WPT runtime

kainino0x:
Tested manually on: https://chromium-review.googlesource.com/c/chromium/src/+/2528311


kainino0x:
@austinEng PTAL

src/common/runtime/wpt.ts
```diff
@@ -1,63 +1,59 @@
+// Implements the wpt-embedded test runner (see also: wpt/cts.html).
+
 import { DefaultTestFileLoader } from '../framework/file_loader.js';
 import { Logger } from '../framework/logging/logger.js';
 import { parseQuery } from '../framework/query/parseQuery.js';
-import { TestTreeLeaf } from '../framework/tree.js';
-import { AsyncMutex } from '../framework/util/async_mutex.js';
```
austinEng, line 5: Is this the only usage of `AsyncMutex`? Can we remove it?

austinEng (approved):
Nice :)

src/common/runtime/wpt.ts
kainino0x: I think it is, and I was thinking about removing it but forgot. May as well do it now.

PR #334, opened 2020-11-10
Fix Actions on CI

mehmetoguzderin:
Sorry about not testing this on an org- I directly did the development on my fork which was not a great strategy. But this arrangement should end the checkout problem the last run had.


kainino0x (approved):
Thanks!

PR #338, opened 2020-11-11
Add tests for indexing, index format and primitive restart

haoxli:
There are some fail tests in primitive restart becuase the primitive restart value not only works for strip primitive topologies, but also for list primitive topologies ( 'line-list' and 'triangle-list')
without primitive restart:
line-list-wo-restart triangle-list-wo-restart
with primitive restart (expect same with above):
line-list triangle-list

For point-list tests, the problem is the render position not correct. There is only the bottom right point in the render target when using the vertices:
-1.0,  1.0, 0.0, 1.0, 
 1.0, -1.0, 0.0, 1.0,
 1.0,   1.0, 0.0, 1.0, 
-1.0, -1.0, 0.0, 1.0 
point-list

When we use ±0.99 instead of ±1.0 in the vertex array, we can see all 4 points: 
point-list-2


haoxli:
@kainino0x @Kangz PTAL! Thanks.

Kangz:
On which backends do the test fail? At least on D3D12 and Vulkan it should be possible to disable the primitive restart for list topologies. (Maybe [not on Metal apparently](https://developer.apple.com/documentation/metal/mtlrendercommandencoder/1515520-drawindexedprimitives?language=objc), that's quite terrible.)

Kangz:
I'll let @kainino0x take a first look first for CTS-isms.

haoxli:
This issue is reproduced on D3D12 backend.

Kangz:
Then it's likely https://bugs.chromium.org/p/dawn/issues/detail?id=575

kainino0x:
Oops, I just realized we already had some tests for this:
https://github.com/gpuweb/cts/blob/glsl-dependent/src/webgpu/api/operation/render_pipeline/primitive_topology.spec.ts

However I guess they didn't catch the Dawn bug that you found, so I'm really glad that we have more tests. Let's keep both.

src/webgpu/api/operation/vertex_state.spec.ts
```diff
@@ -0,0 +1,348 @@
...
+    }
+
+    const expectedTextureValues = t.CreateExpectedUint8Array(expectedShape);
+    t.expectContents(t.result, expectedTextureValues);
+  });
+
+// Test primitive restart with each primitive topology. The primitive restart value can used with
+// strip primitive topologies ('line-strip' or 'triangle-strip').
+// For line-strip, [0, 1, PRIM_RESTART, 2, 3] only render a X shape.
+// For triangle-strip, [0, 1, 3, PRIM_RESTART, 2] only the first triangle.
```
kainino0x, line 287: This description of these index arrays looks incomplete relative to the code. Can just delete this part of the comment.

src/webgpu/api/operation/vertex_state.spec.ts
```diff
@@ -0,0 +1,348 @@
...
+    }
+    return arrayBuffer;
+  }
+}
+
+export const g = makeTestGroup(IndexFormatTest);
+
+// Test indexing draw with index format of uint16. If this is interpreted as uint32, it will have
+// index 1 and 2 be both 0 and render nothing. And the index buffer size - offset must be not less
+// than the size required by triangle list, otherwise it also render nothing.
```
kainino0x, line 234: This description should be inside `g.test(...).desc( HERE )` instead of as a comment. Then it will show up in /standalone/ as part of the test description.

Same below.

src/webgpu/api/operation/vertex_state.spec.ts
```diff
@@ -0,0 +1,348 @@
...
+      } else if (primitiveTopology === 'triangle-strip') {
+        indexArray = new Uint16Array([0, 1, 3, 0xffff, 2, 2]);
+      } else {
+        indexArray = new Uint16Array([0, 1, 0xffff, 2, 3, 3]);
+      }
+    } else {
+      if (primitiveTopology === 'triangle-list') {
+        indexArray = new Uint32Array([0, 1, 3, 0xffffffff, 2, 1, 0]);
+      } else if (primitiveTopology === 'triangle-strip') {
+        indexArray = new Uint32Array([0, 1, 3, 0xffffffff, 2]);
```
kainino0x, line 319: If there were a bug where 'triangle-strip' were translated as 'triangle-list', I think this would still pass, because it would only be one triangle (0,1,3 followed by the incomplete triangle 0xffffffff,2).

src/webgpu/api/operation/vertex_state.spec.ts
```diff
@@ -0,0 +1,348 @@
...
+        ])
+      )
+  )
+  .fn(t => {
+    const { indexFormat, primitiveTopology } = t.params;
+
+    let indexArray: Uint16Array | Uint32Array;
+    if (indexFormat === 'uint16') {
+      if (primitiveTopology === 'triangle-list') {
+        indexArray = new Uint16Array([0, 1, 3, 0xffff, 2, 1, 0, 0]);
```
kainino0x, line 309: If primitive restart can indeed be disabled for lists, 0xffff will go out of bounds of the vertex buffer. Maybe we should make the vertex buffer large enough that they wouldn't. Do we have a strict guarantee on what happens when you go out of bounds of the vertex buffer? I think we allow a range of behaviors, which we shouldn't test here. Of course we can't do this for uint32 but at least we can do it for uint16.

kainino0x, line 309: Comments on each of these explaining the expected result would be helpful, e.g. simply
`// triangles (0, 1, 3), (0xffff, 2, 1)`
(or
`// triangles (0, 1, 3), (2, 1, 0)`
depending on spec for primitive restart for non-strip topos)

src/webgpu/api/operation/vertex_state.spec.ts
```diff
@@ -0,0 +1,348 @@
...
+    // The indexFormat must be set in render pipeline descriptor that specifys a strip primitive
+    // topology for primitive restart testing
+    if (primitiveTopology === 'line-strip' || primitiveTopology === 'triangle-strip') {
+      pipeline = this.MakeRenderPipeline(primitiveTopology, indexFormat);
+    } else {
+      pipeline = this.MakeRenderPipeline(primitiveTopology);
+    }
+
+    /* prettier-ignore */
+    const vertexArray = new Float32Array([
```
kainino0x, line 160: uNit:
```suggestion
    const vertexArray = new Float32Array(/* prettier-ignore */ [
```
so it applies only to the array `[...]` instead of the whole `const ...;`

src/webgpu/api/operation/vertex_state.spec.ts
```diff
@@ -0,0 +1,348 @@
...
+}
+
+class IndexFormatTest extends GPUTest {
+  private colorAttachment!: GPUTexture;
+  byteLength!: number;
+  rowsPerImage!: number;
+  bytesPerRow!: number;
+  result!: GPUBuffer;
+
+  async init(): Promise<void> {
```
kainino0x, line 55: Is it necessary to do this work in init() and store into class members, instead of just as a helper called in `run()`? I think it looks like only `result` is needed, which could just be returned from `run()`.

src/webgpu/api/operation/vertex_state.spec.ts
```diff
@@ -0,0 +1,348 @@
+export const description = '';
```
kainino0x, line 1: We'll need tests that use the primitive restart values as real indices with non-strip topologies, to make sure those behave properly (and primitive restart is appropriately enabled/disabled) across backends. However let's leave these as a TODO for the moment and do them as a followup?

These tests will be important for https://github.com/gpuweb/gpuweb/issues/1220.

src/webgpu/api/operation/vertex_state.spec.ts
haoxli: I don't clear that what the 0xffff will be used for in non-strip topologies when primitive restart can indeed be disabled for lists. If it's used as a big number in index buffer, I think we need the tests, currently it draws nothing with an index out of bounds of the vertex buffer, no error is thrown.  But if it's checked as an invalid index, and skipped, our tests will cover it. I'm not sure what kind of behavior it will be.

src/webgpu/api/operation/vertex_state.spec.ts
haoxli: I think even without primitive restart, it's also pass for such bug. But here，I should draw another bottom right triangle to make the primitives after the restart values also can be renderred.

src/webgpu/api/operation/vertex_state.spec.ts
haoxli: > Also note that if ANY index goes out of bounds, I'm pretty sure we allow discarding the entire draw call (not just that primitive).

So for the lists, our expected result should be nothing is renderred, right?

kainino0x: I think there are multiple possible expected results: the entire draw could be discarded, just the invalid triangle could be discarded, or the invalid triangle could be rendered by clamping the index (0xFFFF) to the max valid index value (3). This is why it would be a good idea to make the vertex buffer large enough that 0xFFFF does not go out of bounds. We can test out of bounds behavior, including for 0xFFFF, in another test.

src/webgpu/api/operation/vertex_state.spec.ts
```diff
@@ -0,0 +1,366 @@
...
+    const { indexFormat, primitiveTopology, _expectedShape } = t.params;
+
+    let indexArray: Uint16Array | Uint32Array;
+    // The primitive restart value can used with strip primitive topologies ('line-strip' or 'triangle-strip').
+    // For lists, they are expected to be renderred the same as without primitive restart.
+    if (indexFormat === 'uint16') {
+      if (primitiveTopology === 'triangle-list') {
+        // triangle-list: (0, 1, 3), (0xffff, 2, 1)
+        // triangle-list with restart: (0, 1, 3), (2, 1, 0)
+        // triangle-strip: (0, 1, 3), (2, 1, 0)
```
kainino0x, line 331: nit: Mark up which one is expected, like:
```suggestion
        // -> triangle-list: (0, 1, 3), (0xffff, 2, 1)
        //    triangle-list with restart: (0, 1, 3), (2, 1, 0)
        //    triangle-strip: (0, 1, 3), (2, 1, 0)
```
and same below.

kainino0x, line 331: I realized we should also show what it looks like for "triangle-strip without restart", since it's possible a backend could enable triangle strip without enabling restart, I think.

src/webgpu/api/operation/vertex_state.spec.ts
```diff
@@ -0,0 +1,366 @@
...
+        indexArray = new Uint16Array([0, 1, 3, 0xffff, 2, 1, 0, 0]);
+      } else if (primitiveTopology === 'triangle-strip') {
+        // triangles: (0, 1, 3), (2, 3, 1), (3, 1, 1)
```
kainino0x, line 334: Enumerate what the triangles would look like for the other three possible topologies as well.

src/webgpu/api/operation/vertex_state.spec.ts
```diff
@@ -0,0 +1,366 @@
...
+        indexArray = new Uint16Array([0, 1, 3, 0xffff, 2, 3, 1, 1]);
+      } else {
+        // point: (0), (1), (0xffff), (2), (3), (3)
+        // line-list: (0, 1), (0xffff, 2), (3, 3)
+        // line-list with restart: (0, 1), (2, 3)
+        // line-strip: (0, 1), (2, 3), (3, 3)
```
kainino0x, line 340: example:
```suggestion
        // -> point: (0), (1), (0xffff), (2), (3), (3)
        // -> line-list: (0, 1), (0xffff, 2), (3, 3)
        //    line-list with restart: (0, 1), (2, 3)
        // -> line-strip: (0, 1), (2, 3), (3, 3)
```

src/webgpu/api/operation/vertex_state.spec.ts
kainino0x: Another (smaller) reason to make sure 0xFFFF is a valid index in the vertex buffer is that, if an implementation did triangle-strip with primitive restart disabled, they would still pass if just invalid triangles were discarded.

src/webgpu/api/operation/vertex_state.spec.ts
```diff
@@ -0,0 +1,366 @@
...
+    params()
+      .combine(poptions('indexFormat', ['uint16', 'uint32'] as const))
+      .combine(kPrimitiveTopologiesForRestart)
+  )
+  .fn(t => {
+    const { indexFormat, primitiveTopology, _expectedShape } = t.params;
+
+    let indexArray: Uint16Array | Uint32Array;
+    // The primitive restart value can used with strip primitive topologies ('line-strip' or 'triangle-strip').
+    // For lists, they are expected to be renderred the same as without primitive restart.
```
kainino0x, line 326:
```suggestion
    // For lists, the primitive restart value isn't special and should be treated as a regular index value.
```

src/webgpu/api/operation/vertex_state.spec.ts
```diff
@@ -0,0 +1,366 @@
...
+    const indexArray = new Uint32Array([1, 2, 0, 0, 0, 0, 0, 1, 3, 0]);
+    const indexBuffer = t.CreateIndexBuffer(indexArray, 'uint32');
+    const result = t.run(indexBuffer, indexArray.length, 'uint32', indexOffset);
+
+    const expectedTextureValues = t.CreateExpectedUint8Array(_expectedShape);
+    t.expectContents(result, expectedTextureValues);
+  });
+
+g.test('primitive_restart')
+  .desc('Test primitive restart with each primitive topology.')
```
kainino0x, line 315: Add a TODO to try to test indexing into a vertex buffer that actually has `2**32` vertices.

src/webgpu/api/operation/vertex_state.spec.ts
```diff
@@ -171,8 +184,19 @@ class IndexFormatTest extends GPUTest {
...
       -1.0, -1.0, 0.0, 1.0]
     );
+
+    let bufferSize = vertexArray.byteLength;
+    // Create a large enough buffer for uint16 primitive restart value used as a big index.
+    if (
+      indexFormat === 'uint16' &&
+      primitiveTopology !== 'line-strip' &&
+      primitiveTopology !== 'triangle-strip'
+    ) {
```
kainino0x, line 194: You can just unconditionally allocate the buffer to have exactly `2**16` vertex elements regardless of the parameters. Since that's large-ish (1MiB) it would be good to explicitly destroy the buffer at the end of this function after the submit.

src/webgpu/api/operation/vertex_state.spec.ts
```diff
@@ -171,8 +184,19 @@ class IndexFormatTest extends GPUTest {
...
+      bufferSize = 0xffff * 16;
```
kainino0x, line 195: This is 0xffff vertex elements, but you need one more.

src/webgpu/api/operation/vertex_state.spec.ts
```diff
@@ -0,0 +1,366 @@
...
+  .desc('Test primitive restart with each primitive topology.')
+  .params(
+    params()
+      .combine(poptions('indexFormat', ['uint16', 'uint32'] as const))
+      .combine(kPrimitiveTopologiesForRestart)
+  )
+  .fn(t => {
+    const { indexFormat, primitiveTopology, _expectedShape } = t.params;
+
+    let indexArray: Uint16Array | Uint32Array;
```
kainino0x, line 324: Whoops, I thought I suggested this before, but can you make the construction of the array (`number[]`) not depend on the indexFormat, and then just use the indexFormat to decide whether to pack the result into Uint16Array or Uint32Array? It would halve the amount of lines here.

Note if you do this, you can just use `-1` as the restart value, as `new Uint16Array([-1])` gives 0xFFFF and `new Uint32Array([-1])` gives 0xFFFF_FFFF.

kainino0x (commented):
I will be OOO the rest of the week, but maybe @Kangz can also review and help get this landed (otherwise I'll get to it next week)

src/webgpu/api/operation/vertex_state.spec.ts
```diff
@@ -0,0 +1,366 @@
+export const description = `Test indexing, index format and primitive restart.
```
Kangz, line 1: Why is the test file called vertex_state.spec.ts and not index_format.spec.ts?

src/webgpu/api/operation/vertex_state.spec.ts
```diff
@@ -0,0 +1,366 @@
...
+    const vertexArray = new Float32Array(
+      /* prettier-ignore */ [
+      // float4 position
+      -1.0, 1.0, 0.0, 1.0,
+      1.0, -1.0, 0.0, 1.0,
+      1.0, 1.0, 0.0, 1.0,
+      -1.0, -1.0, 0.0, 1.0]
+    );
+
+    let bufferSize = vertexArray.byteLength;
```
Kangz, line 188: I think we could avoid the need for allocating large index buffers by hardcoding the vertex values in the shader. This would also allow testing 0xFFFF and 0xFFFFFFFF by hardcoding if (VertexIndex == 0xFFFFFFFF) {Position = ...} in the vertex shader.

We'd have 4 points that interest us, and the primitive restart index for non-strip topologies should get used as a regular index. WDYT?

src/webgpu/api/operation/vertex_state.spec.ts
Kangz: Why? I'm not sure I followed.

src/webgpu/api/operation/vertex_state.spec.ts
```diff
@@ -0,0 +1,366 @@
...
+    });
+    pass.setPipeline(pipeline);
+    pass.setVertexBuffer(0, vertexBuffer);
+    pass.setIndexBuffer(indexBuffer, indexFormat, indexOffset);
+    pass.drawIndexed(indexCount, 1, 0, 0, 0);
+    pass.endPass();
+    encoder.copyTextureToBuffer(
+      { texture: colorAttachment, origin: { x: 0, y: 0, z: 0 } },
+      { buffer: result, bytesPerRow, rowsPerImage },
+      { width: kWidth, height: kHeight, depth: 1 }
```
Kangz, line 224: nit: depth default to 1 so it can be removed.

src/webgpu/api/operation/vertex_state.spec.ts
```diff
@@ -0,0 +1,366 @@
...
+        },
+      ],
+    });
+    pass.setPipeline(pipeline);
+    pass.setVertexBuffer(0, vertexBuffer);
+    pass.setIndexBuffer(indexBuffer, indexFormat, indexOffset);
+    pass.drawIndexed(indexCount, 1, 0, 0, 0);
+    pass.endPass();
+    encoder.copyTextureToBuffer(
+      { texture: colorAttachment, origin: { x: 0, y: 0, z: 0 } },
```
Kangz, line 222: nit: origin default to {1, 1, 1} so it can just be {}

src/webgpu/api/operation/vertex_state.spec.ts
```diff
@@ -0,0 +1,366 @@
...
+    return result;
+  }
+
+  CreateExpectedUint8Array(renderShape: RenderShape): Uint8Array {
+    let texelValueBytes;
+    const arrayBuffer = new Uint8Array(byteLength);
+    for (let row = 0; row < kHeight; row++) {
+      for (let col = 0; col < kWidth; col++) {
+        if (
+          renderShape === RenderShape.Points &&
```
Kangz, line 237: Instead of this complicated logic, could be define an array of boolean for each shape and format them in a way that visually looks like the shape?

src/webgpu/api/operation/vertex_state.spec.ts
Kangz: See comment above for a way to do this without a vertex buffer.

src/webgpu/api/operation/vertex_state.spec.ts
kainino0x: The origin defaults to `{}` so it can be omitted entirely.

src/webgpu/api/operation/vertex_state.spec.ts
kainino0x: Great idea. Then this test will be more targeted, covering just index format/primitive restart, and not vertex fetch. Vertex fetch coverage can be separate (I forget how much we have already, if any).

src/webgpu/api/operation/vertex_state.spec.ts
kainino0x: Index 0xffff is the last element in an array of size 0x10000.

Obsolete with the suggestion about doing this without a vertex buffer, though.

src/webgpu/api/operation/vertex_state.spec.ts
kainino0x: We're not really following the spec chapters, although we probably really should be (good idea) because it's hard to figure out where tests should be right now. Unfortunately the organization of the spec will change though (current plan is https://github.com/gpuweb/gpuweb/wiki/Table-of-Contents but render pipeline section organization isn't set).

Let's rename this to `vertex_state/index_format.spec.ts`? We have a bunch of other tests in the test plan doc for vertex state and they shouldn't be in the same file.

kainino0x: See #392

src/webgpu/api/operation/vertex_state/index_format.spec.ts
```diff
@@ -0,0 +1,369 @@
...
+
+/**
+ * The expected rendering in raster grids of 4x4 pixels:
+ * Points:
+ *  ------------
+ *  |#        #|
+ *  |          |
+ *  |          |
+ *  |#        #|
+ *  ------------
```
kainino0x: I removed these comments because now that we have the literals, they're redundant. (They don't appear in the doc comment, unfortunately, but better to get rid of redundancy.)

kainino0x (commented):
Made a number of small changes and pushed them to your branch.

src/webgpu/api/operation/vertex_state/index_format.spec.ts
```diff
@@ -0,0 +1,313 @@
...
+          vec2<f32>(1.0, -1.0),
+          vec2<f32>(1.0, 1.0),
+          vec2<f32>(-1.0, -1.0));
+
+        [[builtin(position)]] var<out> Position : vec4<f32>;
+        [[builtin(vertex_idx)]] var<in> VertexIndex : u32;
+
+        [[stage(vertex)]]
+        fn main() -> void {
+          Position = vec4<f32>(pos[VertexIndex], 0.0 , 1.0);
```
kainino0x, line 114: Handle `VertexIndex == 0xFFFF || VertexIndex == 0xFFFFFFFF` explicitly here. OK to just do both because we don't use 0xFFFF in any of the uint32 index buffers. (Though maybe we should have a test for that too)

kainino0x (commented):
LGTM except for the outstanding thing about handling 0xffff/ffffffff in the shader (https://github.com/gpuweb/cts/pull/338#discussion_r529497371), and making sure the tests catch wrong behavior with those values.

src/webgpu/api/operation/vertex_state/index_format.spec.ts
haoxli: Ack.

src/webgpu/api/operation/vertex_state/index_format.spec.ts
```diff
@@ -0,0 +1,317 @@
...
+    // the primitive restart value isn't special and should be treated as a regular index value.
+    //
+    // The value -1 gets uploaded as 0xFFFF or 0xFFFF_FFFF according to the format.
+    //
+    // An arrow '->' indicates the expected result for that primitiveTopology.
+    if (primitiveTopology === 'triangle-list') {
+      // -> triangle-list:              (0, 1, 3), (-1, 2, 1)
+      //    triangle-list with restart: (0, 1, 3), (2, 1, 0)
+      //    triangle-strip:             (0, 1, 3), (2, 1, 0), (1, 0, 0)
+      //    triangle-strip w/o restart: (0, 1, 3), (1, 3, -1), (3, -1, 2), (-1, 2, 1), (2, 1, 0), (1, 0, 0)
```
kainino0x, line 295: In going through these to make sure the test were robust I realized I did need to draw the expected result for every incorrect topology. So I added it.

I found that some of the incorrect topologies are not distinguishable from the expected result, due to the position value (0,0,0,0) for index -1, which iiuc will never render anything.

src/webgpu/api/operation/vertex_state/index_format.spec.ts
```diff
@@ -0,0 +1,380 @@
...
+`;
+
+import { params, poptions } from '../../../../common/framework/params_builder.js';
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { GPUTest } from '../../../gpu_test.js';
+import { getTextureCopyLayout } from '../../../util/texture/layout.js';
+
+const kHeight = 4;
+const kWidth = 8;
+const kTextureFormat = 'r8uint' as const;
```
kainino0x, line 12: small change: I changed this to r8uint instead of bgra8unorm because it simplifies the code slightly.

src/webgpu/api/operation/vertex_state/index_format.spec.ts
```diff
@@ -0,0 +1,380 @@
...
+Test indexing, index format and primitive restart.
+`;
+
+import { params, poptions } from '../../../../common/framework/params_builder.js';
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { GPUTest } from '../../../gpu_test.js';
+import { getTextureCopyLayout } from '../../../util/texture/layout.js';
+
+const kHeight = 4;
+const kWidth = 8;
```
kainino0x, line 11: I expanded the 4x4 image by 4 more pixels on the left, and moved index -1 over there, so that it would render things.

Mostly this just meant adding space to the left of the expected results and then adding extra triangles on top of the old expected results.
I noticed that some of the triangles cut right along pixel centers, so hopefully rounding issues don't make this differ on different hardware. We'll want to update the test if we encounter that.

kainino0x:
@Kangz PTAL at my revisions since I can't review them :)

kainino0x:
(@haoxli also please review my changes)

Kangz (approved):
LGTM!

src/webgpu/api/operation/vertex_state/index_format.spec.ts
```diff
@@ -0,0 +1,317 @@
+export const description = `Test indexing, index format and primitive restart.
+
+TODO(hao.x.li@intel.com): Test that use the primitive restart values as real indices with
+non-strip topologies, to make sure those behave properly (and primitive restart is appropriately
+enabled/disabled) across backends. These tests will be important for gpuweb/gpuweb#1220.
```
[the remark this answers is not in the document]
haoxli: It's done.

PR #346, opened 2020-11-12
Merge 'main' in 'glsl-dependent'

Kangz:


kainino0x:
oops, this doesn't make a proper merge commit due to squash-and-merge. Usually I just push the merge directly to the glsl-dependent branch (after doing a local `grunt pre`). Guess we should update the roll instructions to indicate this

PR #347, opened 2020-11-12
Add gen_wpt_cts_html as an NPM script

Kangz:
PTAL! This is needed for the change to `regenerate.sh` in https://chromium-review.googlesource.com/c/chromium/src/+/2534730

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,8 @@
   "version": "0.1.0",
   "description": "",
   "scripts": {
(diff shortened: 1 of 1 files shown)


kainino0x (approved):

kainino0x:
@mehmetoguzderin fyi the build seems to have failed due to firebase rate limits. I don't know what exactly the limit is but we can chat about it sometime.

PR #352, opened 2020-11-13
Implement texelData for rgb9e5ufloat and add texelData unit tests

austinEng:
- Adds `dataType` and `componentType` to the texture format table for encodable texture formats. `dataType` is the format's data representation whereas `componentType` is the `GPUTextureComponentType` - the type you get in the shader.
 - Implements texelData helpers for packed format `rgb9e5ufloat`
 - Adds unit tests for upload texel data as `dataType`, loading the `componentType` in the shader, and writing it out.

### DIFF
--- a/src/webgpu/capability_info.ts
+++ b/src/webgpu/capability_info.ts
@@ -106,12 +106,12 @@ export type TextureDataType = 'uint' | 'sint' | 'unorm' | 'snorm' | 'float' | 'u
 
 export const kRegularTextureFormatInfo: {
   readonly [k in RegularTextureFormat]: {
-    color: true;
-    bytesPerBlock: number;
-    blockWidth: 1;
-    blockHeight: 1;
-    dataType: TextureDataType;
-    componentType: GPUTextureComponentType;
+    readonly color: true;
+    readonly bytesPerBlock: number;
+    readonly blockWidth: 1;
+    readonly blockHeight: 1;
+    readonly dataType: TextureDataType;
+    readonly componentType: GPUTextureComponentType;
   } & TextureFormatInfo;
 } = /* prettier-ignore */ {
   // 8-bit formats
@@ -136,7 +136,7 @@ export const kRegularTextureFormatInfo: {
   'rg16float':              { renderable:  true, color:  true, depth: false, stencil: false, storage: false, copySrc:  true, copyDst:  true, bytesPerBlock:  4, blockWidth: 1, blockHeight: 1, dataType:  'float', componentType: 'float' },
   'rgba8unorm':             { renderable:  true, color:  true, depth: false, stencil: false, storage:  true, copySrc:  true, copyDst:  true, bytesPerBlock:  4, blockWidth: 1, blockHeight: 1, dataType:  'unorm', componentType: 'float' },
   'rgba8unorm-srgb':        { renderable:  true, color:  true, depth: false, stencil: false, storage: false, copySrc:  true, copyDst:  true, bytesPerBlock:  4, blockWidth: 1, blockHeight: 1, dataType:  'unorm', componentType: 'float' },
-  'rgba8snorm':             { renderable: false, color:  true, depth: false, stencil: false, storage:  true, copySrc:  true, copyDst:  true, bytesPerBlock:  4, blockWidth: 1, blockHeight: 1, dataType:   'sint', componentType: 'float' },
+  'rgba8snorm':             { renderable: false, color:  true, depth: false, stencil: false, storage:  true, copySrc:  true, copyDst:  true, bytesPerBlock:  4, blockWidth: 1, blockHeight: 1, dataType:   'snorm', componentType: 'float' },
   'rgba8uint':              { renderable:  true, color:  true, depth: false, stencil: false, storage:  true, copySrc:  true, copyDst:  true, bytesPerBlock:  4, blockWidth: 1, blockHeight: 1, dataType:   'uint', componentType:  'uint' },
   'rgba8sint':              { renderable:  true, color:  true, depth: false, stencil: false, storage:  true, copySrc:  true, copyDst:  true, bytesPerBlock:  4, blockWidth: 1, blockHeight: 1, dataType:   'sint', componentType:  'sint' },
   'bgra8unorm':             { renderable:  true, color:  true, depth: false, stencil: false, storage: false, copySrc:  true, copyDst:  true, bytesPerBlock:  4, blockWidth: 1, blockHeight: 1, dataType:  'unorm', componentType: 'float' },
@@ -166,8 +166,8 @@ export const kSizedDepthStencilFormatInfo: {
     readonly bytesPerBlock: number;
     readonly blockWidth: 1;
     readonly blockHeight: 1;
-    dataType: TextureDataType;
-    componentType: GPUTextureComponentType;
+    readonly dataType: TextureDataType;
+    readonly componentType: GPUTextureComponentType;
   } & TextureFormatInfo;
 } = /* prettier-ignore */ {
   'depth32float': { renderable:  true, color: false, depth:  true, stencil: false, storage: false, copySrc:  true, copyDst: false, bytesPerBlock:  4, blockWidth: 1, blockHeight: 1, dataType:  'float', componentType: 'float' },
--- a/src/webgpu/capability_info.ts
+++ b/src/webgpu/capability_info.ts
@@ -102,56 +102,60 @@ type TextureFormatInfo = {
   // Add fields as needed
 };
 
+export type TextureDataType = 'uint' | 'sint' | 'unorm' | 'snorm' | 'float' | 'ufloat';
+
 export const kRegularTextureFormatInfo: {
   readonly [k in RegularTextureFormat]: {
     color: true;
     bytesPerBlock: number;
     blockWidth: 1;
     blockHeight: 1;
+    dataType: TextureDataType;
+    componentType: GPUTextureComponentType;
   } & TextureFormatInfo;
 } = /* prettier-ignore */ {
   // 8-bit formats
-  'r8unorm':                { renderable:  true, color:  true, depth: false, stencil: false, storage: false, copySrc:  true, copyDst:  true, bytesPerBlock:  1, blockWidth: 1, blockHeight: 1 },
-  'r8snorm':                { renderable: false, color:  true, depth: false, stencil: false, storage: false, copySrc:  true, copyDst:  true, bytesPerBlock:  1, blockWidth: 1, blockHeight: 1 },
-  'r8uint':                 { renderable:  true, color:  true, depth: false, stencil: false, storage: false, copySrc:  true, copyDst:  true, bytesPerBlock:  1, blockWidth: 1, blockHeight: 1 },
-  'r8sint':                 { renderable:  true, color:  true, depth: false, stencil: false, storage: false, copySrc:  true, copyDst:  true, bytesPerBlock:  1, blockWidth: 1, blockHeight: 1 },
+  'r8unorm':                { renderable:  true, color:  true, depth: false, stencil: false, storage: false, copySrc:  true, copyDst:  true, bytesPerBlock:  1, blockWidth: 1, blockHeight: 1, dataType:  'unorm', componentType: 'float' },
+  'r8snorm':                { renderable: false, color:  true, depth: false, stencil: false, storage: false, copySrc:  true, copyDst:  true, bytesPerBlock:  1, blockWidth: 1, blockHeight: 1, dataType:  'snorm', componentType: 'float' },
+  'r8uint':                 { renderable:  true, color:  true, depth: false, stencil: false, storage: false, copySrc:  true, copyDst:  true, bytesPerBlock:  1, blockWidth: 1, blockHeight: 1, dataType:   'uint', componentType:  'uint' },
+  'r8sint':                 { renderable:  true, color:  true, depth: false, stencil: false, storage: false, copySrc:  true, copyDst:  true, bytesPerBlock:  1, blockWidth: 1, blockHeight: 1, dataType:   'sint', componentType:  'sint' },
   // 16-bit formats
-  'r16uint':                { renderable:  true, color:  true, depth: false, stencil: false, storage: false, copySrc:  true, copyDst:  true, bytesPerBlock:  2, blockWidth: 1, blockHeight: 1 },
-  'r16sint':                { renderable:  true, color:  true, depth: false, stencil: false, storage: false, copySrc:  true, copyDst:  true, bytesPerBlock:  2, blockWidth: 1, blockHeight: 1 },
-  'r16float':               { renderable:  true, color:  true, depth: false, stencil: false, storage: false, copySrc:  true, copyDst:  true, bytesPerBlock:  2, blockWidth: 1, blockHeight: 1 },
-  'rg8unorm':               { renderable:  true, color:  true, depth: false, stencil: false, storage: false, copySrc:  true, copyDst:  true, bytesPerBlock:  2, blockWidth: 1, blockHeight: 1 },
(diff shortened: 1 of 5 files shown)


austinEng:
Running the tests in https://chromium-review.googlesource.com/c/chromium/src/+/2536758
it looks okay modulo precision..

Going to need to have some way to have a tolerance that varies based on ULPs, the expected value, and the component index  (since some formats have different precision for each component).

src/webgpu/util/texture/texelData.ts
```diff
@@ -66,21 +75,28 @@ const sint = (bitLength: number) => (n: number) => ({
...
 
-const sint8 = { write: sint(8), bitLength: 8 };
-const sint16 = { write: sint(16), bitLength: 16 };
-const sint32 = { write: sint(32), bitLength: 32 };
+const sint8 = { decode: identity, write: sint(8), bitLength: 8 };
+const sint16 = { decode: identity, write: sint(16), bitLength: 16 };
+const sint32 = { decode: identity, write: sint(32), bitLength: 32 };
 
 const float10 = {
+  decode: identity,
```
austinEng, line 99: Ideally, this wouldn't actually be the identity function but would cut the precision of the input value down so that it could be represented by a float10

I'm not sure if that will actually help the precision issues though or simply need a better way to compute tolerance.

src/webgpu/capability_info.ts
```diff
@@ -102,56 +102,60 @@ type TextureFormatInfo = {
...
+export type TextureDataType = 'uint' | 'sint' | 'unorm' | 'snorm' | 'float' | 'ufloat';
+
 export const kRegularTextureFormatInfo: {
   readonly [k in RegularTextureFormat]: {
     color: true;
     bytesPerBlock: number;
     blockWidth: 1;
     blockHeight: 1;
+    dataType: TextureDataType;
+    componentType: GPUTextureComponentType;
```
austinEng, line 114: The table is getting huge... maybe I should make a separate table for it?

src/webgpu/util/texture/texelData.spec.ts
```diff
@@ -0,0 +1,377 @@
...
+          // Test some values
+          makeParam(format, () => 0.119873046875),
+          makeParam(format, () => 1.40625),
+          makeParam(format, () => 24896),
+
+          // Test mixed values
+          makeParam(format, (bitLength, i) => {
+            if (format === 'rgb9e5ufloat') {
+              // These values can't be too scattered or the representation
+              // will be very imprecise.
```
kainino0x, line 367: We should also test values that are scattered though, as long as we know what to expect from the test (and I think we should be able to here)

src/webgpu/util/texture/texelData.spec.ts
```diff
@@ -0,0 +1,377 @@
...
+  }
+
+  const shader = `
+  [[set(0), binding(0)]] var<uniform_constant> tex : texture_2d<${shaderType}>;
+
+  [[block]] struct Output {
+    ${rep.componentOrder
+      .map((C, i) => {
+        return `[[offset(${i * 4})]] result${C} : ${shaderType};`;
+      })
```
kainino0x, line 66: nit: use plain lambda (omit `{ return` and `}`)

src/webgpu/util/texture/texelData.spec.ts
```diff
@@ -0,0 +1,377 @@
...
+      .join('\n')}
+  };
+  [[set(0), binding(1)]] var<storage_buffer> output : Output;
+
+  [[stage(compute)]]
+  fn main() -> void {
+      ${rep.componentOrder
+        .map(C => {
+          return `output.result${C} = textureLoad(tex, vec2<i32>(0, 0), 0).${C.toLowerCase()};`;
+        })
```
kainino0x, line 76: nit: plain lambda

src/webgpu/util/texture/texelData.spec.ts
```diff
@@ -0,0 +1,377 @@
...
+    format,
+    size: [1, 1, 1],
+    usage: GPUTextureUsage.COPY_DST | GPUTextureUsage.SAMPLED,
+  });
+
+  t.device.defaultQueue.writeTexture(
+    { texture },
+    texelData,
+    {
+      bytesPerRow: texelData.byteLength,
```
kainino0x, line 36: For Wx1x1 it should be possible to omit this.

src/webgpu/util/texture/texelData.spec.ts
```diff
@@ -0,0 +1,377 @@
...
+
+  const shader = `
+  [[set(0), binding(0)]] var<uniform_constant> tex : texture_2d<${shaderType}>;
+
+  [[block]] struct Output {
+    ${rep.componentOrder
+      .map((C, i) => {
+        return `[[offset(${i * 4})]] result${C} : ${shaderType};`;
+      })
+      .join('\n')}
```
kainino0x, line 67: Could just return a `vec4<shaderType>`. Then you can just `output.result = textureLoad(...)` I think. In the result you can just leave the unused components untested so we're not testing that behavior here (should be a separate test).

src/webgpu/util/texture/texelData.spec.ts
```diff
@@ -0,0 +1,377 @@
...
+  };
+};
+
+g.test('unorm_texel_data_in_shader')
+  .params(
+    params()
+      .combine(poptions('format', kEncodableTextureFormats))
+      .filter(({ format }) => {
+        return (
+          kEncodableTextureFormatInfo[format].copySrc &&
```
kainino0x, line 174: Seems only copyDst is needed? Same for other tests.

src/webgpu/capability_info.ts
```diff
@@ -162,9 +166,11 @@ export const kSizedDepthStencilFormatInfo: {
     readonly bytesPerBlock: number;
     readonly blockWidth: 1;
     readonly blockHeight: 1;
+    dataType: TextureDataType;
+    componentType: GPUTextureComponentType;
   } & TextureFormatInfo;
 } = /* prettier-ignore */ {
-  'depth32float': { renderable:  true, color: false, depth:  true, stencil: false, storage: false, copySrc:  true, copyDst: false, bytesPerBlock:  4, blockWidth: 1, blockHeight: 1 },
+  'depth32float': { renderable:  true, color: false, depth:  true, stencil: false, storage: false, copySrc:  true, copyDst: false, bytesPerBlock:  4, blockWidth: 1, blockHeight: 1, dataType:  'float', componentType: 'float' },
```
Kangz, line 173: Mmmh can't this also be "depth-comparison"?  How are we going to handle formats that support multiple components types? (or is this just for sampling in a texture to read the value and nothing else?)

src/webgpu/util/texture/texelData.spec.ts
```diff
@@ -0,0 +1,377 @@
...
+      rep.componentOrder.map(c => {
+        const value = rep.decode(componentData)[c];
+        assert(value !== undefined);
+        return value;
+      })
+    )
+  );
+}
+
+const testFn = (
```
Kangz, line 131: Uh I don't understand what this function is doing, can we have a comment?

src/webgpu/util/conversion.ts
```diff
@@ -54,6 +69,67 @@ export function float32ToFloatBits(
   return (sign << (exponentBits + fractionBits)) | (newBiasedExp << fractionBits) | newFraction;
 }
 
+// Three partial-precision floating-point numbers encoded into a single 32-bit value all
+// sharing the same 5-bit exponent (variant of s10e5, which is sign bit, 10-bit mantissa,
+// and 5-bit biased (15) exponent).
```
Kangz, line 74: I'm not sure what s10e5 does here since it isn't a format that's related (it isn't a compressed format).

src/webgpu/util/conversion.ts
```diff
@@ -54,6 +69,67 @@ export function float32ToFloatBits(
...
+  buf.setFloat32(0, r, true);
+  const rBits = buf.getUint32(0, true);
+
+  buf.setFloat32(0, g, true);
+  const gBits = buf.getUint32(0, true);
+
+  buf.setFloat32(0, b, true);
+  const bBits = buf.getUint32(0, true);
+
+  // >> to remove fraction, & to remove sign, - 127 to remove bias.
```
Kangz, line 98:
```suggestion
  // >> to remove mantissa, & to remove sign, - 127 to remove bias.
```

src/webgpu/util/conversion.ts
```diff
@@ -54,6 +69,67 @@ export function float32ToFloatBits(
...
+// There is no sign bit, and there is a shared 5-bit biased (15) exponent and a 9-bit
+// mantissa for each channel.
+export function encodeRGB9E5UFloat(r: number, g: number, b: number): number {
+  assert(r >= 0 && g >= 0 && b >= 0);
+  if (r === 0 && g === 0 && b === 0) {
+    return 0;
+  }
+
+  const bias = 15;
+  const fractionBits = 9;
```
Kangz, line 84:
```suggestion
  const mantissaBits = 9;
```

Kangz: ditto below

src/webgpu/util/texture/texelData.ts
```diff
@@ -331,6 +405,25 @@ class TexelDataRepresentationImpl implements TexelDataRepresentation {
...
+      assert(bytesPerBlock === 4);
+      assert(components.R !== undefined);
+      assert(components.G !== undefined);
+      assert(components.B !== undefined);
+
+      const buf = new ArrayBuffer(bytesPerBlock);
+      new DataView(buf).setUint32(
+        0,
+        encodeRGB9E5UFloat(components.R, components.G, components.B),
+        this.isGPULittleEndian
```
Kangz, line 422: I think we said this GPU were all going to be little endian?

Kangz (approved):

src/webgpu/util/conversion.ts
kainino0x: Documentation on rgb9e5 like
https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_texture_shared_exponent.txt
https://docs.microsoft.com/en-us/windows/win32/api/dxgiformat/ne-dxgiformat-dxgi_format
refer to rgb9e5 as related to s10e5 in that it has the same number of mantissa and exponent bits. (Though it's critically different in that the implicit leading value is 0, instead of 1)

src/webgpu/util/texture/texelData.spec.ts
austinEng: Maybe, but then I can't just use t.expectContents of the whole thing?

src/webgpu/util/texture/texelData.spec.ts
```diff
@@ -0,0 +1,377 @@
...
+            const offset = [0.13, 0.63, 0.42, 0.89];
+            const range = 2 * max(bitLength);
+            return -max(bitLength) + Math.floor(offset[i] * range);
+          }),
+        ];
+      })
+  )
+  .fn(testFn);
+
+g.test('uint_texel_data_in_shader')
```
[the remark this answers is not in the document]
austinEng: Could be, but I think `sint` and `snorm` should be separate at least since snorm tests -128 and -127 whereas sint cares only about -128. And if they're separate I'd rather keep uint and unorm separate for consistency.

src/webgpu/util/texture/texelData.ts
```diff
@@ -331,6 +405,25 @@ class TexelDataRepresentationImpl implements TexelDataRepresentation {
     const bytesPerBlock = kUncompressedTextureFormatInfo[this.format].bytesPerBlock;
     assert(!!bytesPerBlock);
 
+    if (this.format === 'rgb9e5ufloat') {
```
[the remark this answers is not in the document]
austinEng: Yes, we should definitely do this. I think originally I didn't realize we would have some formats that are packed such as this one.

src/webgpu/capability_info.ts
austinEng: I'm probably going to need to make this an array or we'll have to add per-aspect information at some point.. but for now it's just for the sampling part.

src/webgpu/util/conversion.ts
```diff
@@ -12,6 +12,21 @@ export function floatAsNormalizedInteger(float: number, bits: number, signed: bo
   }
 }
 
+export function normalizedIntegerAsFloat(integer: number, bits: number, signed: boolean): number {
```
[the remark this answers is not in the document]
austinEng: Done

src/webgpu/util/conversion.ts
```diff
@@ -54,6 +69,67 @@ export function float32ToFloatBits(
...
+  let exp = 0;
+  if (r !== 0) exp = rExp;
+  if (g !== 0) exp = gExp;
+  if (b !== 0) exp = bExp;
+
+  // Then get the smallest exponent
+  if (r !== 0) exp = Math.min(exp, rExp);
+  if (g !== 0) exp = Math.min(exp, gExp);
+  if (b !== 0) exp = Math.min(exp, bExp);
+
```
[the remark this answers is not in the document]
austinEng: Done

src/webgpu/util/texture/texelData.spec.ts
austinEng: Types aren't updated to have that yet

src/webgpu/util/texture/texelData.ts
austinEng: Ah, we can remove this globally in another PR then.

austinEng (commented):
PTAL again

src/webgpu/capability_info.ts
kainino0x: Given I think we're changing 'depth-comparison' to 'depth-stencil' upstream, maybe we don't even need to support 'float' component type for depth formats?

src/webgpu/util/texture/texelData.spec.ts
kainino0x: right.

src/webgpu/util/texture/texelData.spec.ts
kainino0x: True. Couldn't you still make this a simple vec4 and write into it selectively in main()? (Or can you not put swizzles on the left of `=` in wgsl?) Anyway, either is okay.

src/webgpu/util/texture/texelData.spec.ts
```diff
@@ -61,19 +73,15 @@ function doTest(
 
   [[block]] struct Output {
     ${rep.componentOrder
-      .map((C, i) => {
-        return `[[offset(${i * 4})]] result${C} : ${shaderType};`;
-      })
+      .map((C, i) => `[[offset(${i * 4})]] result${C} : ${shaderType};`)
       .join('\n')}
```
kainino0x, line 77: Maybe `pp` should allow interpolants to be arrays of strings that will get joined by `\n` automatically (and maybe even indented properly?)

src/webgpu/util/texture/texelData.spec.ts
```diff
@@ -61,19 +73,15 @@ function doTest(
...
   };
   [[set(0), binding(1)]] var<storage_buffer> output : Output;
 
   [[stage(compute)]]
   fn main() -> void {
       ${rep.componentOrder
-        .map(C => {
-          return `output.result${C} = textureLoad(tex, vec2<i32>(0, 0), 0).${C.toLowerCase()};`;
-        })
+        .map(C => `output.result${C} = textureLoad(tex, vec2<i32>(0, 0), 0).${C.toLowerCase()};`)
```
kainino0x, line 84: uNit: could textureLoad just once, then this code generation would be a little easier to read.

kainino0x (approved):
LGTM

src/webgpu/util/texture/texelData.spec.ts
austinEng: Okay made this load once but then write into individual scalar slots of the struct. I guess I could have made the struct vecN<type> but it doesn't seem like it makes much of a difference.

PR #353, opened 2020-11-14
Refactor texture zero init tests and use WGSL

austinEng:
- Use WGSL
- Remove base and derived test classes. There is one test class
  which imports implementations of the read method
- Reorder parameters to make more sense from a test suppression point-of-view
- Handling of texture formats with extensions
- Parameter for with/without canary values
- More param filtering for invalid cases
- Various changes for concision and readability


austinEng:
https://github.com/gpuweb/cts/compare/glsl-dependent...austinEng:zero-init-2

austinEng:
Sadly looks like I changed too much stuff for it to diff nicely.

Related to #352 . I think the way I'm doing the data values there is better. Going from data rep -> shader rep instead of the other way around in these zero init tests.
Main goal here is to port this to the `main` branch where the other texelData stuff is happening so I can then get rid of the funky indirection from abstract "state" value to texel component value. Open to suggestions for further simplifications.

Kangz:
+1, the way to do data rep is better in #352. Should we wait for that to land + this PR be rebased on top of it before reviewing?

austinEng:
I think you could review now as the conversion to WGSL and parameterization changes are orthogonal to inverting the way data rep works, so I won't do all of that in this PR.

src/webgpu/api/operation/resource_init/texture_zero_init.spec.ts
```diff
@@ -467,127 +423,197 @@ export abstract class TextureZeroInitTest extends GPUTest {
...
+  .combine(poptions('format', kUncompressedTextureFormats))
+  .expand(({ format }) => {
+    const info = kUncompressedTextureFormatInfo[format];
+    let aspects: { aspect: GPUTextureAspect }[] = [];
+    if (info.depth && info.stencil) {
+      aspects = [{ aspect: 'depth-only' }, { aspect: 'stencil-only' }];
+      return aspects;
+    }
+    aspects = [{ aspect: 'all' }];
+    return aspects;
```
kainino0x, line 447: nit: this could be simplified as below. Alternatively, could merge the `.filter(({ readMethod, aspect }) => ...` logic in here I think.

```suggestion
    if (info.depth && info.stencil) {
      return [{ aspect: 'depth-only' }, { aspect: 'stencil-only' }];
    }
    return [{ aspect: 'all' }];
```
or
```suggestion
    return poptions('aspect', (info.depth && info.stencil) ? ['depth-only', 'stencil-only'] : ['all']);
```

src/webgpu/api/operation/resource_init/texture_zero_init.spec.ts
```diff
@@ -467,127 +423,197 @@ export abstract class TextureZeroInitTest extends GPUTest {
...
+  })
+  .combine(poptions('mipLevelCount', kMipLevelCounts))
+  .combine(poptions('sampleCount', kSampleCounts))
+  // Multisampled textures may only have one mip
+  .unless(({ sampleCount, mipLevelCount }) => sampleCount > 1 && mipLevelCount > 1)
+  .combine(poptions('uninitializeMethod', kUninitializeMethods))
+  .unless(
```
kainino0x, line 454: Nit: these two `.unless`s only depend on readMethod and format, so can come immediately after both have been defined, near the top.

src/webgpu/api/operation/resource_init/texture_zero_init.spec.ts
```diff
@@ -467,127 +423,197 @@ export abstract class TextureZeroInitTest extends GPUTest {
...
-        .combine(pbool('nonPowerOfTwo'))
+      (readMethod === ReadMethod.DepthTest && !info.depth) ||
+      (readMethod === ReadMethod.StencilTest && !info.stencil) ||
+      (readMethod === ReadMethod.ColorBlending && !info.color) ||
+      // TODO: Test with depth sampling
+      (readMethod === ReadMethod.Sample && info.depth)
     );
-  }
+  })
+  .filter(({ readMethod, aspect }) => {
```
kainino0x, line 473: Same, can be just after aspect is defined.

src/webgpu/api/operation/resource_init/texture_zero_init.spec.ts
```diff
@@ -467,127 +423,197 @@ export abstract class TextureZeroInitTest extends GPUTest {
...
+  )
+  .combine(poptions('format', kUncompressedTextureFormats))
+  .expand(({ format }) => {
+    const info = kUncompressedTextureFormatInfo[format];
+    let aspects: { aspect: GPUTextureAspect }[] = [];
+    if (info.depth && info.stencil) {
+      aspects = [{ aspect: 'depth-only' }, { aspect: 'stencil-only' }];
+      return aspects;
+    }
+    aspects = [{ aspect: 'all' }];
```
kainino0x, line 446: This loses separate coverage for 'depth-only' and 'all' when both would be valid (same for stencil). Should we keep that coverage? It might behave a bit differently since implementations could have an error where they treat them differently for the purpose of identifying which aspects need to be cleared.

kainino0x (approved):
Shaders and params LGTM. Should we merge this first or 352?

src/webgpu/api/operation/resource_init/check_texture/by_ds_test.ts
```diff
@@ -0,0 +1,197 @@
+import { assert } from '../../../../../common/framework/util/util.js';
+import { GPUTest } from '../../../../gpu_test.js';
+import { CheckContents } from '../texture_zero_init.spec.js';
+
+function fullscreenVertexModule(device: GPUDevice) {
```
Kangz, line 5: nit: function name doesn't have a verb in it.

src/webgpu/api/operation/resource_init/check_texture/by_ds_test.ts
```diff
@@ -0,0 +1,197 @@
...
+        vec2<f32>( 3.0,  1.0),
+        vec2<f32>(-1.0,  1.0));
+      Position = vec4<f32>(pos[VertexIndex], 0.0, 1.0);
+      return;
+    }
+    `,
+  });
+}
+
+function getDepthTestReadbackPipeline(
```
Kangz, line 24: nit: I don't think this would be readback because it doesn't sample the texture. `getDepthTestEqualPipeline`?

src/webgpu/api/operation/resource_init/check_texture/by_ds_test.ts
```diff
@@ -0,0 +1,197 @@
...
+  params,
+  texture,
+  state,
+  subresourceRange
+) => {
+  for (const viewDescriptor of t.generateTextureViewDescriptorsForRendering(
+    params.aspect,
+    subresourceRange
+  )) {
+    const width = t.textureWidth >> viewDescriptor.baseMipLevel!;
```
Kangz, line 120: Shouldn't this be using a helper function to do the max(1, ...) ? ditto in other files.

src/webgpu/api/operation/resource_init/check_texture/by_ds_test.ts
```diff
@@ -0,0 +1,197 @@
...
+    depthStencilState: {
+      format,
+      depthCompare: 'equal',
+    },
+    primitiveTopology: 'triangle-list',
+    sampleCount,
+  });
+}
+
+function getStencilTestReadbackPipeline(
```
Kangz, line 65: ditto

src/webgpu/api/operation/resource_init/check_texture/by_sampling.ts
```diff
@@ -0,0 +1,149 @@
...
+      shaderType = 'i32';
+      ReadbackTypedArray = Int32Array;
+    } else if (params.format.indexOf('uint') !== -1) {
+      shaderType = 'u32';
+      ReadbackTypedArray = Uint32Array;
+    }
+
+    const componentOrder = rep.componentOrder;
+    const componentCount = componentOrder.length;
+    const indexExpression =
```
Kangz, line 42: Can we have a small example of the string that this produces? Is it a swizzle expression? (the variable could be renamed extraSwizzle?). Ah never mind we need the index expression for `vec2<f32>` but not `f32`. Can we have the comment that shows the code still?

src/webgpu/api/operation/resource_init/check_texture/by_sampling.ts
```diff
@@ -0,0 +1,149 @@
...
+        entries: [
+          {
+            binding: 0,
+            resource: { buffer: ubo },
+          },
+          {
+            binding: 1,
+            resource: texture.createView({
+              baseMipLevel: 0,
+              mipLevelCount: params.mipLevelCount,
```
Kangz, line 111: don't we want to check each mipLevel independently? Right now I believe only the first one will be checked.

src/webgpu/api/operation/resource_init/texture_zero_init.spec.ts
```diff
@@ -81,91 +87,49 @@ export const enum InitializedState {
...
-      unreachable();
-  }
-}
+const initializedStateAsSint = {
+  [InitializedState.Zero]: 0,
+  [InitializedState.Canary]: -1,
+};
 
-export function initializedStateAsColor(
+function initializedStateAsColor(
```
Kangz, line 105: I think this function could use the textureComponentType of the format instead of matching on substring?

src/webgpu/api/operation/resource_init/texture_zero_init.spec.ts
Kangz: Maybe this would help make the parameter generation marginally faster too?

Kangz (approved):

src/webgpu/api/operation/resource_init/check_texture/by_sampling.ts
austinEng: the level is passed in as a uniform

src/webgpu/api/operation/resource_init/texture_zero_init.spec.ts
austinEng: It could, but that's in the other patch. I'll fix this up after landing both.

src/webgpu/api/operation/resource_init/texture_zero_init.spec.ts
austinEng: We should indeed keep that coverage. I'll add it back.

austinEng:
@kainino0x PTAL again - I think this can land either before or after #352 - it shouldn't matter

src/webgpu/api/operation/resource_init/texture_zero_init.spec.ts
```diff
@@ -436,71 +440,47 @@ const paramsBuilder = params()
...
-  .filter(({ readMethod, aspect }) => {
-    if (readMethod === ReadMethod.DepthTest && aspect === 'stencil-only') {
-      return false;
-    }
-    if (readMethod === ReadMethod.StencilTest && aspect === 'depth-only') {
-      return false;
-    }
-    return true;
-  })
+  .combine(poptions('mipLevelCount', kMipLevelCounts))
```
kainino0x, line 462: This big params builder is a good candidate for subcases when we add them.

kainino0x (approved):

PR #358, opened 2020-11-18
 Merge branch 'main' into glsl-dependent

austinEng:


kainino0x:
Unfortunately I don't think there's any way to create a clean merge using a pull request.

- Squash will squash all the changes into a single-parent commit
- Rebase will create new copies of all of the `main` branch commits on the branch (and I think would fail)
- Merge will create an extra merge commit, so there are two merge commits.

The last one is ok, but I think we should just keep pushing the merges directly to the branch. Can you do that?

austinEng:
oh interesting. I pushed to origin/glsl-dependent and it autoclosed this

kainino0x:
whoa, cool! I wonder if that means we can require checks to have passed on the PR in order for the push to succeed?

PR #360, opened 2020-11-18
Add pp.join([...]), which joins an array of strings, preserving vertical alignment

### LINKED ISSUE #356: preprocessor: Join arrays of strings with '\n' for convenience
Maybe `pp` should allow interpolants to be arrays of strings that will get joined by `\n` automatically (and maybe even indented properly?)

_Originally posted by @kainino0x in https://github.com/gpuweb/cts/pull/352#discussion_r525585792_

kainino0x:
And use `pp` in a few places.

Fixes #356

### DIFF
--- a/src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
+++ b/src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
@@ -39,6 +39,7 @@ Test Coverage:
 `;
 
 import { pbool, poptions, params } from '../../../../common/framework/params_builder.js';
+import { pp } from '../../../../common/framework/preprocessor.js';
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
 import { assert } from '../../../../common/framework/util/util.js';
 import {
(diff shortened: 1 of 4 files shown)


src/webgpu/api/validation/resource_usages/textureUsageInPassEncoder.spec.ts
```diff
@@ -925,9 +926,9 @@ g.test('unused_bindings_in_pipeline')
       entry_point vertex = main;
     `;
     // TODO: revisit the shader code once 'image' can be supported in wgsl.
-    const wgslFragment = `
-      ${useBindGroup0 ? '[[set 0, binding 0]] var<image> image0;' : ''}
-      ${useBindGroup1 ? '[[set 1, binding 0]] var<image> image1;' : ''}
+    const wgslFragment = pp`
+      ${pp._if(useBindGroup0)}[[set 0, binding 0]] var<image> image0;${pp._endif}
+      ${pp._if(useBindGroup1)}[[set 1, binding 0]] var<image> image1;${pp._endif}
```
kainino0x, line 931: Maybe not better.

kainino0x (commented):
WDYT?

austinEng:
Hmm yea I don't see so much value in having it join lines by preserving indentation.

Overall, I think what's most important is readability of the template code which is better both without pp.join and without manual string formatting.

What could be neat though is a simple post process formatter to fixup indentation for logging purposes?

kainino0x:
Maintaining the indentation is mostly just incidental since I could do it easily. Originally, I was going to have any interpolant of type `string[]` automatically get joined by `\n`, but it seemed better to be explicit (at which point it's not really better than `.join('\n')`, I guess). Do you think it would be more useful that way?

> manual string formatting

what do you mean by this?

austinEng:
> Maintaining the indentation is mostly just incidental since I could do it easily. Originally, I was going to have any interpolant of type `string[]` automatically get joined by `\n`, but it seemed better to be explicit (at which point it's not really better than `.join('\n')`, I guess). Do you think it would be more useful that way?

I do think it would be easier to use. Seems like pp.join is roughly as much typing as `.join('\n')`

> 
> > manual string formatting
> 
> what do you mean by this?

Doing `.join('\n')`

kainino0x:
I may not keep this but if I do something like this, TODO: update `join` in setVertexBuffer.spec.ts

PR #361, opened 2020-11-18
Fix vscode workspace, add src/webgpu as a separate folder

kainino0x:
Only `"shell"` and `"process"` tasks are allowed in `.code-workspace` files. Replace all the `"grunt"` tasks with `"shell"` tasks.
Configure `package.json`'s `"scripts"` to expose the relevant build commands from Grunt.

Also adds `src/webgpu` as a separate folder in the vscode workspace, so the test file hierarchy is clearly visible without getting mixed in with everything else.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -185,24 +185,25 @@ module.exports = function (grunt) {
     'run:unittest',
     'run:lint',
   ]);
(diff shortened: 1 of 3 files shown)


kainino0x:
(PTAL)

austinEng (approved):
LGTM

PR #365, opened 2020-11-19
Simplify texelData helpers

austinEng:
- Removes componentType from capability_info since it can be computed
  from dataType.
- Removes dataType from capability_info since it is only used by
  texelData.
- Removes multiple impl of X -> packed representation. Now just data rep -> packed.
   Previously:
  - shader rep -> packed
  - data rep -> packed
- Standardizes helpers for decoding from or encoding into the data
  representation.
- Assorted code cleanup

### DIFF
--- a/src/webgpu/api/operation/resource_init/check_texture/by_copy.ts
+++ b/src/webgpu/api/operation/resource_init/check_texture/by_copy.ts
@@ -3,7 +3,7 @@ import {
   EncodableTextureFormat,
   kEncodableTextureFormatInfo,
 } from '../../../../capability_info.js';
-import { CheckContents } from '../textureZero.spec.js';
+import { CheckContents } from '../texture_zero.js';
 
 export const checkContentsByBufferCopy: CheckContents = (
   t,
--- a/src/webgpu/api/operation/resource_init/check_texture/by_copy.ts
+++ b/src/webgpu/api/operation/resource_init/check_texture/by_copy.ts
@@ -3,7 +3,7 @@ import {
   EncodableTextureFormat,
   kEncodableTextureFormatInfo,
 } from '../../../../capability_info.js';
-import { CheckContents } from '../texture_zero_init.spec.js';
+import { CheckContents } from '../textureZero.spec.js';
 
 export const checkContentsByBufferCopy: CheckContents = (
   t,
--- a/src/webgpu/util/texture/texelData.ts
+++ b/src/webgpu/util/texture/texelData.ts
@@ -1,18 +1,13 @@
 import { assert, unreachable } from '../../../common/framework/util/util.js';
-import {
-  EncodableTextureFormat,
-  kEncodableTextureFormatInfo,
-  kUncompressedTextureFormatInfo,
-  UncompressedTextureFormat,
-} from '../../capability_info.js';
+import { UncompressedTextureFormat } from '../../capability_info.js';
 import {
   assertInIntegerRange,
   float32ToFloatBits,
   floatAsNormalizedInteger,
   gammaCompress,
-  encodeRGB9E5UFloat,
-  normalizedIntegerAsFloat,
   gammaDecompress,
+  normalizedIntegerAsFloat,
+  packRGB9E5UFloat,
 } from '../conversion.js';
 
 export const enum TexelComponent {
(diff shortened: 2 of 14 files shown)


src/webgpu/util/texture/texelData.ts
```diff
@@ -26,263 +21,101 @@ export const enum TexelComponent {
...
-
-const uint8 = { decode: identity, write: uint(8), bitLength: 8 };
-const uint16 = { decode: identity, write: uint(16), bitLength: 16 };
-const uint32 = { decode: identity, write: uint(32), bitLength: 32 };
+}>;
 
-const sint8 = { decode: identity, write: sint(8), bitLength: 8 };
-const sint16 = { decode: identity, write: sint(16), bitLength: 16 };
-const sint32 = { decode: identity, write: sint(32), bitLength: 32 };
+// A function which maps component values to component values.
```
kainino0x, line 31: nit: Could use `/** doc comment */` for many of these.

src/webgpu/util/texture/texelData.ts
```diff
@@ -322,170 +155,364 @@ class TexelDataRepresentationImpl implements TexelDataRepresentation {
...
-      assert(componentValue !== undefined);
-      this.setComponentBytes(data, c, componentValue);
-    }
-    return data;
-  }
+export const kTexelRepresentationInfo: {
+  readonly [k in UncompressedTextureFormat]: {
+    readonly componentOrder: TexelComponent[];
+    readonly componentInfo: TexelComponentInfo;
+    readonly encode: ComponentMapFn; // Encode shader values into their data representation. ex.) float 1.0 -> unorm8 255
```
kainino0x, line 307: Can also be doc comments (needs to be on the line before the documented item)

src/webgpu/util/texture/texelData.ts
```diff
@@ -322,170 +155,364 @@ class TexelDataRepresentationImpl implements TexelDataRepresentation {
...
+      },
+      encode: components => {
+        components = applyEach(
+          (n: number) => (floatAsNormalizedInteger(n, 10, false), n),
+          kRGB
+        )(components);
+        components = applyEach((n: number) => (floatAsNormalizedInteger(n, 2, false), n), [
+          TexelComponent.A,
+        ])(components);
+        return components;
```
kainino0x, line 365: I think perhaps this could be more clearly written as a literal with `{ R: ..., G: ..., B: ..., A: ... }` and a little repetition.

Or if the floatAsNormalizedInteger results are supposed to be discarded, simply call those 4 and then return `components` unmodified?

src/webgpu/util/texture/texelData.ts
```diff
@@ -322,170 +155,364 @@ class TexelDataRepresentationImpl implements TexelDataRepresentation {
...
+      componentOrder: kRGBA,
+      componentInfo: {
+        [TexelComponent.R]: { dataType: 'unorm', bitLength: 10 },
+        [TexelComponent.G]: { dataType: 'unorm', bitLength: 10 },
+        [TexelComponent.B]: { dataType: 'unorm', bitLength: 10 },
+        [TexelComponent.A]: { dataType: 'unorm', bitLength: 2 },
+      },
+      encode: components => {
+        components = applyEach(
+          (n: number) => (floatAsNormalizedInteger(n, 10, false), n),
```
kainino0x, line 359: This is discarding the result of floatAsNormalizedInteger? I don't get why. Are the packed formats special in that they're encoded later?

src/webgpu/util/texture/texelData.ts
```diff
@@ -322,170 +155,364 @@ class TexelDataRepresentationImpl implements TexelDataRepresentation {
...
+      },
+      decode: components => {
+        components = applyEach(
+          (n: number) => (normalizedIntegerAsFloat(n, 10, false), n),
+          kRGB
+        )(components);
+        components = applyEach((n: number) => (normalizedIntegerAsFloat(n, 2, false), n), [
+          TexelComponent.A,
+        ])(components);
+        return components;
```
kainino0x, line 375: same comments

src/webgpu/util/texture/texelData.ts
```diff
@@ -322,170 +155,364 @@ class TexelDataRepresentationImpl implements TexelDataRepresentation {
...
+      componentInfo: {
+        [TexelComponent.R]: { dataType: 'ufloat', bitLength: 11 },
+        [TexelComponent.G]: { dataType: 'ufloat', bitLength: 11 },
+        [TexelComponent.B]: { dataType: 'ufloat', bitLength: 10 },
+      },
+      pack: components => {
+        components = applyEach((n: number) => float32ToFloatBits(n, 0, 5, 6, 15), [
+          TexelComponent.R,
+          TexelComponent.G,
+        ])(components);
```
kainino0x, line 403: kRG

src/webgpu/util/texture/texelData.ts
```diff
@@ -322,170 +155,364 @@ class TexelDataRepresentationImpl implements TexelDataRepresentation {
...
+export function getComponentReadbackTraits(dataType: ComponentDataType) {
+  let ReadbackTypedArray: Float32ArrayConstructor | Int32ArrayConstructor | Uint32ArrayConstructor;
+  let shaderType: 'i32' | 'u32' | 'f32';
+  switch (dataType) {
+    case 'ufloat':
+    case 'float':
+    case 'unorm':
+    case 'snorm':
+      ReadbackTypedArray = Float32Array;
+      shaderType = 'f32';
```
kainino0x, line 504: nit: Could just `return` inline inside each case.

kainino0x (commented):
Overall seems like a nice improvement. I think I mostly understand it. Few comments.

Kangz (approved):
No real additional comment.

src/webgpu/util/texture/texelData.ts
```diff
@@ -26,263 +21,101 @@ export const enum TexelComponent {
...
-
-  private writeTexelData(
-    data: ArrayBuffer,
-    bitOffset: number,
-    bitLength: number,
-    type: TexelWriteType,
-    value: number
-  ) {
+  components: PerTexelComponent<number>,
+  bitLengths: number | PerTexelComponent<number>,
```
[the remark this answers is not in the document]
austinEng: You could, but that would result in 4 overloads for a function like this? Or I guess it could be 2 to be simple, but the implementation code would be pretty much the same - like it still has to check the type of the arguments.

src/webgpu/util/texture/texelData.ts
austinEng: hm it's not - this is doing something akin to:

components = Object.assign({}, components, {
  R: floatAsNormalized(components.R),
  G: floatAsNormalized(components.G),
  B: floatAsNormalized(components.B),
});
components = Object.assign({}, components, {
  A: floatAsNormalized(components.A),
});

austinEng: but yes as mentioned below maybe it would be more clear to just repeat the code

src/webgpu/util/texture/texelData.ts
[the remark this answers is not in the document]
kainino0x: You can't have real overloads; you can only provide multiple signatures for a single function. This is because JS has no overloading (of course, since there are no types) and TS won't do things like create mangled names to implement overloads.

src/webgpu/util/texture/texelData.ts
kainino0x: but shouldn't `(n: number) => (floatAsNormalizedInteger(n, 10, false), n)` be an identity function (sequence operator `,`), except for side effects?

src/webgpu/api/operation/resource_init/check_texture/by_copy.ts
```diff
@@ -3,7 +3,7 @@ import {
   EncodableTextureFormat,
   kEncodableTextureFormatInfo,
 } from '../../../../capability_info.js';
-import { CheckContents } from '../texture_zero_init.spec.js';
+import { CheckContents } from '../textureZero.spec.js';
```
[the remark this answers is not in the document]
austinEng: Ah, I renamed the other one to texel_data then

src/webgpu/util/texture/texelData.ts
```diff
@@ -26,263 +21,101 @@ export const enum TexelComponent {
...
+  const componentDataTypeMap =
+    typeof componentDataTypes === 'string' || componentDataTypes === null
+      ? makePerTexelComponent(componentOrder, componentDataTypes)
+      : componentDataTypes;
+
+  const totalBitLength = Object.entries(bitLengthMap).reduce((acc, [, value]) => {
+    assert(value !== undefined);
+    return acc + value;
+  }, 0);
+  assert(8 * Math.floor(totalBitLength / 8) === totalBitLength);
```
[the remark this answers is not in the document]
austinEng: Done.

src/webgpu/util/texture/texelData.ts
austinEng: Done.

src/webgpu/util/texture/texelData.ts
austinEng: Done. The discarding was a bug.

src/webgpu/util/texture/texelData.ts
austinEng: Done. The discarding was a bug.

src/webgpu/util/texture/texelData.ts
austinEng: Done.

src/webgpu/util/texture/texelData.ts
austinEng: Done.

src/webgpu/util/texture/texel_data.ts
```diff
@@ -28,13 +28,26 @@ type TexelComponentInfo = PerTexelComponent<{
   bitLength: number;
 }>;
 
-// A function which maps component values to component values.
+/**
+ * Maps component values to component values
+ * @param {PerTexelComponent<number>} components - The input components.
+ * @returns {PerTexelComponent<number>} The new output components.
```
kainino0x, line 34: FYI it would have also been okay to omit params/returns if they're somewhat trivial (like here); `/** */` is still more useful than `//`. More documentation is better though.

src/webgpu/util/texture/texel_data.ts
```diff
@@ -355,24 +404,28 @@ export const kTexelRepresentationInfo: {
...
-        return components;
+        assert(components.R !== undefined);
+        assert(components.G !== undefined);
+        assert(components.B !== undefined);
+        assert(components.A !== undefined);
+        return {
+          [TexelComponent.R]: floatAsNormalizedInteger(components.R, 10, false),
+          [TexelComponent.G]: floatAsNormalizedInteger(components.G, 10, false),
+          [TexelComponent.B]: floatAsNormalizedInteger(components.B, 10, false),
+          [TexelComponent.A]: floatAsNormalizedInteger(components.A, 2, false),
```
kainino0x, line 415: nit: Just `R:`, `G:`, etc?

src/webgpu/util/texture/texel_data.ts
```diff
@@ -355,24 +404,28 @@ export const kTexelRepresentationInfo: {
...
-        components = applyEach((n: number) => (floatAsNormalizedInteger(n, 2, false), n), [
-          TexelComponent.A,
-        ])(components);
-        return components;
+        assert(components.R !== undefined);
+        assert(components.G !== undefined);
+        assert(components.B !== undefined);
+        assert(components.A !== undefined);
+        return {
+          [TexelComponent.R]: floatAsNormalizedInteger(components.R, 10, false),
```
kainino0x, line 412: This is silly but maybe:
```suggestion
          [TexelComponent.R]: floatAsNormalizedInteger(components.R ?? unreachable(), 10, false),
```

kainino0x (approved):

src/webgpu/util/texture/texel_data.ts
austinEng: Kinda like it. Done.

src/webgpu/util/texture/texel_data.ts
austinEng: Done.

kainino0x (approved):

PR #370, opened 2020-11-20
Port culling tests from GLSL to WGSL

Kangz:


Kangz:
Nice diff is here: https://github.com/gpuweb/cts/compare/glsl-dependent..Kangz:wgsl_culling#diff-b20e284525e71afca4de5f6f7c328e54da9e2dab52e04707308f799be204e474

Kangz:
Ah also note that `[front_facing]` being a `bool` is technically not correct per the WGSL spec but is needed to make things work. I opened a WGSL PR for it https://github.com/gpuweb/gpuweb/pull/1242

austinEng (approved):
LGTM

kainino0x (approved):

PR #372, opened 2020-11-20
Port createRenderPipeline validation tests from GLSL to WGSL

Kangz:
Currently fails because of https://bugs.chromium.org/p/tint/issues/detail?id=352

### DIFF
--- a/src/webgpu/api/validation/createRenderPipeline.spec.ts
+++ b/src/webgpu/api/validation/createRenderPipeline.spec.ts
@@ -1,0 +1,228 @@
+export const description = `
+createRenderPipeline validation tests.
+`;
+
+import { poptions } from '../../../common/framework/params_builder.js';
+import { makeTestGroup } from '../../../common/framework/test_group.js';
+import { kAllTextureFormats, kAllTextureFormatInfo } from '../../capability_info.js';
+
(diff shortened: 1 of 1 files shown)


Kangz:
PTAL https://github.com/gpuweb/cts/compare/glsl-dependent..Kangz:wgsl_createRenderPipeline#diff-71ad9b35ae56b357c4f2a0eb3b3943cc794dace1fb39d5f0634396297a815ac7

src/webgpu/api/validation/createRenderPipeline.spec.ts
```diff
@@ -0,0 +1,226 @@
...
+            }`,
+        }),
+        entryPoint: 'main',
+      },
+      fragmentStage: {
+        module: this.device.createShaderModule({
+          code: `
+            [[location(0)]] var<out> fragColor : vec4<${fragColorType}>;
+            [[stage(fragment)]] fn main() -> void {
+              fragColor = vec4<${fragColorType}>(vec4<f32>(0.0, 1.0, 0.0, 1.0));
```
austinEng, line 57: maybe it will be better to choose the suffix above which is either ".0", "u", or ""

austinEng (approved):
explicit return should no longer be needed

PR #377, opened 2020-11-20
[wgsl] Add runtime array validation tests

sarahM0:
Adding runtime array validation tests for following rules:
v-0015: The last member of the structure type defining the "store type" for variable in the storage storage class may be a runtime-sized array.
v-0030: A runtime-sized array must not be used as the store type or contained within a store type except as allowed by v-0015
v-0031: The type of an expression must not be a runtime-sized array type.
v-0032: A runtime-sized array must have a stride attribute.

### DIFF
--- a/src/webgpu/shader/validation/wgsl/runtime-array-is-store-type.fail.wgsl
+++ b/src/webgpu/shader/validation/wgsl/runtime-array-is-store-type.fail.wgsl
@@ -1,12 +1,12 @@
-# v-0031 - This fails because the runtime array must not be the store type of variable 'buf'.
+# v-0015: variable 's' store type is struct 'SArr' that has a runtime-sized member but its storage class is not 'storage'.
 
+type RTArr = [[stride (16)]] array<vec4<f32>>;
 [[block]]
-struct Foo {
-  [[offset (0)]] a : [[stride (4)]] array<f32>;
+struct SArr{
+  [[offset(0)]] data : RTArr;
 };
-[[set(0), binding(1)]] var<storage> buf: Foo;
 
 [[stage(vertex)]]
 fn main() -> void {
-  return;
+  var s : SArr;
 }
--- a/src/webgpu/shader/validation/wgsl/runtime-array-is-store-type.fail.wgsl
+++ b/src/webgpu/shader/validation/wgsl/runtime-array-is-store-type.fail.wgsl
@@ -1,0 +1,12 @@
+# v-0031 - This fails because the runtime array must not be the store type of variable 'buf'.
+
+[[block]]
+struct Foo {
+  [[offset (0)]] a : [[stride (4)]] array<f32>;
+};
+[[set(0), binding(1)]] var<storage> buf: Foo;
+
+[[stage(vertex)]]
+fn main() -> void {
+  return;
+}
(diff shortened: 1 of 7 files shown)


kainino0x:
Just invited you to the org so you can request reviews from specific people. (oops, I thought I had done this before, but must have forgot)

src/webgpu/shader/validation/wgsl/runtime-array-not-last-v3.fail.wgsl
```diff
@@ -0,0 +1,8 @@
+# v-0015 - This fails because of the runtime array is not last member of the struct.
```
kainino0x, line 1: This is a weird error message for this case since there's no struct. I think it could simply say "of a struct" or "of a [[block]] struct" at the end and that would be ok.

src/webgpu/shader/validation/wgsl/runtime-array-not-last.fail.wgsl
```diff
@@ -1,5 +1,6 @@
 # v-0015 - This fails because of the runtime array is not last member of the struct.
 
+[[block]]
 struct Foo {
   [[offset (0)]] a : array<f32>;
```
kainino0x, line 5: Isn't this also invalid because the array has no stride?

kainino0x (commented):
Few comments but as usual I'll let someone else review (presumably @dj2)

src/webgpu/shader/validation/wgsl/runtime-array-not-last-v3.fail.wgsl
sarahM0: Yes, that's right! fixed.

src/webgpu/shader/validation/wgsl/runtime-array-not-last.fail.wgsl
sarahM0: I cannot find this in spec. Could you point my to what you're referring to?

src/webgpu/shader/validation/wgsl/runtime-array-is-store-type.fail.wgsl
```diff
@@ -0,0 +1,12 @@
+# v-0031 - This fails because the runtime array must not be the store type of variable 'buf'.
```
dj2, line 1: This sentence doesn't seem quite right, the store type of `buf` is `Foo` not the runtime array.

src/webgpu/shader/validation/wgsl/runtime-array-not-block-decorated.fail.wgsl
```diff
@@ -0,0 +1,10 @@
+# v-0031: struct 'S' has runtime-sized member but it's storage class is not 'storage'.
```
dj2, line 1: There is no `var` in this example, so nothing with a storage class, is the failure due to the missing `block` decoration?

src/webgpu/shader/validation/wgsl/runtime-array-not-last-v3.fail.wgsl
```diff
@@ -0,0 +1,8 @@
+# v-0015 - This fails because of the runtime array is not last member of a [[block]] struct.
```
dj2, line 1: This doesn't seem quite right as there is no `struct`, the runtime array is used as the store type?

src/webgpu/shader/validation/wgsl/runtime-array-not-last-v2.fail.wgsl
```diff
@@ -0,0 +1,12 @@
+# v-0015 - This fails because of the runtime array is not last member of the struct.
```
dj2, line 1: Maybe mention the array is aliased to show the difference to runtime-array-not-last.fail.wgsl

dj2 (approved):

PR #378, opened 2020-11-20
Update the standalone test runner to compile Typescript sources at runtime

austinEng:


service-worker/index.ts
```diff
@@ -0,0 +1,118 @@
...
+
+            compilerOptions.module = asModuleKind(
+              tsconfig['compilerOptions']['module']);
+            compilerOptions.target = asScriptTarget(
+              tsconfig['compilerOptions']['target'] || 'latest');
+            compilerOptions.moduleResolution = asModuleResolutionKind(
+              tsconfig['compilerOptions']['moduleResolution']);
+
+            // https://github.com/microsoft/TypeScript/wiki/Using-the-Compiler-API
+            const output = ts.transpile(file, compilerOptions);
```
austinEng, line 107: Would be good if this could use Babel as well since that's what we use for the actual build.

standalone/index.html
```diff
@@ -218,7 +218,21 @@ <h1>WebGPU Conformance Test Suite</h1>
     <p>
       <input type="button" id="copyResultsJSON" value="Copy results as JSON">
     </p>
+    <script>
+      if ('serviceWorker' in navigator) {
```
austinEng, line 222: Need to guard this on some define like IS_STANDALONE_DEV or something

package.json
```diff
@@ -44,13 +45,19 @@
...
     "grunt-ts": "^6.0.0-beta.22",
     "gts": "^3.0.1",
+    "json5-loader": "^4.0.1",
     "quiet-grunt": "^0.2.3",
+    "ts-loader": "^8.0.5",
     "ts-node": "^9.0.0",
-    "typescript": "^4.0.3"
-  }
+    "typescript": "^4.0.3",
+    "val-loader": "^2.1.2",
```
austinEng, line 57: `val-loader` is unused.

austinEng (commented):
Here's the overall structure, still need some tweaks. Sending it out now for early feedback.

service-worker/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
```
kainino0x, line 1: May not be relevant with babel, but this is only used in service-worker/index.ts I assume? Could we just hardcode these into that file?

standalone/index.html
kainino0x: What would that do? This file is only used for standalone dev. Should probably throw an exception if service workers aren't available.

standalone/index.html
```diff
@@ -218,7 +218,21 @@ <h1>WebGPU Conformance Test Suite</h1>
...
+          .then(async function (registration) {
+            await registration.update();
+            console.debug("Service worker registered");
 
-    <script type="module" src="../out/common/runtime/standalone.js"></script>
+            const script = document.createElement('script');
+            script.type = "module";
+            script.src = "../src/common/runtime/standalone.ts";
+
+            document.body.append(script);
```
kainino0x, line 233: Could this be just `import('../src/common/runtime/standalone.ts')`?

build/webpack.dev.js
```diff
@@ -0,0 +1,67 @@
+const path = require('path');
```
kainino0x, line 1: How does webpack get triggered?

build/listing-loader.js
```diff
@@ -0,0 +1,54 @@
...
+const { Compilation } = require('webpack');
+
+require('../src/common/tools/setup-ts-in-node');
+
+const { crawl } = require('../src/common/tools/crawl.ts');
+const { assert } = require('../src/common/framework/util/util.ts');
+
+// See https://webpack.js.org/api/loaders/.
+// Module loader that replaces an import with a module that contains the generated
+// listing for a test suite. Assumes the listing is at /path/to/{suite}/listing.ts.
```
kainino0x, line 11: I was thinking for the moment we might still require a build step when new files are added. OTOH, if we have code to generate listing.ts running server-side (which I assume this is?) then should we just do all the TS compilation server-side...?

service-worker/tsconfig.json
austinEng: This tsconfig is for the actual service-worker/index.ts file. But it imports the root tsconfig.json for runtime compilation of the other sources.

standalone/index.html
austinEng: Yea, probably.

build/webpack.dev.js
austinEng: This is happening with `npm run dev`. Which will both start and http server and watch changes to recompile if the service worker file changes (maybe a but overkill but was very nice for development).

build/listing-loader.js
austinEng: I didn't actually verify this yet, but Webpack's watcher should regenerate the listing on changes. If it doesn't we should be able to set it up to do that.

But yea you're right we could probably make all the compilation happen server side without the need for service worker. It'll probably be faster that way too. 

It might be cleaner to just write our own http server for this.

standalone/index.html
austinEng: It's also used for the standalone build that's published to gh-pages.

Maybe we could also turn this file into a template.

build/listing-loader.js
austinEng: I'm going to try our own http server. It avoids bundling shenanigans and is less memory and is much simpler, especially if we intend for this to be dev only.

standalone/index.html
kainino0x: ah, right, got it.

build/listing-loader.js
kainino0x: Sounds good. I really would like to go the service worker route for WPT someday, but given we have no way to do this now with a reasonable file size, this does seem more straightforward.

austinEng: I was pretty excited about this, but now seeing that the compiles take tens and sometimes 100-something milliseconds (maybe cause the browser is busy doing something else like fetching, etc.) it seems like it potentially could be rather detrimental for test runner performance and increase run times on CI a lot. Maybe the code structure could be better, because it looks like with the service worker, fetches and compiles aren't happening in parallel. If we did Promise.all of all the dynamic `import`s instead of `await`ing each, maybe this would be less of a problem.

How do you imagine we would go the service worker route for WPT while maintaining a reasonable file size?
 - WPT gains a fetch deps + build step? (might as well just compile the TS here)
 - We use a tiny TS compiler?

If we want WPT service worker to happen in the nearish-term, maybe we should stick with service worker on standalone so most of the code is there.

austinEng: another important bit here is that service worker won't work if you serve the CTS from a remote server. It's explicitly allowed for `localhost`, but if you try to get at it from some random IP address, service worker isn't available - unless you run with --disable-web-security

kainino0x: We can definitely get the fetches to happen in parallel. Though I assume by default the service worker won't actually parallelize the synchronous TS compilation?

My only idea for WPT is to use a tiny TS compiler (I haven't been able to find any such thing, even though it's theoretically possible), or get approval to drop a big dep into WPT (no idea how likely).

Good to know about the `localhost` thing, is that it would work on localhost or https but not http?

austinEng: Yup, exactly

austinEng: Server version is in #386

austinEng:
Closing in favor of #386

PR #368, opened 2020-11-20
From HackMD plan, part 2: some more validation tests

kainino0x:
Also implemented a few very simple tests.

PR #383, opened 2020-11-21
iterateCollapsedQueries: don't yield files with no tests

kainino0x:
This prevents gen_wpt_cts_html, which generates a list of test queries to cover the whole CTS, from generating "variant" entries for test files (groups) that don't have any tests in them (which throws an error when the browser tries to actually run those variants). I added several such files, as TODOs, in #368.


austinEng (approved):
LGTM

PR #384, opened 2020-11-23
Memory sync test for buffer: write after write in the same pass

Richard-Yunchao:
PTAL. Thanks. The naming like verifyAlternativeData and expectAlternativeContents might not be very good, though. But I can't think of a better one.

### DIFF
--- a/src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
+++ b/src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
@@ -256,11 +256,11 @@ export class BufferSyncTest extends GPUTest {
     this.expectContents(buffer, bufferData);
   }
 
-  verifyAlternativeData(buffer: GPUBuffer, expectedValue1: number, expectedValue2: number) {
+  verifyDataMultipleValidValues(buffer: GPUBuffer, expectedValue1: number, expectedValue2: number) {
     const bufferData1 = new Uint32Array(1);
     bufferData1[0] = expectedValue1;
     const bufferData2 = new Uint32Array(1);
     bufferData2[0] = expectedValue2;
-    this.expectAlternativeContents(buffer, bufferData1, bufferData2);
+    this.expectContentsMultipleValidValues(buffer, bufferData1, bufferData2);
   }
 }
--- a/src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
+++ b/src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
@@ -255,4 +255,12 @@ export class BufferSyncTest extends GPUTest {
     bufferData[0] = expectedValue;
     this.expectContents(buffer, bufferData);
   }
+
+  verifyAlternativeData(buffer: GPUBuffer, expectedValue1: number, expectedValue2: number) {
+    const bufferData1 = new Uint32Array(1);
+    bufferData1[0] = expectedValue1;
+    const bufferData2 = new Uint32Array(1);
+    bufferData2[0] = expectedValue2;
+    this.expectAlternativeContents(buffer, bufferData1, bufferData2);
+  }
 }
--- a/src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
+++ b/src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
@@ -256,11 +256,11 @@ export class BufferSyncTest extends GPUTest {
     this.expectContents(buffer, bufferData);
   }
 
-  verifyDataMultipleValidValues(buffer: GPUBuffer, expectedValue1: number, expectedValue2: number) {
+  verifyDataTwoValidValues(buffer: GPUBuffer, expectedValue1: number, expectedValue2: number) {
     const bufferData1 = new Uint32Array(1);
     bufferData1[0] = expectedValue1;
     const bufferData2 = new Uint32Array(1);
     bufferData2[0] = expectedValue2;
-    this.expectContentsMultipleValidValues(buffer, bufferData1, bufferData2);
+    this.expectContentsTwoValidValues(buffer, bufferData1, bufferData2);
   }
 }
--- a/src/webgpu/api/operation/memory_sync/buffer/ww.spec.ts
+++ b/src/webgpu/api/operation/memory_sync/buffer/ww.spec.ts
@@ -83,16 +83,16 @@ g.test('two_draws_in_the_same_render_pass')
     a storage buffer. The second write will write 2 into the same buffer in the same pass. Expected
     data in buffer is either 1 or 2. It may use bundle in each draw.`
   )
-  .params(pbool('useBundle'))
+  .params(params().combine(pbool('firstDrawUseBundle')).combine(pbool('secondDrawUseBundle')))
   .fn(async t => {
-    const { useBundle } = t.params;
+    const { firstDrawUseBundle, secondDrawUseBundle } = t.params;
     const buffer = await t.createBufferWithValue(0);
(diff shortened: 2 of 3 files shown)


src/webgpu/gpu_test.ts
```diff
@@ -167,6 +167,33 @@ export class GPUTest extends Fixture {
...
+    );
+
+    this.eventualAsyncExpectation(async niceStack => {
+      const constructor = expected1.constructor as TypedArrayBufferViewConstructor;
+      await dst.mapAsync(GPUMapMode.READ);
+      const actual = new constructor(dst.getMappedRange());
+      const check1 = this.checkBuffer(actual.subarray(begin, end), expected1);
+      const check2 = this.checkBuffer(actual.subarray(begin, end), expected2);
+      if (check1 !== undefined && check2 !== undefined) {
+        niceStack.message = check1 + check2;
```
kainino0x, line 190: IIRC just concatenating these messages isn't going to make much sense. Maybe add an explanation (and a newline):
```suggestion
        niceStack.message = `Expected one of the following two checks to succeed:
- ${check1}
- ${check2}`;
```

src/webgpu/gpu_test.ts
```diff
@@ -167,6 +167,33 @@ export class GPUTest extends Fixture {
     });
   }
 
+  expectAlternativeContents(
```
kainino0x, line 170: expectContentsMultipleValidValues?
expectContentsNonDeterministic?

src/webgpu/api/operation/memory_sync/buffer/ww.spec.ts
```diff
@@ -83,12 +83,54 @@ g.test('two_draws_in_the_same_render_pass')
     a storage buffer. The second write will write 2 into the same buffer in the same pass. Expected
     data in buffer is either 1 or 2. It may use bundle in each draw.`
   )
-  .unimplemented();
+  .params(pbool('useBundle'))
```
kainino0x, line 86: Now that I'm looking at the implementation, I think we should test at least these three options:
- same pass, no bundle
- same pass, different bundles
- same bundle

kainino0x (approved):

src/webgpu/gpu_test.ts
```diff
@@ -187,7 +187,9 @@ export class GPUTest extends Fixture {
       const check1 = this.checkBuffer(actual.subarray(begin, end), expected1);
       const check2 = this.checkBuffer(actual.subarray(begin, end), expected2);
       if (check1 !== undefined && check2 !== undefined) {
-        niceStack.message = check1 + check2;
+        niceStack.message = <span class="x x-first x-last">`Expected one of the following two checks to succeed:</span>
+- ${check1}
+- ${check2}`;
```
Richard-Yunchao, line 192: @kainino0x , do you know why this code snippet failed to build?

src/webgpu/gpu_test.ts
Richard-Yunchao: Done. I chose the first one.

src/webgpu/api/operation/memory_sync/buffer/ww.spec.ts
Richard-Yunchao: Agreed. I was meant to test same pass, no bundle, same pass, one bundle, same pass, two separate bundles. But I wrote it wrongly. And I ignored same bundle (definitely one pass). 
Now I added all of them. Note that I separated the tests two-draws-in-the-same-bundle from the original one. I tried to merge into the same one, but it requires unless and multiple ifs, making the code logic in a mess. So I think it is better to write a standalone test for same-bundle test.

src/webgpu/gpu_test.ts
```diff
@@ -167,6 +167,35 @@ export class GPUTest extends Fixture {
     });
   }
 
+  expectContentsMultipleValidValues(
+    src: GPUBuffer,
+    expected1: TypedArrayBufferView,
```
Kangz, line 172: If anything, I think this should take a `sequence<TypedArrayBufferView>`. Maybe we could also refactor the `expectContents` to take a validator function so it's easy to build additional variations in the future.

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
```diff
@@ -255,4 +255,12 @@ export class BufferSyncTest extends GPUTest {
     bufferData[0] = expectedValue;
     this.expectContents(buffer, bufferData);
   }
+
+  verifyDataMultipleValidValues(buffer: GPUBuffer, expectedValue1: number, expectedValue2: number) {
```
Kangz, line 259: nit: this could take a `sequence<number>`

Kangz (approved):
LGTM

src/webgpu/gpu_test.ts
Richard-Yunchao: Thanks for your suggestions, Corentin. I considered about this too but didn't do this because: 1) it is somehow over-engineering. 2) actually we have two directions to expand this function: 
- expand two values to an array (or a sequence), and the expected result should fall into any value of the array. it is just like any() function. This direction is what you said. 
- expand two values to two vectors, and the expected result should be a vector whose value at any index should be either  of the two values at the same index of the two vectors. Take (R, G, B, A) as an example, if two vectors are (0, 0, 0, 0) and (255, 255, 255, 255), then there are 16 correct values like (0, 0, 0, 0), (0, 0, 0, 255), ..., (0, 255, 255, 0), ..., (255, 255, 255, 255). This is another possible expansion. If we have multiple values in the buffer, a race condition of write after write is more likely to behave in this direction. But I don't want to make the test too complicated. 

So I didn't expand this function. 

How about we keep it and expand this function to whatever we want if we do need it in future?

src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
Richard-Yunchao: same reason

src/webgpu/gpu_test.ts
Kangz: While I think it is ok in this test fixture, the method I have concerns keeping the method in the harness as-is: it says 'multiple values" but only take two. In general the harness needs to be higher quality and more future proof than test code.

What we could do is do the refactor of `expectContents` to to have a version that takes a validator and make the validator check agains the two values in the fixture.

Richard-Yunchao: Yeah. I have concerns about the naming stuff too (I said about the naming problem at the very beginning). I hesitated to pick up a name like alternative, either, two, multiple, etc. Let's be clear and just say "two" valid values here. And we can expand it to multiple valid values or two mixed vectors if needed in future. I understand that the code should be future-proof but I don't think it is very clear that which direction is the future (maybe neither of them is needed). So I added a comment about it. WDYT?

src/webgpu/gpu_test.ts
kainino0x: I have no idea where the span stuff came from?? It must have been a GitHub bug.

src/webgpu/gpu_test.ts
kainino0x: The comment is inconsistent with the idea of passing a validator function, but it links to this discussion, so it's fine. I also considered suggesting passing an array here (instead of 2 values); it wouldn't really be over-engineering as it wouldn't be any more complex. However I think a validator function makes more sense in the future. A validator function pattern also works much better for texture content expectation. (I was thinking of passing a `function(x,y,z)->expected value` into expectTextureContents but a validator is more flexible.)

kainino0x (approved):

PR #386, opened 2020-11-24
Add dev server for standalone runner which compiles TS at runtime

austinEng:
- Custom http server which
  - compiles typescript sources at runtime using Babel to serve requests to `out/**/*.js`
  - crawls the filesystem to serve requests to `out/:suite/listing.js`
- Includes a ts compile and listing cache so subsequent requests don't take a long time
- Watches the filesystem to dirty the cache


dev/server
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env node
```
Kangz, line 1: Can we have something to run that automatically like `npx grunt dev` or similar and edit the `README.md` to mention it?

dev/server.ts
```diff
@@ -0,0 +1,149 @@
...
+    const tsUrl = path.sep + relPath;
+    if (compileCache.has(tsUrl)) {
+      console.debug('Dirtying compile cache', tsUrl);
+    }
+    compileCache.delete(tsUrl);
+  }
+}
+
+// Handler to dirty the listing cache for directory changes. Also dirties
+// the compile cache for changed files.
```
Kangz, line 47: nit: I think the listing only cares about the `.spec.ts` files so we can dirty the listing only in changes of such files. Also does the listing list the tests inside each `.spec.ts` or just the list of all the `.spec.ts`? If it's the former I think we also need to dirty the listing cache on `change`.

Kangz (approved):
LGTM although I'm not too familiar with node stuff.

dev/server
austinEng: It's available at `npm run dev` - I added this to the README.md as well.

dev/server.ts
austinEng: Indeed it only cares about dirs and .spec.ts. I added a condition for that.

It doesn't care about the contents of the `.spec.ts` except for the `description`. I think we're probably okay leaving that outdated..? Updating efficiently would be more involved since compiling the listing takes seconds.

Kangz: Ah yeah if the listing only cares about the description I agree that it's not worth forcing a listing recompile that takes seconds, and the description being outdated will be a small gotcha.

Kangz:
Still LGTM

dev/server.ts
kainino0x: I was thinking about taking the description out of the listing anyway. I am pretty sure it's not needed.

README.md
```diff
@@ -75,6 +75,20 @@ npm install
...
 
+### Dev Server
+
+To start the development server, use:
+
+```sh
+npm run dev
+```
+
+Then, browse to the standalone test runner at `http://127.0.0.1:8080/standalone/`. If this port is used, the server will print what port it is using to the console.
```
kainino0x, line 86:
```suggestion
Then, browse to the standalone test runner at `http://localhost:8080/standalone/`. If port 8080 isn't available, the server will print what port it is using to the console.
```

dev/server.ts
```diff
@@ -0,0 +1,159 @@
...
+    if (compileCache.has(tsUrl)) {
+      console.debug('Dirtying compile cache', tsUrl);
+    }
+    compileCache.delete(tsUrl);
+  }
+}
+
+/**
+ * Handler to dirty the listing cache for directory changes and .spec.ts changes.
+ * Also dirties the compile cache for changed files.
```
kainino0x, line 53: Ideally would also dirty on README files being added/removed/changed.

If we remove descriptions from the listing (which I think we can), then that should leave us with no staleness problem. (We could even remove README contents from the listing, and README changes wouldn't invalidate the listing, but that's not a big deal.)

dev/server
kainino0x, line 1: Maybe `tools/dev_server`?

dev/server.ts
```diff
@@ -0,0 +1,159 @@
+import * as fs from 'fs';
```
kainino0x, line 1: And `src/common/tools/dev_server.ts`

dev/server.ts
```diff
@@ -0,0 +1,159 @@
...
+const babelConfig = {
+  ...require(path.resolve(__dirname, '../babel.config.js'))({
+    cache: () => {
+      /* not used */
+    },
+  }),
+  sourceMaps: 'inline',
+};
+
+// Caches for the generated listing file and compiled TS sources to speed up reloads.
```
kainino0x, line 26: Comment what the key is here (request path?)

dev/server.ts
```diff
@@ -0,0 +1,159 @@
...
+watcher.on('unlinkDir', dirtyListingAndCompileCache);
+watcher.on('change', dirtyCompileCache);
+
+const app = express();
+
+// Set up logging
+app.use(morgan('dev'));
+
+// Serve the standalone runner directory
+app.use('/standalone', express.static(path.resolve(srcDir, '../standalone')));
```
kainino0x, line 85: `src/**/*.ts` (or all of `src/`) should also be served statically so that source maps work.

dev/server.ts
```diff
@@ -0,0 +1,159 @@
...
+
+  if (listingCache.has(suite)) {
+    res.setHeader('Content-Type', 'application/javascript');
+    res.send(listingCache.get(suite));
+    return;
+  }
+
+  try {
+    const listing = await makeListing(path.resolve(srcDir, suite, 'listing.ts'));
+    const result = `export const listing = Promise.resolve(${JSON.stringify(
```
kainino0x, line 99: nit: Promise technically not required; gen_listings doesn't generate it. `interface ListingFile` in `file_loader.ts` allows either.

dev/server.ts
```diff
@@ -0,0 +1,159 @@
...
+      res.send(result.code);
+    } else {
+      throw new Error(`Failed compile ${tsUrl}.`);
+    }
+  } catch (err) {
+    next(err);
+  }
+});
+
+const host = '0.0.0.0';
```
kainino0x, line 139: Should serve on 127.0.0.1 (`lo` only, rather than all interfaces), at least by default.

dev/server.ts
austinEng: Any more comments Kai, or shall we land this? :)

dev/server.ts
```diff
@@ -0,0 +1,159 @@
...
+
+import * as babel from '@babel/core';
+import * as chokidar from 'chokidar';
+import * as express from 'express';
+import * as morgan from 'morgan';
+import * as portfinder from 'portfinder';
+
+import { makeListing } from '../src/common/tools/crawl.js';
+
+// Make sure that makeListing doesn't cache imported spec files.
```
kainino0x, line 13:
```suggestion
// Make sure that makeListing doesn't cache imported spec files. See crawl().
```

kainino0x (commented):
I was just reviewing it. :)

dev/server.ts
austinEng: Hm? Not sure what you mean here

austinEng: Oh, the file name, ok

dev/server.ts
austinEng: but.. why not? I believe this is the default behavior of `http-server` now and it's super nice for cross-device development.

dev/server.ts
austinEng: I'm using inline source maps which doesn't require this. I could make them not inline, but then I need to write the .map files to a tmp directory or have a cache to answer .map requests

dev/server
austinEng: Done.

dev/server.ts
austinEng: Done.

dev/server.ts
austinEng: Done.

dev/server.ts
austinEng: Done.

dev/server.ts
austinEng: Done.

dev/server.ts
kainino0x: I guess it's fine since it's part of this project so you know all you're serving up is the stuff in the project dir. Usually I default to serving locally so I don't have to worry about exposing anything (or triggering the port scraper bot on our internal network)

kainino0x (approved):

dev/server.ts
kainino0x: Ah, didn't realize. Sounds good.

PR #390, opened 2020-12-02
From HackMD plan, part 3: most of the rest of the validation tests

kainino0x:
This one's more messy because I really need to get this done, so I didn't clean up (look for duplicates, organize, etc.) as much. As usual not super necessary to review the plans because we'll look at them again later when doing the TODOs, but if you notice anything lmk.

### DIFF
--- a/src/webgpu/api/operation/compute_pipeline/entry_point_name.spec.ts
+++ b/src/webgpu/api/operation/compute_pipeline/entry_point_name.spec.ts
@@ -2,6 +2,8 @@ export const description = `
 TODO:
 - Test some weird but valid values for entry point name (both module and pipeline creation
   should succeed).
+- Test using each of many entry points in the module (should succeed).
+- Test using an entry point with the wrong stage (should fail).
 `;
 
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
--- a/src/webgpu/api/operation/compute_pipeline/entry_point_name.spec.ts
+++ b/src/webgpu/api/operation/compute_pipeline/entry_point_name.spec.ts
@@ -1,0 +1,10 @@
+export const description = `
+TODO:
+- Test some weird but valid values for entry point name (both module and pipeline creation
+  should succeed).
+`;
+
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { GPUTest } from '../../../gpu_test.js';
+
+export const g = makeTestGroup(GPUTest);
--- a/src/webgpu/api/operation/render_pipeline/entry_point_name.spec.ts
+++ b/src/webgpu/api/operation/render_pipeline/entry_point_name.spec.ts
@@ -2,6 +2,8 @@ export const description = `
 TODO:
 - Test some weird but valid values for entry point name (both module and pipeline creation
   should succeed).
+- Test using each of many entry points in the module (should succeed).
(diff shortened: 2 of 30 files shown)


kainino0x:
Organizing this is really hard. I keep forgetting where stuff is, not knowing whether we already have something implemented, forgetting that we already have a plan when someone writes new plan/tests, etc.

src/webgpu/api/operation/compute_pipeline/entry_point_name.spec.ts
```diff
@@ -0,0 +1,10 @@
+export const description = `
+TODO:
+- Test some weird but valid values for entry point name (both module and pipeline creation
```
Kangz, line 3: nit: also test using one of many entrypoints in the shadermodule, and creating on compute pipeline from each entrypoint in this large shadermodule (and using them?)

src/webgpu/api/operation/render_pipeline/entry_point_name.spec.ts
```diff
@@ -0,0 +1,10 @@
+export const description = `
+TODO:
+- Test some weird but valid values for entry point name (both module and pipeline creation
+  should succeed).
```
Kangz, line 4: ditto

src/webgpu/api/operation/sampling/anisotropy.spec.ts
```diff
@@ -0,0 +1,15 @@
+export const description = `
+Tests the behavior of anisotropic filtering.
+
+Note that anisotropic filtering is never guaranteed to occur, but we might be able to test that
+no *more* than the provided maxAnisotropy samples are used, by testing how many unique
+sample values come out of the sample operation.
```
Kangz, line 6: Or have a test that check the result with and without anisotropy are different, or that aniso is done in the correct direction (by having a 2D gradient and checking we get more of the color in the correct direction). There's a bunch of implementation dependent behavior we'll still want to test, so maybe it is worth having some sort of soft expectations or non-failing expectations for them?

src/webgpu/api/validation/buffer/map_state.spec.ts
```diff
@@ -0,0 +1,8 @@
+export const description = `
+Validation tests for GPUBuffer.mapAsync, GPUBuffer.unmap and GPUBuffer.getMappedRange.
```
Kangz, line 2: These are already done in mapping.spec.ts

src/webgpu/api/validation/create_pipeline.ts
```diff
@@ -0,0 +1,15 @@
+export const description = `
```
Kangz, line 1: nit: This file is `create_pipeline` while the other one is `createPipeline`. What's the convention?

Kangz (approved):

src/webgpu/api/validation/create_pipeline.ts
kainino0x: camelCase is named after a function in the API, but createPipeline is not one (it's createRenderPipeline + createComputePipeline)

src/webgpu/api/validation/buffer/map_state.spec.ts
kainino0x: oops, accidentally duplicated

PR #394, opened 2020-12-03
organize pass state tests a bit

kainino0x:


src/webgpu/api/operation/command_buffer/render/state_tracking.spec.ts
```diff
@@ -1,17 +1,13 @@
...
 TODO: plan and implement
-- {viewport, scissor rect, blend color, stencil reference, setIndexBuffer, setVertexBuffer}: test rendering result with:
-    - state {unset (= default), explicitly set default value, another value}
-    - persistence: [set, draw, draw] (fn should differentiate from [set, draw] + [draw])
-    - overwriting: [set(1), draw, set(2), draw] (fn should differentiate from [set(1), set(2), draw, draw])
-    - overwriting: [set(1), set(2), draw] (fn should differentiate from [set(1), draw] but not [set(2), draw])
-- setIndexBuffer: specifically test changing the format, offset, size, without changing the buffer
-- setVertexBuffer: specifically test changing the offset, size, without changing the buffer
+- try setting states multiple times in different orders, check state is correct in a draw call.
+    - setIndexBuffer: specifically test changing the format, offset, size, without changing the buffer
```
Kangz, line 9: nit: we should also take into account changing the pipeline before or after the setVertexBuffer. In D3D12 the VB stride is part of the SetVertexBuffer call and not the pipeline.

src/webgpu/api/validation/encoding/programmable/pipeline_bind_group_compat.spec.ts
```diff
@@ -0,0 +1,15 @@
+export const description = `
+TODO:
+- test compatibility between bind groups and pipelines
```
Kangz, line 3: Should we also test inheritance between pipelines with different layouts for 0 or for the last bindgroup? Also what happens when a pipeline has an empty bindgroup for group 1 but not group 2 and we never set anything to group 1 for example.

Is it possible to unset bindgroups?

Kangz (approved):

src/webgpu/api/validation/encoding/programmable/pipeline_bind_group_compat.spec.ts
kainino0x: It's not possible to unset bind groups. If they go unused by a pipeline that's supposed to be OK.

kainino0x: > Should we also test inheritance between pipelines with different layouts for 0 or for the last bindgroup? Also what happens when a pipeline has an empty bindgroup for group 1 but not group 2 and we never set anything to group 1 for example.

Added to operation because I think these wouldn't be invalid in validation.

PR #399, opened 2020-12-04
Update and organize documentation for contributors

kainino0x:
View here:
https://github.com/kainino0x/cts/blob/docs/README.md
https://github.com/kainino0x/cts/tree/docs/docs


Kangz (approved):
It looks great! I'm not sure about the need for the docs/more directory but don't feel strongly.

kainino0x:
Wanted to keep it separate from the 4 docs that are part of the "intro". I put those in a folder and moved more/* out.

PR #400, opened 2020-12-04
Rename texture_zero.ts -> texture_zero.spec.ts

austinEng:
Realized we lost the .spec.ts extension so this test disappeared.


kainino0x:
I noticed this last night too and just rolled it into a PR I already had open. https://github.com/gpuweb/cts/pull/394

PR #401, opened 2020-12-04
Replace pull_request_target with pull_request+workflow_run

mehmetoguzderin:
For improving security

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -1,0 +1,71 @@
+name: Workflow CI
+
+on:
+  workflow_run:
+    workflows:
+      - "Pull Request CI"
+    types:
+      - completed
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2.3.1
+        with:
+          persist-credentials: false
+      - run: |
+          PR=$(curl https://api.github.com/search/issues?q=${{ github.event.workflow_run.head_sha }} |
+          grep -Po "(?<=${{ github.event.workflow_run.repository.full_name }}\/pulls\/)\d*" | head -1)
+          echo "PR=$PR" >> $GITHUB_ENV
+      - run: |
+          git fetch origin ${{ github.event.workflow_run.head_sha }}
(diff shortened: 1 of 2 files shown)


.github/workflows/workflow.yml
```diff
@@ -0,0 +1,71 @@
...
+          git fetch origin ${{ github.event.workflow_run.head_sha }}
+          git checkout ${{ github.event.workflow_run.head_sha }}
+      - uses: actions/setup-node@v2-beta
+        with:
+          node-version: "15.x"
+      - run: npm ci
+      - run: |
+          npm test
+          sed -i '/out\//d' .gitignore
+          sed -i '/out-wpt\//d' .gitignore
```
kainino0x, line 31: unrelated: I think these two lines are obsolete now that we don't upload by git push to gh-pages?

.github/workflows/workflow.yml
```diff
@@ -0,0 +1,71 @@
...
+      - uses: actions/checkout@v2.3.1
+        with:
+          persist-credentials: false
+      - run: |
+          PR=$(curl https://api.github.com/search/issues?q=${{ github.event.workflow_run.head_sha }} |
+          grep -Po "(?<=${{ github.event.workflow_run.repository.full_name }}\/pulls\/)\d*" | head -1)
+          echo "PR=$PR" >> $GITHUB_ENV
+      - run: |
+          git fetch origin ${{ github.event.workflow_run.head_sha }}
+          git checkout ${{ github.event.workflow_run.head_sha }}
```
kainino0x, line 23: Doesn't this job still run code (`npm test`) from an untrusted head_sha, with the secrets in scope? Not sure if I'm right about that.

Is there any way to get the build results from "Pull Request CI", which were built without secrets in scope, and upload them from this workflow which does have secrets in scope?

mehmetoguzderin: @kainino0x right that's possible, I will update this PR to use the build from base Action

mehmetoguzderin:
@kainino0x I have updated the code to make use of build artifacts, could you take a look?

kainino0x (approved):

PR #406, opened 2020-12-07
Test CI

mehmetoguzderin:


kainino0x:
Looks like it's working to me! (Took a while to load, but that's just because there are a ton of files. I guess Firebase is not serving them up super fast.)

PR #409, opened 2020-12-09
Add helper to make capability tables prettier

### LINKED ISSUE #354: Condense capability info tables to be more readable/authorable
_Originally posted by @kainino0x in https://github.com/gpuweb/cts/pull/352#discussion_r523300691_

---

I don't think it makes sense to split it here in such a way that it affects other files, just for the sake of making this file more editable. I think it's okay if this file is a little special (have to zoom out your editor maybe).

Maybe we could come up with a little table-making helper thing so we don't have to repeat "renderable:" "color:" etc on every line?

---

`componentType` can be computed from `dataType` right? So we could separate that into a separate table. Or use `, ...kDataTypeUnorm` where `const kDataTypeUnorm = { dataType: 'unorm', componentType: 'float' };`.

Let's consider both of these as followups.

kainino0x:
Fixes #354

-----

- [x] WebGPU readability
- [x] TypeScript readability

### DIFF
--- a/src/common/framework/util/types.ts
+++ b/src/common/framework/util/types.ts
@@ -24,9 +24,6 @@ type TypeOr<T, Default> = T extends undefined ? Default : T;
 /**
  * Zips a key tuple type and a value tuple type together into an object.
  *
- * As of this writing, VSCode bundles an older version of TypeScript which cannot support this.
- * Use "TypeScript: Select TypeScript Version..." in the command palette to switch.
- *
  * @template K Keys of the resulting object.
  * @template V Values of the resulting object. If an item is `undefined` or past the end, it defaults.
  * @template D Default values. If an item is past the end, it defaults to `undefined`.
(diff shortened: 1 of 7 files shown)


kainino0x:
VSCode has just updated so now I think this is actually good to land. PTAL

austinEng:
Sorry I forgot about this; I'll review now.

src/common/framework/util/types.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Computes the intersection of a set of types, given the union of those types.
+ *
+ * From: https://stackoverflow.com/a/56375136
+ */
+export type UnionToIntersection<U> =
+  /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
+  (U extends any ? (k: U) => void : never) extends (k: infer I) => void ? I : never;
+
+/** "Type asserts" that `X` is a subtype of `Y`. */
+type Cast<X, Y> = X extends Y ? X : never;
```
austinEng, line 18: maybe a better name is "EnsureExtends" or "EnsureSubtype" ? "Cast" is odd because the input X comes back as X still.

austinEng (approved):
One naming nit;
WebGPU readability N/A; Typescript readability ok.

PR #410, opened 2020-12-09
Move the tests about Z-axis or depth or zero-copy into separated tests

Jiawei-Shao:
This patch moves the test configurations about Z-axis or depth from
't2t_non_compressed_color_formats' into a separated test
't2t_2d_array_non_compressed_color_textures' so that they will no
longer be used together with the 2D non-array textures and generate
meaningless tests (previously when the copy depth is out of range
the copyExtent.depth will be set to 0).

### DIFF
--- a/src/webgpu/api/operation/command_buffer/copyTextureToTexture.spec.ts
+++ b/src/webgpu/api/operation/command_buffer/copyTextureToTexture.spec.ts
@@ -1,4 +1,7 @@
-export const description = 'copyTexturetoTexture operation tests';
+export const description = `copyTexturetoTexture operation tests
+
+  TODO(jiawei.shao@intel.com): support all WebGPU texture formats.
+  `;
 
 import { poptions, params } from '../../../../common/framework/params_builder.js';
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
@@ -250,7 +253,6 @@ class F extends GPUTest {
 
 export const g = makeTestGroup(F);
 
-// TODO(jiawei.shao@intel.com): support all WebGPU texture formats
 g.test('color_textures,non_compressed,non_array')
   .desc(
     `
@@ -306,7 +308,6 @@ g.test('color_textures,non_compressed,non_array')
     );
   });
 
-// TODO(jiawei.shao@intel.com): support all WebGPU texture formats
 g.test('color_textures,non_compressed,array')
   .desc(
     `
--- a/src/webgpu/api/operation/command_buffer/copyTextureToTexture.spec.ts
+++ b/src/webgpu/api/operation/command_buffer/copyTextureToTexture.spec.ts
@@ -87,27 +87,41 @@ class F extends GPUTest {
     // Copy the region specified by copyBoxOffsets from srcTexture to dstTexture.
     const dstTextureSizeAtLevel = physicalMipSize(dstTextureSize, format, '2d', dstCopyLevel);
     const minWidth = Math.min(srcTextureSizeAtLevel.width, dstTextureSizeAtLevel.width);
-    const appliedCopyWidth =
-      minWidth +
-      copyBoxOffsets.copyExtent.width -
-      Math.max(copyBoxOffsets.srcOffset.x, copyBoxOffsets.dstOffset.x);
     const minHeight = Math.min(srcTextureSizeAtLevel.height, dstTextureSizeAtLevel.height);
-    const appliedCopyHeight =
+
+    const appliedSrcOffset = {
+      x: Math.min(copyBoxOffsets.srcOffset.x, minWidth),
+      y: Math.min(copyBoxOffsets.srcOffset.y, minHeight),
+      z: copyBoxOffsets.srcOffset.z,
+    };
+    const appliedDstOffset = {
+      x: Math.min(copyBoxOffsets.dstOffset.x, minWidth),
+      y: Math.min(copyBoxOffsets.dstOffset.y, minHeight),
+      z: copyBoxOffsets.dstOffset.z,
+    };
+
+    const appliedCopyWidth = Math.max(
+      minWidth + copyBoxOffsets.copyExtent.width - Math.max(appliedSrcOffset.x, appliedDstOffset.x),
+      0
+    );
+    const appliedCopyHeight = Math.max(
       minHeight +
-      copyBoxOffsets.copyExtent.height -
-      Math.max(copyBoxOffsets.srcOffset.y, copyBoxOffsets.dstOffset.y);
+        copyBoxOffsets.copyExtent.height -
+        Math.max(appliedSrcOffset.y, appliedDstOffset.y),
+      0
+    );
     assert(appliedCopyWidth % blockWidth === 0 && appliedCopyHeight % blockHeight === 0);
 
     const appliedCopyDepth =
       srcTextureSize.depth +
       copyBoxOffsets.copyExtent.depth -
-      Math.max(copyBoxOffsets.srcOffset.z, copyBoxOffsets.dstOffset.z);
+      Math.max(appliedSrcOffset.z, appliedDstOffset.z);
     assert(appliedCopyDepth >= 0);
 
     const encoder = this.device.createCommandEncoder();
     encoder.copyTextureToTexture(
-      { texture: srcTexture, mipLevel: srcCopyLevel, origin: copyBoxOffsets.srcOffset },
-      { texture: dstTexture, mipLevel: dstCopyLevel, origin: copyBoxOffsets.dstOffset },
+      { texture: srcTexture, mipLevel: srcCopyLevel, origin: appliedSrcOffset },
+      { texture: dstTexture, mipLevel: dstCopyLevel, origin: appliedDstOffset },
       { width: appliedCopyWidth, height: appliedCopyHeight, depth: appliedCopyDepth }
     );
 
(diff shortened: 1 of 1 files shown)


Kangz:
Note that testing noop copies is a valid test and something we should probably do anyway. That's why I'm not sure it's necessary to split these tests. WDYT?

Jiawei-Shao:
Hi @Kangz , I agree that we should test no-op copies, but I don't think it necessary to test it so many times. In the current code actually all the test configurations on Z-component are actually testing no-op copie when the texture is 2D non-array texture, which I think is a bit too many (we have 4 2D non-array texture configurations but only 1 2D array texture configuration right now, which means 4/5 of the whole tests on Z-component are testing no-op copies). WDYT?

Jiawei-Shao:
Maybe we can put all no-op copy tests in one test.

src/webgpu/api/operation/command_buffer/copyTextureToTexture.spec.ts
```diff
@@ -35,12 +35,161 @@ class F extends GPUTest {
...
+    }
+
+    // Verify the content of the whole subresouce of dstTexture at dstCopyLevel (in dstBuffer) is expected.
+    this.expectContents(dstBuffer, expectedUint8DataWithPadding);
+  }
 }
 
 export const g = makeTestGroup(F);
 
 // TODO(jiawei.shao@intel.com): support all WebGPU texture formats
```
kainino0x, line 191: nit (new guidance): move this TODO inside the description so it's visible in /standalone/. Also make it just a little more descriptive.

src/webgpu/api/operation/command_buffer/copyTextureToTexture.spec.ts
```diff
@@ -75,9 +223,92 @@ g.test('t2t_non_compressed_color_formats')
...
+      textureSize.srcTextureSize,
+      textureSize.dstTextureSize,
+      format,
+      copyBoxOffsets,
+      srcCopyLevel,
+      dstCopyLevel
+    );
+  });
+
+// TODO(jiawei.shao@intel.com): support all WebGPU texture formats
```
kainino0x, line 293: Same

src/webgpu/api/operation/command_buffer/copyTextureToTexture.spec.ts
```diff
@@ -75,9 +223,92 @@ g.test('t2t_non_compressed_color_formats')
...
+g.test('t2t_2d_array_non_compressed_color_textures')
```
kainino0x, line 294: nit: name these two so they are siblings in the tree. e.g.
`color_textures,non_compressed,array`
`color_textures,non_compressed,non_array`
(And can remove "t2t" because it's redundant with the filename.)

src/webgpu/api/operation/command_buffer/copyTextureToTexture.spec.ts
```diff
@@ -75,9 +223,92 @@ g.test('t2t_non_compressed_color_formats')
             srcTextureSize: { width: 32, height: 32, depth: 1 },
             dstTextureSize: { width: 63, height: 61, depth: 1 },
           },
+        ])
+      )
+      .combine(poptions('format', kRegularTextureFormats))
+      .combine(
+        poptions('copyBoxOffsets', [
```
kainino0x, line 230: Most of this list is a subset of the other, right? I think you should factor it out into a `const` and then use `...copyBoxOffsetsForWholeDepth,` in both.

src/webgpu/api/operation/command_buffer/copyTextureToTexture.spec.ts
```diff
@@ -75,9 +223,92 @@ g.test('t2t_non_compressed_color_formats')
...
+            dstOffset: { x: 0, y: 0, z: 0 },
+            copyExtent: { width: -1, height: 0, depth: 0 },
+          },
+          // From (0, 1) of src to (0, 0) of dst, and the copy extent will not cover the last
+          // row of both source and destination texture.
+          {
+            srcOffset: { x: 0, y: 1, z: 0 },
+            dstOffset: { x: 0, y: 0, z: 0 },
+            copyExtent: { width: 0, height: -1, depth: 0 },
+          },
```
kainino0x, line 274: Add no-op copy cases to this list (that are not factored out into the common cases)?

src/webgpu/api/operation/command_buffer/copyTextureToTexture.spec.ts
```diff
@@ -86,7 +317,7 @@ g.test('t2t_non_compressed_color_formats')
         poptions('copyBoxOffsets', [
```
kainino0x, line 317: I don't think there are any empty-copy cases in here either?

kainino0x (commented):
I think this split is reasonable, but would tweak it slightly.

src/webgpu/api/operation/command_buffer/copyTextureToTexture.spec.ts
Jiawei-Shao: I've put all the zero-copy cases into one test group "zero_copies", PTAL.

src/webgpu/api/operation/command_buffer/copyTextureToTexture.spec.ts
Jiawei-Shao: I've put all the zero-copy cases into one test group "zero_copies", PTAL.

src/webgpu/api/operation/command_buffer/copyTextureToTexture.spec.ts
```diff
@@ -184,16 +199,63 @@ class F extends GPUTest {
...
+      copyExtent: { width: 0, height: -1, depth: 0 },
+    },
+  ];
 }
 
 export const g = makeTestGroup(F);
 
-// TODO(jiawei.shao@intel.com): support all WebGPU texture formats
-g.test('t2t_non_array_non_compressed_color_formats')
+g.test('color_textures_non_compressed_non_array')
```
kainino0x, line 253: Name should use commas to separate parts of the test "path" (like slashes in a file path, but we use comma, at least right now) e.g.
```suggestion
g.test('color_textures,non_compressed,non_array')
```

src/webgpu/api/operation/command_buffer/copyTextureToTexture.spec.ts
```diff
@@ -422,3 +394,87 @@ supported by WebGPU, doing CopyTextureToTexture() copy, and verifying the conten
...
+            dstOffset: { x: 64, y: 0, z: 0 },
+            copyExtent: { width: -64, height: 0, depth: 0 },
+          },
+          // copyExtent.height === 0
+          {
+            srcOffset: { x: 0, y: 0, z: 0 },
+            dstOffset: { x: 0, y: 0, z: 0 },
+            copyExtent: { width: 0, height: -32, depth: 0 },
+          },
+          // copyExtent.height === 0 && srcOffset.y === textureHeight - 1
```
kainino0x, line 431: Should be `=== textureHeight`, I think? Same in 3 places below.

src/webgpu/api/operation/command_buffer/copyTextureToTexture.spec.ts
```diff
@@ -325,5 +197,284 @@ g.test('t2t_non_compressed_color_formats')
...
+      format,
+      copyBoxOffsets,
+      srcCopyLevel,
+      dstCopyLevel
+    );
+  });
+
+g.test('zero_copies')
+  .desc(
+    `Validate the correctness of the functionality of CopyTextureToTexture() with zero copies
```
kainino0x, line 400:
```suggestion
g.test('zero_sized')
  .desc(
    `Validate the correctness of zero-sized copies (should be no-ops).

- Copies that are zero-sized in only one dimension {x, y, z}, each touching the {lower, upper} end of that dimension.
```

src/webgpu/api/operation/command_buffer/copyTextureToTexture.spec.ts
```diff
@@ -325,5 +197,284 @@ g.test('t2t_non_compressed_color_formats')
...
+      textureSize.srcTextureSize,
+      textureSize.dstTextureSize,
+      format,
+      copyBoxOffsets,
+      srcCopyLevel,
+      dstCopyLevel
+    );
+  });
+
+g.test('color_textures_non_compressed_array')
```
kainino0x, line 308:
```suggestion
g.test('color_textures,non_compressed,array')
```

src/webgpu/api/operation/command_buffer/copyTextureToTexture.spec.ts
```diff
@@ -325,5 +197,284 @@ g.test('t2t_non_compressed_color_formats')
...
+}
+
+export const g = makeTestGroup(F);
+
+g.test('color_textures_non_compressed_non_array')
+  .desc(
+    `
+  Validate the correctness of the copy by filling the srcTexture with testable data and any non-compressed color format that is
+  supported by WebGPU, doing CopyTextureToTexture() copy, and verifying the content of the whole dstTexture
+  TODO(jiawei.shao@intel.com): support all WebGPU texture formats
```
kainino0x, line 258: nit (sorry, missed this before): Having the TODO here implies we'll add it to this test. If you think we'll add it to this test, update the test name to not include "non_compressed"; if you think it'll be a separate test, move this TODO into the file description. Same below.

kainino0x (approved):

kainino0x:
Thanks!

PR #411, opened 2020-12-10
Add attachment compatibility tests

egalli:
-----

- [x] WebGPU readability
- [x] TypeScript readability

### DIFF
--- a/src/webgpu/api/validation/attachment_compatibility.spec.ts
+++ b/src/webgpu/api/validation/attachment_compatibility.spec.ts
@@ -1,0 +1,305 @@
+export const description = `
+Validation for attachment compatibility between render passes, bundles, and pipelines
+
+TODO: Add sparse color attachment compatibility test when defined by specification
+`;
+
+import { poptions, params } from '../../../common/framework/params_builder.js';
+import { makeTestGroup } from '../../../common/framework/test_group.js';
+import { range } from '../../../common/framework/util/util.js';
+import {
+  kRegularTextureFormatInfo,
+  kRegularTextureFormats,
+  kSizedDepthStencilFormats,
+  kUnsizedDepthStencilFormats,
+  RegularTextureFormat,
+} from '../../capability_info.js';
+
+import { ValidationTest, CommandBufferMaker } from './validation_test.js';
+
+// TODO: Update maximum color attachments when defined
+const kColorAttachmentCounts = range(4, i => i + 1);
+const kDepthStencilAttachmentFormats = [
+  undefined,
+  ...kSizedDepthStencilFormats,
+  ...kUnsizedDepthStencilFormats,
+] as const;
+
+class F extends ValidationTest {
+  createAttachmentTextureView(format: GPUTextureFormat) {
+    return this.device
+      .createTexture({
+        size: [1, 1, 1],
+        format,
+        usage: GPUTextureUsage.OUTPUT_ATTACHMENT,
+      })
+      .createView();
+  }
+
+  createColorAttachment(format: GPUTextureFormat): GPURenderPassColorAttachmentDescriptor {
+    return {
+      attachment: this.createAttachmentTextureView(format),
+      loadValue: [0, 0, 0, 0],
+    };
+  }
+
+  createDepthAttachment(format: GPUTextureFormat): GPURenderPassDepthStencilAttachmentDescriptor {
+    return {
+      attachment: this.createAttachmentTextureView(format),
+      depthLoadValue: 0,
+      depthStoreOp: 'clear',
+      stencilLoadValue: 1,
+      stencilStoreOp: 'clear',
+    };
+  }
+
+  createPassOrBundleEncoder(
+    encoderType: 'render pass' | 'render bundle',
+    colorFormats: Iterable<GPUTextureFormat>,
+    depthStencilFormat?: GPUTextureFormat
+  ): CommandBufferMaker<GPURenderPassEncoder | GPURenderBundleEncoder> {
+    const encoder = this.device.createCommandEncoder();
+    const passDesc: GPURenderPassDescriptor = {
+      colorAttachments: Array.from(colorFormats, format => this.createColorAttachment(format)),
+      depthStencilAttachment:
+        depthStencilFormat !== undefined
+          ? this.createDepthAttachment(depthStencilFormat)
+          : undefined,
+    };
+    const pass = encoder.beginRenderPass(passDesc);
+    switch (encoderType) {
+      case 'render bundle': {
+        const bundleEncoder = this.device.createRenderBundleEncoder({
+          colorFormats,
+          depthStencilFormat,
+        });
+
+        return {
+          encoder: bundleEncoder,
+          finish() {
+            const bundle = bundleEncoder.finish();
+            pass.executeBundles([bundle]);
+            pass.endPass();
+            return encoder.finish();
+          },
+        };
+      }
+      case 'render pass':
+        return {
+          encoder: pass,
+          finish() {
+            pass.endPass();
+            return encoder.finish();
+          },
+        };
+    }
+  }
+
+  createRenderPipeline(
+    colorStates: Iterable<GPUColorStateDescriptor>,
+    depthStencilState?: GPUDepthStencilStateDescriptor
+  ) {
+    const wgslVertex = `
+    [[builtin(position)]] var<out> position : vec4<f32>;
+
+    [[stage(vertex)]]
+    fn main() -> void {
+      position = vec4<f32>(0.0, 0.0, 0.0, 0.0);
+      return;
+    }
+  `;
+    const wgslFragment = `
+    [[stage(fragment)]]
+    fn main() -> void {
+      return;
+    }
+  `;
+
+    return this.device.createRenderPipeline({
+      vertexStage: {
+        module: this.device.createShaderModule({
+          code: wgslVertex,
+        }),
+        entryPoint: 'main',
+      },
+      fragmentStage: {
+        module: this.device.createShaderModule({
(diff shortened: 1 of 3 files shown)


src/webgpu/api/validation/attachment_compatibility.spec.ts
```diff
@@ -0,0 +1,305 @@
...
+    colorStates: Iterable<GPUColorStateDescriptor>,
+    depthStencilState?: GPUDepthStencilStateDescriptor
+  ) {
+    const wgslVertex = `
+    [[builtin(position)]] var<out> position : vec4<f32>;
+
+    [[stage(vertex)]]
+    fn main() -> void {
+      position = vec4<f32>(0.0, 0.0, 0.0, 0.0);
+      return;
```
Kangz, line 108: nit: the return is now optional! ditto below

src/webgpu/api/validation/attachment_compatibility.spec.ts
```diff
@@ -0,0 +1,305 @@
...
+
+  createRenderPipeline(
+    colorStates: Iterable<GPUColorStateDescriptor>,
+    depthStencilState?: GPUDepthStencilStateDescriptor
+  ) {
+    const wgslVertex = `
+    [[builtin(position)]] var<out> position : vec4<f32>;
+
+    [[stage(vertex)]]
+    fn main() -> void {
```
Kangz, line 106: nit: we try to have a convention of putting the `[[stage]]` and `fn main` on the same line.

src/webgpu/api/validation/attachment_compatibility.spec.ts
```diff
@@ -0,0 +1,305 @@
...
+          },
+        };
+    }
+  }
+
+  createRenderPipeline(
+    colorStates: Iterable<GPUColorStateDescriptor>,
+    depthStencilState?: GPUDepthStencilStateDescriptor
+  ) {
+    const wgslVertex = `
```
Kangz, line 102: nit: You can put that shader code inline in the pipeline descriptor

src/webgpu/api/validation/attachment_compatibility.spec.ts
```diff
@@ -0,0 +1,305 @@
...
+    });
+    pass.executeBundles([bundle]);
+    pass.endPass();
+    t.expectValidationError(() => {
+      t.queue.submit([encoder.finish()]);
+    }, passFormat !== bundleFormat);
+  });
+
+g.test('render_pass_and_bundle_color_count')
+  .desc('Test that the number of color attachments in render passes and bundles must match.')
```
Kangz, line 173: nit: the TODO about sparse attachments could be here too.

src/webgpu/api/validation/attachment_compatibility.spec.ts
```diff
@@ -0,0 +1,305 @@
...
+export const g = makeTestGroup(F);
+
+const kColorAttachmentFormats = kRegularTextureFormats
+  .map((key): [RegularTextureFormat, typeof kRegularTextureFormatInfo[RegularTextureFormat]] => [
+    key,
+    kRegularTextureFormatInfo[key],
+  ])
+  .filter(([, format]) => format.color && format.renderable)
+  .map(([key]) => key);
+
```
Kangz, line 147: We should also test that the sampleCount matches.

src/webgpu/api/validation/attachment_compatibility.spec.ts
Kangz: Ditto below

src/webgpu/api/validation/attachment_compatibility.spec.ts
```diff
@@ -0,0 +1,305 @@
...
+  .desc('Test that the number of color attachments in render passes and bundles must match.')
+  .params(
+    params()
+      .combine(poptions('passCount', kColorAttachmentCounts))
+      .combine(poptions('bundleCount', kColorAttachmentCounts))
+  )
+  .fn(t => {
+    const { passCount, bundleCount } = t.params;
+    const bundleEncoder = t.device.createRenderBundleEncoder({
+      colorFormats: range(bundleCount, () => 'rgba8unorm'),
```
Kangz, line 182: It would be nice to test varying formats in postitions other than 0 too, but maybe that can be done with the sparse attachments.

Kangz: Ditto below

Kangz (commented):
Looks good overall

src/webgpu/api/validation/attachment_compatibility.spec.ts
```diff
@@ -0,0 +1,305 @@
...
+      range(encoderCount, () => 'rgba8unorm')
+    );
+    encoder.setPipeline(pipeline);
+
+    t.expectValidationError(() => {
+      t.queue.submit([finish()]);
+    }, encoderCount !== pipelineCount);
+  });
+
+g.test('render_pass_and_pipeline_depth_format')
```
kainino0x, line 276: `render_pass_or_bundle_and_`

src/webgpu/api/validation/attachment_compatibility.spec.ts
kainino0x: +1, can be a TODO and followup, if you want (because this PR is almost done)

kainino0x (approved):

src/webgpu/api/validation/attachment_compatibility.spec.ts
egalli: I'm concerned about the test case count explosion. `kColorAttachmentFormats` currently has 31 formats. The `color_format` tests currently create 961 and 1922 test cases (`render_pass_and_bundle` and  `render_pass_or_bundle_and_pipeline` respectively). Combining `color_format` and `color_count` would create a combined tests with 15376 and 30752 tests cases.

Not sure that sparse attachments would create any less test cases.

src/webgpu/api/validation/attachment_compatibility.spec.ts
egalli: Done

src/webgpu/api/validation/attachment_compatibility.spec.ts
```diff
@@ -303,3 +337,36 @@ Test that the depth attachment format in render passes or bundles match the pipe
...
+  .desc(
+    `
+Test that the sample count in render passes or bundles match the pipeline sample count.
+`
+  )
+  .params(
+    params()
+      .combine(poptions('encoderType', ['render pass', 'render bundle'] as const))
+      .combine(poptions('encoderSampleCount', [1, 4] as const))
+      .combine(poptions('pipelineSampleCount', [1, 4] as const))
```
kainino0x, line 351: nit: kSampleCounts

src/webgpu/api/validation/attachment_compatibility.spec.ts
```diff
@@ -12,11 +12,13 @@ import {
   kRegularTextureFormats,
   kSizedDepthStencilFormats,
   kUnsizedDepthStencilFormats,
-  RegularTextureFormat,
 } from '../../capability_info.js';
 
 import { ValidationTest, CommandBufferMaker } from './validation_test.js';
 
+// TODO: Update with all possible sample counts when defined
+const kSampleCounts = [1, 4] as const;
```
kainino0x, line 20: nit: Could you put this definition in capability_info.ts? We probably already have other tests that hardcode 1 and 4, but putting it there helps keep track of it for optional capabilities, like sample count 8, in the future. When you do that, please also add a TODO to find other tests for sample counts and use kSampleCounts in them.

kainino0x (approved):
Sorry, more tiny nits :)

src/webgpu/api/validation/attachment_compatibility.spec.ts
```diff
@@ -0,0 +1,305 @@
...
+  kRegularTextureFormats,
+  kSizedDepthStencilFormats,
+  kUnsizedDepthStencilFormats,
+  RegularTextureFormat,
+} from '../../capability_info.js';
+
+import { ValidationTest, CommandBufferMaker } from './validation_test.js';
+
+// TODO: Update maximum color attachments when defined
+const kColorAttachmentCounts = range(4, i => i + 1);
```
kainino0x, line 23: I guess this should be in capability_info too

src/webgpu/api/validation/attachment_compatibility.spec.ts
kainino0x: I don't think we need to test every combination, just some. So e.g. we could just test:

- pass [rgba8unorm, rgba8unorm, rgba8unorm, ...] vs bundle with same array except one element changed to bgra8unorm or something (+ `kMaxColorAttachments` cases)
- pass [fmt1, fmt1, fmt1, ...] vs bundle with same array except element `[2]` changed to each color format (+ `kColorAttachmentFormats.length` cases)

@kangz is that what you had in mind?

And yeah I definitely think it's fine to do this in the PR where we update for sparse attachments, or as a separate PR before or after that

src/webgpu/api/validation/attachment_compatibility.spec.ts
egalli: Done

src/webgpu/api/validation/attachment_compatibility.spec.ts
egalli: Done

src/webgpu/api/validation/attachment_compatibility.spec.ts
egalli: Done

kainino0x (approved):
Thanks!

PR #416, opened 2020-12-18
Catch exceptions in expectValidationError

kainino0x:
And fix unhandled rejection in testMapAsyncCall.

Was previously caught in WPT (due to window.onunhandledrejection) but
not in standalone.

-----

- [x] WebGPU readability
- [x] TypeScript readability

### DIFF
--- a/src/webgpu/api/validation/buffer/mapping.spec.ts
+++ b/src/webgpu/api/validation/buffer/mapping.spec.ts
@@ -44,15 +44,16 @@ class F extends ValidationTest {
       const p = buffer.mapAsync(mode, offset, size);
       await p;
     } else {
-      await this.expectValidationErrorImmediate(async () => {
-        const p = buffer.mapAsync(mode, offset, size);
-        try {
-          await p;
-          assert(rejectName === null, 'mapAsync unexpectedly passed');
-        } catch (ex) {
-          assert(rejectName === ex.name, `mapAsync rejected unexpectedly with: ${ex}`);
-        }
+      let p: Promise<void>;
+      this.expectValidationError(() => {
+        p = buffer.mapAsync(mode, offset, size);
       });
+      try {
+        await p!;
+        assert(rejectName === null, 'mapAsync unexpectedly passed');
+      } catch (ex) {
+        assert(rejectName === ex.name, `mapAsync rejected unexpectedly with: ${ex}`);
+      }
     }
   }
 
--- a/src/webgpu/api/validation/buffer/mapping.spec.ts
+++ b/src/webgpu/api/validation/buffer/mapping.spec.ts
@@ -26,7 +26,7 @@ TODO: review existing tests and merge with this plan:
 
 import { pbool, poptions, params } from '../../../../common/framework/params_builder.js';
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
-import { unreachable } from '../../../../common/framework/util/util.js';
+import { assert, unreachable } from '../../../../common/framework/util/util.js';
 import { kBufferUsages } from '../../../capability_info.js';
 import { GPUConst } from '../../../constants.js';
 import { ValidationTest } from '../validation_test.js';
@@ -43,12 +43,15 @@ class F extends ValidationTest {
     if (success) {
       const p = buffer.mapAsync(mode, offset, size);
       await p;
-      this.shouldResolve(p);
     } else {
-      this.expectValidationError(async () => {
+      await this.expectValidationErrorImmediate(async () => {
         const p = buffer.mapAsync(mode, offset, size);
-        await p;
-        this.shouldReject(rejectName!, p);
+        try {
+          await p;
+          assert(rejectName === null, 'mapAsync unexpectedly passed');
+        } catch (ex) {
+          assert(rejectName === ex.name, `mapAsync rejected unexpectedly with: ${ex}`);
+        }
       });
     }
   }
--- a/src/webgpu/api/validation/validation_test.ts
+++ b/src/webgpu/api/validation/validation_test.ts
@@ -281,52 +281,39 @@ export class ValidationTest extends GPUTest {
     }
   }
 
-  async expectValidationErrorImmediate<T>(
-    fn: () => T | Promise<T>,
-    shouldError: boolean = true
-  ): Promise<T> {
-    if (!shouldError) {
-      return fn();
-    }
-
-    this.device.pushErrorScope('validation');
-    const returnValue = fn();
-
-    const gpuValidationError = await this.device.popErrorScope();
-    assert(gpuValidationError instanceof GPUValidationError, 'Validation error was expected.');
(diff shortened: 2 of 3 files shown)


src/webgpu/api/validation/validation_test.ts
```diff
@@ -280,15 +281,41 @@ export class ValidationTest extends GPUTest {
     }
   }
 
-  expectValidationError(fn: Function, shouldError: boolean = true): void {
+  async expectValidationErrorImmediate<T>(
+    fn: () => T | Promise<T>,
```
Kangz, line 285: Isn't `T` sufficient as it can be a `Promise<T>` as well?

Kangz (approved):

src/webgpu/api/validation/validation_test.ts
kainino0x: `T` is also used in the return type `Promise<T>`. They input should be flexible and the output should not be.

kainino0x: I guess doing this makes the type inference ambiguous, but somehow it works out and doesn't end up with `Promise<Promise<void>>` at the end.

austinEng: That's because if the return value is already a promise, an async function returns it directly. It only wraps the result if it's not a Promise

austinEng (commented):
Could we have less code duplication here if we updated expectValidationError to return a Promise created by this.eventualAsyncExpectation? Then if the callee waits for that it would be functionally equivalent to expectValidationErrorImmediate.

kainino0x:
> Could we have less code duplication here if we updated expectValidationError to return a Promise created by this.eventualAsyncExpectation? Then if the callee waits for that it would be functionally equivalent to expectValidationErrorImmediate.

I thought about doing this but it was going to be a bit more involved. May as well do it though.

kainino0x:
> I thought about doing this but it was going to be a bit more involved. May as well do it though.

I realized both the involved thing and the thing I did in this PR were not the right solutions. Fixed in a different way.

src/webgpu/api/validation/validation_test.ts
```diff
@@ -280,26 +281,39 @@ export class ValidationTest extends GPUTest {
...
-  expectValidationError(fn: Function, shouldError: boolean = true): void {
+  /**
+   * Expect a validation error inside the callback.
+   *
+   * Awaiting the returned promise is only necessary if you want to await on the completion
+   * of an async callback. The validation error check will still happen "eventually".
+   *
+   * Tests should always do just one WebGPU call in the callback, to make sure that's what's tested.
+   */
+  expectValidationError(fn: () => unknown, shouldError: boolean = true): void {
```
austinEng, line 292:
```suggestion
  expectValidationError(fn: () => undefined, shouldError: boolean = true): void {
```

if we have the assertion below at [1] on the return value, can we just say the type here should be undefined?

src/webgpu/api/validation/validation_test.ts
```diff
@@ -280,26 +281,39 @@ export class ValidationTest extends GPUTest {
...
     }
 
-    this.device.pushErrorScope('validation');
-    fn();
-    const promise = this.device.popErrorScope();
+    const returnValue = fn();
+    assert(
+      returnValue === undefined,
+      'expectValidationError callback should not return a value (or be async)'
+    );
```
austinEng, line 302: [1]
```suggestion
    fn();
```

src/webgpu/api/validation/validation_test.ts
```diff
@@ -280,26 +281,39 @@ export class ValidationTest extends GPUTest {
     }
   }
 
-  expectValidationError(fn: Function, shouldError: boolean = true): void {
+  /**
+   * Expect a validation error inside the callback.
+   *
+   * Awaiting the returned promise is only necessary if you want to await on the completion
+   * of an async callback. The validation error check will still happen "eventually".
```
austinEng, line 288: There's no returned promise here

src/webgpu/api/validation/buffer/mapping.spec.ts
```diff
@@ -43,13 +43,17 @@ class F extends ValidationTest {
...
       await p;
-      this.shouldResolve(p);
     } else {
-      this.expectValidationError(async () => {
-        const p = buffer.mapAsync(mode, offset, size);
-        await p;
-        this.shouldReject(rejectName!, p);
+      let p: Promise<void>;
+      this.expectValidationError(() => {
+        p = buffer.mapAsync(mode, offset, size);
```
austinEng, line 49: Hm, maybe I don't understand the problem - Why can't we return `p` instead of assigning it in this way?

```typescript
const p = this.expectValidationError(() => {
  return buffer.mapAsync(..)
});
```

Or, shouldn't the async version work here too?
```typescript
const p = this.expectValidationError(async () => {
  return await buffer.mapAsync(..)
});
```

Or to make it more explicit we could do

```typescript
const { callbackResult: p } = this.expectValidationError(async () => ...);
```

or... if we need the promise after popErrroScope too?

```typescript
const { callbackResult: p, validationPromise } = this.expectValidationError(async () => ...);
```

src/webgpu/api/validation/validation_test.ts
kainino0x: Unfortunately there is a weird quirk in TypeScript here. `() => {}` is not a valid `() => undefined` (because `void` is not assignable to `undefined`) but `async () => {}` and `() => 3` *are* valid `() => void`s (even though `Promise<void>` and `number` are not assignable to `void`). So AFAICT no matter what we put here we can't get better type safety without requiring `return undefined;` in all of the user callbacks.

I'll change it to void, it should be equivalent but more self-documenting.

src/webgpu/api/validation/validation_test.ts
kainino0x: oops, outdated doc

src/webgpu/api/validation/buffer/mapping.spec.ts
kainino0x: We can, but I realized that passing an async function into expectValidationError is confusing. Either we:
- make expectValidationError async and await on fn(): Causes an async split between pushErrorScope and popErrorScope, so if you forget to await expectValidationError (or choose not to because you don't care about the result), then other code will end up getting caught by the error scope
- make expectValidationError NOT await fn(), but just execute its first microtask and return the resulting promise: Confusing because you might think that the error scope would include the whole function and not just the microtask

austinEng: Right, okay that makes sense because we want it to be very explicit what's in-between the two synchronous push/pop calls. Sounds good to me then, let's document the reasoning in a comment.

austinEng (approved):
LGTM!

kainino0x:
thanks for the thorough review!

PR #424, opened 2020-12-27
[wgsl] Add validation test - v-0038: A module scope variable in the in or out storage class must be IO-shareable.

sarahM0:
A variable in the in or out storage classes:
Must use a store type as described in § [3.4.5 Storage Classes.](https://gpuweb.github.io/gpuweb/wgsl.html#storage-class).
ie. it must be IO shareable.

The following types are IO-shareable:

- numeric scalar types
- numeric vector types
- § 3.3.6 Matrix Types
- § 3.3.7 Array Types if its element type is IO-shareable, and the array is not runtime-sized
- § 3.3.8 Structure Types if all its members are IO-shareable

As a result these are not IO-shareable:

- boolean
- vector of bools
- array of bools
- array runtime sized -> cannot be used outside of a struct, so no cts for this
- struct with bool component
- struct with runtime array

-----

- [ ] WebGPU readability
- [ ] TypeScript readability

### DIFF
--- a/src/webgpu/shader/validation/wgsl/in-storage-class-not-IO-shareable-boolean-array.fail.wgsl
+++ b/src/webgpu/shader/validation/wgsl/in-storage-class-not-IO-shareable-boolean-array.fail.wgsl
@@ -1,0 +1,8 @@
+# v-0038: A module scope variable in 'out' storage class must be IO-sharable.
+# An array type with boolean elements is not IO-shareable.
+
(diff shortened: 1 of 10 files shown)


dj2 (commented):
Should we add passing tests as well?

sarahM0:
These tests are redundant because of https://github.com/gpuweb/cts/pull/448

PR #425, opened 2020-12-30
Add validation test for createSampler with different maxAnisotropy values

shrekshao:
Related to spec update at: https://github.com/gpuweb/gpuweb/pull/1323

~~I'm using Visual Code and seeing intelligence warning for assigning 'nearest' 'linear' to enum type `GPUFilterMode`. But the code builds and runs fine. Wondering if there needs any extra coding pattern for typescript.
(Well seems it doesn't build)~~

-----

- [ ] WebGPU readability
- [ ] TypeScript readability

### DIFF
--- a/src/webgpu/api/validation/createSampler.spec.ts
+++ b/src/webgpu/api/validation/createSampler.spec.ts
@@ -52,5 +52,5 @@ g.test('maxAnisotropy')
         mipmapFilter,
         maxAnisotropy,
       });
-    }, typeof maxAnisotropy !== 'number' || maxAnisotropy < 1 || (maxAnisotropy > 1 && !(minFilter === 'linear' && magFilter === 'linear' && mipmapFilter === 'linear')));
+    }, maxAnisotropy < 1 || (maxAnisotropy > 1 && !(minFilter === 'linear' && magFilter === 'linear' && mipmapFilter === 'linear')));
   });
--- a/src/webgpu/api/validation/createSampler.spec.ts
+++ b/src/webgpu/api/validation/createSampler.spec.ts
@@ -42,5 +42,5 @@ g.test('maxAnisotropy')
         mipmapFilter: t.params.mipmapFilter as GPUFilterMode,
         maxAnisotropy: t.params.maxAnisotropy,
       });
-    }, t.params.maxAnisotropy < 1 || (t.params.maxAnisotropy > 1 && !(t.params.minFilter == 'linear' && t.params.magFilter == 'linear' && t.params.mipmapFilter == 'linear') ));
+    }, t.params.maxAnisotropy < 1 || (t.params.maxAnisotropy > 1 && !(t.params.minFilter === 'linear' && t.params.magFilter === 'linear' && t.params.mipmapFilter === 'linear') ));
   });
--- a/src/webgpu/api/validation/createSampler.spec.ts
+++ b/src/webgpu/api/validation/createSampler.spec.ts
@@ -37,9 +37,9 @@ g.test('maxAnisotropy')
   .fn(async t => {
     t.expectValidationError(() => {
       t.device.createSampler({
-        minFilter: t.params.minFilter,
-        magFilter: t.params.magFilter,
-        mipmapFilter: t.params.mipmapFilter,
+        minFilter: t.params.minFilter as GPUFilterMode,
+        magFilter: t.params.magFilter as GPUFilterMode,
+        mipmapFilter: t.params.mipmapFilter as GPUFilterMode,
         maxAnisotropy: t.params.maxAnisotropy,
       });
     }, t.params.maxAnisotropy < 1 || (t.params.maxAnisotropy > 1 && !(t.params.minFilter == 'linear' && t.params.magFilter == 'linear' && t.params.mipmapFilter == 'linear') ));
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/createSampler.spec.ts
```diff
@@ -24,3 +24,23 @@ g.test('lodMinAndMaxClamp')
...
+      .combine(poptions('mipmapFilter', ['nearest', 'linear']))
+  )
+  .fn(async t => {
+    t.expectValidationError(() => {
+      t.device.createSampler({
+        minFilter: t.params.minFilter as GPUFilterMode,
+        magFilter: t.params.magFilter as GPUFilterMode,
+        mipmapFilter: t.params.mipmapFilter as GPUFilterMode,
+        maxAnisotropy: t.params.maxAnisotropy,
+      });
```
kainino0x, line 44: Since this is a validation test it's not really necessary to hit this entire test space - most of them are invalid for multiple reasons so they don't add coverage.

It's possible to reduce the combinatorics with something like this:
```suggestion
  .params([
    ...poptions('maxAnisotropy', [-1, undefined, 0, 1, 2, 4, 7, 16, 32, 33, 1024]),
    { minFilter: 'nearest' as const },
    { magFilter: 'nearest' as const },
    { mipmapFilter: 'nearest' as const },
  ])
  .fn(async t => {
    const {
      maxAnisotropy = 1,
      minFilter = 'linear',
      magFilter = 'linear',
      mipmapFilter = 'linear',
    } = t.params;
    t.expectValidationError(() => {
      t.device.createSampler({ minFilter, magFilter, mipmapFilter, maxAnisotropy });
```

src/webgpu/api/validation/createSampler.spec.ts
```diff
@@ -24,3 +24,23 @@ g.test('lodMinAndMaxClamp')
       });
     }, t.params.lodMinClamp > t.params.lodMaxClamp || t.params.lodMinClamp < 0 || t.params.lodMaxClamp < 0);
   });
+
+g.test('maxAnisotropy')
+  .desc('test different maxAnisotropy values and combinations with min/mag/mipmapFilter')
+  .params(
+    params()
+      .combine(poptions('maxAnisotropy', [0, 1, 2, 4, 16, 32, 1024]))
```
kainino0x, line 32: Probably good to test a few more values just for completeness - see below.

kainino0x (commented):
Test coverage looks pretty good - some tips below

src/webgpu/api/validation/createSampler.spec.ts
shrekshao: Having this error when trying to add null, NaN to the test case. Is typescript compiler making these the same type as undefined?
`Duplicate public test case params for test maxAnisotropy: maxAnisotropy=null`

Modified a bit to pass code style check, might be suboptimal.

src/webgpu/api/validation/createSampler.spec.ts
```diff
@@ -27,20 +27,30 @@ g.test('lodMinAndMaxClamp')
...
   .desc('test different maxAnisotropy values and combinations with min/mag/mipmapFilter')
-  .params(
-    params()
-      .combine(poptions('maxAnisotropy', [0, 1, 2, 4, 16, 32, 1024]))
-      .combine(poptions('minFilter', ['nearest', 'linear']))
-      .combine(poptions('magFilter', ['nearest', 'linear']))
-      .combine(poptions('mipmapFilter', ['nearest', 'linear']))
-  )
+  .params([
+    ...poptions('maxAnisotropy', [-1, undefined, 0, 1, 2, 4, 7, 16, 32, 33, 1024]),
```
kainino0x, line 31: re: some comment somewhere: It's okay not to test other values; we just want to test some values that are allowed by IDL. (Tests for IDL type check rules will be separate from validation tests.) Strictly speaking, we don't even have to test `undefined` here; we should test it when we have an operation test (and it should have the same expectation as `1`, because that is the default value).

src/webgpu/api/validation/createSampler.spec.ts
```diff
@@ -27,20 +27,30 @@ g.test('lodMinAndMaxClamp')
...
-        magFilter: t.params.magFilter as GPUFilterMode,
-        mipmapFilter: t.params.mipmapFilter as GPUFilterMode,
-        maxAnisotropy: t.params.maxAnisotropy,
+        minFilter,
+        magFilter,
+        mipmapFilter,
+        maxAnisotropy,
       });
-    }, t.params.maxAnisotropy < 1 || (t.params.maxAnisotropy > 1 && !(t.params.minFilter === 'linear' && t.params.magFilter === 'linear' && t.params.mipmapFilter === 'linear')));
+    }, typeof maxAnisotropy !== 'number' || maxAnisotropy < 1 || (maxAnisotropy > 1 && !(minFilter === 'linear' && magFilter === 'linear' && mipmapFilter === 'linear')));
```
kainino0x, line 55: Pretty sure this isn't needed with the current parameterization. It's never hit.
```suggestion
    }, maxAnisotropy < 1 || (maxAnisotropy > 1 && !(minFilter === 'linear' && magFilter === 'linear' && mipmapFilter === 'linear')));
```

kainino0x (approved):

kainino0x:
Thanks!

PR #426, opened 2021-01-04
Add validation tests for createQuerySet

haoxli:
- Add validation tests and test plan for createQuerySet with all types
- Validate all valid and invalid query count and pipeline statistics names.

-----

- [x] WebGPU readability
- [ ] TypeScript readability


src/webgpu/api/validation/query_set/create.spec.ts
```diff
@@ -0,0 +1,178 @@
...
+  - pipelineStatistics is undefined
+- invalid cases:
+  - count > kMaxQueryCount
+  - pipelineStatistics is set with pipeline statistic names
+  `
+  )
+  .params(
+    params()
+      .combine(poptions('count', kQueryCounts))
+      .combine(
```
kainino0x, line 39: It's not necessary to test cases that are invalid for multiple reasons (count is too large AND pipelineStatistics is invalid). You can use something like this: https://github.com/gpuweb/cts/pull/425#discussion_r554262676

kainino0x (commented):
Coverage looking good!

src/webgpu/api/validation/query_set/create.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+      extensions: ['pipeline-statistics-query'],
+    });
+
+    const { count = 1, pipelineStatistics = ['vertex-shader-invocations'] } = t.params as {
+      count?: number;
+      pipelineStatistics?: GPUPipelineStatisticName[];
+    };
+
+    // If the pipelineStatistics is passed as undefined, the default value 'vertex-shader-invocations'
+    // is got here, test it with undefined seperately.
```
Kangz, line 89: nit: separately

src/webgpu/api/validation/query_set/create.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+      // valid
+      ['clipper-invocations'],
+      // invalid
+      undefined,
+      [],
+      ['clipper-invocations', 'clipper-invocations'],
+    ]),
+  ])
+  .fn(async t => {
+    await t.selectDeviceOrSkipTestCase({
```
Kangz, line 79: Can we also test that creating a pipeline stastistics query / timestamp query set is invalid if we don't have the extension?

src/webgpu/api/validation/query_set/create.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+      extensions: ['pipeline-statistics-query'],
+    });
+
+    const { count = 1, pipelineStatistics = ['vertex-shader-invocations'] } = t.params as {
```
Kangz, line 83: I'm not sure I understand, why do we need the defaulting to vertex-shader-invocations to then detect that case and go back to setting undefined?

Kangz (approved):
LGTM overall

src/webgpu/api/validation/query_set/create.spec.ts
Kangz: Nevermind, that's done in another PR.

src/webgpu/api/validation/query_set/create.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+    ...poptions('pipelineStatistics', [
+      // valid
+      undefined,
+      [],
+      // invalid
+      ['clipper-invocations'],
+    ]),
+  ])
+  .fn(t => {
+    const { count = 1, pipelineStatistics } = t.params as {
```
kainino0x, line 45: The `= 1` here seems like it's not doing anything. Same below.

src/webgpu/api/validation/query_set/create.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+  )
+  .params([
+    ...poptions('count', kQueryCounts),
+    ...poptions('pipelineStatistics', [
+      // valid
+      undefined,
+      [],
+      // invalid
+      ['clipper-invocations'],
+    ]),
```
kainino0x, line 42: You can use `as const` here and then the type cast on `t.params` should no longer be needed.
```suggestion
    ] as const),
```

src/webgpu/api/validation/query_set/create.spec.ts
kainino0x: +1, I don't think any of that is necessary.

kainino0x (commented):
Few more comments. Sorry for the delay, I've been a bit swamped. Could you perhaps ask Shaobo or Jiawei for early reviews to help speed things along especially with the time zone difference?

Aside: looking at this with fresh eyes, I might have considered splitting up the test coverage differently. However, the coverage would be the same, so there is definitely no need to change anything.

- g.test('count'), parameterized over `type` and `count`
- g.test('pipelineStatistics'), parameterized over `type` and `pipelineStatistics`

src/webgpu/api/validation/query_set/create.spec.ts
haoxli: We use `...poptions` for setting the parameters count and pipelineStatistics. When testing count, pipelineStatistics is undefined, and vice versa, so we need to set a valid default value for them. But for pipeline statistics query, undefined is also a checkpoint, which overwrite by the default value, so I set different value from the parameter, and detect it should be checked as undefined. 

Kai's suggestion may avoid such problem.

haoxli (commented):
> Few more comments. Sorry for the delay, I've been a bit swamped. Could you perhaps ask Shaobo or Jiawei for early reviews to help speed things along especially with the time zone difference?
> 
> Aside: looking at this with fresh eyes, I might have considered splitting up the test coverage differently. However, the coverage would be the same, so there is definitely no need to change anything.
> 
> * g.test('count'), parameterized over `type` and `count`
> * g.test('pipelineStatistics'), parameterized over `type` and `pipelineStatistics`

That's OK, I just update these cases recently, and I will also cc Shaobo and Jiawei for reviews.

We need to separate the tests by query type because timestamp and pipeline statistics need to enable extensions. If we group the tests by 'count' and 'pipelineStatistics', and pass `type` as a parameter,  the whole test suite will be skipped
 if one of the query extension is not supported.

src/webgpu/api/validation/query_set/create.spec.ts
kainino0x: Ah yes sorry, I misread.

kainino0x:
> If we group the tests by 'count' and 'pipelineStatistics', and pass `type` as a parameter, the whole test suite will be skipped
> if one of the query extension is not supported.

The test would only need to request the extension for the type that's being tested. That way nothing would get unnecessarily skipped. Anyway let's not worry about this unnecessary refactor. Just a thought which might help designing future tests.

src/webgpu/api/validation/query_set/create.spec.ts
kainino0x: Ah I see I didn't really think about the fact that this is [...poptions] instead of .combine(poptions).

We should definitely rephrase this so that this complexity isn't needed. Probably by simply splitting this test into two - one for .count and one for .pipelineStatistics - as I think it would be simpler.

src/webgpu/api/validation/query_set/create.spec.ts
```diff
@@ -2,140 +2,110 @@ export const description = `
...
 
-import { poptions } from '../../../../common/framework/params_builder.js';
+import { params, poptions } from '../../../../common/framework/params_builder.js';
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
 import { ValidationTest } from '../validation_test.js';
 
 export const g = makeTestGroup(ValidationTest);
 
 const kMaxQueryCount = 8192;
+const kQuerytype = ['occlusion', 'pipeline-statistics', 'timestamp'] as const;
```
kainino0x, line 12: nit: These two constants can be in capability_info.ts. I pushed this change to your branch now.

src/webgpu/api/validation/query_set/create.spec.ts
```diff
@@ -2,140 +2,110 @@ export const description = `
...
+          ] as const,
+        ])
+      )
+      // Except pipeline statistics query, there is no need to test duplicate values and all values of GPUPipelineStatisticName
+      .unless(
+        ({ type, pipelineStatistics }) =>
+          type !== 'pipeline-statistics' &&
+          pipelineStatistics !== undefined &&
+          pipelineStatistics.length >= 2
+      )
```
kainino0x, line 82: Thanks, these tests did end up clearer with this split.

Now that I am looking at them this way, this is the last remaining source of complexity. I think we can clarify even further by splitting this test - one for each validation rule. And now that I think about it, that's a very useful guiding rule. What do you think of this split?

- One test for the rule "pipelineStatistics must be [] for non-pipeline-statistics queries": type in {all types} &times; pipelineStatistics in { undefined, [], ['clipper-invocations'] }
- One test for the rule "pipelineStatistics must not have duplicates": type = pipeline-statistics, pipelineStatistics in { ... }

src/webgpu/api/validation/query_set/create.spec.ts
```diff
@@ -2,140 +2,110 @@ export const description = `
...
+
+    if (type === 'pipeline-statistics') {
+      await t.selectDeviceOrSkipTestCase({
+        extensions: ['pipeline-statistics-query'],
+      });
+    } else if (type === 'timestamp') {
+      await t.selectDeviceOrSkipTestCase({
+        extensions: ['timestamp-query'],
+      });
+    }
```
kainino0x, line 38: Perhaps consider making a helper to deduplicate this code (`async function selectDeviceForQueryType(t: GPUTest, type: GPUQueryType): Promise<void>` in this file)

src/webgpu/api/validation/query_set/create.spec.ts
kainino0x: If there's a rule that pipelineStatistics must have at least one entry for `'pipeline-statistics`' query sets, that can roll into the second test I suggested.

haoxli: A little unclear. Do you mean split the 'pipelineStatistics' tests to two tests, one for non-pipeline-statistics queries and other for pipeline-statistics query? or spilt the checkpoint of 'pipelineStatistics' to two parts, one is { undefined, [], ['clipper-invocations'] } for all query types, the left is duplicated values and full values for pipeline-statistics query?

kainino0x: Sorry it's not that clear, it's hard to explain :)

> spilt the checkpoint of 'pipelineStatistics' to two parts, one is { undefined, [], ['clipper-invocations'] } for all query types, the left is duplicated values and full values for pipeline-statistics query?

This exactly. Message me on slack if you would like to chat more about it; I'll try to be online a bit.

haoxli: Thanks, I got your point.

src/webgpu/api/validation/query_set/create.spec.ts
haoxli: Added the helper function.

kainino0x (approved):
Looks great! Thanks, and sorry for the many revisions :)

src/webgpu/api/validation/query_set/create.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+
+    const pipelineStatistics =
+      type === 'pipeline-statistics' ? (['clipper-invocations'] as const) : ([] as const);
+
+    t.expectValidationError(() => {
+      t.device.createQuerySet({ type, count, pipelineStatistics });
+    }, count > kMaxQueryCount);
+  });
+
+g.test('pipelineStatistics')
```
kainino0x, line 49:
```suggestion
g.test('pipelineStatistics,all_types')
```

src/webgpu/api/validation/query_set/create.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+        pipelineStatistics.length > 0) ||
+      (type === 'pipeline-statistics' &&
+        (pipelineStatistics === undefined || pipelineStatistics.length === 0));
+
+    t.expectValidationError(() => {
+      t.device.createQuerySet({ type, count, pipelineStatistics });
+    }, shouldError);
+  });
+
+g.test('pipelineStatistics_duplicates_and_all')
```
kainino0x, line 83:
```suggestion
g.test('pipelineStatistics,duplicates_and_all')
```

PR #431, opened 2021-01-07
Added range validation tests for GPUQueue.writeBuffer

toji:
Adds tests that check various validation edge cases for the offset and size ranges provided to `GPUQueue.writeBuffer`.

I would have liked each TypedArray variant to be a different `params` for the test, but couldn't figure out how to make TypeScript happy with that. Suggestions welcome!

-----

- [x] WebGPU readability
- [ ] TypeScript readability

### DIFF
--- a/src/webgpu/api/operation/queue/writeBuffer.spec.ts
+++ b/src/webgpu/api/operation/queue/writeBuffer.spec.ts
@@ -8,3 +8,89 @@ import { makeTestGroup } from '../../../../common/framework/test_group.js';
 import { GPUTest } from '../../../gpu_test.js';
 
 export const g = makeTestGroup(GPUTest);
+
+g.test('writeBuffer,Ranges')
+  .params([
+    {},
+  ])
+  .fn(async t => {
+    const queue = t.device.defaultQueue;
+
+    function runTest(arrayType : any, testBuffer : boolean) {
+      const elementSize = arrayType.BYTES_PER_ELEMENT;
+      const buffer = t.device.createBuffer({ size: 16 * elementSize, usage: GPUBufferUsage.COPY_DST });
+      let arraySm : any = new arrayType(8);
+      let arrayMd : any = new arrayType(16);
+      let arrayLg : any = new arrayType(32);
+
+      if (testBuffer) {
+        arraySm = arraySm.buffer;
+        arrayMd = arrayMd.buffer;
+        arrayLg = arrayLg.buffer;
+
+        const array15 = new Uint8Array(15).buffer;
+
+        // Writing the full buffer that isn't 4-byte aligned.
+        t.shouldThrow('OperationError', () => queue.writeBuffer(buffer, 0, array15));
+
+        // Writing from an offset that causes source to be 4-byte aligned.
+        t.expectGPUError('validation', () => queue.writeBuffer(buffer, 0, array15, 3), false);
+      }
+
+      if (elementSize < 4) {
+        // Writing from an offset that causes the source to not be 4-byte aligned.
+        t.shouldThrow('OperationError', () => queue.writeBuffer(buffer, 0, arrayMd, 3));
+
+        // Writing with a size that is not 4-byte aligned.
+        t.shouldThrow('OperationError', () => queue.writeBuffer(buffer, 0, arraySm, 0, 7));
+      }
+
+      // Writing the full buffer without offsets.
+      t.expectGPUError('validation', () => queue.writeBuffer(buffer, 0, arraySm), false);
+      t.expectGPUError('validation', () => queue.writeBuffer(buffer, 0, arrayMd), false);
+      t.expectGPUError('validation', () => queue.writeBuffer(buffer, 0, arrayLg), true);
+
+      // Writing the full buffer with a 4-byte aligned offset.
+      t.expectGPUError('validation', () => queue.writeBuffer(buffer, 8, arraySm), false);
+      t.expectGPUError('validation', () => queue.writeBuffer(buffer, 8, arrayMd), true);
+
+      // Writing the full buffer with a unaligned offset.
+      t.shouldThrow('OperationError', () => queue.writeBuffer(buffer, 3, arraySm));
+
+      // Writing remainder of buffer from offset.
+      t.expectGPUError('validation', () => queue.writeBuffer(buffer, 0, arraySm, 4), false);
+      t.expectGPUError('validation', () => queue.writeBuffer(buffer, 0, arrayMd, 4), false);
+      t.expectGPUError('validation', () => queue.writeBuffer(buffer, 0, arrayLg, 4), true);
+
+      // Writing a larger buffer from an offset that allows it to fit in the destination.
+      t.expectGPUError('validation', () => queue.writeBuffer(buffer, 0, arrayLg, 16), false);
+
+      // Writing with both an offset and size.
+      t.expectGPUError('validation', () => queue.writeBuffer(buffer, 0, arraySm, 4, 4), false);
+
+      // Writing with a size that extends past the source buffer length.
+      t.shouldThrow('OperationError', () => queue.writeBuffer(buffer, 0, arraySm, 0, 16));
+      t.shouldThrow('OperationError', () => queue.writeBuffer(buffer, 0, arraySm, 4, 8));
+
+      // Writing with a size that is 4-byte aligned but an offset that is not.
+      t.expectGPUError('validation', () => queue.writeBuffer(buffer, 0, arraySm, 3, 4), false);
+    }
+
+    const arrayTypes = [
+      Uint8Array,
+      Uint8Array,
+      Int8Array,
+      Uint16Array,
+      Int16Array,
+      Uint32Array,
+      Int32Array,
+      Float32Array,
+      Float64Array,
+    ];
+
+    runTest(Uint8Array, true);
+
+    for (const arrayType of arrayTypes) {
+      runTest(arrayType, false);
+    }
+  });
--- a/src/webgpu/api/operation/queue/writeBuffer.spec.ts
+++ b/src/webgpu/api/operation/queue/writeBuffer.spec.ts
@@ -1,96 +1,11 @@
 export const description = `
 TODO:
 - source.origin is unaligned
-- ?
+- data given as ArrayBuffer
+- data given as each TypedArray variant
 `;
 
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
 import { GPUTest } from '../../../gpu_test.js';
 
 export const g = makeTestGroup(GPUTest);
-
-g.test('writeBuffer,Ranges')
-  .params([
-    {},
-  ])
-  .fn(async t => {
-    const queue = t.device.defaultQueue;
-
-    function runTest(arrayType : any, testBuffer : boolean) {
-      const elementSize = arrayType.BYTES_PER_ELEMENT;
-      const buffer = t.device.createBuffer({ size: 16 * elementSize, usage: GPUBufferUsage.COPY_DST });
-      let arraySm : any = new arrayType(8);
-      let arrayMd : any = new arrayType(16);
-      let arrayLg : any = new arrayType(32);
-
-      if (testBuffer) {
-        arraySm = arraySm.buffer;
-        arrayMd = arrayMd.buffer;
-        arrayLg = arrayLg.buffer;
-
-        const array15 = new Uint8Array(15).buffer;
-
-        // Writing the full buffer that isn't 4-byte aligned.
-        t.shouldThrow('OperationError', () => queue.writeBuffer(buffer, 0, array15));
-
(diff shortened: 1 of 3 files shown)


kainino0x:
You can autoformat with `npx grunt fix`. (Probably should make an alias as `npm run fix` I guess.)

src/webgpu/api/operation/queue/writeBuffer.spec.ts
```diff
@@ -8,3 +8,89 @@ import { makeTestGroup } from '../../../../common/framework/test_group.js';
...
 export const g = makeTestGroup(GPUTest);
+
+g.test('writeBuffer,Ranges')
+  .params([
+    {},
+  ])
+  .fn(async t => {
+    const queue = t.device.defaultQueue;
+
+    function runTest(arrayType : any, testBuffer : boolean) {
```
kainino0x, line 19: You can add an `export` on this type:
https://github.com/gpuweb/cts/blob/main/src/webgpu/gpu_test.ts#L25
and use it here.

src/webgpu/api/operation/queue/writeBuffer.spec.ts
```diff
@@ -8,3 +8,89 @@ import { makeTestGroup } from '../../../../common/framework/test_group.js';
...
+      const elementSize = arrayType.BYTES_PER_ELEMENT;
+      const buffer = t.device.createBuffer({ size: 16 * elementSize, usage: GPUBufferUsage.COPY_DST });
+      let arraySm : any = new arrayType(8);
```
kainino0x, line 22: Can use the `TypedArrayBufferView` type next to `TypedArrayBufferViewConstructor`.

src/webgpu/api/operation/queue/writeBuffer.spec.ts
```diff
@@ -8,3 +8,89 @@ import { makeTestGroup } from '../../../../common/framework/test_group.js';
 import { GPUTest } from '../../../gpu_test.js';
 
 export const g = makeTestGroup(GPUTest);
+
+g.test('writeBuffer,Ranges')
+  .params([
+    {},
+  ])
```
kainino0x, line 15: This is equivalent to just omitting the .params call.

src/webgpu/api/operation/queue/writeBuffer.spec.ts
```diff
@@ -8,3 +8,89 @@ import { makeTestGroup } from '../../../../common/framework/test_group.js';
 import { GPUTest } from '../../../gpu_test.js';
 
 export const g = makeTestGroup(GPUTest);
+
+g.test('writeBuffer,Ranges')
```
kainino0x, line 12: Needs a `.desc()` with a test plan that explains what the test tests in a bit more detail. BTW, have you read https://github.com/gpuweb/cts/blob/main/docs/intro/README.md?

kainino0x, line 12: `writeBuffer` can be omitted here because it's redundant with the filename. So just `'ranges'` or maybe `typed_array_types`

src/webgpu/api/operation/queue/writeBuffer.spec.ts
```diff
@@ -8,3 +8,89 @@ import { makeTestGroup } from '../../../../common/framework/test_group.js';
...
+        arrayMd = arrayMd.buffer;
+        arrayLg = arrayLg.buffer;
+
+        const array15 = new Uint8Array(15).buffer;
+
+        // Writing the full buffer that isn't 4-byte aligned.
+        t.shouldThrow('OperationError', () => queue.writeBuffer(buffer, 0, array15));
+
+        // Writing from an offset that causes source to be 4-byte aligned.
+        t.expectGPUError('validation', () => queue.writeBuffer(buffer, 0, array15, 3), false);
```
kainino0x, line 37: nit: expectGPUError is not needed for cases where an error is never expected; if there's an error it'll be caught by the "ambient" error checking which occurs around every test. The shouldError param is only needed when it's computed somehow from the params.

src/webgpu/api/operation/queue/writeBuffer.spec.ts
```diff
@@ -8,3 +8,89 @@ import { makeTestGroup } from '../../../../common/framework/test_group.js';
...
+  ])
+  .fn(async t => {
+    const queue = t.device.defaultQueue;
+
+    function runTest(arrayType : any, testBuffer : boolean) {
+      const elementSize = arrayType.BYTES_PER_ELEMENT;
+      const buffer = t.device.createBuffer({ size: 16 * elementSize, usage: GPUBufferUsage.COPY_DST });
+      let arraySm : any = new arrayType(8);
+      let arrayMd : any = new arrayType(16);
+      let arrayLg : any = new arrayType(32);
```
kainino0x, line 24: Would it be practical to parameterize over the array size, and doing every check on all sizes? It would require some computation of the shouldError values though, so maybe it wouldn't be a net win.

src/webgpu/api/operation/queue/writeBuffer.spec.ts
```diff
@@ -8,3 +8,89 @@ import { makeTestGroup } from '../../../../common/framework/test_group.js';
...
+      Uint8Array,
+      Uint8Array,
+      Int8Array,
+      Uint16Array,
+      Int16Array,
+      Uint32Array,
+      Int32Array,
+      Float32Array,
+      Float64Array,
+    ];
```
kainino0x, line 89: If you wanted to parameterize you could do something like this https://github.com/gpuweb/cts/blob/main/src/webgpu/api/operation/vertex_state/index_format.spec.ts#L105

src/webgpu/api/operation/queue/writeBuffer.spec.ts
```diff
@@ -8,3 +8,89 @@ import { makeTestGroup } from '../../../../common/framework/test_group.js';
 import { GPUTest } from '../../../gpu_test.js';
 
 export const g = makeTestGroup(GPUTest);
```
kainino0x, line 10: Validation tests should be in `api/validation` instead of `api/operation`. Looks like this test only tests validation and exceptions, doesn't actually test uploaded results for valid cases, so it can just move there wholesale. When you do so, take a look at the file description that's there and update it so it's clear what is still TODO.

kainino0x, line 10: The typed array types are something that we SHOULD have operation tests for though, so please add that to the TODO list in this file.

kainino0x (commented):
Thanks for the contribution! Many small comments :)

src/webgpu/api/operation/queue/writeBuffer.spec.ts
toji: Feels like that ends up being more complex and less readable than what I have now without too much concrete benefit.

toji:
Addressed feedback, thanks!

src/webgpu/api/validation/queue/writeBuffer.spec.ts
```diff
@@ -11,10 +11,115 @@ Tests writeBuffer validation.
...
+        : new arrayType(16);
+      const arrayLg: TypedArrayBufferView | ArrayBuffer = testBuffer
+        ? new arrayType(32).buffer
+        : new arrayType(32);
+
+      if (testBuffer) {
+        const array15 = new Uint8Array(15).buffer;
+
+        // Writing the full buffer that isn't 4-byte aligned.
+        t.shouldThrow('OperationError', () => queue.writeBuffer(buffer, 0, array15));
```
kainino0x, line 63: For negative tests to be as strong as possible, I think this should be "minimally unaligned", i.e. a multiple of the biggest power of 2 that's still invalid. That is, I think testing "14 (multiple of 2) is invalid" is stronger than testing "15 (multiple of 1) is invalid". Similar in other places in this test.

src/webgpu/api/validation/queue/writeBuffer.spec.ts
```diff
@@ -11,10 +11,115 @@ Tests writeBuffer validation.
...
+
+        // Writing from an offset that causes source to be 4-byte aligned.
+        queue.writeBuffer(buffer, 0, array15, 3);
```
kainino0x, line 66: Since this is a positive test, being maximally unaligned, as this is now, is best. Of course we will also want operation tests for corner cases like this.

src/webgpu/api/validation/queue/writeBuffer.spec.ts
```diff
@@ -11,10 +11,115 @@ Tests writeBuffer validation.
 
 Note: destroyed buffer is tested in destroyed/.
 
-TODO: implement.
+TODO: implement usage flag validation.
```
kainino0x, line 14: Another TODO I thought of while looking at copyBufferToBuffer: try to test huge values for the copy size that could overflow (e.g. if multiplied to bytes before comparing the size with the TypedArray) and end up being incorrectly treated as valid.

src/webgpu/api/validation/queue/writeBuffer.spec.ts
```diff
@@ -11,10 +11,115 @@ Tests writeBuffer validation.
...
+
+      // Writing remainder of buffer from offset.
+      queue.writeBuffer(buffer, 0, arraySm, 4);
+      queue.writeBuffer(buffer, 0, arrayMd, 4);
+      t.expectGPUError('validation', () => queue.writeBuffer(buffer, 0, arrayLg, 4));
+
+      // Writing a larger buffer from an offset that allows it to fit in the destination.
+      queue.writeBuffer(buffer, 0, arrayLg, 16);
+
+      // Writing with both an offset and size.
```
kainino0x, line 97: Somewhere (maybe here, maybe later) we should have tests that call `writeBuffer(dst, 0, src, undefined, size)` which is also valid.

toji: Added a couple of tests to ensure it's treated as 0.

src/webgpu/api/validation/queue/writeBuffer.spec.ts
toji: Added the TODO

src/webgpu/api/validation/queue/writeBuffer.spec.ts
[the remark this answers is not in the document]
toji: Done

src/webgpu/api/validation/queue/writeBuffer.spec.ts
```diff
@@ -11,10 +11,115 @@ Tests writeBuffer validation.
...
+      }
+
+      // Writing the full buffer without offsets.
+      queue.writeBuffer(buffer, 0, arraySm);
+      queue.writeBuffer(buffer, 0, arrayMd);
+      t.expectGPUError('validation', () => queue.writeBuffer(buffer, 0, arrayLg));
+
+      // Writing the full buffer with a 4-byte aligned offset.
+      queue.writeBuffer(buffer, 8, arraySm);
+      t.expectGPUError('validation', () => queue.writeBuffer(buffer, 8, arrayMd));
```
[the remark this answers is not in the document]
toji: That's handy! Done.

src/webgpu/api/validation/queue/writeBuffer.spec.ts
```diff
@@ -11,10 +11,115 @@ Tests writeBuffer validation.
...
+interpreted correctly for both.
+
+  - When passing a TypedArray the data offset and size is given in elements.
+  - When passing an ArrayBuffer the data offset and size is given in bytes.
+
+Also verifies that the specified data range:
+
+  - Describes a valid range of the source buffer.
+  - Fits fully within the destination buffer.
+  - Is 4-byte aligned.
```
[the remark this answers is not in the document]
toji: Updated the phrasing.

toji:
Addressed most of the feedback. Don't know that I've fully groked the "Minimally unaligned" comments but I'll read up on it.

src/webgpu/api/validation/queue/writeBuffer.spec.ts
kainino0x: I don't think there's anything to read up on, I kind of made it up, but basically: Alignments are generally to powers of two. If I call 8 "more aligned" than 4, 4 "more aligned" than 2, etc., and we want to test two things:
- The implementation doesn't require a bigger alignment (say, 8 instead of 4) than the spec.
- The implementation doesn't require a smaller alignment (say, 2 instead of 4) than the spec.

Thus for positive cases you want to test "least aligned" valid values, (12, say, to check the impl requires a multiple of 4 or smaller)
while for negative cases you want to test "most aligned" invalid values (10, say, to check the impl requires a multiple of 4 or bigger)

kainino0x (approved):
lgtm other than possible strengthening for alignments

kainino0x:
No reason to block landing this on possibly making some cases slightly stronger. Landing now.

PR #433, opened 2021-01-14
Organize more and add more test stubs

kainino0x:
-----

- [x] WebGPU readability
- [x] TypeScript readability


src/webgpu/api/operation/onuncapturederror.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+
+g.test('only_original_device_is_event_target')
+  .desc(
+    `GPUDevices copied to other threads should not be EventTargets and should not have onuncapturederror.`
+  )
+  .unimplemented();
+
+g.test('uncapturederror_from_non_originating_thread')
+  .desc(
+    `Uncaptured errors on other threads propagate to originating thread's uncapturederror event.`
```
Kangz, line 30: What does this mean? That each thread has its own uncapturederror event?

src/webgpu/api/operation/resource_init/buffer.spec.ts
```diff
@@ -0,0 +1,10 @@
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { GPUTest } from '../../../gpu_test.js';
+
+export const description = `
+Test uninitialized buffers are initialized to zero when read.
```
Kangz, line 5: or written, for example with atomics (technically a combined read / write but I think it's good to call it out).

Kangz (approved):

PR #437, opened 2021-01-16
Add sampler anisotropy tests

shrekshao:
Add sampler anisotropy tests (first part), which is mostly a port from the dawn test.
Plans for the other part is added as description. Let me send it for review first.

-----

- [x] WebGPU readability
- [x] TypeScript readability


shrekshao:
Commit separator for part 2 (thinking it might be unnecessary to open a separate pull request)

shrekshao:
Part 2 adds a checkerboard tests that uses no mipmap and compares between the render results with different sampler to see if they match or not. Also some refactor for better code usage.

Open question: how we tell (query?) the maxAnisotropy limit/support for different hardware? (The test now just assumes 16 is the limit which is universal support of metal device)

src/webgpu/api/operation/sampling/anisotropy.spec.ts
```diff
@@ -6,14 +6,506 @@ Note that anisotropic filtering is never guaranteed to occur, but we might be ab
...
+            [[stage(fragment)]] fn main() -> void {
+                fragColor = textureSample(texture0, sampler0, fragUV);
+            }
+            `,
+        }),
+        entryPoint: 'main',
+      },
+      primitiveTopology: 'triangle-list',
+      colorStates: [{ format: 'rgba8unorm' }],
+      vertexState: {
```
kainino0x, line 129: Consider using something like this:
https://github.com/gpuweb/cts/blob/main/src/webgpu/api/operation/rendering/basic.spec.ts#L65
instead of using vertex buffers. Helps cut down the size of the test.

You can also hardcode the matrix into a const in the shader. Optionally you could even pre-multiply the matrix into the position data so no matrix is needed.

src/webgpu/api/operation/sampling/anisotropy.spec.ts
```diff
@@ -6,14 +6,506 @@ Note that anisotropic filtering is never guaranteed to occur, but we might be ab
...
+    ]);
+    this.uniformBuffer = this.device.createBuffer({
+      size: matrixData.byteLength,
+      usage: GPUBufferUsage.VERTEX | GPUBufferUsage.UNIFORM | GPUBufferUsage.COPY_DST,
+    });
+    this.device.defaultQueue.writeBuffer(this.uniformBuffer, 0, matrixData);
+
+    // position : vec4, uv : vec2
+    // uv is scaled
+    const vertexData = new Float32Array([
```
kainino0x, line 178: (Code may get deleted, but) You could use prettier-ignore something like this:
```suggestion
    const vertexData = new Float32Array(/* prettier-ignore */ [
```
so you can format the block of data more nicely.

src/webgpu/api/operation/sampling/anisotropy.spec.ts
```diff
@@ -6,14 +6,506 @@ Note that anisotropic filtering is never guaranteed to occur, but we might be ab
...
+    );
+    this.queue.submit([commandEncoder.finish()]);
+
+    return buffer;
+  }
+
+  private pipeline: GPURenderPipeline | undefined;
+  private uniformBuffer: GPUBuffer | undefined;
+  private vertexBuffer: GPUBuffer | undefined;
+  setup(): void {
```
kainino0x, line 83: Should be possible to have this override GPUTest.init and then you won't have to call it in the test function.

src/webgpu/api/operation/sampling/anisotropy.spec.ts
```diff
@@ -6,14 +6,506 @@ Note that anisotropic filtering is never guaranteed to occur, but we might be ab
...
+        mipmapFilter: 'linear',
+        maxAnisotropy: 1024,
+      }),
+    ];
+
+    const byteLength = kRTSize * kBytesPerRow;
+
+    const dst: GPUBuffer[] = [];
+
+    Promise.all(
```
kainino0x, line 363: I think you need to await this, or the test could end before the expectations are registered (if you didn't see this fail locally, I wonder why).

Personally I'd rephrase this to just use a regular loop instead of Promise.all, it would be easier to read with straight-line async code.
The loop could just be `for (const maxAnisotropy of [1, 16, 1024])` I think. Then you would only need one createSampler() line, as well.

src/webgpu/api/operation/sampling/anisotropy.spec.ts
```diff
@@ -6,14 +6,506 @@ Note that anisotropic filtering is never guaranteed to occur, but we might be ab
...
+    for (let i = 0, len = a.length; i < len; i++) {
+      if (a[i] !== b[i]) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  checkCheckerBoardResults(results: Uint8Array[]) {
+    this.eventualAsyncExpectation(async niceStack => {
```
kainino0x, line 48: I think eventualAsyncExpectation isn't needed unless the inner function is actually async, so here you could just not use it and use regular `t.expect` and so on.

src/webgpu/api/operation/sampling/anisotropy.spec.ts
```diff
@@ -6,14 +6,506 @@ Note that anisotropic filtering is never guaranteed to occur, but we might be ab
...
+    We will check the rendering result using sampler with maxAnisotropy values to be
+    different from each other, as the sampling rate is different.
+    We will also check if those large maxAnisotropy values are clamped so that rendering is the
+    same as the supported upper limit say 16.
+    A smiliar webgl demo is at https://jsfiddle.net/yqnbez24`
+  )
+  // .unimplemented();
+  // .params([
+  //   ...poptions('maxAnisotropy', [1, 2, 4, 16]),
+  // ])
```
kainino0x, line 282: old comment

src/webgpu/api/operation/sampling/anisotropy.spec.ts
```diff
@@ -6,14 +6,506 @@ Note that anisotropic filtering is never guaranteed to occur, but we might be ab
...
 
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { assert } from '../../../../common/framework/util/util.js';
 import { GPUTest } from '../../../gpu_test.js';
+import { mipSize } from '../../../util/texture/subresource.js';
 
-export const g = makeTestGroup(GPUTest);
+const kRTSize = 16;
+const kBytesPerRow = 256;
+const xm = kRTSize / 2;
```
kainino0x, line 20: nit: Pick a more specific name, or maybe just inline this at the usage sites

src/webgpu/api/operation/sampling/anisotropy.spec.ts
```diff
@@ -6,14 +6,506 @@ Note that anisotropic filtering is never guaranteed to occur, but we might be ab
...
+        data[c + 1] = color[1];
+        data[c + 2] = color[2];
+        data[c + 3] = color[3];
+      }
+    }
+    const buffer = t.device.createBuffer({
+      usage: GPUBufferUsage.COPY_SRC | GPUBufferUsage.COPY_DST,
+      size: data.byteLength,
+    });
+    t.device.defaultQueue.writeBuffer(buffer, 0, data);
```
kainino0x, line 316: You can use GPUTest.makeBufferWithContents to make this slightly simpler.

Unfortunately we don't have a makeTextureWithContents but it's something I'd like to do eventually.

src/webgpu/api/operation/sampling/anisotropy.spec.ts
```diff
@@ -6,14 +6,506 @@ Note that anisotropic filtering is never guaranteed to occur, but we might be ab
...
+
+g.test('anisotropic_filter_checkerboard')
+  .desc(
+    `anisotropic filter rendering tests that draws a slanted plane and samples from a texture
+    that only has a top level mipmap, the content of which is like a checkerboard.
+    We will check the rendering result using sampler with maxAnisotropy values to be
+    different from each other, as the sampling rate is different.
+    We will also check if those large maxAnisotropy values are clamped so that rendering is the
+    same as the supported upper limit say 16.
+    A smiliar webgl demo is at https://jsfiddle.net/yqnbez24`
```
kainino0x, line 277: nit: similar

src/webgpu/api/operation/sampling/anisotropy.spec.ts
```diff
@@ -6,14 +6,506 @@ Note that anisotropic filtering is never guaranteed to occur, but we might be ab
...
+      if (a[i] !== b[i]) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  checkCheckerBoardResults(results: Uint8Array[]) {
+    this.eventualAsyncExpectation(async niceStack => {
+      if (true === this.compareUint8ArrayResult(results[0], results[1])) {
```
kainino0x, line 49: Should be possible to use the (not that well-named) GPUTest.checkBuffer here.

kainino0x, line 49: nit: `true ===` is redundant since compareUint8ArrayResult returns boolean.

src/webgpu/api/operation/sampling/anisotropy.spec.ts
```diff
@@ -6,14 +6,506 @@ Note that anisotropic filtering is never guaranteed to occur, but we might be ab
...
+        niceStack.message =
+          'Render results with sampler.maxAnisotropy being 1 and 16 should be different.';
+        this.rec.expectationFailed(niceStack);
+      }
+      if (false === this.compareUint8ArrayResult(results[1], results[2])) {
+        niceStack.message =
+          'Render results with sampler.maxAnisotropy being 16 and 1024 should be the same.';
+        this.rec.expectationFailed(niceStack);
```
kainino0x, line 57: (This doesn't need to be a warning, though.)
You can use GPUTest.expectBuffer here I think.

src/webgpu/api/operation/sampling/anisotropy.spec.ts
```diff
@@ -6,14 +6,506 @@ Note that anisotropic filtering is never guaranteed to occur, but we might be ab
...
+    }
+    return true;
+  }
+
+  checkCheckerBoardResults(results: Uint8Array[]) {
+    this.eventualAsyncExpectation(async niceStack => {
+      if (true === this.compareUint8ArrayResult(results[0], results[1])) {
+        niceStack.message =
+          'Render results with sampler.maxAnisotropy being 1 and 16 should be different.';
+        this.rec.expectationFailed(niceStack);
```
kainino0x, line 52: Unfortunately we can't make a strong guarantee that maxAnisotropy=1 and =16 actually produce different results. Instead you can use `this.warn` here.

Later, we will want to add a way that we can catch warnings on automated testing (probably an equivalent of `-Werror` - sadly WPT doesn't have a concept of warnings so we can't put in warning expectations entries).

kainino0x (commented):
I haven't gotten to anisotropic_filter_mipmap_color or gpu_test yet, but here are some comments. I'll review anisotropic_filter_mipmap_color once you've addressed them as I think a few of them apply to both tests. Will try to get to gpu_test tomorrow.

kainino0x:
> Open question: how we tell (query?) the maxAnisotropy limit/support for different hardware? (The test now just assumes 16 is the limit which is universal support of metal device)

Oops, lost my reply to this. The spec chose not to expose a maxAnisotropy value as there are no guarantees about what actually happens when you enable anisotropic filtering. Even on a device that advertises it has anisotropic filtering with a max of 16, it's within spec in Vulkan for the rendering to be exactly the same as with maxAnisotropy=1. Because of this, we decided (at least for now) not to expose any info about whether anisotropic filtering is available at all, or what the maximum maxAnisotropy is, and instead treat it as a pure hint.

shrekshao:
Thanks for the review (addressing). So we will just run these tests and skip on those devices that failing the test (not supporting any anisotropic filtering)?

kainino0x:
> Thanks for the review (addressing). ~So we will just run these tests and skip on those devices that failing the test (not supporting any anisotropic filtering)?~ (use warning instead)

Yup, we can use a warning, set the equivalent of Werror, and like with Chrome's WebglExtension tests, skip if we have any test devices that fail it (I don't think we will, actually).

src/webgpu/api/operation/sampling/anisotropy.spec.ts
```diff
@@ -334,67 +219,56 @@ g.test('anisotropic_filter_checkerboard')
...
-        results[i] = new Uint8Array(dst[i].getMappedRange());
-      }
+        maxAnisotropy,
+      });
+      const d = t.createAlignedCopyForMapRead(
+        t.copyRenderTargetToBuffer(t.drawSlantedPlane(textureView, sampler)),
+        byteLength,
+        0
+      ).dst;
+      dst.push(d);
```
kainino0x, line 238: nit: Is this array actually needed?

src/webgpu/api/operation/sampling/anisotropy.spec.ts
```diff
@@ -6,14 +6,377 @@ Note that anisotropic filtering is never guaranteed to occur, but we might be ab
...
+g.test('anisotropic_filter_mipmap_color')
+  .desc(
+    `anisotropic filter rendering tests that draws a slanted plane and samples from a texture
+    containing mipmaps of different colors.
+    A similar webgl demo is at https://jsfiddle.net/t8k7c95o/5/`
+  )
+  .params([
+    {
+      maxAnisotropy: 1,
+      _results: [
```
kainino0x, line 265: Elegant data-driven expectations :)

src/webgpu/api/operation/sampling/anisotropy.spec.ts
```diff
@@ -6,14 +6,377 @@ Note that anisotropic filtering is never guaranteed to occur, but we might be ab
...
+        {
+          coord: { x: xMiddle, y: 2 },
+          expected: colors[2],
+        },
```
kainino0x, line 269: uNit: I think each of these `{}`s could fit on one line

src/webgpu/api/operation/sampling/anisotropy.spec.ts
```diff
@@ -6,14 +6,377 @@ Note that anisotropic filtering is never guaranteed to occur, but we might be ab
...
+    }
+    t.device.defaultQueue.submit([textureEncoder.finish()]);
+
+    const textureView = texture.createView();
+
+    const sampler = t.device.createSampler({
+      magFilter: 'linear',
+      minFilter: 'linear',
+      mipmapFilter: 'linear',
+      maxAnisotropy: t.params.maxAnisotropy,
```
kainino0x, line 361: Commenting out this line should not cause test failures, as it's valid for implementations to ignore this parameter. I guess the best way to achieve this is to change errors to warnings? Unfortunately this is not currently doable with the expect functions :(

Maybe it's possible to loosen the expectations? (maybe have both warnings and hard errors)

kainino0x (commented):
Sorry I didn't manage to finish reviewing this today. The first test looks great though.

src/webgpu/api/operation/sampling/anisotropy.spec.ts
shrekshao: What about expecting the result to be equal to either that of `maxAnisotropy = t.params.maxAnisotropy` or `maxAnisotropy = 1`? (which is assuming if device support anisotropy, it supports up to 16)

src/webgpu/api/operation/sampling/anisotropy.spec.ts
shrekshao: Not any more...

src/webgpu/api/operation/sampling/anisotropy.spec.ts
shrekshao: (define 'elegant' ... 😭 )

src/webgpu/api/operation/sampling/anisotropy.spec.ts
kainino0x: > (which is assuming if device support anisotropy, it supports up to 16)

IIRC there are a few devices out there that only support anisotropy up to 4, unfortunately. Though I suppose we could check 1 and 4 instead of 16.

> What about expecting the result to be equal to either that of `maxAnisotropy = t.params.maxAnisotropy` or `maxAnisotropy = 1`?

I don't quite understand this, expecting the result of what to be equal to one of those? Isn't the "result" the case with maxAnisotropy=t.params.maxAnisotropy?

Unfortunately since there are no guarantees about the implementation of anisotropic filtering (and the fact that the Vulkan spec's reference implementation is really complicated), probably the best we can do here is warnings. Warnings would be strong enough to guarantee that WebGPU implementations are correct in automated testing - and it's not within-spec to strictly test hardware behavior.

If the tests can only produce warnings anyway, then it's okay to warn on hardware that implements anisotropic filtering poorly (e.g. <16, or in a substandard way) or not at all.

kainino0x (commented):
LGTM, only remaining questions are in the test expectations of anisotropic_filter_mipmap_color

src/webgpu/api/operation/sampling/anisotropy.spec.ts
```diff
@@ -6,14 +6,357 @@ Note that anisotropic filtering is never guaranteed to occur, but we might be ab
...
+    const textureSizeMipmap0 = 1 << (mipLevelCount - 1);
+    const texture = t.device.createTexture({
+      mipLevelCount,
+      size: { width: textureSizeMipmap0, height: textureSizeMipmap0, depth: 1 },
+      format: kTextureFormat,
+      usage: GPUTextureUsage.COPY_DST | GPUTextureUsage.SAMPLED,
+    });
+
+    const textureEncoder = t.device.createCommandEncoder();
+    // Populate each level with a different color
```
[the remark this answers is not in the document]
shrekshao: Add a createTexture2DWithMipmaps in gpu_test

src/webgpu/api/operation/sampling/anisotropy.spec.ts
shrekshao: Now testing maxAnisotropy = [1, 4]. And make the maxAnisotorpy=4 case to generate warning when fails
"Unimportant: I thought I mean when maxAnisotropy = 4, the result could be equal to that on 1 or 4 on a device with full support of anisotoropic rendering"

src/webgpu/gpu_test.ts
```diff
@@ -314,22 +425,14 @@ got [${failedByteActualValues.join(', ')}]`;
     this.expectContents(buffer, new Uint8Array(arrayBuffer));
   }
 
-  // TODO: Add check for values of depth/stencil, probably through sampling of shader
-  // TODO(natashalee): Can refactor this and expectSingleColor to use a similar base expect
-  expectSinglePixelIn2DTexture(
+  // return a GPUBuffer that data are going to be written into
+  private readSinglePixelFrom2DTexture(
```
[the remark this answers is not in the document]
shrekshao: Thanks for handling any potential conflicts

src/webgpu/api/operation/sampling/anisotropy.spec.ts
```diff
@@ -6,14 +6,310 @@ Note that anisotropic filtering is never guaranteed to occur, but we might be ab
...
+    A similar webgl demo is at https://jsfiddle.net/t8k7c95o/5/`
+  )
+  .params([
+    {
+      maxAnisotropy: 1,
+      _results: [
+        { coord: { x: xMiddle, y: 2 }, expected: colors[2] },
+        { coord: { x: xMiddle, y: 6 }, expected: [colors[0], colors[1]] },
+      ],
+      generateWarningOnly: false,
```
kainino0x, line 269: nit: Should be a private param (or maybe it should even be per-entry in the _results array)

src/webgpu/gpu_test.ts
```diff
@@ -147,7 +147,12 @@ export class GPUTest extends Fixture {
     return { dst, begin: offsetDifference, end: offsetDifference + size };
   }
 
-  expectContents(src: GPUBuffer, expected: TypedArrayBufferView, srcOffset: number = 0): void {
+  expectContents(
+    src: GPUBuffer,
+    expected: TypedArrayBufferView,
+    srcOffset: number = 0,
+    generateWarningOnly: boolean = false
```
kainino0x, line 154: nit: Would be nice as a named argument (in an options dictionary) so the caller always specifies what the boolean means.

kainino0x (approved):
LGTM with small nits and merge conflicts, I'll just fix them now. Thanks again for working on this test!

kenrussell:
Very nice work on this test @shrekshao ! It looks like it was challenging to get all of the pieces in place! Kudos @kainino0x for your reviews and help getting this landed!

PR #442, opened 2021-01-20
Add validation tests on the member 'aspect' in GPUTextureCopyView

Jiawei-Shao:
-----

- [ ] WebGPU readability
- [ ] TypeScript readability

### DIFF
--- a/src/webgpu/api/validation/encoding/cmds/copyTextureToTexture.spec.ts
+++ b/src/webgpu/api/validation/encoding/cmds/copyTextureToTexture.spec.ts
@@ -517,9 +517,9 @@ g.test('copy_aspects')
     params()
       .combine(
         poptions('formatAndValidAspects', [
-          { format: 'rgba8unorm', validAspects: ['all'] as const },
-          { format: 'depth24plus-stencil8', validAspects: ['all'] as const },
-          { format: 'depth32float', validAspects: ['all', 'depth-only'] as const },
+          { format: 'rgba8unorm', validAspects: ['all'] },
+          { format: 'depth24plus-stencil8', validAspects: ['all'] },
+          { format: 'depth32float', validAspects: ['all', 'depth-only'] },
         ] as const)
       )
       .combine(poptions('sourceAspect', ['all', 'depth-only', 'stencil-only'] as const))
--- a/src/webgpu/api/validation/encoding/cmds/copyTextureToTexture.spec.ts
+++ b/src/webgpu/api/validation/encoding/cmds/copyTextureToTexture.spec.ts
@@ -510,46 +510,55 @@ g.test('copy_within_same_texture')
     );
   });
 
-// TODO(jiawei.shao@intel.com): test 'stencil-only' is also valid for the texture format 'stencil8'
 g.test('copy_aspects')
   .desc(
     `
 Test the validations on the member 'aspect' of GPUTextureCopyView in CopyTextureToTexture().
 - for all the color and depth-stencil formats: the texture copy aspects must be both 'all'.
-- for all the depth-only formats: the texture copy aspects must be both 'all' or 'depth-only'.
+- for all the depth-only formats: the texture copy aspects must be either 'all' or 'depth-only'.
+- for all the stencil-only formats: the texture copy aspects must be either 'all' or 'stencil-only'.
 `
   )
   .params(
     params()
       .combine(
-        poptions('formatAndValidAspects', [
-          { format: 'rgba8unorm', _validAspects: ['all'] },
-          { format: 'depth24plus-stencil8', _validAspects: ['all'] },
-          { format: 'depth32float', _validAspects: ['all', 'depth-only'] },
+        poptions('format', [
+          'rgba8unorm',
+          'depth24plus-stencil8',
+          'depth32float',
+          'stencil8',
         ] as const)
       )
       .combine(poptions('sourceAspect', ['all', 'depth-only', 'stencil-only'] as const))
       .combine(poptions('destinationAspect', ['all', 'depth-only', 'stencil-only'] as const))
   )
   .fn(async t => {
-    const { formatAndValidAspects, sourceAspect, destinationAspect } = t.params;
+    const { format, sourceAspect, destinationAspect } = t.params;
 
     const kTextureSize = { width: 16, height: 8, depth: 1 };
 
     const srcTexture = t.device.createTexture({
       size: kTextureSize,
-      format: formatAndValidAspects.format,
+      format,
       usage: GPUTextureUsage.COPY_SRC,
     });
     const dstTexture = t.device.createTexture({
       size: kTextureSize,
-      format: formatAndValidAspects.format,
+      format,
       usage: GPUTextureUsage.COPY_DST,
     });
 
+    // TODO(jiawei.shao@intel.com): get the valid aspects from capability_info.ts.
+    const formatAndValidAspecs = {
+      rgba8unorm: ['all'],
+      'depth24plus-stencil8': ['all'],
+      depth32float: ['all', 'depth-only'],
+      stencil8: ['all', 'stencil-only'],
+    };
+
     let isSourceAspectValid = false;
     let isDestinationAspectValid = false;
-    for (const aspect of formatAndValidAspects._validAspects) {
+    for (const aspect of formatAndValidAspecs[format]) {
       if (sourceAspect === aspect) {
         isSourceAspectValid = true;
       }
--- a/src/webgpu/api/validation/encoding/cmds/copyTextureToTexture.spec.ts
+++ b/src/webgpu/api/validation/encoding/cmds/copyTextureToTexture.spec.ts
@@ -7,7 +7,6 @@ Test Plan: (TODO(jiawei.shao@intel.com): add tests on 1D/3D textures)
   - mipLevel {>, =, <} the mipmap level count of the {source, destination} texture.
   - the source texture is created {with, without} GPUTextureUsage::CopySrc.
   - the destination texture is created {with, without} GPUTextureUsage::CopyDst.
-  - the texture copy aspects must be both 'all'.
 * sample count
   - the sample count of the source texture {is, isn't} equal to the one of the destination texture
   - when the sample count is greater than 1:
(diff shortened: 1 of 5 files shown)


src/webgpu/api/validation/encoding/cmds/copyTextureToTexture.spec.ts
```diff
@@ -510,6 +511,38 @@ g.test('copy_within_same_texture')
     );
   });
 
+g.test('copy_aspects')
+  .params(
+    params()
+      .combine(poptions('format', ['rgba8unorm', 'depth24plus-stencil8'] as const))
+      .combine(poptions('sourceAspect', ['all', 'depth-only', 'stencil-only'] as const))
+      .combine(poptions('destinationAspect', ['all', 'depth-only', 'stencil-only'] as const))
```
Kangz, line 519: Can we also test that `depth-only` is valid for depth-only formats, and a TODO to test `stencil-only` is valid for `stencil8`?

Kangz (approved):
LGTM, one suggestion

src/webgpu/api/validation/encoding/cmds/copyTextureToTexture.spec.ts
Jiawei-Shao: In current Dawn using ```depth-only```  for ```depth32float``` will report a validation error:
 "Texture aspect must be "all" for texture to texture copies"

I will fix it in Dawn.

Kangz: Sounds good, but you just filing a bug is ok too (so you don't need to multi-task as much).

src/webgpu/api/validation/encoding/cmds/copyTextureToTexture.spec.ts
```diff
@@ -510,6 +512,51 @@ g.test('copy_within_same_texture')
     );
   });
 
+// TODO(jiawei.shao@intel.com): test 'stencil-only' is also valid for the texture format 'stencil8'
+g.test('copy_aspects')
```
kainino0x, line 516: This can be a follow-up, but let's please adopt the [new style](https://github.com/gpuweb/cts/blob/main/docs/intro/plans.md) and have a description for each test.

src/webgpu/api/validation/encoding/cmds/copyTextureToTexture.spec.ts
```diff
@@ -510,6 +512,51 @@ g.test('copy_within_same_texture')
...
+  .params(
+    params()
+      .combine(
+        poptions('formatAndValidAspects', [
+          { format: 'rgba8unorm', validAspects: ['all'] },
+          { format: 'depth24plus-stencil8', validAspects: ['all'] },
+          { format: 'depth32float', validAspects: ['all', 'depth-only'] },
+        ] as const)
+      )
```
kainino0x, line 525: nit: use something like this:
```suggestion
      .combine([
        { format: 'rgba8unorm', _validAspects: ['all'] },
        { format: 'depth24plus-stencil8', _validAspects: ['all'] },
        { format: 'depth32float', _validAspects: ['all', 'depth-only'] },
      ] as const)
```
The underscore makes it "private" (not part of the test path).

src/webgpu/api/validation/encoding/cmds/copyTextureToTexture.spec.ts
```diff
@@ -510,6 +512,51 @@ g.test('copy_within_same_texture')
     );
   });
 
+// TODO(jiawei.shao@intel.com): test 'stencil-only' is also valid for the texture format 'stencil8'
```
kainino0x, line 515: This would be easy to add to the test now, we just wouldn't be able to pass it, right? If so we should just add it.

src/webgpu/api/validation/encoding/cmds/copyTextureToTexture.spec.ts
```diff
@@ -510,6 +512,51 @@ g.test('copy_within_same_texture')
...
+    });
+    const dstTexture = t.device.createTexture({
+      size: kTextureSize,
+      format: formatAndValidAspects.format,
+      usage: GPUTextureUsage.COPY_DST,
+    });
+
+    let isSuccess = false;
+    for (const aspect of formatAndValidAspects.validAspects) {
+      if (sourceAspect === aspect && destinationAspect === aspect) {
```
kainino0x, line 547: This would compute isSuccess incorrectly e.g. for `format='depth32float';sourceAspect='all';destinationAspect='depth-only'`

Jiawei-Shao: I am not sure if in this situation we can use either ```all``` or ```depth-only``` on the depth-only formats? In current SPEC it is not very clear about this part.

kainino0x: The intent was that it would be valid. The set of aspects in the copy view must be the same, which is true here.

Jiawei-Shao: Oh. Sorry for my misunderstanding.

src/webgpu/api/validation/encoding/cmds/copyTextureToTexture.spec.ts
Jiawei-Shao: ```stencil8``` is not defined in gpuweb/types. I've just written the PR https://github.com/gpuweb/types/pull/39 to add it.

src/webgpu/api/validation/encoding/cmds/copyTextureToTexture.spec.ts
kainino0x: The change you made doesn't make _validAspects private, as there is still a parameter `formatAndValidAspects` with value `{'format':'rgba8unorm','_validAspects':['all']}`. You should be able to see this in the standalone runner, in the query strings of the cases.

Instead, I think you should do something like this which makes two different parameters, `format`, and `_validAspects` which is private.

src/webgpu/api/validation/encoding/cmds/copyTextureToTexture.spec.ts
kainino0x: Done and published a new version of the package, 0.0.34. Feel free to update the test in this PR or a followup.

src/webgpu/api/validation/encoding/cmds/copyTextureToTexture.spec.ts
kainino0x: Actually, come to think of it, the information of which aspects are valid for which format should ultimately live in capability_info.ts. No need to do it now, it'll be a bit messy (have to do it for every format). Just leave a TODO here saying the info should be pulled from capability_info once it's added there.

Kangz (approved):
Still LGTM but I'd like Kai to review the texture table stuff as I'm less familiar with it.

Jiawei-Shao:
We haven't supported Stencil8 in Chromium/Dawn yet. Shall we support it  before WebGPU OT?

Kangz:
No, it's an optional feature and would distract from more important work for OT. We'll just add a suppression for the test.

kainino0x:
> No, it's an optional feature and would distract from more important work for OT. We'll just add a suppression for the test.

It's not an optional feature but it's still something we don't really need for OT. We'll suppress it when it rolls into chromium.

src/webgpu/api/validation/encoding/cmds/copyTextureToTexture.spec.ts
```diff
@@ -510,46 +510,55 @@ g.test('copy_within_same_texture')
...
     });
     const dstTexture = t.device.createTexture({
       size: kTextureSize,
-      format: formatAndValidAspects.format,
+      format,
       usage: GPUTextureUsage.COPY_DST,
     });
 
+    // TODO(jiawei.shao@intel.com): get the valid aspects from capability_info.ts.
+    const formatAndValidAspecs = {
```
kainino0x, line 552: nit: kValidAspectsForFormat or something like that

src/webgpu/api/validation/encoding/cmds/copyTextureToTexture.spec.ts
```diff
@@ -510,46 +510,55 @@ g.test('copy_within_same_texture')
...
+      rgba8unorm: ['all'],
+      'depth24plus-stencil8': ['all'],
+      depth32float: ['all', 'depth-only'],
+      stencil8: ['all', 'stencil-only'],
+    };
+
     let isSourceAspectValid = false;
     let isDestinationAspectValid = false;
-    for (const aspect of formatAndValidAspects._validAspects) {
+    for (const aspect of formatAndValidAspecs[format]) {
```
kainino0x, line 561: I think you could now use e.g. `const isSourceAspectValid = formatAndValidAspects[format].includes(sourceAspect);` instead of this loop.

src/webgpu/api/validation/encoding/cmds/copyTextureToTexture.spec.ts
```diff
@@ -510,6 +510,70 @@ g.test('copy_within_same_texture')
...
+  )
+  .params(
+    params()
+      .combine(
+        poptions('format', [
+          'rgba8unorm',
+          'depth24plus-stencil8',
+          'depth32float',
+          'stencil8',
+        ] as const)
```
kainino0x, line 530: nit: May as well test every depth/stencil format in kDepthStencilFormats, in addition to rgba8unorm. Something like:

```suggestion
        poptions('format', [
          'rgba8unorm',
          ...kDepthStencilFormats,
        ] as const)
```

src/webgpu/util/texture/texel_data.ts
```diff
@@ -490,6 +490,17 @@ export const kTexelRepresentationInfo: {
...
+      componentOrder: [TexelComponent.Stencil],
+      encode: applyEach((n: number) => (assertInIntegerRange(n, 8, false), n), [
+        TexelComponent.Stencil,
+      ]),
+      decode: applyEach((n: number) => (assertInIntegerRange(n, 8, false), n), [
+        TexelComponent.Stencil,
+      ]),
+      componentInfo: { Stencil: { dataType: 'uint', bitLength: 8 } },
+      pack: components => packComponents([TexelComponent.Stencil], components, 8, 'uint'),
+    },
```
kainino0x, line 503: nit: This is a little hard to read due to being over-general (not to say that other code here isn't hard to read, but); we can make it more readable by using a similar pattern to the one in depth24plus-stencil8 just below here. Rough sketch:

```ts
    'depth24plus-stencil8': {
      componentOrder: [TexelComponent.Stencil],
      componentInfo: {
        Stencil: { dataType: 'uint', bitLength: 8 },
      },
      encode: components => {
        assert(components.Stencil !== undefined);
        assertInIntegerRange(components.Stencil, 8, false);
        return components;
      },
      decode: components => {
        assert(components.Stencil !== undefined);
        assertInIntegerRange(components.Stencil, 8, false);
        return components;
      },
      pack: components => packComponents([TexelComponent.Stencil], components, 8, 'uint'),
    },
```

kainino0x (approved):
Sorry for the delay, LGTM with nits

kainino0x (approved):
LGTM!

PR #443, opened 2021-01-21
Add ShaderValidationTest fixture

kainino0x:
And basic tests to make sure it's working.

-----

- [x] WebGPU readability
- [x] TypeScript readability

### DIFF
--- a/src/webgpu/shader/validation/shader_validation_test.ts
+++ b/src/webgpu/shader/validation/shader_validation_test.ts
@@ -1,0 +1,42 @@
+import { GPUTest } from '../../gpu_test.js';
+
+export class ShaderValidationTest extends GPUTest {
+  /**
+   * Add a test expectation for whether a createShaderModule call succeeds or not.
+   *
+   * @example
+   * ```ts
+   * t.expectCompileResult(true, `wgsl code`); // Expect success
+   * t.expectCompileResult(false, `wgsl code`); // Expect validation error with any error string
+   * t.expectCompileResult('v-0000', `wgsl code`); // Expect validation error containing 'v-0000'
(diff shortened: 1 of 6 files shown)


src/webgpu/shader/validation/shader_validation_test.ts
```diff
@@ -0,0 +1,42 @@
...
+   *
+   * @example
+   * ```ts
+   * t.expectCompileResult(true, `wgsl code`); // Expect success
+   * t.expectCompileResult(false, `wgsl code`); // Expect validation error with any error string
+   * t.expectCompileResult('v-0000', `wgsl code`); // Expect validation error containing 'v-0000'
+   * ```
+   */
+  expectCompileResult(result: boolean | string, code: string) {
+    // If no result is expected, we let the scope surrounding the test catch it.
```
austinEng, line 15: nit: perhaps this comment should be reversed: "if an error is expected, push a scope.."

austinEng (approved):
Does this rely on Chrome-specific behavior? i.e. I think the error messages are supposed to be surfaced through `GPUCompilationInfo` but we don't implement that. Or do the messages go to both places?

kainino0x:
> Does this rely on Chrome-specific behavior? i.e. I think the error messages are supposed to be surfaced through `GPUCompilationInfo` but we don't implement that. Or do the messages go to both places?

Ohh that's a good point, I totally forgot about that. Yes, that makes this chrome-specific behavior; the spec does not require the error message to appear in the validation error message.

kainino0x:
@austinEng WDYT of this approach?

austinEng:
Yup, still looks good

Issue #305, opened 2020-10-13
Consider "subcases"


kainino0x:
Right now we have `.params()` which generates "cases" from a "test" (runs the "test function" once for each case).

Each test case gets a unique identifier and shows up individually in test results (in /standalone/, cmdline, wpt, and the JSON result format). However sometimes this isn't really necessary - it's not important which case is failing unless you're actually digging deeply into it. For example, cases that just differ in buffer ranges.

Consider renaming `.params()` to `.cases()`, and adding a `.subcases()`\* (e.g. `g.test('...').desc('...').cases([...]).subcases(p => [...])`), which is basically like a `for (...) { t.debug(subcase_params); fn(subcase_params); }`. The result of the case would be the combined result of all of the subcases.

Ideally, it would still be possible to manually drill down into a subcase by specifying the subcase params (as printed by `t.debug()`), for developers fixing a bug in an implementation.

**\*Note:** .subcases() needs to be able to receive the parameterization from .cases() because sometimes it might want to depend on it (similar to `.expand()`).

kainino0x:
Likely an important use-case for this is parameterizations over hardware capabilities (e.g. having a test parameterized over all of the allowed sample counts, not just 1 and 4). In order to do this we would need to pass adapter capabilities into subcase creation.

kainino0x:
On the subcase "drill down" feature: https://github.com/gpuweb/cts/pull/444#issuecomment-767967546

> I don't think it's super critical. It's really only for a browser developer testing changes in their browser, and they should have a local copy of the cts anyway where they can just cut down the .subcases() list manually.

Kangz:
Isn't the subcase feature done?

kainino0x:
I kept this open for the "drill down" into subcases but at this point I don't think we need to add it. I think it would be pretty easy to add since Austin already implemented the meat of it for WPT but we're more likely to do a different rework to subcases.

PR #444, opened 2021-01-25
Adding subcases() to TestGroupBuilder

egalli:
First pass at issue #305.

We are still missing to the ability to drill down into subcases. It shouldn't be too hard to add, but there is no reason for new test to have to wait for it.

Since we moved to `.cases()` instead of `.params()`. I have tagged `.params()` with `@deprecated` as well as updated `example.spec.ts` and `test_group.spec.ts`,

-----

- [ ] WebGPU readability
- [ ] TypeScript readability

### DIFF
--- a/src/common/framework/test_group.ts
+++ b/src/common/framework/test_group.ts
@@ -1,6 +1,6 @@
 import { Fixture, SkipTestCase } from './fixture.js';
 import { TestCaseRecorder } from './logging/test_case_recorder.js';
-import { CaseParams, CaseParamsIterable, extractPublicParams } from './params_utils.js';
+import { CaseParams, extractPublicParams, Merged, mergeParams } from './params_utils.js';
 import { kPathSeparator } from './query/separators.js';
 import { stringifyPublicParams, stringifyPublicParamsUniquely } from './query/stringify_params.js';
 import { validQueryPart } from './query/validQueryPart.js';
@@ -20,7 +20,7 @@ export interface RunCase {
 
 // Interface for defining tests
 export interface TestGroupBuilder<F extends Fixture> {
-  test(name: string): TestBuilderWithName<F, never>;
+  test(name: string): TestBuilderWithName<F>;
 }
 export function makeTestGroup<F extends Fixture>(fixture: FixtureClass<F>): TestGroupBuilder<F> {
   return new TestGroup(fixture);
@@ -43,13 +43,15 @@ export function makeTestGroupForUnitTesting<F extends Fixture>(
   return new TestGroup(fixture);
 }
 
-type FixtureClass<F extends Fixture> = new (log: TestCaseRecorder, params: CaseParams) => F;
-type TestFn<F extends Fixture, P extends {}> = (t: F & { params: P }) => Promise<void> | void;
+type FixtureClass<F extends Fixture> = new <P extends {}>(log: TestCaseRecorder, params: P) => F;
+type TestFn<F extends Fixture, P extends {}, SubP extends {}> = (
+  t: F & { params: Merged<P, SubP> }
+) => Promise<void> | void;
 
 class TestGroup<F extends Fixture> implements TestGroupBuilder<F> {
   private fixture: FixtureClass<F>;
   private seen: Set<string> = new Set();
-  private tests: Array<TestBuilder<F, never>> = [];
+  private tests: Array<TestBuilder<F, {}, {}>> = [];
 
   constructor(fixture: FixtureClass<F>) {
     this.fixture = fixture;
@@ -72,7 +74,7 @@ class TestGroup<F extends Fixture> implements TestGroupBuilder<F> {
   }
 
   // TODO: This could take a fixture, too, to override the one for the group.
-  test(name: string): TestBuilderWithName<F, never> {
+  test(name: string): TestBuilderWithName<F> {
     const testCreationStack = new Error(`Test created: ${name}`);
 
     this.checkName(name);
@@ -82,7 +84,7 @@ class TestGroup<F extends Fixture> implements TestGroupBuilder<F> {
       assert(validQueryPart.test(p), `Invalid test name part ${p}; must match ${validQueryPart}`);
     }
 
-    const test = new TestBuilder<F, never>(parts, this.fixture, testCreationStack);
+    const test = new TestBuilder<F, {}, {}>(parts, this.fixture, testCreationStack);
     this.tests.push(test);
     return test;
   }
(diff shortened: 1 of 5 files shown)


src/common/framework/test_group.ts
```diff
@@ -223,8 +276,41 @@ class RunCaseSpecific<F extends Fixture> implements RunCase {
...
     }
+  }
+
+  async run(rec: TestCaseRecorder): Promise<void> {
+    rec.start();
+    if (this.subParamGen) {
+      let totalCount = 0;
+      let skipCount = 0;
+      for (const subParams of this.subParamGen(this.params)) {
+        rec.debug(new Error('subcase: ' + stringifyPublicParamsUniquely(subParams)));
```
kainino0x: nit: This should probably just JSON.stringify since there isn't the concept of a "query string" (or "public params") for subcases.

src/common/framework/test_group.ts
```diff
@@ -223,8 +276,41 @@ class RunCaseSpecific<F extends Fixture> implements RunCase {
...
+        try {
+          await this.runTest(rec, mergeParams(this.params, subParams), true);
+        } catch (ex) {
+          if (ex instanceof SkipTestCase) {
+            // Convert SkipTestCase to debug messages
```
kainino0x, line 286: Nice catch, I probably would have missed this :)

src/common/framework/test_group.ts
kainino0x: Perhaps we should have a different log level for this so these logs don't get hidden by default.

src/webgpu/examples.spec.ts
```diff
@@ -79,10 +79,11 @@ g.test('basic,async').fn(async t => {
 // They can also be private by starting with an underscore (_result), which passes
 // them into the test but does not make them part of the case name:
 //
-// - webgpu:examples:basic/params={"x":2,"y":4}    runs with t.params = {x: 2, y: 5, _result: 6}.
-// - webgpu:examples:basic/params={"x":-10,"y":18} runs with t.params = {x: -10, y: 18, _result: 8}.
-g.test('basic,params')
-  .params([
+// - webgpu:examples:basic/cases={"x":2,"y":4}    runs with t.params = {x: 2, y: 5, _result: 6}.
+// - webgpu:examples:basic/cases={"x":-10,"y":18} runs with t.params = {x: -10, y: 18, _result: 8}.
```
kainino0x: oops, this had outdated query syntax. Should fix that here.

src/common/framework/test_group.ts
```diff
@@ -94,23 +96,44 @@ class TestGroup<F extends Fixture> implements TestGroupBuilder<F> {
...
+  ): TestBuilderWithSubParams<F, NewSubP>;
 }
 
-interface TestBuilderWithParams<F extends Fixture, P extends {}> {
-  fn(fn: TestFn<F, P>): void;
+interface TestBuilderWithSubParams<F extends Fixture, SubP extends {}>
+  extends TestBuilderWithParamsAndSubParams<F, {}, SubP> {
+  /** @deprecated use cases() and subcases() instead */
+  params<NewP extends {}>(specs: Iterable<NewP>): TestBuilderWithParamsAndSubParams<F, NewP, SubP>;
+  cases<NewP extends {}>(specs: Iterable<NewP>): TestBuilderWithParamsAndSubParams<F, NewP, SubP>;
```
kainino0x: Once .subcases() has been called, I don't think it makes sense to furthe call .cases(). So the hierarchy can be simplified a bit.

src/common/framework/test_group.ts
```diff
@@ -94,23 +96,44 @@ class TestGroup<F extends Fixture> implements TestGroupBuilder<F> {
...
-interface TestBuilderWithName<F extends Fixture, P extends {}> extends TestBuilderWithParams<F, P> {
+interface TestBuilderWithName<F extends Fixture>
+  extends TestBuilderWithParamsAndSubParams<F, {}, {}> {
   desc(description: string): this;
+  /** @deprecated use cases() and subcases() instead */
   params<NewP extends {}>(specs: Iterable<NewP>): TestBuilderWithParams<F, NewP>;
+  cases<NewP extends {}>(specs: Iterable<NewP>): TestBuilderWithParams<F, NewP>;
+  subcases<NewSubP extends {}>(
+    specs: (_: {}) => Iterable<NewSubP>
+  ): TestBuilderWithSubParams<F, NewSubP>;
```
kainino0x: TestBuilderWithName could inherit .subcases() from another interface

src/common/framework/test_group.ts
```diff
@@ -20,7 +20,7 @@ export interface RunCase {
 
 // Interface for defining tests
 export interface TestGroupBuilder<F extends Fixture> {
-  test(name: string): TestBuilderWithName<F, never>;
```
kainino0x, line 23: This `never` usage of never (hah) really made sense, and I have considered several times trying to do something about it. Thanks for cleaning it up!

kainino0x (approved):
This looks great!! Awesome work diving into this really obtuse type metaprogramming mess :)

I had a few suggestions, but they're really small and I wanted to play with them locally before I made them, so I just implemented them and pushed to your branch. Would you mind reviewing my last commit before we land this?

src/common/framework/test_group.ts
kainino0x: ... Actually, thinking about this again, since we would like to add the "drill-down" functionality later, "query string" and "public params" do apply here. Reverted this.

kainino0x:
> We are still missing to the ability to drill down into subcases. It shouldn't be too hard to add, but there is no reason for new test to have to wait for it.

No worries, definitely best to separate that change out into a separate patch.

That said, I don't think it's super critical. It's really only for a browser developer testing changes in their browser, and they should have a local copy of the cts anyway where they can just cut down the .subcases() list manually. Depending on whether you've already started looking at it, I'd say we can just defer this particular feature until later.

src/common/framework/test_group.ts
```diff
@@ -223,8 +265,41 @@ class RunCaseSpecific<F extends Fixture> implements RunCase {
...
+      let skipCount = 0;
+      for (const subParams of this.subParamGen(this.params)) {
+        rec.info(new Error('subcase: ' + stringifyPublicParamsUniquely(subParams)));
+        try {
+          await this.runTest(rec, mergeParams(this.params, subParams), true);
+        } catch (ex) {
+          if (ex instanceof SkipTestCase) {
+            // Convert SkipTestCase to debug messages
+            ex.message = 'subcase skipped: ' + ex.message;
+            rec.debug(ex);
```
kainino0x, line 288: As agreed on chat, makes sense for this to be .info() as well.

PR #445, opened 2021-01-25
[wgsl] validation v-0033 - the initializer’s type must match the store type of the variable.

sarahM0:
Validating rule v-0033: 
If present, the initializer’s type must match the store type of the variable ([link](https://gpuweb.github.io/gpuweb/wgsl.html#variables)).

-----

- [ ] WebGPU readability
- [ ] TypeScript readability


src/webgpu/shader/validation/wgsl/variable.spec.ts
```diff
@@ -0,0 +1,198 @@
...
+      e1_type === e2_type &&
+      e2_type === e3_type
+    ) {
+      t.expectCompileResult(true, code);
+    } else {
+      t.expectCompileResult('v-0033', code);
+    }
+  });
+
+g.test('v_0033_vec3_e1_e2')
```
sarahM0, line 115: @kainino0x 
Is it possible to combine g.test('v_0033_vec3_e1_e2') with g.test('v_0033_vec3_e2_e1')?

src/webgpu/shader/validation/wgsl/variable.spec.ts
```diff
@@ -0,0 +1,198 @@
...
+
+// TODO(sarahM0): randomaize rhs_value
+const kTypeConstructorExpression = [
+  { rhs_type: 'i32', rhs_value: '0' },
+  { rhs_type: 'f32', rhs_value: '0.0' },
+  { rhs_type: 'u32', rhs_value: '0u' },
+  { rhs_type: 'bool', rhs_value: 'false' },
+] as const;
+
+g.test('v_0033_scalar')
```
dj2, line 21: Does `v_0033_scalar` end up being the test name? If so, can we make it more descriptive?

`v_0033_validate_lhs_matches_rhs_in_initializer` or something? (similar for test strings below)

src/webgpu/shader/validation/wgsl/variable.spec.ts
```diff
@@ -0,0 +1,198 @@
...
+  )
+  .fn(async t => {
+    const { lhs_type, rhs_type, e1_type, e2_type } = t.params;
+    const code = `
+
+    [[stage(vertex)]]
+    fn main() -> void {
+      var e1: ${e1_type} = ${e1_type}();
+      var e2: ${e2_type} = ${e2_type}();
+      var v : vec2<${lhs_type}> = vec2<${rhs_type}>(e1, e2);
```
dj2, line 54: Aren't there multiple possible errors in here? If `e1_type` or `e2_type` are not the same as `rhs_type` then that's one error and if `rhs_type` is not the same as `lhs_type`?

kainino0x (commented):
Here is a rough sketch of how JUST the v-0033 part of this test could possibly work:
https://github.com/sarahM0/cts/compare/cts-example...kainino0x:v0033-sketch

The other part, testing vector constructors, I didn't try to tackle yet - it's harder, though I think is the part you were asking about.

src/webgpu/shader/validation/v_0033.spec.ts
```diff
@@ -0,0 +1,72 @@
+export const description = `
+Tests for validation rule v-0033.
```
dj2, line 2: Can we expand this to list what validation rule v-0033 is so we don't have to go lookup in the spec?

src/webgpu/shader/validation/v_0033.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+import { assert } from '../../../common/framework/util/util.js';
+
+import { ShaderValidationTest } from './shader_validation_test.js';
+
+export const g = makeTestGroup(ShaderValidationTest);
+
+type ScalarType = 'f32' | 'i32' | 'u32' | 'bool';
+const kScalarType = ['i32', 'f32', 'u32', 'bool'] as const;
+
+function getVecTypeInfo(scalarType: ScalarType, x: number, y: number) {
```
dj2, line 16: This is more like `getTypeInfo` as it's vec, or scalar or matrix.

src/webgpu/shader/validation/v_0033.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  let type: string = '';
+  if (y === 1) {
+    if (x === 1) {
+      type = scalarType;
+    } else {
+      type = `vec${x}<${scalarType}>`;
+    }
+  } else {
+    type = `mat${y}x${x}<${scalarType}>`;
```
dj2, line 28: Why ${y}x${x} instead of ${x}x${y}?

src/webgpu/shader/validation/v_0033.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+      lhs_x,
+      lhs_y,
+      rhs_x,
+      rhs_y,
+    } = t.params;
+
+    const lhsType = getVecTypeInfo(lhsScalarType, lhs_x, lhs_y);
+    const rhsType = getVecTypeInfo(rhsScalarType, rhs_x, rhs_y);
+
+    t.debug(`${variable_or_constant} a : ${lhsType} = ${rhsType}()`);
```
dj2, line 61: Is this to be left in? (I don't know what t.debug does)

kainino0x: A t.debug() entry shows up in the logs if you pass debug=1 to the /standalone/ runner.

It would be okay to leave in, but it should probably be removed, I just added it so I could see what was running.

src/webgpu/shader/validation/v_0033.spec.ts
kainino0x: It isn't actually in the spec yet so I didn't know what to write here, but +1

src/webgpu/shader/validation/v_0033.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+import { assert } from '../../../common/framework/util/util.js';
+
+import { ShaderValidationTest } from './shader_validation_test.js';
+
+export const g = makeTestGroup(ShaderValidationTest);
+
+type ScalarType = 'f32' | 'i32' | 'u32' | 'bool';
+const kScalarType = ['i32', 'f32', 'u32', 'bool'] as const;
+
+function getTypeInfo(scalarType: ScalarType, x: number, y: number) {
```
kainino0x, line 17: uNit: Since you've simplified this to not have to return the zero values (nice! I totally didn't think of that...), a better name could just be `getType`, `getTypeString`, `getTypeForMatrixVectorScalar`, or something like that.

OTOH, this function could go away; see below.

src/webgpu/shader/validation/v_0033.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  let type: string = '';
+  if (y === 1) {
+    if (x === 1) {
+      type = scalarType;
+    } else {
+      type = `vec${x}<${scalarType}>`;
+    }
+  } else {
+    type = `mat${x}x${y}<${scalarType}>`;
```
kainino0x, line 29: If y==2 and x==1 then I think this generates an invalid type name. For simplicity, you could simply generate a vec in that case even though it would be duplicated with other cases. (You could put an assert, but it would  make the .params() too complicated.)

An alternative idea, though: Now that this function is much simpler, it's easy to enumerate all of the possible vec/mat types

```ts
const kContainerTypes = [undefined, 'vec2', 'vec3', 'vec4', 'mat2x2', 'mat2x3', 'mat2x4', 'mat3x2', 'mat3x3', 'mat3x4', 'mat4x2', 'mat4x3', 'mat4x4'];
```

With that you could do something like the suggestions below...

src/webgpu/shader/validation/v_0033.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  assert(type.length !== 0, 'type is not set');
+  return type;
+}
+
+g.test('wgsl-v-0033')
+  .desc(`v-033: If present, the initializer’s type must match the store type of the variable.`)
+  .params(
+    params()
+      .combine(poptions('variable_or_constant', ['var', 'const']))
```
kainino0x, line 40: nit: We usually use camelCase since these correspond to variable names below.

src/webgpu/shader/validation/v_0033.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+      .combine(poptions('lhsScalarType', kScalarType))
+      .combine(poptions('rhsScalarType', kScalarType))
+      .combine(poptions('lhs_x', [1, 2, 3, 4] as const))
+      .combine(poptions('lhs_y', [1, 2, 3, 4] as const))
+      .combine(poptions('rhs_x', [1, 2, 3, 4] as const))
+      .combine(poptions('rhs_y', [1, 2, 3, 4] as const))
```
kainino0x, line 46:
```suggestion
      .combine(poptions('variableOrConstant', ['var', 'const']))
      .combine(poptions('lhsContainerType', kContainerTypes))
      .combine(poptions('lhsScalarType', kScalarType))
      .combine(poptions('rhsContainerType', kContainerTypes))
      .combine(poptions('rhsScalarType', kScalarType))
```

src/webgpu/shader/validation/v_0033.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+      lhsScalarType,
+      rhsScalarType,
+      lhs_x,
+      lhs_y,
+      rhs_x,
+      rhs_y,
+    } = t.params;
+
+    const lhsType = getTypeInfo(lhsScalarType, lhs_x, lhs_y);
+    const rhsType = getTypeInfo(rhsScalarType, rhs_x, rhs_y);
```
kainino0x, line 60:
```suggestion
      variable_or_constant,
      lhsContainerType,
      lhsScalarType,
      rhsContainerType,
      rhsScalarType,
    } = t.params;

    const lhsType = lhsContainerType ? `${lhsContainerType}<${lhsScalarType}>` : lhsScalarType;
    const rhsType = rhsContainerType ? `${rhsContainerType}<${rhsScalarType}>` : rhsScalarType;
```

(Unfortunately we can't very easily just define the container types as templates - but it could be done by delegating to a helper function instead of doing it inline here. So if we wanted to test other container types, say, `array<>`, this test could be generalized.)

kainino0x (commented):
Test looks good (though there's one thing I think might not behave properly), though I had some ideas below if you are interested.

src/webgpu/shader/validation/v_0033.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+  } else {
+    type = `mat${x}x${y}<${scalarType}>`;
+  }
+
+  assert(type.length !== 0, 'type is not set');
+  return type;
+}
+
+g.test('wgsl-v-0033')
+  .desc(`v-033: If present, the initializer’s type must match the store type of the variable.`)
```
kainino0x, line 37: typo in v-033.

nit: No need to be redundant with the filename and file description, IMO.

```suggestion
g.test('scalar_vec_mat')
  .desc(`Tests for v-0033 with scalars, vectors, and matrices of every dimension and inner type.`)
```

src/webgpu/shader/validation/variable_and_const.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+    }
+    default: {
+      type = containerType ? `${containerType}<${scalarType}>` : scalarType;
+      break;
+    }
+  }
+  return type;
+}
+
+// TODO(sarahM0): v-0033: test structs
```
kainino0x, line 68: (not a comment, just a discussion)

This came up in this week's call but I don't know the answer: are structs nominal or structural types? That is, is the following valid or invalid?:

```
struct A { x: i32; }
struct B { x: i32; }

const a: A = ...;
const b: B = a;
```

If it's valid (structs are structural), then we should extend the test here. If it's invalid (structs are nominal), it would make most sense to have a separate test for this that is just like what I wrote above.

src/webgpu/shader/validation/variable_and_const.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+      break;
+    }
+    case 'array': {
+      // TODO(sarahM0): 12 is a random number here. find a solution to replace it.
+      // TODO(sarahM0): test array of vectors and matrices. eg. array<vec4, 9>
+      type = `array<${scalarType}, 12>`;
+      break;
+    }
+    default: {
+      type = containerType ? `${containerType}<${scalarType}>` : scalarType;
```
kainino0x, line 61: nit: `containerType ?` is now redundant since you have the (clearer) `case undefined:` above.

src/webgpu/shader/validation/variable_and_const.spec.ts
kainino0x, line 68: nit: Please put this TODO in either the test description or the file description so that it'll be visible at runtime (via /standalone/)

src/webgpu/shader/validation/variable_and_const.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+      break;
+    }
+  }
+  return type;
+}
```
kainino0x, line 66: Do you think we'd be likely to reuse this in other tests? If so, we could move it out to another file. (But we can also just defer and do that later.)

kainino0x (approved):
New getType looks nice. LGTM with tiny nits. Feel free to squash+merge when you're ready.

Issue #388, opened 2020-12-01
Make expectContents take a "validator" function


kainino0x:
Update expectContents (and expectTextureContents?) to take a validator function of the form `(index, value) -> boolean` or `(x, y, z, value) -> boolean`.

For expectTextureContents, it would probably be called once per texel. But for expectContents, it might be necessary to also tell it the chunk size (e.g. you might want bytes, or floats, or float4s).

This is a much more general way of testing buffer/texture contents than having to provide an ArrayBuffer like now.
- Handles expecting multiple possible values
- Avoids having to write expected values out into an ArrayBuffer
- Makes a lot more sense for texture expectations than writing strided values into an ArrayBuffer

Credit: @Kangz in https://github.com/gpuweb/cts/pull/384#discussion_r533391453

PR #446, opened 2021-01-26
Loosen precision in multisample resolve operation test

kainino0x:
Some platforms produce 0x7f and others 0x80 when a pixel is
half-covered.

I'd like to do this with #388 but that's a bigger project, so here's a more targeted fix.

-----

- [ ] New helpers, if any, are documented in `helper_index.md`. **Left undocumented because none of the siblings are documented and I expect this to go away with #388 anyway.**

- [x] WebGPU readability
- [x] TypeScript readability

### DIFF
--- a/src/webgpu/gpu_test.ts
+++ b/src/webgpu/gpu_test.ts
@@ -444,8 +444,7 @@ got [${failedByteActualValues.join(', ')}]`;
     src: GPUTexture,
     format: SizedTextureFormat,
     { x, y }: { x: number; y: number },
-    slice = 0,
-    layout?: TextureLayoutOptions
+    { slice = 0, layout }: { slice?: number; layout?: TextureLayoutOptions }
   ): GPUBuffer {
     const { byteLength, bytesPerRow, rowsPerImage, mipSize } = getTextureCopyLayout(
       format,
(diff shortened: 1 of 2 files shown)


kainino0x:
Actually, adding @bjjones for review instead, since you wrote the original test.

bjjones (approved):
Looks good. 

Just a thought to consider: instead of allowing two valid values - just giving an expected value and allowing +/- 1 or whatever tolerance might be better. If this test is expanded in the future (like more formats) you wouldn't have to worry about minor differences between devices.

src/webgpu/gpu_test.ts
```diff
@@ -314,22 +314,12 @@ got [${failedByteActualValues.join(', ')}]`;
     this.expectContents(buffer, new Uint8Array(arrayBuffer));
   }
 
-  // TODO: Add check for values of depth/stencil, probably through sampling of shader
-  // TODO(natashalee): Can refactor this and expectSingleColor to use a similar base expect
-  expectSinglePixelIn2DTexture(
+  makeBufferFromTextureSinglePixel(
```
kainino0x, line 317: oops, Shrek made the same refactor in #437. I'll rebase this on that when it lands.

kainino0x:
> Just a thought to consider: instead of allowing two valid values - just giving an expected value and allowing +/- 1 or whatever tolerance might be better. If this test is expanded in the future (like more formats) you wouldn't have to worry about minor differences between devices.

Thanks for the review! Since Shrek just added a helper to check if a pixel is between two values, I used that. Makes it easier to loosen in the future if needed - and good point that it would be needed for different texture formats.

PR #448, opened 2021-01-28
[wgsl] Add validation test - v-0038: A module scope variable in the in or out storage class must be IO-shareable.

sarahM0:
Tests for validation rule v-0038:
  The following types are IO-shareable:
  - numeric scalar types
  - numeric vector types
  - Matrix Types
  - Array Types if its element type is IO-shareable, and the array is not runtime-sized
  - Structure Types if all its members are IO-shareable

  As a result these are not IO-shareable:
  - boolean
  - vector of booleans
  - array of booleans
  - matrix of booleans
  - array runtime sized -> cannot be used outside of a struct, so no cts for this
  - struct with bool component
  - struct with runtime array
 
  TODO: add test for: struct - struct with bool component - struct with runtime array

-----

- [ ] WebGPU readability
- [ ] TypeScript readability

### DIFF
--- a/src/webgpu/shader/validation/variable_and_const.spec.ts
+++ b/src/webgpu/shader/validation/variable_and_const.spec.ts
@@ -108,43 +108,42 @@ g.test('v_0038')
   .desc(
     `Tests for validation rule v-0038:
   The following types are IO-shareable:
-  numeric scalar types
-  numeric vector types
-  Matrix Types
-  Array Types if its element type is IO-shareable, and the array is not runtime-sized
-  Structure Types if all its members are IO-shareable
+  - numeric scalar types
+  - numeric vector types
+  - Matrix Types
+  - Array Types if its element type is IO-shareable, and the array is not runtime-sized
+  - Structure Types if all its members are IO-shareable
 
   As a result these are not IO-shareable:
-  boolean
-  vector of booleans
-  array of booleans
-  matrix of booleans
-  array runtime sized -> cannot be used outside of a struct, so no cts for this
-  struct with bool component
-  struct with runtime array
-  TODO: add test for structs`
+  - boolean
+  - vector of booleans
+  - array of booleans
+  - matrix of booleans
+  - array runtime sized -> cannot be used outside of a struct, so no cts for this
+  - struct with bool component
+  - struct with runtime array
+
+  Control case: 'private' is used to make sure when only the storage class changes, the shader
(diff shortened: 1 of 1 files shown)


src/webgpu/shader/validation/variable_and_const.spec.ts
```diff
@@ -103,3 +103,48 @@ g.test('v_0033')
...
+  struct with bool component
+  struct with runtime array
+  TODO: add test for structs`
+  )
+  .params(
+    params()
+      .combine([
+        { variableOrConstant: 'var', inputOrOutput: 'in' },
+        { variableOrConstant: 'var', inputOrOutput: 'out' },
+        { variableOrConstant: 'const', inputOrOutput: 'out' },
```
kainino0x, line 132: I don't think `const` can have a storage class (variable_storage_decoration)?
(In which case you can un-parameterize over variableOrConstant.)

src/webgpu/shader/validation/variable_and_const.spec.ts
```diff
@@ -103,3 +103,48 @@ g.test('v_0033')
...
+  matrix of booleans
+  array runtime sized -> cannot be used outside of a struct, so no cts for this
+  struct with bool component
+  struct with runtime array
+  TODO: add test for structs`
+  )
+  .params(
+    params()
+      .combine([
+        { variableOrConstant: 'var', inputOrOutput: 'in' },
```
kainino0x, line 130: nit: inputOrOutput could be storageClass

src/webgpu/shader/validation/variable_and_const.spec.ts
```diff
@@ -103,3 +103,48 @@ g.test('v_0033')
...
+        { variableOrConstant: 'var', inputOrOutput: 'out' },
```
kainino0x, line 131: It would be ideal to have a "control case" to make sure that when *only* the storage class changes, the shader becomes invalid (to make sure nothing else is wrong).

Here I think the control case could be `var<private>` - all test cases should be valid (I think) with storage class private.

kainino0x (commented):
Looks good overall, maybe @dj2 can help with details

src/webgpu/shader/validation/variable_and_const.spec.ts
dj2: Yea, no storage class for const.

dj2 (approved):

PR #449, opened 2021-01-29
Update defaultQueue->queue, OUTPUT_ATTACHMENT->RENDER_ATTACHMENT to match latest spec

toji:
Simple update of any  instance of `defaultQueue` to `queue`. This will bring the CTS inline with the most recent spec changes and, once it rolls, significantly reduce the amount of deprecation warnings spamming the Chrome build bots. :)

-----

- [x] WebGPU readability
- [ ] TypeScript readability

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1240,9 +1240,9 @@
       }
     },
     "@webgpu/types": {
-      "version": "0.0.34",
-      "resolved": "https://registry.npmjs.org/@webgpu/types/-/types-0.0.34.tgz",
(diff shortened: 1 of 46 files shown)


toji:
Hm... What does the CI run against and how do we update it to something that recognizes the new attribute name?

Kangz:
It uses https://github.com/gpuweb/types that should be updated and rolled in the CTS.

kainino0x:
We can do this as an incremental fix to the types, but we're also really overdue for a full update to match the latest spec.

kainino0x (commented):
LGTM but will need types update

toji:
Also added the matUTPUT_ATTACHMENT->RENDER_ATTACHMENT to this change to match the corresponding change in https://github.com/gpuweb/types/pull/41

kainino0x (approved):

PR #451, opened 2021-02-02
mipLevelCount and arrayLayerCount should not be 0

Richard-Yunchao:
We should report errors if mipLevelCount and/or arrayLayerCount is 0. 

If we want to use the default behavior, just leave them undefined.

### DIFF
--- a/src/webgpu/api/validation/createView.spec.ts
+++ b/src/webgpu/api/validation/createView.spec.ts
@@ -101,11 +101,33 @@ g.test('creating_texture_view_on_a_2D_non_array_texture')
     // mip level is in range
     { mipLevelCount: 1, baseMipLevel: MIP_LEVEL_COUNT - 1, _success: true },
     { mipLevelCount: 2, baseMipLevel: MIP_LEVEL_COUNT - 2, _success: true },
-    // baseMipLevel == k && mipLevelCount == 0 means to use levels k..end
-    { mipLevelCount: 0, baseMipLevel: 0, _success: true },
-    { mipLevelCount: 0, baseMipLevel: 1, _success: true },
-    { mipLevelCount: 0, baseMipLevel: MIP_LEVEL_COUNT - 1, _success: true },
-    { mipLevelCount: 0, baseMipLevel: MIP_LEVEL_COUNT, _success: false },
+    // mipLevelCount == 0 is invalid
+    { mipLevelCount: 0, baseMipLevel: 0, _success: false },
+    // baseMipLevel == k && mipLevelCount == undefined means to use levels k..end
+    {
+      mipLevelCount: undefined,
+      baseMipLevel: 0,
+      _mipLevelCountKeepUndefined: true,
+      _success: true,
+    },
+    {
+      mipLevelCount: undefined,
+      baseMipLevel: 1,
+      _mipLevelCountKeepUndefined: true,
+      _success: true,
+    },
+    {
+      mipLevelCount: undefined,
+      baseMipLevel: MIP_LEVEL_COUNT - 1,
+      _mipLevelCountKeepUndefined: true,
+      _success: true,
+    },
+    {
+      mipLevelCount: undefined,
+      baseMipLevel: MIP_LEVEL_COUNT,
+      _mipLevelCountKeepUndefined: true,
+      _success: false,
+    },
     // it is an error to make the mip level out of range
     { mipLevelCount: MIP_LEVEL_COUNT + 1, baseMipLevel: 0, _success: false },
     { mipLevelCount: MIP_LEVEL_COUNT, baseMipLevel: 1, _success: false },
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/createView.spec.ts
```diff
@@ -101,19 +97,48 @@ g.test('creating_texture_view_on_a_2D_non_array_texture')
...
-    // baseMipLevel == k && mipLevelCount == 0 means to use levels k..end
-    { mipLevelCount: 0, baseMipLevel: 0, _success: true },
-    { mipLevelCount: 0, baseMipLevel: 1, _success: true },
-    { mipLevelCount: 0, baseMipLevel: MIP_LEVEL_COUNT - 1, _success: true },
-    { mipLevelCount: 0, baseMipLevel: MIP_LEVEL_COUNT, _success: false },
+    // mipLevelCount == 0 is invalid
+    { mipLevelCount: 0, baseMipLevel: 0, _success: false },
+    // baseMipLevel == k && mipLevelCount == undefined means to use levels k..end
+    {
+      mipLevelCount: undefined,
```
Richard-Yunchao, line 104: mipLevleCount and arrayLayerCount below are not useful. The explicit assignment as undefined is used to be displayed on the test webpage.

src/webgpu/api/validation/createView.spec.ts
```diff
@@ -101,19 +97,48 @@ g.test('creating_texture_view_on_a_2D_non_array_texture')
...
+      baseMipLevel: 0,
+      _mipLevelCountKeepUndefined: true,
```
Kangz, line 106: I don't know Typescript so well, but can't we just skip this variable and have mipLevelCount: undefined mean that?

src/webgpu/api/validation/createView.spec.ts
```diff
@@ -124,34 +149,62 @@ g.test('creating_texture_view_on_a_2D_non_array_texture')
...
-    { arrayLayerCount: 0, baseArrayLayer: 1, _success: true },
-    { arrayLayerCount: 0, baseArrayLayer: ARRAY_LAYER_COUNT_2D - 1, _success: true },
-    { arrayLayerCount: 0, baseArrayLayer: ARRAY_LAYER_COUNT_2D, _success: false },
+    // arrayLayerCount == 0 is invalid.
+    { arrayLayerCount: 0, baseArrayLayer: 0, _success: false },
+    // baseArrayLayer == k && arrayLayerCount == undefined means to use layers k..end.
+    {
+      arrayLayerCount: undefined,
+      baseArrayLayer: 0,
+      _arrayLayerCountKeepUndefined: true,
```
Kangz, line 173: ditto

Kangz (approved):

src/webgpu/api/validation/createView.spec.ts
Richard-Yunchao: Hmm, the helper function above explicitly assigns values to mipLevelCount and arrayLayerCount.

src/webgpu/api/validation/createView.spec.ts
Richard-Yunchao: same

src/webgpu/api/validation/createView.spec.ts
Kangz: Which helper function? `params`?

Richard-Yunchao: GetDescriptor() at line 59.

Kangz: Ah, got it thanks! Either way sounds find. Maybe @kainino0x you have a suggestion to make this more idiomatic?

Richard-Yunchao: Thanks for your detailed suggestion, Kai. I looked further for the current validation tests for createTexture and creatView. And I totally agree with you that the tests are very old and we need to add a lot more tests and we may need significant changes to these two tests. I will revisit the tests and revise them incrementally (at the same time, I will work on 3D texture feature implementation in Dawn project, like copy from/to 3D texture, etc.), not just add a couple of random tests for 3D texture validation. Let's discuss the issues you just mentioned in the coming PRs. 

And I'd like to close this PR.

kainino0x: I don't think it's necessary to redo everything to incrementally fix these outdated cases, but on the other hand maybe we want to redesign the testing approach anyway. Either way is OK with me - if we redesign, we should consider both validation and operation testing at once, have a test plan review, and try to make sure it's covering all the same stuff.
We could also design new tests but keep these old tests; test duplication doesn't really hurt. If we want to keep these then we'll have to do incremental updates to make them correct (like the suggested code block above).

Richard-Yunchao: Thanks for the suggestion. I looked into the existing test plan for createTexture, it is good. I didn't find significant missing parts. But the tests have not been added yet. The tests lag far behind.

PR #455, opened 2021-02-04
Add validation tests for mipLevels for texture creation

Richard-Yunchao:
-----

- [x] WebGPU readability
- [x] TypeScript readability

### DIFF
--- a/src/webgpu/api/validation/createTexture.spec.ts
+++ b/src/webgpu/api/validation/createTexture.spec.ts
@@ -33,6 +33,7 @@ TODO: move destroy tests out of this file
 import { poptions, params } from '../../../common/framework/params_builder.js';
 import { makeTestGroup } from '../../../common/framework/test_group.js';
 import { kAllTextureFormats, kAllTextureFormatInfo } from '../../capability_info.js';
+import { maxMipLevelCount } from '../../util/texture/base.js';
 
 import { ValidationTest } from './validation_test.js';
 
@@ -165,37 +166,37 @@ g.test('mipLevelCount,bound_check')
     TODO: test compressed texture`
   )
   .subcases(() => [
-    { size: [32, 32], mipLevelCount: 6, _success: true }, // Mip level sizes: 32x32, 16x16, 8x8, 4x4, 2x2, 1x1
-    { size: [31, 32], mipLevelCount: 6, _success: true }, // Mip level sizes: 31x32, 15x16, 7x8, 3x4, 1x2, 1x1
-    { size: [32, 31], mipLevelCount: 6, _success: true }, // Mip level sizes: 32x31, 16x15, 8x7, 4x3, 2x1, 1x1
-    { size: [31, 32], mipLevelCount: 7, _success: false }, // Mip level sizes: 31x32, 15x16, 7x8, 3x4, 1x2, 1x1, ?x?
-    { size: [32, 31], mipLevelCount: 7, _success: false }, // Mip level sizes: 32x31, 16x15, 8x7, 4x3, 2x1, 1x1, ?x?
-    { size: [31, 31], mipLevelCount: 5, _success: true }, // Mip level sizes: 31x31, 15x15, 7x7, 3x3, 1x1
-    { size: [31, 31], mipLevelCount: 6, _success: false }, // Mip level sizes: 31x31, 15x15, 7x7, 3x3, 1x1, ?x?
-    { size: [32], dimension: '1d' as const, mipLevelCount: 6, _success: true }, // Mip level sizes: 32, 16, 8, 4, 2, 1
-    { size: [32], dimension: '1d' as const, mipLevelCount: 7, _success: false }, // Mip level sizes: 32, 16, 8, 4, 2, 1, ?
-    { size: [31], dimension: '1d' as const, mipLevelCount: 5, _success: true }, // Mip level sizes: 31, 15, 7, 3, 1
-    { size: [31], dimension: '1d' as const, mipLevelCount: 6, _success: false }, // Mip level sizes: 31, 15, 7, 3, 1, ?
-    { size: [32, 32, 32], dimension: '3d' as const, mipLevelCount: 6, _success: true }, // Mip level sizes: 32x32x32, 16x16x16, 8x8x8, 4x4x4, 2x2x2, 1x1x1
-    { size: [32, 32, 32], dimension: '3d' as const, mipLevelCount: 7, _success: false }, // Mip level sizes: 32x32x32, 16x16x16, 8x8x8, 4x4x4, 2x2x2, 1x1x1, ?x?x?
-    { size: [32, 31, 31], dimension: '3d' as const, mipLevelCount: 6, _success: true }, // Mip level sizes: 32x31x31, 16x15x15, 8x7x7, 4x3x3, 2x1x1, 1x1x1
-    { size: [32, 31, 31], dimension: '3d' as const, mipLevelCount: 7, _success: false }, // Mip level sizes: 32x31x31, 16x15x15, 8x7x7, 4x3x3, 2x1x1, 1x1x1, ?x?x?
-    { size: [31, 32, 31], dimension: '3d' as const, mipLevelCount: 6, _success: true }, // Mip level sizes: 31x32x31, 15x16x15, 7x8x7, 3x4x3, 1x2x1, 1x1x1
-    { size: [31, 32, 31], dimension: '3d' as const, mipLevelCount: 7, _success: false }, // Mip level sizes: 31x32x31, 15x16x15, 7x8x7, 3x4x3, 1x2x1, 1x1x1, ?x?x?
-    { size: [31, 31, 32], dimension: '3d' as const, mipLevelCount: 6, _success: true }, // Mip level sizes: 31x31x32, 15x15x16, 7x7x8, 3x3x4, 1x1x2, 1x1x1
-    { size: [31, 31, 32], dimension: '3d' as const, mipLevelCount: 7, _success: false }, // Mip level sizes: 31x31x32, 15x15x16, 7x7x8, 3x3x4, 1x1x2, 1x1x1, ?x?x?
-    { size: [31, 31, 31], dimension: '3d' as const, mipLevelCount: 5, _success: true }, // Mip level sizes: 31x31x31, 15x15x15, 7x7x7, 3x3x3, 1x1x1
-    { size: [31, 31, 31], dimension: '3d' as const, mipLevelCount: 6, _success: false }, // Mip level sizes: 31x31x31, 15x15x15, 7x7x7, 3x3x3, 1x1x1, ?x?x?
-    { size: [32, 32], mipLevelCount: 100, _success: false }, // undefined shift check if miplevel is bigger than the integer bit width
-    { size: [32, 8], mipLevelCount: 6, _success: true }, // Mip levels: 32x8, 16x4, 8x2, 4x1, 2x1, 1x1
-    { size: [32, 8], mipLevelCount: 7, _success: true }, // Mip levels: 32x8, 16x4, 8x2, 4x1, 2x1, 1x1
-    { size: [32, 32, 64], mipLevelCount: 7, _success: false }, // Mip levels: 32x32x64, 16x16x64, 8x8x64, 4x4x64, 2x2x64, 1x1x64, ?x?x64
-    { size: [32, 32, 64], dimension: '3d' as const, mipLevelCount: 7, _success: true }, // Mip levels: 32x32x64, 16x16x32, 8x8x16, 4x4x8, 2x2x4, 1x1x2, 1x1x1
+    { size: [32, 32], mipLevelCount: 6 }, // Mip level sizes: 32x32, 16x16, 8x8, 4x4, 2x2, 1x1
+    { size: [31, 32], mipLevelCount: 6 }, // Mip level sizes: 31x32, 15x16, 7x8, 3x4, 1x2, 1x1
+    { size: [32, 31], mipLevelCount: 6 }, // Mip level sizes: 32x31, 16x15, 8x7, 4x3, 2x1, 1x1
+    { size: [31, 32], mipLevelCount: 7 }, // Mip level sizes: 31x32, 15x16, 7x8, 3x4, 1x2, 1x1, ?x?
+    { size: [32, 31], mipLevelCount: 7 }, // Mip level sizes: 32x31, 16x15, 8x7, 4x3, 2x1, 1x1, ?x?
+    { size: [31, 31], mipLevelCount: 5 }, // Mip level sizes: 31x31, 15x15, 7x7, 3x3, 1x1
+    { size: [31, 31], mipLevelCount: 6 }, // Mip level sizes: 31x31, 15x15, 7x7, 3x3, 1x1, ?x?
+    { size: [32], dimension: '1d' as const, mipLevelCount: 6 }, // Mip level sizes: 32, 16, 8, 4, 2, 1
+    { size: [32], dimension: '1d' as const, mipLevelCount: 7 }, // Mip level sizes: 32, 16, 8, 4, 2, 1, ?
+    { size: [31], dimension: '1d' as const, mipLevelCount: 5 }, // Mip level sizes: 31, 15, 7, 3, 1
+    { size: [31], dimension: '1d' as const, mipLevelCount: 6 }, // Mip level sizes: 31, 15, 7, 3, 1, ?
+    { size: [32, 32, 32], dimension: '3d' as const, mipLevelCount: 6 }, // Mip level sizes: 32x32x32, 16x16x16, 8x8x8, 4x4x4, 2x2x2, 1x1x1
+    { size: [32, 32, 32], dimension: '3d' as const, mipLevelCount: 7 }, // Mip level sizes: 32x32x32, 16x16x16, 8x8x8, 4x4x4, 2x2x2, 1x1x1, ?x?x?
+    { size: [32, 31, 31], dimension: '3d' as const, mipLevelCount: 6 }, // Mip level sizes: 32x31x31, 16x15x15, 8x7x7, 4x3x3, 2x1x1, 1x1x1
+    { size: [32, 31, 31], dimension: '3d' as const, mipLevelCount: 7 }, // Mip level sizes: 32x31x31, 16x15x15, 8x7x7, 4x3x3, 2x1x1, 1x1x1, ?x?x?
+    { size: [31, 32, 31], dimension: '3d' as const, mipLevelCount: 6 }, // Mip level sizes: 31x32x31, 15x16x15, 7x8x7, 3x4x3, 1x2x1, 1x1x1
+    { size: [31, 32, 31], dimension: '3d' as const, mipLevelCount: 7 }, // Mip level sizes: 31x32x31, 15x16x15, 7x8x7, 3x4x3, 1x2x1, 1x1x1, ?x?x?
+    { size: [31, 31, 32], dimension: '3d' as const, mipLevelCount: 6 }, // Mip level sizes: 31x31x32, 15x15x16, 7x7x8, 3x3x4, 1x1x2, 1x1x1
+    { size: [31, 31, 32], dimension: '3d' as const, mipLevelCount: 7 }, // Mip level sizes: 31x31x32, 15x15x16, 7x7x8, 3x3x4, 1x1x2, 1x1x1, ?x?x?
+    { size: [31, 31, 31], dimension: '3d' as const, mipLevelCount: 5 }, // Mip level sizes: 31x31x31, 15x15x15, 7x7x7, 3x3x3, 1x1x1
+    { size: [31, 31, 31], dimension: '3d' as const, mipLevelCount: 6 }, // Mip level sizes: 31x31x31, 15x15x15, 7x7x7, 3x3x3, 1x1x1, ?x?x?
+    { size: [32, 32], mipLevelCount: 100 }, // undefined shift check if miplevel is bigger than the integer bit width
+    { size: [32, 8], mipLevelCount: 6 }, // Mip levels: 32x8, 16x4, 8x2, 4x1, 2x1, 1x1
+    { size: [32, 8], mipLevelCount: 7 }, // Mip levels: 32x8, 16x4, 8x2, 4x1, 2x1, 1x1
+    { size: [32, 32, 64], mipLevelCount: 7 }, // Mip levels: 32x32x64, 16x16x64, 8x8x64, 4x4x64, 2x2x64, 1x1x64, ?x?x64
+    { size: [32, 32, 64], dimension: '3d' as const, mipLevelCount: 7 }, // Mip levels: 32x32x64, 16x16x32, 8x8x16, 4x4x8, 2x2x4, 1x1x2, 1x1x1
   ])
   .fn(async t => {
-    const { size, mipLevelCount, dimension, _success } = t.params;
+    const { size, mipLevelCount, dimension } = t.params;
 
-    const descriptor = {
+    const descriptor: GPUTextureDescriptor = {
       size,
       mipLevelCount,
       dimension,
@@ -203,9 +204,11 @@ g.test('mipLevelCount,bound_check')
       usage: GPUTextureUsage.SAMPLED,
     };
 
+    const success =
+      mipLevelCount <= maxMipLevelCount(descriptor.size as GPUExtent3DDict, descriptor.dimension);
     t.expectValidationError(() => {
       t.device.createTexture(descriptor);
-    }, !_success);
+    }, !success);
   });
 
 g.test('sampleCount')
--- a/src/webgpu/api/validation/createTexture.spec.ts
+++ b/src/webgpu/api/validation/createTexture.spec.ts
@@ -4,9 +4,6 @@ createTexture validation tests.
 TODO: review existing tests and merge with this plan:
 > All x= every texture format
 >
-> - mip level count validation
->     - x= every texture format
->     - x= any dimension
 > - sampleCount = {0, 1, 4, 8, 16, 256} with format/dimension that supports multisample
 >     - x= every texture format
 > - sampleCount = {1, 4}
@@ -33,7 +30,7 @@ TODO: review existing tests and merge with this plan:
 TODO: move destroy tests out of this file
 `;
 
-import { poptions } from '../../../common/framework/params_builder.js';
+import { poptions, params } from '../../../common/framework/params_builder.js';
 import { makeTestGroup } from '../../../common/framework/test_group.js';
 import { kAllTextureFormats, kAllTextureFormatInfo } from '../../capability_info.js';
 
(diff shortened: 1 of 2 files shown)


src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -70,45 +71,74 @@ class F extends ValidationTest {
...
+      mipLevelCount: 6,
+      format,
+      usage: GPUTextureUsage.SAMPLED,
+    };
 
-    t.expectValidationError(() => {
-      t.device.createTexture(descriptor);
-    }, !_success);
+    // const extension: GPUExtensionName | undefined = kAllTextureFormatInfo[format].extension;
+    const extension: GPUExtensionName = kAllTextureFormatInfo[format].extension;
```
Richard-Yunchao, line 87: I keep get error info like below:
Type '"texture-compression-bc" | undefined' is not assignable to type 'GPUExtensionName'

I tried a few methods (like the code on line 86), but they don't work. 

Any graceful solution for this if I don't split the tests into two: one for Uncompressed texture format (no extension needed), another for Compressed texture format (need extension)?  

In addition, it is valid to create mipmaps for compressed texture? I simply looked into Vulkan, looks like it is OK because it doesn't say no.

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -4,8 +4,9 @@ createTexture validation tests.
 TODO: review existing tests and merge with this plan:
 > All x= every texture format
 >
-> - {any dimension, mipLevelCount} = 0
```
Richard-Yunchao, line 7: Not sure whether I misunderstood this. But I suppose we don't need to test different dimensions (texture sizes) * all formats. I think two tests to cover these two factors are enough.

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -70,45 +71,74 @@ class F extends ValidationTest {
...
     { width: 32, height: 31, mipLevelCount: 6, _success: true }, // full mip chains are allowed (Mip level sizes: 32x31, 16x15, 8x7, 4x3, 2x1, 1x1)
     { width: 31, height: 32, mipLevelCount: 7, _success: false }, // too big mip chains on width are disallowed (Mip level sizes: 31x32, 15x16, 7x8, 3x4, 1x2, 1x1, ?x?)
     { width: 32, height: 31, mipLevelCount: 7, _success: false }, // too big mip chains on height are disallowed (Mip level sizes: 32x31, 16x15, 8x7, 4x3, 2x1, 1x1, ?x?)
     { width: 32, height: 32, mipLevelCount: 100, _success: false }, // undefined shift check if miplevel is bigger than the integer bit width
     { width: 32, height: 8, mipLevelCount: 6, _success: true }, // non square mip map halves the resolution until a 1x1 dimension. (Mip maps: 32 * 8, 16 * 4, 8 * 2, 4 * 1, 2 * 1, 1 * 1)
+    { width: 32, height: 32, arrayLayerCount: 64, mipLevelCount: 7, _success: false }, // array layer count for 2D texture should not be taken account to calculate mip levels. (Mip maps: 32 * 32 * 64, 16 * 16 * 64, 8 * 8 * 64, 4 * 4 * 64, 2 * 2 * 64, 1 * 1 * 64)
+    {
+      width: 32,
+      height: 32,
+      arrayLayerCount: 64,
```
Richard-Yunchao, line 109: I really want to change parameter name "arrayLayerCount" in this file into "depthOrArrayLayers" or "depthOrArrayLayerCount", WDYT? It's weird to say that arrayLayerCount is 64 for a 3D texture, just like this one.

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -70,45 +71,74 @@ class F extends ValidationTest {
...
-    { sampleCount: 2, _success: false }, // sampleCount of 2 is not allowed
-    { sampleCount: 3, _success: false }, // sampleCount of 3 is not allowed
-    { sampleCount: 4, _success: true }, // sampleCount of 4 is allowed
-    { sampleCount: 8, _success: false }, // sampleCount of 8 is not allowed
-    { sampleCount: 16, _success: false }, // sampleCount of 16 is not allowed
-    { sampleCount: 4, mipLevelCount: 2, _success: false }, // multisampled multi-level not allowed
-    { sampleCount: 4, arrayLayerCount: 2, _success: false }, // multisampled multi-layer is not allowed
-  ])
+g.test('mipLevelCount_validation_on_formats')
+  .params(poptions('format', kAllTextureFormats))
```
kainino0x, line 75: All of these are expected to pass, which means there's no control case or boundary condition being tested. How about something like this (using subcases from #444):

```ts
g.test('mipLevelCount_bound,formats')
  .desc(`Test mipLevelCount=6 at the boundary between size 31 and 32,
  with every dimension, and every possible format.`)
  .cases(params()
    .combine(poptions('dimension', ['1d', '2d', '3d']))
    .combine(poptions('format', kAllTextureFormats))
  ).subcases(({ dimension }) =>
    dimension === '1d' ? [
      { ok: true, size: [32, 1] },
      { ok: false, size: [31, 1] },
    ] : dimension === '2d' ? [
      { ok: true, size: [32, 31, 1] },
      { ok: true, size: [31, 32, 1] },
      { ok: false, size: [31, 31, 1] },
    ] : [
      { ok: true, size: [32, 31, 31] },
      { ok: true, size: [31, 32, 31] },
      { ok: true, size: [31, 31, 32] },
      { ok: false, size: [31, 31, 31] },
    ])
```

This tests different dimensions, but some dimension+format combinations might not be valid, so we'd have to filter those out (e.g. I don't think we allow 3d compressed textures yet). We can skip testing multiple dimensions, but if we do, but be sure to leave a TODO in the `.desc()`.

Perhaps kAllTextureFormats should be replaced by a hand-picked representative list (say, a few regular formats, all depth/stencil if valid\*, a few compressed).

This does end up overlapping a bit with the next test; I think a few cases from that can just be removed if they're covered here.

\* Not sure if depth/stencil mipmapping is allowed. We should check into this, perhaps first by checking against Dawn and then Vulkan. If you don't find a conclusive answer, we should err on the side of testing too many formats.

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -70,45 +71,74 @@ class F extends ValidationTest {
...
+
+    const descriptor = t.getDescriptor({ width, height, arrayLayerCount, mipLevelCount });
+
+    t.expectValidationError(() => {
+      t.device.createTexture(descriptor);
+    }, !_success);
+  });
 
-    const descriptor = t.getDescriptor({ width, height, mipLevelCount });
+g.test('validation_of_sampleCount')
```
kainino0x, line 125: uNit: why did this test move? it made the diff a little more verbose.

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: Thanks for your update!

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: Got it. So I added a separate test named zero_size to cover this situation. PTAL.

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: OK. Then we don't need to write down the new variable name in this test if we use `size` and assign values directly.

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -4,8 +4,9 @@ createTexture validation tests.
 TODO: review existing tests and merge with this plan:
 > All x= every texture format
 >
-> - {any dimension, mipLevelCount} = 0
+> - mip level count validation
 >     - x= every texture format
+>     - x= any dimension
```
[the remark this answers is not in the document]
Richard-Yunchao: Got it. Done.

src/webgpu/api/validation/createTexture.spec.ts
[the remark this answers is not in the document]

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -70,45 +71,74 @@ class F extends ValidationTest {
...
+
+    t.expectValidationError(() => {
+      t.device.createTexture(descriptor);
+    }, !_success);
+  });
 
-    const descriptor = t.getDescriptor({ width, height, mipLevelCount });
+g.test('validation_of_sampleCount')
+  .params([
+    // TODO: Consider making a list of "valid"+"invalid" texture descriptors in capability_info.
```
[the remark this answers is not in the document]
Richard-Yunchao: I will update this test in the next PR. The current one is mainly for mipLevelCount.

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -70,45 +71,74 @@ class F extends ValidationTest {
...
-    { sampleCount: 1, _success: true }, // sampleCount of 1 is allowed
-    { sampleCount: 2, _success: false }, // sampleCount of 2 is not allowed
-    { sampleCount: 3, _success: false }, // sampleCount of 3 is not allowed
-    { sampleCount: 4, _success: true }, // sampleCount of 4 is allowed
-    { sampleCount: 8, _success: false }, // sampleCount of 8 is not allowed
-    { sampleCount: 16, _success: false }, // sampleCount of 16 is not allowed
-    { sampleCount: 4, mipLevelCount: 2, _success: false }, // multisampled multi-level not allowed
-    { sampleCount: 4, arrayLayerCount: 2, _success: false }, // multisampled multi-layer is not allowed
-  ])
+g.test('mipLevelCount_validation_on_formats')
```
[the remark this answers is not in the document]

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -70,45 +71,74 @@ class F extends ValidationTest {
...
-    }, !_success);
+    // const extension: GPUExtensionName | undefined = kAllTextureFormatInfo[format].extension;
+    const extension: GPUExtensionName = kAllTextureFormatInfo[format].extension;
+    await t.selectDeviceOrSkipTestCase({ extensions: [extension] });
+
+    t.device.createTexture(descriptor);
   });
 
-g.test('validation_of_mipLevelCount')
+g.test('mipLevelCount_validation_on_dimensions')
```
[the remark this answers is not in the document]
Richard-Yunchao: done

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: Thanks for your suggestion, Kai. I think the test mipLevelCount,bound_check (the following one) intend to test the boundary of mipLevelCount (how to calculate the max value of mipLevelCount correctly) according to texture size/dimension. So I think we don't need to add a lot of boundary check for mipLevelCount. However, the `subcases` is interesting. I will try to use it if needed.

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: I moved it in order to be consistent to the sequence of TODOs.

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -71,58 +68,149 @@ class F extends ValidationTest {
...
+g.test('mipLevelCount,format')
+  .desc(
+    `Test texture creation with no mipmap chain, partial mipmap chain, full mipmap chain, out-of-bounds mipmap chain
+    for every format with different texture dimension types.`
+  )
+  .params(
+    params()
+      .combine(poptions('dimension', ['1d', '2d', '3d'] as const))
+      .combine(poptions('format', kAllTextureFormats))
+      .combine(poptions('mipLevelCount', [1, 3, 5, 6]))
```
kainino0x, line 145: Won't this test combinations that aren't valid? I don't think all formats can be 1d, 3d, and have mipmapping. We'll need capability info about this somewhere (in both cts and spec).

The validation tests for those things would be separate I think, but we still need to skip the invalid combinations here somehow.

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -71,58 +68,149 @@ class F extends ValidationTest {
...
+    { size: [31, 32], mipLevelCount: 7, _success: false }, // too big mip chains on width for 2D texture are disallowed (Mip level sizes: 31x32, 15x16, 7x8, 3x4, 1x2, 1x1, ?x?)
+    { size: [32, 31], mipLevelCount: 7, _success: false }, // too big mip chains on height for 2D texture are disallowed (Mip level sizes: 32x31, 16x15, 8x7, 4x3, 2x1, 1x1, ?x?)
+    { size: [31, 31], mipLevelCount: 5, _success: true }, // full mip chains for non-power-of-two 2D texture are allowed (Mip level sizes: 31x31, 15x15, 7x7, 3x3, 1x1)
+    { size: [31, 31], mipLevelCount: 6, _success: false }, // too big mip chains for non-power-of-two 2D texture are disallowed (Mip level sizes: 31x31, 15x15, 7x7, 3x3, 1x1, ?x?)
+    { size: [32, 1, 1], dimension: '1d' as const, mipLevelCount: 6, _success: true }, // full mip chains for 1D texture are allowed (Mip level sizes: 32, 16, 8, 4, 2, 1)
+    { size: [32, 1, 1], dimension: '1d' as const, mipLevelCount: 7, _success: false }, // too big mip chains for 1D texture are not allowed (Mip level sizes: 32, 16, 8, 4, 2, 1)
```
kainino0x, line 179: optional nit: You can put `as const` at the end (`.subcases([...] as const)` if you want to avoid repeating it

src/webgpu/api/validation/createTexture.spec.ts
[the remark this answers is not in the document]
kainino0x: Updating it later is fine, but IMO it's also fine to just delete the comment now since we noticed it here. It's okay to have tiny cleanups in unrelated commits, it's better than not getting around to them because of the overhead, or forgetting about them. Especially since we have quite a bit of debt in this codebase right now (missing descriptions, lack of subcases, lack of centralization of common complex stuff, incomplete tests, duplicated test plans).

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -71,58 +68,149 @@ class F extends ValidationTest {
...
+    t.expectValidationError(() => {
+      t.device.createTexture(descriptor);
+    }, !success);
+  });
+
+g.test('mipLevelCount,bound_check')
+  .desc(
+    `Test mip level count bound check upon different texture size and different texture dimension types.`
+  )
   .params([
```
kainino0x, line 170: Please switch this to either `cases` or `subcases` as `params` is deprecated. (I think it should be `subcases`.)

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -71,58 +68,149 @@ class F extends ValidationTest {
...
-    { width: 31, height: 32, mipLevelCount: 7, _success: false }, // too big mip chains on width are disallowed (Mip level sizes: 31x32, 15x16, 7x8, 3x4, 1x2, 1x1, ?x?)
-    { width: 32, height: 31, mipLevelCount: 7, _success: false }, // too big mip chains on height are disallowed (Mip level sizes: 32x31, 16x15, 8x7, 4x3, 2x1, 1x1, ?x?)
-    { width: 32, height: 32, mipLevelCount: 100, _success: false }, // undefined shift check if miplevel is bigger than the integer bit width
-    { width: 32, height: 8, mipLevelCount: 6, _success: true }, // non square mip map halves the resolution until a 1x1 dimension. (Mip maps: 32 * 8, 16 * 4, 8 * 2, 4 * 1, 2 * 1, 1 * 1)
-    { width: 32, height: 32, arrayLayerCount: 64, mipLevelCount: 7, _success: false }, // array layer count for 2D texture should not be taken account to calculate mip levels. (Mip maps: 32 * 32 * 64, 16 * 16 * 64, 8 * 8 * 64, 4 * 4 * 64, 2 * 2 * 64, 1 * 1 * 64)
+    { size: [32, 32], mipLevelCount: 6, _success: true }, // full mip chains for 2D texture are allowed (Mip level sizes: 32x32, 16x32, 8x8, 4x4, 2x2, 1x1)
```
kainino0x, line 171: nit: I think these old detailed comments are unnecessarily verbose. The "32x32, 16x16, ..." would be good to keep though.

The general premise can be in the description (above) instead of repeated here.

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -71,58 +68,149 @@ class F extends ValidationTest {
...
+    { size: [31, 31, 31], dimension: '3d' as const, mipLevelCount: 5, _success: true }, // full mip chains for non-power-of-two 3D texture are allowed (Mip level sizes: 31x31x31, 15x15x15, 7x7x7, 3x3x3, 1x1x1)
+    { size: [31, 31, 31], dimension: '3d' as const, mipLevelCount: 6, _success: false }, // too big mip chains for non-power-of-two 3D texture are not allowed (Mip level sizes: 31x31x31, 15x15x15, 7x7x7, 3x3x3, 1x1x1)
+    { size: [32, 32], mipLevelCount: 100, _success: false }, // undefined shift check if miplevel is bigger than the integer bit width
+    { size: [32, 8], mipLevelCount: 6, _success: true }, // non square mip map halves the resolution until a 1x1 dimension. (Mip maps: 32x8, 16x4, 8x2, 4x1, 2x1, 1x1)
+    { size: [32, 8], mipLevelCount: 7, _success: true }, // too big mip chains for non square mip map halves are disallowed (Mip maps: 32x8, 16x4, 8x2, 4x1, 2x1, 1x1)
+    { size: [32, 32, 64], mipLevelCount: 7, _success: false }, // array layer count for 2D texture should not be taken account to calculate mip levels. (Mip maps: 32x32x64, 16x16x64, 8x8x64, 4x4x64, 2x2x64, 1x1x64)
     {
```
kainino0x, line 190: uNit: see if this can be coerced to stay on one line

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -71,58 +68,149 @@ class F extends ValidationTest {
...
+    { size: [31, 32], mipLevelCount: 6, _success: true }, // full mip chains for 2D texture are allowed (Mip level sizes: 31x32, 15x16, 7x8, 3x4, 1x2, 1x1)
+    { size: [32, 31], mipLevelCount: 6, _success: true }, // full mip chains for 2D texture are allowed (Mip level sizes: 32x31, 16x15, 8x7, 4x3, 2x1, 1x1)
+    { size: [31, 32], mipLevelCount: 7, _success: false }, // too big mip chains on width for 2D texture are disallowed (Mip level sizes: 31x32, 15x16, 7x8, 3x4, 1x2, 1x1, ?x?)
+    { size: [32, 31], mipLevelCount: 7, _success: false }, // too big mip chains on height for 2D texture are disallowed (Mip level sizes: 32x31, 16x15, 8x7, 4x3, 2x1, 1x1, ?x?)
+    { size: [31, 31], mipLevelCount: 5, _success: true }, // full mip chains for non-power-of-two 2D texture are allowed (Mip level sizes: 31x31, 15x15, 7x7, 3x3, 1x1)
+    { size: [31, 31], mipLevelCount: 6, _success: false }, // too big mip chains for non-power-of-two 2D texture are disallowed (Mip level sizes: 31x31, 15x15, 7x7, 3x3, 1x1, ?x?)
+    { size: [32, 1, 1], dimension: '1d' as const, mipLevelCount: 6, _success: true }, // full mip chains for 1D texture are allowed (Mip level sizes: 32, 16, 8, 4, 2, 1)
```
kainino0x, line 178: uNit: These could be just `size: [32]`

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -71,58 +68,149 @@ class F extends ValidationTest {
...
+    { size: [31, 31, 31], dimension: '3d' as const, mipLevelCount: 5, _success: true }, // full mip chains for non-power-of-two 3D texture are allowed (Mip level sizes: 31x31x31, 15x15x15, 7x7x7, 3x3x3, 1x1x1)
+    { size: [31, 31, 31], dimension: '3d' as const, mipLevelCount: 6, _success: false }, // too big mip chains for non-power-of-two 3D texture are not allowed (Mip level sizes: 31x31x31, 15x15x15, 7x7x7, 3x3x3, 1x1x1)
+    { size: [32, 32], mipLevelCount: 100, _success: false }, // undefined shift check if miplevel is bigger than the integer bit width
+    { size: [32, 8], mipLevelCount: 6, _success: true }, // non square mip map halves the resolution until a 1x1 dimension. (Mip maps: 32x8, 16x4, 8x2, 4x1, 2x1, 1x1)
+    { size: [32, 8], mipLevelCount: 7, _success: true }, // too big mip chains for non square mip map halves are disallowed (Mip maps: 32x8, 16x4, 8x2, 4x1, 2x1, 1x1)
+    { size: [32, 32, 64], mipLevelCount: 7, _success: false }, // array layer count for 2D texture should not be taken account to calculate mip levels. (Mip maps: 32x32x64, 16x16x64, 8x8x64, 4x4x64, 2x2x64, 1x1x64)
```
kainino0x, line 189: nit: Should go to `?x?x64`

And please do the same fix for other comments with the same problem.

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -71,58 +68,149 @@ class F extends ValidationTest {
...
+
+    const success = mipLevelCount <= 5;
+    t.expectValidationError(() => {
+      t.device.createTexture(descriptor);
+    }, !success);
+  });
+
+g.test('mipLevelCount,bound_check')
+  .desc(
+    `Test mip level count bound check upon different texture size and different texture dimension types.`
```
kainino0x, line 168: Describe more specifically what this is testing, e.g.: "There must be no mip levels after a 1x1x1 level. Array layers are not mip-mapped. Test power-of-two and non-power-of-two base sizes."

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -71,58 +68,149 @@ class F extends ValidationTest {
...
+    { size: [32, 1, 1], dimension: '1d' as const, mipLevelCount: 6, _success: true }, // full mip chains for 1D texture are allowed (Mip level sizes: 32, 16, 8, 4, 2, 1)
+    { size: [32, 1, 1], dimension: '1d' as const, mipLevelCount: 7, _success: false }, // too big mip chains for 1D texture are not allowed (Mip level sizes: 32, 16, 8, 4, 2, 1)
+    { size: [31, 1, 1], dimension: '1d' as const, mipLevelCount: 5, _success: true }, // full mip chains for non-power-of-two 1D texture are allowed (Mip level sizes: 31, 15, 7, 3, 1)
+    { size: [31, 1, 1], dimension: '1d' as const, mipLevelCount: 6, _success: false }, // too big mip chains for non-power-of-two 1D texture are not allowed (Mip level sizes: 31, 15, 7, 3, 1)
+    { size: [32, 32, 32], dimension: '3d' as const, mipLevelCount: 6, _success: true }, // full mip chains for 3D texture are allowed (Mip level sizes: 32x32x32, 16x16x16, 8x8x8, 4x4x4, 2x2x2, 1x1x1)
+    { size: [32, 32, 32], dimension: '3d' as const, mipLevelCount: 7, _success: false }, // too big mip chains for 3D texture are not allowed (Mip level sizes: 32x32x32, 16x16x16, 8x8x8, 4x4x4, 2x2x2, 1x1x1)
```
kainino0x, line 183: nit: Should end with `?x?x?`

src/webgpu/api/validation/createTexture.spec.ts
kainino0x, line 171: nit: 16x32 -> 16x16

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -71,58 +68,149 @@ class F extends ValidationTest {
...
+      size: [32, 32, 64],
+      dimension: '3d' as const,
       mipLevelCount: 7,
       _success: true,
-    }, // depth of 3D texture should be taken account to calculate mip levels. (Mip maps: 32 * 32 * 64, 16 * 16 * 32, 8 * 8 * 16, 4 * 4 * 8, 2 * 2 * 4, 1 * 1 * 2, 1 * 1 * 1)
+    }, // depth of 3D texture should be taken account to calculate mip levels. (Mip maps: 32x32x64, 16x16x32, 8x8x16, 4x4x8, 2x2x4, 1x1x2, 1x1x1)
   ])
   .fn(async t => {
-    const { width, height, mipLevelCount, arrayLayerCount, _success } = t.params;
+    const { size, mipLevelCount, dimension, _success } = t.params;
```
kainino0x, line 198: thought to consider: It would be much easier to add cases if some part of this were computed instead of written manually. How about computing the max mipLevelCount and checking that's valid and that+1 is invalid? Computing it isn't quite trivial but it's fairly simple. A helper for it could live in `webgpu/util/texture`.

This would require separating the mipLevelCount:100 case, but that's testing a different thing anyway so it should be separate.

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -71,58 +68,149 @@ class F extends ValidationTest {
...
+
+    await t.selectDeviceOrSkipTestCase(kAllTextureFormatInfo[format].extension);
+
+    const success = mipLevelCount <= 5;
+    t.expectValidationError(() => {
+      t.device.createTexture(descriptor);
+    }, !success);
+  });
+
+g.test('mipLevelCount,bound_check')
```
kainino0x, line 166: I think we need tests like this for a compressed texture format as well. Can you add that as a TODO to file description?

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -71,58 +68,149 @@ class F extends ValidationTest {
...
+    { size: [31, 1, 1], dimension: '1d' as const, mipLevelCount: 6, _success: false }, // too big mip chains for non-power-of-two 1D texture are not allowed (Mip level sizes: 31, 15, 7, 3, 1)
+    { size: [32, 32, 32], dimension: '3d' as const, mipLevelCount: 6, _success: true }, // full mip chains for 3D texture are allowed (Mip level sizes: 32x32x32, 16x16x16, 8x8x8, 4x4x4, 2x2x2, 1x1x1)
+    { size: [32, 32, 32], dimension: '3d' as const, mipLevelCount: 7, _success: false }, // too big mip chains for 3D texture are not allowed (Mip level sizes: 32x32x32, 16x16x16, 8x8x8, 4x4x4, 2x2x2, 1x1x1)
+    { size: [31, 31, 31], dimension: '3d' as const, mipLevelCount: 5, _success: true }, // full mip chains for non-power-of-two 3D texture are allowed (Mip level sizes: 31x31x31, 15x15x15, 7x7x7, 3x3x3, 1x1x1)
+    { size: [31, 31, 31], dimension: '3d' as const, mipLevelCount: 6, _success: false }, // too big mip chains for non-power-of-two 3D texture are not allowed (Mip level sizes: 31x31x31, 15x15x15, 7x7x7, 3x3x3, 1x1x1)
```
kainino0x, line 185: Ideally we should have test cases for [32, 31, 31], [31, 32, 31], and [31, 31, 32] as well

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -71,58 +68,149 @@ class F extends ValidationTest {
...
+          'depth24plus-stencil8',
+          'bc1-rgba-unorm',
+        ] as const)
+      )
+  )
   .fn(async t => {
-    const format: GPUTextureFormat = t.params.format;
+    const { dimension, zeroArgument, format } = t.params;
+
+    const size = dimension === '1d' ? [31, 1, 1] : dimension === '2d' ? [31, 31, 1] : [31, 31, 31];
```
kainino0x, line 100: These base sizes will be invalid for bc formats.

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -71,58 +68,149 @@ class F extends ValidationTest {
...
+          'height',
+          'depthOrArrayLayers',
+          'mipLevelCount',
+        ] as const)
+      )
+      .combine(
+        poptions('format', [
+          'rgba8unorm',
+          'rgb10a2unorm',
+          'depth24plus-stencil8',
```
kainino0x, line 92: I'm not even sure whether depth formats can have mipmapping, we need to check this.

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -71,58 +68,149 @@ class F extends ValidationTest {
...
+          'bc1-rgba-unorm',
```
kainino0x, line 93: 1d/3d compressed textures are an unknown, I don't know what should happen here.

src/webgpu/api/validation/createTexture.spec.ts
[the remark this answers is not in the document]
Richard-Yunchao: OK. Done.

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: I suppose this is resolved.

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: Done. TODO added.

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: Done. Added a function named maxMipLevelCount in util/texture/base.ts (please propose a new file name if you think it is not good enough). MipLevelCount:100 should still fail even according to the calculation, should we remove this test, or put it in a standalone test category?

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: Removed it and added a TODO. Looks like depth is OK for mipmapping but stencil is not.

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: We disallow 1D/3D compressed texture in Dawn project. I suppose it is correct. So I added a filter for compressed format.

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: Yeah. Revised to 32.

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: Removed test for 1D/3D, and added a TODO.

src/webgpu/api/validation/createTexture.spec.ts

src/webgpu/api/validation/createTexture.spec.ts

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: Done. But when I run the test in latest Chromium, it complains that size's length should be 3. Is there anything wrong in the test (or in Chromium/Dawn)?

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: Not sure how to do this. Classify the whole test into 3 categories (1D, 2D, and 3D) and in each category, extract dimension parameter and add a combine function for dimension parameter?

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: Done

src/webgpu/api/validation/createTexture.spec.ts

src/webgpu/api/validation/createTexture.spec.ts

src/webgpu/api/validation/createTexture.spec.ts

Richard-Yunchao: BTW, subcases will combine all sub-tests in subcases into one. We can only see much less error information. While the previous (params) will isolate each sub-test with a separate checkbox and show the error info for each sub-test.

src/webgpu/api/validation/createTexture.spec.ts
kainino0x: never mind, looks like this suggestion makes typescript unhappy for reasons I don't understand (could be a bug in our type metaprogramming code)

src/webgpu/api/validation/createTexture.spec.ts
kainino0x: We intend to fix this for subcases, see "drill down" in #305, but no one is currently working on it as I don't think it's that critical.

src/webgpu/api/validation/createTexture.spec.ts
kainino0x: This restriction should have been lifted in https://chromium-review.googlesource.com/c/chromium/src/+/2478976, but maybe we missed something, can you check and file an issue if we did?

src/webgpu/api/validation/createTexture.spec.ts
kainino0x: It should be a separate test.

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -70,45 +69,164 @@ class F extends ValidationTest {
...
+    const descriptor: GPUTextureDescriptor = {
+      size,
+      mipLevelCount,
+      dimension,
+      format: 'rgba8unorm' as const,
+      usage: GPUTextureUsage.SAMPLED,
+    };
 
+    const success =
+      mipLevelCount <= maxMipLevelCount(descriptor.size as GPUExtent3DDict, descriptor.dimension);
```
kainino0x, line 208: Instead of passing mipLevelCount as a param, you can cut the subcase list by ~half by doing this:
```ts
      const mipLevelCount = maxMipLevelCount(descriptor.size as GPUExtent3DDict, descriptor.dimension);
      device.createTexture( with mipLevelCount );
        device.createTexture( with mipLevelCount + 1 );
      });
```

kainino0x, line 208: This type cast (`as GPUExtent3DDict`) doesn't make sense. It's casting an array to an object and weird stuff will happen as a result (Math.max with `undefined`). (Please make sure the tests seem to be functioning before we land this PR.)

Incidentally, I just added `standardizeExtent3D` to handle this. You can change maxMipLevelCount to take `GPUExtent3D` instead of `GPUExtent3DDict` and use that helper to "parse" it.
https://github.com/gpuweb/cts/blob/main/src/webgpu/util/unions.ts

src/webgpu/util/texture/base.ts
```diff
@@ -0,0 +1,9 @@
+export function maxMipLevelCount(
+  size: GPUExtent3DDict,
+  dimension: GPUTextureDimension | undefined
+): number {
+  let maxMippedDimension = size.width;
+  if (dimension !== '1d') maxMippedDimension = Math.max(maxMippedDimension, size.height);
+  if (dimension === '3d') maxMippedDimension = Math.max(maxMippedDimension, size.depth);
+  return Math.log2(maxMippedDimension) + 1;
```
kainino0x, line 8:
```suggestion
  return Math.floor(Math.log2(maxMippedDimension)) + 1;
```

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: Got it. But it is not so convenient as before when I run the new tests I just wrote to verify whether the new tests are good. Previously, most sub-tests can pass (they are green!) if the new written CTS tests are ready for review. Right now, the whole test will fail if one single sub-test fail (that 's very common because there implementation in Chromium is not 100% ready) and I don't know whether other sub-tests can pass by first sight.

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: Done!

You are right. I thought undefined values were changed into "1" (or "2D", or other default values) implicitly somewhere. It is not true. We need to do this explicitly via functions like standardizeExtent3D. BTW, I do check my CTS in latest Chromium on my local machine before every request for review. But as I said, it is not so convenient to verify the new tests because of `subcases` right now.

src/webgpu/util/texture/base.ts
Richard-Yunchao: Done

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: Done. see the crbug at https://bugs.chromium.org/p/chromium/issues/detail?id=1180301

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -70,45 +70,173 @@ class F extends ValidationTest {
...
+
+g.test('mipLevelCount,format')
+  .desc(
+    `Test texture creation with no mipmap chain, partial mipmap chain, full mipmap chain, out-of-bounds mipmap chain
+    for every format with different texture dimension types.
+    TODO: test 1D and 3D dimensions. Note that it is invalid for some formats with 1D/3D and/or mipmapping.`
+  )
+  .subcases(() =>
+    params()
+      .combine(poptions('format', kAllTextureFormats))
```
Richard-Yunchao, line 142: I noticed that a few texture formats are missing in CTS and Dawn, compared to spec.
The texture formats we are missing in CTS: depth16unorm, depth24unorm-stencil8, depth32float-stencil8
The texture formats we are missing in Dawn: stencil8 + all missing formats in CTS. 

Do we need to file bugs for this? I am not sure whether you are aware of this, though.

Richard-Yunchao:
Added Corentin and Austin for code review. The first PR is not ready for review but I met a problem, so I didn't add you. But I forgot it then.

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: Done

kainino0x (approved):
LGTM! Pushed some tiny last changes.

Richard-Yunchao:
> LGTM! Pushed some tiny last changes.

Thanks for your changes, Kai. 

I look forward to your (and Corentin and Austin's) feedback about the missing texture formats in CTS and Dawn. Please see my comment above for details. :)

austinEng:
We're aware they're not implemented yet, but you're right there are no bugs tracking it. You could file them if you'd like. I don't think they're very, very important to implement in the near term.

src/webgpu/api/validation/createTexture.spec.ts
kainino0x: Sorry, forgot about this. Yeah, we should add them in the CTS. Filed https://github.com/gpuweb/cts/issues/467

Richard-Yunchao:
> We're aware they're not implemented yet, but you're right there are no bugs tracking it. You could file them if you'd like. I don't think they're very, very important to implement in the near term.

OK. I simply searched Dawn issues, there are actually two bugs related to this missing texture formats issue: 
Implement the depth16 format: https://bugs.chromium.org/p/dawn/issues/detail?id=570
Implement the stencil8 format: https://bugs.chromium.org/p/dawn/issues/detail?id=666 

But we are still missing some formats. So I filed a bug at https://bugs.chromium.org/p/dawn/issues/detail?id=690

PR #458, opened 2021-02-05
Update the test plan for validation/vertex_state

Kangz:
-----

- [X] Incomplete tests, if any, are marked with TODO or `.unimplemented()`.

- [x] WebGPU readability
- [x] TypeScript readability

### DIFF
--- a/src/webgpu/api/validation/vertex_state.spec.ts
+++ b/src/webgpu/api/validation/vertex_state.spec.ts
@@ -1,18 +1,58 @@
 export const description = `
 vertexState validation tests.
 
-TODO: review existing tests, write descriptions, and make sure tests are complete.
-      Make sure the following is covered.
-> - In createRenderPipeline():
-> - An attribute is unused by the shader
-> - If invalid, test these (if valid, they should be operation tests instead):
->     - Vertex buffer with zero attributes
->     - Overlapping attributes
->         - Verify correct sizing of every vertex format
->     - Overlapping vertex buffer elements (an attribute offset + its size > arrayStride)
->     - Shader tries to use an attribute location that's not bound
->     - Alignment constraints on attributes, if any
->     - Alignment constraints on arrayStride, if any
+TODO: implement the combinations tests below.
+
+Test the number of vertex buffers allowed:
+ - For buffers = 0, 1, limit, limit + 1
+  - Test using empty vertex buffers except the last one with 0 or 1 element
+   - Test should fail IFF buffers > limit
+(note this also tests that empty vertex buffers are allowed)
+
+Test the limit of attributes:
+ - For attributePerBuffer = 0, 1, 4:
+  - For totalAttribs = 4, limit, limit + 1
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/vertex_state.spec.ts
```diff
@@ -1,18 +1,58 @@
...
+  - Check error IFF stride > limit
+
+Test the stride alignment check:
+ - For stride = 0, 2, 4, limit -4, limit -2, limit:
+  - Check error IFF stride not aligned to 4
+
+Test check that the end attribute must be contained in the stride:
+ - For stride = 0 (special case), 4, 128, limit
+   - For each vertex format
+    - For offset stride, stride - componentsize(format), stride - sizeof(format), stride - sizeof(format) + componentsize(format), 0, 2^32 - componentsize(format)
```
kainino0x, line 48: nit: perhaps `2**32` as well?

Tempted to try to come up with a case that tests whether you do all your offset computations in fp64 (and therefore would have rounding errors) but seems hard. Maybe just `2**60` or something.

src/webgpu/api/validation/vertex_state.spec.ts
```diff
@@ -1,18 +1,58 @@
...
+Test the max stride check:
+ - For stride = 0, 4, 256, limit - 4, limit, limit +1
+  - Check error IFF stride > limit
+
+Test the stride alignment check:
+ - For stride = 0, 2, 4, limit -4, limit -2, limit:
+  - Check error IFF stride not aligned to 4
+
+Test check that the end attribute must be contained in the stride:
+ - For stride = 0 (special case), 4, 128, limit
```
kainino0x, line 46: Does the stride=0 special case affect the limit on offset+size since there's no stride to check it against? Or is offset+size still limited to maxStride? (I assume the latter)

src/webgpu/api/validation/vertex_state.spec.ts
```diff
@@ -1,18 +1,58 @@
...
+   - For each vertex format
+    - For offset stride, stride - componentsize(format), stride - sizeof(format), stride - sizeof(format) + componentsize(format), 0, 2^32 - componentsize(format)
+      - Check error IFF offset + sizeof(format) > stride (or 2048 for 0)
+
+Test check that the end attribute must be contained in the stride:
```
kainino0x, line 51: "Test that an attribute must be aligned to component size"?

kainino0x (approved):

src/webgpu/api/validation/vertex_state.spec.ts
Kangz: Added 2**60 but TBH I think it's unlikely an implementation will do stuff with f64

src/webgpu/api/validation/vertex_state.spec.ts
Kangz: Yep it's the latter.

Kangz:
Thanks!

PR #459, opened 2021-02-08
Add operation tests for GPUQueue.writeBuffer()

egalli:
I figured that "source.origin is unaligned" meant a non-multiple of 4 `dataOffset` on `writeBuffer`.

-----

- [x] WebGPU readability
- [x] TypeScript readability

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,11 +9,11 @@
     "wpt": "grunt wpt",
     "fix": "grunt fix",
     "unittest": "grunt unittest",
-    "gen_wpt_cts_html": "tools/gen_wpt_cts_html",
-    "dev": "tools/dev_server"
+    "gen_wpt_cts_html": "node tools/gen_wpt_cts_html",
+    "dev": "node tools/dev_server"
   },
   "engines": {
-    "node": ">=11.14.0"
+    "node": ">=12.0.0"
   },
   "repository": {
     "type": "git",
--- a/src/webgpu/api/operation/queue/writeBuffer.spec.ts
+++ b/src/webgpu/api/operation/queue/writeBuffer.spec.ts
@@ -61,7 +61,7 @@ class F extends GPUTest {
     });
 
     // Initialize buffer to non-zero data (0xff)
-    const expectedData = new Uint8Array(range<number>(bufferSize, i => 0xff));
+    const expectedData = new Uint8Array(range<number>(bufferSize, () => 0xff));
     this.queue.writeBuffer(buffer, 0, expectedData);
 
     for (const { bufferOffset, data, arrayType, useArrayBuffer, dataOffset, dataSize } of writes) {
--- a/src/webgpu/api/operation/queue/writeBuffer.spec.ts
+++ b/src/webgpu/api/operation/queue/writeBuffer.spec.ts
@@ -1,11 +1,53 @@
-export const description = `
-TODO:
-- source.origin is unaligned
-- data given as ArrayBuffer
-- data given as each TypedArray variant
-`;
+export const description = 'Operation tests for GPUQueue.writeBuffer()';
 
+import { params, pbool, poptions } from '../../../../common/framework/params_builder.js';
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { range } from '../../../../common/framework/util/util.js';
 import { GPUTest } from '../../../gpu_test.js';
 
+const kTypedArrays = [
+  'Uint8Array',
+  'Uint16Array',
+  'Uint32Array',
+  'Int8Array',
+  'Int16Array',
+  'Int32Array',
+  'Float32Array',
+  'Float64Array',
+] as const;
+
 export const g = makeTestGroup(GPUTest);
+
+const kTestData = range<number>(16, i => i);
+
+g.test('array_types')
+  .desc('Tests that writeBuffer correctly handles different TypedArrays and ArrayBuffer.')
+  .cases(params().combine(poptions('arrayType', kTypedArrays)).combine(pbool('useArrayBuffer')))
+  .fn(t => {
+    const { arrayType, useArrayBuffer } = t.params;
+    const typeBuilder = self[arrayType];
+    const typedData = new typeBuilder(kTestData);
+    const buffer = t.device.createBuffer({
+      size: typedData.byteLength,
+      usage: GPUBufferUsage.COPY_SRC | GPUBufferUsage.COPY_DST,
+    });
+    t.queue.writeBuffer(buffer, 0, useArrayBuffer ? typedData.buffer : typedData);
+    t.expectContents(buffer, typedData);
+  });
+
+g.test('unaligned_source')
+  .desc('Tests that writeBuffer correctly handles unaligned sources.')
+  .fn(t => {
+    const kStartOffset = 1;
+    const arrayData = new ArrayBuffer(kStartOffset + kTestData.length);
+    const typedData = new Uint8Array(arrayData, kStartOffset);
+    kTestData.every((val, idx) => {
+      typedData[idx] = val;
+    });
+    const buffer = t.device.createBuffer({
+      size: typedData.byteLength,
+      usage: GPUBufferUsage.COPY_SRC | GPUBufferUsage.COPY_DST,
+    });
+    t.queue.writeBuffer(buffer, 0, arrayData, kStartOffset);
+    t.expectContents(buffer, typedData);
+  });
--- a/src/webgpu/api/operation/queue/writeBuffer.spec.ts
+++ b/src/webgpu/api/operation/queue/writeBuffer.spec.ts
@@ -4,6 +4,7 @@ import { params, pbool, poptions } from '../../../../common/framework/params_bui
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
 import { range } from '../../../../common/framework/util/util.js';
 import { GPUTest } from '../../../gpu_test.js';
+import { align } from '../../../util/math.js';
 
 const kTypedArrays = [
   'Uint8Array',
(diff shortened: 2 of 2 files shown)


src/webgpu/api/operation/queue/writeBuffer.spec.ts
```diff
@@ -1,11 +1,53 @@
...
+  'Int32Array',
+  'Float32Array',
+  'Float64Array',
+] as const;
+
 export const g = makeTestGroup(GPUTest);
+
+const kTestData = range<number>(16, i => i);
+
+g.test('array_types')
```
Kangz, line 23: Can we also have tests that the offset and size arguments are handled correctly? (for typed arrays they are in elements while they are in bytes for ArrayBuffer).

What about a 0-sized writeBuffer?

What about many writeBuffers on the same location? (to check barriers are done correctly), overlapping location but only on one word at the start or the end?

Any other tests you can think of?

Kangz (commented):
Looks good but I added suggestions for more tests.

src/webgpu/api/operation/queue/writeBuffer.spec.ts
egalli: I added the tests you suggested: 
- using offset and size on TypedArrays and ArrayBuffers
- 0-sized writes to non-zero and zero sized buffers
- multiple overlapping writes with decreasing sizes

I added another tests with overlap in the middle. Although, I'm not sure that it adds much.

I also refactored the tests so that they calculate the expected result given a list of writes.

src/webgpu/api/operation/queue/writeBuffer.spec.ts
```diff
@@ -1,11 +1,228 @@
...
+  }
+
+  testWriteBuffer(...writes: WriteBufferSignature[]) {
+    const bufferSize = this.calculateRequiredBufferSize(writes);
+    const buffer = this.device.createBuffer({
+      size: bufferSize,
+      usage: GPUBufferUsage.COPY_SRC | GPUBufferUsage.COPY_DST,
+    });
+
+    // Initialize buffer to non-zero data (0xff)
```
Kangz, line 63: Uh, this is using writeBuffer to test writeBuffer. Maybe use mappedAtCreation instead?

src/webgpu/api/operation/queue/writeBuffer.spec.ts
```diff
@@ -1,11 +1,228 @@
...
+    const expectedData = new Uint8Array(range<number>(bufferSize, () => 0xff));
+    this.queue.writeBuffer(buffer, 0, expectedData);
+
+    for (const { bufferOffset, data, arrayType, useArrayBuffer, dataOffset, dataSize } of writes) {
+      const typeBuilder = self[arrayType];
```
Kangz, line 68: What does the `self[stuff]` construct do in Typescript?

src/webgpu/api/operation/queue/writeBuffer.spec.ts
```diff
@@ -1,11 +1,228 @@
...
+      if (dataOffset) {
+        requiredSize -= dataOffset * bytesPerElement;
+      }
+      if (dataSize) {
+        requiredSize = Math.min(requiredSize, dataSize * bytesPerElement);
+      }
+
+      requiredSize += bufferOffset;
+
+      if (requiredSize > bufferSize) {
```
Kangz, line 48: nit: Math.max would make this more clear.

src/webgpu/api/operation/queue/writeBuffer.spec.ts
```diff
@@ -1,11 +1,228 @@
...
+      let requiredSize = data.length * typeBuilder.BYTES_PER_ELEMENT;
+
+      // When passing data as an ArrayBuffer, dataOffset, dataSize use byte instead of number of
+      // elements
+      const bytesPerElement = useArrayBuffer ? 1 : typeBuilder.BYTES_PER_ELEMENT;
+      if (dataOffset) {
+        requiredSize -= dataOffset * bytesPerElement;
+      }
+      if (dataSize) {
+        requiredSize = Math.min(requiredSize, dataSize * bytesPerElement);
```
Kangz, line 43: It's a bit weird that we modify requiredSize to just set it here, maybe we could do:

```
bytesPerElement = ...
requiredSize = bufferOffset;
if (dataSize) {
  requiredSize += ...;
} else {
    requiredSize += data.length * BYTES_PER_ELEMENT  - (dataOffset || 0) * bytesPerElement
}
```

src/webgpu/api/operation/queue/writeBuffer.spec.ts
```diff
@@ -1,11 +1,228 @@
...
+    {
+      // Zero buffer
+      writes: [],
+    }, // Expected []
+    {
+      // Unaligned source
+      writes: [
+        {
+          bufferOffset: 0,
+          data: [0, ...kTestData],
```
Kangz, line 183: nit: maybe put a garbage number instead of 0 to make it clearer that it's ignored.

src/webgpu/api/operation/queue/writeBuffer.spec.ts
```diff
@@ -1,11 +1,228 @@
...
+    },
+    {
+      // Write over with empty buffer
+      writes: [
+        { bufferOffset: 0, data: kTestData, arrayType: 'Uint8Array', useArrayBuffer: false },
+        { bufferOffset: 0, data: [], arrayType: 'Uint8Array', useArrayBuffer: false },
+      ], // Expected [0, 1, 2, 3, 4, 5 ,6 ,7, 8, 9, 10, 11, 12, 13, 14, 15]
+    },
+    {
+      // Zero buffer
```
Kangz, line 175: I'm not sure what this does, don't we want to do a single writebuffer of size 0 instead?

src/webgpu/api/operation/queue/writeBuffer.spec.ts
```diff
@@ -1,11 +1,228 @@
...
+
+g.test('array_types')
+  .desc('Tests that writeBuffer correctly handles different TypedArrays and ArrayBuffer.')
+  .cases(params().combine(poptions('arrayType', kTypedArrays)).combine(pbool('useArrayBuffer')))
+  .fn(t => {
+    const { arrayType, useArrayBuffer } = t.params;
+
+    t.testWriteBuffer({ bufferOffset: 0, data: kTestData, arrayType, useArrayBuffer });
+  });
+
```
Kangz, line 104: Can we also have tests generic over the array typed / arraybuffer that set the dataOffset and the dataLength to check that the browser computes it correctly? Right now there's only some special cases tested below.

Kangz (approved):

src/webgpu/api/operation/queue/writeBuffer.spec.ts
kainino0x: I think this is okay; if this writeBuffer fails, then the expectContents will still fail.

src/webgpu/api/operation/queue/writeBuffer.spec.ts
```diff
@@ -1,11 +1,228 @@
...
+  }
+}
+
+export const g = makeTestGroup(F);
+
+const kTestData = range<number>(16, i => i);
+
+g.test('array_types')
+  .desc('Tests that writeBuffer correctly handles different TypedArrays and ArrayBuffer.')
+  .cases(params().combine(poptions('arrayType', kTypedArrays)).combine(pbool('useArrayBuffer')))
```
kainino0x, line 98: optional nit: Arguably redundant because I think with useArrayBuffer:true the test is the same regardless of the arrayType? Not a big deal though.

kainino0x (commented):
Overall looks good, I will leave it to Corentin to continue the detailed review

egalli (commented):
Updates code with change, sorry for the delay.

kainino0x:
No worries, sorry I can't get to this tonight, will get to it tomorrow.

kainino0x (approved):
LGTM, and the changes look small enough that Corentin's approval should still stand. Few nits.
Corentin, if you actually still want to rereview, please feel free, or re-request review from yourself, otherwise I'll land this after nits.

src/webgpu/api/operation/queue/writeBuffer.spec.ts
```diff
@@ -21,33 +21,37 @@ type WriteBufferSignature = {
...
-      const typeBuilder = self[arrayType];
-      let requiredSize = data.length * typeBuilder.BYTES_PER_ELEMENT;
-
-      // When passing data as an ArrayBuffer, dataOffset, dataSize use byte instead of number of
-      // elements
-      const bytesPerElement = useArrayBuffer ? 1 : typeBuilder.BYTES_PER_ELEMENT;
-      if (dataOffset) {
-        requiredSize -= dataOffset * bytesPerElement;
-      }
+      const TypedArrayConstructor = self[arrayType];
```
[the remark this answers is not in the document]
egalli: `globalThis` works. Although, I had to bump the node version on the `package.json`.

package.json
```diff
@@ -9,11 +9,11 @@
     "wpt": "grunt wpt",
     "fix": "grunt fix",
     "unittest": "grunt unittest",
-    "gen_wpt_cts_html": "tools/gen_wpt_cts_html",
-    "dev": "tools/dev_server"
+    "gen_wpt_cts_html": "node tools/gen_wpt_cts_html",
```
egalli, line 12: `node` is required to use these in windows if you don't have ".js" associated with node.

kainino0x: oops, thanks for catching this! Are things otherwise working well on Windows? I haven't tried in, like, years.

kainino0x (approved):

PR #464, opened 2021-02-10
Remove explicit storage class from samplers/textures

jrprice:
WGSL specifies that texture and sampler variables must not have a
storage class decoration, as they implicitly have the `handle` storage
class.

-----

- [x] WebGPU readability
- [x] TypeScript readability


kainino0x (approved):
thanks!

PR #465, opened 2021-02-12
Vertex state validation

Kangz:
-----

- [ ] WebGPU readability
- [ ] TypeScript readability

### DIFF
--- a/src/webgpu/api/validation/vertex_state.spec.ts
+++ b/src/webgpu/api/validation/vertex_state.spec.ts
@@ -40,13 +40,13 @@ Test that an attribute must be aligned to the component size:
 
 import { params, pbool, poptions } from '../../../common/framework/params_builder.js';
 import { makeTestGroup } from '../../../common/framework/test_group.js';
+import {
+  kMaxVertexAttributes,
+  kMaxVertexBufferArrayStride,
+  kMaxVertexBuffers,
+} from '../../capability_info.js';
 
 import { ValidationTest } from './validation_test.js';
-
-const MAX_VERTEX_ATTRIBUTES: number = 16;
-const MAX_VERTEX_BUFFER_END: number = 2048;
-const MAX_VERTEX_BUFFER_ARRAY_STRIDE: number = 2048;
-const MAX_VERTEX_BUFFERS: number = 8;
 
 const SIZEOF_FLOAT = Float32Array.BYTES_PER_ELEMENT;
 
@@ -129,7 +129,7 @@ g.test('max_vertex_buffer_limit')
   )
   .subcases(() =>
     params()
-      .combine(poptions('count', [0, 1, MAX_VERTEX_BUFFERS, MAX_VERTEX_BUFFERS + 1]))
+      .combine(poptions('count', [0, 1, kMaxVertexBuffers, kMaxVertexBuffers + 1]))
       .combine(pbool('lastEmpty'))
   )
   .fn(t => {
@@ -147,19 +147,19 @@ g.test('max_vertex_buffer_limit')
       }
     }
 
-    const success = count <= MAX_VERTEX_BUFFERS;
+    const success = count <= kMaxVertexBuffers;
     t.testVertexState(success, { vertexBuffers });
   });
 
 g.test('max_vertex_attribute_limit')
   .desc(
     `Test that only up to <maxVertexAttributes> vertex attributes are allowed.
-   - Tests with 0, 1, limit, limits + 1 vertex attributes.
+   - Tests with 0, 1, limit, limits + 1 vertex attribute.
    - Tests with 0, 1, 4 attributes per buffer (with remaining attributes in the last buffer).`
   )
   .subcases(() =>
     params()
-      .combine(poptions('attribCount', [0, 1, MAX_VERTEX_ATTRIBUTES, MAX_VERTEX_ATTRIBUTES + 1]))
+      .combine(poptions('attribCount', [0, 1, kMaxVertexAttributes, kMaxVertexAttributes + 1]))
       .combine(poptions('attribsPerBuffer', [0, 1, 4]))
   )
   .fn(t => {
@@ -171,7 +171,7 @@ g.test('max_vertex_attribute_limit')
     while (attribsAdded !== attribCount) {
       // Choose how many attributes to add for this buffer. The last buffer gets all remaining attributes.
       let targetCount = Math.min(attribCount, attribsAdded + attribsPerBuffer);
-      if (vertexBuffers.length === MAX_VERTEX_BUFFERS - 1) {
+      if (vertexBuffers.length === kMaxVertexBuffers - 1) {
         targetCount = attribCount;
       }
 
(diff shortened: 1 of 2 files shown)


src/webgpu/api/validation/vertex_state.spec.ts
```diff
@@ -303,73 +310,169 @@ g.test('vertex_attribute_shaderLocation_limit')
...
+    // Add attribute A
+    if (testAttributeAtStartA) {
+      attributesA.unshift({ format: 'float', offset: 0, shaderLocation: shaderLocationA });
+    } else {
+      attributesA.push({ format: 'float', offset: 0, shaderLocation: shaderLocationA });
+    }
+
+    // Add attribute B potentially in the same list of attributes as A
+    const attributesB = vertexBufferAttributes[vertexBufferIndexB];
+    if (testAttributeAtStartB) {
```
kainino0x, line 367: isn't this array empty at this point so this has no effect?

src/webgpu/api/validation/vertex_state.spec.ts
```diff
@@ -303,73 +310,169 @@ g.test('vertex_attribute_shaderLocation_limit')
...
+      attributesB.unshift({ format: 'float', offset: 0, shaderLocation: shaderLocationB });
+    } else {
+      attributesB.push({ format: 'float', offset: 0, shaderLocation: shaderLocationB });
+    }
+
+    // Use the attributes to make the list of vertex buffers not that we might be setting the same vertex
+    // buffer twice but that only happens when it is the only vertex buffer.
```
kainino0x, line 374: I can't parse this, needs some punctuation

src/webgpu/api/validation/vertex_state.spec.ts
```diff
@@ -303,73 +310,169 @@ g.test('vertex_attribute_shaderLocation_limit')
...
+      vertexBufferIndexA,
+      vertexBufferIndexB,
+      testAttributeAtStartA,
+      testAttributeAtStartB,
+      shaderLocationA,
+      shaderLocationB,
+      extraAttributes,
+    } = t.params;
+
+    const vertexBufferAttributes: GPUVertexAttributeDescriptor[][] = [];
```
kainino0x, line 341: Not actually used, could just be `attributesA = [];` `attributesB = [];`

src/webgpu/api/validation/vertex_state.spec.ts
```diff
@@ -116,6 +103,26 @@ class F extends ValidationTest {
       });
     }, !success);
   }
+
+  generateTestVertexShader(inputs: { type: string; location: number }[]): string {
+    let interfaces = '';
+    let body = '';
+
+    const count = 0;
```
kainino0x, line 111: is this meant to be incremented in the loop?

src/webgpu/api/validation/vertex_state.spec.ts
```diff
@@ -303,73 +310,169 @@ g.test('vertex_attribute_shaderLocation_limit')
...
+    vertexBuffers[vertexBufferIndexB] = { arrayStride: 256, attributes: attributesB };
+
+    const success = shaderLocationA !== shaderLocationB;
+    t.testVertexState(success, { vertexBuffers });
+  });
+
+g.test('vertex_shader_input_location_limit')
+  .desc(
+    `Test that vertex shader's input's location decoration must be less than maxVertexAttributes.
+   - Test for shaderLocation 0, 1, limit - 1, limit`
```
kainino0x, line 386: Maybe add some weird cases like `-1`, `2**32`? but more likely those will need to be wgsl tests

src/webgpu/api/validation/vertex_state.spec.ts
```diff
@@ -303,73 +310,169 @@ g.test('vertex_attribute_shaderLocation_limit')
...
-    `;
-    const descriptor = t.getDescriptor(vertexState, code);
-    t.device.createRenderPipeline(descriptor);
-  }
-  {
-    // Check for an error when the pipeline uses an attribute not in the vertex input
-    const code = `
-      [[location(2)]] var<in> a : vec4<f32>;
+g.test('vertex_shader_input_location_in_vertex_state')
+  .desc(``)
```
kainino0x, line 421: should have at least a short description

src/webgpu/api/validation/vertex_state.spec.ts
```diff
@@ -303,73 +310,169 @@ g.test('vertex_attribute_shaderLocation_limit')
...
+    }
+
+    // Use the attributes to make the list of vertex buffers not that we might be setting the same vertex
+    // buffer twice but that only happens when it is the only vertex buffer.
+    const vertexBuffers: GPUVertexBufferLayoutDescriptor[] = [];
+    vertexBuffers[vertexBufferIndexA] = { arrayStride: 256, attributes: attributesA };
+    vertexBuffers[vertexBufferIndexB] = { arrayStride: 256, attributes: attributesB };
+
+    const success = shaderLocationA !== shaderLocationB;
+    t.testVertexState(success, { vertexBuffers });
```
kainino0x, line 380: just realized testVertexState captures three calls (createShaderModule, createShaderModule, createRenderPipeline) in its error scope, ideally it should only be one (probably createRenderPipeline for all of these tests)

src/webgpu/api/validation/vertex_state.spec.ts
```diff
@@ -303,73 +310,169 @@ g.test('vertex_attribute_shaderLocation_limit')
...
+   - Test for various shaderLocations that conflict or not`
+  )
+  .subcases(() =>
+    params()
+      .combine(poptions('vertexBufferIndexA', [0, 1, kMaxVertexBuffers - 1]))
+      .combine(poptions('vertexBufferIndexB', [0, 1, kMaxVertexBuffers - 1]))
+      .combine(pbool('testAttributeAtStartA'))
+      .combine(pbool('testAttributeAtStartB'))
+      .combine(poptions('shaderLocationA', [0, 1, 7, kMaxVertexAttributes - 1]))
+      .combine(poptions('shaderLocationB', [0, 1, 7, kMaxVertexAttributes - 1]))
```
kainino0x, line 327: didn't have a chance to double check, but I think this is creating duplicate shaderLocations in the vertex state AND duplicate locations in the WGSL? which means there are two potentially invalid things being tested together. Probably should have at least one test/case for each separately if not covered already

kainino0x (approved):

src/webgpu/api/validation/vertex_state.spec.ts
Kangz: It can be the same object as `attrributesA`. Made the comment more clear.

src/webgpu/api/validation/vertex_state.spec.ts
Kangz: Actually we use it so we can have both attributes be aliased or not. Added a comment to clarify this.

src/webgpu/api/validation/vertex_state.spec.ts
Kangz: Whoops :)

src/webgpu/api/validation/vertex_state.spec.ts
Kangz: This test doesn't trigger an issue in the shaderModule because it uses an empty VS. I'll add a comment.

Kangz:
Addressed all comments, PTAL again.

Kangz:
Up!

kainino0x (approved):
Assuming the only changes are in the last commit, LGTM

PR #466, opened 2021-02-18
Some more cases for index_access.spec.ts

shrekshao:
Related to https://crbug.com/dawn/622

~~Now drawIndexed with index count out of bounds of index buffer size should be invalid~~

-----

- [ ] WebGPU readability
- [ ] TypeScript readability

### DIFF
--- a/src/webgpu/api/validation/encoding/cmds/index_access.spec.ts
+++ b/src/webgpu/api/validation/encoding/cmds/index_access.spec.ts
@@ -134,6 +134,7 @@ g.test('out_of_bounds')
       .combine(pbool('indirect')) // indirect drawIndexed
       .combine([
         { indexCount: 6, firstIndex: 1 }, // indexCount + firstIndex out of bound
+        { indexCount: 0, firstIndex: 6 }, // indexCount is 0 but firstIndex out of bound
         { indexCount: 6, firstIndex: 6 }, // only firstIndex out of bound
         { indexCount: 6, firstIndex: 10000 }, // firstIndex much larger than the bound
         { indexCount: 7, firstIndex: 0 }, // only indexCount out of bound
@@ -150,8 +151,6 @@ g.test('out_of_bounds')
     if (indirect) {
       t.drawIndexedIndirect(new Uint32Array([indexCount, instanceCount, firstIndex, 0, 0]), 0);
     } else {
-      t.expectValidationError(() => {
-        t.drawIndexed(indexCount, instanceCount, firstIndex, 0, 0);
-      });
+      t.drawIndexed(indexCount, instanceCount, firstIndex, 0, 0);
     }
   });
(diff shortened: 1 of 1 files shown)


shrekshao:
Some params will crash cur chromium with `--disable-dawn-features=disallow_unsafe_apis` (on D3D12)

What is current state of validation for draw indirect (as unsafe apis)in webgpu? (Maybe it's not the focus now)
Is driver responsible for stop index out of bounds indirect draw? (Quickly skimming through d3d12 and metal indirect draw I don't see validation against this)

austinEng:
We haven't implemented any validation for drawIndirect yet. We likely need to run a compute shader to check the arguments and turn it into a no-op.

kainino0x:
I can't remember for sure, but I thought this was not a validation error, that way it matches the result you get if you do the same thing with indirect, and instead it would be treated as a dynamic out-of-bounds access and we'd no-op it instead?

Also would be good to test zero-sized index buffers as mentioned in https://crbug.com/dawn/622#c9
but that obviously doesn't have to be in this PR (it may end up being a whole new test).

austinEng:
I thought the browser could choose to do a validation error or noop or clamp. Not completely sure though.

kainino0x:
I don't think validation errors should be optional; things like this could perhaps be console warnings (though unobservable unless we add an api for warnings)

austinEng:
That's fair; I can an error could lead to non-portable behavior. Different rendering could too, but an error would be worse.

kainino0x:
Found a reference where it looks like we agreed to not have validation errors in this case: https://github.com/gpuweb/gpuweb/issues/955

shrekshao:
out of bounds drawIndexed is not invalid operation now. So simply add a few new cases to the test.

src/webgpu/api/validation/encoding/cmds/index_access.spec.ts
```diff
@@ -129,19 +129,23 @@ class F extends ValidationTest {
...
         { indexCount: 7, firstIndex: 0 }, // only indexCount out of bound
         { indexCount: 10000, firstIndex: 0 }, // indexCount much larger than the bound
+        { indexCount: 0xffffffff, firstIndex: 0xffffffff }, // max uint32 value
+        { indexCount: 0xffffffff, firstIndex: 2 }, // max uint32 indexCount and small firstIndex
+        { indexCount: 2, firstIndex: 0xffffffff }, // small indexCount and max uint32 firstIndex
       ] as const)
       .combine(poptions('instanceCount', [1, 10000])) // normal and large instanceCount
   )
-  .fn(t => {
+  .fn(async t => {
```
kainino0x, line 148: leftover
```suggestion
  .fn(t => {
```

src/webgpu/api/validation/encoding/cmds/index_access.spec.ts
```diff
@@ -129,19 +129,23 @@ class F extends ValidationTest {
 export const g = makeTestGroup(F);
 
 g.test('out_of_bounds')
-  .params(
+  .cases(
```
kainino0x, line 132: hm, probably should split this into cases and subcases while we're here. I figure only 'indirect' should be a case, the rest can be subcases because they're not very distinct/separable.

kainino0x (approved):
LGTM, 2 nits, will fix and land

PR #462, opened 2021-02-10
partially update image copy operation tests; assorted small refactors

kainino0x:
These aren't close to passing yet, but this is a best-effort to partially update them.

-----

- [x] New helpers, if any, are documented in `helper_index.md`.
- [x] Incomplete tests, if any, are marked with TODO or `.unimplemented()`.

**[Review requirements](https://github.com/gpuweb/cts/blob/main/docs/reviews.md) are satisfied for:**

- [x] WebGPU readability
- [x] TypeScript readability

PR #468, opened 2021-02-23
Fix copyBetweenLinearDataAndTexture operation tests

kainino0x:
Should be reviewed one commit at a time.

-----

- [ ] WebGPU readability
- [x] TypeScript readability


src/webgpu/api/operation/copyBetweenLinearDataAndTexture.spec.ts
```diff
@@ -364,6 +364,7 @@ class CopyBetweenLinearDataAndTextureTest extends GPUTest {
       expectedDataLayout,
```
kainino0x, line 390: Note: expectedDataLayout has an `offset` in it. It was previously obeyed on L364 but ignored on L363. It is now obeyed in both places due to the change to `updateLinearTextureDataSubBox`.

src/webgpu/api/operation/copyBetweenLinearDataAndTexture.spec.ts
```diff
@@ -733,15 +733,15 @@ g.test('copy_with_various_offsets_and_data_sizes')
...
+    const { validMinDataSize } = dataBytesForCopy(
       { offset, bytesPerRow, rowsPerImage },
       format,
       copySize,
       { method: initMethod }
     );
-    assert(valid);
+    assert(validMinDataSize !== undefined);
 
-    const dataSize = offset + minDataSize + dataPaddingInBytes;
```
kainino0x, line 744: the double-counting of `offset` was a regression in #462

src/webgpu/api/operation/copyBetweenLinearDataAndTexture.spec.ts
```diff
@@ -413,20 +414,27 @@ class CopyBetweenLinearDataAndTextureTest extends GPUTest {
...
+        texel,
+        sourceOrigin
+      );
       const destinationOffset = this.getTexelOffsetInBytes(
-        { bytesPerRow, rowsPerImage, offset: 0 },
+        destinationDataLayout,
         format,
-        texel
+        texel,
+        destinationOrigin
```
kainino0x, line 463: I think somehow the omission of offset and the omission of destinationOrigin were previously canceling each other out. I don't understand why these tests passed to begin with, but this seems right, and at least in Chrome, this makes the tests pass.

src/webgpu/util/texture/image_copy.ts
```diff
@@ -25,7 +25,7 @@ export function dataBytesForCopy(
   format: SizedTextureFormat,
   copyExtentValue: GPUExtent3D,
   { method }: { method: ImageCopyType }
-): { minDataSize: number; valid: boolean } {
+): { validMinDataSize: number | undefined; bestGuessMinDataSize: number } {
```
Kangz, line 28: "Guess" is unfortunate here, because I'm sure some tests will rely on the value of the guess. I actually liked the previous return value that said 1) whether your copy was valid 2) what the minimum size of the buffer should be. Also we can see in this commit that no test uses both `validMinDataSize` and `bestGuessMinDataSize`.

src/webgpu/util/texture/image_copy.ts
```diff
@@ -41,24 +41,33 @@ export function dataBytesForCopy(
...
-
-    bytesPerRow ??= align(info.bytesPerBlock * widthInBlocks, 256);
-    rowsPerImage ??= heightInBlocks;
+    // (a) If heightInBlocks > 1, layout.bytesPerRow must be specified.
+    // (b) If copyExtent.depth > 1, layout.bytesPerRow and layout.rowsPerImage must be specified.
+    // (c) If specified, layout.bytesPerRow must be greater than or equal to bytesInLastRow.
+    // (d) If specified, layout.rowsPerImage must be greater than or equal to heightInBlocks.
+    //
+    // But for the sake of various tests that don't actually care about the exact value, guess.
+    if (bytesPerRow !== undefined && bytesPerRow < bytesInLastRow) {
```
Kangz, line 55: This is structured a bit weird, shouldn't we do all the checks for bytesPerRow (potentially as separate boolean) and then if invalid compute the best guess? Here the state between the ifs makes things a bit complicated.

src/webgpu/api/operation/copyBetweenLinearDataAndTexture.spec.ts
```diff
@@ -79,6 +84,19 @@ const kMethodsToTest = [
   { initMethod: 'WriteTexture', checkMethod: 'PartialCopyT2B' },
 ] as const;
 
+const kExcludedFormats: Set<SizedTextureFormat> = new Set([
```
Kangz, line 95: Can we add a comment that this list should be reduced to 0?

src/webgpu/api/operation/copyBetweenLinearDataAndTexture.spec.ts
```diff
@@ -790,42 +791,42 @@ g.test('copy_with_various_origins_and_copy_extents')
     const info = kSizedTextureFormatInfo[format];
     await t.selectDeviceOrSkipTestCase(info.extension);
 
-    const origin: Required<GPUOrigin3DDict> = { x: info.blockWidth, y: info.blockHeight, z: 1 };
-    const copySize = { width: 2 * info.blockWidth, height: 2 * info.blockHeight, depth: 2 };
-    const textureSize: [number, number, number] = [3 * info.blockWidth, 3 * info.blockHeight, 3];
+    const originBlks = [1, 1, 1];
```
Kangz, line 841: nit: just Blocks? We're saving only two chars here.

src/webgpu/api/operation/copyBetweenLinearDataAndTexture.spec.ts
```diff
@@ -1013,69 +1014,43 @@ g.test('copy_various_mip_levels')
 // slice we can set rowsPerImage to 0. Origin, offset, mipLevel and rowsPerImage values will be set
 // to undefined when appropriate.
 g.test('copy_with_no_image_or_slice_padding_and_undefined_values')
+  .desc(
+    `TODO: description
+
+  TODO: "0" below really means "undefined" (it gets replaced with undefined later).
```
Kangz, line 1068: Didn't you just do this cleanup?

src/webgpu/api/operation/copyBetweenLinearDataAndTexture.spec.ts
Kangz: I'm not sure :(

Kangz (approved):
LGTM overall

src/webgpu/util/texture/image_copy.ts
kainino0x: The reason for guessing is that a number of validation tests are attempting to test conditions on bytesPerRow/rowsPerImage but we want to make sure we don't *also* trigger an error on "buffer wasn't big enough".

This is really just a refactor; same semantics but exposed differently. Exposing validMinDataSize and bestGuessMinDataSize separately ensures callers don't forget to `assert(valid)` if they should.

I'm not sure what would be better than this. Two separate entry points, one which returns best guess and one which throws if not valid?

src/webgpu/api/operation/copyBetweenLinearDataAndTexture.spec.ts
kainino0x: Only partly, I only did it for bytesPerRow/rowsPerImage (in the params, but not in the rest of the test code). Though I suppose I can easily do it for the others too right now.

src/webgpu/util/texture/image_copy.ts
kainino0x: There's still a bit of inbetween state, but I think this is clearer. PTAL.

src/webgpu/util/texture/image_copy.ts
kainino0x: Made two helper entry points.

src/webgpu/api/operation/command_buffer/image_copy.spec.ts
```diff
@@ -1059,31 +1061,28 @@ g.test('mip_levels')
...
       // copying two slices
-      { copySize: [3, 3, 2], origin: [0, 0, 0], bytesPerRow: 256, rowsPerImage: 3 },
-      // origin.x = undefined
-      { copySize: [1, 1, 1], origin: [0, 1, 1], bytesPerRow: undefined, rowsPerImage: undefined },
-      // origin.y = undefined
-      { copySize: [1, 1, 1], origin: [1, 0, 1], bytesPerRow: undefined, rowsPerImage: undefined },
-      // origin.z = undefined
-      { copySize: [1, 1, 1], origin: [1, 1, 0], bytesPerRow: undefined, rowsPerImage: undefined },
+      { copySize: [3, 3, 2], origin: [UND, UND, UND], bytesPerRow: 256, rowsPerImage: 3 },
+      // origin.x = und
```
Kangz, line 1077: nit: undefined.

Kangz (approved):

PR #469, opened 2021-02-23
Add validation tests for Timestamp Query

haoxli:
- Test writeTimestamp in all possible encoders.
- Test resolveQuerySet for all types of queries.
- Test destoryed query set in writeTimestamp and resolveQuerySet.

-----

- [ ] WebGPU readability
- [ ] TypeScript readability

### DIFF
--- a/src/webgpu/api/validation/encoding/queries/general.spec.ts
+++ b/src/webgpu/api/validation/encoding/queries/general.spec.ts
@@ -5,10 +5,100 @@ TODO:
     - queryIndex {in, out of} range for GPUQuerySet
     - GPUQuerySet {valid, invalid}
         - or {undefined}, for occlusionQuerySet
-    - ?
+    - x = {occlusion, pipeline statistics, timestamp} query
 `;
 
+import { params, poptions } from '../../../../../common/framework/params_builder.js';
 import { makeTestGroup } from '../../../../../common/framework/test_group.js';
+import { kQueryTypes } from '../../../../capability_info.js';
 import { ValidationTest } from '../../validation_test.js';
 
+async function selectDeviceForQueryType(t: ValidationTest, type: GPUQueryType): Promise<void> {
+  const extensions: GPUExtensionName[] = [];
+  if (type === 'pipeline-statistics') {
+    extensions.push('pipeline-statistics-query');
+  } else if (type === 'timestamp') {
+    extensions.push('timestamp-query');
+  }
+
+  await t.selectDeviceOrSkipTestCase({ extensions });
+}
+
+export const enum EncoderType {
+  CommandEncoder = 'CommandEncoder',
+  ComputeEncoder = 'ComputeEncoder',
+  RenderEncoder = 'RenderEncoder',
+}
+
 export const g = makeTestGroup(ValidationTest);
+
+g.test('writeTimestamp')
+  .desc(
+    `
+Tests that write timestamp to all types of query set on all possible encoders:
+- queryIndex {in, out of} range for GPUQuerySet
+- GPUQuerySet {valid, invalid}
+- x= {occlusion, pipeline statistics, timestamp} query
+- x= {compute, render, non-pass} enconder
+  `
+  )
+  .params(
+    params()
+      .combine(
+        poptions('encoderType', [
+          EncoderType.CommandEncoder,
+          EncoderType.ComputeEncoder,
+          EncoderType.RenderEncoder,
+        ] as const)
+      )
+      .combine(poptions('type', kQueryTypes))
+      .combine(poptions('queryIndex', [0, 1, 2]))
+      .unless(({ type, queryIndex }) => type !== 'timestamp' && queryIndex !== 0)
+  )
+  .fn(async t => {
+    const { encoderType, type, queryIndex } = t.params;
+
+    await selectDeviceForQueryType(t, type);
+
+    const count = 2;
+    const pipelineStatistics =
+      type === 'pipeline-statistics' ? (['clipper-invocations'] as const) : ([] as const);
+    const querySet = t.device.createQuerySet({ type, count, pipelineStatistics });
+
+    const encoder = t.device.createCommandEncoder();
+
+    switch (encoderType) {
+      case EncoderType.CommandEncoder: {
+        encoder.writeTimestamp(querySet, queryIndex);
+        break;
+      }
+      case EncoderType.ComputeEncoder: {
+        const pass = encoder.beginComputePass();
+        pass.writeTimestamp(querySet, queryIndex);
+        pass.endPass();
+        break;
+      }
+      case EncoderType.RenderEncoder: {
+        const colorAttachment = t.device.createTexture({
+          format: 'rgba8unorm',
+          size: { width: 4, height: 4, depth: 1 },
+          usage: GPUTextureUsage.RENDER_ATTACHMENT,
+        });
+        const pass = encoder.beginRenderPass({
+          colorAttachments: [
+            {
+              attachment: colorAttachment.createView(),
+              loadValue: { r: 1.0, g: 1.0, b: 1.0, a: 1.0 },
+            },
+          ],
+        });
+        pass.writeTimestamp(querySet, queryIndex);
+        pass.endPass();
+        break;
+      }
+    }
+
+    t.expectValidationError(() => {
+      encoder.finish();
+    }, type !== 'timestamp' || queryIndex >= count);
+  });
--- a/src/webgpu/api/validation/encoding/queries/general.spec.ts
+++ b/src/webgpu/api/validation/encoding/queries/general.spec.ts
@@ -13,92 +13,68 @@ import { makeTestGroup } from '../../../../../common/framework/test_group.js';
 import { kQueryTypes } from '../../../../capability_info.js';
 import { ValidationTest } from '../../validation_test.js';
 
-async function selectDeviceForQueryType(t: ValidationTest, type: GPUQueryType): Promise<void> {
-  const extensions: GPUExtensionName[] = [];
-  if (type === 'pipeline-statistics') {
-    extensions.push('pipeline-statistics-query');
-  } else if (type === 'timestamp') {
-    extensions.push('timestamp-query');
+class F extends ValidationTest {
+  async selectDeviceForQuerySetOrSkipTestCase(type: GPUQueryType): Promise<void> {
+    return this.selectDeviceOrSkipTestCase(
+      type === 'pipeline-statistics'
+        ? 'pipeline-statistics-query'
+        : type === 'timestamp'
+        ? 'timestamp-query'
+        : undefined
+    );
   }
-
-  await t.selectDeviceOrSkipTestCase({ extensions });
 }
 
-export const enum EncoderType {
-  CommandEncoder = 'CommandEncoder',
-  ComputeEncoder = 'ComputeEncoder',
-  RenderEncoder = 'RenderEncoder',
-}
+export const g = makeTestGroup(F);
 
-export const g = makeTestGroup(ValidationTest);
-
-g.test('writeTimestamp')
+g.test('writeTimestamp,query_type_and_index')
   .desc(
     `
 Tests that write timestamp to all types of query set on all possible encoders:
+- type {occlusion, pipeline statistics, timestamp}
 - queryIndex {in, out of} range for GPUQuerySet
-- GPUQuerySet {valid, invalid}
-- x= {occlusion, pipeline statistics, timestamp} query
-- x= {compute, render, non-pass} enconder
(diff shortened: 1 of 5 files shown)


src/webgpu/api/validation/encoding/queries/general.spec.ts
```diff
@@ -5,10 +5,100 @@ TODO:
...
+    await selectDeviceForQueryType(t, type);
+
+    const count = 2;
+    const pipelineStatistics =
+      type === 'pipeline-statistics' ? (['clipper-invocations'] as const) : ([] as const);
+    const querySet = t.device.createQuerySet({ type, count, pipelineStatistics });
+
+    const encoder = t.device.createCommandEncoder();
+
+    switch (encoderType) {
```
kainino0x, line 70: It's possible to simplify this by using `ValidationTest.createEncoder` but you'd need to add one more overload. I took a look at this and actually managed to generalize it so that's not necessary, see #471 and feel free to depend on it in this PR.

For this change, the `encoderType` will have to change to use the strings defined by `kEncoderTypes` in `validation_test.ts`.

src/webgpu/api/validation/encoding/queries/general.spec.ts
```diff
@@ -5,10 +5,100 @@ TODO:
...
+      .combine(
+        poptions('encoderType', [
+          EncoderType.CommandEncoder,
+          EncoderType.ComputeEncoder,
+          EncoderType.RenderEncoder,
+        ] as const)
+      )
+      .combine(poptions('type', kQueryTypes))
+      .combine(poptions('queryIndex', [0, 1, 2]))
+      .unless(({ type, queryIndex }) => type !== 'timestamp' && queryIndex !== 0)
```
kainino0x, line 56: Can you explain why these cases are skipped in a comment?

Generally we avoid skipping cases in validation tests, if it's not too hard, because it makes it easier to reason about what's being tested (don't have to think about the contents of the `unless`).

src/webgpu/api/validation/encoding/queries/general.spec.ts
```diff
@@ -5,10 +5,100 @@ TODO:
...
+  .desc(
+    `
+Tests that write timestamp to all types of query set on all possible encoders:
+- queryIndex {in, out of} range for GPUQuerySet
+- GPUQuerySet {valid, invalid}
+- x= {occlusion, pipeline statistics, timestamp} query
+- x= {compute, render, non-pass} enconder
+  `
+  )
+  .params(
```
kainino0x, line 45: Please use `.cases` and `.subcases`. In my opinion, `encoderType` and `type` should be cases, while `queryIndex` should be subcases.

src/webgpu/api/validation/queue/destroyed/query_set.spec.ts
```diff
@@ -9,7 +9,107 @@ TODO: implement. (Search for other places some of these cases may have already b
...
 export const g = makeTestGroup(ValidationTest);
+
+g.test('writeTimestamp')
+  .desc(
+    `
+Tests that use a destroyed query set in writeTimestamp on {compute, render, non-pass} encoder.
+- x= {destroyed, not destroyed (control case)}
+  `
+  )
+  .params(
```
kainino0x, line 31: Please use `cases` `encoderType` and `subcases` `destroyed`

src/webgpu/api/validation/encoding/queries/resolveQuerySet.spec.ts
```diff
@@ -1,13 +1,94 @@
 export const description = `
-TODO:
-- invalid GPUQuerySet
```
kainino0x, line 3: I don't see any tests with _invalid_ `GPUQuerySet`s, i.e. attempting to use `GPUQuerySet`s that failed during creation.

Can you either leave a TODO (or better, a `.unimplemented()` test) for that?

src/webgpu/api/validation/encoding/queries/resolveQuerySet.spec.ts
```diff
@@ -1,13 +1,94 @@
...
-- firstQuery and/or queryCount out of range
-- invalid destination buffer
```
kainino0x, line 5: This is referring to an invalid buffer, i.e. failed during creation. See `ValidationTest.createBufferWithState` but feel free to leave this `TODO` or `.unimplemented()` in this PR.

src/webgpu/api/validation/encoding/queries/resolveQuerySet.spec.ts
```diff
@@ -1,13 +1,94 @@
...
-- destinationOffset out of range
-- ?
+Validation tests for resolveQuerySet.
 `;
```
kainino0x, line 3: nit: It would have been helpful to have a first PR which wrote an expanded plan but didn't start implementing tests. It would have caught these issues more quickly. Though I know the review latency makes this harder, so it wasn't really a problem for these relatively simple tests.

src/webgpu/api/validation/encoding/queries/resolveQuerySet.spec.ts
```diff
@@ -1,13 +1,94 @@
...
+
+    const encoder = t.device.createCommandEncoder();
+    encoder.resolveQuerySet(querySet, firstQuery, queryCount, destination, 0);
+
+    t.expectValidationError(() => {
+      encoder.finish();
+    }, firstQuery + queryCount > kQueryCount);
+  });
+
+g.test('resolveQuerySet,destination_buffer')
```
kainino0x, line 43: Good catch, missed this in the initial plan. nit: `destination_usage` or `destination,usage` or something

src/webgpu/api/validation/encoding/queries/resolveQuerySet.spec.ts
```diff
@@ -1,13 +1,94 @@
...
+  .desc(
+    `
+Tests that resolve query set with invalid destinationBuffer:
+- Buffer usage {with, without} QUERY_RESOLVE
+  `
+  )
+  .params(
```
kainino0x, line 50: could be subcases

src/webgpu/api/validation/encoding/queries/resolveQuerySet.spec.ts
```diff
@@ -1,13 +1,94 @@
...
+export const kQueryCount = 2;
+
+g.test('resolveQuerySet,first_query_and_query_count')
+  .desc(
+    `
+Tests that resolve query set with invalid firstQuery and queryCount:
+- firstQuery and/or queryCount out of range
+  `
+  )
+  .params([
```
kainino0x, line 20: could be subcases

src/webgpu/api/validation/encoding/queries/resolveQuerySet.spec.ts
```diff
@@ -1,13 +1,94 @@
...
+g.test('resolveQuerySet,destination_offset')
+  .desc(
+    `
+Tests that resolve query set with invalid destinationOffset:
+- destinationOffset is not a multiple of 8
+- The size of destinationBuffer - destinationOffset < queryCount * 8
+- destinationOffset out of range
+  `
+  )
+  .params(poptions('destinationOffset', [0, 6, 8, 16]))
```
kainino0x, line 80: could be subcases

src/webgpu/api/validation/queue/destroyed/query_set.spec.ts
```diff
@@ -9,7 +9,107 @@ TODO: implement. (Search for other places some of these cases may have already b
...
+
+    const querySet = t.device.createQuerySet({ type: 'timestamp', count: 2 });
+
+    if (t.params.destroyed) {
+      querySet.destroy();
+    }
+
+    const encoder = t.device.createCommandEncoder();
+
+    switch (t.params.encoderType) {
```
kainino0x, line 53: Same as in the other file

src/webgpu/api/validation/queue/destroyed/query_set.spec.ts
```diff
@@ -9,7 +9,107 @@ TODO: implement. (Search for other places some of these cases may have already b
 Consider whether these tests should be distributed throughout the suite, instead of centralized.
```
kainino0x, line 9: Do you have an opinion about this after working with these tests? Feedback would be useful.

src/webgpu/api/validation/queue/destroyed/query_set.spec.ts
```diff
@@ -9,7 +9,107 @@ TODO: implement. (Search for other places some of these cases may have already b
...
+  });
+
+g.test('resolveQuerySet')
+  .desc(
+    `
+Tests that use a destroyed query set in resolveQuerySet.
+- x= {destroyed, not destroyed (control case)}
+  `
+  )
+  .params(pbool('destroyed'))
```
kainino0x, line 96: subcases

kainino0x (commented):
Overall looking good but I left a lot of small comments. Reviewed `general` closely, but I haven't reviewed the tests in `resolveQuerySet` and `query_set` VERY closely yet (they look good overall).

src/webgpu/api/validation/queue/destroyed/query_set.spec.ts
haoxli: These validations happen on queue submit, it is a good choice to put together the cases with same check points.

src/webgpu/api/validation/encoding/queries/general.spec.ts
```diff
@@ -13,92 +13,68 @@ import { makeTestGroup } from '../../../../../common/framework/test_group.js';
...
+g.test('writeTimestamp,query_type_and_index')
   .desc(
     `
 Tests that write timestamp to all types of query set on all possible encoders:
+- type {occlusion, pipeline statistics, timestamp}
 - queryIndex {in, out of} range for GPUQuerySet
-- GPUQuerySet {valid, invalid}
-- x= {occlusion, pipeline statistics, timestamp} query
-- x= {compute, render, non-pass} enconder
+- x= {non-pass, compute, render} enconder
```
kainino0x, line 36:
```suggestion
- x= {non-pass, compute, render} encoder
```

src/webgpu/api/validation/encoding/queries/resolveQuerySet.spec.ts
```diff
@@ -10,14 +10,47 @@ export const g = makeTestGroup(ValidationTest);
...
+Tests that resolve query set with invalid object.
+- invalid GPUQuerySet that failed during creation.
+- invalid destination buffer that failed during creation.
+  `
+  )
+  .subcases(
+    () =>
+      [
+        { querySetState: 'invalid', destinationState: 'valid' },
+        { querySetState: 'valid', destinationState: 'invalid' },
```
kainino0x, line 25: Please also test with both being valid, as a control case, to make sure the test code is valid aside from the querySetState/destinationState.

kainino0x (approved):
Sorry for the delay. Just a few tiny things, but LGTM!

src/webgpu/api/validation/encoding/queries/resolveQuerySet.spec.ts
haoxli: The test with both being valid should be same with the control case in `resolveQuerySet,first_query_and_query_count`. Do we need to add control cases for each `cases` or `subcases`, even they have been tested?

kainino0x: We should, yes. The point of the control case is not to the test the browser, but to test the test - that is, make sure there aren't any other mistakes in the test function (and especially that no mistakes get introduced in the future).

src/webgpu/api/validation/encoding/queries/resolveQuerySet.spec.ts
```diff
@@ -21,6 +21,7 @@ Tests that resolve query set with invalid object.
   .subcases(
     () =>
       [
+        { querySetState: 'valid', destinationState: 'valid' }, // control case
```
kainino0x, line 24: Doesn't this require changing the expectValidationError call below too?

kainino0x (approved):

src/webgpu/api/validation/encoding/queries/general.spec.ts
```diff
@@ -5,10 +5,76 @@ TODO:
     - queryIndex {in, out of} range for GPUQuerySet
     - GPUQuerySet {valid, invalid}
         - or {undefined}, for occlusionQuerySet
-    - ?
+    - x = {occlusion, pipeline statistics, timestamp} query
```
kainino0x, line 8: I just realized that this PR leaves this TODO in here, but I think most or all of it is done. Can you please update the TODO? It's really important for us to track what work remains. Sorry for missing this the first time. @haoxli

PR #477, opened 2021-02-26
Bug fixes for new texture formats

Richard-Yunchao:
-----

- [ ] WebGPU readability
- [ ] TypeScript readability

### DIFF
--- a/src/webgpu/api/validation/attachment_compatibility.spec.ts
+++ b/src/webgpu/api/validation/attachment_compatibility.spec.ts
@@ -14,7 +14,6 @@ import {
   kUnsizedDepthStencilFormats,
   kTextureSampleCounts,
   kMaxColorAttachments,
-  kAllTextureFormatInfo,
 } from '../../capability_info.js';
 
 import { ValidationTest, CommandBufferMaker } from './validation_test.js';
@@ -135,27 +134,6 @@ class F extends ValidationTest {
       depthStencilState,
       sampleCount,
     });
-  }
-
-  async selectDeviceForTextureFormatOrSkipTestCase(
-    formats: GPUTextureFormat | undefined | (GPUTextureFormat | undefined)[]
-  ): Promise<void> {
-    if (!Array.isArray(formats)) {
-      formats = [formats];
-    }
-    const extensions = new Set<GPUExtensionName>();
-    for (const format of formats) {
-      if (format !== undefined) {
-        const formatExtension = kAllTextureFormatInfo[format].extension;
-        if (formatExtension !== undefined) {
-          extensions.add(formatExtension);
-        }
-      }
-    }
-
-    if (extensions.size) {
-      await this.selectDeviceOrSkipTestCase({ extensions });
-    }
   }
 }
 
--- a/src/webgpu/api/validation/attachment_compatibility.spec.ts
+++ b/src/webgpu/api/validation/attachment_compatibility.spec.ts
@@ -142,8 +142,7 @@ class F extends ValidationTest {
     formats: (DepthStencilFormat | undefined)[]
   ): GPUExtensionName[] {
     const extensions: GPUExtensionName[] = [];
-    for (let i = 0; i < formats.length; i++) {
-      const format = formats[i];
+    for (const format of new Set(formats)) {
       if (format !== undefined) {
         const formatExtension = kDepthStencilFormatInfo[format].extension;
         if (formatExtension !== undefined) {
--- a/src/webgpu/api/validation/attachment_compatibility.spec.ts
+++ b/src/webgpu/api/validation/attachment_compatibility.spec.ts
@@ -137,19 +137,22 @@ class F extends ValidationTest {
     });
   }
 
-  getTextureFormatsRequiredExtensions(
+  async selectDeviceForTextureFormatOrSkipTestCase(
     formats: (GPUTextureFormat | undefined)[]
-  ): GPUExtensionName[] {
-    const extensions: GPUExtensionName[] = [];
-    for (const format of new Set(formats)) {
+  ): Promise<void> {
+    const extensions = new Set<GPUExtensionName>();
+    for (const format of formats) {
       if (format !== undefined) {
         const formatExtension = kAllTextureFormatInfo[format].extension;
         if (formatExtension !== undefined) {
-          extensions.push(formatExtension);
+          extensions.add(formatExtension);
         }
       }
     }
-    return extensions;
+
+    if (extensions.size) {
+      await this.selectDeviceOrSkipTestCase({ extensions });
+    }
   }
 }
 
@@ -224,11 +227,7 @@ g.test('render_pass_and_bundle,depth_format')
   )
   .fn(async t => {
     const { passFormat, bundleFormat } = t.params;
-
-    const extensions = t.getTextureFormatsRequiredExtensions([passFormat, bundleFormat]);
-    if (extensions.length) {
-      await t.selectDeviceOrSkipTestCase({ extensions });
-    }
+    await t.selectDeviceForTextureFormatOrSkipTestCase([passFormat, bundleFormat]);
 
     const bundleEncoder = t.device.createRenderBundleEncoder({
       colorFormats: ['rgba8unorm'],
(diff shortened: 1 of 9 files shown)


Richard-Yunchao:
PTAL. Thanks!

Now that this PR https://github.com/gpuweb/cts/pull/475 was merged, it introduced a few new texture formats and may cause CTS failures. I didn't see a lot failures. One issue introduced is that we need to query the extensions needed for some formats introduced when we create device, in order to skip the test if the extensions are not supported. Otherwise, it might fail unexpectedly.

austinEng (approved):
Looks good, up to my knowledge

src/webgpu/api/validation/createRenderPipeline.spec.ts
```diff
@@ -141,6 +141,8 @@ g.test('color_formats_must_be_renderable')
 
     const descriptor = t.getDescriptor({ colorStates: [{ format }] });
 
+    await t.selectDeviceOrSkipTestCase(info.extension);
```
kainino0x, line 144: nit: Do this as early as possible (right after `info`)

kainino0x (commented):
LG, but would be nice to make this nicer for the whole test suite while we're here.

src/webgpu/api/validation/attachment_compatibility.spec.ts
```diff
@@ -135,6 +137,22 @@ class F extends ValidationTest {
       sampleCount,
     });
   }
+
+  getTextureFormatsRequiredExtensions(
+    formats: (DepthStencilFormat | undefined)[]
```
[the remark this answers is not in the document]
Richard-Yunchao: Not really, `undefined` is added on line 26. In addition, we can't use GPUTextureFormat because of the type check. Which is very strict and very silly. For example: 
very strict: type subset and type will be thought mismatch and not allowed. For example, DepthStencilFormat is a subset of GPUTextureFormat / AllTextureFormats, but it will report mismatch if the formats are DepthStencilFormat and you use kAllTextureFormatInfo(format)).
 very silly: you may exclude `undefined` somewhere, but it will complain type mismatch because it doesn't know you exclude `undefined` in its caller or caller's caller, etc. It seems to me that you need to exclued `undefined` exactly where you use it.

src/webgpu/api/validation/attachment_compatibility.spec.ts
```diff
@@ -206,8 +224,14 @@ g.test('render_pass_and_bundle,depth_format')
       .combine(poptions('passFormat', kDepthStencilAttachmentFormats))
       .combine(poptions('bundleFormat', kDepthStencilAttachmentFormats))
   )
-  .fn(t => {
+  .fn(async t => {
     const { passFormat, bundleFormat } = t.params;
+
+    const extensions = t.getTextureFormatsRequiredExtensions([passFormat, bundleFormat]);
```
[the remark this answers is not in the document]
Richard-Yunchao: See the comment above. It is hard to generalize this function, it will report type mismatch because the types might be:
UnsizedDepthStencilFormat,
DepthStencilFormat,
UncompressedTextureFormat,
AllTextureFormat / GPUTextureFormat

But it requires strict match.

src/webgpu/api/validation/attachment_compatibility.spec.ts
```diff
@@ -135,6 +137,22 @@ class F extends ValidationTest {
       sampleCount,
     });
   }
+
+  getTextureFormatsRequiredExtensions(
+    formats: (DepthStencilFormat | undefined)[]
+  ): GPUExtensionName[] {
+    const extensions: GPUExtensionName[] = [];
+    for (let i = 0; i < formats.length; i++) {
+      const format = formats[i];
```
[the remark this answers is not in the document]
Richard-Yunchao: Done.

src/webgpu/api/validation/createRenderPipeline.spec.ts
Richard-Yunchao: Ack. Done.

src/webgpu/api/validation/attachment_compatibility.spec.ts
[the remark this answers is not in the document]
kainino0x: Ah, sorry, didn't realize why undefined was needed, but it makes sense and should be kept.

Let me think about the other problem.

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -297,6 +297,8 @@ g.test('it_is_invalid_to_have_an_output_attachment_texture_with_non_renderable_f
 
     const descriptor = t.getDescriptor({ width: 1, height: 1, format });
 
+    await t.selectDeviceOrSkipTestCase(info.extension);
```
kainino0x, line 300: nit: Sorry, other nit was supposed to apply to all instances. Please move this earlier here too.

src/webgpu/api/validation/encoding/cmds/copyTextureToTexture.spec.ts
```diff
@@ -349,6 +349,9 @@ g.test('depth_stencil_copy_restrictions')
     } = t.params;
 
     const kMipLevelCount = 3;
+
+    await t.selectDeviceOrSkipTestCase(kAllTextureFormatInfo[format].extension);
```
kainino0x, line 353: same

src/webgpu/api/validation/attachment_compatibility.spec.ts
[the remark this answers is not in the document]
kainino0x: I didn't have an issue with it. I will push a commit to your branch.

src/webgpu/api/validation/attachment_compatibility.spec.ts
```diff
@@ -135,6 +137,21 @@ class F extends ValidationTest {
       sampleCount,
     });
   }
+
+  getTextureFormatsRequiredExtensions(
+    formats: (DepthStencilFormat | undefined)[]
+  ): GPUExtensionName[] {
+    const extensions: GPUExtensionName[] = [];
+    for (const format of new Set(formats)) {
```
kainino0x: Oops, I realized this isn't actually going to prevent duplicate extension names. My bad. Fixed in push.

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: Done.

src/webgpu/api/validation/encoding/cmds/copyTextureToTexture.spec.ts
Richard-Yunchao: Done. Sorry, I didn't look into other tests when you pointed out this problem.

src/webgpu/api/validation/attachment_compatibility.spec.ts
```diff
@@ -138,8 +138,11 @@ class F extends ValidationTest {
   }
 
   async selectDeviceForTextureFormatOrSkipTestCase(
-    formats: (GPUTextureFormat | undefined)[]
+    formats: GPUTextureFormat | undefined | (GPUTextureFormat | undefined)[]
```
Richard-Yunchao, line 141: Interesting. This can resolve the type mismatch issue... Thanks!

kainino0x: I'm not exactly sure what type mismatch issue you had, but: in general if you can't get the types to do what you actually want, you can try to do the "right thing" and use `as` casts to ignore the type issues - and then sometimes it's possible later to get rid of `as` with the right magic words.

src/webgpu/api/validation/attachment_compatibility.spec.ts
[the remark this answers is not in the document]
kainino0x: > The resulting helper would be generally useful (it could be extended to all texture formats) and could live in GPUTest. (In which case please document it in `helper_index`.) It should work for all of the cases in the rest of this PR I think.

I think this is the only outstanding comment

src/webgpu/api/validation/attachment_compatibility.spec.ts
Richard-Yunchao: Got it. I realized I used `as` in previous patches, but forgot to try it here. Thanks!

src/webgpu/api/validation/attachment_compatibility.spec.ts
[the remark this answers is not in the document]
Richard-Yunchao: Done. I also documented another helper function I added a few days ago.

kainino0x (approved):
lgtm, thanks for the maintenance work! (and remembering that other helper)

PR #478, opened 2021-03-02
Add sample count validation test for createTexture

Richard-Yunchao:
-----

- [x] WebGPU readability
- [x] TypeScript readability


src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -216,6 +216,86 @@ g.test('mipLevelCount,bound_check,bigger_than_integer_bit_width')
...
+      size: [32, 32, 1],
+      sampleCount,
+      format,
+      usage: GPUTextureUsage.SAMPLED,
+    };
+
+    await t.selectDeviceOrSkipTestCase(kAllTextureFormatInfo[format].extension);
+
+    const success =
+      sampleCount === 1 || (sampleCount === 4 && format in kUncompressedTextureFormats);
```
Richard-Yunchao, line 238: I intended to use the code snippet to say that it will success if: 
1) sample count is 1, or
2) sample count is 4 and the format is not compressed texture formats. 

But looks like it is not consistent with my intention. 

Is there a way to debug the script code?

kainino0x: You can use the standard JS debugger in Chrome dev tools. If you are using `npm run dev` then it will source-map the generated JS back to TypeScript.

You can use the little "log stack to console" button image on a test or log line to print a stack you can use to jump to the test code.

The issue here is that `in` checks on object properties, not array elements; you have to use `Array.includes`:
```suggestion
      sampleCount === 1 || (sampleCount === 4 && kUncompressedTextureFormats.includes(format));
```

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -216,6 +216,86 @@ g.test('mipLevelCount,bound_check,bigger_than_integer_bit_width')
...
+    t.expectValidationError(() => {
+      t.device.createTexture(descriptor);
+    }, !success);
+  });
+
+g.test('sampleCount,valid_sampleCount_with_other_parameter_varies')
+  .desc(
+    `Test texture creation with valid sample count when dimensions, arrayLayerCount, mipLevelCount, format, and usage varies.`
```
kainino0x, line 246: Please describe what validation conditions are being tested here based on the condition below ("multisample textures have dimension 2d, a supported format, 1 mip level, ...")

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -216,6 +216,86 @@ g.test('mipLevelCount,bound_check,bigger_than_integer_bit_width')
     });
   });
 
+g.test('sampleCount,various_sampleCount_with_all_formats')
+  .desc(`Test texture creation with various (valid or invalid) sample count and all formats`)
+  .subcases(() =>
+    params()
+      .combine(poptions('sampleCount', [1, 2, 4, 8, 16, 32, 256]))
```
kainino0x, line 223: Should test 0 as well, which is an invalid sampleCount.

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -216,6 +216,86 @@ g.test('mipLevelCount,bound_check,bigger_than_integer_bit_width')
...
+        mipLevelCount === 1 &&
+        arrayLayerCount === 1 &&
+        (usage & GPUConst.TextureUsage.STORAGE) === 0);
+
+    t.expectValidationError(() => {
+      t.device.createTexture(descriptor);
+    }, !success);
+  });
+
 g.test('sampleCount')
```
kainino0x, line 299: I think this entire test is subsumed by your more thorough tests. If you agree, let's go ahead and delete it.

kainino0x (approved):
LGTM!

src/webgpu/api/validation/createTexture.spec.ts

src/webgpu/api/validation/createTexture.spec.ts

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: Agreed. This PR is not ready at that time, so I didn't remove the original code. 
Done.

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: It leads to type mismatch between kUncompressedTextureFormats and format. And I can't use `as` here. However, if I use a user-defined constant exactly the same with kUncompressedTextureFormats, it will ignore the type check. It is like a workaround. Maybe there are better solutions.

I really think the type check between kUncompressedTextureFormats and format is not necessary. Otherwise, `includes` is not effective as it supposed to be (if they are exactly the same, why we name it `includes`. `include` means a set and a subset).

Richard-Yunchao:
FYI. @bbernhar and @RafaelCintron

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: > You can use the standard JS debugger in Chrome dev tools. If you are using `npm run dev` then it will source-map the generated JS back to TypeScript.
> 
> You can use the little "log stack to console" button image on a test or log line to print a stack you can use to jump to the test code.

Thank you, Kai.

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -29,14 +29,59 @@ import { makeTestGroup } from '../../../common/framework/test_group.js';
...
   kAllTextureFormatInfo,
-  kUncompressedTextureFormats,
   kTextureUsages,
 } from '../../capability_info.js';
 import { GPUConst } from '../../constants.js';
 import { maxMipLevelCount } from '../../util/texture/base.js';
 
 import { ValidationTest } from './validation_test.js';
 
+const kValidTextureFormatsForMSAA = [
```
kainino0x, line 39: If we can't simply used "Uncompressed" then IMO we should add this info to the format capability tables (as a new column, `multisample: boolean`). Then instead of `kValidTextureFormatsForMSAA.includes(fmt)`, `kAllTextureFormatInfo[fmt].multisample`

kainino0x:
One more comment (and there are some minor conflicts, which wouldn't have happened if I got to this sooner, sorry about the long delay 😞 )

src/webgpu/api/validation/createTexture.spec.ts

Richard-Yunchao:
> One more comment (and there are some minor conflicts, which wouldn't have happened if I got to this sooner, sorry about the long delay  )

Resolved (no problem. :)  )

kainino0x:
lgtm, thanks!

PR #479, opened 2021-03-03
Add texture size tests for 1D/2D/3D textures

Richard-Yunchao:
-----

- [x] WebGPU readability
- [x] TypeScript readability

### DIFF
--- a/src/webgpu/api/validation/createTexture.spec.ts
+++ b/src/webgpu/api/validation/createTexture.spec.ts
@@ -10,15 +10,6 @@ TODO: review existing tests and merge with this plan:
 >     - with format that supports multisample, with all possible dimensions
 >     - with dimension that support multisample, with all possible formats
 >     - with format-dimension that support multisample, with {mipLevelCount, array layer count} = {1, 2}
-> - 1d, {width, height, depthOrArrayLayers} > whatever the max is
->     - height max is 1 (unless 1d-array is added)
->     - depthOrArrayLayers max is 1
->     - x= every texture format
-> - 2d, {width, height, depthOrArrayLayers} > whatever the max is
->     - depthOrArrayLayers max differs from width/height
->     - x= every texture format
-> - 3d, {width, height, depthOrArrayLayers} > whatever the max is
->     - x= every texture format
 > - usage flags
 >     - {0, ... each single usage flag}
 >     - x= every texture format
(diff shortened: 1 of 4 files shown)


Richard-Yunchao:
This PR is pending on https://github.com/gpuweb/types/pull/59. FYI @austinEng

Kangz (approved):
LGTM

kainino0x:
Doesn't sound like I need to review; feel free to land after rolling in types v0.0.42 :)

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -238,6 +230,140 @@ g.test('sampleCount')
...
+          DefaultLimits.maxTextureDimension2D + 1,
+        ])
+      )
+      .combine(
+        poptions('depthOrArrayLayers', [
+          DefaultLimits.maxTextureArrayLayers - 1,
+          DefaultLimits.maxTextureArrayLayers,
+          DefaultLimits.maxTextureArrayLayers + 1,
+        ])
+      )
```
Richard-Yunchao, line 293: It turned out that width*height*depthOrArrayLayers will lead to out-of-memory issue. I remember that we don't have any limit about the entire texture memory footprint (width*height*dpehtOrArrayLayers). Right? So I think I need to change other parameters to 1 when I test the bound of each parameter.

Richard-Yunchao:
FYI. @bbernhar and @RafaelCintron . I will cc non-trivial PRs to you via @ you.

src/webgpu/api/validation/createTexture.spec.ts
kainino0x: Since these are just validation tests, we don't need to test create max * max * max sized textures. We're only concerned here with making sure the implementation enforces the maximum exactly. So you could just have 9 cases instead of 3 * 3 * 3:

{ width: max - 1, height: 1, depth: 1 }
{ width: max, height: 1, depth: 1 }
{ width: max + 1, height: 1, depth: 1 }
{ width: 1, height: max - 1, depth: 1 }
...

Richard-Yunchao: Exactly. I am doing so right now.

Richard-Yunchao: Another problem is the alignment issue for compressed texture: max - 1 is valid for uncompressed texture but it is invalid for compressed texture . I just skipped compressed formats and added a TODO. I will submit separate PR soon.

kainino0x: Sounds good. Maybe the way to do this is parameterize over "diff from max in blocks" and test -1,0,1 and compute the actual size inside the test.

kainino0x (approved):

kainino0x:
Reviewed only the latest commit; it LGTM.

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: Yep.

Richard-Yunchao:
Can we merge this PR?

Kangz (approved):
Still LGTM, merging.

PR #481, opened 2021-03-05
Add texture usage validation test for createTexture

Richard-Yunchao:
-----

- [x] WebGPU readability
- [x] TypeScript readability


src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -19,11 +19,6 @@ TODO: review existing tests and merge with this plan:
 >     - x= every texture format
 > - 3d, {width, height, depthOrArrayLayers} > whatever the max is
 >     - x= every texture format
-> - usage flags
->     - {0, ... each single usage flag}
->     - x= every texture format
-> - every possible pair of usage flags
->     - with one common texture format
```
Richard-Yunchao, line 26: The tests is not exactly the same with this TODO. I added more tests for different dimension types (1D, 2D, 3D). And I tested pair of usage flags (combined usages) for every single formats, not only a few typical formats as the TODO said. I am not sure whether it is over-tested, though. But I remember Kai said that 100% full coverage is more important for CTS, and sometimes it might be a little redundant/duplicated.

Richard-Yunchao:
FYI @bbernhar and @RafaelCintron

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -289,19 +289,43 @@ g.test('it_is_invalid_to_submit_a_destroyed_texture_before_and_after_encode')
...
+    const descriptor = {
+      size,
+      dimension,
+      format,
+      usage,
+    };
+
+    let success = true;
+    if (!info.copySrc && (usage & GPUTextureUsage.COPY_SRC) !== 0) success = false;
+    if (!info.copyDst && (usage & GPUTextureUsage.COPY_DST) !== 0) success = false;
```
Richard-Yunchao, line 322: According to capability_info.ts, depth/stencil related formats can not be used as copySrc/copyDst. But we didn't validate this rule in Dawn.

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -289,19 +289,43 @@ g.test('it_is_invalid_to_submit_a_destroyed_texture_before_and_after_encode')
...
+  )
   .fn(async t => {
-    const format: GPUTextureFormat = t.params.format;
+    const { dimension, format, usage0, usage1 } = t.params;
     const info = kAllTextureFormatInfo[format];
 
     await t.selectDeviceOrSkipTestCase(info.extension);
 
-    const descriptor = t.getDescriptor({ width: 1, height: 1, format });
+    const size = dimension === '1d' ? [32, 1, 1] : dimension === '2d' ? [32, 32, 1] : [32, 32, 32];
```
Kangz, line 310: nit: the size could even be `[1, 1, 1]` unconditionally.

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -289,19 +289,43 @@ g.test('it_is_invalid_to_submit_a_destroyed_texture_before_and_after_encode')
     }, !_success);
   });
 
-g.test('it_is_invalid_to_have_an_output_attachment_texture_with_non_renderable_format')
-  .params(poptions('format', kAllTextureFormats))
+g.test('texture_usage')
+  .desc(
+    `Test texture usage (single usage or combined usages) for every texture formats and every dimension types`
```
Kangz, line 294: I believe there are also restrictions based on the sample count, can we add 1 and 4 sample counts?

src/webgpu/api/validation/createTexture.spec.ts
Kangz: We still don't have the specification for `createTexture` >_>. I believe it should be allowed to create the formats with the COPY flags because full T2T copies should be allowed.

Kangz (commented):
LGTM overall

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: Good point. I have tested multisample and `storage` usage conflict at sampleCount: https://github.com/gpuweb/cts/pull/478/files#diff-e0bc33c614ca9f7cdbbf7fdec58243d4af5681749e8d56739c95e970e521d15fR339. And the current tests are really a lot. So I didn't test that again. Otherwise, it will double the amount of test. Is this OK?

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: That's right. I just want to make the CTS more like cases in real world. Anyway, I suppose it's not important.

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: Yeah. We have submitted a lot issues and PRs for the spec of createTexture (and createView), but none of them got approval. I don't want to make a proposal and I'd like to leave it to other guys (maybe Google's Brandon?) because 1) you have discussed it a lot, and 2) Writing a large paragraph of spec is not my expertise because appropriate wording in spec is not easy for me but spec does need precise wording. I am glad to make some revisions if needed after the basic one is there.

Richard-Yunchao: > I believe it should be allowed to create the formats with the COPY flags because full T2T copies should be allowed.
Done. Changed the corresponding values in capability_info.ts.

Kangz: Interesting, we might need to do additional splitting of the capability because I believe this flag is used by some other tests to try copyB2T and copyT2B. CC @austinEng

src/webgpu/api/validation/createTexture.spec.ts
Kangz: I think the current generalization looks good. Do we need to add a sampleCount as well? I think the sampleCount != 1 disallows storage.

Kangz: It could be a different test too if that makes this one too big.

Kangz (approved):

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: Yeah. Here is the separate test for sampleCount: https://github.com/gpuweb/cts/pull/478. Those tests can cover sampleCount != 1 and usage includes STORAGE.

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: Hmm. I looked into current tests. Some tests don’t respect copySrc/copyDst in capability_info.ts at all, they just think depth/stencil formats are copyable but never check that information. For example:
depth_stencil_format,copy_usage_and_aspect in validation/encoding/cmds/buffer_texture_copies.spec.ts
and depth_stencil_copy_restrictions in validation/encoding/cmds/copyTextureToTexture.spec.ts. These tests can pass (say copyT2B for depth32float).

However, other tests do respect copySrc/copyDst in capability_info.js. For example, 
rowsPerImage_and_bytesPerRow in operation/command_buffer/image_copy.spec.ts and 
required_bytes_in_copy in validation/image_copy/layout_related.spec.ts. They filtered out sized depth/stencil before (because copySrc/Dst is false). With my change, it will include depth/stencil formats and test them, then those new tests fail ((say copyT2B for depth32float).

Which one is correct?

Richard-Yunchao:
@kainino0x , not sure whether you missed this PR. Please take a look when you have time but there is no hurry.

kainino0x:
Do you need a second reviewer? If not I'd like to save some time and rely on Corentin's review.

kainino0x:
Sorry, thought there were outstanding comments here. If not you can go ahead and squash-and-merge this yourself (pretty sure you should have access)

Richard-Yunchao:
> Do you need a second reviewer? If not I'd like to save some time and rely on Corentin's review.

I think you are the gatekeeper for CTS, :)     OK. I can leave it to Corentin, including the unresolved comment.

kainino0x:
Historically I am, but my time doesn't always scale :)

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -4,11 +4,6 @@ createTexture validation tests.
 TODO: review existing tests and merge with this plan:
 > All x= every texture format
 >
-> - usage flags
->     - {0, ... each single usage flag}
->     - x= every texture format
-> - every possible pair of usage flags
->     - with one common texture format
 > - any other conditions from the spec
 > - ...?
```
kainino0x, line 8: It looks like this TODO can be removed? or at least reduced simply to "TODO: check the spec for any other validation conditions that need to be tested here" (we should do this TODO very soon, since the file is getting fairly complete, so we can call the file "Done").

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -4,11 +4,6 @@ createTexture validation tests.
 TODO: review existing tests and merge with this plan:
```
kainino0x, line 4: I noticed not a lot of the old tests got replaced in your past few PRs on this file. It's okay to have duplicates but could you confirm there are at least no _obvious_ candidates to deduplicate?

Richard-Yunchao: The old tests are basically removed (or replaced?) because they are too simple. Currently I think there are no significant duplicates. You see, every test focus on one single parameter or a small field for all formats, like mipLevelCount, sampleCount, maxTextureSize, texture-usage, zero-size-not-allowed, etc. There might be small overlap but I am trying to avoid: for example, for the conflict between multisample support and STORAGE usage, It can be tested in sampleCount or texture usage. So I only test it in sampleCount related test, but don't test it again at texture usage.

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: How about to have a separate patch to clean createTexture? I noticed a few things like some comments, some change in capability_info.ts, and maybe a few more tests for corner cases.

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: I think copySrc/copyDst info for each depth/stencil format in capability_info.js should be true (just as what I revised in this PR). Because it should be OK to create depth/stencil texture with copy usages and T2T copy should be allowed. For CopyB2T or CopyT2B, there is another structure called kDepthStencilFormatCapabilityInBufferTextureCopy in capability_info.js (added by Jiawei recently), which will tell you whether is valid to copy all, depth-only or stencil-only aspect for CopyB2T, and whether it is valid to copy those aspects for CopyT2B. For the failures I pointed out above, there might be some bugs either in CTS or in Danw/Chromium implementation. I guess Jiawei (or Austin?) might fix them.

Kangz: Thanks for the investigation. It seems we need to fix a bunch of things. Maybe for this PR we should remove the changes to `capability_info` and unconditionally test with `COPY_SRC` and `COPY_DST` because now all the formats are copyable (`copySrc` and `copyDst` are true for all formats with this PR).

Someone can later figure out what we want to do for the copyability of texture formats, and how to split it per method. (for example it might be ok to `writeTexture` a format but not `copyBufferToTexture`).

Kangz (approved):

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: Done. Recovered changes in capability_info.js. One more place where we don't respect copySrc/copyDst for TextureFormatInfo. I left a comment and don't remove the code.

src/webgpu/api/validation/createTexture.spec.ts
kainino0x: Thanks for checking this!

src/webgpu/api/validation/createTexture.spec.ts
kainino0x: SGTM

PR #483, opened 2021-03-06
Validate compressed texture

Richard-Yunchao:
-----

- [x] WebGPU readability
- [x] TypeScript readability

### DIFF
--- a/src/webgpu/api/validation/createTexture.spec.ts
+++ b/src/webgpu/api/validation/createTexture.spec.ts
@@ -26,6 +26,8 @@ import { makeTestGroup } from '../../../common/framework/test_group.js';
 import {
   kAllTextureFormats,
   kAllTextureFormatInfo,
+  kCompressedTextureFormats,
+  kCompressedTextureFormatInfo,
   kUncompressedTextureFormats,
   kUncompressedTextureFormatInfo,
 } from '../../capability_info.js';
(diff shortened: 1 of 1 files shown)


Richard-Yunchao:
FYI @bbernhar and @RafaelCintron .

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -361,6 +460,92 @@ g.test('texture_size,3d_texture')
     }, !success);
   });
 
+g.test('texture_size,3d_texture,compressed_format')
```
Richard-Yunchao, line 463: Not sure whether this is needed. We say that compressed formats (BC formats currently) are not supported in 3D texture in Dawn project: https://dawn-review.googlesource.com/c/dawn/+/34922/1/src/tests/unittests/validation/TextureValidationTests.cpp#561

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -316,11 +331,95 @@ g.test('texture_size,2d_texture')
...
-  .desc(
-    `Test texture size requirement for 3D texture.
-	TODO: add tests for compressed texture.`
+g.test('texture_size,2d_texture,compressed_format')
+  .desc(`Test texture size requirement for 2D texture with compressed format.`)
+  .subcases(() =>
+    params()
+      .combine(poptions('format', kCompressedTextureFormats))
+      .combine(poptions('dimension', [undefined, '2d'] as const))
+      .combine([
```
Kangz, line 340: [1] this might need to become:

```
 .expand(p => {
     const {blockWidth, blockHeight} = kTextureFormatInfo[p.format];
     return poptions('size', [ same list that you have now ])
})
```

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -159,32 +165,44 @@ g.test('mipLevelCount,format')
...
+        { size: [31, 31, 32], dimension: '3d' as const }, // Mip level sizes: 31x31x32, 15x15x16, 7x7x8, 3x3x4, 1x1x2, 1x1x1
+        { size: [31, 31, 31], dimension: '3d' as const }, // Mip level sizes: 31x31x31, 15x15x15, 7x7x7, 3x3x3, 1x1x1
+        { size: [32, 8] }, // Mip levels: 32x8, 16x4, 8x2, 4x1, 2x1, 1x1
+        { size: [32, 32, 64] }, // Mip levels: 32x32x64, 16x16x64, 8x8x64, 4x4x64, 2x2x64, 1x1x64
+        { size: [32, 32, 64], dimension: '3d' as const }, // Mip levels: 32x32x64, 16x16x32, 8x8x16, 4x4x8, 2x2x4, 1x1x2, 1x1x1
+      ])
+      .unless(
+        ({ format, size, dimension }) =>
+          format === 'bc1-rgba-unorm' &&
+          (dimension === '1d' ||
```
Kangz, line 192: this might need to be dimension !== 2d. I don't think we allow 3D compressed textures.

src/webgpu/api/validation/createTexture.spec.ts
Kangz: I think it might not be needed because my understanding (based on your comment) is that WebGPU shouldn't enable 3D compressed textures at this point.

Kangz (approved):
LGTM

src/webgpu/api/validation/createTexture.spec.ts
kainino0x: +1, no 3d compressed textures at this point. I tried to investigate it but got confused by the D3D docs which seem to say BC texture blocks are 4x4x4 but I don't understand how that's possible. It's too niche and too complicated for V1.

However if you think it would be nice to keep these tests around for the future, feel free to leave them but with a `t.skip()` at the top and explanation of why.

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: Thanks. Looks like we can access parameters only in `unless`, `filter`, `expand` and something like that. But we can't access  previous parameters in another `.combine`.

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: OK. I use dimensions === 1d || dimension === 3d because the dimension is `undefined` for some tests, which are not 2d yet at this point.

Richard-Yunchao:
This PR should be ready for being merged now. Please take another look when you have time.

PR #485, opened 2021-03-09
Vertex state validation

Kangz:
PTAL! Each commit should be easy to review individually (except maybe the last one, but it just contains a bunch of new tests).

-----

- [X] Incomplete tests, if any, are marked with TODO or `.unimplemented()`.

- [x] WebGPU readability
- [x] TypeScript readability

### DIFF
--- a/src/webgpu/api/validation/vertex_state.spec.ts
+++ b/src/webgpu/api/validation/vertex_state.spec.ts
@@ -146,7 +146,7 @@ g.test('max_vertex_buffer_limit')
   .fn(t => {
     const { count, lastEmpty } = t.params;
 
-    const vertexBuffers: GPUVertexBufferLayoutDescriptor[] = [];
+    const vertexBuffers = [];
     for (let i = 0; i < count; i++) {
       if (lastEmpty || i !== count - 1) {
         vertexBuffers.push({ attributes: [], arrayStride: 0 });
@@ -154,7 +154,7 @@ g.test('max_vertex_buffer_limit')
         vertexBuffers.push({
           attributes: [{ format: 'float32', offset: 0, shaderLocation: 0 }],
           arrayStride: 0,
-        });
+        } as const);
       }
     }
 
@@ -176,7 +176,7 @@ g.test('max_vertex_attribute_limit')
   .fn(t => {
     const { attribCount, attribsPerBuffer } = t.params;
 
-    const vertexBuffers: GPUVertexBufferLayoutDescriptor[] = [];
+    const vertexBuffers = [];
 
     let attribsAdded = 0;
     while (attribsAdded !== attribCount) {
@@ -186,9 +186,9 @@ g.test('max_vertex_attribute_limit')
         targetCount = attribCount;
       }
 
-      const attributes: GPUVertexAttributeDescriptor[] = [];
+      const attributes = [];
       while (attribsAdded !== targetCount) {
-        attributes.push({ format: 'float32', offset: 0, shaderLocation: attribsAdded });
+        attributes.push({ format: 'float32', offset: 0, shaderLocation: attribsAdded } as const);
         attribsAdded++;
       }
 
@@ -222,7 +222,7 @@ g.test('max_vertex_buffer_array_stride_limit')
   .fn(t => {
     const { vertexBufferIndex, arrayStride } = t.params;
 
-    const vertexBuffers: GPUVertexBufferLayoutDescriptor[] = [];
+    const vertexBuffers = [];
     vertexBuffers[vertexBufferIndex] = { arrayStride, attributes: [] };
 
     const success = arrayStride <= kMaxVertexBufferArrayStride;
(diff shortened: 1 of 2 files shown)


src/webgpu/api/validation/vertex_state.spec.ts
```diff
@@ -488,183 +459,202 @@ g.test('vertex_shader_input_location_in_vertex_state')
...
+        const offsetsToTest = [0, bytesPerComponent, p.arrayStride - formatSize];
+
+        // Avoid adding duplicate offsets because the harness will assert each case is different.
+        const halfAlignment = Math.floor(bytesPerComponent / 2);
+        if (halfAlignment !== 0) {
+          offsetsToTest.push(halfAlignment);
+          offsetsToTest.push(p.arrayStride - formatSize - halfAlignment);
+        }
+
+        return poptions('offset', offsetsToTest);
```
kainino0x, line 540: nit: Could do something like
```ts
        return poptions(
          'offset',
          new Set([
            0,
            halfAlignment,
            bytesPerComponent,
            p.arrayStride - formatSize,
            p.arrayStride - formatSize - halfAlignment,
          ])
        );
```
I think.

OTOH, you could filter out formats with bytesPerComponent of 1. Aren't they kind of pointless for this test?

src/webgpu/api/validation/vertex_state.spec.ts
```diff
@@ -488,183 +459,202 @@ g.test('vertex_shader_input_location_in_vertex_state')
...
+    const formatInfo = kVertexFormatInfo[format];
+    const formatSize = formatInfo.bytesPerComponent * formatInfo.componentCount;
+    const limit = arrayStride === 0 ? kMaxVertexBufferArrayStride : arrayStride;
+
+    const success = offset + formatSize <= limit;
+    t.testVertexState(success, { vertexBuffers });
+  });
+
+g.test('many_attributes_overlapping')
+  .desc(`Test that it is valid to have many vertex attributes overlap`)
```
kainino0x, line 649: nit: Maybe test with different formats (like rotate the `format` between the attributes)?

Or maybe this should be an operation test later that actually tests the load behavior? (or both validation and operation test)

src/webgpu/api/validation/vertex_state.spec.ts
```diff
@@ -488,183 +459,202 @@ g.test('vertex_shader_input_location_in_vertex_state')
...
+    Test that vertex attribute [offset, offset + formatSize) must be contained in the arrayStride if arrayStride is not 0:
+    - Test for all formats.
+    - Test for various arrayStrides and offsets within that stride
+    - Test for various vertex buffer indices
+    - Test for various amounts of attributes in that vertex buffer`
+  )
+  .cases(
+    params()
+      .combine(poptions('format', kVertexFormats))
+      .combine(
```
kainino0x, line 586: nit: Probably arrayStride and offset (at least offset) make more sense as subcases

src/webgpu/api/validation/vertex_state.spec.ts
```diff
@@ -488,183 +459,202 @@ g.test('vertex_shader_input_location_in_vertex_state')
...
+        ])
+      )
+      .expand(p => {
+        // Compute a bunch of test offsets to test.
+        const { bytesPerComponent, componentCount } = kVertexFormatInfo[p.format];
+        const formatSize = bytesPerComponent * componentCount;
+        const offsetsToTest = [0, bytesPerComponent];
+
+        // Avoid adding offset based on the arrayStride if it would make duplicated tests at it
+        // makes the harness assert.
```
kainino0x, line 601: I don't think this comment is applicable here?

kainino0x (approved):

src/webgpu/api/validation/vertex_state.spec.ts
Kangz: Went with the Set approach.

src/webgpu/api/validation/vertex_state.spec.ts
Kangz: There should be an operation test that does complicated cases like this, but I still added the format rotation.

src/webgpu/api/validation/vertex_state.spec.ts
kainino0x: It's a little weird but it can be done, see the commit I pushed

kainino0x: (I have ideas about how to improve this - subcases could be passed a pre-constructed single-value params() object so you can just `p => p.combine`, but haven't had a chance to try this)

kainino0x (approved):
nit: prefer to push new commits instead of force-pushing, to make re-review easier (though it was ok in this case because there was no rebase / the "force-push" diff on github was clean)

Kangz:
Yeah I should avoid force push. What do you think of my adding small commits like "fix for X, fix for Foo" then I'll do a rebase -i before and force-push before merging?

src/webgpu/api/validation/vertex_state.spec.ts
Kangz: Thanks for the fix!

kainino0x:
I use squash-and-merge for almost all PRs, but if you want multiple commits in a PR that sounds good, just let me know so I won't land it.

PR #489, opened 2021-03-09
Split createEncoder finish into three parts for strict validation

kainino0x:
Validation tests should specifically expect validation to occur during
finish(), and not during endPass(). This allows the tests to check this
strictly.

NOT READY YET

-----

- [ ] WebGPU readability
- [ ] TypeScript readability

### DIFF
--- a/src/webgpu/api/validation/encoding/cmds/compute_pass.spec.ts
+++ b/src/webgpu/api/validation/encoding/cmds/compute_pass.spec.ts
@@ -54,10 +54,15 @@ setPipeline should generate an error iff using an 'invalid' pipeline.
   .params(poptions('state', ['valid', 'invalid'] as const))
   .fn(t => {
     const pipeline = t.createComputePipeline(t.params.state);
-    const { encoder, finish } = t.createEncoder('compute pass');
-    encoder.setPipeline(pipeline);
+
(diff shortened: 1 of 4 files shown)


kainino0x:
Good point on being unable to tell what caused the error. Have you considered, instead of exposing more details to the test, just moving the expectValidationError checks to `finish`? and/or creating another expectValidationError-like function that understands CommandBufferMaker? Otherwise, it seems like each test that uses createEncoder would gain extra boilerplate.

Regarding `api/validation/queue/destroyed/*`, I'm not a huge fan of having similar test logic in 2 places. Maybe we can add a method to `ValidationTest` that conditionally checks either finish or submit depending on a state?
```ts
  expectValidationErrorWithState(
    state: 'valid' | 'invalid' | 'destroyed',
    finish: () => Iterable<GPUCommandBuffer>,
    shouldError: boolean = true
  ): void {
    if (state === 'destroyed') {
      const commandBuffers = finish();
        this.queue.submit(commandBuffers);
      }, shouldError);
    } else {
        finish();
      }, shouldError);
    }
  }
```

_Originally posted by @egalli in https://github.com/gpuweb/cts/pull/493#discussion_r592828542_

PR #491, opened 2021-03-10
Add validation tests plan for Occlusion Query

haoxli:
This is tests plan for Occlusion Query validation on:
- {all types, invalid, undefined, destoryed} query set
- query index {out of bounds, write twice}
- begin/end balance
- nesting

-----

- [ ] WebGPU readability
- [ ] TypeScript readability

### DIFF
--- a/src/webgpu/api/validation/encoding/beginRenderPass.spec.ts
+++ b/src/webgpu/api/validation/encoding/beginRenderPass.spec.ts
@@ -1,4 +1,6 @@
 export const description = `
+Note: render pass 'occlusionQuerySet' validation is tested in queries/general.spec.ts
+
 TODO: check for duplication (render_pass/, etc.), plan, and implement.
 Note possibly a lot of this should be operation tests instead.
 Notes:
--- a/src/webgpu/api/validation/encoding/beginRenderPass.spec.ts
+++ b/src/webgpu/api/validation/encoding/beginRenderPass.spec.ts
@@ -17,8 +17,6 @@ Notes:
 >     - {all possible {depth, stencil} load ops, load values {in range, negative, too large}}
 >     - all possible {depth, stencil} store ops
 >     - depthReadOnly {t,f}, stencilReadOnly {t,f}
-> - occlusion query set
->     - tested in queries/general.spec.ts
 `;
 
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
--- a/src/webgpu/api/validation/encoding/beginRenderPass.spec.ts
+++ b/src/webgpu/api/validation/encoding/beginRenderPass.spec.ts
@@ -15,6 +15,8 @@ Notes:
 >     - {all possible {depth, stencil} load ops, load values {in range, negative, too large}}
 >     - all possible {depth, stencil} store ops
 >     - depthReadOnly {t,f}, stencilReadOnly {t,f}
+> - occlusion query set
+>     - tested in queries/general.spec.ts
 `;
 
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
--- a/src/webgpu/api/validation/encoding/queries/begin_end.spec.ts
+++ b/src/webgpu/api/validation/encoding/queries/begin_end.spec.ts
@@ -24,11 +24,7 @@ g.test('occlusion_query,begin_end_balance')
   .desc(
     `
 Tests that begin/end occlusion queries mismatch on render pass:
-- begin 0, end 1 (begin no queries, then end one query)
-- begin 1, end 0 (begin one query, then end no queries)
-- begin 1, end 1 (begin one query, then end one query)(control case)
-- begin 1, end 2 (begin one query, then end two queries)
-- begin 2, end 1 (begin two queries, then end one query)
+- begin n queries, then end m queries, for various n and m.
   `
   )
   .subcases(
@@ -36,7 +32,7 @@ Tests that begin/end occlusion queries mismatch on render pass:
       [
         { begin: 0, end: 1 },
         { begin: 1, end: 0 },
-        { begin: 1, end: 1 },
+        { begin: 1, end: 1 }, // control case
         { begin: 1, end: 2 },
         { begin: 2, end: 1 },
       ] as const
@@ -47,11 +43,19 @@ g.test('occlusion_query,begin_end_invalid_nesting')
   .desc(
     `
 Tests the invalid nesting of begin/end occlusion queries:
+- begin index 0, end, begin index 0, end (control case)
 - begin index 0, begin index 0, end, end
 - begin index 0, begin index 1, end, end
   `
   )
-  .subcases(() => [{ calls: [0, 0, 'end', 'end'] }, { calls: [0, 1, 'end', 'end'] }] as const)
+  .subcases(
+    () =>
+      [
+        { calls: [0, 'end', 1, 'end'] }, // control case
+        { calls: [0, 0, 'end', 'end'] },
+        { calls: [0, 1, 'end', 'end'] },
+      ] as const
+  )
   .unimplemented();
 
 g.test('occlusion_query,disjoint_queries_with_same_query_index')
@@ -69,16 +73,7 @@ g.test('nesting')
   .desc(
     `
 Tests that whether it's allowed to nest various types of queries:
-- beginOcclusionQuery, writeTimestamp, endOcclusionQuery
-- beginOcclusionQuery, beginOcclusionQuery, endOcclusionQuery, endOcclusionQuery
-- beginOcclusionQuery, beginPipelineStatisticsQuery, endPipelineStatisticsQuery, endOcclusionQuery
-- beginOcclusionQuery, beginPipelineStatisticsQuery, endOcclusionQuery, endPipelineStatisticsQuery
-- beginPipelineStatisticsQuery, writeTimestamp, endPipelineStatisticsQuery
-- beginPipelineStatisticsQuery, beginPipelineStatisticsQuery, endPipelineStatisticsQuery, endPipelineStatisticsQuery
-- beginPipelineStatisticsQuery, beginOcclusionQuery, endOcclusionQuery, endPipelineStatisticsQuery
-- beginPipelineStatisticsQuery, beginOcclusionQuery, endPipelineStatisticsQuery, endOcclusionQuery
-- writeTimestamp, beginOcclusionQuery, writeTimestamp, endOcclusionQuery
-- writeTimestamp, beginPipelineStatisticsQuery, writeTimestamp, endPipelineStatisticsQuery
+- call {occlusion, pipeline-statistics, timestamp} query in same type or other type.
   `
   )
   .subcases(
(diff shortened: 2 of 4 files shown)


src/webgpu/api/validation/encoding/beginRenderPass.spec.ts
```diff
@@ -15,6 +15,8 @@ Notes:
 >     - {all possible {depth, stencil} load ops, load values {in range, negative, too large}}
 >     - all possible {depth, stencil} store ops
 >     - depthReadOnly {t,f}, stencilReadOnly {t,f}
+> - occlusion query set
+>     - tested in queries/general.spec.ts
```
kainino0x, line 19: This whole notes section is part of the TODO. This should go outside of the TODO (it can go above) like "Note: render pass 'occlusionQuerySet' validation is tested in queries/general.spec.ts"

src/webgpu/api/validation/encoding/queries/begin_end.spec.ts
```diff
@@ -9,15 +9,50 @@ TODO:
...
 import { makeTestGroup } from '../../../../../common/framework/test_group.js';
 import { ValidationTest } from '../../validation_test.js';
 
 export const g = makeTestGroup(ValidationTest);
+
+g.test('occlusion_query,begin_end')
+  .desc(
+    `
+Tests that begin/end occlusion query on render pass:
+- begin 0
```
kainino0x, line 26: For this description we should be a bit more explicit, e.g. "begin index 0, begin index 1, end" or something.

(Note, in the notes above, "begin 0, end 1" means "begin no queries and end one query". That's unclear, so best to be clear here.)

src/webgpu/api/validation/encoding/queries/begin_end.spec.ts
```diff
@@ -9,15 +9,50 @@ TODO:
...
 
 import { makeTestGroup } from '../../../../../common/framework/test_group.js';
 import { ValidationTest } from '../../validation_test.js';
 
 export const g = makeTestGroup(ValidationTest);
+
+g.test('occlusion_query,begin_end')
+  .desc(
+    `
+Tests that begin/end occlusion query on render pass:
```
kainino0x, line 25: More explicitly describe which validation this is testing.

src/webgpu/api/validation/encoding/queries/begin_end.spec.ts
```diff
@@ -9,15 +9,50 @@ TODO:
...
+- begin 0
+- begin 0, end (control case)
+- begin 0, begin 1, end
+- begin 0, end, end
```
kainino0x, line 29: Nested occlusion query (beginOcclusionQuery, beginOcclusionQuery, endOcclusionQuery, endOcclusionQuery) should be tested either here or in the 'nesting' test below. A separate test is probably ideal, since it tests a different rule, but you can share some code since it will be very simple to describe. In the notation of this test, you could test both:

- begin index 0, begin index 0, end, end
- begin index 0, begin index 1, end, end

src/webgpu/api/validation/encoding/queries/begin_end.spec.ts
```diff
@@ -9,15 +9,50 @@ TODO:
...
+- end
+  `
+  )
+  .unimplemented();
+
+g.test('occlusion_query,begin_with_same_query_index')
+  .desc(
+    `
+Tests that begin occlusion query with same query index twice:
```
kainino0x, line 38: Does this mean (begin, end, begin, end)? If so, let's be clearer about it ("Test two disjoint occlusion queries in the same command buffer")

src/webgpu/api/validation/encoding/queries/begin_end.spec.ts
```diff
@@ -9,15 +9,50 @@ TODO:
...
+- on {same, different (control case)} render pass
```
kainino0x, line 39: Is one of these cases invalid? (Please specify)

src/webgpu/api/validation/encoding/queries/begin_end.spec.ts
```diff
@@ -9,15 +9,50 @@ TODO:
...
+  .desc(
+    `
+Tests that begin/end occlusion query on render pass:
+- begin 0
+- begin 0, end (control case)
+- begin 0, begin 1, end
+- begin 0, end, end
+- end
+  `
+  )
```
kainino0x, line 32: Consider sketching out the `.cases`/`.subcases` of this test as part of this test plan. That can replace the list in the description and would help understand how the test is intended to work.

Perhaps something like
```ts
[
  { _valid: false, calls: [0] },
  { _valid: true, calls: [0, 'end'] },
  { _valid: false, calls: [0, 1, 'end'] },
  { _valid: false, calls: [0, 'end', 'end'] },
  { _valid: false, calls: ['end'] },
] as const
```

src/webgpu/api/validation/encoding/queries/general.spec.ts
```diff
@@ -27,6 +27,26 @@ class F extends ValidationTest {
 
 export const g = makeTestGroup(F);
 
+g.test('occlusion_query,invalid_query_set')
+  .desc(
+    `
+Tests that set occlusion query set with all types in render pass descriptor:
+- type {occlusion (control case), pipeline statistics, timestamp}
+- invalid query set that failed during creation
```
kainino0x, line 35: This case would probably be easier as a separate test which has just 2 cases (valid and invalid).

src/webgpu/api/validation/queue/destroyed/query_set.spec.ts
```diff
@@ -21,6 +21,15 @@ export const enum EncoderType {
 
 export const g = makeTestGroup(ValidationTest);
 
+g.test('beginRenderPass')
+  .desc(
+    `
+Tests that use a destroyed query set in beginRenderPass for occlusion query.
```
kainino0x, line 27: Should note here that the render pass can be empty, no begin/end query calls are needed to test this.

src/webgpu/api/validation/encoding/queries/begin_end.spec.ts
```diff
@@ -9,15 +9,50 @@ TODO:
...
+Tests that begin occlusion query with same query index twice:
+- on {same, different (control case)} render pass
+  `
+  )
+  .unimplemented();
+
+g.test('nesting')
+  .desc(
+    `
+Tests that whether it's allowed to nest various types of queries:
```
kainino0x, line 47: Either test occlusion-in-occlusion/pipelinestatistics-in-pipelinestatistics here, or add a note to the description saying they're tested elsewhere. IMO should just test them here, it's okay if there is duplicated test coverage.

kainino0x, line 47: Maybe we should test cases like
beginOcclusionQuery, beginPipelineStatisticsQuery, endOcclusionQuery, endPipelineStatisticsQuery
just in case

src/webgpu/api/validation/queue/destroyed/query_set.spec.ts
kainino0x: I see. That's a little surprising but I guess it makes sense. Would you mind filing a spec issue so editors can make sure that's intended behavior?

kainino0x: Sorry, I realized this is probably in https://github.com/gpuweb/gpuweb/pull/1494 which I haven't gotten a chance to  review yet.

src/webgpu/api/validation/encoding/queries/begin_end.spec.ts
```diff
@@ -14,45 +14,91 @@ TODO:
...
+
+g.test('occlusion_query,begin_end_invalid_nesting')
+  .desc(
+    `
+Tests the invalid nesting of begin/end occlusion queries:
+- begin index 0, begin index 0, end, end
+- begin index 0, begin index 1, end, end
+  `
+  )
+  .subcases(() => [{ calls: [0, 0, 'end', 'end'] }, { calls: [0, 1, 'end', 'end'] }] as const)
```
kainino0x, line 54: If neither of these is valid (I'm not sure), there should be some control cases here like `[0, 'end']`, `[0, 'end', 1, 'end']` which would be valid.

src/webgpu/api/validation/encoding/queries/begin_end.spec.ts
```diff
@@ -14,45 +14,91 @@ TODO:
...
-- begin 0, end (control case)
-- begin 0, begin 1, end
-- begin 0, end, end
-- end
+Tests that begin/end occlusion queries mismatch on render pass:
+- begin 0, end 1 (begin no queries, then end one query)
+- begin 1, end 0 (begin one query, then end no queries)
+- begin 1, end 1 (begin one query, then end one query)(control case)
+- begin 1, end 2 (begin one query, then end two queries)
+- begin 2, end 1 (begin two queries, then end one query)
```
kainino0x, line 31: It's okay to omit the "gory details" in this description, so we don't have to keep it perfectly in sync with the case list. Instead can just say "begin n queries, then end m queries, for various n and m".

src/webgpu/api/validation/encoding/beginRenderPass.spec.ts
```diff
@@ -15,6 +17,8 @@ Notes:
...
```
kainino0x, line 21:
```suggestion
```

src/webgpu/api/validation/encoding/queries/begin_end.spec.ts
```diff
@@ -14,45 +14,91 @@ TODO:
...
 - beginOcclusionQuery, writeTimestamp, endOcclusionQuery
+- beginOcclusionQuery, beginOcclusionQuery, endOcclusionQuery, endOcclusionQuery
 - beginOcclusionQuery, beginPipelineStatisticsQuery, endPipelineStatisticsQuery, endOcclusionQuery
 - beginOcclusionQuery, beginPipelineStatisticsQuery, endOcclusionQuery, endPipelineStatisticsQuery
 - beginPipelineStatisticsQuery, writeTimestamp, endPipelineStatisticsQuery
+- beginPipelineStatisticsQuery, beginPipelineStatisticsQuery, endPipelineStatisticsQuery, endPipelineStatisticsQuery
 - beginPipelineStatisticsQuery, beginOcclusionQuery, endOcclusionQuery, endPipelineStatisticsQuery
 - beginPipelineStatisticsQuery, beginOcclusionQuery, endPipelineStatisticsQuery, endOcclusionQuery
 - writeTimestamp, beginOcclusionQuery, writeTimestamp, endOcclusionQuery
 - writeTimestamp, beginPipelineStatisticsQuery, writeTimestamp, endPipelineStatisticsQuery
```
kainino0x, line 81: Same here, don't need the gory details in the description.

src/webgpu/api/validation/encoding/queries/begin_end.spec.ts
```diff
@@ -14,45 +14,91 @@ TODO:
...
   `
   )
+  .subcases(
+    () =>
+      [
+        { begin: 'occlusion', nest: 'timestamp', end: 'occlusion', _valid: true },
+        { begin: 'occlusion', nest: 'occlusion', end: 'occlusion', _valid: false },
+        { begin: 'occlusion', nest: 'pipeline-statistcs', end: 'occlusion', _valid: true },
```
kainino0x, line 89: nit: statistcs -> statistics

src/webgpu/api/validation/queue/destroyed/query_set.spec.ts
kainino0x: Is this actually covered by that PR? I couldn't find it. If it's not, could you make a PR or open an issue on the spec?

kainino0x (approved):

src/webgpu/api/validation/queue/destroyed/query_set.spec.ts
haoxli: I made a separate PR for it (https://github.com/gpuweb/gpuweb/pull/1531) which also contains the query set state at its creation and destory.

kainino0x (approved):

kainino0x:
@haoxli BTW when you're working on these tests, make sure any obsolete TODOs are removed, I think there might be a few that are duplicated with the new plans you just added. Thanks!

haoxli:
Yes, there are some TODOs in other files under queries/, they may can be removed. And some TODOs are partially completed, which need update. I can make a PR to clean up them.

Issue #391, opened 2020-12-02
Remove descriptions from listing.js?


kainino0x:
When listing.js is generated, it reads the description from each .spec.ts file. This was done so that the file list with descriptions could be loaded without loading the .spec.ts files.

I think this is not really a necessary optimization. Saving /standalone/ from having to load every .spec.ts file in the suite is probably not that worth having duplicated info in the build.

This would fix a problem where the description can get out of date in generated listing.js both in builds and in dev_server.

_Originally posted by @kainino0x in https://github.com/gpuweb/cts/pull/386#discussion_r533625136_

PR #492, opened 2021-03-10
Break gen_listings dep on ts-node / fast-glob. Add src_dir/out_dir args

austinEng:
Previous to this patch, it was necessary to run gen_listings from the
root of the source tree using ts-node. These requirements make it
necessary for listing generation to take extra dependencies which may
be problematic in environments where the set of dependencies are vendored
and downloading many dependencies with npm is undesirable. Even with no
dependencies, running from the root of the source tree means the
transpiled .js file can't be run directly.

This patch adds a tsconfig for transpiling tools/ for usage directly
with Node.js, and a build step in the grunt presubmit check for it.

-----

- [[x]](url) WebGPU readability
- [ ] TypeScript readability

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -8,7 +8,7 @@ module.exports = function (grunt) {
     pkg: grunt.file.readJSON('package.json'),
 
     clean: {
-      out: ['out/', 'out-wpt/'],
+      out: ['out/', 'out-wpt/', 'out-node/'],
     },
 
     run: {
@@ -18,7 +18,7 @@ module.exports = function (grunt) {
       },
       'generate-listings': {
         cmd: 'node',
-        args: ['tools/gen_listings', 'webgpu', 'unittests', 'demo'],
+        args: ['tools/gen_listings', 'out/', 'src/', 'webgpu', 'unittests', 'demo'],
       },
       'generate-wpt-cts-html': {
         cmd: 'node',
@@ -54,6 +54,16 @@ module.exports = function (grunt) {
           // These files will be generated, instead of compiled from TypeScript.
           '--ignore=src/common/framework/version.ts',
           '--ignore=src/webgpu/listing.ts',
+        ],
+      },
+      'build-out-node': {
+        cmd: 'node',
+        args: [
+          'node_modules/typescript/lib/tsc.js',
+          '--project', 'node.tsconfig.json',
+          '--outDir', 'out-node/',
+          '--noEmit', 'false',
+          '--declaration', 'false'
         ],
       },
       lint: {
@@ -146,6 +156,7 @@ module.exports = function (grunt) {
     'clean',
     'build-standalone',
     'build-wpt',
+    'run:build-out-node',
     'build-done-message',
     'ts:check',
     'run:unittest',
--- a/node.tsconfig.json
+++ b/node.tsconfig.json
@@ -1,0 +1,13 @@
+// Typescript configuration for compile sources and
+// dependent files for usage directly with Node.js. This
+// is useful for running scripts in tools/ directly with Node
+// without including extra dependencies.
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "module": "commonjs"
+  },
+
+  // Can't be compiled for Node.js as it uses import.metal.url.
+  "exclude": ["src/common/runtime/**/*.ts"]
+}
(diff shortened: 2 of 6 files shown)


Gruntfile.js
```diff
@@ -56,6 +56,16 @@ module.exports = function (grunt) {
           '--ignore=src/webgpu/listing.ts',
         ],
       },
+      'build-out-node': {
```
austinEng, line 59: using `tsc` instead of `babel`, so we don't have the generated comment at the top of the file. Is that okay? If Chromium doesn't want a dep on babel, we would lose the generated comment in /gen/..wpt/ too. But, it would be in /gen so I think it should be fairly clear.

austinEng:
Nevermind.. this effort might not be worth it. `tsc` doesn't work in a `--no-check` mode so we still have to download a few `node_modules` in order to have the proper types for the transpilation. That could mean we may as well vendor our own deps.

austinEng:
Hm maybe it's fine though. `tsc` still emits the JS even though it fails a lot of the type checks (I think).

Gruntfile.js
kainino0x: I think that's fine since we aren't checking in the generated code.

node.tsconfig.json
```diff
@@ -0,0 +1,13 @@
...
+// dependent files for usage directly with Node.js. This
+// is useful for running scripts in tools/ directly with Node
+// without including extra dependencies.
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "module": "commonjs"
+  },
+
+  // Can't be compiled for Node.js as it uses import.metal.url.
```
kainino0x, line 11:
```suggestion
  // Can't be compiled for Node.js as it uses import.meta.url.

```

src/common/tools/gen_listings.ts
```diff
@@ -6,27 +6,26 @@ import { crawl } from './crawl.js';
 
 function usage(rc: number): void {
   console.error('Usage:');
-  console.error('  tools/gen_listings [SUITES...]');
-  console.error('  tools/gen_listings webgpu unittests');
+  console.error('  tools/gen_listings [OUT_DIR] [SRC_DIR] [SUITES...]');
+  console.error('  tools/gen_listings out/ src/ webgpu unittests');
```
kainino0x, line 10: nit: Could probably just be
```suggestion
  console.error('  tools/gen_listings [OUT_DIR] [PATHS_TO_SUITES...]');
  console.error('  tools/gen_listings out/ src/webgpu src/unittests');
```

src/common/tools/crawl.ts
```diff
@@ -45,26 +75,25 @@ export async function crawl(suite: string): Promise<TestSuiteListingEntry[]> {
 
       mod.g.validate();
```
kainino0x, line 76: followup?: I think it may be possible to skip importing the module entirely when using `crawl` for runtime generation. We can skip validate(), and with #391 we don't need the description either. I suspect this would make it a lot faster.

kainino0x (approved):

src/common/tools/crawl.ts
austinEng: Oh it would be very, very nice. Right now we need to compile almost all the sources because crawl does the import. But if the listing just cares about file names, then we don't need to do real imports. And we can get away with a Node.js build just for stuff in tools/ (maybe some other utilities too).

src/common/tools/gen_listings.ts
austinEng: Done.

src/common/tools/crawl.ts
austinEng: did this now.

austinEng:
@kainino0x  I cherry-picked #499 into this PR since they were both updating the `gen_listing` args. It now has some very basic arg parsing.

src/common/tools/gen_listings.ts
```diff
@@ -5,25 +5,46 @@ import * as process from 'process';
...
+Options:
+  --help          Print this message and exit.
+  --no-validate   Whether to validate test modules while crawling.
+`);
   process.exit(rc);
 }
 
-if (process.argv.length <= 4) {
+const argv = process.argv;
+if (argv.indexOf('--help') !== -1) {
```
kainino0x, line 22: FYI we have one other utility with command line flags: `runtime/cmdline.ts`. This looks fine though.

kainino0x (approved):
LGTM

PR #486, opened 2021-03-09
Deduplicate texture table header

kainino0x:
Not sure why I duplicated it in the first place.

-----

- [x] New helpers, if any, are documented in `helper_index.md`.
- [x] Incomplete tests, if any, are marked with TODO or `.unimplemented()`.

**[Review requirement](https://github.com/gpuweb/cts/blob/main/docs/reviews.md) checklist:**

- [x] WebGPU readability
- [x] TypeScript readability

PR #494, opened 2021-03-11
Clean code for createTexture validation tests

Richard-Yunchao:
It includes:
  - move destroy related tests in createTexture.spec.ts to texture/destroy.spec.ts
  - remove duplicated tests in texture/destroy.spec.ts
  - remove unneeded comments in createTexture.spec.ts
  - add a few more tests about texture size in createTexture.spec.ts
  - add 'undefined' as a dimension type in tests
  - move kTexFmtInfoHeader to the top in capability_info.js
  - set and use default values in TextureFormatInfo if we can in capability_info.js

-----

- [ ] WebGPU readability
- [ ] TypeScript readability

### DIFF
--- a/src/webgpu/api/validation/createTexture.spec.ts
+++ b/src/webgpu/api/validation/createTexture.spec.ts
@@ -1,14 +1,4 @@
-export const description = `
-createTexture validation tests.
-
-TODO: review existing tests and merge with this plan:
-> All x= every texture format
->
-> - any other conditions from the spec
-> - ...?
-
-TODO: move destroy tests out of this file
-`;
+export const description = `createTexture validation tests.`;
 
 import { poptions, params } from '../../../common/framework/params_builder.js';
 import { makeTestGroup } from '../../../common/framework/test_group.js';
@@ -66,10 +56,9 @@ g.test('zero_size')
     width, height, depthOrArrayLayers and mipLevelCount for every dimension, and representative formats.
     TODO: add tests for depth/stencil format if depth/stencil format can support mipmaps.`
   )
-  // .cases(poptions('dimension', ['1d', '2d', '3d'] as const))
   .subcases(() =>
     params()
-      .combine(poptions('dimension', ['1d', '2d', '3d'] as const))
+      .combine(poptions('dimension', [undefined, ...kTextureDimensions]))
       .combine(
         poptions('zeroArgument', [
           'none',
@@ -80,7 +69,10 @@ g.test('zero_size')
         ] as const)
       )
       .combine(poptions('format', ['rgba8unorm', 'rgb10a2unorm', 'bc1-rgba-unorm'] as const))
-      .unless(({ format, dimension }) => format === 'bc1-rgba-unorm' && dimension !== '2d')
+      .unless(
+        ({ format, dimension }) =>
+          format === 'bc1-rgba-unorm' && dimension !== '2d' && dimension !== undefined
+      )
   )
   .fn(async t => {
     const { dimension, zeroArgument, format } = t.params;
@@ -251,14 +243,17 @@ g.test('sampleCount,valid_sampleCount_with_other_parameter_varies')
   .desc(
     `Test texture creation with valid sample count when dimensions, arrayLayerCount, mipLevelCount, format, and usage varies.
      Texture can be single sample (sampleCount is 1) or multi-sample (sampleCount is 4).
-     Multisample texture requires that 1) its dimension is 2d, 2) its format is a uncompressed format, 3) its mipLevelCount and arrayLayerCount are 1, 4) its usage doesn't include STORAGE.`
+     Multisample texture requires that 1) its dimension is 2d or undefined, 2) its format is a uncompressed format, 3) its mipLevelCount and arrayLayerCount are 1, 4) its usage doesn't include STORAGE.`
   )
   .subcases(() =>
     params()
       .combine(poptions('sampleCount', [1, 4]))
-      .combine(poptions('dimension', ['1d', '2d', '3d'] as const))
+      .combine(poptions('dimension', [undefined, ...kTextureDimensions]))
       .combine(poptions('arrayLayerCount', [1, 2]))
-      .unless(({ arrayLayerCount, dimension }) => arrayLayerCount === 2 && dimension !== '2d')
+      .unless(
+        ({ arrayLayerCount, dimension }) =>
+          arrayLayerCount === 2 && dimension !== '2d' && dimension !== undefined
+      )
       .combine(poptions('mipLevelCount', [1, 2]))
       .combine(poptions('format', kAllTextureFormats))
       .combine(poptions('usage', kTextureUsages))
@@ -278,7 +273,7 @@ g.test('sampleCount,valid_sampleCount_with_other_parameter_varies')
     const size =
       dimension === '1d'
         ? [32, 1, 1]
-        : dimension === '2d'
+        : dimension === '2d' || dimension === undefined
         ? [32, 32, arrayLayerCount]
         : [32, 32, 32];
     const descriptor = {
@@ -293,7 +288,7 @@ g.test('sampleCount,valid_sampleCount_with_other_parameter_varies')
     const success =
       sampleCount === 1 ||
       (sampleCount === 4 &&
-        dimension === '2d' &&
+        (dimension === '2d' || dimension === undefined) &&
         kAllTextureFormatInfo[format].multisample &&
         mipLevelCount === 1 &&
         arrayLayerCount === 1 &&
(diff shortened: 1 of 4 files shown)


Richard-Yunchao:
As I promised in another PR, here comes the code clean PR for createTexture. PTAL.

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -28,48 +18,17 @@ import { maxMipLevelCount } from '../../util/texture/base.js';
...
-export const g = makeTestGroup(F);
+export const g = makeTestGroup(ValidationTest);
 
 g.test('zero_size')
   .desc(
     `Test texture creation with zero or nonzero size of
     width, height, depthOrArrayLayers and mipLevelCount for every dimension, and representative formats.
     TODO: add tests for depth/stencil format if depth/stencil format can support mipmaps.`
   )
-  // .cases(poptions('dimension', ['1d', '2d', '3d'] as const))
```
Richard-Yunchao, line 69: Do you think that it is good to make 'dimension' as a parameter in `.cases`, while keep other parameters in `.subcases`? Because 2d or undefined textures are the most important one. 3d textures are not that important, and 1d texture is not important at all (look like we may not support 1d texture in MVP, and GLES 3.2 removed 1d texture). 

If we move `dimension` to `.cases`, then it is easier to see the test results of different dimension types. Then the most important one (2d / undefined, or maybe 3d?) will not mixed with others. While failures in other dimension types are not important. So it is convenient to see the top-priority failures. 

WDYT?

src/webgpu/capability_info.ts
```diff
@@ -70,8 +70,8 @@ export const kBufferUsages = numericKeysOf<GPUBufferUsage>(kBufferUsageInfo);
 
 // Textures
 
-export const kRegularTextureFormatInfo = /* prettier-ignore */ makeTable(
-                           ['renderable', 'multisample', 'color', 'depth', 'stencil', 'storage', 'copySrc', 'copyDst', 'bytesPerBlock', 'blockWidth', 'blockHeight',              'extension'] as const,
+const kTexFmtInfoHeader =  ['renderable', 'multisample', 'color', 'depth', 'stencil', 'storage', 'copySrc', 'copyDst', 'bytesPerBlock', 'blockWidth', 'blockHeight',              'extension'] as const;
```
Richard-Yunchao, line 73: I think it is better to move the template of kTexFmtInfoHeader to the top. But `grunt pre` fail because I don't want to run `grunt fix`, the latter will re-layout and make all parameters of kTexFmtInfoHeader on a separate line. I run `grunt pre --force`. It can pass.

Richard-Yunchao:
FYI, @bbernhar and @RafaelCintron

src/webgpu/api/validation/texture/destroy.spec.ts
```diff
@@ -3,17 +3,61 @@ Destroying a texture more than once is allowed.
...
-    format: 'r8unorm',
-    usage: GPUTextureUsage.SAMPLED,
-  });
+g.test('it_is_valid_to_destroy_a_texture').fn(t => {
+  const texture = t.device.createTexture(descriptor);
+  texture.destroy();
+});
 
-  tex.destroy();
-  tex.destroy();
```
Richard-Yunchao, line 18: Exactly duplicated with this test: it is valid to destroy a destroyed texture.

src/webgpu/capability_info.ts
kainino0x: I actually just considered but then didn't do this in #486. Reasoning was that having the header multiple times makes it easier to read. But I am also totally fine with this change; whatever you think.

The lint fix is to use `/* prettier-ignore */` on the previous line, see [1]

src/webgpu/capability_info.ts
```diff
@@ -117,20 +117,19 @@ export const kRegularTextureFormatInfo = /* prettier-ignore */ makeTable(
   'rgba32float':           [        true,              ,        ,        ,          ,      true,          ,          ,              16],
 } as const);
 /* prettier-ignore */
```
kainino0x, line 119: [1] this should have moved up along with the header

src/webgpu/api/validation/texture/destroy.spec.ts
```diff
@@ -3,17 +3,61 @@ Destroying a texture more than once is allowed.
...
+g.test('it_is_valid_to_destroy_a_destroyed_texture').fn(t => {
+  const texture = t.device.createTexture(descriptor);
+  texture.destroy();
+  texture.destroy();
 });
+
+g.test('it_is_invalid_to_submit_a_destroyed_texture_before_and_after_encode')
```
kainino0x, line 28: To conform to newer style, please make these test names shorter and put the detail in `.desc()` instead. Every test should have a `.desc()` and we should update them as opportunistically as we can, even if the descriptions are shortish, e.g. "It is invalid to submit with a texture that is destroyed {before, after} encoding finishes"

src/webgpu/api/validation/createTexture.spec.ts
kainino0x: I think it makes sense for `dimension` to be a case instead of a subcase.
For example chromium might pass the 2d and 3d cases but need to suppress the 1d case in automated testing.

btw see https://github.com/gpuweb/cts/pull/485/commits/689ff05ec5a284940719291724d97ded03e6c8c2 on how this can be done

src/webgpu/api/validation/texture/destroy.spec.ts
```diff
@@ -3,17 +3,61 @@ Destroying a texture more than once is allowed.
 `;
 
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { GPUConst } from '../../../constants.js';
 import { ValidationTest } from '../validation_test.js';
 
+const descriptor: GPUTextureDescriptor = {
```
kainino0x, line 9: nit: kDescriptor

But actually I think you can remove this and use `t.getSampledTexture()` or similar.

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -304,6 +269,72 @@ g.test('sampleCount,valid_sampleCount_with_other_parameter_varies')
     }, !success);
   });
 
+g.test('texture_size,default_value_and_smallest_size,uncompressed_format')
```
kainino0x, line 272: I haven't reviewed these two new tests just yet. Would you mind putting them in a separate PR? I think they should separate cleanly.

kainino0x (commented):
Haven't reviewed the two new tests but rest LGTM

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: Great! Done.

src/webgpu/capability_info.ts
Richard-Yunchao: What a coincidence... Hmm, I agree that the current one is more friendly for reading info from the table (and reading is more frequently being used than writing, just as you and Austin discussed). So I changed back, but added a comment to explain it.

src/webgpu/capability_info.ts
Richard-Yunchao: No need now.

src/webgpu/api/validation/texture/destroy.spec.ts
Richard-Yunchao: Done.

src/webgpu/api/validation/texture/destroy.spec.ts
Richard-Yunchao: Done.

src/webgpu/api/validation/createTexture.spec.ts

src/webgpu/api/validation/texture/destroy.spec.ts
```diff
@@ -7,13 +7,57 @@ import { ValidationTest } from '../validation_test.js';
...
+  .desc(`Test that it is valid to destroy a destroyed texture.`)
+  .fn(t => {
+    const texture = t.getSampledTexture();
+    texture.destroy();
+    texture.destroy();
+  });
+
+g.test('submit_a_destroyed_texture')
+  .desc(
+    `Test that it is invalid to submit a destroyed texture before or after command encoder is finished.`
```
kainino0x, line 27:
```suggestion
    `Test that it is invalid to submit with a texture that was destroyed {before, after} encoding finishes.`
```

kainino0x (approved):

src/webgpu/api/validation/texture/destroy.spec.ts
```diff
@@ -24,7 +24,7 @@ g.test('twice')
 
 g.test('submit_a_destroyed_texture')
   .desc(
-    `Test that it is invalid to submit a destroyed texture before or after command encoder is finished.`
+    `Test that it is invalid to submit with a texture that was destroyed {before, after} encoding finishes.`
```
Richard-Yunchao, line 27: Thanks for review, Kai. I should have used the exact description you used as example...

PR #498, opened 2021-03-12
Implement api,operation,rendering,draw:arguments:*

austinEng:
-----

- [ ] WebGPU readability
- [ ] TypeScript readability


src/webgpu/api/operation/rendering/draw.spec.ts
```diff
@@ -44,7 +28,330 @@ TODO:
...
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { assert } from '../../../../common/framework/util/util.js';
 import { GPUTest } from '../../../gpu_test.js';
 
 export const g = makeTestGroup(GPUTest);
+
+g.test('arguments')
+  .desc(
+    `Test that draw arguments are passed correctly by drawing triangles in a grid.
+Horizontally across the screen are triangles with increasing "primitive id".
```
Kangz, line 41: nit: screen -> texture since we aren't using the swapchain.

Also maybe this test could be simpler if it were using points? WDYT?

src/webgpu/api/operation/rendering/draw.spec.ts
```diff
@@ -44,7 +28,330 @@ TODO:
...
+  - index_buffer_offset= {0, 16} - only for indexed draws
+  - base_vertex= {0, 9} - only for indexed draws
+  `
+  )
+  .cases(
+    params()
+      .combine(poptions('first', [0, 3] as const))
+      .combine(poptions('count', [0, 6] as const))
+      .combine(poptions('first_instance', [0, 2] as const))
+      .combine(poptions('instance_count', [0, 4] as const))
```
Kangz, line 66: Maybe also test 1 instance? 0 instance means 0 triangles will be drawn. Ditto for count.

src/webgpu/api/operation/rendering/draw.spec.ts
```diff
@@ -44,7 +28,330 @@ TODO:
...
+[[builtin(instance_index)]] var<in> instance_id : u32;
+[[location(0)]] var<in> vertexPosition : vec2<f32>;
+
+[[builtin(position)]] var<out> Position : vec4<f32>;
+[[stage(vertex)]] fn vert_main() -> void {
+  // 9u is the base_vertex offset. We select different vertices
+  // but apply the same x offset in the grid.
+  // 3u is the number of points in a triangle to convert from index
+  // to id.
+  var vertex_id : u32 = (vertex_index % 9u) / 3u;
```
Kangz, line 124: It could be possible to use a larger render target to now have to mess with the best vertex in the shader.

src/webgpu/api/operation/rendering/draw.spec.ts
```diff
@@ -44,7 +28,330 @@ TODO:
...
+  y = -2.0 * y + 1.0;
+  Position = vec4<f32>(x, y, 0.0, 1.0);
+}
+`,
+    });
+
+    const fragmentModule = t.device.createShaderModule({
+      code: `
+[[block]] struct Output {
+  [[offset(0)]] value : u32;
```
Kangz, line 140: nit: `[[offset(0)]]` has been removed from WGSL

src/webgpu/api/operation/rendering/draw.spec.ts
```diff
@@ -44,7 +28,330 @@ TODO:
...
+
+[[location(0)]] var<out> fragColor : vec4<f32>;
+[[stage(fragment)]] fn frag_main() -> void {
+  output.value = 1u;
+  fragColor = vec4<f32>(0.0, 1.0, 0.0, 1.0);
+}
+`,
+    });
+
+    const pipeline = t.device.createRenderPipeline({
```
Kangz, line 153: nit: this could use the new render pipeline descriptor when webgpu/types is rolled to 0.0.44

Kangz (approved):
LGTM overall

src/webgpu/api/operation/rendering/draw.spec.ts
austinEng: I think it's easier to test points in a test like this since we're translating from 0-1 or NDC coordinates to pixel offsets for the expectations. Triangles let us avoid getting the exact point and instead just check where the center of the triangle is.

src/webgpu/api/operation/rendering/draw.spec.ts
austinEng: 0 is intentional. I think @kainino0x said we don't need to be testing 1 vs. non-zero here?

src/webgpu/api/operation/rendering/draw.spec.ts
austinEng: This was before that landed, but I'll update this now.

kainino0x:
Skipping reviewing this unless you think it needs a second reviewer

src/webgpu/api/operation/rendering/draw.spec.ts
austinEng: Eh, added the cases for 1 triangle and 1 instance.

src/webgpu/api/operation/rendering/draw.spec.ts
austinEng: Done.

src/webgpu/api/operation/rendering/draw.spec.ts
austinEng: Done.

src/webgpu/api/operation/rendering/draw.spec.ts
austinEng: Done.

src/webgpu/api/operation/rendering/draw.spec.ts
Kangz: Ack

Kangz: The nit: about screen->texture in the comment could still be addressed. LGTM otherwise.

Kangz (approved):

src/webgpu/api/operation/rendering/draw.spec.ts
austinEng, line 41:
```suggestion
Horizontally across the texture are triangles with increasing "primitive id".
```

PR #500, opened 2021-03-12
Updating CTS tests to use the new RenderPipelineDescriptor layout

toji:
This change updates all of the RenderPipelineDescriptors (that I could find) to the new format. It isn't merge-able just yet, primarily due to it's dependency on https://github.com/gpuweb/types/pull/61 which has not yet been published to NPM. Posting here for early feedback and to prevent duplicate work.

### DIFF
--- a/src/webgpu/api/operation/resource_init/check_texture/by_ds_test.ts
+++ b/src/webgpu/api/operation/resource_init/check_texture/by_ds_test.ts
@@ -54,8 +54,8 @@ function getDepthTestEqualPipeline(
       format,
       depthCompare: 'equal',
     },
-    primitive: { topology: 'triangle-list', },
-    multisample: { count: sampleCount, },
+    primitive: { topology: 'triangle-list' },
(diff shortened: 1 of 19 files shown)


Kangz:
I published a new version of types. 0.0.44 has your changes

src/webgpu/api/operation/resource_init/check_texture/by_ds_test.ts
```diff
@@ -82,19 +82,19 @@ function getStencilTestEqualPipeline(
         }
         `,
       }),
+      targets: [
```
Kangz, line 85: nit: the definition of targets could be on a single line.

Kangz (approved):
LGTM

src/webgpu/api/operation/resource_init/check_texture/by_ds_test.ts
toji: Done, thanks!

austinEng (approved):

PR #501, opened 2021-03-12
Make the intent of the expand/collapse UI clearer using details elements

toji:
When working with the CTS I found that I was repeatedly confused by the fact that the expand/collapse widget was represented by a checkbox. To me this suggested that only the "checked" tests would be run (as is the case with the WebGL conformance suite), and gave the impression that the test suite was much shallower than it actually was.

To avoid this confusion, I've made a change that uses the HTML `<details>` element rather than a checkbox to handled expanding and collapsing sections. I believe this helps make the intent clearer.

Before:
Screenshot 2021-03-12 151134

After:
Screenshot 2021-03-12 151030


kainino0x (approved):
Great improvement! Never occurred to me to use `<details>` when I was building this, I thought I needed some stupidly complex thing to restyle the checkbox.

PR #503, opened 2021-03-15
add tools/deno runner

lucacasonato:
This commit adds a `./tools/deno` runner for running CTS in Deno. It is
essentially a JS copy of the node runner, using Deno syscalls instead.

It requires you to have run `npm run standalone` before invoking.

image

### DIFF
--- a/tools/deno
+++ b/tools/deno
@@ -1,4 +1,4 @@
-#!/usr/bin/env -S deno run --unstable --allow-read --no-check
+#!/usr/bin/env -S deno run --unstable --allow-read --allow-write --no-check
 
 /* eslint no-console: "off" */
 /* eslint no-process-exit: "off" */
@@ -25,6 +25,15 @@ if (!existsSync("src/common/runtime/cmdline.ts")) {
   console.log("Must be run from repository root");
   usage(1);
 }
+
+const badComment = `/// <reference types="@webgpu/types" />`;
+const navigatorGpuPath = "./out/webgpu/util/navigator_gpu.js";
+let navigatorGpuFile = await Deno.readTextFile(navigatorGpuPath);
+navigatorGpuFile = navigatorGpuFile.replace(
+  badComment,
+  " ".repeat(badComment.length)
+);
+await Deno.writeTextFile(navigatorGpuPath, navigatorGpuFile);
 
 let verbose = false;
 let debug = false;
--- a/tools/deno
+++ b/tools/deno
@@ -15,9 +15,11 @@ function usage(rc) {
   console.log("  tools/deno [OPTIONS...] QUERIES...");
   console.log("  tools/deno 'unittests:*' 'webgpu:buffers,*'");
   console.log("Options:");
-  console.log("  --verbose     Print result/log of every test as it runs.");
-  console.log("  --debug       Include debug messages in logging.");
-  console.log("  --print-json  Print the complete result JSON in the output.");
+  console.log("  --verbose             Print result/log of every test as it runs.");
+  console.log("  --debug               Include debug messages in logging.");
+  console.log("  --print-json          Print the complete result JSON in the output.");
+  console.log("  --trace               Record a trace of the WebGPU operations into a ./trace directory.");
+  console.log("  --ignore-file=<file>  A line delimited text file containing names of tests to skip.");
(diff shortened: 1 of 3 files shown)


tools/deno
```diff
@@ -0,0 +1,185 @@
...
+    }
+  }
+}
+
+if (!existsSync("src/common/runtime/cmdline.ts")) {
+  console.log("Must be run from repository root");
+  usage(1);
+}
+
+const badComment = `/// <reference types="@webgpu/types" />`;
```
kainino0x, line 40: Could you add a comment explaining why this has to be removed? I'd expect it to be parsed as a normal comment.

tools/deno
```diff
@@ -0,0 +1,185 @@
...
+    if (a === "--verbose") {
+      verbose = true;
+    } else if (a === "--trace") {
+      trace = true;
+    } else if (a === "--debug") {
+      debug = true;
+    } else if (a === "--print-json") {
+      printJSON = true;
+    } else if (a.startsWith("--ignore-file=")) {
+      ignoreFile = a.slice("--ignore-file=".length);
```
kainino0x, line 66: FYI, I think soon we might be able to replace this with something integrated into the CTS, as I think we will soon teach it to understand "expectations". That's not ready yet though; we can let you know.
cc @austinEng

tools/deno
```diff
@@ -0,0 +1,185 @@
...
+/* eslint no-process-exit: "off" */
+
+import { existsSync } from "https://deno.land/std@0.90.0/fs/exists.ts";
+
+import { DefaultTestFileLoader } from "../out/common/framework/file_loader.js";
+import { Logger } from "../out/common/framework/logging/logger.js";
+import { parseQuery } from "../out/common/framework/query/parseQuery.js";
+import { assert, unreachable } from "../out/common/framework/util/util.js";
+
+function usage(rc) {
```
kainino0x, line 13: Would it be possible to move some of this code into TypeScript so it can be shared with the code in `src/common/runtime/cmdline.ts`, e.g. `runtime/deno.ts`? Since you're using `npm run standalone` it will transpile these to .js as well.

Happy to land this PR without this, but would be great to work on it as a follow-up.

kainino0x (approved):
Awesome work! Exciting to see this support especially where the Node work hasn't panned out yet :)

> It requires you to have run `npm run standalone` before invoking.

Could you document some simple steps on how to run this? Including this step, and anything else that has to be done (e.g. particular deno version? build from source?) `/docs/deno.md` would be a good place for this, though it could also just be a comment in `/tools/deno`.

tools/deno
lucacasonato: There are two possible options, im open to either:

1. Deno can run TS out of the box. I can convert the `./tools/deno` file to `./tools/deno.ts`, but this would still have to import from the `./out` dir, and would thus not get typings for the imports. This is because I can't import the TS files directly from source, because Deno does not understand them, because deno uses fully specified import specifiers, like the web, which TSC does not support.
2. Add a `src/common/runtime/deno.ts`, add `// @ts-ignore` to the fully qualified http import (because tsc does not understand it), and add inline type definitions for the parts of the `Deno` global we use.

I think 2 is far easier, but maybe also overcomplicated. A possibly better third approach might be to just factor all of the common code into a single ./runtime/cmdline.ts module, which exports a `init` function that takes an object containing all of the system APIs (like `readFile`, `writeFile`, `env`, `args`, etc). We can then have a simple `./tools/node` and `./tools/deno` which both import this init function and pass their respective APIs to it.

tools/deno
lucacasonato: Awesome. Looking forward to that :-)

tools/deno
kainino0x: 3 sounds great to me. (My only other suggestion would be 2 but using a bunch of `any` instead of actually writing inline type definitions, it would be annoying.)

I don't fully understand the issue with 1. Except for a few utilities that only run in Node, all of our import specifiers are relative (and web-compatible as that's the main place we run them). We can have TSC ignore the whole file if that would help?
Anyway, it's okay, I don't need to understand the issue :)

kainino0x (approved):
Thorough docs, thanks!

PR #509, opened 2021-03-17
Support passing skip/fail test expectations in wpt or by command line

austinEng:
On the command line, test expectations are accepted by a command line
flag `--expectations path/to/expectations.ts`, where `expectations.ts` is
a module that exports `RawTestQueryStringWithExpectation[]` as
`expectations`. In WPT, the page should define a global variable
`loadWebGPUExpectations` which is a
`Promise<RawTestQueryStringWithExpectation[]>`.

The test expectations are filtered down to only what is relevant for
the running test query, and applied to the testcases before they run.
A `'skip'` expectation will entirely skip a testcase, and a `'fail'`
expectation will surface failures as passes. If a `'fail'` expectation
is found, but a test actually passes, the harness will log an warning
for the unexpected pass.

-----

- [x] WebGPU readability
- [ ] TypeScript readability

### DIFF
--- a/src/common/framework/test_group.ts
+++ b/src/common/framework/test_group.ts
@@ -1,12 +1,18 @@
-import { Fixture, SkipTestCase } from './fixture.js';
+import { Fixture, SkipTestCase, UnexpectedPassError } from './fixture.js';
+import { Expectation } from './logging/result.js';
 import { TestCaseRecorder } from './logging/test_case_recorder.js';
 import { CaseParams, extractPublicParams, Merged, mergeParams } from './params_utils.js';
+import { compareQueries, Ordering } from './query/compare.js';
+import { TestQuerySingleCase, TestQueryWithExpectation } from './query/query.js';
 import { kPathSeparator } from './query/separators.js';
 import { stringifyPublicParams, stringifyPublicParamsUniquely } from './query/stringify_params.js';
 import { validQueryPart } from './query/validQueryPart.js';
-import { assert } from './util/util.js';
+import { assert, unreachable } from './util/util.js';
 
-export type RunFn = (rec: TestCaseRecorder) => Promise<void>;
+export type RunFn = (
+  rec: TestCaseRecorder,
+  expectations?: TestQueryWithExpectation[]
+) => Promise<void>;
 
 export interface TestCaseID {
   readonly test: readonly string[];
@@ -15,7 +21,11 @@ export interface TestCaseID {
 
 export interface RunCase {
   readonly id: TestCaseID;
-  run: RunFn;
+  run(
+    rec: TestCaseRecorder,
+    selfQuery: TestQuerySingleCase,
+    expectations: TestQueryWithExpectation[]
+  ): Promise<void>;
 }
 
 // Interface for defining tests
@@ -212,7 +222,8 @@ class TestBuilder<F extends Fixture, P extends {}, SubP extends {}> {
         params,
         this.subcaseParams,
         this.fixture,
-        this.testFn
+        this.testFn,
+        this.testCreationStack
       );
     }
   }
@@ -230,27 +241,35 @@ class RunCaseSpecific<
   private readonly subParamGen?: (_: P) => Iterable<SubP>;
   private readonly fixture: FixtureClass<F>;
   private readonly fn: FN;
+  private readonly testCreationStack: Error;
 
   constructor(
     testPath: string[],
     params: P,
     subParamGen: ((_: P) => Iterable<SubP>) | undefined,
     fixture: FixtureClass<F>,
-    fn: FN
+    fn: FN,
+    testCreationStack: Error
   ) {
     this.id = { test: testPath, params: extractPublicParams(params) };
     this.params = params;
     this.subParamGen = subParamGen;
     this.fixture = fixture;
     this.fn = fn;
+    this.testCreationStack = testCreationStack;
   }
 
   async runTest(
     rec: TestCaseRecorder,
     params: P | Merged<P, SubP>,
-    throwSkip: boolean
+    throwSkip: boolean,
+    expectedStatus: Expectation
   ): Promise<void> {
     try {
+      rec.beginSubCase();
+      if (expectedStatus === 'skip') {
+        throw new SkipTestCase('Skipped by expectations');
+      }
       const inst = new this.fixture(rec, params);
 
       try {
@@ -270,10 +289,47 @@ class RunCaseSpecific<
         throw ex;
       }
       rec.threw(ex);
+    } finally {
+      try {
+        rec.endSubCase(expectedStatus);
+      } catch (ex) {
+        assert(ex instanceof UnexpectedPassError);
+        ex.message = `Testcase passed unexpectedly.`;
+        ex.stack = this.testCreationStack.stack;
+        rec.warn(ex);
+      }
     }
   }
 
-  async run(rec: TestCaseRecorder): Promise<void> {
+  async run(
+    rec: TestCaseRecorder,
+    selfQuery: TestQuerySingleCase,
+    expectations: TestQueryWithExpectation[]
+  ): Promise<void> {
+    const getExpectedStatus = (selfQueryWithSubParams: TestQuerySingleCase) => {
+      let didSeeFail = false;
+      for (const exp of expectations) {
+        const ordering = compareQueries(exp.query, selfQueryWithSubParams);
+        if (ordering === Ordering.Unordered || ordering === Ordering.StrictSubset) {
+          continue;
+        }
+
+        switch (exp.expectation) {
+          // Skip takes precendence. If there is any expecation indicating a skip,
+          // signal it immediately.
+          case 'skip':
+            return 'skip';
+          case 'fail':
+            // Otherwise, indicate that we might expect a failure.
+            didSeeFail = true;
+            break;
+          default:
+            unreachable();
+        }
+      }
+      return didSeeFail ? 'fail' : 'pass';
+    };
+
     rec.start();
     if (this.subParamGen) {
       let totalCount = 0;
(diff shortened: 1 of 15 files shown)


austinEng:
@lucacasonato could you check what I've done in `tools/deno` ?

tools/deno
```diff
@@ -74,6 +79,9 @@ for (const a of Deno.args) {
       printJSON = true;
     } else if (a.startsWith("--ignore-file=")) {
       ignoreFile = a.slice("--ignore-file=".length);
+    } else if (a === '--expectations') {
+      const expectationsFile = pathResolve(Deno.cwd(), process.argv[++i]);
```
lucacasonato, line 83:
```suggestion
      const expectationsFile = pathResolve(Deno.cwd(), Deno.args[++i]);
```

lucacasonato (commented):
Looks great!

src/common/runtime/cmdline.ts
```diff
@@ -54,7 +67,33 @@ if (queries.length === 0) {
...
+          return;
+        }
+
+        switch (entry.expectation) {
+          case 'pass':
+          case 'skip':
+          case 'fail':
+            break;
+          default:
+            unreachable();
```
kainino0x, line 88: nit: Probably best to have a message here since this isn't an internal error.

optional nit: `assert(['pass', 'skip', 'fail'].includes(entry.expectation), '...');`

src/common/runtime/wpt.ts
```diff
@@ -17,19 +22,60 @@ declare function setup(properties: { explicit_done?: boolean }): void;
...
   const loader = new DefaultTestFileLoader();
   const qs = new URLSearchParams(window.location.search).getAll('q');
   assert(qs.length === 1, 'currently, there must be exactly one ?q=');
-  const testcases = await loader.loadCases(parseQuery(qs[0]));
+  const testcaseQuery = parseQuery(qs[0]);
+  const testcases = await loader.loadCases(testcaseQuery);
+
+  const expectations: TestQueryWithExpectation[] = [];
+  if (typeof loadWebGPUExpectations !== 'undefined') {
+    (await loadWebGPUExpectations).forEach(entry => {
```
kainino0x, line 45: same as above, and perhaps move it into a helper since it's quite long, and shadows some variables in the surrounding scope too

src/common/runtime/wpt.ts
```diff
@@ -17,19 +22,60 @@ declare function setup(properties: { explicit_done?: boolean }): void;
 declare function promise_test(f: (t: WptTestObject) => Promise<void>, name: string): void;
 declare function done(): void;
 
+declare let loadWebGPUExpectations: Promise<RawTestQueryStringWithExpectation[]> | undefined;
```
kainino0x, line 25: nit: declare const?

kainino0x, line 25: Perhaps add a line in the default `cts.html` template that sets this to `undefined` explicitly, so there's a reference for how to use it.

tools/deno
```diff
@@ -99,7 +107,33 @@ try {
     queries.length === 1,
     "currently, there must be exactly one query on the cmd line",
   );
-  const testcases = await loader.loadCases(parseQuery(queries[0]));
+  const testcaseQuery = parseQuery(queries[0]);
+  const testcases = await loader.loadCases(testcaseQuery);
+
+  const expectations = [];
+  if (typeof loadWebGPUExpectations !== 'undefined') {
+    (await loadWebGPUExpectations).forEach(entry => {
```
kainino0x, line 115: same as above

src/common/runtime/wpt.ts
```diff
@@ -17,19 +22,60 @@ declare function setup(properties: { explicit_done?: boolean }): void;
...
   const qs = new URLSearchParams(window.location.search).getAll('q');
   assert(qs.length === 1, 'currently, there must be exactly one ?q=');
-  const testcases = await loader.loadCases(parseQuery(qs[0]));
+  const testcaseQuery = parseQuery(qs[0]);
+  const testcases = await loader.loadCases(testcaseQuery);
+
+  const expectations: TestQueryWithExpectation[] = [];
+  if (typeof loadWebGPUExpectations !== 'undefined') {
+    (await loadWebGPUExpectations).forEach(entry => {
+      const url = new URL(`${window.location.origin}/${entry.query}`);
```
kainino0x, line 46: Can you add a comment with an example of what this URL will look like? at first glance it looks like it'll be `http://localhost:8080/webgpu:examples:*` but that appears not to be the case.

src/common/runtime/wpt.ts
```diff
@@ -17,19 +22,60 @@ declare function setup(properties: { explicit_done?: boolean }): void;
...
+      if (url.pathname !== window.location.pathname) {
```
kainino0x, line 47: when would this happen? could this be an assert instead?

src/common/runtime/wpt.ts
```diff
@@ -17,19 +22,60 @@ declare function setup(properties: { explicit_done?: boolean }): void;
...
+      }
+
+      const params = url.searchParams;
+      if (workerEnabled !== optionEnabled('worker', params)) {
+        return;
+      }
 
-  const worker = optionEnabled('worker') ? new TestWorker(false) : undefined;
+      const qs = params.getAll('q');
+      assert(qs.length === 1, 'currently, there must be exactly one ?q=');
```
kainino0x, line 57: this is checking the expectation string, right? if so, should clarify the error message

src/common/runtime/cmdline.ts
```diff
@@ -54,7 +67,33 @@ if (queries.length === 0) {
   try {
     const loader = new DefaultTestFileLoader();
     assert(queries.length === 1, 'currently, there must be exactly one query on the cmd line');
-    const testcases = await loader.loadCases(parseQuery(queries[0]));
+    const testcaseQuery = parseQuery(queries[0]);
+    const testcases = await loader.loadCases(testcaseQuery);
+
+    const expectations: TestQueryWithExpectation[] = [];
+    if (typeof loadWebGPUExpectations !== 'undefined') {
+      (await loadWebGPUExpectations).forEach(entry => {
```
kainino0x, line 75: I think almost all of this can be factored out, it's almost the same across runtimes

kainino0x (commented):
Nice! The subcase stuff is a bit gnarly, but I think it will work.

src/common/framework/test_group.ts
```diff
@@ -270,18 +289,68 @@ class RunCaseSpecific<
...
       }
       rec.threw(ex);
+    } finally {
+      try {
+        rec.endSubCase(expectedStatus);
+      } catch (ex) {
+        assert(ex instanceof UnexpectedPassError);
+        ex.message = `Testcase passed unexpectedly.`;
+        ex.stack = this.testCreationStack.stack;
+        rec.warn(ex);
```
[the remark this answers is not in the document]
austinEng: Maybe? I'd like to address that whenever we add the unexpected warns.

src/common/framework/test_group.ts
```diff
@@ -270,18 +289,68 @@ class RunCaseSpecific<
...
         rec.info(new Error('subcase: ' + stringifyPublicParams(subParams)));
         try {
-          await this.runTest(rec, mergeParams(this.params, subParams), true);
+          const params = mergeParams(this.params, subParams);
+          await this.runTest(
+            rec,
+            params,
+            true,
+            getExpectedStatus(
+              new TestQuerySingleCase(
```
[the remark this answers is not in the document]
austinEng: Done

src/common/runtime/cmdline.ts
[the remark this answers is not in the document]
austinEng: Done

src/common/runtime/cmdline.ts
austinEng: I kept the switch statement. The other way doesn't narrow the type so it doesn't infer correctly.

src/common/runtime/wpt.ts
austinEng: Done.

src/common/runtime/wpt.ts
austinEng: Done.

austinEng: Done.

tools/deno
austinEng: Done.

src/common/runtime/wpt.ts
austinEng: Added a jsdoc comment. It's not super great..? But the format is slightly different in/out of WPT.

src/common/runtime/wpt.ts
austinEng: in case you want to have a cts1.html and a cts2.html? Maybe we don't need that though

src/common/runtime/wpt.ts
austinEng: Done.

src/common/runtime/cmdline.ts
austinEng: Done.

src/common/runtime/wpt.ts
kainino0x: Let's assume it can't happen for now

src/common/framework/test_group.ts
```diff
@@ -338,19 +338,13 @@ class RunCaseSpecific<
...
-                <CaseParams>params
-              )
-            )
+          const subcaseQuery = new TestQuerySingleCase(
+            selfQuery.suite,
+            selfQuery.filePathParts,
+            selfQuery.testPathParts,
+            <CaseParams>params
           );
+          await this.runTest(rec, params, true, getExpectedStatus(subcaseQuery));
```
kainino0x, line 347: Thinking about this again. How much simpler would things be if we didn't support subcase expectations? Probably not enough that it's worth undoing the work. But I don't think we necessarily need them. We can generally keep dimensions out of the subcases if they are useful to suppress on.

src/common/runtime/wpt.ts
```diff
@@ -17,19 +18,29 @@ declare function setup(properties: { explicit_done?: boolean }): void;
...
 
 (async () => {
+  const workerEnabled = optionEnabled('worker');
+  const worker = workerEnabled ? new TestWorker(false) : undefined;
+
   const loader = new DefaultTestFileLoader();
   const qs = new URLSearchParams(window.location.search).getAll('q');
   assert(qs.length === 1, 'currently, there must be exactly one ?q=');
-  const testcases = await loader.loadCases(parseQuery(qs[0]));
+  const testcaseQuery = parseQuery(qs[0]);
```
kainino0x, line 36: nit: this usually identifies more than one "testcase", so it should be called something like `filterQuery`. Same in deno and cmdline.

kainino0x (approved):

src/common/framework/test_group.ts
austinEng: I'm going to keep it for now as it's nice for the test author to not have to care about what is/is not possible with expectations.

PR #511, opened 2021-03-23
Clean up TODOs for queries validation

haoxli:
- Some TODOs have been completed, remove them from TODO list.
- Start a query without draw call is allowed, move them to operation tests.
- Add pipelineStatistics checking in pipeline_statistics.spec.ts

-----

- [ ] WebGPU readability
- [ ] TypeScript readability


src/webgpu/api/validation/queue/destroyed/query_set.spec.ts
```diff
@@ -1,8 +1,7 @@
 export const description = `
 Tests using a destroyed query set on a queue.
 
-- used in {resolveQuerySet, timestamp {compute, render, non-pass},
-    pipeline statistics {compute, render}, occlusion}
+- used in pipeline statistics {compute, render}
```
Kangz, line 4: This could just be "TODO: Test with pipeline statistics queries as well." and the TODO at the bottom removed.

src/webgpu/api/validation/encoding/queries/occlusion.spec.ts
```diff
@@ -1,14 +0,0 @@
-export const description = `
```
Kangz, line 1: It's a bit weird to have `general.spec.ts` and `begin_end.spec.ts` when the tests are basically different for all query types. But that matches the test plan so let's do that for now and see if we need to resplit tests later.

Kangz (approved):

PR #512, opened 2021-03-24
Fix check of `loadWebGPUExpectations` for undefined

austinEng:
`if (someVariable)` does not work if the variable is
truly undefined. `typeof someVariable !== 'undefined'` should
be used instead.

-----

- [x] WebGPU readability
- [ ] TypeScript readability


Kangz (approved):
LGTM

PR #515, opened 2021-03-30
WPT: Pass expectations to worker.run

austinEng:
-----

- [x] WebGPU readability
- [x] TypeScript readability


kainino0x (approved):
LGTM though it occurs to me: the expectations object is pretty large right? we shouldn't really need to post it multiple times; ideally we should probably send it to the worker once a the beginning

PR #516, opened 2021-04-01
Fix multicase test expectation filtering

austinEng:
The filter was removing multicase expectations that did not have
overlapping params with the test query. Strip the params since a
multicase expectation (ending in "*") could match a test query
even if the explicitly listed params are different.

ex.) foo=2;* may match bar=3;*

-----

- [x] WebGPU readability
- [x] TypeScript readability

### DIFF
--- a/src/common/framework/query/query.ts
+++ b/src/common/framework/query/query.ts
@@ -207,7 +207,20 @@ Expectation should be of the form path/to/cts.html?worker=0&q=suite:test_path:te
       expectationQuery = parseQuery(entry.query);
     }
 
-    if (compareQueries(query, expectationQuery) === Ordering.Unordered) {
+    // Strip params from multicase expectations so that an expectation of foo=2;*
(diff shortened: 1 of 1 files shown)


src/common/framework/query/query.ts
```diff
@@ -207,7 +207,20 @@ Expectation should be of the form path/to/cts.html?worker=0&q=suite:test_path:te
       expectationQuery = parseQuery(entry.query);
     }
 
-    if (compareQueries(query, expectationQuery) === Ordering.Unordered) {
+    // Strip params from multicase expectations so that an expectation of foo=2;*
+    // is stored if the test query is bar=3;*
+    const queryForFilter =
+      expectationQuery instanceof TestQueryMultiCase
```
kainino0x, line 213: uNit: I prooobably would have done this on the query instead of the expectationQuery but I think they're equivalent

kainino0x (approved):

PR #519, opened 2021-04-02
Some depth related tests

shrekshao:
Some depth related tests. For complete depth/stencil tests I think there's still a lot to port from dawn end2end tests. But this should be something to get started and can fill in in the future.

The reverse depth test actually only test depth compare func and clipping range.

depth16unorm fails because it's not implemented yet.

-----

- [x] WebGPU readability
- [x] TypeScript readability

### DIFF
--- a/src/webgpu/api/operation/rendering/depth.spec.ts
+++ b/src/webgpu/api/operation/rendering/depth.spec.ts
@@ -1,0 +1,266 @@
+export const description = `
+Test related to depth buffer, depth op, compare func, etc.
+`;
+
+import { pbool, params, poptions } from '../../../../common/framework/params_builder.js';
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { kDepthStencilFormats } from '../../../capability_info.js';
+import { GPUTest } from '../../../gpu_test.js';
+
+const backgroundColor = [0x00, 0x00, 0x00, 0xff];
+const triangleColor = [0xff, 0xff, 0xff, 0xff];
+
+export const g = makeTestGroup(GPUTest);
+
+g.test('depth_compare_func')
+  .desc(
+    `Tests each depth compare function works properly. The depth of the triangle is always 0.5.`
+  )
+  .cases(
+    params()
+      .combine(
+        poptions(
+          'format',
+          kDepthStencilFormats.filter(format => format !== 'stencil8')
+        )
+      )
+      .combine([
+        { depthCompare: 'never', depthLoadValue: 1.0, expected: backgroundColor },
+        { depthCompare: 'never', depthLoadValue: 0.5, expected: backgroundColor },
+        { depthCompare: 'never', depthLoadValue: 0.0, expected: backgroundColor },
+        { depthCompare: 'less', depthLoadValue: 1.0, expected: triangleColor },
+        { depthCompare: 'less', depthLoadValue: 0.5, expected: backgroundColor },
+        { depthCompare: 'less', depthLoadValue: 0.0, expected: backgroundColor },
+        { depthCompare: 'less-equal', depthLoadValue: 1.0, expected: triangleColor },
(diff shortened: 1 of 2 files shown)


src/webgpu/api/operation/rendering/depth.spec.ts
```diff
@@ -0,0 +1,266 @@
...
+                  vec2<f32>(-1.0, -3.0),
+                  vec2<f32>(3.0, 1.0),
+                  vec2<f32>(-1.0, 1.0));
+              // TODO: remove workaround for Tint unary array access broke
+              const zv : array<vec2<f32>, 4> = array<vec2<f32>, 4>(
+                  vec2<f32>(0.2, 0.2),
+                  vec2<f32>(0.3, 0.3),
+                  vec2<f32>(-0.1, -0.1),
+                  vec2<f32>(1.1, 1.1));
+              const z : f32 = zv[InstanceIndex].x;
```
kainino0x, line 198: nit: Seems like InstanceIndex is only 0 or 1 so some of the entries in `zv` and `colors` are unused?

src/webgpu/api/operation/rendering/depth.spec.ts
```diff
@@ -0,0 +1,266 @@
...
+    pass.draw(3, 2);
+    pass.endPass();
+    encoder.copyTextureToBuffer(
+      { texture: colorAttachment, mipLevel: 0, origin: { x: 0, y: 0, z: 0 } },
+      { buffer: dst, bytesPerRow: 256 },
+      { width: 1, height: 1, depthOrArrayLayers: 1 }
+    );
+    t.device.queue.submit([encoder.finish()]);
+
+    t.expectContents(
```
kainino0x, line 262: same comment about `expectSinglePixelIn2DTexture`

kainino0x (approved):
LGTM!

src/webgpu/api/operation/rendering/depth.spec.ts
shrekshao: oops. Fixed draw instance count to 4 to utilize those.

shrekshao:
Thanks for the review!

src/webgpu/api/operation/rendering/depth.spec.ts
```diff
@@ -84,11 +92,7 @@ g.test('depth_compare_func')
             [[location(0)]] var<out> color : vec4<f32>;
 
             [[stage(vertex)]] fn main() -> void {
-              const pos : array<vec2<f32>, 3> = array<vec2<f32>, 3>(
-                  vec2<f32>(-1.0, -3.0),
-                  vec2<f32>(3.0, 1.0),
-                  vec2<f32>(-1.0, 1.0));
-              Position = vec4<f32>(pos[VertexIndex], 0.5, 1.0);
+              Position = vec4<f32>(0.5, 0.5, 0.5, 1.0);
```
kainino0x, line 95: I just realized I think(?) the middle of the screen would have been 0,0 not 0.5,0.5. Sorry. There should be nothing wrong with 0.5,0.5 though so just leaving it.

kainino0x (approved):

kainino0x:
Thanks!

src/webgpu/api/operation/rendering/depth.spec.ts
shrekshao: Ah... I didn't realize that either...

shrekshao: Well since we are on this topic. What is the equivalent for gl_PointSize in webgpu/wgsl ? (didn't find one)

kainino0x: There is no gl_PointSize in WebGPU, you have to use instanced rectangles if you want bigger points. This is to avoid the portability issues with larger points like in WebGL. (On some AMD they clip if the center of the point goes off the screen, so half of the large point disappears suddenly)

PR #521, opened 2021-04-06
Fix a bug in createTexture validation for multisample

Richard-Yunchao:
If a texture format support multisample, it should be renderable:
color-renderable, depth-renderable, or stencil-renderable.

See gpuweb issue at https://github.com/gpuweb/gpuweb/issues/1522.

This change will fix Dawn issue 731:
https://bugs.chromium.org/p/dawn/issues/detail?id=731.

-----

- [x] WebGPU readability
- [x] TypeScript readability


Richard-Yunchao:
Note that I still use a separate column `multisample` to manifest whether each format can support multisample because 1) it is more clear for multisample, we don't need to refer to .renderable for multisample support,  2) it is very unlikely but multisample support might be different with `renderable` in future?  

But if you think we can just remove `multisample` column in capability_info.ts, I am OK to do that :)

Kangz (approved):

kainino0x:
> Note that I still use a separate column `multisample` to show whether each format can support multisample because 1) it is more clear for multisample, we don't need to refer to .renderable for multisample support, 2) it is very unlikely but multisample support might be different with `renderable` in future?

LGTM, I agree.

austinEng:
You can run the third_party/blink/web_tests/webgpu/roll_webgpu_cts.sh script. Probably needs test expectation updates in third_party/blink/web_tests/WebGPUExpectations

Richard-Yunchao:
Got it. I will have a try.

PR #526, opened 2021-04-07
Add tests for GPUBlendComponent

austinEng:
-----

- Do these tests pass (or partially pass without unexpected issues) in an implementation?
Assuming the decision in https://github.com/gpuweb/gpuweb/issues/65 is to follow Vulkan/Metal, these tests pass in a build of Chrome on Windows updated to reflect that decision. Also pass in a local build on Mac, but surprisingly are flaky in Canary. Seems like a caching bug. After a bunch of tests pass, the randomly all start failing. Rerunning the failing tests in a new tab passes. Probably best to avoid merging until this is confirmed.
- [x] Incomplete tests/files and known test issues, if any, are marked with "TODO" (in a description) or are `.unimplemented()`.

- [x] WebGPU readability
- [x] TypeScript readability

### DIFF
--- a/src/webgpu/api/operation/rendering/blending.spec.ts
+++ b/src/webgpu/api/operation/rendering/blending.spec.ts
@@ -10,6 +10,7 @@ TODO:
 
 import { params, poptions } from '../../../../common/framework/params_builder.js';
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { assert } from '../../../../common/framework/util/util.js';
 import { GPUTest } from '../../../gpu_test.js';
 
 export const g = makeTestGroup(GPUTest);
@@ -53,7 +54,7 @@ function mapColor(
 function computeBlendFactor(
   src: GPUColorDict,
   dst: GPUColorDict,
-  col: GPUColorDict,
+  blendColor: GPUColorDict | undefined,
   factor: GPUBlendFactor
 ): GPUColorDict {
   switch (factor) {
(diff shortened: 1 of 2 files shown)


src/webgpu/api/operation/rendering/blending.spec.ts
```diff
@@ -7,7 +8,269 @@ TODO:
...
+    case 'dst-color':
+      return { ...dst };
+    case 'one-minus-dst-color':
+      return mapColor(dst, v => 1 - v);
+    case 'dst-alpha':
+      return mapColor(dst, () => dst.a);
+    case 'one-minus-dst-alpha':
+      return mapColor(dst, () => 1 - dst.a);
+    case 'src-alpha-saturated': {
+      const f = Math.min(src.a, 1 - src.a);
```
kainino0x, line 81:
```suggestion
      const f = Math.min(src.a, 1 - dst.a);
```
according to
https://docs.microsoft.com/en-us/windows/win32/api/d3d12/ne-d3d12-d3d12_blend

src/webgpu/api/operation/rendering/blending.spec.ts
```diff
@@ -7,7 +8,269 @@ TODO:
...
+                srcFactor: 'one',
+                dstFactor: 'zero',
+                operation: 'add',
+              },
+              alpha: {
+                srcFactor: 'one',
+                dstFactor: 'zero',
+                operation: 'add',
+              },
+              // And then override the component we're testing.
```
kainino0x, line 175: :galaxy_brain:

But also the above values for `color` and `alpha` are the defaults so you should be able to just omit them for the same effect

src/webgpu/api/operation/rendering/blending.spec.ts
```diff
@@ -7,7 +8,269 @@ TODO:
...
+    const textureFormat: GPUTextureFormat = 'rgba32float';
+    const srcColor: GPUColorDict = { r: 0.11, g: 0.61, b: 0.81, a: 0.44 };
+    const dstColor: GPUColorDict = { r: 0.51, g: 0.22, b: 0.71, a: 0.33 };
+    const varColor: GPUColorDict = { r: 0.91, g: 0.82, b: 0.73, a: 0.64 };
+
+    const srcFactor = computeBlendFactor(srcColor, dstColor, varColor, t.params.srcFactor);
+    const dstFactor = computeBlendFactor(srcColor, dstColor, varColor, t.params.dstFactor);
+
+    const expectedColor =
+      t.params.operation === 'min' || t.params.operation === 'max'
```
kainino0x, line 137: computeBlendOperation already switches on the GPUBlendOperation, so maybe just fold this logic into it?

I didn't realize min/max were different. Confusing. Didn't see mention of this in D3D12 docs either.

src/webgpu/api/operation/rendering/blending.spec.ts
```diff
@@ -7,7 +8,269 @@ TODO:
...
+  are done correctly by blending a single pixel. The test uses rgba32float as the format
+  to avoid checking clamping behavior (tested in api,operation,rendering,blending:clamp,*).
+
+  Params:
+    - component= {color, alpha} - whether to test blending the color or the alpha component.
+    - srcFactor= {...all GPUBlendFactors}
+    - dstFactor= {...all GPUBlendFactors}
+    - operation= {...all GPUBlendOperations}
+  `)
+  .cases(
```
kainino0x, line 120: Not sure where this should be, but we should leave a TODO _somewhere_ for blending into srgb render targets. (Just thought of this while [looking at](https://www.khronos.org/registry/vulkan/specs/1.2/html/chap27.html#framebuffer-blendoperations) the Vulkan spec.)

kainino0x (approved):

src/webgpu/api/operation/rendering/blending.spec.ts
kainino0x: Oh, and something to make sure blending uses src0 properly (in case an implementation reorders render targets or something).

src/webgpu/api/operation/rendering/blending.spec.ts
kainino0x: nvm I was wrong about this, the dictionaries have defaults but `color` and `alpha` are required.

src/webgpu/api/operation/rendering/blending.spec.ts
```diff
@@ -7,7 +8,294 @@ TODO:
...
+      assert(blendColor !== undefined);
+      return { ...blendColor };
+    case 'one-minus-blend-color':
+      assert(blendColor !== undefined);
+      return mapColor(blendColor, v => 1 - v);
+  }
+}
+
+function computeBlendOperation(src: GPUColorDict, srcFactor: GPUColorDict,
+  dst: GPUColorDict, dstFactor: GPUColorDict,operation: GPUBlendOperation) {
```
kainino0x, line 95: Confused why the linter is not forcing this to get autoformatted. Something must have broken in the CI at some point...

PR #527, opened 2021-04-09
Fix bugs for texture creation validation

Richard-Yunchao:
This change filters out these dimenstion type and format combination
errors for texture creation validation:
  - Compressed formats are invalid for 1D and 3D,
  - Depth/stencil formats are invalid for 3D,

See the discussion about these validation rules at
https://github.com/gpuweb/gpuweb/issues/1522 and
https://bugs.chromium.org/p/dawn/issues/detail?id=730

<hr>

### DIFF
--- a/src/webgpu/api/validation/createTexture.spec.ts
+++ b/src/webgpu/api/validation/createTexture.spec.ts
@@ -12,6 +12,7 @@ import {
   kTextureUsages,
   kUncompressedTextureFormats,
   kUncompressedTextureFormatInfo,
+  dimensionTypeAndFormatCombinationError,
 } from '../../capability_info.js';
 import { DefaultLimits, GPUConst } from '../../constants.js';
 import { maxMipLevelCount } from '../../util/texture/base.js';
@@ -79,9 +80,34 @@ g.test('zero_size')
     };
 
     const success = zeroArgument === 'none';
+
     t.expectValidationError(() => {
       t.device.createTexture(descriptor);
     }, !success);
+  });
+
+g.test('dimension_type_on_formats')
+  .desc(
+    `Test every dimension types on every formats. Note that 1) compressed formats are not valid for 1D and 3D dimension types and, 2) depth/stencil formats are not valid for 3D dimension type.`
+  )
+  .cases(poptions('dimension', [undefined, ...kTextureDimensions]))
+  .subcases(() => params().combine(poptions('format', kAllTextureFormats)))
+  .fn(async t => {
+    const { dimension, format } = t.params;
+
+    const info = kAllTextureFormatInfo[format];
+    await t.selectDeviceOrSkipTestCase(info.extension);
+
+    const descriptor: GPUTextureDescriptor = {
+      size: [info.blockWidth, info.blockHeight, 1],
+      dimension,
+      format,
+      usage: GPUTextureUsage.SAMPLED,
+    };
+
+    t.expectValidationError(() => {
+      t.device.createTexture(descriptor);
+    }, dimensionTypeAndFormatCombinationError(dimension, format));
   });
 
 g.test('mipLevelCount,format')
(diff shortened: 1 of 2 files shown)


src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -535,6 +573,8 @@ g.test('texture_size,3d_texture,uncompressed_format')
     }, !success);
   });
 
+// Compressed formats are not supported in 3D in WebGPU v1 because they are complicated but not very useful.
+/*
 g.test('texture_size,3d_texture,compressed_format')
```
Richard-Yunchao, line 578: According to the discussion at https://github.com/gpuweb/gpuweb/issues/1522, we will not support compressed formats 3D texture. Dawn project follows this rule too. So comment out this test. It spent some time on it, it might be useful if we decided to support compressed formats for 3D texture. So I didn't remove this test. 

If you think you should remove this test. I am fine to do so.

Richard-Yunchao (commented):
PTAL, thanks!

src/webgpu/api/validation/createTexture.spec.ts
kainino0x: If we want to keep the test around (I think it's good, I _think_ we will probably eventually add 3d compressed textures) then I would prefer to keep it uncommented so the code doesn't get too stale. The skip at the beginning is OK I think. Running the skipped test should only waste a few milliseconds at most.

src/webgpu/capability_info.ts
```diff
@@ -290,6 +290,17 @@ export function depthStencilFormatAspectSize(
   return texelAspectSize;
 }
 
+export function dimensionTypeAndFormatCombinationError(
```
kainino0x, line 293: uNit: Prefer to invert this as that better describes "capability info" per the filename: true if the combination is valid, false if error. e.g. "textureDimensionAndFormatCompatible"

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -79,11 +80,36 @@ g.test('zero_size')
...
     const success = zeroArgument === 'none';
+
     t.expectValidationError(() => {
       t.device.createTexture(descriptor);
     }, !success);
   });
 
+g.test('dimension_type_on_formats')
+  .desc(
+    `Test every dimension types on every formats. Note that 1) compressed formats are not valid for 1D and 3D dimension types and, 2) depth/stencil formats are not valid for 3D dimension type.`
```
kainino0x, line 91: Just to confirm, 1d depth/stencil textures are valid? I don't think you can render into 1d textures, so not sure what the usefulness is. Even if the APIs support it (why?) I wouldn't be surprised if it's an underutilized path that turns out to be buggy...

kainino0x (approved):
LGTM

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: Yeah. Just noted the comments that depth/stencil formats are invalid for 1D on Metal (from the log Austin put on dawn issue 730, and Corentin's comment on WebGPU issue 1522). I don't have a Mac device to try though. But the log should be reliable. And just as you said. Depth/stencil formats for 1D textures can be buggy in real world drivers.
Done.

src/webgpu/capability_info.ts
Richard-Yunchao: Done

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: You are right. We have already discussed this before. I missed the skip() part below and thought I didn't do that.

kainino0x: No worries, I'd have forgotten too but I just happened to see the skip() line.

kainino0x (approved):

PR #538, opened 2021-04-17
Update bindings API usage

kainino0x:
<hr>

### DIFF
--- a/src/webgpu/api/validation/createBindGroupLayout.spec.ts
+++ b/src/webgpu/api/validation/createBindGroupLayout.spec.ts
@@ -5,21 +5,18 @@ TODO: update for new binding structure, remove eslint-disable.
 TODO: review existing tests, write descriptions, and make sure tests are complete.
 `;
 
-import { pbool, poptions, params } from '../../../common/framework/params_builder.js';
+import { poptions, params } from '../../../common/framework/params_builder.js';
 import { makeTestGroup } from '../../../common/framework/test_group.js';
 import {
-  kBindingTypeInfo,
-  kBindingTypes,
-  kBufferBindingTypeInfo,
   kMaxBindingsPerBindGroup,
   kShaderStages,
   kShaderStageCombinations,
-  kTextureBindingTypeInfo,
-  kTextureComponentTypes,
   kTextureViewDimensions,
-  kTextureViewDimensionInfo,
-  kAllTextureFormats,
-  kAllTextureFormatInfo,
+  allBindingEntries,
+  bindingTypeInfo,
+  bufferBindingTypeInfo,
+  kBufferBindingTypes,
+  BGLEntry,
 } from '../../capability_info.js';
 
 import { ValidationTest } from './validation_test.js';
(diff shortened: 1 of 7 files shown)


src/webgpu/api/validation/createBindGroupLayout.spec.ts
```diff
@@ -51,91 +48,17 @@ g.test('some_binding_index_was_specified_more_than_once').fn(async t => {
...
+    const success = (visibility & ~info.validStages) === 0;
 
     t.expectValidationError(() => {
       t.device.createBindGroupLayout({
-        entries: [{ binding: 0, visibility, type, storageTextureFormat }],
-      });
-    }, !success);
-  });
-
-g.test('bindingTypeSpecific_optional_members')
```
kainino0x, line 74: Test removed because:
- these rules are mostly no longer relevant with the new bindings style
- testing that `undefined` is _allowed_ is really an IDL test
- it makes more sense to test the _behavior_ of the defaulting in operation tests.

kainino0x:
@Richard-Yunchao PTAL when you have a chance!

src/webgpu/capability_info.ts
```diff
@@ -494,65 +478,134 @@ const kBindingKind: {
...
+    { texture: { multisampled: true } },
+  ] as const;
+}
+export function storageTextureBindingEntries(format: GPUTextureFormat): readonly BGLEntry[] {
+  return [
+    { storageTexture: { access: 'read-only', format } },
+    { storageTexture: { access: 'write-only', format } },
+  ] as const;
+}
+export function sampledAndStorageBindingEntries(
```
Richard-Yunchao, line 581: The naming stuff for textureBindingEntries, storageTextureBindingEntries, and sampledAndStorageBindingEntries sounds a little bit weird for me. I may like to use sampledTextureBindingEntries, storageTextureBindingEntries, and TextureBindingEntries (including sampledTexture and storageTexture). But the spec does use TextureBindingLayout and StorageTextureBindingLayout, which is a little weird for me too (Why not use SampledTextureBindingLayout and StorageTextureBindingLayout?). 

I assume that You and other folks in WebGPU WG have already discussed this and chose the right names, though.

Richard-Yunchao (approved):
LGTM. Please ignore the comment if you have already discussed the names.

src/webgpu/capability_info.ts
kainino0x: Makes sense to make this more explicit in this PR, will do that. As for the spec, sampled textures are by far the most common case, so I think we just wanted to make "texture" look like the more common case over "storageTexture". There's certainly an argument that both should be explicit, but I don't think it's worth another change at this point...

Richard-Yunchao: I revisited the discussions when (Google's) Brandon submitted a PR to change BGL. There were a lot of discussions. A few guys did raise the concern about these names like GPUTextureBindingLayout and GPUStorageTextureBindingLayout at that time. But the WG chose the current names. It's fine. I will not waste my time to submit a PR to change the names.

PR #544, opened 2021-04-24
gen_wpt_cts_html: always split down to individual tests

kainino0x:
This increases the number of variants from 117 to 262.
Since all tests are handwritten, this number won't grow out of control.

An equivalent change was made in the standalone viewer a while ago,
to give a better overview of the CTS and test plan.

<hr>
    
### DIFF
--- a/src/common/framework/tree.ts
+++ b/src/common/framework/tree.ts
@@ -78,16 +78,13 @@ export class TestTree {
   }
 
   /**
-   * If a parent and its child are at different levels, then
-   * generally the parent has only one child, i.e.:
+   * Dissolve nodes which have only one child, e.g.:
    *   a,* { a,b,* { a,b:* { ... } } }
-   * Collapse that down into:
+   * collapses down into:
    *   a,* { a,b:* { ... } }
    * which is less needlessly verbose when displaying the tree in the standalone runner.
-   *
-   * TODO: fix the name of this
    */
-  dissolveLevelBoundaries(): void {
+  dissolveSingleChildTrees(): void {
     const newRoot = dissolveSingleChildTrees(this.root);
     assert(newRoot === this.root);
   }
--- a/src/common/framework/tree.ts
+++ b/src/common/framework/tree.ts
@@ -62,17 +62,44 @@ export interface TestTreeLeaf {
 
 export type TestTreeNode = TestSubtree | TestTreeLeaf;
 
+/**
+ * When iterating through "collapsed" tree nodes, indicates how many "query levels" to traverse
+ * through before starting to collapse nodes.
+ *
+ * Corresponds with TestQueryLevel, but excludes 4 (SingleCase):
+ * - 1 = MultiFile. Expands so every file is in the collapsed tree.
+ * - 2 = MultiTest. Expands so every test is in the collapsed tree.
+ * - 3 = MultiCase. Expands so every case is in the collapsed tree (i.e. collapsing disabled).
+ */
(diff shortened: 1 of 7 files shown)


kainino0x:
@austinEng WDYT?

austinEng (approved):
This was surprisingly easier than I expected. LGTM though - looks like we need to update tests?

kainino0x:
> looks like we need to update tests?

Yup, I skipped this initially. Will update them.

kainino0x:
PTAL again, bigger patch now but I came up with a more straightforward solution: just specify which level collapsing should start at (`collapseStartingAtLevel`) and removed some weird special cases in the code

kainino0x:
nvm, realized it's not behaving as expected for the standalone demo...

kainino0x:
> nvm, realized it's not behaving as expected for the standalone demo...

fixed by being less ambitious. PTAL

src/common/framework/tree.ts
```diff
@@ -62,30 +62,37 @@ export interface TestTreeLeaf {
 
 export type TestTreeNode = TestSubtree | TestTreeLeaf;
 
+export type ExpandThroughLevel = 1 | 2 | 3;
```
austinEng, line 65: Could use a comment - do these levels mean suite, file, test ?

Do we plan to extend this further so you can pass any number and it'll expand params as well?
Oh I see - maybe that's what the `const expandThrough = Math.max(...` is for?

src/common/tools/gen_wpt_cts_html.ts
```diff
@@ -40,67 +40,64 @@ const [
...
-          urlQueryString.length < 185,
-          'Generated test variant would produce too-long -actual.txt filename. \
+  const loader = new DefaultTestFileLoader();
+  const lines: Array<string | undefined> = [];
+  for (const prefix of argsPrefixes) {
+    const rootQuery = new TestQueryMultiFile(suite, []);
+    const tree = await loader.loadTree(rootQuery, expectations.get(prefix)!);
+
+    lines.push(undefined); // output blank line between prefixes
+    for (const q of tree.iterateCollapsedQueries(false, 2)) {
```
austinEng, line 86: nit: comment why this is 2

src/common/tools/checklist.ts
```diff
@@ -51,7 +51,7 @@ function checkForUnmatchedSubtrees(tree: TestTree, matchQueries: TestQuery[]): n
   let subtreeCount = 0;
   const unmatchedSubtrees: TestQuery[] = [];
   const overbroadMatches: [TestQuery, TestQuery][] = [];
-  for (const collapsedSubtree of tree.iterateCollapsedQueries(true)) {
+  for (const collapsedSubtree of tree.iterateCollapsedQueries(true, 1)) {
```
austinEng, line 54: nit: comment why this is 1

src/common/framework/tree.ts
```diff
@@ -62,30 +62,37 @@ export interface TestTreeLeaf {
...
+
 export class TestTree {
+  readonly forQuery: TestQuery;
```
austinEng, line 68: nit: comment what forQuery does - from what I understand, makes it so that iterating the tree filters by the query

kainino0x:
Done

austinEng (approved):
still LGTM, thanks for the comments

kainino0x:
@toji FYI I think this will block the autoroller; to fix we'll probably need to expand out any file-level expectations to possibly-multiple test-level expectations.

PR #548, opened 2021-04-30
Added some CompilationInfo tests.

toji:
Adds some simple tests to ensure CompilationInfo reported from ShaderModules is sane. This is difficult to test strictly since different implementations have a fair amount of leeway in how the identify and report shader errors, but we can at least validate that the messages reported are sane (i.e: reference actual positions within the shader code) and internally consistent.

These will pass in Chrome once https://chromium-review.googlesource.com/c/chromium/src/+/2861210 lands, though there will be known issues with reporting offsets if there is unicode in the shader.

<hr>

### DIFF
--- a/src/webgpu/api/operation/shader_module/compilation_info.spec.ts
+++ b/src/webgpu/api/operation/shader_module/compilation_info.spec.ts
@@ -12,19 +12,19 @@ const kValidShaderSources = [
   {
     valid: true,
     unicode: false,
-    codeFn: () => `
-  [[stage(vertex)]] fn main() -> [[builtin(position)]] vec4<f32> {
-    return vec4<f32>(0.0, 0.0, 0.0, 1.0);
-  }`,
+    _code: `
+      [[stage(vertex)]] fn main() -> [[builtin(position)]] vec4<f32> {
+        return vec4<f32>(0.0, 0.0, 0.0, 1.0);
+      }`,
   },
   {
     valid: true,
     unicode: true,
-    codeFn: () => `
-  // 頂点シェーダー
-  [[stage(vertex)]] fn main() -> [[builtin(position)]] vec4<f32> {
-    return vec4<f32>(0.0, 0.0, 0.0, 1.0);
-  }`,
+    _code: `
+      // 頂点シェーダー 👩‍💻
+      [[stage(vertex)]] fn main() -> [[builtin(position)]] vec4<f32> {
+        return vec4<f32>(0.0, 0.0, 0.0, 1.0);
+      }`,
   },
 ];
 
@@ -32,23 +32,23 @@ const kInvalidShaderSources = [
   {
     valid: false,
     unicode: false,
-    errorLine: 4,
-    codeFn: () => `
-  [[stage(vertex)]] fn main() -> [[builtin(position)]] vec4<f32> {
-    // Expected Error: vec4 should be vec4<f32>
-    return vec4(0.0, 0.0, 0.0, 1.0);
-  }`,
+    _errorLine: 4,
+    _code: `
+      [[stage(vertex)]] fn main() -> [[builtin(position)]] vec4<f32> {
+        // Expected Error: vec4 should be vec4<f32>
+        return vec4(0.0, 0.0, 0.0, 1.0);
+      }`,
   },
   {
     valid: false,
     unicode: true,
-    errorLine: 5,
-    codeFn: () => `
-  // 頂点シェーダー
-  [[stage(vertex)]] fn main() -> [[builtin(position)]] vec4<f32> {
-    // Expected Error: vec4 should be vec4<f32>
-    return vec4(0.0, 0.0, 0.0, 1.0);
-  }`,
+    _errorLine: 5,
+    _code: `
+      // 頂点シェーダー 👩‍💻
+      [[stage(vertex)]] fn main() -> [[builtin(position)]] vec4<f32> {
+        // Expected Error: vec4 should be vec4<f32>
+        return vec4(0.0, 0.0, 0.0, 1.0);
+      }`,
   },
 ];
 
@@ -65,11 +65,11 @@ g.test('compilationInfo_returns')
   )
   .cases(kAllShaderSources)
   .fn(async t => {
-    const { codeFn, valid } = t.params;
+    const { _code, valid } = t.params;
 
     const shaderModule = t.expectGPUError(
       'validation',
-      () => t.device.createShaderModule({ code: codeFn() }),
+      () => t.device.createShaderModule({ code: _code }),
       !valid
     );
 
@@ -105,10 +105,10 @@ g.test('line_number_and_position')
   )
   .cases(kInvalidShaderSources)
   .fn(async t => {
-    const { codeFn, errorLine } = t.params;
+    const { _code, _errorLine } = t.params;
 
     const shaderModule = t.expectGPUError('validation', () =>
-      t.device.createShaderModule({ code: codeFn() })
+      t.device.createShaderModule({ code: _code })
     );
 
     const info = await shaderModule.compilationInfo();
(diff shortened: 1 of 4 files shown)


src/webgpu/api/operation/shader_module/compilation_info.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+        if (message.lineNum == 4) {
+          foundAppropriateError = true;
+
+          // Various backends may choose to report the error at different positions within the line,
+          // so it's difficult to meaningfully validate them.
+          break;
+        }
+      }
+    }
+    t.expect(foundAppropriateError, "Expected to find an error which corresponded with the erronious line");
```
kainino0x, line 113:
```suggestion
    t.expect(foundAppropriateError, "Expected to find an error which corresponded with the erroneous line");

```

src/webgpu/api/operation/shader_module/compilation_info.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+          code: BAD_SHADER_SOURCE,
+        })
+    );
+
+    const info = await shaderModule.compilationInfo();
+
+    for (const message of info.messages) {
+      // Any offsets and lengths should reference valid spans of the shader code.
+      t.expect(message.offset < BAD_SHADER_SOURCE.length, "Message offset should be within the shader source");
+      t.expect(message.offset + message.length < BAD_SHADER_SOURCE.length, "Message offset and length should be within the shader source");
```
kainino0x, line 136:
```suggestion
      t.expect(message.offset <= BAD_SHADER_SOURCE.length, "Message offset should be within the shader source");
      t.expect(message.offset + message.length <= BAD_SHADER_SOURCE.length, "Message offset and length should be within the shader source");
```
Should be `<=` I think?

src/webgpu/api/operation/shader_module/compilation_info.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+    const info = await shaderModule.compilationInfo();
+
+    let foundAppropriateError = false;
+    for (const message of info.messages) {
+      if (message.type == 'error') {
+        // Some backends may not be able to indicate a precise location for the error. In those
+        // cases a line and position of 0 should be reported.
+        if (message.lineNum == 0) {
+          foundAppropriateError = true;
+          t.expect(message.linePos == 0, "GPUCompilationMessages that don't report a line number should not report a line position.")
```
kainino0x, line 99: nit: Test the converse as well, e.g.:
```ts
t.expect((message.lineNum === 0) === (message.lineNum === 0));
```
(outside of the `if (message.lineNum == 0)`)

src/webgpu/api/operation/shader_module/compilation_info.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+export const g = makeTestGroup(GPUTest);
+
+const BAD_SHADER_SOURCE = `
+  [[stage(vertex)]] fn main() -> [[builtin(position)]] vec4<f32> {
+    // Expected Error: vec4 should be vec4<f32>
+    return vec4(0.0, 0.0, 0.0, 1.0);
+  }
+`;
+
+g.test('compilationInfo_valid_shader')
```
kainino0x, line 18: nit: File is already called `compilation_info`, so probably no need to include `compilationInfo` in the test name (even though it's more specifically referring to `.compilationInfo()`).

kainino0x (approved):
Will need an `npm run fix`, but right now is also failing typechecks because the types aren't complete. Would you be willing to try updating the types so we have more people who are familiar with it? The process is documented here:
https://github.com/gpuweb/types#how-to-update-these-types

src/webgpu/api/operation/shader_module/compilation_info.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+ShaderModule CompilationInfo tests.
+`;
+
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { GPUTest } from '../../../gpu_test.js';
+import { assert } from '../../../../common/framework/util/util.js';
+
+export const g = makeTestGroup(GPUTest);
+
+const BAD_SHADER_SOURCE = `
```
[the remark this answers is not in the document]
toji: Refactored this into a list of sources used for the cases, using my newfound knowledge from the CTS-a-thon. :)

toji:
Addressed all feedback and ran `npm run fix`. Will work on the types next.

toji:
Type roll at https://github.com/gpuweb/types/pull/73

src/webgpu/api/operation/shader_module/compilation_info.spec.ts
```diff
@@ -1,156 +1,184 @@
...
+import { assert } from '../../../../common/framework/util/util.js';
+import { GPUTest } from '../../../gpu_test.js';
+
+export const g = makeTestGroup(GPUTest);
+
+const kValidShaderSources = [
+  {
+    valid: true,
+    unicode: false,
+    codeFn: () => `
```
kainino0x, line 15: You can start a parameter name with `_` to make it "private" (not a part of the "identity" of the case). Here I think you can replace the function with the string, i.e. ``_code: `...` ``

(All of the parameters in this list _could_ be private, except that then the cases would not have distinguishable IDs. Using valid+unicode as the ID makes sense.)

src/webgpu/api/operation/shader_module/compilation_info.spec.ts
```diff
@@ -1,156 +1,184 @@
...
+  [[stage(vertex)]] fn main() -> [[builtin(position)]] vec4<f32> {
+    return vec4<f32>(0.0, 0.0, 0.0, 1.0);
+  }`,
```
kainino0x, line 18: nit: Nicer to indent this to the right level relative to the surrounding JS (even though that affects the string).

src/webgpu/api/operation/shader_module/compilation_info.spec.ts
```diff
@@ -1,156 +1,184 @@
...
+  },
+];
+
+const kInvalidShaderSources = [
+  {
+    valid: false,
+    unicode: false,
+    errorLine: 4,
```
kainino0x, line 35: errorLine can also be private.

src/webgpu/api/operation/shader_module/compilation_info.spec.ts
```diff
@@ -1,156 +1,184 @@
...
+    codeFn: () => `
+  [[stage(vertex)]] fn main() -> [[builtin(position)]] vec4<f32> {
+    return vec4<f32>(0.0, 0.0, 0.0, 1.0);
+  }`,
+  },
+  {
+    valid: true,
+    unicode: true,
+    codeFn: () => `
+  // 頂点シェーダー
```
kainino0x, line 24: If we want to be tougher here, we should include a character which is encoded with a UTF-16 surrogate pair, like `🔺`,  or even better a grapheme cluster with at least one codepoint encoded with UTF-16 surrogate pairs, like `👩‍💻`.
(Can be plugged into https://r12a.github.io/app-conversion/ to inspect the structure)

src/webgpu/api/operation/shader_module/compilation_info.spec.ts
toji: Oh, good! That's why I ended up going with the function in the first place, so I'm glad to see there's a better solution in place for it.

src/webgpu/api/operation/shader_module/compilation_info.spec.ts
toji: 👩‍💻 seems appropriate, (and takes a whole pile a code points.) I'll add it to the comment!

src/webgpu/api/operation/shader_module/compilation_info.spec.ts
toji: Hm... thought I had. Done!

kainino0x (approved):

PR #550, opened 2021-05-04
docs: nits, add pointer to a good reference test

kainino0x:


Kangz (approved):
Yay my tests is the "good reference" one :D

PR #553, opened 2021-05-04
Update variable_and_const.spec.ts shader to match current WGSL spec

zoddicus:


zoddicus:
@kainino0x PTAL

src/webgpu/shader/validation/variable_and_const.spec.ts
```diff
@@ -137,12 +137,37 @@ g.test('v_0038')
...
-      [[location(0)]] var<${storageClass}> a : ${type} = ${type}();
+    var code;
+    if (`${storageClass}` === 'in') {
+      code = `
+        struct MyInputs {
+          [[location(0)]] a : ${type};
+        };
+
+        [[stage(vertex)]]
+        fn main(inputs : MyInputs) {
```
jrprice, line 148: The WGSL spec now requires that all vertex shaders return at least `[[builtin(position)]]`.

src/webgpu/shader/validation/variable_and_const.spec.ts
```diff
@@ -137,12 +137,37 @@ g.test('v_0038')
...
+        }
+      `;
+    } else if (`${storageClass}` === 'out') {
+      code = `
+        struct MyOutputs {
+          [[location(0)]] a : ${type};
```
jrprice, line 154: As above, this struct needs to include a position builtin.

src/webgpu/shader/validation/variable_and_const.spec.ts
```diff
@@ -137,12 +137,37 @@ g.test('v_0038')
...
+        fn main() -> MyOutputs {
+          return MyOutputs();
         }
       `;
+    } else {
+      code = `
+      [[location(0)]] var<${storageClass}> a : ${type} = ${type}();
+
+      [[stage(vertex)]]
+      fn main() {
```
jrprice, line 167: As above, needs to return position builtin.

src/webgpu/shader/validation/variable_and_const.spec.ts
jrprice: Actually, there's probably no reason this test needs to use a vertex shader, so could just make these all fragment shaders to keep things simpler.

src/webgpu/shader/validation/variable_and_const.spec.ts
```diff
@@ -137,12 +137,37 @@ g.test('v_0038')
...
+
         [[stage(vertex)]]
-        fn main() {
+        fn main() -> MyOutputs {
+          return MyOutputs();
         }
       `;
+    } else {
+      code = `
+      [[location(0)]] var<${storageClass}> a : ${type} = ${type}();
```
jrprice, line 164: Presumably this is just for the `private` storage class case, in which case the `[[location(0)]]` decoration is not valid here.

src/webgpu/shader/validation/variable_and_const.spec.ts
zoddicus: Converted to fragment

src/webgpu/shader/validation/variable_and_const.spec.ts
zoddicus: Converted to fragment

src/webgpu/shader/validation/variable_and_const.spec.ts
zoddicus: Done

src/webgpu/shader/validation/variable_and_const.spec.ts
zoddicus: Converted to fragment

jrprice (approved):

kainino0x (approved):

kainino0x:
Two tiny linting failures:
```
Error:   140:5  error  Unexpected var, use let or const instead  no-var
Error:   169:8  error  Insert `;`                                prettier/prettier
```
The former can be fixed manually and the latter with `npm run fix`. You can run the full CI locally with `npm test`.

PR #554, opened 2021-05-04
Reviewed createBindGroup spec

toji:
- Added descriptions to all tests
- Fixed naming where appropriate
- Added more test cases to the viewDimensions test.

<hr>

### DIFF
--- a/src/webgpu/api/validation/createBindGroup.spec.ts
+++ b/src/webgpu/api/validation/createBindGroup.spec.ts
@@ -1,4 +1,8 @@
-export const description = `createBindGroup validation tests.`;
+export const description = `
+  createBindGroup validation tests.
+
+  TODO: Ensure sure tests cover all createBindGroup validation rules.
+`;
 
 import { poptions, params } from '../../../common/framework/params_builder.js';
 import { makeTestGroup } from '../../../common/framework/test_group.js';
@@ -232,8 +236,9 @@ g.test('texture_must_have_correct_dimension')
     - Test for every GPUTextureViewDimension`
   )
   .cases(poptions('viewDimension', kTextureViewDimensions))
+  .subcases(() => poptions('dimension', kTextureViewDimensions))
   .fn(async t => {
-    const { viewDimension } = t.params;
+    const { viewDimension, dimension } = t.params;
     const bindGroupLayout = t.device.createBindGroupLayout({
       entries: [
         {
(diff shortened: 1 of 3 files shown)


src/webgpu/api/validation/createBindGroup.spec.ts
```diff
@@ -204,49 +217,55 @@ g.test('texture_must_have_correct_component_type')
...
+      const viewDescriptor = { dimension, arrayLayerCount };
 
-  t.expectValidationError(() => {
-    t.device.createBindGroup({
-      entries: [{ binding: 0, resource: t.device.createTexture(badDescriptor).createView() }],
-      layout: bindGroupLayout,
-    });
+      t.expectValidationError(() => {
+        t.device.createBindGroup({
+          entries: [{ binding: 0, resource: texture.createView(viewDescriptor) }],
```
kainino0x, line 254: This expectValidationError currently captures multiple calls (createBindGroup and createView). Need to move the createView outside so we know we are testing the right thing. (Also I think createView might actually fail in some cases, not positive)

src/webgpu/api/validation/createBindGroup.spec.ts
```diff
@@ -25,60 +22,69 @@ function clone<T extends GPUTextureDescriptor>(descriptor: T): T {
 
 export const g = makeTestGroup(ValidationTest);
 
-g.test('binding_count_mismatch').fn(async t => {
-  const bindGroupLayout = t.device.createBindGroupLayout({
-    entries: [{ binding: 0, visibility: GPUShaderStage.COMPUTE, buffer: { type: 'storage' } }],
-  });
+g.test('binding_count_mismatch')
+  .desc('Test that the number of entries must match the number of entries in the BindGroupLayout.')
```
kainino0x, line 26: I think it would be ideal to slightly refactor these two tests (this one and the next one) to parameterize between the control case and bad case. This makes it possible by construction to see exactly how the two cases differ.

src/webgpu/api/validation/createBindGroup.spec.ts
toji: Moved `createView()` outside the validation error check, but now every case of the test fails because they all attempt to create every view variant to test compatibility and Dawn doesn't support 1D or 3D views yet. That's a little discouraging, but I'm not sure what a better pattern would be. I could check and see if `createView()` fails and skip that iteration of loop, since this isn't intended to test `createView()`? I'd appreciate your thoughts.

src/webgpu/api/validation/createBindGroup.spec.ts
toji: Done, please make sure that this is what you had in mind.

src/webgpu/api/validation/createBindGroup.spec.ts
kainino0x: Ah, true. How about replacing the loop with subcases? The setup part of the test will end up running more, but that seems fine.

kainino0x (commented):
oops, forgot to mark reviewed. (last change lgtm)

toji:
Is there a reason you didn't actually mark it approved? Can't be merged without it.

kainino0x (approved):
Thought you were going to revise the for-loop over the texture view formats first, but that's not strictly needed to land.

src/webgpu/api/validation/createBindGroupLayout.spec.ts
```diff
@@ -1,7 +1,7 @@
 export const description = `
 createBindGroupLayout validation tests.
 
-TODO: update for new binding structure, remove eslint-disable.
+TODO: update for new binding structure.
```
kainino0x, line 4: I just realized this TODO should be obsolete, I did it when fixing deprecations.

src/webgpu/api/validation/createBindGroup.spec.ts
toji: Oh, sorry I missed this. Github didn't include it in the page auto-refresh for some reason. I can do that.

src/webgpu/api/validation/createBindGroup.spec.ts
```diff
@@ -1,8 +1,4 @@
-export const description = `
-createBindGroup validation tests.
-
-TODO: review existing tests, write descriptions, and make sure tests are complete.
```
kainino0x, line 4: Just to double check before we remove this, do you think this area is completely tested aside from outstanding TODOs?

kainino0x: I.e. do we test every validation rule in createBindGroup() - I haven't reviewed the spec, so not sure. (Sorry I forgot about this until now)

toji: Added the todo back indicating I still need to check to ensure all validation rules are covered.

src/webgpu/api/validation/createBindGroup.spec.ts
toji: And done.

PR #555, opened 2021-05-04
Add compute,basic,large_dispatch test.

kenrussell:
Test compute dispatches with a single large parameter along one of the
X, Y, or Z workgroup grid dimensions. Within this, also test a few
workgroup sizes - which increase runtime significantly since they
increase the size along all 3 dimensions simultaneously.

Add and document helpers for ensuring a buffer's contents are all a
single value.

<hr>

### DIFF
--- a/src/webgpu/api/operation/compute/basic.spec.ts
+++ b/src/webgpu/api/operation/compute/basic.spec.ts
@@ -116,8 +116,7 @@ Test reasonably-sized large dispatches (see also stress tests).
     const data = new Uint32Array([val]);
 
     const wgSize = t.params.workgroupSize;
-    const bufferSize =
-      Uint32Array.BYTES_PER_ELEMENT * t.params.dispatchSize * wgSize;
+    const bufferSize = Uint32Array.BYTES_PER_ELEMENT * t.params.dispatchSize * wgSize;
     const dst = t.device.createBuffer({
       size: bufferSize,
       usage: GPUBufferUsage.COPY_SRC | GPUBufferUsage.STORAGE,
--- a/src/webgpu/api/operation/compute/basic.spec.ts
+++ b/src/webgpu/api/operation/compute/basic.spec.ts
@@ -75,39 +75,17 @@ Test reasonably-sized large dispatches (see also stress tests).
   .cases(
     params()
       .combine(
-        poptions('dispatchSize', [
-          // Reasonably-sized powers of two
-          256,
-          512,
-          1024,
-          2048,
-          // Some stranger larger sizes
-          315,
-          628,
-          1053,
-          2179,
-        ] as const)
+        // Reasonably-sized powers of two, and some stranger larger sizes.
+        poptions('dispatchSize', [256, 512, 1024, 2048, 315, 628, 1053, 2179] as const)
       )
       .combine(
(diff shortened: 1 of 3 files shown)


src/webgpu/api/operation/compute/basic.spec.ts
```diff
@@ -64,5 +65,110 @@ g.test('memcpy').fn(async t => {
...
+            [[group(0), binding(0)]] var<storage> dst : [[access(read_write)]] OutputBuffer;
+
+            [[stage(compute), workgroup_size(${wgSize}, ${wgSize}, ${wgSize})]]
+            fn main(
+              [[builtin(global_invocation_id)]] GlobalInvocationID : vec3<u32>
+            ) {
+              // This indexing only works because this is conceptually a 1D array
+              // which spans a single 3D axis.
+              dst.value[GlobalInvocationID.x + GlobalInvocationID.y + GlobalInvocationID.z] = ${val}u;
+              return;
```
kainino0x, line 147: nit: Return can be omitted

src/webgpu/api/operation/compute/basic.spec.ts
```diff
@@ -64,5 +65,110 @@ g.test('memcpy').fn(async t => {
...
+    const val = 0x01020304;
+    const data = new Uint32Array([val]);
+
+    const src = t.device.createBuffer({
+      mappedAtCreation: true,
+      size: Uint32Array.BYTES_PER_ELEMENT,
+      usage: GPUBufferUsage.COPY_DST | GPUBufferUsage.STORAGE,
+    });
+    new Uint32Array(src.getMappedRange()).set(data);
+    src.unmap();
```
kainino0x, line 120: Unused buffer?

src/webgpu/gpu_test.ts
```diff
@@ -416,6 +442,72 @@ got [${failedByteActualValues.join(', ')}]`;
     return undefined;
   }
 
+  checkSingleValueBuffer(
+    actual: TypedArrayBufferView,
+    exp: TypedArrayBufferView,
+    tolerance: number | ((i: number) => number) = 0
+  ): string | undefined {
```
kainino0x, line 449: I'm sure we'll refactor all these helpers eventually anyway, but maybe this could be written on top of checkBufferFn to make it duplicate less?

kenrussell: We discussed this online and unfortunately checkBufferFn doesn't generate the error messages itself - the function passed to it does - so it doesn't share a lot of code.

A better solution would be to carefully refactor checkBuffer to take functions that index into the actual and expected arrays, or similar. Per discussion, it sounds like we could do this in a follow-up.

src/webgpu/api/operation/compute/basic.spec.ts
kenrussell: Yes, thanks - fixed.

src/webgpu/api/operation/compute/basic.spec.ts
kenrussell: Thanks, fixed.

kainino0x (approved):

kainino0x (approved):

PR #557, opened 2021-05-05
Update texture zero init tests for 3D textures

austinEng:
I'm not 100% sure if this works since 3D textures are still being implemented. Running with unsafe_apis doesn't yield validation errors though.

<hr>

### DIFF
--- a/src/webgpu/api/operation/resource_init/check_texture/by_sampling.ts
+++ b/src/webgpu/api/operation/resource_init/check_texture/by_sampling.ts
@@ -1,4 +1,4 @@
-import { assert } from '../../../../../common/framework/util/util.js';
+import { assert, unreachable } from '../../../../../common/framework/util/util.js';
 import {
   EncodableTextureFormat,
   kEncodableTextureFormatInfo,
@@ -17,7 +17,7 @@ export const checkContentsBySampling: CheckContents = (
   state,
   subresourceRange
 ) => {
-  assert(params.dimension === '2d');
+  assert(params.dimension !== '1d');
   assert(params.format in kEncodableTextureFormatInfo);
   const format = params.format as EncodableTextureFormat;
   const rep = kTexelRepresentationInfo[format];
@@ -43,20 +43,27 @@ export const checkContentsBySampling: CheckContents = (
 
     const _xd = '_' + params.dimension;
     const _multisampled = params.sampleCount > 1 ? '_multisampled' : '';
+    const texelIndexExpresion =
+      params.dimension === '2d'
+        ? 'vec2<i32>(GlobalInvocationID.xy)'
+        : params.dimension === '3d'
+        ? 'vec3<i32>(i32(GlobalInvocationID.x), i32(GlobalInvocationID.y), constants.slice)'
+        : unreachable();
     const computePipeline = t.device.createComputePipeline({
       compute: {
         entryPoint: 'main',
         module: t.device.createShaderModule({
           code: `
             [[block]] struct Constants {
-              [[offset(0)]] level : i32;
+              level : i32;
+              slice: i32;
             };
 
             [[group(0), binding(0)]] var<uniform> constants : Constants;
             [[group(0), binding(1)]] var myTexture : texture${_multisampled}${_xd}<${shaderType}>;
 
             [[block]] struct Result {
-              [[offset(0)]] values : [[stride(4)]] array<${shaderType}>;
+              values : [[stride(4)]] array<${shaderType}>;
             };
             [[group(0), binding(3)]] var<storage> result : [[access(read_write)]] Result;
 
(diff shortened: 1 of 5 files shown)


src/webgpu/api/operation/resource_init/check_texture/by_sampling.ts
```diff
@@ -79,19 +84,33 @@ export const checkContentsBySampling: CheckContents = (
...
       const resultBuffer = t.device.createBuffer({
         size: byteLength,
         usage: GPUBufferUsage.STORAGE | GPUBufferUsage.COPY_SRC,
       });
 
+      const textureView =
+        params.dimension === '2d'
+          ? texture.createView({
+              baseMipLevel: 0,
+              mipLevelCount: params.mipLevelCount,
```
kainino0x, line 105: is `mipLevelCount` the mipLevelCount of the`texture`? If so, this is the default (and I think it would be clearer to omit - right now, I can't tell whether this is supposed to be able to differ from the `texture`). In fact, all of these arguments in both branches are the defaults, except for `baseArrayLayer: slice`.

src/webgpu/util/texture/subresource.ts
```diff
@@ -43,46 +46,33 @@ export class SubresourceRange {
 
   *each(): Generator<{ level: number; slice: number }> {
     for (let level = this.mipRange.begin; level < this.mipRange.end; ++level) {
-      for (let slice = this.sliceRange.begin; slice < this.sliceRange.end; ++slice) {
+      const end =
+        this.dimension === '3d'
+          ? this.depthOrArrayLayerCount >> level
+          : this.depthOrArrayLayerCount;
+      for (let slice = this.sliceRange.begin; slice < Math.min(end, this.sliceRange.end); ++slice) {
```
kainino0x, line 53: I think this function is supposed to iterate over the subresources of a resource, but 3D textures just have one subresource per level (not per level per slice). Seems like this is returning a subresource for each slice as well? See also [1]s

> Each such texture subresource is uniquely identified by a texture, mipmap level, and (for 2d textures only) array layer, and aspect.

src/webgpu/util/texture/subresource.ts
```diff
@@ -43,46 +46,33 @@ export class SubresourceRange {
...
         yield { level, slice };
       }
     }
   }
 
   *mipLevels(): Generator<{ level: number; slices: Generator<number> }> {
```
kainino0x, line 59: Not sure what "slices" should mean here...

src/webgpu/api/operation/resource_init/texture_zero.spec.ts
```diff
@@ -251,7 +244,9 @@ export class TextureZeroInitTest extends GPUTest {
   *iterateInitializedSubresources(): Generator<SubresourceRange> {
     const uninitialized: boolean[][] = new Array(this.p.mipLevelCount);
     for (let level = 0; level < uninitialized.length; ++level) {
-      uninitialized[level] = new Array(this.p.sliceCount);
+      uninitialized[level] = new Array(
+        this.p.dimension === '3d' ? this.p.sliceCount >> level : this.p.sliceCount
+      );
```
kainino0x, line 249: [1]

src/webgpu/api/operation/resource_init/texture_zero.spec.ts
```diff
@@ -261,10 +256,14 @@ export class TextureZeroInitTest extends GPUTest {
...
-            sliceRange: { begin: slice, count: 1 },
-          });
+          yield new SubresourceRange(
+            {
+              mipRange: { begin: level, count: 1 },
+              sliceRange: { begin: slice, count: 1 },
+            },
+            this.p.sliceCount,
+            this.p.dimension
+          );
```
kainino0x, line 266: [1]

austinEng:
Yea, I realized the thing about subresources as I was implementing the lazy clear in Dawn. Instead of shoving everything into this one test, it probably makes sense to have a separate zero init test for 1D/3D textures because they work differently and have more restrictions than 2D - no depth/stencil, etc.

Richard-Yunchao:
> Yea, I realized the thing about subresources as I was implementing the lazy clear in Dawn. Instead of shoving everything into this one test, it probably makes sense to have a separate zero init test for 1D/3D textures because they work differently and have more restrictions than 2D - no depth/stencil, etc.

Totally agreed. Maybe you need to have significant change for this PRs (or via another PR). For example, lots of tests use kEncoderableTextureFormatInfo, which include depth/stencil formats. And depth/stencil formats are invalid for 3D textures.

In addition, I don't like term `slice` throughout CTS and Dawn projects. If it is a layer, just say layer. If it is a depth, just say depth. The only exception is that if you really want a single term to combine depth and layer. 2 reasons of not using the term `slice`: 1) as you know, a depth slice (of a 3D texture) is not a standalone subresource, you can't use it to calculate subresource index. However, a layer (of a 2D texture) is a standalone subresource. So assumptions and computations based on term `slice`  in 2D textures may not apply to 3D textures. 2) D3D uses term `slice` to refer to all similar things, including mip slice, layer slice, and plane/aspect slice. A slice in D3D is (very likely) a subresource. It is confusing that Dawn/CTS excludes mipmaps and aspects from the term `slice` if someone is familiar with D3D.

Richard-Yunchao:
I forgot to add the link about uses of term `slice` in D3D12 (and the definition about subresource) I mentioned above, including mip slice, array slice, and plane slice: https://docs.microsoft.com/en-us/windows/win32/direct3d12/subresources.

austinEng:
Closing in favor of #572

PR #561, opened 2021-05-12
Use more subcases

kainino0x:
Use subcases more liberally, specifically for tests that had WAY too
many cases (and are spamming the output when run via WPT).

<hr>

### DIFF
--- a/src/webgpu/api/operation/rendering/blending.spec.ts
+++ b/src/webgpu/api/operation/rendering/blending.spec.ts
@@ -127,12 +127,21 @@ g.test('GPUBlendComponent')
     - dstFactor= {...all GPUBlendFactors}
     - operation= {...all GPUBlendOperations}`
   )
-  .cases(poptions('component', ['color', 'alpha'] as const))
-  .subcases(() =>
-    params()
-      .combine(poptions('operation', kBlendOperations))
+  .cases(
+    params() //
+      .combine(poptions('component', ['color', 'alpha'] as const))
       .combine(poptions('srcFactor', kBlendFactors))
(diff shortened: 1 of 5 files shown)


src/webgpu/api/operation/rendering/blending.spec.ts
```diff
@@ -121,33 +125,33 @@ g.test('GPUBlendComponent')
     - component= {color, alpha} - whether to test blending the color or the alpha component.
     - srcFactor= {...all GPUBlendFactors}
     - dstFactor= {...all GPUBlendFactors}
-    - operation= {...all GPUBlendOperations}
-  `)
-  .cases(
```
kainino0x, line 126: Maybe this change won't be necessary once we don't collect ALL stdout into one big file in our infra.

src/webgpu/api/operation/resource_init/texture_zero.spec.ts
```diff
@@ -532,7 +532,7 @@ const checkContentsImpl: { [k in ReadMethod]: CheckContents } = {
 export const g = makeTestGroup(TextureZeroInitTest);
 
 g.test('uninitialized_texture_is_zero')
-  .params(paramsBuilder)
```
kainino0x, line 535: Maybe some parts of this should be cases and some should be subcases?

austinEng (approved):

src/webgpu/api/operation/resource_init/texture_zero.spec.ts
austinEng: Oh mm yea I think we need some to be cases. Can you leave this one out for now? Probably need to revisit this test.

PR #563, opened 2021-05-17
Check features support when requesting device

haoxli:
The features may be not supported, we should check them before
requesting device, otherwise the GPU process will lost due to
the failure of device creation.

Bug: chromium:1175091

<hr>

### DIFF
--- a/src/webgpu/api/operation/adapter/requestDevice.spec.ts
+++ b/src/webgpu/api/operation/adapter/requestDevice.spec.ts
@@ -1,0 +1,16 @@
+export const description = `
+Tests for GPUAdapter.requestDevice.
+
+TODO:
+- descriptor is {null, undefined}
+- descriptor with the combinations of features and limits
+`;
+
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { GPUTest } from '../../../gpu_test.js';
+
+export const g = makeTestGroup(GPUTest);
+
+g.test('request_default_after_error')
+  .desc(`Request default device after a failed request with unsupported features.`)
+  .unimplemented();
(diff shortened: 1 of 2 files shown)


kainino0x:
If I understand correctly, this changes this from a DOMException(OperationError) to a FeatuuresNotSupported exception. However both exceptions will still be caught by the same `try` block:
https://github.com/gpuweb/cts/blob/main/src/webgpu/util/device_pool.ts#L139

> otherwise the GPU process will lost due to the failure of device creation.

This should not be the case. `requestDevice` should simply reject, the GPU process should not exit.

haoxli:
Although both exceptions are caught by the same `try` block, they are still different. The OperationError is caught when requesting device failed,  the FeaturesNotSupported exception is before the requesting. For a optional feature which may be not supported, we'd better check its availability in advance.

For the case of requestDevice reject but GPU process does not lost, I added a test in adapter to cover it.

kainino0x:
Bad things (like GPU process crash) should never happen when we requestDevice with invalid capabilities - remember that any untrusted website on the internet can use the API in any way. I don't think we need the CTS to be "more polite" in this way unless we're trying to specifically detect if _other_ exceptions (other than checking capabilities) come from requestDevice.

haoxli:
Got it, close this PR.

Kangz:
Hao made a CL to fix the GPU process crash in Chromium, but independently of this, WDYT of this PR? It seems ok to be nicer when requesting devices with optional features.

kainino0x:
> It seems ok to be nicer when requesting devices with optional features.

It's fine, but right now I think it's extra code for no gain. However, there _is_ a possible benefit to this: we can detect erroneous exceptions from requestDevice - with this check, it should never throw per spec.

In the `try`/`catch` block of `DescriptorToHolderMap.getOrCreate`, we should only `SkipTestCase` if the exception is a FeaturesNotSupported. We can insert this at the top of the `catch` block to re-throw if any other error occurs:
```ts
if (!(ex instanceof FeaturesNotSupported)) throw ex;
```

kainino0x (approved):
Sorry I missed your update, LGTM!

PR #565, opened 2021-05-21
Simplify TestBuilder, generalize ParamArgument, rename CaseParams

kainino0x:
- the TestBuilder types now just have one "params" type (i.e. the value
  passed to the test function), instead of tracking cases and subcases
  params types separately.
- this file already has to use type system escape hatches, so just lean
  into it and remove all of the generics on TestBuilder and
  RunCaseSpecific as they're not really providing any real benefit
  (except to verify TestBuilder implements the associated interfaces)
- replace ParamArgument with JSONWithUndefined, enforce CaseParams in .cases()/.subcases()
- rename CaseParams to TestParams ("the parameters passed to a test (function)")

<hr>

### DIFF
--- a/src/common/framework/test_group.ts
+++ b/src/common/framework/test_group.ts
@@ -111,13 +111,13 @@ class TestGroup<F extends Fixture> implements TestGroupBuilder<F> {
 interface TestBuilderWithName<F extends Fixture> extends TestBuilderWithCases<F, {}> {
   desc(description: string): this;
   /** @deprecated use cases() and/or subcases() instead */
-  params<NewP extends {}>(specs: Iterable<NewP>): TestBuilderWithSubcases<F, NewP>;
-  cases<NewP extends {}>(specs: Iterable<NewP>): TestBuilderWithCases<F, NewP>;
+  params<NewP extends CaseParams>(specs: Iterable<NewP>): TestBuilderWithSubcases<F, NewP>;
+  cases<NewP extends CaseParams>(specs: Iterable<NewP>): TestBuilderWithCases<F, NewP>;
 }
 
 interface TestBuilderWithCases<F extends Fixture, P extends {}>
   extends TestBuilderWithSubcases<F, P> {
-  subcases<SubP extends {}>(
+  subcases<SubP extends CaseParams>(
     specs: (_: P) => Iterable<SubP>
   ): TestBuilderWithSubcases<F, Merged<P, SubP>>;
 }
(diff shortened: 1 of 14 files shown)


src/common/framework/test_group.ts
```diff
@@ -107,36 +108,36 @@ class TestGroup<F extends Fixture> implements TestGroupBuilder<F> {
...
 }
 
-interface TestBuilderWithName<F extends Fixture> extends TestBuilderWithParams<F, {}> {
+interface TestBuilderWithName<F extends Fixture> extends TestBuilderWithCases<F, {}> {
   desc(description: string): this;
   /** @deprecated use cases() and/or subcases() instead */
-  params<NewP extends {}>(specs: Iterable<NewP>): TestBuilderWithParams<F, NewP>;
-  cases<NewP extends {}>(specs: Iterable<NewP>): TestBuilderWithParams<F, NewP>;
+  params<NewP extends {}>(specs: Iterable<NewP>): TestBuilderWithSubcases<F, NewP>;
+  cases<NewP extends {}>(specs: Iterable<NewP>): TestBuilderWithCases<F, NewP>;
```
austinEng: Shouldn't these return the same type? I only expect it to return TestBuilderWithSubcases when you call .subcases 

or maybe it's because if you're using deprecated .params, you shouldn't use the new .subcases

src/common/framework/test_group.ts
```diff
@@ -194,29 +195,25 @@ class TestBuilder<F extends Fixture, P extends {}, SubP extends {}> {
...
-    const newSelf = (this as unknown) as TestBuilder<F, NewP, SubP>;
-    newSelf.caseParams = Array.from(casesIterable);
-
-    return newSelf;
+    this.caseParams = Array.from(casesIterable);
+    return this;
   }
 
-  subcases<NewSubP extends {}>(specs: (_: P) => Iterable<NewSubP>): TestBuilder<F, P, NewSubP> {
+  subcases(specs: (_: {}) => Iterable<{}>): TestBuilder {
```
austinEng, line 208: does this mean we lose the type information on `params` when doing

.subcases(params => {
})

?
I found it kind of nice

austinEng (approved):

src/common/framework/test_group.ts
kainino0x: > or maybe it's because if you're using deprecated .params, you shouldn't use the new .subcases

yeah, it's not necessary to do it this way, but this is why.

src/common/framework/test_group.ts
kainino0x: It still works, because the test code sees it as a `TestBuilderWithCases<F, P>`, not as a `TestBuilder`.

PR #566, opened 2021-05-22
Add {Case,Subcase}ParamsBuilder to replace .cases().subcases()

kainino0x:
Baking the split between cases and subcases into the params builder
should make it more ergonomic to use: especially in cases where the
cases/subcases are not defined inline, this means the cases and subcases
don't have to be defined as two separate things (an iterable and a
function), and instead can be defined as a single SubcaseParamsBuilder.

It also makes it easy to change whether a parameter is a case or subcase parameter.

<hr>

    https://chromium-review.googlesource.com/c/chromium/src/+/2938946
    
### DIFF
--- a/src/common/framework/params_builder.ts
+++ b/src/common/framework/params_builder.ts
@@ -1,154 +1,312 @@
-import {
-  TestParams,
-  TestParamsIterable,
-  FlattenUnionOfInterfaces,
-  Merged,
-  mergeParams,
-  publicParamsEquals,
-} from './params_utils.js';
-import { ResolveType, UnionToIntersection } from './util/types.js';
+import { Merged, mergeParams } from './params_utils.js';
 
-/** Conditionally chooses between two types depending on whether T is a union. */
-type CheckForUnion<T, TErr, TOk> = [T] extends [UnionToIntersection<T>] ? TOk : TErr;
+// ================================================================
+// "Public" ParamsBuilder API / Documentation
+// ================================================================
 
-/** Conditionally chooses a type (or void) depending on whether T is a string. */
-type CheckForStringLiteralType<T, TOk> = string extends T ? void : CheckForUnion<T, void, TOk>;
-
-/* eslint-disable-next-line @typescript-eslint/no-unused-vars */
-function typeAssert<T extends 'pass'>() {}
-{
-  type Test<T, U> = [T] extends [U]
-    ? [U] extends [T]
-      ? 'pass'
-      : { actual: ResolveType<T>; expected: U }
-    : { actual: ResolveType<T>; expected: U };
-
-  type T01 = { a: number } | { b: string };
-  type T02 = { a: number } | { b?: string };
-  type T03 = { a: number } | { a?: number };
-  type T04 = { a: number } | { a: string };
-  type T05 = { a: number } | { a?: string };
-
-  type T11 = { a: number; b?: undefined } | { a?: undefined; b: string };
-
-  type T21 = { a: number; b?: undefined } | { b: string };
-  type T22 = { a: number; b?: undefined } | { b?: string };
-  type T23 = { a: number; b?: undefined } | { a?: number };
-  type T24 = { a: number; b?: undefined } | { a: string };
-  type T25 = { a: number; b?: undefined } | { a?: string };
-  type T26 = { a: number; b?: undefined } | { a: undefined };
-  type T27 = { a: number; b?: undefined } | { a: undefined; b: undefined };
-
-  /* prettier-ignore */ {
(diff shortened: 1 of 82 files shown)


kainino0x:
@austinEng let me know what you think of this idea

kainino0x:
@egalli as well if you happen to want to look at this

egalli (commented):
I like the idea of integrating cases/subcases with the params builder.

That said, I'm not a huge fan of `params2`. Are you using it because `params` was already in use? If so, have you consider using deprecated overloads?

Also, `beginSubcases`, IMHO, hides the fact that the remaining calls are defining subcases. Have you consider using something like `subcases(unit: SubcaseParamsBuilder<CaseP, {}>) => SubcaseParamsBuilder<{}, {}>)): void` instead? I'm using a void return in order to make it harder to accidentally continue defining more cases after starting to define subcases.

kainino0x:
Thanks for taking a look!

> That said, I'm not a huge fan of `params2`. Are you using it because `params` was already in use? If so, have you consider using deprecated overloads?

I used it temporarily while working on it, but now I've picked real names for everything and that's gone.

> Also, `beginSubcases`, IMHO, hides the fact that the remaining calls are defining subcases. Have you consider using something like `subcases(unit: SubcaseParamsBuilder<CaseP, {}>) => SubcaseParamsBuilder<{}, {}>)): void` instead? I'm using a void return in order to make it harder to accidentally continue defining more cases after starting to define subcases.

I thought about doing something like this, but from experience I've found that we keep needing to change which parameters are cases and which are subcases, and this structure makes it easier to change them. Also, generally, it's not very important whether something is a case or subcase, so I think it's okay that they look the same.

I also thought about making the method names slightly distinct (like, have a leading `_` on the subcase methods). Maybe this would be clearer.

(Thinking about that, maybe they don't even really need to be two different builders. Maybe it should be possible to add more cases after adding a subcase. But I think the "lifting" behavior where cases would get lifted above subcases might be too weird. And I'm not even sure the idea is possible once you start getting into it.)

kainino0x:
Yeah, the lifting behavior wouldn't work. The ordering that params are defined in is very important (to the type system and to the way filtering and stuff works). Technically I think it might work, at least at runtime, since we don't allow overwriting params, but it would be too weird.

kainino0x:
@austinEng PTAL. I've split the changes into several commits. IMO you can just skip the big migration commit at the end. It does have some minor consequential changes (I finally migrated all the old .params() to use cases/subcases) but I did a chromium dry run and all seemed good.

kainino0x:
Although you can look at that commit to find samples of how this change actually looks in practice, of course. Including a few cases where kUnitCaseParamsBuilder is used explicitly.

src/webgpu/api/validation/createSampler.spec.ts
```diff
@@ -27,12 +26,16 @@ g.test('lodMinAndMaxClamp')
...
   .desc('test different maxAnisotropy values and combinations with min/mag/mipmapFilter')
-  .params([
-    ...poptions('maxAnisotropy', [-1, undefined, 0, 1, 2, 4, 7, 16, 32, 33, 1024]),
-    { minFilter: 'nearest' as const },
-    { magFilter: 'nearest' as const },
-    { mipmapFilter: 'nearest' as const },
-  ])
+  .params(u =>
+    u //
+      .beginSubcases()
```
austinEng, line 31: Is there a difference between this and `paramsSubcasesOnly`?  Do we need `paramsSubcasesOnly`?

src/common/framework/params_builder.ts
```diff
@@ -1,154 +1,291 @@
...
+  /**
+   * Expands each subcase in `this` into zero or more subcases.
+   *
+   * **Note:** In most situations, `expand` is a simpler and more readable alternative.
+   *
+   * ```text
+   *               this = [     a       ,      b     ,       c       ]
+   * this.map(expander) = [   f(a)           f(b)          f(c)      ]
+   *                    = [[a1, a2, a3] ,    [ b1 ]  ,       []      ]
+   *   flattened result = [ a1, a2, a3  ,      b1                    ]
```
austinEng, line 199: IIUC, I think `a1` in the flattened result is more like `merge(a, a1)` ? THe example makes it look like just a map + flatten, when really it also merges with the existing params.

Also, I don't love the "expandP" name, but don't have better suggestions. Maybe if it's not used very often we can make it more verbose

austinEng (approved):
LGTM - I think?

src/webgpu/api/validation/createSampler.spec.ts
kainino0x: Oh, there is a difference here. In `.params()`, `u` is a `CaseParamsBuilder` which can also be used as a generic combinatorial generator in places other than just the return value of `.params()`'s callback. It is used here on line 33.

PR #568, opened 2021-05-27
Update validation tests of copyImageBitmapToTexture

shaoboyan:
Update copyImageBitmapToTexture validation rules based on the new
CL (https://github.com/gpuweb/gpuweb/pull/1737)

<hr>

### DIFF
--- a/src/webgpu/api/validation/queue/copyToTexture/ImageBitmap.spec.ts
+++ b/src/webgpu/api/validation/queue/copyToTexture/ImageBitmap.spec.ts
@@ -321,7 +321,7 @@ g.test('destination_texture,format')
     });
     t.device.popErrorScope();
 
-    const success = ([...kValidTextureFormatsForCopyIB2T] as string[]).includes(format);
+    const success = (kValidTextureFormatsForCopyIB2T as readonly string[]).includes(format);
 
     t.runTest({ imageBitmap }, { texture: dstTexture }, copySize, success);
   });
--- a/src/webgpu/api/validation/queue/copyToTexture/ImageBitmap.spec.ts
+++ b/src/webgpu/api/validation/queue/copyToTexture/ImageBitmap.spec.ts
@@ -47,10 +47,7 @@ const kValidTextureFormatsForCopyIB2T = [
   'bgra8unorm',
   'bgra8unorm-srgb',
   'rgb10a2unorm',
-  'rgba16float',
-  'rgba32float',
   'rg8unorm',
-  'rg16float',
 ];
 
 function computeMipMapSize(width: number, height: number, mipLevel: number) {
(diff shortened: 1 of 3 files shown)


shaoboyan:
@kainino0x  PTAL!

kainino0x (approved):
LGTM, thanks! Please check the "Author" checklist. Do these tests pass in your local build?

src/webgpu/web_platform/copyToTexture/ImageBitmap.spec.ts
```diff
@@ -339,10 +336,7 @@ g.test('from_canvas')
           'rgba8unorm-srgb',
           'bgra8unorm-srgb',
           'rgb10a2unorm',
-          'rgba16float',
-          'rgba32float',
           'rg8unorm',
-          'rg16float',
         ] as const)
```
kainino0x, line 340: nit: We have three copies of this list. We should deduplicate them into capability_info.ts so we can update them more easily later.

shaoboyan: One issue when I try to move the defination to capability_info.ts as const kValidTextureFormatForCopyE2T= [...] as const.
I couldn't use kValidTextureFormatForCopyE2T.include(format) (format is kAllTextureFormats) in the validation tests. The compile reports type mismatch assignment.
Do you have any suggestions?

shaoboyan:
Yes, these tests passed in my local except for some failures with some depth stencil texture format in destination_texture,format validation test(e.g. stencil8, ), which reports that the enum is not a valid enum for createTexture(Suspect it is my build chromium reason).

src/webgpu/web_platform/copyToTexture/ImageBitmap.spec.ts
kainino0x: You could use something like `const kValidTextureFormatForCopyE2T: read-only GPUTextureFormat[] = ...` if that works (but maybe it will cause other issues, I don't know

If that doesn't work just put in `as` casts wherever needed to fix the problems.

shaoboyan: FYI, I discuss with my collegues working on TFJS for suggestions and we'd like to find a solution. So I'd like to hold the CL for one more day and use tricks `as` later if I cannot get some ideas.

shaoboyan: Finally adopt ([...object] as string[]).includes(...) :p

src/webgpu/api/validation/queue/copyToTexture/ImageBitmap.spec.ts
```diff
@@ -329,19 +317,13 @@ g.test('destination_texture,format')
     const dstTexture = t.device.createTexture({
       size: { width: 1, height: 1, depthOrArrayLayers: 1 },
       format,
-      usage: GPUTextureUsage.COPY_DST,
+      usage: GPUTextureUsage.COPY_DST | GPUTextureUsage.RENDER_ATTACHMENT,
     });
     t.device.popErrorScope();
 
-    const success = kValidTextureFormatsForCopyIB2T.includes(format);
+    const success = ([...kValidTextureFormatsForCopyIB2T] as string[]).includes(format);
```
kainino0x, line 324: This introduces unnecessary runtime cost, can it simply be:

```suggestion
    const success = (kValidTextureFormatsForCopyIB2T as readonly string[]).includes(format);
```

kainino0x (approved):
LGTM

PR #569, opened 2021-06-02
Add validation for creating compute pipeline

jzm-intel:
Add validation test for both createComputePipeline and createComputePipelieAsync.

<hr>

### DIFF
--- a/src/webgpu/api/validation/createComputePipeline.spec.ts
+++ b/src/webgpu/api/validation/createComputePipeline.spec.ts
@@ -1,16 +1,5 @@
 export const description = `
 createComputePipeline and createComputePipelineAsync validation tests.
-
-For createComputePipeline and its async version createComputePipelineAsync,
-each start with a valid descriptor, than for the only one compute stage, make
-following errors:
-- compute stage's module is an invalid object
-- compute stage's module is of other stage, i.e., vertex or gragment
-- stage's entryPoint doesn't exist, including following situation:
-  - the entryPoint in module is / isn't 'main'
-  - the assigned name is different from module's name
-  - the assigned name is empty string
-  - the assigned name contain illegal character
 `;
 
 import { params, poptions } from '../../../common/framework/params_builder.js';
--- a/src/webgpu/api/validation/createComputePipeline.spec.ts
+++ b/src/webgpu/api/validation/createComputePipeline.spec.ts
@@ -1,4 +1,5 @@
 export const description = `
+createComputePipeline and createComputePipelineAsync validation tests.
 
 TODO:
 For createComputePipeline and its async version createComputePipelineAsync,
@@ -115,30 +116,23 @@ g.test('shader_module_stage_must_be_compute')
   .cases(
     params()
       .combine(poptions('isAsync', [true, false]))
-      .combine([
-        {
-          shaderModuleStage: 'compute' as TShaderStage,
-          _success: true,
-        },
-        {
-          shaderModuleStage: 'vertex' as TShaderStage,
-          _success: false,
-        },
-        {
-          shaderModuleStage: 'fragment' as TShaderStage,
-          _success: false,
-        },
-      ])
+      .combine(
+        poptions('shaderModuleStage', [
+          'compute' as TShaderStage,
+          'vertex' as TShaderStage,
+          'fragment' as TShaderStage,
+        ])
+      )
   )
   .fn(async t => {
-    const { isAsync, shaderModuleStage, _success } = t.params;
+    const { isAsync, shaderModuleStage } = t.params;
     const descriptor = {
       compute: {
         module: t.getShaderModule(shaderModuleStage, 'main'),
         entryPoint: 'main',
       },
     };
-    t.doCreateComputePipelineTest(isAsync, _success, descriptor);
+    t.doCreateComputePipelineTest(isAsync, shaderModuleStage === 'compute', descriptor);
   });
 
 g.test('enrty_point_name_must_match')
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/createComputePipeline.spec.ts
```diff
@@ -0,0 +1,164 @@
+export const description = `
+createComputePipeline and createComputePipelineAsync validation tests.
+
+TODO:
+For createComputePipeline and its async version createComputePipelineAsync,
```
Kangz, line 5: We usually use these TODOs to say what work should be done in the file. Everything that's in this TODO is implemented so it doesn't need to appear. Is there something else that should be tested?

src/webgpu/api/validation/createComputePipeline.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+
+class F extends ValidationTest {
+  getShaderModule(
+    shaderStage: TShaderStage = 'compute',
+    entryPoint: string = 'main'
+  ): GPUShaderModule {
+    let code;
+    switch (shaderStage) {
+      case 'compute': {
+        code = `[[stage(compute)]] fn ${entryPoint}() {}`;
```
Kangz, line 33: nit: can you add `workgroup_size(1)` too? The group just decided to require these annotations on all compute entrypoints now.

src/webgpu/api/validation/createComputePipeline.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+      }
+      case 'vertex': {
+        code = `
+        [[stage(vertex)]] fn ${entryPoint}() -> [[builtin(position)]] vec4<f32> {
+          return vec4<f32>(0.0, 0.0, 0.0, 1.0);
+        }`;
+        break;
+      }
+      case 'fragment': {
+        const fragColorType = 'i32';
```
Kangz, line 44: Any reason for these variables? Since they never change I think it might be clearer to just put them in the code.

src/webgpu/api/validation/createComputePipeline.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+    });
+  });
+
+g.test('shader_module_stage_must_be_compute')
+  .cases(
+    params()
+      .combine(poptions('isAsync', [true, false]))
+      .combine(
+        poptions('shaderModuleStage', [
+          'compute' as TShaderStage,
```
Kangz, line 121: I think adding `as const` after the list might be sufficient instead of all the `as TShaderStage`. Though I don't understand why ^^

src/webgpu/api/validation/createComputePipeline.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+        });
+      }
+    }
+  }
+}
+
+export const g = makeTestGroup(F);
+
+g.test('basic_use_of_createComputePipeline')
+  .cases(poptions('isAsync', [true, false]))
```
Kangz, line 95: Can we have a `.desc` for all the tests? See an example here: https://github.com/gpuweb/cts/blob/main/docs/intro/plans.md

src/webgpu/api/validation/createComputePipeline.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+      .combine([
+        { shaderModuleEntryPoint: 'main', stageEntryPoint: 'main' },
+        { shaderModuleEntryPoint: 'main', stageEntryPoint: '' },
+        { shaderModuleEntryPoint: 'main', stageEntryPoint: 'main\0' },
+        { shaderModuleEntryPoint: 'main', stageEntryPoint: 'main\0a' },
+        { shaderModuleEntryPoint: 'main', stageEntryPoint: 'mian' },
+        { shaderModuleEntryPoint: 'main', stageEntryPoint: 'main ' },
+        { shaderModuleEntryPoint: 'main', stageEntryPoint: 'main\n' },
+        { shaderModuleEntryPoint: 'mian', stageEntryPoint: 'mian' },
+        { shaderModuleEntryPoint: 'mian', stageEntryPoint: 'main' },
```
Kangz, line 151: Could we also have a couple tests with `shaderModuleEntryPoint` != "main"? I'm asking because for a very long time Dawn didn't support entrypoints other than "main".

Kangz (approved):

src/webgpu/api/validation/createComputePipeline.spec.ts
jzm-intel: In fact the last two tests have shaderModuleEntryPoint === "mian" rather than "main". Maybe I can add more cases like "mainmain" or "entryPoint0"?

src/webgpu/api/validation/createComputePipeline.spec.ts
jzm-intel: I change it into "TShaderStage[]".

src/webgpu/api/validation/createComputePipeline.spec.ts
Kangz: I sorry I meant that we could remove the test plan as well, since it is now in the `.desc()` for each test.

src/webgpu/api/validation/createComputePipeline.spec.ts
Kangz: Great!

Kangz (approved):
LGTM

You can mark the comments you addressed as "Resolved" for easy tracking.

PR #577, opened 2021-06-11
Fix unintended compilation and validation failures

ben-clayton:
A number of fixes for updating WGSL to match new spec changes.

The `[[offset]]` decoration has been removed for `[[size]]` and `[[align]]` decorations. See https://github.com/gpuweb/gpuweb/pull/1447.

Vertex stage entry points must now return a `[[builtin(position)]]` value.

Most of this is done without testing, as there's many so many pre-existing failing tests, I simply couldn't use the test runner.

### DIFF
--- a/src/webgpu/api/operation/compute/basic.spec.ts
+++ b/src/webgpu/api/operation/compute/basic.spec.ts
@@ -28,7 +28,7 @@ g.test('memcpy').fn(async t => {
       module: t.device.createShaderModule({
         code: `
           [[block]] struct Data {
-              [[offset(0)]] value : u32;
+              value : u32;
           };
 
           [[group(0), binding(0)]] var<storage, read> src : Data;
--- a/src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
+++ b/src/webgpu/api/operation/memory_sync/buffer/buffer_sync_test.ts
@@ -54,7 +54,7 @@ export class BufferSyncTest extends GPUTest {
   createStorageWriteComputePipeline(value: number): GPUComputePipeline {
     const wgslCompute = `
       [[block]] struct Data {
-        [[offset(0)]] a : i32;
+        a : i32;
       };
 
       [[group(0), binding(0)]] var<storage, read_write> data : Data;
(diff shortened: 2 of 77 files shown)


ben-clayton:
Okay, what formatter do I need to run?

ben-clayton:
`esbenp.prettier-vscode`, apparently. Would be good to get this documented! :)

kainino0x:
I was sure it was documented somewhere, but the command is `npm run fix`. I'll make sure it's documented.

ben-clayton:
> I was sure it was documented somewhere, but the command is `npm run fix`. I'll make sure it's documented.

Aha, I've found it in `docs/intro/developing.md`, when I was looking in `docs/intro/tests.md`. `docs/intro/tests.md` does also say:

> Before uploading, you can run pre-submit checks (`npm test`) to make sure it will pass CI.

Which gives the following errors:

```
src/webgpu/api/operation/shader_module/compilation_info.spec.ts:162:24 - error TS2339: Property 'offset' does not exist on type 'GPUCompilationMessage'.

162       t.expect(message.offset <= _code.length, 'Message offset should be within the shader source');
                           ~~~~~~

src/webgpu/api/operation/shader_module/compilation_info.spec.ts:164:17 - error TS2339: Property 'offset' does not exist on type 'GPUCompilationMessage'.

164         message.offset + message.length <= _code.length,
                    ~~~~~~

src/webgpu/api/operation/shader_module/compilation_info.spec.ts:164:34 - error TS2339: Property 'length' does not exist on type 'GPUCompilationMessage'.

164         message.offset + message.length <= _code.length,
                                     ~~~~~~

src/webgpu/api/operation/shader_module/compilation_info.spec.ts:179:19 - error TS2339: Property 'offset' does not exist on type 'GPUCompilationMessage'.

179           message.offset === lineOffset + message.linePos - 1,
                      ~~~~~~

src/webgpu/util/device_pool.ts:217:31 - error TS2345: Argument of type 'string' is not assignable to parameter of type 'GPUFeatureName'.

217     if (!adapter.features.has(feature.toString())) {
```

However, I'm moderately sure I didn't introduce these, and the CI clearly passed.


`npm run fix` now returns without having to change anything, and the CI seems happy.

Kangz:
Maybe try updating packages with `npm install`? That could come from using an outdated version of webgpu-types.

ben-clayton:
> Maybe try updating packages with npm install? That could come from using an outdated version of webgpu-types.

Yup, that did it. Thanks.

kainino0x (approved):
Thanks for the maintenance!

PR #585, opened 2021-06-20
Validation tests for queue.writeBuffer

haoxli:
- The ranges tests have covered these validations:
  - the write size is  too large for the destination buffer,
  - the write size extends the source buffer size.
- Add tests for buffer offset and data offset out of range.
- Add tests for the buffer missed COPY_DST usage.
- Cleanup TODOs.

<hr>

### DIFF
--- a/src/webgpu/api/validation/queue/writeBuffer.spec.ts
+++ b/src/webgpu/api/validation/queue/writeBuffer.spec.ts
@@ -1,18 +1,8 @@
 export const description = `
 Tests writeBuffer validation.
 
-- buffer missing usage flag
-- bufferOffset {ok, unaligned, too large for buffer}
-- dataOffset {ok, too large for data}
-- buffer size {ok, too small for copy}
-- data size {ok, too small for copy}
-- size {aligned, unaligned}
-- size unspecified; default {ok, too large for buffer}
-
 Note: destroyed buffer is tested in destroyed/.
-
-TODO: implement usage flag validation.
-TODO: validate large write sizes that may overflow.
+Note: buffer map state is tested in ./buffer_mapped.spec.ts.
 `;
 
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
@@ -20,6 +10,7 @@ import {
   TypedArrayBufferView,
   TypedArrayBufferViewConstructor,
 } from '../../../../common/util/util.js';
+import { GPUConst } from '../../../constants.js';
 import { ValidationTest } from '../validation_test.js';
 
 export const g = makeTestGroup(ValidationTest);
@@ -36,7 +27,7 @@ interpreted correctly for both.
 
 Also verifies that the specified data range:
 
-  - Describes a valid range of the source buffer.
+  - Describes a valid range of the destination buffer and source buffer.
   - Fits fully within the destination buffer.
   - Has a byte size which is a multiple of 4.
 `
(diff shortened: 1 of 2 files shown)


src/webgpu/api/validation/queue/writeBuffer.spec.ts
```diff
@@ -1,21 +1,12 @@
...
-- dataOffset {ok, too large for data}
-- buffer size {ok, too small for copy}
-- data size {ok, too small for copy}
-- size {aligned, unaligned}
-- size unspecified; default {ok, too large for buffer}
-
 Note: destroyed buffer is tested in destroyed/.
-
-TODO: implement usage flag validation.
-TODO: validate large write sizes that may overflow.
```
kainino0x, line 15: This TODO is about `size` or `offset+size` overflowing a 64-bit integer, and I don't think we test it yet, so please leave the TODO (can add clarification) - or add the test cases in this PR.

src/webgpu/api/validation/queue/writeBuffer.spec.ts
```diff
@@ -109,16 +100,21 @@ Also verifies that the specified data range:
...
+      // Writing with a data offset that is out of range of data size
+      t.shouldThrow('OperationError', () => queue.writeBuffer(buffer, 0, arraySm, 9, 0));
+
       // A data offset of undefined should be treated as 0
       queue.writeBuffer(buffer, 0, arraySm, undefined, 8);
       t.shouldThrow('OperationError', () => queue.writeBuffer(buffer, 0, arraySm, undefined, 12));
     }
 
     const arrayTypes = [
-      Uint8Array,
```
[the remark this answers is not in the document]
haoxli: Maybe, only `Uint8ClampedArray` is missed.

src/webgpu/api/validation/queue/writeBuffer.spec.ts
haoxli: The maximum safe value of Number is 2^54 - 1, we cannot make `size` or `data size`  in bytes overflow a 64-bit integer, the value larger than Number.MAX_SAFE_INTEGER are not acceptable.

kainino0x: That's the largest safe value, which means that every integer can be represented. After that, the next representable value after `x` is `x+2`, `x+4`, ..., `y`, `y+4`, `y+8`..., and so on. However you can still have values larger than MAX_SAFE_INTEGER that can be converted to uint64 (`unsigned long long`) by WebIDL, I believe.

The largest JS number (double) less than 2^64 appears to be `0xfffffffffffff800`, which should get converted to uint64, and then you could add an offset like `0x800` to it, and it would overflow to 0.

src/webgpu/api/validation/queue/writeBuffer.spec.ts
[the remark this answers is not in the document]
kainino0x: If the tests pass with Uint8ClampedArray, let's add it now.

src/webgpu/api/validation/queue/writeBuffer.spec.ts
```diff
@@ -132,6 +133,29 @@ Also verifies that the specified data range:
...
+g.test('write_sizes_overflow')
+  .desc(
+    `
+Tests writing large sizes that may overflow a 64-bit integer.
+`
+  )
+  .paramsSubcasesOnly([
+    { size: 16, _valid: true }, // control case
+    { size: 0xfffffffffffff800, _valid: false }, // The largest number less than the maximum of uint64
+    { size: 0xffffffffffffff00, _valid: false }, // The smallest number overflow uint64
```
kainino0x, line 145: Sorry I wasn't clear about what I meant by "add an offset". These are useful to test WebIDL implementations, but they should already be robust to this kind of overflow. The WebGPU-specific implementation (e.g. Blink or Dawn), however, is likely to do something like `offset + size` and we want to test that the implementation doesn't allow that to overflow. So cases are also needed which set `offset` to a larger value.

haoxli: Thanks for your clarification. 
As we discussed offline, I had filed an issue for the TypeError thrown if a Number larger that MAX_SAFE_INTEGER (https://bugs.chromium.org/p/dawn/issues/detail?id=961), and still expected it throw OperationError in our cases. If it's actually the WebIDL-required behavior, we will remove  the overflow case at that time.

kainino0x: We got an answer on the bug, and indeed it turns out it's invalid in WebIDL! Sorry for my misreading of the WebIDL spec. Let's remove the overflow cases then 👍

kainino0x (approved):

src/webgpu/api/validation/queue/writeBuffer.spec.ts
haoxli: Got it! Removed.

PR #589, opened 2021-06-22
Review and expand api,validation,buffer,destroy:*

toji:
Adds tests for destroying buffers that:
 - have all usages
 - are error buffers
 - are mapped

Error buffer test does not currently pass in Chrome, but the behavior is simple enough that we can confidently add the test now.

<hr>

- [ ] **N/A** All outstanding work is tracked with "TODO" in a test/file description or `.unimplemented()` on a test.
- [ ] **N/A** New helpers, if any, are documented using `/** doc comments */` and can be found via `helper_index.txt`.

### DIFF
--- a/src/webgpu/api/validation/buffer/destroy.spec.ts
+++ b/src/webgpu/api/validation/buffer/destroy.spec.ts
@@ -1,15 +1,43 @@
 export const description = `
-Destroying a buffer more than once is allowed.
+Validation tests for GPUBuffer.destroy.
 `;
 
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { kBufferUsages } from '../../../capability_info.js';
 import { GPUConst } from '../../../constants.js';
 import { ValidationTest } from '../validation_test.js';
 
 export const g = makeTestGroup(ValidationTest);
 
+g.test('all_usages')
+  .desc('Test destroying buffers of every usage type.')
+  .paramsSubcasesOnly(u =>
+    u //
+      .combine('usage', kBufferUsages)
+  )
+  .fn(async t => {
+    const { usage } = t.params;
+    const buf = t.device.createBuffer({
+      size: 4,
+      usage,
+    });
+
+    buf.destroy();
+  });
+
+g.test('error_buffer')
+  .desc('Test that error buffers may be destroyed without generating validation errors.')
+  .fn(async t => {
+    const buf = t.getErrorBuffer();
+    buf.destroy();
+  });
+
 g.test('twice')
-  .desc('Tests various mapping-related descripton options that could affect how state is tracked.')
+  .desc(
+    `Test that destroying a buffer more than once is allowed.
+      - Tests buffers which are mapped at creation or not
+      - Tests buffers with various usages`
+  )
   .paramsSubcasesOnly(u =>
     u //
       .combine('mappedAtCreation', [false, true])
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/buffer/destroy.spec.ts
```diff
@@ -28,7 +56,37 @@ g.test('twice')
...
-  .unimplemented();
+  .fn(async t => {
+    const { usage, mapMode, mappedAtCreation } = t.params;
+    const buf = t.device.createBuffer({
+      size: 4,
+      usage,
+      mappedAtCreation,
+    });
+
+    if (!mappedAtCreation && mapMode !== undefined) {
```
kainino0x, line 87: I suppose we should test mappedAtCreation -> unmap -> mapAsync -> destroy as well, and probably destroy-after-unmap as well, making the full list:

- MAP_READ; mapAsync(read) -> [unmap ->] destroy
- MAP_READ; mappedAtCreation -> [unmap ->] destroy
- MAP_READ; mappedAtCreation -> unmap -> mapAsync(read) -> [unmap ->] destroy
- MAP_WRITE; mapAsync(write) -> [unmap ->] destroy
- MAP_WRITE; mappedAtCreation -> [unmap ->] destroy
- MAP_WRITE; mappedAtCreation -> unmap -> mapAsync(write) -> [unmap ->] destroy

I think?

src/webgpu/api/validation/buffer/destroy.spec.ts
```diff
@@ -28,7 +56,37 @@ g.test('twice')
...
+        {
+          usage: GPUConst.BufferUsage.MAP_WRITE | GPUConst.BufferUsage.COPY_SRC,
+          mapMode: GPUConst.MapMode.WRITE,
+        },
+        {
+          usage: GPUConst.BufferUsage.COPY_DST | GPUConst.BufferUsage.MAP_READ,
+          mapMode: GPUConst.MapMode.READ,
+        },
+      ])
+      .filter(p => p.mappedAtCreation === false && p.mapMode === undefined)
```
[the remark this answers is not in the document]
toji: You are correct. Resolved.

src/webgpu/api/validation/buffer/destroy.spec.ts
toji: Deleting while unmapped after being mapped previously seems like a separate test altogether, so I've added it as one rather than try to subcase `while_mapped` futher.

kainino0x: I'm not sure they're really so different, at least in terms of how the test is written: it's just an extra boolean parameter and one line that says `if (unmapBeforeDestroy) buf.unmap()`.

I think the new changes also don't cover `mappedAtCreation -> unmap -> mapAsync -> destroy`, which admittedly is a weird case, but I usually suggest covering it, since re-mapping mappedAtCreation buffers seems like it could run into some state tracking corner case.

toji: Okay, revised it to condense the test down into one again.

src/webgpu/api/validation/buffer/destroy.spec.ts
```diff
@@ -56,13 +56,14 @@ g.test('twice')
...
       - Tests {mappable, unmappable mapAtCreation, mappable mapAtCreation}
-      - Tests while {mapped, mapped at creation}`
+      - Tests while {mapped, mapped at creation, unmapped}`
   )
   .paramsSubcasesOnly(u =>
     u //
       .combine('mappedAtCreation', [false, true])
+      .combine('unmapBeforeDestoy', [false, true])
       .combineWithParams([
         { usage: GPUConst.BufferUsage.COPY_SRC },
```
kainino0x, line 68: Cases where it's mappable without actually being mapped (in case the presence of a usage flag changes the implementation of mappedAtCreation)
```suggestion
        { usage: GPUConst.BufferUsage.COPY_SRC },
        { usage: GPUConst.BufferUsage.MAP_WRITE | GPUConst.BufferUsage.COPY_SRC },
        { usage: GPUConst.BufferUsage.COPY_DST | GPUConst.BufferUsage.MAP_READ },
```

src/webgpu/api/validation/buffer/destroy.spec.ts
```diff
@@ -74,18 +75,26 @@ g.test('while_mapped')
...
+      buf.unmap();
+    }
+
+    if ((!mappedAtCreation || unmapBeforeDestoy) && mapMode !== undefined) {
       await buf.mapAsync(mapMode);
+
+      if (unmapBeforeDestoy) {
+        buf.unmap();
+      }
     }
```
kainino0x: I had hard time following this logic, in particular the guarantee between `mappedAtCreation && unmapBeforeDestroy` and `!mappedAtCreation || unmapBeforeDestroy`. After working through the possibilities I am pretty sure it's right, but this seems like it would be more direct:

````suggestion
    if (mapMode !== undefined) {
      if (mappedAtCreation) {
        buf.unmap();
      }
      await buf.mapAsync(mapMode);
    }

    if (unmapBeforeDestroy) {
      buf.unmap();
    }
```

src/webgpu/api/validation/buffer/destroy.spec.ts
```diff
@@ -56,13 +56,14 @@ g.test('twice')
...
-    `Test destroying buffers while mapped.
+    `Test destroying buffers while mapped or after being unmapped.
       - Tests {mappable, unmappable mapAtCreation, mappable mapAtCreation}
-      - Tests while {mapped, mapped at creation}`
+      - Tests while {mapped, mapped at creation, unmapped}`
   )
   .paramsSubcasesOnly(u =>
     u //
       .combine('mappedAtCreation', [false, true])
+      .combine('unmapBeforeDestoy', [false, true])
```
kainino0x, line 66: unmapBeforeDestoy -> unmapBeforeDestroy

kainino0x (approved):

src/webgpu/api/validation/buffer/destroy.spec.ts
toji: In combination with your two new suggested cases, yes. This looks like it would cover it.

toji:
Thanks for the feedback!

PR #592, opened 2021-06-24
Refactor ArrayBuffer/GPUBuffer content checking helpers

kainino0x:
Requires #593 #594 to compile, but doesn't have conflicts with them, so completely separate PR.

<hr>

### DIFF
--- a/src/webgpu/api/operation/buffers/map.spec.ts
+++ b/src/webgpu/api/operation/buffers/map.spec.ts
@@ -18,6 +18,7 @@ TODO: Some testing (probably minimal) of accessing with different TypedArray/Dat
 
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
 import { assert } from '../../../../common/util/util.js';
+import { checkElementsEqual } from '../../../util/check_contents.js';
 
 import { MappingTest } from './mapping_test.js';
 
@@ -133,7 +134,7 @@ map-read and check the read-back result.`
     const mapRegion = getRegionForMap(size, [rangeOffset, rangeSize], t.params);
     await buffer.mapAsync(GPUMapMode.READ, ...mapRegion);
     const actual = new Uint8Array(buffer.getMappedRange(...range));
-    t.expectBuffer(actual, new Uint8Array(expected.buffer));
+    t.expectOK(checkElementsEqual(actual, new Uint8Array(expected.buffer)));
   });
 
 g.test('mappedAtCreation')
--- a/src/webgpu/api/operation/resource_init/buffer.spec.ts
+++ b/src/webgpu/api/operation/resource_init/buffer.spec.ts
@@ -2,6 +2,7 @@ import { makeTestGroup } from '../../../../common/framework/test_group.js';
 import { assert, unreachable } from '../../../../common/util/util.js';
 import { GPUConst } from '../../../constants.js';
 import { GPUTest } from '../../../gpu_test.js';
+import { checkElementsEqual } from '../../../util/check_contents.js';
 import { getTextureCopyLayout } from '../../../util/texture/layout.js';
 
 export const description = `
@@ -48,7 +49,7 @@ class F extends GPUTest {
     // usage contains COPY_SRC.
     if (bufferUsage & GPUBufferUsage.MAP_READ) {
       await buffer.mapAsync(GPUMapMode.READ);
-      this.expectBuffer(new Uint8Array(buffer.getMappedRange()), expectedData);
+      this.expectOK(checkElementsEqual(new Uint8Array(buffer.getMappedRange()), expectedData));
       buffer.unmap();
     } else {
       assert((bufferUsage & GPUBufferUsage.COPY_SRC) !== 0);
--- a/src/webgpu/gpu_test.ts
+++ b/src/webgpu/gpu_test.ts
@@ -1,6 +1,10 @@
 import { Fixture } from '../common/framework/fixture.js';
 import { attemptGarbageCollection } from '../common/util/collect_garbage.js';
-import { assert } from '../common/util/util.js';
+import {
+  assert,
+  TypedArrayBufferView,
+  TypedArrayBufferViewConstructor,
+} from '../common/util/util.js';
 
 import {
   EncodableTextureFormat,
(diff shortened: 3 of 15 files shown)


kainino0x:
@Kangz PTAL
2 commits, can review separately. The second one is a bit large but I think it should be fairly digestible (look at check_contents.ts, then gpu_test.ts, then the rest is small bits).

src/webgpu/api/operation/buffers/map.spec.ts
```diff
@@ -125,7 +126,7 @@ map-read and check the read-back result.`
     const mapRegion = getRegionForMap(size, [rangeOffset, rangeSize], t.params);
     await buffer.mapAsync(GPUMapMode.READ, ...mapRegion);
     const actual = new Uint8Array(buffer.getMappedRange(...range));
-    t.expectBuffer(actual, new Uint8Array(expected.buffer));
+    t.expectOK(checkElementsEqual(actual, new Uint8Array(expected.buffer)));
```
Kangz, line 129: nit: `this.expectGPUBufferValuesEqual` instead? In general it feels very few places in the code should need to use `expectOk`

src/webgpu/api/operation/resource_init/buffer.spec.ts
```diff
@@ -49,7 +50,7 @@ class F extends GPUTest {
     // usage contains COPY_SRC.
     if (bufferUsage & GPUBufferUsage.MAP_READ) {
       await buffer.mapAsync(GPUMapMode.READ);
-      this.expectBuffer(new Uint8Array(buffer.getMappedRange()), expectedData);
+      this.expectOK(checkElementsEqual(new Uint8Array(buffer.getMappedRange()), expectedData));
```
Kangz, line 53: nit: `this.expectGPUBufferValuesEqual` instead?

src/webgpu/util/check_contents.ts
```diff
@@ -0,0 +1,234 @@
...
+  range,
+  TypedArrayBufferView,
+  TypedArrayBufferViewConstructor,
+} from '../../common/util/util.js';
+
+export type BufferCheckGenerator = (index: number) => number;
+export type BufferCheckPredicate = (index: number, value: number) => boolean;
+export type BufferCheckPredicatePrinter = {
+  leftHeader: string;
+  cell: (index: number) => number;
```
Kangz, line 14: nit: tableCellValue? If I understand correctly what this does.

src/webgpu/util/check_contents.ts
```diff
@@ -0,0 +1,234 @@
...
+      value >= Math.min(expected[0][index], expected[1][index]) &&
+      value <= Math.max(expected[0][index], expected[1][index]),
+    {
+      predicatePrinter: [
+        { leftHeader: 'between', cell: index => expected[0][index] },
+        { leftHeader: 'and', cell: index => expected[1][index] },
+      ],
+    }
+  );
+  return e === undefined
```
Kangz, line 50: I'm not sure I understand what this does. Maybe it could be a method on ErrorWithExtra so we can do `e??.dostuff` the method name would help make what happens more explicit

src/webgpu/util/check_contents.ts
```diff
@@ -0,0 +1,234 @@
...
+  const size = actual.length;
+  const ctor = actual.constructor as TypedArrayBufferViewConstructor;
+  const printAsFloat = ctor === Float32Array || ctor === Float64Array;
+
+  let failedElementsFirstMaybe: number | undefined = undefined;
+  const failedElements: (true | undefined)[] = [];
+  for (let i = 0; i < size; ++i) {
+    if (!predicate(i, actual[i])) {
+      failedElementsFirstMaybe ??= i;
+      failedElements[i] = true;
```
Kangz, line 120: nit: Maybe all out that we use the "sparse array" feature of JS.

src/webgpu/util/check_contents.ts
```diff
@@ -0,0 +1,234 @@
...
+ * The first argument is the printing options:
+ *  - fillToWidth: Keep printing columns (as long as there is data) until this width is passed.
+ *    If there is more data, "..." is appended.
+ *  - numberToString: if a cell value is a number, this is used to stringify it.
+ *
+ * Each remaining argument provides one row for the table.
+ */
+function generatePrettyTable(
+  { fillToWidth, numberToString }: { fillToWidth: number; numberToString: (n: number) => string },
+  ...rows: ReadonlyArray<Iterable<string | number>>
```
Kangz, line 199: nit: we could pass a list directly here instead of relying on `...` this seems quite obscure in JS/TS.

Kangz (approved):

src/webgpu/api/operation/buffers/map.spec.ts
kainino0x: This isn't checking a GPUBuffer though, so I'd be adding a new `expectElementsEqual` to the fixture for just a few cases.

I thought about adding an option to make expectGPUBufferValuesEqual readback using mapAsync. I was thinking it wouldn't be able to be `eventual` but actually it could consume the input GPUBuffer so it wouldn't have to be.

kainino0x: I think I found a preexisting bug in expectGPUBufferValuesPassCheck (`begin` and `end` are byte offsets but it treats them as array offsets), so I'm going to try to fix that separately then do this.

src/webgpu/gpu_test.ts
```diff
@@ -216,377 +199,52 @@ export class GPUTest extends Fixture {
   }
 
   /**
-   * Expect a GPUBuffer's contents to equal the values in the provided TypedArray.
+   * Expect a GPUBuffer's contents to pass the provided check.
    */
-  expectGPUBufferValuesEqual(
+  expectGPUBufferValuesPassCheck(
```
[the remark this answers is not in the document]
kainino0x: Nice catch. This turned out not to be as simple to fix as I expected.

src/webgpu/api/operation/buffers/map.spec.ts
Kangz: ok

kainino0x: I didn't end up fixing this one because it can't be folded with expectGPUBufferValuesEqual because it is testing the map itself, but it would also add a one-off expectElementsEqual that isn't used anywhere else. I think this is okay. Also I may adopt the expectOK(checkX()) pattern in the future if if it's handy, as it's more composable.

Kangz: Ack

PR #600, opened 2021-06-28
CopyToTexture,canvas basic tests : Add premultiplied alpha tests

shaoboyan:
This PR add premultiplied alpha tests to test 2d/webgl/webgl2
canvas/offscreenCanvas.

The tests covers webgl/webgl2 context created with premultipliedAlpha
attribute values and the CopyExternalImageToTexture with different
premultipliedAlpha config value

<hr>

### DIFF
--- a/src/webgpu/web_platform/copyToTexture/canvas.spec.ts
+++ b/src/webgpu/web_platform/copyToTexture/canvas.spec.ts
@@ -15,24 +15,11 @@ import { CopyToTextureUtils } from '../../util/copyToTexture.js';
 import { kTexelRepresentationInfo } from '../../util/texture/texel_data.js';
 
 class F extends CopyToTextureUtils {
-  initCanvasContent({
-    canvasType,
-    contextName,
-    width,
-    height,
-  }: {
-    canvasType: 'onscreen' | 'offscreen';
-    contextName: '2d' | 'webgl' | 'webgl2';
-    width: number;
-    height: number;
-  }): {
-    canvas: HTMLCanvasElement | OffscreenCanvas;
-    canvasContext:
-      | WebGLRenderingContext
-      | WebGL2RenderingContext
-      | CanvasRenderingContext2D
-      | OffscreenCanvasRenderingContext2D;
-  } {
+  createCanvas(
+    canvasType: 'onscreen' | 'offscreen',
+    width: number,
+    height: number
+  ): HTMLCanvasElement | OffscreenCanvas | null {
     let canvas: HTMLCanvasElement | OffscreenCanvas | null = null;
     if (canvasType === 'onscreen') {
       if (typeof document !== 'undefined') {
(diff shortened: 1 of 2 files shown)


shaoboyan:
@kainino0x @Kangz  PTAL!

src/webgpu/web_platform/copyToTexture/canvas.spec.ts
```diff
@@ -51,53 +38,137 @@ class F extends CopyToTextureUtils {
...
-      | WebGLRenderingContext
-      | WebGL2RenderingContext
+    let canvasContext = null;
+    canvasContext = canvas.getContext('2d') as
       | CanvasRenderingContext2D
       | OffscreenCanvasRenderingContext2D
       | null;
+
     if (canvasContext === null) {
       this.skip(canvasType + ' canvas context not available');
```
kainino0x, line 70:
```suggestion
      this.skip(canvasContext + ' canvas context not available');
```

src/webgpu/web_platform/copyToTexture/canvas.spec.ts
```diff
@@ -51,53 +38,137 @@ class F extends CopyToTextureUtils {
...
+      | null;
+
+    if (canvasContext === null) {
+      this.skip(canvasType + ' canvas context not available');
+    }
+
+    const rectWidth = Math.floor(width / 2);
+    const rectHeight = Math.floor(height / 2);
+
+    const gl = canvasContext as WebGLRenderingContext | WebGL2RenderingContext;
```
kainino0x, line 137: nit: This `as` cast shouldn't be needed. You could remove it (`const gl = canvasContext;`) or just rename `canvasContext` to `gl` and remove this line.

src/webgpu/web_platform/copyToTexture/canvas.spec.ts
```diff
@@ -214,17 +317,131 @@ g.test('copy_contents_from_canvas')
 
     // Construct expected value for different dst color format
     const dstBytesPerPixel = kTextureFormatInfo[dstColorFormat].bytesPerBlock;
+    const format: RegularTextureFormat =
+      dstColorFormat === 'rgba8unorm-srgb' || dstColorFormat === 'bgra8unorm-srgb'
+        ? dstColorFormat === 'rgba8unorm-srgb'
+          ? 'rgba8unorm'
+          : 'bgra8unorm'
+        : dstColorFormat;
+    dstColorFormat;
```
kainino0x, line 326: Last line appears to be a no-op? And the rest can be a bit more straightforward.

Since this is also used below, factor it out into a small helper function that documents the intent.
```ts
/**
 * If the destination format specifies a transfer function,
 * copyExternalImageToTexture (like B2T/T2T copies) should ignore it.
 */
function formatForExpectedPixels(format: RegularTextureFormat): RegularTextureFormat {
  return dstColorFormat === 'rgba8unorm-srgb'
    ? 'rgba8unorm'
    : dstColorFormat === 'bgra8unorm-srgb'
    ? 'bgra8unorm'
    : dstColorFormat;
}
```

src/webgpu/web_platform/copyToTexture/canvas.spec.ts
```diff
@@ -214,17 +317,131 @@ g.test('copy_contents_from_canvas')
...
+
+    // For 2d canvas, get expected pixels with getImageData(), which returns origin
+    // values(not applied premultipy alpha) always.
```
kainino0x, line 329: nit: "origin" (noun) should be "original" (adjective).

Although actually I wouldn't call them "original" values. Canvas normally operates in premultiplied space. So we can just say:
```suggestion
    // For 2d canvas, get expected pixels with getImageData(), which returns unpremultiplied values.
```

src/webgpu/web_platform/copyToTexture/canvas.spec.ts
```diff
@@ -51,53 +38,137 @@ class F extends CopyToTextureUtils {
...
+    // rgb10a2unorm dst textures, we'll set alphaValue to 0.0 to test.
+    const alphaValue = isLosePrecisionDstFormat ? 1.0 : 0.6;
+
+    // For webgl/webgl2 context canvas, if the context created with premultipliedAlpha attributes,
+    // it means that the value in drawing buffer is premultiplied or not. So we should set
+    // premultipliedAlpha value for premultipliedAlpha true gl context and unpremultipliedAlpha value
+    // for the premulitpliedAlpha false gl context.
+    gl.enable(gl.SCISSOR_TEST);
+    gl.scissor(0, 0, rectWidth, rectHeight);
+    premultiplied
```
kainino0x, line 150: nit: factor out a `const colorValue = premultiplied ? alphaValue : 1.0;` and use it below; easier to read than one conditional on each gl.clearColor.

src/webgpu/web_platform/copyToTexture/canvas.spec.ts
```diff
@@ -214,17 +317,131 @@ g.test('copy_contents_from_canvas')
...
+  And the expected results are all passed.
+  `
+  )
+  .params(u =>
+    u
+      .combine('canvasType', ['onscreen', 'offscreen'] as const)
+      .combine('contextName', ['webgl', 'webgl2'] as const)
+      .combine('dstColorFormat', kValidTextureFormatsForCopyIB2T)
+      .combine('webglCanvasPremultipliedAlpha', [true, false])
+      .combine('premultipliedAlpha', [true, false])
```
kainino0x, line 381: nit: I think `srcPremultiplied` and `dstPremultiplied`, as you already use in getExpectedPixels, are more self-explanatory than these.

src/webgpu/web_platform/copyToTexture/canvas.spec.ts
```diff
@@ -214,17 +317,131 @@ g.test('copy_contents_from_canvas')
...
+
+    // Construct expected value for different dst color format
+    const dstBytesPerPixel = kTextureFormatInfo[dstColorFormat].bytesPerBlock;
+    const format: RegularTextureFormat =
+      dstColorFormat === 'rgba8unorm-srgb' || dstColorFormat === 'bgra8unorm-srgb'
+        ? dstColorFormat === 'rgba8unorm-srgb'
+          ? 'rgba8unorm'
+          : 'bgra8unorm'
+        : dstColorFormat;
+    dstColorFormat;
```
kainino0x, line 425: use helper from above

src/webgpu/web_platform/copyToTexture/canvas.spec.ts
```diff
@@ -185,20 +285,23 @@ g.test('copy_contents_from_canvas')
   .params(u =>
     u
       .combine('canvasType', ['onscreen', 'offscreen'] as const)
-      .combine('contextName', ['2d', 'webgl', 'webgl2'] as const)
       .combine('dstColorFormat', kValidTextureFormatsForCopyIB2T)
+      .combine('premultipliedAlpha', [true, false])
```
kainino0x, line 289: nit: `dstPremultiplied` is more self-explanatory

src/webgpu/web_platform/copyToTexture/canvas.spec.ts
```diff
@@ -135,18 +210,44 @@ class F extends CopyToTextureUtils {
...
+
+        rgba = {
+          R: sourcePixels[pixelPos * 4] / divide,
+          G: sourcePixels[pixelPos * 4 + 1] / divide,
+          B: sourcePixels[pixelPos * 4 + 2] / divide,
+          A: sourcePixels[pixelPos * 4 + 3] / divide,
+        };
+
+        if (!srcPremultiplied) {
+          if (dstPremultiplied) {
```
kainino0x, line 229: nit: merge these two with `&&`

src/webgpu/web_platform/copyToTexture/canvas.spec.ts
```diff
@@ -135,18 +210,44 @@ class F extends CopyToTextureUtils {
...
+            rgba.R *= rgba.A;
+            rgba.G *= rgba.A;
+            rgba.B *= rgba.A;
+          }
+        }
+
+        if (srcPremultiplied && !dstPremultiplied) {
+          if (rgba.A !== 0) {
```
kainino0x, line 237: Would it work to assert this instead? Having this as an `if()` creates a corner case where I'm not sure the generated values are correct.

src/webgpu/web_platform/copyToTexture/canvas.spec.ts
```diff
@@ -135,18 +210,44 @@ class F extends CopyToTextureUtils {
...
             rep.encode({
-              R: sourcePixels[pixelPos * 4] / divide,
-              G: sourcePixels[pixelPos * 4 + 1] / divide,
-              B: sourcePixels[pixelPos * 4 + 2] / divide,
-              A: sourcePixels[pixelPos * 4 + 3] / divide,
+              R: rgba.R,
+              G: rgba.G,
+              B: rgba.B,
+              A: rgba.A,
             })
```
kainino0x, line 251: can this be `rep.encode(rgba)`?

src/webgpu/web_platform/copyToTexture/canvas.spec.ts
```diff
@@ -51,53 +38,137 @@ class F extends CopyToTextureUtils {
...
+    premultiplied
+      ? gl.clearColor(0.0, 0.0, alphaValue, alphaValue)
+      : gl.clearColor(0.0, 0.0, 1.0, alphaValue);
+    gl.clear(gl.COLOR_BUFFER_BIT);
+
+    gl.scissor(rectWidth, rectHeight, width - rectWidth, height - rectHeight);
+    premultiplied
+      ? gl.clearColor(alphaValue, alphaValue, alphaValue, alphaValue)
+      : gl.clearColor(1.0, 1.0, 1.0, alphaValue);
+    gl.clear(gl.COLOR_BUFFER_BIT);
```
kainino0x, line 171: Test description says this is black, but it's white

src/webgpu/web_platform/copyToTexture/canvas.spec.ts
```diff
@@ -51,53 +38,137 @@ class F extends CopyToTextureUtils {
...
+    height,
+    premultiplied,
+    isLosePrecisionDstFormat,
+  }: {
+    canvasType: 'onscreen' | 'offscreen';
+    contextName: 'webgl' | 'webgl2';
+    width: number;
+    height: number;
+    premultiplied: boolean;
+    isLosePrecisionDstFormat: boolean;
```
kainino0x, line 114: Please give this a name that describes what it does, rather than when it's needed. I don't fully understand it so I don't have a name suggestion yet.

src/webgpu/web_platform/copyToTexture/canvas.spec.ts
```diff
@@ -51,53 +38,137 @@ class F extends CopyToTextureUtils {
...
+    }
+
+    const rectWidth = Math.floor(width / 2);
+    const rectHeight = Math.floor(height / 2);
+
+    const gl = canvasContext as WebGLRenderingContext | WebGL2RenderingContext;
+
+    // The rgb10a2unorm dst texture will have tiny errors when we compare actual and expectation.
+    // This is due to the convert from 8-bit to 10-bit combined with alpha value ops. So for
+    // rgb10a2unorm dst textures, we'll set alphaValue to 0.0 to test.
```
kainino0x, line 141: Please explain why this fixes the problem. Also, the comment says alphaValue is `0.0` but code says `1.0`.

src/webgpu/web_platform/copyToTexture/canvas.spec.ts
shaoboyan: I expect the error message should be : "onscreen canvas : canvas context not available". So wants the keep 'canvasType' here.

kainino0x: Availability of the canvasType was already checked, so maybe here it makes sense to report both. So e.g. if offscreen + 2d is not available, it says that.

Though it doesn't really matter I suppose, the canvas type and context type will already be in the test case name!

shaoboyan: <  the canvas type and context type will already be in the test case name!
Good point!

src/webgpu/web_platform/copyToTexture/canvas.spec.ts
shaoboyan: Ah, yes, and it reminds me that I'm not updating test description.

src/webgpu/web_platform/copyToTexture/canvas.spec.ts
```diff
@@ -286,22 +271,28 @@ g.test('copy_contents_from_2d_context_canvas')
...
+
+    // When dst texture format is rgb10a2unorm, the generated expected value of the result
+    // may have tiny errors compared to the actual result when the channel value is
+    // not 1.0 or 0.0.
+    // For example, we init the pixel with r channel to 0.6. And the denormalized value for
+    // 10-bit channel is 613.8, which needs to call "round" or other function to get an integer.
+    // It is possible that gpu adopt different "round" as our cpu implementation(we use Math.round())
+    // and it will generate tiny errors.
+    // So the cases with rgb10a2unorm dst texture format are handled specially by painting opaque rects
+    // to ensure they will have stable result after alphaOps(should keep the same value).
```
kainino0x, line 290: Thanks, this makes sense! Does this mean we are losing some coverage, though, as we only test opaque values when the destination is rgb10a2? It's probably not that big of a deal, but perhaps we should leave a TODO to test this with non-opaque values using a tolerance. In theory, this should be much easier at some point later, once we've written good texture checking helpers that can deal with tolerances easily.

src/webgpu/web_platform/copyToTexture/canvas.spec.ts
```diff
@@ -87,12 +99,8 @@ class F extends CopyToTextureUtils {
...
     ctx.fillStyle = 'rgba(0, 0, 255,' + alphaValue + ')';
     ctx.fillRect(0, rectHeight, rectWidth, height - rectHeight);
-    // Black
-    if (isLosePrecisionDstFormat) {
-      ctx.fillStyle = 'rgba(255, 255, 255, 0.0)';
-    } else {
-      ctx.fillStyle = 'rgba(0, 0, 0,' + alphaValue + ')';
-    }
+    // White
+    ctx.fillStyle = 'rgba(255, 255, 255,' + alphaValue + ')';
```
kainino0x, line 103: optional nit (just thought of this): I think this would be slightly more readable as `rgba(255, 255, 255, ${alphaValue})`

kainino0x (approved):
LGTM, please squash-and-merge this once a TODO is added (unless you don't think one is needed)

src/webgpu/web_platform/copyToTexture/canvas.spec.ts
shaoboyan: Yes, I think we lost some coverage and the TODO has already been added in the desc. I'll address it when I figure out the pack(encode) works in framework and how to add the error lost.

kainino0x: Ahh, sorry I forgot about the TODO. I had seen it before and forgot :)

shaoboyan: Also realized that we could do some opt in initialization part to reduce the case running time. Add TODOs for the helper functions.

PR #603, opened 2021-06-29
api,validation,encoding,cmds,render,other plans

haoxli:
- Split validation plans to each cmd file.
- The resource_usages, programmable pass stuff, state tracking are covered in other tests.
- Test destroyed buffer in the other cmds.

<hr>


src/webgpu/api/validation/encoding/cmds/render/indirect_draw.spec.ts
```diff
@@ -0,0 +1,78 @@
+export const description = `
+Validation tests for drawIndirect/drawIndexedIndirect on render pass and render bundle.
```
Jiawei-Shao, line 2: Do we need to test if the draw parameters stored in the indirect buffer are properly validated, just like they are used in ```draw()``` | ```drawIndexed()```?

haoxli: We should validate the indirect buffer parameters like `draw()` and `drawIndexed()` did, but we haven't such validation rules in indirect draw calls, is it missing or not added due to other reasons? @kainino0x

kainino0x: The parameters inside the indirect buffer exist only on the GPU, so they have to be validated on-GPU, so they can't produce validation errors without blocking on all previous GPU work. Instead, various behaviors are allowed - no-op the draw call, or robust buffer access: https://github.com/gpuweb/gpuweb/pull/1835

You can also think of it this way: An indirect draw call's parameters going out of bounds (of an index or vertex buffer) is similar to an index buffer going out of bounds (on a vertex buffer).

src/webgpu/api/validation/encoding/cmds/render/indirect_draw.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+        { indirectOffset: 0, bufferSize: minSize },
+        { indirectOffset: 0, bufferSize: minSize - 1 },
+        { indirectOffset: 0, bufferSize: minSize - 4 },
+        { indirectOffset: 4, bufferSize: minSize + 4 },
+        { indirectOffset: 4, bufferSize: minSize + 3 },
+        { indirectOffset: 3, bufferSize: minSize + 4 },
+        { indirectOffset: 5, bufferSize: minSize + 4 },
+        { indirectOffset: minSize, bufferSize: minSize },
+        { indirectOffset: minSize + 4, bufferSize: minSize },
+      ] as const;
```
kainino0x, line 79: nit: Maybe this would be clearer to express as indirectOffset and additionalBufferSize? Where `bufferSize` gets computed inside the test as `indirectOffset + additionalBufferSize`

src/webgpu/api/validation/encoding/cmds/render/setVertexBuffer.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+  )
+  .unimplemented();
+
+g.test('offset_and_size_oob')
+  .desc(
+    `
+Tests offset and size cannot be larger than vertex buffer size
+- (offset, size) is
+  - (0, 1)
+  - (0, min size) (control case)
```
kainino0x, line 73: The smallest vertex format is only 2 bytes, and could be 1 byte in the future. I don't think there are any restrictions on min size of a vertex buffer binding, at least I don't see them in the spec right now.

kainino0x (commented):
Looks good so far!

src/webgpu/api/validation/encoding/cmds/render/setVertexBuffer.spec.ts
haoxli: Here I want the 'min size' to be the min size of an index / vertex buffer, depends on the data type in the buffer. The OOB validation has nothing to do with the data type, I thought is to use uint32 for index and float32 for vertex to have a same 'min size'.

src/webgpu/api/validation/encoding/cmds/render/indirect_draw.spec.ts
haoxli: Thank you for your answers! I think we can test the invalid parameters in indrect buffer in operation tests.

src/webgpu/api/validation/encoding/cmds/render/setVertexBuffer.spec.ts
kainino0x: A particular pipeline may need a certain size for each bound vertex and index buffer, but I don't see any restrictions on a setVertexBuffer call by itself. Both setIndexBuffer and setVertexBuffer do have alignment restrictions on offset, but not on size (at least in the current spec draft).

src/webgpu/api/validation/encoding/cmds/render/other.spec.ts
```diff
@@ -1,45 +0,0 @@
...
->         - (0, 0)
->         - (0, 1)
->         - (0, 4)
->         - (0, 5)
->         - (0, b.size)
->         - (min alignment, b.size - 4)
->         - (4, b.size - 4)
->         - (b.size - 4, 4)
->         - (b.size, min size)
->         - (0, min size), and if that's valid:
```
haoxli, line 23: [1]

src/webgpu/api/validation/encoding/cmds/render/setVertexBuffer.spec.ts
haoxli: Yes. The 'min size' does not come from SPEC. I saw it at [1]. If I misunderstood the 'min size', what should it be?

src/webgpu/api/validation/encoding/cmds/render/indirect_draw.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+  - (0, min size - min alignment)
+  - (0, min size - 1)
+  - (0, min size) (control case)
+  - (min alignment, min size + min alignment)
+  - (min alignment, min alignment + min size - 1)
+  - (min alignment +/- 1, min size + alignment)
+  - (min size, min size)
+  - (min size + min alignment, min size)
+  `
+  )
```
kainino0x, line 64: nit: Needs a case for whether it's drawIndirect or drawIndexedIndirect

src/webgpu/api/validation/encoding/cmds/render/indirect_draw.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+  .paramsSubcasesOnly(
+    kTestParams.expandWithParams(p => {
+      const minSize = p.indexed ? 20 : 16;
+      return [
+        { indirectOffset: 0, bufferSize: 0 },
+        { indirectOffset: 0, bufferSize: minSize },
+        { indirectOffset: 0, bufferSize: minSize - 1 },
```
kainino0x, line 71: Please also test `{ indirectOffset: 0, bufferSize: minSize + 1 }` as a control case to make sure that doesn't hit any unrelated error.

src/webgpu/api/validation/encoding/cmds/render/indirect_draw.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+  - (min alignment, min size + min alignment)
+  - (min alignment, min alignment + min size - 1)
+  - (min alignment +/- 1, min size + alignment)
+  - (min size, min size)
+  - (min size + min alignment, min size)
+  `
+  )
+  .paramsSubcasesOnly(
+    kTestParams.expandWithParams(p => {
+      const minSize = p.indexed ? 20 : 16;
```
kainino0x, line 67: nit: Perhaps rename minSize to indirectParametersSize or something

src/webgpu/api/validation/encoding/cmds/render/setVertexBuffer.spec.ts
kainino0x: Ah, sorry I didn't find the existing notes that said "min size". I think it means means "whatever is the minimum setIndexBuffer `size` argument value (maybe it's 1, wasn't sure when writing this)". The point is to test the case that's just barely out of bounds, to make sure there are no off-by-one errors.

I think it's kind of redundant, e.g. `(0, 1)` and `(0, min size)` are probably the same thing.

src/webgpu/api/validation/encoding/cmds/render/indirect_draw.spec.ts
haoxli: There will be negative value for `additionalBufferSize` in the last case, it may bring doubts about negative additional size. Keeping `bufferSize`.

kainino0x (approved):

PR #605, opened 2021-06-30
Fix byte vs element units bug; allow buffer checking by mapping

kainino0x:
So, I guess replacing usages of `createAlignedCopyForMapRead` with `readGPUBufferRangeTyped` mostly didn't end up in net simplifications (especially since most of them just use `Uint8Array`), but I still think it's a bit easier to reason about? And it makes it easier for tests to test both read-by-copy and read-by-map (but does it matter?)

WDYT?

Tested `webgpu:api,operation,resource_init,buffer:*` locally and still passes.

<hr>


Kangz (approved):
LGTM. I think it is a small improvements.

PR #606, opened 2021-06-30
Add operation tests on writeTexture() and B2T copies with stencil aspects

Jiawei-Shao:
<hr>


src/webgpu/api/operation/command_buffer/image_copy.spec.ts
```diff
@@ -625,6 +626,149 @@ class ImageCopyTest extends GPUTest {
         unreachable();
     }
   }
+
+  async CheckStencilTextureContent(
```
Kangz, line 630: unit: checkStencilTextureContent

src/webgpu/api/operation/command_buffer/image_copy.spec.ts
```diff
@@ -625,6 +626,149 @@ class ImageCopyTest extends GPUTest {
...
+      expectedStencilTextureData.byteLength ===
+        copyLayout.mipSize[0] * copyLayout.mipSize[1] * copyLayout.mipSize[2] +
+          expectedStencilTextureDataOffset
+    );
+
+    const renderPipeline = this.device.createRenderPipeline({
+      vertex: {
+        module: this.device.createShaderModule({
+          code: `
+            [[stage(vertex)]] fn main(
```
Kangz, line 651: nit: the separation between the function declaration and its body is a bit hard to see. Ditto below for the fragment shader.

src/webgpu/api/operation/command_buffer/image_copy.spec.ts
Kangz: FYI @kainino0x maybe we want to move this texture helper to be shared between tests in the future.

Might be worth doing a perfect copy of the stencil data instead of checking each pixel individually below. It should be possible to do this with 8 draws with blendMode ADD that each check a single bit of the stencil buffer.

src/webgpu/api/operation/command_buffer/image_copy.spec.ts
```diff
@@ -625,6 +626,149 @@ class ImageCopyTest extends GPUTest {
...
+        colorAttachments: [
+          {
+            view: colorTexture.createView(),
+            loadValue: { r: 0.0, g: 0.0, b: 0.0, a: 0.0 },
+            storeOp: 'store',
+          },
+        ],
+        depthStencilAttachment: {
+          view: stencilTexture.createView({
+            baseMipLevel: stencilTextureMipLevel,
```
Kangz, line 718: nit: mipLevelCount: 1 too I'd imagine.

src/webgpu/api/operation/command_buffer/image_copy.spec.ts
Kangz: As a crazy hack we could use the instanceIndex to pass extra data with baseInstance without adding a uniform buffer ^^

src/webgpu/api/operation/command_buffer/image_copy.spec.ts
```diff
@@ -1133,3 +1277,82 @@ g.test('copy_into_stencil_aspect')
   `
   )
   .unimplemented();
+
+g.test('write_into_stencil_aspect')
```
Kangz, line 1281: Sorry this is a big change, but wouldn't there be a way to add stencil8 to the kWorkingFormats and make the texture check use the new method you made, such that all the tests will automatically start testing stencil8?

src/webgpu/api/operation/command_buffer/image_copy.spec.ts
kainino0x: > Might be worth doing a perfect copy of the stencil data instead of checking each pixel individually below. It should be possible to do this with 8 draws with blendMode ADD that each check a single bit of the stencil buffer.

This is one of the things I want to have a texture helper for.

src/webgpu/api/operation/command_buffer/image_copy.spec.ts
Jiawei-Shao: Sorry I may not fully understand your comment here. In fact ```stencil8``` is included in ```kDepthStencilFormats``` and is tested in this test.

src/webgpu/api/operation/command_buffer/image_copy.spec.ts
Jiawei-Shao: As in WebGPU we support copying from the stencil aspect into a buffer, I think it easier to verify the stencil data with T2B copies. In this test I am testing the B2T and T2B copies, so I have to check the stencil data in a render pass, so it seems we just need to keep this code here.

Jiawei-Shao: > Might be worth doing a perfect copy of the stencil data instead of checking each pixel individually below. It should be possible to do this with 8 draws with blendMode ADD that each check a single bit of the stencil buffer.

Do you mean using 8 render pipelines with different stencilReadMask (1 << 0 to 1<< 7), and in each draw we output ( (1<<0) / 255, (1 << 1) / 255, ...  (1 << 7) / 255) to an R8Unorm color output?

Kangz: Yes exactly.

src/webgpu/api/operation/command_buffer/image_copy.spec.ts
Jiawei-Shao: Does it look better now?

src/webgpu/api/operation/command_buffer/image_copy.spec.ts
```diff
@@ -627,6 +629,250 @@ class ImageCopyTest extends GPUTest {
...
+    });
+    assert(
+      expectedStencilTextureData.byteLength ===
+        copyLayout.mipSize[0] * copyLayout.mipSize[1] * copyLayout.mipSize[2] +
+          expectedStencilTextureDataOffset
+    );
+
+    const stencilBitCount = 8;
+
+    // Choose 'rgba8unorm' as the format of referenceTexture because WebGPU doesn't support
```
Kangz, line 652: nit: I'm a bit surprised we're doing a texture to compute the bit -> r8unorm float value, when we could just output `(1 << param.stencilBitIndex) / 255` in the shader. Am I missing something?

src/webgpu/api/operation/command_buffer/image_copy.spec.ts
```diff
@@ -627,6 +629,250 @@ class ImageCopyTest extends GPUTest {
...
+    const uniformBufferSize = kMinDynamicBufferOffsetAlignment * (stencilBitCount - 1) + 4;
+    const uniformBuffer = this.device.createBuffer({
+      size: uniformBufferSize,
+      usage: GPUBufferUsage.COPY_DST | GPUBufferUsage.UNIFORM,
+    });
+    const uniformBufferData = new Uint32Array(uniformBufferSize / 4);
+    for (let i = 1; i < stencilBitCount; ++i) {
+      uniformBufferData[(kMinDynamicBufferOffsetAlignment / 4) * i] = i;
+    }
+    this.queue.writeBuffer(uniformBuffer, 0, uniformBufferData);
```
Kangz, line 684: nit: I think there's a createBufferFromContents helper we can use.

src/webgpu/api/operation/command_buffer/image_copy.spec.ts
```diff
@@ -627,6 +629,250 @@ class ImageCopyTest extends GPUTest {
...
+        }),
+        entryPoint: 'main',
+        targets: [
+          {
+            // As we implement "rendering one bit in each draw() call" with blending operation
+            // 'add', the format of outputTexture must support blending.
+            format: 'r8unorm',
+            blend: {
+              color: { srcFactor: 'one', dstFactor: 'one', operation: 'add' },
+              alpha: { srcFactor: 'one', dstFactor: 'one', operation: 'add' },
```
Kangz, line 749: nit: I think we might be able to skip alpha since we're r8unorm that has no alpha channel.

src/webgpu/api/operation/command_buffer/image_copy.spec.ts
```diff
@@ -627,6 +629,250 @@ class ImageCopyTest extends GPUTest {
...
+        format: stencilTextureFormat,
+        stencilFront: {
+          compare: 'equal',
+        },
+        stencilBack: {
+          compare: 'equal',
+        },
+      };
+      for (let stencilBitIndex = 0; stencilBitIndex < stencilBitCount; ++stencilBitIndex) {
+        const depthStencilState = depthStencilStateBase;
```
Kangz, line 826: u-nit: I think it might be more idiomatic JS to first create the full descriptor and continuously update it here. Especially since the rest of the GPUDepthStencilState doesn't change:

```
renderPipelineDescriptor.depthStencil.stencilReadMask = 1 << stencilBitIndex;
const renderPipeline = this.device.createRenderPipeline(renderPipelineDescriptor);
```

src/webgpu/api/operation/command_buffer/image_copy.spec.ts
Kangz: Ah sorry I wasn't clear. In the file we have a lot of tests exercising many options of copies: bytesPerRow, rowsPerImage, sub copies, copies at an offset, etc. It would be really nice to find a way to "just" add the stencil copy tests to the parametrization of existing tests so that we get all the amazing coverage.

Kangz (commented):
Stencil checking algorithm looks much better, thank you!

src/webgpu/api/operation/command_buffer/image_copy.spec.ts
Jiawei-Shao: Oh we cannot skip setting alpha as GPUBlendState.alpha is not optional.

src/webgpu/api/operation/command_buffer/image_copy.spec.ts
Jiawei-Shao: Oh the reason I prefer adding a separated test is that the copies with depth/stencil formats are a bit different as we can only do full subresource copies with depth/stencil formats and the way to verify the result is totally different from other copies. What do you think?

src/webgpu/api/operation/command_buffer/image_copy.spec.ts
kainino0x: That's right. You can set it to `{}` though (if you want).

kainino0x:
Please let me know if a second review is needed, I am skipping this for now to burn down my review inbox.

src/webgpu/api/operation/command_buffer/image_copy.spec.ts
Kangz: @kainino0x FYI this thread raises interesting points about helpers not handling the notion of aspects as consistently as they should.

Long term I think it would be better to have the tests be generic on the aspects etc, but that can be left for later. The solution you outlined with similar looking tests for various aspects being grouped together sounds good. (although it may be better to use `.expand(commonOffsetsAndSizeParamExpander)`). To make code duplication of code minimal, do you think we could have a helper like `uploadTextureAndVerifyCopy` but for depth / stencil?

All of the above said, I think we can land this to make forward progress first, then add the rest of the tests. This is a good step.

Kangz (approved):

src/webgpu/api/operation/command_buffer/image_copy.spec.ts
```diff
@@ -627,6 +630,220 @@ class ImageCopyTest extends GPUTest {
         unreachable();
     }
   }
+
+  async checkStencilTextureContent(
```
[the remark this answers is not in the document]
kainino0x:
```suggestion
  // TODO(crbug.com/dawn/868): Revisit this when consolidating texture helpers.
  async checkStencilTextureContent(
```

src/webgpu/api/operation/command_buffer/image_copy.spec.ts
kainino0x: The aspects problem is an issue with both the helpers and the capabilities tables. Agree we definitely need to fix it when fixing the helpers.

Also agree with using simple local helpers (just blocks of code extracted out into functions) to reduce code duplication, but it's not critical. Jiawei and I discussed this a bit over chat in the past few days as well.

> code duplication of code

"code duplication," i.e., duplication of the word "code" 👍 🤣

Jiawei-Shao: Thanks for your understanding, @kainino0x and @Kangz!

I plan to consider extracting the common code paths and better organizing all the buffer-to-texture copy tests after adding b2t-stencil-copy, t2b-stencil-copy and t2b-depth-copy.

kainino0x:
@Jiawei-Shao this is ready to land right? The refactor will be a followup? (https://github.com/gpuweb/cts/pull/606#discussion_r670932971)

Jiawei-Shao:
Hi @kainino0x, sorry that I am still preparing some changes locally and I will update them soon.  This PR is quite important to me as I need this PR to set up the basic infrastructure for other depth-stencil copy tests.

src/webgpu/api/operation/command_buffer/image_copy.spec.ts
Jiawei-Shao: Hi @Kangz and @kainino0x, I've updated the PR. PTAL, thanks!

Kangz:
Sorry I didn't get to re-review today. (Github doesn't send an email on just new commits I think). @kainino0x would you be able to take a look? Otherwise I'll do it tomorrow.

kainino0x (approved):

kainino0x:
LGTM, please merge if ready

PR #609, opened 2021-06-30
Test that array values are preserved across multiple WRITE mappings

toji:
Completes a TODO in api,operation,buffers,map:*

Pushing this test separately because it does not currently pass in Chrome, which appears to return a newly cleared TypedArray with every mapping. (Firefox appears to not be able to run the CTS at all.) So I would appreciate some extra review attention ensuring the check here is valid.

<hr>


kainino0x (approved):
LGTM!

PR #610, opened 2021-06-30
Test that buffer writes are preserved after another range is mapped.

toji:
Completes a TODO in api,operation,buffers,map:*

Uses `mappedAtCreation` or `mapAsync` to write to various ranges of variously-sized buffers, then uses `mapAsync` to map a different range and zero it out. Finally uses `expectGPUBufferValuesEqual` to verify that contents originally written outside the
second mapped range were not altered.

<hr>

### DIFF
--- a/src/webgpu/api/operation/buffers/map.spec.ts
+++ b/src/webgpu/api/operation/buffers/map.spec.ts
@@ -10,8 +10,6 @@ mapRegionBoundModes is used to get mapRegion from range:
  - default-expand: expand mapRegion to buffer bound by setting offset/size to undefined
  - explicit-expand: expand mapRegion to buffer bound by explicitly calculating offset/size
  - minimal: make mapRegion to be the same as range which is the minimal range to make getMappedRange input valid
-
-TODO: Test that ranges not written preserve previous contents.
 `;
 
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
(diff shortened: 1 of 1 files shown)


src/webgpu/api/operation/buffers/map.spec.ts
```diff
@@ -99,6 +98,73 @@ g.test('mapAsync,write')
...
+    u
+      .beginSubcases()
+      .combine('mappedAtCreation', [false, true])
+      .combineWithParams([
+        { size: 12, range1: [], range2: [8], verify: [0, 8] },
+        { size: 12, range1: [], range2: [0, 8], verify: [8] },
+        { size: 12, range1: [0, 8], range2: [8], verify: [0, 8] },
+        { size: 12, range1: [8], range2: [0, 8], verify: [8] },
+        { size: 28, range1: [], range2: [8, 8], verify: [0, 8] },
+        { size: 28, range1: [], range2: [8, 8], verify: [16] },
```
kainino0x, line 118: nit: The three regions make this a little hard to reason about (to make sure the right ranges are verified in each test case), and also mean you have to have two cases like this to verify before and after the second range. Maybe you could emulate all of the updates in the "expected" buffer and just check the entire buffer at the end of the test? up to you though, coverage LGTM as is.

kainino0x (approved):

src/webgpu/api/operation/buffers/map.spec.ts
toji: Refactored, thanks!

PR #614, opened 2021-07-02
Implement api,operation,command_buffer,programmable,state_tracking:*

toji:
Marking as draft initially while I work through this to get some feedback.

This has been a difficult set of tests for me to figure out how to effectively parameterize. I've added a couple of tests to get some feedback on how their being handled and solicit ideas on how to make them more flexible and extend to more scenarios. I'll be working on adding to the test set in the meantime regardless of whether or not anyone comments.

<hr>


kainino0x:
I think these look reasonable so far. It's definitely going to be necessary to do things like set the same bind group index multiple times, to make sure it gets updated properly, and interleave pipeline setting with bindgroup setting, etc. So I definitely see how that gets hard to parameterize.

One thing to keep in mind is, if things become too hard to parameterize, it's totally reasonable to just have more hand-written or copy-pasted code. Or things like a big switch block in the middle of the test, switched on a param.

```ts
g.test('switch_on_number')
  .params(u => u.combine('whichCase', [0, 1, 2] as const))  // note "as const"
  .fn(t => {
    // some common logic

    switch (t.params.whichCase) {  // "as const" above makes the type 0|1|2
      case 0:
        // arbitrary stuff
        break;
      case 1:
        // arbitrary stuff
        break;
      case 2:
        // arbitrary stuff
        break;
      case 3:  // <- error because it's not in the params list, guards against forgetting
        // arbitrary stuff
        break;
    }

    // more common logic
  });
```

toji:
Alright, finally looping back to this (a month an a half later). The initial set of tests has been expanded to test things like setting a bind group multiple times, sharing bind groups between pipelines, and setting the bind groups before the pipeline. Also (the biggest change) is that it now bind group state in Compute passes, Render passes, and Render Bundles. PTAL!

src/webgpu/api/operation/command_buffer/programmable/programmable_state_test.ts
```diff
@@ -0,0 +1,223 @@
...
+  a: number;
+  b: number;
+  out: number;
+}
+
+export enum PassType {
+  Compute = 'compute',
+  Render = 'render',
+  RenderBundle = 'renderBundle',
+}
```
kainino0x, line 16: Usually we just use string constant types instead of enums, as it generally cuts down on the amount of boilerplate:
`export type PassType = 'compute' | 'render' | 'renderBundle';`
but also consider using the existing `ProgrammableEncoderType` which is basically the same thing.
(and see below about `createEncoder`)

src/webgpu/api/operation/command_buffer/programmable/programmable_state_test.ts
```diff
@@ -0,0 +1,223 @@
...
+          code: wgslShaders.fragment,
+        }),
+        entryPoint: 'frag_main',
+        targets: [{ format: 'rgba8unorm' }],
+      },
+      primitive: { topology: 'point-list' },
+    });
+  }
+
+  beginSimplePass(passType: PassType): GPUProgrammablePassEncoder {
```
kainino0x, line 150: can probably use the `createEncoder` helper here. Unfortunately it looks like it will need to be moved from ValidationTest to GPUTest, but that's probably the right thing to do.

src/webgpu/api/operation/command_buffer/programmable/programmable_state_test.ts
```diff
@@ -0,0 +1,223 @@
...
+            visibility: GPUShaderStage.COMPUTE | GPUShaderStage.FRAGMENT,
+            buffer: { type: 'storage' },
+          },
+        ],
+      });
+    }
+    return this.commonBindGroupLayout;
+  }
+
+  createBufferWithValue(initValue: number): GPUBuffer {
```
kainino0x, line 37: use GPUTest.makeBufferWithContents

src/webgpu/api/operation/command_buffer/programmable/programmable_state_test.ts
```diff
@@ -0,0 +1,223 @@
...
+      const renderPass = this.beginSimpleRenderPass(this.encoder);
+      renderPass.executeBundles([renderBundle]);
+      renderPass.endPass();
+    }
+
+    this.device.queue.submit([this.encoder.finish()]);
+    this.encoder = null;
+  }
+
+  verifyData(buffer: GPUBuffer, expectedValue: number) {
```
kainino0x, line 218: nit: I think it would be more readable to simply do this inline instead of adding this method:
`this.expectGPUBufferValuesEqual(buffer, new Uint32Array([expectedValue]))`

kainino0x (commented):
partial review (only looked at the one file so far)

src/webgpu/api/operation/command_buffer/programmable/programmable_state_test.ts
toji: Moved this and did all the refactoring that entails. Increased the size of this patch considerably, but I agree that it's better to reuse existing helpers.

src/webgpu/api/operation/command_buffer/programmable/programmable_state_test.ts
```diff
@@ -0,0 +1,169 @@
...
+  a: number;
+  b: number;
+  out: number;
+}
+
+type PipelineByEncoderType<T extends EncoderType> = {
+  'non-pass': null;
+  'compute pass': GPUComputePipeline;
+  'render pass': GPURenderPipeline;
+  'render bundle': GPURenderBundleEncoder;
```
kainino0x, line 14:
```suggestion
  'render bundle': GPURenderPipeline;
```
(Though the presence of this error shows that the types here probably aren't helping - maybe usages of PipelineByEncoderType could just be replaced with `GPUComputePipeline | GPURenderPipeline` since the `as` casts are already there. Unless you want to try to make it more typesafe but I'm not sure offhand how to do that.)

src/webgpu/api/operation/command_buffer/programmable/programmable_state_test.ts
```diff
@@ -0,0 +1,169 @@
...
+      entries: [{ binding: 0, resource: { buffer } }],
+    });
+  }
+
+  // Create a compute pipeline that performs an operation on data from two bind groups,
+  // then writes the result to a third bind group.
+  createBindingStatePipeline<T extends EncoderType>(
+    encoderType: T,
+    groups: BindGroupIndices,
+    algorthim: String = 'a.value - b.value'
```
kainino0x, line 55: nit: `algorthim` -> `algorithm`

also `String` -> `string`

src/webgpu/api/operation/command_buffer/programmable/programmable_state_test.ts
```diff
@@ -0,0 +1,169 @@
...
+      case 'render bundle':
+        return this.createBindingStateRenderPipeline(groups, algorthim) as PipelineByEncoderType<T>;
+      default:
+        return null as PipelineByEncoderType<T>;
+    }
+  }
+
+  createBindingStateComputePipeline(
+    groups: BindGroupIndices,
+    algorthim: String = 'a.value - b.value'
```
kainino0x, line 72: ^

src/webgpu/api/operation/command_buffer/programmable/programmable_state_test.ts
```diff
@@ -0,0 +1,169 @@
...
+          code: wgsl,
+        }),
+        entryPoint: 'main',
+      },
+    });
+  }
+
+  createBindingStateRenderPipeline(
+    groups: BindGroupIndices,
+    algorthim: String = 'a.value - b.value'
```
kainino0x, line 103: ^

src/webgpu/api/operation/command_buffer/programmable/state_tracking.spec.ts
```diff
@@ -10,6 +10,266 @@ TODO: for each programmable pass encoder {compute pass, render pass, render bund
...
 
 import { makeTestGroup } from '../../../../../common/framework/test_group.js';
-import { GPUTest } from '../../../../gpu_test.js';
+import { GPUConst } from '../../../../constants.js';
+import { ProgrammableStateTest } from './programmable_state_test.js';
 
-export const g = makeTestGroup(GPUTest);
+export const g = makeTestGroup(ProgrammableStateTest);
+
+const kEncoderTypes = ['compute pass', 'render pass', 'render bundle'] as const;
```
kainino0x, line 18: nit: I think you can import `kProgrammableEncoderTypes` instead of redefining this list

src/webgpu/api/operation/command_buffer/programmable/programmable_state_test.ts
```diff
@@ -0,0 +1,169 @@
...
+    }
+    return this.commonBindGroupLayout;
+  }
+
+  createBufferWithValue(initValue: number): GPUBuffer {
+    return this.makeBufferWithContents(
+      new Int32Array([initValue]),
+      GPUBufferUsage.COPY_SRC | GPUBufferUsage.STORAGE
+    );
+  }
```
kainino0x, line 41: helper unused?

src/webgpu/api/operation/command_buffer/programmable/state_tracking.spec.ts
```diff
@@ -10,6 +10,266 @@ TODO: for each programmable pass encoder {compute pass, render pass, render bund
 `;
```
kainino0x, line 10: Need to clean up this TODO (I think it's sufficiently done now - though maybe there should be a new TODO for trying to do really complex or pseudorandom sequences of things? rockot may have also thought about this in stress tests, I don't remember)

kainino0x (approved):

src/webgpu/api/operation/command_buffer/programmable/programmable_state_test.ts
toji: I'm actually not seeing an error here? (And the CI checks appear to be passing now.)

kainino0x: I mean that this looks like a mistake (unless I'm wrong), yet there is no error because this type isn't actually taking effect since it is `as` cast everywhere it's used.

toji: Ah, thanks for pointing that out! It was indeed a typo. I reverted to `GPUComputePipeline | GPURenderPipeline` for simplicity.

PR #615, opened 2021-07-02
operation/vertex_state: split runTest into smaller functions

Kangz:
This will allow using the individual components in tests that need
custom handling beyond what runTest can do.

@kainino0x PTAL at the Typescriptness. The types of the helper functions are pretty massive but I couldn't get them to be smaller (I wanted to use generic "extends" but that requires declaring interfaces etc).

<hr>

### DIFF
--- a/src/webgpu/api/operation/vertex_state/correctness.spec.ts
+++ b/src/webgpu/api/operation/vertex_state/correctness.spec.ts
@@ -510,26 +510,35 @@ struct VSOutputs {
     }
   }
 
-  runTest(
-    buffers: VertexState<
+  createPipelineAndTestData<V, A>(
+    state: VertexState<
       {
         stepMode: GPUInputStepMode;
         arrayStride: number;
-        vbOffset?: number;
-      },
+      } & V,
       {
         offset: number;
         format: GPUVertexFormat;
-        shaderComponentCount?: number;
-      }
+      } & A
     >,
-    // Default to using 20 vertices and 20 instances so that we cover each of the test data at least
-    // once (at the time of writing the largest testData has 16 values).
-    vertexCount: number = 20,
-    instanceCount: number = 20
-  ) {
+    vertexCount: number,
+    instanceCount: number
+  ): {
+    pipeline: GPURenderPipeline;
+    testData: VertexState<
+      {
+        stepMode: GPUInputStepMode;
+        arrayStride: number;
+      } & V,
+      {
+        offset: number;
+        format: GPUVertexFormat;
+      } & A &
+        TestData
+    >;
+  } {
     // Gather the test data and some additional test state for attribs.
-    const pipelineAndTestState = mapStateAttribs(buffers, (buffer, attrib) => {
+    const pipelineAndTestState = mapStateAttribs(state, (buffer, attrib) => {
       const maxCount = buffer.stepMode === 'instance' ? instanceCount : vertexCount;
       const formatInfo = kVertexFormatInfo[attrib.format];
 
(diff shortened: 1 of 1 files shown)


src/webgpu/api/operation/vertex_state/correctness.spec.ts
```diff
@@ -510,26 +515,16 @@ struct VSOutputs {
     }
   }
 
-  runTest(
-    buffers: VertexState<
-      {
-        stepMode: GPUInputStepMode;
-        arrayStride: number;
-        vbOffset?: number;
-      },
```
kainino0x, line 519: note: you can get these to flow on one line by replacing `{\n` with `{` and then reformatting. I did this in the definition of VertexLayoutState.

src/webgpu/api/operation/vertex_state/correctness.spec.ts
```diff
@@ -562,15 +562,21 @@ struct VSOutputs {
...
+    return this.device.createBindGroup({
       layout: pipeline.getBindGroupLayout(0),
       entries: bgEntries,
     });
+  }
 
+  createVertexBuffers(
+    state: VertexLayoutState<{ vbOffset?: number }, TestData>,
+    vertexCount: number = 20,
+    instanceCount: number = 20
```
kainino0x, line 574: These two optional arguments are never passed. I assume they should be non-optional and passed through from runTest? [1]

src/webgpu/api/operation/vertex_state/correctness.spec.ts
```diff
@@ -595,7 +601,23 @@ struct VSOutputs {
...
-    // Run the test shader.
+    return vertexBuffers;
+  }
+
+  runTest(
+    buffers: VertexLayoutState<{ vbOffset?: number }, { shaderComponentCount?: number }>,
+    // Default to using 20 vertices and 20 instances so that we cover each of the test data at least
+    // once (at the time of writing the largest testData has 16 values).
+    vertexCount: number = 20,
+    instanceCount: number = 20
```
kainino0x, line 612: These also appear to never be used, but if OK if you intend to use them later

src/webgpu/api/operation/vertex_state/correctness.spec.ts
```diff
@@ -595,7 +601,23 @@ struct VSOutputs {
...
+  ) {
+    const { testData, pipeline } = this.createPipelineAndTestData(
+      buffers,
+      vertexCount,
+      instanceCount
+    );
+    const expectedDataBG = this.createExpectedBG(testData, pipeline);
+    const vertexBuffers = this.createVertexBuffers(testData);
```
kainino0x, line 620: [1]

kainino0x (approved):
Mostly LGTM, with one issue. I also pushed an extra commit which factors out the VertexState with stepMode/arrayStride/format/offset, PTAL

src/webgpu/api/operation/vertex_state/correctness.spec.ts
Kangz: Ah, thanks. It's a bit unfortunate that npm fix has state like that but I'll make sure to test that next time.

Kangz:
Thanks a lot for the simplification suggestion!

src/webgpu/api/operation/vertex_state/correctness.spec.ts
Kangz: I'm 90% sure it will be used later.

PR #622, opened 2021-07-14
Change blending test target texture format to rgba16float and add helpers for float16

shrekshao:
https://gpuweb.github.io/gpuweb/#plain-color-formats indicates `rgba32float` doesn't have blendable `float` capability. Change to `rgba16float`. Also add update helpers to check float16 render target result. Add helpers to convert float16bits to js number as float32.

This will unblock https://bugs.chromium.org/p/dawn/issues/detail?id=726

As we are adding validation for format blending capability.

<hr>

### DIFF
--- a/src/unittests/conversion.spec.ts
+++ b/src/unittests/conversion.spec.ts
@@ -8,14 +8,17 @@ import { UnitTest } from './unit_test.js';
 export const g = makeTestGroup(UnitTest);
 
 const cases = [
-  [0b0011110000000000, 1],
-  [0b0000010000000000, 0.00006103515625],
-  [0b0011010101010101, 0.33325195],
-  [0b0111101111111111, 65504],
-  [0b0, 0],
-  [14336, 0.5],
-  [12902, 0.1999512],
-  [22080, 100],
+  [0b0_01111_0000000000, 1],
+  [0b0_00001_0000000000, 0.00006103515625],
+  [0b0_01101_0101010101, 0.33325195],
+  [0b0_11110_1111111111, 65504],
+  [0b0_00000_0000000000, 0],
+  [0b0_01110_0000000000, 0.5],
+  [0b0_01100_1001100110, 0.1999512],
+  [0b0_01111_0000000001, 1.00097656],
+  [0b0_10101_1001000000, 100],
+  [0b1_01100_1001100110, -0.1999512],
+  [0b1_10101_1001000000, -100],
 ];
 
 g.test('conversion,float16BitsToFloat32').fn(t => {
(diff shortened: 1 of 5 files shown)


src/unittests/conversion.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+
+const cases = [
+  [0b0011110000000000, 1],
+  [0b0000010000000000, 0.00006103515625],
+  [0b0011010101010101, 0.33325195],
+  [0b0111101111111111, 65504],
+  [0b0, 0],
+  [14336, 0.5],
+  [12902, 0.1999512],
+  [22080, 100],
```
kainino0x, line 18: nit: IMO unless there's a particular reason to use these decimal representations, it would be clearest to use exactly-16-bit long `0b` representations for all of the cases. Also you can use underscore as a separator, like: `0b0011_1100_0000_0000` or even `0b0_01111_0000000000` to show the separation between sign, exponent, and mantissa

src/webgpu/gpu_test.ts
```diff
@@ -439,13 +448,47 @@ export class GPUTest extends Fixture {
...
+    src: GPUTexture,
+    format: SizedTextureFormat,
+    { x, y }: { x: number; y: number },
+    {
+      exp,
+      slice = 0,
+      layout,
+      generateWarningOnly = false,
+    }: {
+      exp: [TypedArrayBufferView, TypedArrayBufferView];
```
kainino0x, line 472: I think these could just be Uint16Array specifically, since anything else wouldn't make much sense.

kainino0x (approved):
LGTM

src/webgpu/gpu_test.ts
shrekshao: Tried with `Uint16Array` first but the typescript didn't compile for that (or is the vscode intellisense mark it red?)
`TypedArrayBufferView = ... | Uint16Array | ...` it seems reasonable to use Uint16Array to me. 

Any other syntax we need to make it assignable to
```
checkElementsBetweenFn?: (
        actual: TypedArrayBufferView,
        expected: readonly [TypedArrayBufferView, TypedArrayBufferView]
      ) => Error | undefined;
```
?

shrekshao: Oops. I think this is for the `exp: [Uint16Array, Uint16Array]` instead. fixed

Still helps knowing why typescript wouldn't allow 
```
export function checkElementsFloat16Between(
  actual: Uint16Array,
  expected: readonly [Uint16Array, Uint16Array]
): ErrorWithExtra | undefined
```
resolves to 
```
checkElementsBetweenFn?: (
        actual: TypedArrayBufferView,
        expected: readonly [TypedArrayBufferView, TypedArrayBufferView]
      ) => Error | undefined;
```
(ErrorWithExtra => Error is not a problem)

kainino0x: It's because if you take `checkElementsFloat16Between` and give it the type signature of `checkElementsBetweenFn`, then a caller to `checkElementsBetweenFn` could pass in other typed array types, which would violate the type constraints of `checkElementsFloat16Between`. This is a concept in polymorphism called "variance" (covariance / contravariance)

PR #624, opened 2021-07-14
Add Validation tests for CopyExternalImageToTexture()

shaoboyan:
This PR is based on previous validation tests of
CopyExternalImageToTexture.
The PR added cases to cover canvas/offscreenCanvas as input and more
validation error cases.

<hr>


shaoboyan:
@kainino0x  PTAL, thanks!

Gruntfile.js
```diff
@@ -139,11 +157,13 @@ module.exports = function (grunt) {
 
   grunt.registerTask('build-standalone', 'Build out/ (no checks, no WPT)', [
     'run:build-out',
+    'run:copy-assets',
     'run:generate-version',
     'run:generate-listings',
   ]);
   grunt.registerTask('build-wpt', 'Build out/ (no checks)', [
     'run:build-out-wpt',
+    'run:copy-assets-wpt',
```
shaoboyan, line 166: This has not been tested...

Gruntfile.js
```diff
@@ -139,11 +157,13 @@ module.exports = function (grunt) {
 
   grunt.registerTask('build-standalone', 'Build out/ (no checks, no WPT)', [
     'run:build-out',
+    'run:copy-assets',
```
shaoboyan, line 160: Run local test with this command, and ensure it works

kainino0x:
> @kainino0x PTAL, thanks!

Please add me as a reviewer, that will make sure it goes into my "Review" inbox/to-do list.

src/webgpu/api/validation/queue/copyToTexture/CopyExternalImageToTexture.spec.ts
```diff
@@ -150,19 +164,195 @@ class CopyImageBitmapToTextureTest extends ValidationTest {
...
+
+    const crossOriginUrl = 'https://get.webgl.org/conformance-resources/opengl_logo.jpg';
+    const originCleanUrl = '../out/resources/Di-3d.png';
+
+    const img = document.createElement('img');
+    img.src = isOriginClean ? originCleanUrl : crossOriginUrl;
+    try {
+      await awaitOrTimeout(img.decode());
+    } catch (e) {
+      t.skip('Cannot load image');
```
kainino0x, line 309: This would indicate something went wrong accessing the image, so should probably be a little louder than just a skip:
- If `isOriginClean`, it's a local resource, so it should be a hard error: `throw e;`
- If `!isOriginClean`, something went wrong accessing get.webgl.org, so `t.warn` *and* `t.skip` here.

src/webgpu/api/validation/queue/copyToTexture/CopyExternalImageToTexture.spec.ts
```diff
@@ -150,19 +164,195 @@ class CopyImageBitmapToTextureTest extends ValidationTest {
...
+      sourceImage === 'offscreenCanvas'
+        ? t.getOffscreenCanvas(1, 1)
+        : document.createElement('canvas');
+
+    const ctx = canvas.getContext('2d');
+    if (ctx === null) {
+      t.skip('Cannot get 2d context');
+    } else {
+      ctx.drawImage(img, 0, 0);
+    }
```
kainino0x, line 322: ctx should never be null except in browsers that don't support "2d" on offscreen canvas. So should fail the test (with `assert` or `unreachable`) if it's not offscreen.

src/webgpu/api/validation/queue/copyToTexture/CopyExternalImageToTexture.spec.ts
```diff
@@ -150,19 +164,195 @@ class CopyImageBitmapToTextureTest extends ValidationTest {
...
+
+    let externalImage: HTMLCanvasElement | OffscreenCanvas | ImageBitmap = canvas;
+    if (sourceImage === 'imageBitmap') {
+      externalImage = await createImageBitmap(canvas);
```
kainino0x, line 326: I was thinking the imagebitmap case would be creating the imagebitmap directly from the image, rather than going through canvas. But actually, we should test both!

src/webgpu/api/validation/queue/copyToTexture/CopyExternalImageToTexture.spec.ts
```diff
@@ -184,15 +374,80 @@ g.test('source_imageBitmap,state')
...
       closed ? 'InvalidStateError' : ''
     );
   });
 
+g.test('source_offscreenCanvas,state')
+  .desc(
+    `
+  Test OffscreenCanvas as source image in state [valid, detached].
+
+  Transfer OffsreenCanvas to worker will detach the OffscreenCanvas
```
kainino0x, line 390:
```suggestion
  Transfer OffscreenCanvas to worker will detach the OffscreenCanvas
```

src/webgpu/api/validation/queue/copyToTexture/CopyExternalImageToTexture.spec.ts
```diff
@@ -184,15 +374,80 @@ g.test('source_imageBitmap,state')
...
+    const { detached, copySize } = t.params;
+    const offscreenCanvas = t.getOffscreenCanvas(1, 1);
+    const dstTexture = t.device.createTexture({
+      size: { width: 1, height: 1, depthOrArrayLayers: 1 },
+      format: 'bgra8unorm',
+      usage: GPUTextureUsage.COPY_DST | GPUTextureUsage.RENDER_ATTACHMENT,
+    });
+
+    if (detached) {
+      const workerCode =
```
kainino0x, line 416: You don't actually need to create a worker in order to post to it, you can just create a MessageChannel and post to it:

```ts
const messageChannel = new MessageChannel();
messageChannel.port1.postMessage(offscreenCanvas, [offscreenCanvas]);
```

src/webgpu/api/validation/queue/copyToTexture/CopyExternalImageToTexture.spec.ts
```diff
@@ -184,15 +374,80 @@ g.test('source_imageBitmap,state')
...
+        "self.onmessage = function(e) {e.data.canvas.getContext('2d'); poseMessage('get 2d context');};";
+      const blob = new Blob([workerCode], { type: 'text/javascript' });
+      const url = URL.createObjectURL(blob);
+      const worker = new Worker(url);
+      worker.postMessage({ canvas: offscreenCanvas }, [offscreenCanvas]);
+      worker.onmessage = function (e) {
```
kainino0x, line 422: You don't have to wait for the message before testing InvalidStateError. The OffscreenCanvas becomes detached instantly inside postMessage.

src/webgpu/api/validation/queue/copyToTexture/CopyExternalImageToTexture.spec.ts
```diff
@@ -207,10 +462,47 @@ g.test('destination_texture,state')
...
+
+    t.runTest({ source: imageBitmap }, { texture: dstTexture }, copySize, dimension === '2d');
   });
 
 g.test('destination_texture,usage')
+  .desc(
+    `
+  Test dst texture usages
+
+  Check that an error is generated when texture is created witout usage COPY_DST | RENDER_ATTACHMENT.
```
kainino0x, line 503:
```suggestion
  Check that an error is generated when texture is created without usage COPY_DST | RENDER_ATTACHMENT.
```

kainino0x (commented):
LGTM overall, few comments that will simplify things

src/webgpu/api/validation/queue/copyToTexture/CopyExternalImageToTexture.spec.ts
shaoboyan: If I don't wait this, I'll get another error about the source image doesn't have correct context. Maybe in impl we should change the validation orders?

kainino0x: Ohh, I see that you are creating a context on the offscreen canvas on the receiving thread.

Yes, that shouldn't be necessary. In the [html spec](https://html.spec.whatwg.org/multipage/canvas.html#:~:text=OffscreenCanvas%20objects%20are%20transferable.%20Their%20transfer%20steps%2C%20given%20value%20and%20dataHolder%2C%20are%20as%20follows%3A), the "context mode" of the source OffscreenCanvas object is set to "detached" inside postMessage, so, regardless of whether the receiver created a context, it should fail in the same way. It sounds like our implementation doesn't mirror the spec exactly, so I guess it's best to test both cases: one where the receiver creates a context prior to runTest, and one where it doesn't.

kainino0x: However you can still do that without launching a worker: just grab the message from port2 (on the same thread instead of a worker) and getContext on that.

src/webgpu/api/validation/queue/copyToTexture/CopyExternalImageToTexture.spec.ts
```diff
@@ -1,36 +1,14 @@
...
-    - Check that an error is generated when sample count is not 1.
-    - Check that an error is generated when mipLevel is too large.
-    - Check that an error is generated when texture format is not valid.
-
-- For copySize:
-  - No-op copy shouldn't throw any exception or return any validation error.
-  - Check that an error is generated when destination.texture.origin + copySize is too large.
-
-TODO: copying into slices of 2d array textures. 1d and 3d as well if they're not invalid.
+copyExternalImageToTexture Validation Tests in Queue.
```
kainino0x, line 2: I remembered that for canvas elements we also need to make sure they are not 'placeholder', i.e. they are not a canvas that has had transferControlToOffscreen called on it. Please add a TODO for this - and you can follow up with an implementation in another PR if you'd like.

src/webgpu/api/validation/queue/copyToTexture/CopyExternalImageToTexture.spec.ts
shaoboyan: < Yes, that shouldn't be necessary. In the html spec, the "context mode" of the source OffscreenCanvas object is set to 
<" detached" inside postMessage, so, regardless of whether the receiver created a context, it should fail in the same way.  
< sounds like our implementation doesn't mirror the spec exactly.
 Yes, I think our implementation checks the context first and then check the detached state, which I think it's not correct.

src/webgpu/api/validation/queue/copyToTexture/CopyExternalImageToTexture.spec.ts
shaoboyan: One remain issue here. I try to construct a test case that port1 listening to the port2 message which do context construction for the offscreenCanvas. But I failed to run it due to the case will terminate(return all of the webgpu resources) before the onmessage callback has been called which runs the `runTest` function.
I tried several ways but I don't think current cts support test case waits for a callback and then terminate.

src/webgpu/api/validation/queue/copyToTexture/CopyExternalImageToTexture.spec.ts
shaoboyan: < one where the receiver creates a context prior to runTest, and one where it doesn't.
Implement the second one. And blocks by  case return webgpu device to pool before callback(onmessage) has been hit.

Gruntfile.js
shaoboyan: Verified by running the build command and find the pictures has been copied to the out-wpt/resources

src/webgpu/api/validation/queue/copyToTexture/CopyExternalImageToTexture.spec.ts
```diff
@@ -187,11 +183,14 @@ export const g = makeTestGroup(CopyExternalImageToTextureTest);
 g.test('source_canvas,contexts')
   .desc(
     `
-  Test HTMLCanvasElement as source image with different contexts.
+  Test HTMLCanvasElement and OffscreenCanvas as source image with different contexts.
```
kainino0x, line 186: The test doesn't appear to test OffscreenCanvas..?

src/webgpu/api/validation/queue/copyToTexture/CopyExternalImageToTexture.spec.ts
```diff
@@ -290,40 +295,75 @@ g.test('source_image,crossOrigin')
...
     try {
-      await awaitOrTimeout(img.decode());
+      await raceWithRejectOnTimeout(img.decode(), timeout_ms, 'load image timeout');
     } catch (e) {
-      t.skip('Cannot load image');
+      if (isOriginClean) {
+        throw e;
+      } else {
+        t.warn('Something wrong happens in get.webgl.org');
+        t.skip('Cannot load image in time');
```
kainino0x, line 322: I guess we should leave a TODO in the description to try to make this test run without accessing the internet. WPT has provisions for this:
https://web-platform-tests.org/writing-tests/server-features.html#tests-involving-multiple-origins

However our test suite (and local test server) are not set up to be able to use them.

src/webgpu/api/validation/queue/copyToTexture/CopyExternalImageToTexture.spec.ts
```diff
@@ -382,13 +422,62 @@ g.test('source_imageBitmap,state')
...
+    const dstTexture = t.device.createTexture({
+      size: { width: 1, height: 1, depthOrArrayLayers: 1 },
+      format: 'bgra8unorm',
+      usage: GPUTextureUsage.COPY_DST | GPUTextureUsage.RENDER_ATTACHMENT,
+    });
+
+    if (state === 'placeholder') {
+      canvas.transferControlToOffscreen();
+    } else {
+      assert(canvas.getContext('2d') !== null);
```
kainino0x, line 463: Just checking, do we have any tests that try to use a canvas with _no_ context as a copy source?

There's a little bit too much different between the "placeholder" and "valid" cases. The placeholder case doesn't create a context at all. So it's possible that the "placeholder" case could be rejected by the browser for the wrong reason, and the test would still pass.

Perhaps adding two additional cases would solve this:
- One where the canvas is not transferred AND it has no context
- One where the canvas is transferred, and the OffscreenCanvas has a context

src/webgpu/api/validation/queue/copyToTexture/CopyExternalImageToTexture.spec.ts
```diff
@@ -405,38 +494,27 @@ g.test('source_offscreenCanvas,state')
...
-          { texture: dstTexture },
-          copySize,
-          true, // No validation errors.
-          'InvalidStateError'
-        );
-      };
+      const messageChannel = new MessageChannel();
+      messageChannel.port1.postMessage(offscreenCanvas, [offscreenCanvas]);
     } else {
       offscreenCanvas.getContext('2d');
```
kainino0x, line 508: Same concern here. Would also be solved by adding a case which creates a context on the transferred canvas (after waiting for it to arrive on port2).

kainino0x (approved):
LGTM, except a few nits and I thought of one new concern.

src/webgpu/api/validation/queue/copyToTexture/CopyExternalImageToTexture.spec.ts
shaoboyan: @kainino0x Yes, but I hit an issue when I tried to add a case which creates a context on the transferred canvas. I'll list this in TODO but left comments here so maybe you could help, here is the issue:
- Register onmessage function for port2 and in the onmessage callback, we do `getContext()` and postMessage
- Register onmessage function for port1 and in the onmessage callback we do `t.runTest`.
- Call `messageChannel.port1.postMessage(offscreenCanvas, [offscreenCanvas])`.
But this model doesn't works because when the port1 onmessage callback has been triggered, the Fixture class already return the webgpu device to the pool. With some debugging work, I find that after executing the `messageChannel.port1.postMessage`, case will hit the end of the closure, and this assert(https://github.com/gpuweb/cts/blob/main/src/webgpu/gpu_test.ts#L49) fails when running `t.runTest` in the `port1.onmessage`.
I'm not quite sure whether current cts framework can wait for the callback and then terminate.

src/webgpu/api/validation/queue/copyToTexture/CopyExternalImageToTexture.spec.ts
shaoboyan: Oh, yes, tried to merge them but think better to split them due to the different valid `getContext` param type.

src/webgpu/api/validation/queue/copyToTexture/CopyExternalImageToTexture.spec.ts
kainino0x: I think you would need something like:

```ts
const port2FirstMessage = new Promise(resolve => {
  messageChannel.port2.onmessage = m => resolve(m);
});

messageChannel.port1.postMessage(offscreenCanvas, [offscreenCanvas]);

const receivedOffscreenCanvas = await port2FirstMessage;
```

shaoboyan: Wow, that's quite clever!  And seems work.

shaoboyan:
@kainino0x  Please help to take another look due to the changes of "detached", thanks!

src/webgpu/api/validation/queue/copyToTexture/CopyExternalImageToTexture.spec.ts
```diff
@@ -425,26 +425,35 @@ g.test('source_imageBitmap,state')
...
   Check whether 'InvalidStateError' is generated when HTMLCanvasElement is
   in 'placeholder' state.
     `
   )
   .params(u =>
     u //
-      .combine('state', ['placeholder', 'valid'])
+      .combine('state', ['nocontext', 'placeholder', 'valid'])
       .beginSubcases()
+      .combine('getContextInOffscreenCanvas', [false, true])
```
kainino0x, line 449: nit: This generates 6 subcases but really there are only 4 different ones, I think. How about just `.combine('state', ['nocontext', 'placeholder-nocontext', 'placeholder-hascontext', 'valid'])`?

src/webgpu/api/validation/queue/copyToTexture/CopyExternalImageToTexture.spec.ts
```diff
@@ -477,43 +504,72 @@ g.test('source_offscreenCanvas,state')
...
   Check whether 'InvalidStateError' is generated when OffscreenCanvas is
   detached.
   `
   )
   .params(u =>
     u //
-      .combine('detached', [false, true])
+      .combine('state', ['nocontext', 'detached', 'valid'])
       .beginSubcases()
+      .combine('getContextInOffscreenCanvas', [false, true])
```
kainino0x, line 522: same

kainino0x (approved):
LGTM with one nit (but OK to just land if you don't think it's worth resolving)

shaoboyan:
Thanks for reviewing!

Issue #504, opened 2021-03-15
webgpu:api,operation,buffers,map_oom:mappedAtCreation,smaller_getMappedRange might be wrong


lucacasonato:
I am working on the implementation of WebGPU in Deno. We want to use cts to test WebGPU in Deno. While working on getting the tests to pass I found a spec vs test inconsistency in `webgpu:api,operation,buffers,map_oom:mappedAtCreation,smaller_getMappedRange`.

This test expects `createBuffer({ mappedAtCreation: true, size: MAX_ALIGNED_SAFE_INTEGER });` to succeed. I don't think this can succeed because the spec says in step 5 that if `mappedAtCreation` is set, to create a `[[mapping]]` buffer of size `descriptor.size` (which is MAX_ALIGNED_SAFE_INTEGER in this test's case). This will obviously fail, because there is no way you can allocate that big of a buffer on any system. I would thus expect that `createBuffer` throws a `RangeError: Array buffer allocation failed`.

Is the test wrong, or the spec incomplete / wrong here?

lucacasonato:
Same actually goes for the `webgpu:api,operation,buffers,map_oom:mappedAtCreation:oom=true` tests. It only checks for a `RangeError` at `getMappedRange`, not at `createBuffer`.

austinEng:
The spec might be wrong. Resolution is documented here: https://github.com/gpuweb/gpuweb/issues/872#issuecomment-661306882 but we might need to check the meeting notes for more information.

It's my understanding is that:
 - `createBuffer` never fails synchronously. Though it may generate an OutOfMemory error asynchronously into the current error scope if the data backing cannot be allocated. This is the simplify error handling for an application. `createBuffer` has one failure mode that is asynchronous.
 -  `getMappedRange` "never" fails. If a shared GPU/CPU allocation can't be made, then `createBuffer` would have generated an async OOM error, and `getMappedRange` should return a dummy allocation. It only synchronously fails with `RangeError` because the JS wrapper can't be made for that allocation or the dummy allocation is so large that allocating it fails.

lucacasonato:
Thanks! This is definitely not, or at least very underspecified in the spec. I think I have a good path forward though. A few questions come up though:

1. If a GPU OOM error occurs, and a CPU OOM error occurs, we emit two errors into the error scopes, correct? Only the first error would be kept in the error scope, so the second one is presumably lost.
[...]

PR #627, opened 2021-07-14
map_oom.spec.ts fixes

shrekshao:
Fix map_oom.spec.ts to current webgpu spec

* `unmap` on an invalid buffer should be invalid (https://gpuweb.github.io/gpuweb/#dom-gpubuffer-unmap: unmap on an unmapped or destroyed buffer => validation error)
* `getMappedRange` throws `OperationError` instead of `RangeError` (https://gpuweb.github.io/gpuweb/#dom-gpubuffer-getmappedrange)

<hr>

### DIFF
--- a/src/webgpu/api/operation/buffers/map_oom.spec.ts
+++ b/src/webgpu/api/operation/buffers/map_oom.spec.ts
@@ -76,7 +76,7 @@ g.test('mappedAtCreation,full_getMappedRange')
   .desc(
     `Test creating a very large buffer mappedAtCreation buffer should produce
 an out-of-memory error if allocation fails.
-  - Because the buffer can be immediately mapped, getMappedRange does not throw an OperationError. It throws a OperationError because such a
+  - Because the buffer can be immediately mapped, getMappedRange throws an OperationError only because such a
     large ArrayBuffer cannot be created.
   - unmap() should not throw.
   `
@@ -134,18 +134,8 @@ an out-of-memory error if allocation fails.
       oom
     );
 
-    const f = () => buffer.getMappedRange(0, 16);
-
-    let mapping: ArrayBuffer | undefined = undefined;
-    if (oom) {
-      // Smaller range inside a too-big mapping
-      t.shouldThrow('OperationError', () => {
-        mapping = f();
-        t.expect(mapping.byteLength === 16);
-      });
-    } else {
-      mapping = f();
-    }
+    const mapping = buffer.getMappedRange(0, 16);
+    t.expect(mapping.byteLength === 16);
 
     t.expectGPUError('validation', () => buffer.unmap(), oom);
     if (mapping !== undefined) {
--- a/src/webgpu/api/operation/buffers/map_oom.spec.ts
+++ b/src/webgpu/api/operation/buffers/map_oom.spec.ts
@@ -138,7 +138,5 @@ an out-of-memory error if allocation fails.
     t.expect(mapping.byteLength === 16);
 
     t.expectGPUError('validation', () => buffer.unmap(), oom);
-    if (mapping !== undefined) {
-      t.expect(mapping.byteLength === 0, 'Mapping should be detached');
-    }
+    t.expect(mapping.byteLength === 0, 'Mapping should be detached');
   });
--- a/src/webgpu/api/operation/buffers/map_oom.spec.ts
+++ b/src/webgpu/api/operation/buffers/map_oom.spec.ts
@@ -61,10 +61,8 @@ g.test('mapAsync')
         buffer.getMappedRange();
       });
 
-      // Should throw an OperationError because the buffer is already unmapped.
-      t.shouldThrow('OperationError', () => {
-        buffer.unmap();
-      });
+      // Should be a validation error since the buffer is invalid.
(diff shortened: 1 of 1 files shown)


shrekshao:
`smaller_getMappedRange`: https://github.com/gpuweb/cts/issues/504#issuecomment-801324361

src/webgpu/api/operation/buffers/map_oom.spec.ts
```diff
@@ -78,7 +76,7 @@ g.test('mappedAtCreation,full_getMappedRange')
   .desc(
     `Test creating a very large buffer mappedAtCreation buffer should produce
 an out-of-memory error if allocation fails.
-  - Because the buffer can be immediately mapped, getMappedRange does not throw an OperationError. It throws a RangeError because such a
+  - Because the buffer can be immediately mapped, getMappedRange does not throw an OperationError. It throws a OperationError because such a
```
kainino0x, line 79:
```suggestion
  - Because the buffer can be immediately mapped, getMappedRange throws a OperationError only because such a
```

src/webgpu/api/operation/buffers/map_oom.spec.ts
```diff
@@ -125,15 +126,29 @@ an out-of-memory error if allocation fails.
...
-    t.expect(mapping.byteLength === 16);
-    buffer.unmap();
-    t.expect(mapping.byteLength === 0, 'Mapping should be detached');
+    const f = () => buffer.getMappedRange(0, 16);
+
+    let mapping: ArrayBuffer | undefined = undefined;
+    if (oom) {
+      // Smaller range inside a too-big mapping
+      t.shouldThrow('OperationError', () => {
+        mapping = f();
```
kainino0x, line 143: Shouldn't getMappedRange here succeed, and return a dummy allocation?

I know the content process _can_ already know that there was a shmem OOM, but IIRC we don't want to expose the difference between shmem OOM and vram OOM (cc @austinEng in case I'm forgetting something again, this is really hard to hold in my head ^_^).

Unfortunately we can't easily trigger vram OOM without triggering shmem OOM. Actually... we should probably have a stress test that allocates as much vram as it can, in large chunks, until it reaches vram-OOM, then does this test. I have some ideas about this so I'll put notes in src/stress/README.txt.

austinEng: ah, yea it should attempt to allocate dummy memory. It would only throw an exception if `getMappedRange` was too big. I missed this when reviewing the Dawn patch.

we should make sure this test is right, but we can probably defer creating a dummy allocation until post-OT. GIANT mappedAtCreation + getMappedRange is probably a bit niche. I think this is the only scenario where it's needed?

shrekshao: Fixed the test first. Worry about the chrome impl later (Thanks for filing: https://bugs.chromium.org/p/dawn/issues/detail?id=995)

src/webgpu/api/operation/buffers/map_oom.spec.ts
```diff
@@ -125,15 +126,19 @@ an out-of-memory error if allocation fails.
...
-    // Smaller range inside a too-big mapping
     const mapping = buffer.getMappedRange(0, 16);
     t.expect(mapping.byteLength === 16);
-    buffer.unmap();
-    t.expect(mapping.byteLength === 0, 'Mapping should be detached');
+
+    t.expectGPUError('validation', () => buffer.unmap(), oom);
+    if (mapping !== undefined) {
+      t.expect(mapping.byteLength === 0, 'Mapping should be detached');
+    }
```
kainino0x, line 143:
```suggestion
    t.expect(mapping.byteLength === 0, 'Mapping should be detached');
```
No longer needed I think

src/webgpu/api/operation/buffers/map_oom.spec.ts
```diff
@@ -125,15 +126,19 @@ an out-of-memory error if allocation fails.
...
+      oom
     );
 
-    // Smaller range inside a too-big mapping
     const mapping = buffer.getMappedRange(0, 16);
     t.expect(mapping.byteLength === 16);
-    buffer.unmap();
-    t.expect(mapping.byteLength === 0, 'Mapping should be detached');
+
+    t.expectGPUError('validation', () => buffer.unmap(), oom);
```
kainino0x, line 140: Weird that we get a validation error on unmap, and yet it has an effect (detaches the ArrayBuffer). I think we need to examine this corner case in the spec.

kainino0x (approved):

src/webgpu/api/operation/buffers/map_oom.spec.ts
kainino0x: Oh, actually I think the spec is already correct here:
https://gpuweb.github.io/gpuweb/#dom-gpubuffer-unmap

> Note: It is valid to unmap an error GPUBuffer that is mapped at creation because the Content timeline might not know it is an error GPUBuffer.

Incidentally, I think the spec is redundant, it shouldn't be possible for a buffer to be both mapped and destroyed.

PR #628, opened 2021-07-15
implement api,operation,rendering,basic:large_draw:*

krockot:
Note that the indirect-indexed variant fails because the feature is
currently disabled.

<hr>

### DIFF
--- a/src/webgpu/api/operation/rendering/basic.spec.ts
+++ b/src/webgpu/api/operation/rendering/basic.spec.ts
@@ -109,5 +109,24 @@ g.test('fullscreen_quad').fn(async t => {
 });
 
 g.test('large_draw')
-  .desc(`Test reasonably-sized large {draw, drawIndexed} (see also stress tests).`)
+  .desc(
+    `Test reasonably-sized large {draw, drawIndexed} (see also stress tests).
+
+  Tests that draw calls behave reasonably with large vertex counts for
+  non-indexed draws, large index counts for indexed draws, and large instance
+  counts in both cases. Various combinations of these counts are tested with
+  both direct and indrect draw calls.
+
+  Draw call sizes are increased incrementally over these parameters until the
+  completion of a draw call exceeds a fixed time limit of 500ms.
+
+  Params:
+    - indexed= {true, false} - whether to test indexed or non-indexed draw calls
+    - indirect= {true, false} - whether to use indirect or direct draw calls`
+  )
+  .params(u =>
+    u //
+      .combine('indexed', [true, false])
+      .combine('indirect', [true, false])
+  )
   .unimplemented();
--- a/src/webgpu/api/operation/rendering/basic.spec.ts
+++ b/src/webgpu/api/operation/rendering/basic.spec.ts
@@ -124,7 +124,7 @@ g.test('large_draw')
   100ms.
 
   To validate that the drawn vertices actually made it though the pipeline on
-  each draw call, we render a 3x3 target with position the first and last
+  each draw call, we render a 3x3 target with the position the first and last
   vertices of the first and last instances in different respective corners, an
   everything else positioned to cover only one of the intermediate fragments.
 
@@ -138,10 +138,11 @@ g.test('large_draw')
       .combine('indirect', [true, false])
   )
   .fn(async t => {
-    const {indexed, indirect} = t.params;
+    const { indexed, indirect } = t.params;
 
+    const BYTES_PER_ROW = 256;
     const dst = t.device.createBuffer({
-      size: 768,
+      size: 3 * BYTES_PER_ROW,
       usage: GPUBufferUsage.COPY_SRC | GPUBufferUsage.COPY_DST,
     });
 
@@ -151,27 +152,32 @@ g.test('large_draw')
     });
     const writeIndirectParams = (count: number, instanceCount: number) => {
       const params = new Uint32Array(5);
-      params[0] = count;  // Vertex or index count
+      params[0] = count; // Vertex or index count
       params[1] = instanceCount;
-      params[2] = 0;  // First vertex or index
-      params[3] = 0;  // First instance (non-indexed) or base vertex (indexed)
-      params[4] = 0;  // First instance (indexed)
+      params[2] = 0; // First vertex or index
+      params[3] = 0; // First instance (non-indexed) or base vertex (indexed)
+      params[4] = 0; // First instance (indexed)
       t.device.queue.writeBuffer(indirectBuffer, 0, params, 0, 5);
     };
 
+    const MILLION = 1024 * 1024;
+    const MAX_INDICES = 16 * MILLION;
     const indexBuffer = t.device.createBuffer({
-      size: 2**26,  // 64 MB, enough to hold 16M index values.
+      size: MAX_INDICES * Uint32Array.BYTES_PER_ELEMENT,
       usage: GPUBufferUsage.INDEX | GPUBufferUsage.COPY_DST,
     });
-    const M = 1024 * 1024;
-    let indexData = new Uint32Array(1*M);
-    for (let i = 0; i < 16; ++i) {
-      for (let k = 0; k < 1*M; ++k) {
-        indexData[k] = i * M + k;
+    const indexData = new Uint32Array(MILLION);
+    for (let offset = 0; offset < MAX_INDICES; offset += MILLION) {
+      for (let k = 0; k < MILLION; ++k) {
+        indexData[k] = offset + k;
       }
       t.device.queue.writeBuffer(
-          indexBuffer, i * M * Uint32Array.BYTES_PER_ELEMENT, indexData,
-          0, 1*M);
+        indexBuffer,
+        offset * Uint32Array.BYTES_PER_ELEMENT,
+        indexData,
+        0,
+        MILLION
+      );
     }
 
     const colorAttachment = t.device.createTexture({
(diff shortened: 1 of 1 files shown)


src/webgpu/api/operation/rendering/basic.spec.ts
```diff
@@ -129,4 +137,191 @@ g.test('large_draw')
...
+        indexData[k] = offset + k;
+      }
+      t.device.queue.writeBuffer(
+        indexBuffer,
+        offset * Uint32Array.BYTES_PER_ELEMENT,
+        indexData,
+        0,
+        MILLION
+      );
+    }
```
kainino0x, line 181: nit: Rather than a series of writeBuffer calls, I think you could safely do this by specifying `mappedAtCreation: true` in createBuffer, getting a mapped range with `getMappedRange()`, and then writing the test values into that with a (now simpler) for-loop.

src/webgpu/api/operation/rendering/basic.spec.ts
```diff
@@ -129,4 +137,191 @@ g.test('large_draw')
...
+        primitive: { topology: 'point-list' },
+      });
+
+      const encoder = t.device.createCommandEncoder();
+      const pass = encoder.beginRenderPass({
+        colorAttachments: [
+          {
+            view: colorAttachmentView,
+            storeOp: 'store',
+            loadValue: { r: 1.0, g: 0.0, b: 0.0, a: 1.0 },
```
kainino0x, line 241: This will be the "background" pixel color of any pixel that doesn't get a point drawn to it, but I think it's the same color as some of the points. Maybe change it to another color like (0,1,0,1)?

src/webgpu/api/operation/rendering/basic.spec.ts
```diff
@@ -129,4 +137,191 @@ g.test('large_draw')
...
+      pass.endPass();
+      encoder.copyTextureToBuffer(
+        { texture: colorAttachment, mipLevel: 0, origin: { x: 0, y: 0, z: 0 } },
+        { buffer: dst, bytesPerRow: BYTES_PER_ROW },
+        { width: 3, height: 3, depthOrArrayLayers: 1 }
+      );
+      t.device.queue.submit([encoder.finish()]);
+
+      // Red should go 0, 127, 255 from left to right. Blue should to the same
+      // from top to bottom.
```
kainino0x, line 274: Just a reminder for myself when I'm working on texture checking helpers soon.

```suggestion
      // from top to bottom.
      // TODO(crbug.com/dawn/868): See if this can be simplified with new texture checking helpers.
```

src/webgpu/api/operation/rendering/basic.spec.ts
```diff
@@ -129,4 +137,191 @@ g.test('large_draw')
...
+              let highOrMid = select(1.0, 0.5, index == maxIndex - 1u);
+              return select(0.0, highOrMid, index == 0u);
+            }
+
+            [[stage(vertex)]] fn main(
+                [[builtin(vertex_index)]] v: u32,
+                [[builtin(instance_index)]] i: u32)
+                -> Output {
+              let r = selectValue(v, ${numVertices}u);
+              let b = selectValue(i, ${numInstances}u);
```
kainino0x, line 210: optional: I suppose it's probably not a problem, but recompiling for every iteration seems a bit heavy-handed. Maybe we can have a single pipeline and pass numVertices and numInstances in via a uniform buffer?

src/webgpu/api/operation/rendering/basic.spec.ts
```diff
@@ -129,4 +137,191 @@ g.test('large_draw')
...
+              let x = (r * 4.0 - 2.0) / 3.0;
+              let y = (b * 4.0 - 2.0) / -3.0;
+              return Output(vec4<f32>(x, y, 0.0, 1.0),
+                            vec4<f32>(r, 0.0, b, 1.0));
```
kainino0x, line 214: optional: Hm, is it necessary to output different colors for the points? I think the test coverage would be roughly as good if the output color were constant - and it might make the shader execution slightly faster, enabling the test iterations to achieve higher draw call sizes before reaching the time limit.

src/webgpu/api/operation/rendering/basic.spec.ts
```diff
@@ -129,4 +137,191 @@ g.test('large_draw')
...
+      params[2] = 0; // First vertex or index
+      params[3] = 0; // First instance (non-indexed) or base vertex (indexed)
+      params[4] = 0; // First instance (indexed)
+      t.device.queue.writeBuffer(indirectBuffer, 0, params, 0, 5);
+    };
+
+    const MILLION = 1024 * 1024;
+    const MAX_INDICES = 16 * MILLION;
+    const indexBuffer = t.device.createBuffer({
+      size: MAX_INDICES * Uint32Array.BYTES_PER_ELEMENT,
```
kainino0x, line 166: Also, since this is a rather large buffer, it's probably best to `.destroy()` it at the end, rather than letting the garbage collector handle it (since the GC doesn't understand VRAM pressure).

kainino0x (approved):
This looks awesome, thank you! I hope the test well when automated!

I had a few comments, mostly not too consequential.

src/webgpu/api/operation/rendering/basic.spec.ts
```diff
@@ -129,4 +137,191 @@ g.test('large_draw')
...
+      },
+      {
+        numInstances: 2 ** 16,
+        vertexCounts: [2 ** 4, 2 ** 8, 2 ** 10, 2 ** 12, 2 ** 14],
+      },
+      {
+        numInstances: 2 ** 20,
+        vertexCounts: [2 ** 4, 2 ** 8, 2 ** 10],
+      },
+    ];
```
kainino0x, line 315: I think this would work, and save a not-very-useful type annotation:

```suggestion
    const counts = [
      {
        numInstances: 4,
        vertexCounts: [2 ** 10, 2 ** 16, 2 ** 18, 2 ** 20, 2 ** 22, 2 ** 24],
      },
      {
        numInstances: 2 ** 8,
        vertexCounts: [2 ** 10, 2 ** 16, 2 ** 18, 2 ** 20, 2 ** 22],
      },
      {
        numInstances: 2 ** 10,
        vertexCounts: [2 ** 8, 2 ** 10, 2 ** 12, 2 ** 16, 2 ** 18, 2 ** 20],
      },
      {
        numInstances: 2 ** 16,
        vertexCounts: [2 ** 4, 2 ** 8, 2 ** 10, 2 ** 12, 2 ** 14],
      },
      {
        numInstances: 2 ** 20,
        vertexCounts: [2 ** 4, 2 ** 8, 2 ** 10],
      },
    ] as const;
```

(the `as const` might not even be necessary, not sure)

src/webgpu/api/operation/rendering/basic.spec.ts
krockot: That's a fair point. If all fragments are covered, it's a reasonable signal that all vertices were processed.

src/webgpu/api/operation/rendering/basic.spec.ts
krockot: Aha. Missed this wonderful detail of the spec, that mappedAtCreation can be used without MAP_WRITE or MAP_READ. Nice.

src/webgpu/api/operation/rendering/basic.spec.ts
```diff
@@ -129,4 +137,191 @@ g.test('large_draw')
       .combine('indexed', [true, false])
       .combine('indirect', [true, false])
   )
-  .unimplemented();
+  .fn(async t => {
+    const { indexed, indirect } = t.params;
+
+    const BYTES_PER_ROW = 256;
```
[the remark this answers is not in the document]
krockot: Done

src/webgpu/api/operation/rendering/basic.spec.ts
```diff
@@ -129,4 +137,191 @@ g.test('large_draw')
...
+      params[1] = instanceCount;
+      params[2] = 0; // First vertex or index
+      params[3] = 0; // First instance (non-indexed) or base vertex (indexed)
+      params[4] = 0; // First instance (indexed)
+      t.device.queue.writeBuffer(indirectBuffer, 0, params, 0, 5);
+    };
+
+    const MILLION = 1024 * 1024;
+    const MAX_INDICES = 16 * MILLION;
+    const indexBuffer = t.device.createBuffer({
```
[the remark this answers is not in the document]
krockot: Done

src/webgpu/api/operation/rendering/basic.spec.ts
krockot: Made it blue, and shader output yellow

src/webgpu/api/operation/rendering/basic.spec.ts
krockot: Done

src/webgpu/api/operation/rendering/basic.spec.ts
krockot: Done

src/webgpu/api/operation/rendering/basic.spec.ts
krockot: Done

kainino0x:
FYI this test appears to be failing on Mac. I'm adding a suppression in Chromium but I'm not sure what's going on.
https://ci.chromium.org/ui/p/chromium/builders/try/dawn-mac-x64-deps-rel/17526/overview

austinEng:
idk if this is the root cause (I would be surprised because Dawn has *tons* of tests that render points), but TIL:

> The vertex shader must provide [[point_size]], or the point size is undefined
https://developer.apple.com/documentation/metal/mtlprimitivetype/mtlprimitivetypepoint?language=objc

austinEng:
Found the root cause:

WGSL:
> select(a,b,c) evaluates to a when c is true, and b otherwise.

MSL evaluates this flip-flopped

austinEng:
It's WGSL that is backward, not MSL. Fixed in https://github.com/gpuweb/gpuweb/pull/1957 @krockot could you revise this test to switch the order of the operands?

PR #632, opened 2021-07-15
More changes to match spec getMappedRange/unmap behavior

kainino0x:
I _think_ all of these changes are correct, but not positive.

Adds a new test about getMappedRange on buffers that fail validation, which seems reasonable?

<hr>

### DIFF
--- a/src/webgpu/api/operation/buffers/map_oom.spec.ts
+++ b/src/webgpu/api/operation/buffers/map_oom.spec.ts
@@ -61,7 +61,7 @@ g.test('mapAsync')
         buffer.getMappedRange();
       });
 
-      // Should be a validation error since the buffer is invalid.
+      // Should be a validation error since the buffer failed to be mapped.
       t.expectGPUError('validation', () => buffer.unmap());
     } else {
       await promise;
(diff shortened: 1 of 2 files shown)


src/webgpu/api/operation/buffers/map_oom.spec.ts
```diff
@@ -99,14 +99,15 @@ an out-of-memory error if allocation fails.
 
     let mapping: ArrayBuffer | undefined = undefined;
     if (oom) {
-      // Note: It is always valid to get mapped ranges of a GPUBuffer that is mapped at creation, even if it is invalid,
-      // because the Content timeline might not know it is invalid.
-      t.shouldThrow('OperationError', f);
+      // getMappedRange is normally valid on OOM buffers, but this one fails because the
+      // (default) range is too large to create the returned ArrayBuffer.
+      t.shouldThrow('RangeError', f);
```
shrekshao, line 104: Is it because this should be a `RangeError`? (spec now says `OperationError` but that should be changed)

shrekshao (approved):
Thanks for the change

src/webgpu/api/operation/buffers/map_oom.spec.ts
kainino0x: I believe this getMappedRange call is not actually hitting any of the validation cases that result in OperationError in the spec. Instead, it's failing inside:
> Let m be a new ArrayBuffer of size rangeSize

which can fail with a RangeError if the size is too large.

kainino0x: e.g.: `new ArrayBuffer(0x2000000000)`

```
VM67:1 Uncaught RangeError: Array buffer allocation failed
    at new ArrayBuffer (<anonymous>)
    at <anonymous>:1:1
```

PR #636, opened 2021-07-15
Some minor visual updates for the standalone page

toji:
Trying to make it feel less busy, help the critical elements stand out, and make it a bit easier to read.
 
 - Gave the entire page some padding
 - Removed some left-hand borders to reduce visual noise
 - Increased indentation of each level
 - Adding some padding for individual entries
 - Bolded the path for each entry
 - Allowed the entry paths to be wider, so not as much is cut off on narrow displays.
 - Added icons for pass/warn/fail states to aid colorblind users
 - Put WebGPU logo on there, because why not?

Updated style vs Current style:
Screenshot 2021-07-15 151526


standalone/index.html
```diff
@@ -209,7 +246,7 @@
     </style>
   </head>
   <body>
-    <h1>WebGPU Conformance Test Suite</h1>
+    <h1><img class="logo" src="webgpu-logo-notext.svg"></img>WebGPU Conformance Test Suite</h1>
```
kainino0x, line 249:
```suggestion
    <h1><img class="logo" src="webgpu-logo-notext.svg">WebGPU Conformance Test Suite</h1>
```
Closing tag only needed in XHTML, but we're HTML5 :)

kainino0x (approved):
Thanks, this is better!

PR #638, opened 2021-07-16
Disallow zero-sized buffer bind group bindings

### LINKED ISSUE #234: Update validation tests to disallow zero-sized bindings
[`createBindGroup:buffer_offset_and_size_for_bind_groups_match`](https://github.com/gpuweb/cts/blob/40e337a38784ad72fa5c7b9afd1b9c358a9e0f1a/src/webgpu/api/validation/createBindGroup.spec.ts#L251) tests currently check that zero-sized bindings **are valid**. This needs to be fixed to reflect the decision in https://github.com/gpuweb/gpuweb/issues/686 to **disallow zero-sized bindings**. The reason is that we need to require at least one element in a buffer to perform clamping.

austinEng:
Bug: crbug.com/dawn/985

Fixes #234 

<hr>

### DIFF
--- a/src/webgpu/api/validation/createBindGroup.spec.ts
+++ b/src/webgpu/api/validation/createBindGroup.spec.ts
@@ -296,8 +296,7 @@ g.test('buffer_offset_and_size_for_bind_groups_match')
   .desc(
     `
     Test that a buffer binding's [offset, offset + size) must be contained in the BindGroup entry's buffer.
-    - Test for various offsets and sizes
-    - TODO(#234): disallow zero-sized bindings`
+    - Test for various offsets and sizes`
   )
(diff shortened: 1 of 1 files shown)


austinEng:
Context: https://github.com/gpuweb/gpuweb/issues/686#issuecomment-657790808

kainino0x (approved):
What's the minimum size we can bind? We should test that too.

I know it was discussed somewhere at some point, but re: the context link, space for one element mean it depends on what's in the shader. I think we might have said that on platforms without robustness we could clamp memory accesses after resolving struct/array offsets etc, but that still means we need space for one primitive element, and I'm not sure if that's 4 bytes, or 16 bytes (vec4).

PR #639, opened 2021-07-16
Fix robust_access_vertex.spec.ts

austinEng:
- Was using incorrect (old) enums for GPUVertexFormat
- queue.writeBuffer was taking an offset in bytes, not in elements

Bug: crbug.com/dawn/983

<hr>

### DIFF
--- a/src/webgpu/shader/execution/robust_access_vertex.spec.ts
+++ b/src/webgpu/shader/execution/robust_access_vertex.spec.ts
@@ -167,7 +167,12 @@ class DrawCall {
     if (partialLastNumber) {
       size -= 3;
     }
-    this.device.queue.writeBuffer(vertexBuffer, 0, vertexArray, size);
+    this.device.queue.writeBuffer(
+      vertexBuffer,
+      0,
+      vertexArray,
+      size / vertexArray.BYTES_PER_ELEMENT
+    );
     return vertexBuffer;
   }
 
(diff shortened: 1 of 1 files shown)


Kangz (approved):

jzm-intel:
For `queue.writeBuffer(buffer, bufferOffset, data, dataOffset, size)`, now we have `dataOffset` as `size / vertexArray.BYTES_PER_ELEMENT`, which actually skip all data element in the provided vertexArray and leave the whole buffer 0. Meanwhile, the validation function in the vertex shader just check if the data read back from vertex buffer is *one of the validValues or **just 0***(for OOB case), so a vertex buffer full of 0 pass the validation. In fact I tried totally removing the writeBuffer here, and find no different in testing result. Actually I think it should be 
```
this.device.queue.writeBuffer(
      vertexBuffer,
      0,
      vertexArray,
      0, // dataOffset
      size / vertexArray.BYTES_PER_ELEMENT // size, in elements
    );
```
I am now working on this testsuit to deal with the validation error generated in Draw and DrawIndexed, and maybe I can fix it in that PR.

Kangz:
Makes sense. Can you make a PR that does that change and also see if it fails horribly on Chromium or it still works?

jzm-intel:
I am working on webgpu:shader,execution,robust_access_vertex:vertexAccess:* now, and changing this writeBuffer doesn't change the result at all, as the validation function accept 0.
Currently the test is failed due to buffer OOB validation errors for draw and drawIndexed, and disallowing of drawIndexedIndirect.

src/webgpu/shader/execution/robust_access_vertex.spec.ts
```diff
@@ -167,7 +167,12 @@ class DrawCall {
     if (partialLastNumber) {
       size -= 3;
     }
-    this.device.queue.writeBuffer(vertexBuffer, 0, vertexArray, size);
+    this.device.queue.writeBuffer(
+      vertexBuffer,
+      0,
+      vertexArray,
+      size / vertexArray.BYTES_PER_ELEMENT
```
kainino0x, line 174: Rounding here also means that the `size -= 3` probably doesn't do as intended.

kainino0x (commented):
I'll open a PR for this.

kainino0x:
Even with the fix I am seeing no difference in the test result. The tests aren't passing on my machine (mac, intel, canary, use_tint_generator), though. Are they expected to?

jzm-intel:
> Even with the fix I am seeing no difference in the test result. The tests aren't passing on my machine (mac, intel, canary, use_tint_generator), though. Are they expected to?

Yes, it is due to the OOB validation added in dawn:808 for Draw and DrawIndexed, and the disallowing for DrawIndexedIndirect. I am working on fixing this test.

PR #643, opened 2021-07-16
Visually indicate test results of entire subtrees

toji:
Fixes a longstanding annoyance of mine: Clicking "run" on a subtree without expanding the test cases gave no visual indication that tests were being run.

This approach isn't perfect but it at least gives the user some indication of progress being made. Each subtree underneath the one you initially run from will now color itself based on the results of all test cases underneath it, as well as displaying an icon for accessibility. If all tests under the subtree succeed it will turn green, if all of them fail it will turn red, and if it's a mixture of pass and fail you'll get a red/green stripe.

Areas for improvement:
  - Running a subtree with not propagate results up to levels above the one you ran from.
  - Currently collects cumulative time it took to run a subtree but is not displaying it. (Tried, but it felt very noisy.)
  - Could show numeric results (ie: "15/20 passed" or "75% passed")
  - Would be nice to visually indicate which tests are currently being run.

Preview:
Screenshot 2021-07-16 140049

### DIFF
--- a/src/common/runtime/standalone.ts
+++ b/src/common/runtime/standalone.ts
@@ -94,10 +94,13 @@ function makeCaseHTML(t: TestTreeLeaf): VisualizedSubtree {
   let caseResult: LiveTestCaseResult | undefined;
 
   // Becomes set once the DOM for this case exists.
+  let clearRenderedResult: (() => void) | undefined;
   let updateRenderedResult: (() => void) | undefined;
 
   const name = t.query.toString();
   const runSubtree = async () => {
+    if (clearRenderedResult) clearRenderedResult();
+
     const result : SubtreeResult = emptySubtreeResult();
 
     haveSomeResults = true;
@@ -136,6 +139,12 @@ function makeCaseHTML(t: TestTreeLeaf): VisualizedSubtree {
     div.appendChild(casehead);
     div.appendChild(caselogs[0]);
 
+    clearRenderedResult = () => {
+      div.removeAttribute('data-status');
+      casetime.text('ms');
+      caselogs.empty();
+    };
+
     updateRenderedResult = () => {
       if (caseResult) {
         div.setAttribute('data-status', caseResult.status);
@@ -171,6 +180,7 @@ function makeCaseHTML(t: TestTreeLeaf): VisualizedSubtree {
 function makeSubtreeHTML(n: TestSubtree, parentLevel: TestQueryLevel): VisualizedSubtree {
   let subtreeResult : SubtreeResult = emptySubtreeResult();
   // Becomes set once the DOM for this case exists.
+  let clearRenderedResult: (() => void) | undefined;
   let updateRenderedResult: (() => void) | undefined;
 
   const { runSubtree, generateSubtreeHTML } = makeSubtreeChildrenHTML(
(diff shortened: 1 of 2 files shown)


austinEng:
hmm, what if we put colored boxes next to the subtree name? I think the stripes make it a bit hard to read

toji:
Aww, but the CSS for it is fun! 😆

Fair enough, though I do want it to be fairly obvious that a level has been completed. We could choose a third (solid) color to indicate mixed results? Yellow, maybe?

austinEng:
third color, or you could do like a split red+green box? I think yellow works for in-progress though, or maybe one of those spinning loading symbols (idk if we want the page animated though).

toji:
What about a light purple for the mixed results?

Screenshot 2021-07-16 143825

austinEng:
Works for me, though purple is the color of Chrome infra failure, so maybe a very minor preference for something else

toji:
I guess the other way to look at it is a partial success is still an overall failure from a conformance point of view, so we'd just color mixed results as red.

kainino0x:
Haven't had a chance to review this yet, but I think we should just make it red. We have more than just two status results - we also have "warn" and "skip" (though I don't think "warn" has a color), and I think what we'll want to do is have each subtree show the "worst" status of any of its leaves.

Also just a thought: Maybe for the subtrees we could highlight just the left margin, or something? Not necessary, but would provide one more visual distinction between leaves and non-leaves in the tree.

toji:
Okay, so a mix of pass/fail will simply show up as red. Also, to give show the difference between subtrees and tests, I've got two different variants that we could use:

1) Left margin only:
Screenshot 2021-07-17 143024

2) Gradient:
Screenshot 2021-07-17 135853

The latest commit implements the gradient, which I think is a bit easier to parse at-a-glance, but I can also appreciate that it may be a bit more overwhelming to some, so I'd be fine with either approach.

src/common/runtime/standalone.ts
```diff
@@ -83,7 +112,23 @@ function makeCaseHTML(t: TestTreeLeaf): VisualizedSubtree {
       await t.run(rec);
     }
 
+    result.total++;
+    result.timems = caseResult.timems;
```
kainino0x, line 116: Is this supposed to be `+=`?

src/common/runtime/standalone.ts
```diff
@@ -83,7 +112,23 @@ function makeCaseHTML(t: TestTreeLeaf): VisualizedSubtree {
...
+    switch (caseResult.status) {
+      case 'pass':
+        result.pass++;
+        break;
+      case 'fail':
+        result.fail++;
+        break;
+      case 'skip':
+        result.fail++;
```
kainino0x, line 125: "skip" is a passing status. (we use it for cases where a webgpu feature isn't available, or the test is running on a worker and `document` isn't available, etc.)

kainino0x (approved):
Personal preference for option 1 (margin only, no gradient) but LGTM.

Status clearing is nice too!

src/common/runtime/standalone.ts
toji: That was a typo, thanks for catching it! I'm recording the skip count now, but it's not factoring into the subtree's visuals in any way.

src/common/runtime/standalone.ts
toji: Technically it doesn't matter in this case because we're starting from zero and it's only measuring one test, but I made it `+=` for consistency.

toji:
Forgot to mention that I switched it to color the margins only prior to merging.

PR #644, opened 2021-07-16
WPT: Add an option to fail on warnings

austinEng:
<hr>

### DIFF
--- a/src/common/runtime/wpt.ts
+++ b/src/common/runtime/wpt.ts
@@ -63,7 +63,7 @@ setup({
       }
 
       // Unfortunately, it seems not possible to surface any logs for warn/skip.
-      if (res.status === 'fail') {
+      if (res.status === 'fail' || (res.status === 'warn' && optionEnabled('failOnWarnings'))) {
         const logs = (res.logs ?? []).map(prettyPrintLog);
         assert_unreached('\n' + logs.join('\n') + '\n');
       }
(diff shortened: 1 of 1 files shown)


austinEng:
Second commit changes it to be a global Promise<boolean>. We could pass it in the same way we do the expectations

src/common/runtime/wpt.ts
```diff
@@ -21,6 +21,7 @@ declare function done(): void;
 declare function assert_unreached(description: string): void;
 
 declare const loadWebGPUExpectations: Promise<unknown> | undefined;
+declare const shouldWebGPUCTSFailOnWarnings: Promise<boolean> | undefined;
```
kainino0x, line 24: Does it need to be a promise? I think loadWebGPUExpectations is only a promise because it normally involves an import() call.

kainino0x (approved):

src/common/runtime/wpt.ts
kainino0x: Like, we should be able to just set it in cts.html (and in fact maybe this PR should add it to `src/common/templates/cts.html` as a reference, just like we have for loadWebGPUExpectations).

austinEng: I think a promise would be most flexible because we could also `import` it. That allows the browser test runner to simply switch a flag instead of generating a different .html file.

Chromium would have a single `import`:
```javascript
const loadWebGPUConfig = import('some/path/to/config.js');
const loadWebGPUExpectations = loadWebGPUConfig.then(c => c.expectations);
const shouldWebGPUCTSFailOnWarnings = loadWebGPUConfig.then(c => c.shouldWebGPUCTSFailOnWarnings);
```

PR #646, opened 2021-07-17
Do validation checks inside createEncoder's finish()

kainino0x:
And add a finishAndSubmit() helper.

With the fixes, this should ~eliminate cases where we had multiple
(endPass, finish, submit) calls inside one validation error scope and
couldn't make sure the errors came from the right functions.
This also substantially simplifies the finish/submit in most tests.

Folds two other helpers into this one so that all of the `CommandBufferMaker`s are centralized.

Tested locally, and all of the changed tests pass (except for existing known issues).

<hr>

### DIFF
--- a/src/webgpu/api/validation/attachment_compatibility.spec.ts
+++ b/src/webgpu/api/validation/attachment_compatibility.spec.ts
@@ -5,7 +5,7 @@ TODO: Add sparse color attachment compatibility test when defined by specificati
 `;
 
 import { makeTestGroup } from '../../../common/framework/test_group.js';
-import { range, unreachable } from '../../../common/util/util.js';
+import { range } from '../../../common/util/util.js';
 import {
   kRegularTextureFormats,
   kSizedDepthStencilFormats,
--- a/src/webgpu/api/validation/attachment_compatibility.spec.ts
+++ b/src/webgpu/api/validation/attachment_compatibility.spec.ts
@@ -5,7 +5,7 @@ TODO: Add sparse color attachment compatibility test when defined by specificati
 `;
 
 import { makeTestGroup } from '../../../common/framework/test_group.js';
-import { range } from '../../../common/util/util.js';
+import { range, unreachable } from '../../../common/util/util.js';
 import {
   kRegularTextureFormats,
   kSizedDepthStencilFormats,
@@ -15,7 +15,7 @@ import {
   kTextureFormatInfo,
 } from '../../capability_info.js';
 
-import { ValidationTest, CommandBufferMaker } from './validation_test.js';
+import { ValidationTest } from './validation_test.js';
 
 const kColorAttachmentCounts = range(kMaxColorAttachments, i => i + 1);
 const kDepthStencilAttachmentFormats = [
@@ -28,7 +28,8 @@ class F extends ValidationTest {
   createAttachmentTextureView(format: GPUTextureFormat, sampleCount?: number) {
     return this.device
       .createTexture({
-        size: [1, 1, 1],
+        // Size matching the "arbitrary" size used by ValidationTest helpers.
+        size: [16, 16, 1],
         format,
         usage: GPUTextureUsage.RENDER_ATTACHMENT,
         sampleCount,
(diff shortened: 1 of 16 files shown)


kainino0x:
@egalli Would you be able to take a look at this change? (Feel free to defer and I can find someone else.) Thanks!

src/webgpu/api/validation/attachment_compatibility.spec.ts
```diff
@@ -350,44 +289,50 @@ Test that the sample count in render passes or bundles match the pipeline sample
...
-      finish: finishWithDepthStencilOnly,
-    } = t.createPassOrBundleEncoder(encoderType, [], 'depth24plus-stencil8', encoderSampleCount);
-    encoderWithDepthStencilOnly.setPipeline(pipelineWithDepthStencilOnly);
-
-    t.expectValidationError(() => {
-      t.queue.submit([finishWithDepthStencilOnly()]);
-    }, encoderSampleCount !== pipelineSampleCount);
+    const { encoderType, attachmentType, encoderSampleCount, pipelineSampleCount } = t.params;
+
+    switch (attachmentType) {
```
egalli, line 300: nit(since the original code was just chaining the expects): can't we replace the switch with something like
```ts
  let colorFormats = attachmentType === "color" ? ["rgba8unorm"] : [];
  let depthStencilFormat =
    attachmentType === "depthstencil" ? "depth24plus-stencil8" : undefined;

  const pipeline = t.createRenderPipeline(
    colorFormats.map((format) => ({ format })),
    depthStencilFormat === undefined ? { format: depthStencilFormat } : undefined,
    pipelineSampleCount
  );

  const { encoder, validateFinishAndSubmit } = t.createEncoder(encoderType, {
    attachmentInfo: {
      colorFormats,
      depthStencilFormat,
      sampleCount: encoderSampleCount,
    },
  });

  encoder.setPipeline(pipeline);
  validateFinishAndSubmit(encoderSampleCount === pipelineSampleCount, true);
```

src/webgpu/api/validation/util/command_buffer_maker.ts
```diff
@@ -0,0 +1,76 @@
...
+   * - Only `submit()` should fail if the resource is 'destroyed'.
+   */
+  readonly validateFinishAndSubmitGivenState: (resourceState: ResourceState) => void;
+
+  constructor(
+    t: ValidationTest,
+    encoder: EncoderByEncoderType<EncoderType>,
+    finish: (shouldSucceed: boolean) => GPUCommandBuffer
+  ) {
+    this.t = t;
```
egalli, line 54: It doesn't seem that we need to store the test.

src/webgpu/api/validation/validation_test.ts
```diff
@@ -326,71 +301,90 @@ export class ValidationTest extends GPUTest {
...
       }
       case 'render bundle': {
         const device = this.device;
-        const encoder = device.createRenderBundleEncoder({
-          colorFormats: [colorFormat],
+        const rbEncoder = device.createRenderBundleEncoder(fullAttachmentInfo);
+        const pass = this.createEncoder('render pass', { attachmentInfo });
+
+        return new CommandBufferMaker(this, rbEncoder, (shouldSucceed: boolean) => {
+          // Bundle should be invalid.
```
egalli, line 336: nit: comment makes it seem like render bundles will always be invalid.

egalli (approved):
LGTM!

This change will make it much easier to check that the errors are getting thrown at the correct time.

src/webgpu/api/validation/attachment_compatibility.spec.ts
```diff
@@ -28,7 +28,8 @@ class F extends ValidationTest {
   createAttachmentTextureView(format: GPUTextureFormat, sampleCount?: number) {
     return this.device
       .createTexture({
-        size: [1, 1, 1],
+        // Size matching the "arbitrary" size used by ValidationTest helpers.
```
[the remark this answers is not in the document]
kainino0x: I think it's OK as we could easily introduce other tests that "accidentally" depend on the arbitrary size, so we can never change it anyway (without careful testing). This is just one more such place.

src/webgpu/api/validation/attachment_compatibility.spec.ts
kainino0x: Good call!

PR #648, opened 2021-07-17
Depth/stencil formats should be invalid for 3D textures

Richard-Yunchao:
<hr>

### DIFF
--- a/src/webgpu/api/validation/createTexture.spec.ts
+++ b/src/webgpu/api/validation/createTexture.spec.ts
@@ -9,6 +9,7 @@ import {
   kTextureDimensions,
   kTextureUsages,
   kUncompressedTextureFormats,
+  kRegularTextureFormats,
   textureDimensionAndFormatCompatible,
 } from '../../capability_info.js';
 import { DefaultLimits, GPUConst } from '../../constants.js';
@@ -531,11 +532,11 @@ g.test('texture_size,2d_texture,compressed_format')
 
 g.test('texture_size,3d_texture,uncompressed_format')
   .desc(
-    `Test that depth/stencil formats are invalid for 3D texture and regular formats have particualr textures size limits for 3D texture.`
+    `Test texture size requirement for 3D texture with uncompressed format. Note that depth/stencil formats are invalid for 3D textures, so we only test regular formats.`
   )
   .paramsSubcasesOnly(u =>
     u //
-      .combine('format', kUncompressedTextureFormats)
+      .combine('format', kRegularTextureFormats)
       .combine('size', [
         // Test the bound of width
         [DefaultLimits.maxTextureDimension3D - 1, 1, 1],
@@ -553,8 +554,7 @@ g.test('texture_size,3d_texture,uncompressed_format')
   )
   .fn(async t => {
     const { format, size } = t.params;
-    const info = kTextureFormatInfo[format];
-    await t.selectDeviceOrSkipTestCase(info.feature);
+    await t.selectDeviceOrSkipTestCase(kTextureFormatInfo[format].feature);
 
     const descriptor: GPUTextureDescriptor = {
       size,
@@ -566,9 +566,7 @@ g.test('texture_size,3d_texture,uncompressed_format')
     const success =
       size[0] <= DefaultLimits.maxTextureDimension3D &&
       size[1] <= DefaultLimits.maxTextureDimension3D &&
-      size[2] <= DefaultLimits.maxTextureDimension3D &&
-      !info.depth &&
-      !info.stencil;
+      size[2] <= DefaultLimits.maxTextureDimension3D;
 
     t.expectValidationError(() => {
       t.device.createTexture(descriptor);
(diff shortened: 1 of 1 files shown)


Richard-Yunchao:
PTAL. This bug was found by https://chromium-review.googlesource.com/c/chromium/src/+/3025322. In that Chromium CL, I searched '3d' in WebGPUExpectations and removed all related failed/crashed cases and run the bots, this is the only bug now.

kainino0x (commented):
Changes LGTM, but this test name/description says it's testing texture size limits. Can you either update the name/description or see if there's a different test that it makes more sense to put these in?

Richard-Yunchao:
> Changes LGTM, but this test name/description says it's testing texture size limits. Can you either update the name/description or see if there's a different test that it makes more sense to put these in?

Done. I changed the description. The name should be OK because depth/stencil formats are uncompressed formats. another test named dimension_type_and_format_compability can cover all valid/invalid formats for 3D textures. But it is not a problem to have some redundant tests in CTS. In addition, these test names for 3D textures are consistent with 2D's (2d_texture,uncompressed_formats, 2d_texture,compressed_formats). It looks like nothing is missed from the test names for 2D and 3D. So I'd like to keep it. Alternatively, we can just remove depth/stencil formats and test kReguarTextureFormats only in this test.

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -530,7 +530,9 @@ g.test('texture_size,2d_texture,compressed_format')
   });
 
 g.test('texture_size,3d_texture,uncompressed_format')
-  .desc(`Test texture size requirement for 3D texture with uncompressed format.`)
+  .desc(
+    `Test that depth/stencil formats are invalid for 3D texture and regular formats have particualr textures size limits for 3D texture.`
```
kainino0x, line 534:
```suggestion
    `Test that depth/stencil formats are invalid for 3D texture and regular formats have particular textures size limits for 3D texture.`
```

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -530,7 +530,9 @@ g.test('texture_size,2d_texture,compressed_format')
   });
 
 g.test('texture_size,3d_texture,uncompressed_format')
```
kainino0x, line 532: The test name still also needs to be updated. It says "texture_size" but it is testing more than just that.

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: Ack.

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: Good point. The name is not accurate if so. Considering tests nearby are all testing texture_size, it may think it makes more sense to keep the name but remove depth/stencil formats for 3D textures because they have already been tested in another test. 

I keep the name as uncompressed_formats (not regular_format) to show we don't miss any formats (uncompressed formats + compressed formats below = all formats), and add some statement to explain that we are testing regular formats because depth/stencil formats are invalid for 3D textures.

kainino0x: > because they have already been tested in another test

Where are these tests? I thought the reason you were making this PR was because we didn't have any yet.

Richard-Yunchao: As I mentioned above, another test in this test file named dimension_type_and_format_compability can cover all valid/invalid formats for 3D textures (and other texture dimension types).

kainino0x: Ok, these are new tests then? If so, we should either either add them, or add a TODO/.unimplemented(), and should do so in this PR so the work is tracked now.

Richard-Yunchao: No new tests. dimension_type_and_format_compatibility() is already there in this file, which have already covered formats and dimension types (1d, 2d, and 3d) compatibility. As a result, depth/stencil formats for 3D textures (which are invalid) in this test are totally redundant, so I think we can remove them.

kainino0x: > dimension_type_and_format_compatibility() is already there in this file

Ohh, sorry. I copied and pasted "dimension_type_and_format_compability" into the search box but didn't notice it has a very subtle typo in it!

Yes, I understand now, all looks good. Sorry for the delay!

kainino0x (approved):

src/webgpu/api/validation/createTexture.spec.ts
Richard-Yunchao: Sorry for the typo... Thanks for your review!

Richard-Yunchao: Can you roll CTS into Chromium? A Chromium patch is pending on this PR. Alternatively, I can roll WebGL CTS into Chromium.

kainino0x: It rolls automatically:
https://autoroll.skia.org/r/webgpu-cts-chromium-autoroll
however the latest roll is failing, apparently due to the commit right before yours...

Richard-Yunchao: Got it. No worries. I can keep an eye on when this PR is rolled.

PR #651, opened 2021-07-20
Fix Index Buffer OOB validation for DrawIndexed

jzm-intel:
Change the test to expect the validation errors for index buffer OOB in DrawIndexed. Tests for DrawIndexedIndirect (indirect=true) is also included but should be skipped in WebGPUExpectation (I will upload that CL once this PR is landed), as DrawIndexedIndirect doesn't validate the index buffer range yet and thus disallowed in Dawn.

<hr>

### DIFF
--- a/src/webgpu/api/validation/encoding/cmds/index_access.spec.ts
+++ b/src/webgpu/api/validation/encoding/cmds/index_access.spec.ts
@@ -1,13 +1,10 @@
 export const description = `
 indexed draws validation tests.
 
-TODO: review and make sure these notes are covered:
-> - indexed draws:
->     - index access out of bounds (make sure this doesn't overlap with robust access)
->         - bound index buffer **range** is {exact size, just under exact size} needed for draws with:
->             - indexCount largeish
->             - firstIndex {=, >} 0
->     - x= {drawIndexed, drawIndexedIndirect}
+TODO: Implement robustness test or operation test for drawIndexedIndirect. This validation test
+only test that drawIndexed can catch index buffer OOB and generate a validation error, but for
+drawIndexedIndirect no CPU validation is available, and the robustness access in that case should
+be tested.
 `;
 
 import { makeTestGroup } from '../../../../../common/framework/test_group.js';
@@ -100,38 +97,6 @@ class F extends ValidationTest {
       });
     }
   }
-
-  drawIndexedIndirect(
-    indexBuffer: GPUBuffer,
-    bufferArray: Uint32Array,
-    indirectOffset: number,
-    isSuccess: boolean
-  ) {
-    const indirectBuffer = this.device.createBuffer({
-      mappedAtCreation: true,
-      size: bufferArray.byteLength,
-      usage: GPUBufferUsage.INDIRECT,
-    });
-    new Uint32Array(indirectBuffer.getMappedRange()).set(bufferArray);
-    indirectBuffer.unmap();
-
-    const pipeline = this.createRenderPipeline();
-
-    const encoder = this.device.createCommandEncoder();
-    const pass = this.beginRenderPass(encoder);
-    pass.setPipeline(pipeline);
-    pass.setIndexBuffer(indexBuffer, 'uint32');
-    pass.drawIndexedIndirect(indirectBuffer, indirectOffset);
-    pass.endPass();
-
-    if (isSuccess) {
-      this.device.queue.submit([encoder.finish()]);
-    } else {
-      this.expectValidationError(() => {
-        encoder.finish();
-      });
-    }
-  }
 }
 
 export const g = makeTestGroup(F);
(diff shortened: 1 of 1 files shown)


jzm-intel:
PTAL, thanks!

src/webgpu/api/validation/encoding/cmds/index_access.spec.ts
```diff
@@ -115,17 +124,22 @@ class F extends ValidationTest {
     pass.drawIndexedIndirect(indirectBuffer, indirectOffset);
     pass.endPass();
 
-    this.device.queue.submit([encoder.finish()]);
+    if (isSuccess) {
```
Kangz, line 127: The WebGPU specification doesn't require an error to be produced for `drawIndexedIndirect` because we don't know on the CPU how many indices will be used. I think we need to keep the TODO to check robust behavior for the indirect case. (note that for OT Chromium disables `drawIndecedIndirect` because it is missing the validation).

src/webgpu/api/validation/encoding/cmds/index_access.spec.ts
```diff
@@ -140,40 +154,43 @@ g.test('out_of_bounds')
...
-          { indexCount: 6, firstIndex: 1 }, // indexCount + firstIndex out of bound
-          { indexCount: 0, firstIndex: 6 }, // indexCount is 0 but firstIndex out of bound
-          { indexCount: 6, firstIndex: 6 }, // only firstIndex out of bound
-          { indexCount: 6, firstIndex: 10000 }, // firstIndex much larger than the bound
-          { indexCount: 7, firstIndex: 0 }, // only indexCount out of bound
-          { indexCount: 10000, firstIndex: 0 }, // indexCount much larger than the bound
-          { indexCount: 0xffffffff, firstIndex: 0xffffffff }, // max uint32 value
-          { indexCount: 0xffffffff, firstIndex: 2 }, // max uint32 indexCount and small firstIndex
-          { indexCount: 2, firstIndex: 0xffffffff }, // small indexCount and max uint32 firstIndex
+          { indexCount: 5, firstIndex: 1, isSuccess: true }, // draw the last 5 out of 6 index
```
Kangz, line 158: nit: I think we could compute the isSuccess directly inside the branch for `!indirect`.

src/webgpu/api/validation/encoding/cmds/index_access.spec.ts
```diff
@@ -185,25 +202,26 @@ g.test('out_of_bounds_zero_sized_index_buffer')
       u
         .combine('indirect', [false, true])
         .combineWithParams([
-          { indexCount: 3, firstIndex: 1 }, // indexCount + firstIndex out of bound
-          { indexCount: 0, firstIndex: 1 }, // indexCount is 0 but firstIndex out of bound
-          { indexCount: 3, firstIndex: 0 }, // only indexCount out of bound
-          { indexCount: 0, firstIndex: 0 }, // just zeros
+          { indexCount: 3, firstIndex: 1, isSuccess: false }, // indexCount + firstIndex out of bound
```
Kangz, line 205: ditto

Kangz (approved):
LGTM overall

shrekshao:
Based on https://bugs.chromium.org/p/dawn/issues/detail?id=871
We don't expect it to be validation error. We'd like to move to operation test (maybe `src/webgpu/api/operation/rendering`) and test the actual drawing results

jzm-intel:
> Based on https://bugs.chromium.org/p/dawn/issues/detail?id=871
> We don't expect it to be validation error. We'd like to move to operation test (maybe `src/webgpu/api/operation/rendering`) and test the actual drawing results

In [dawn:808](https://bugs.chromium.org/p/dawn/issues/detail?id=808) we change the behavior of vertex buffer and index buffer OOB in Draw and DrawIndexed from no-op to validation error. I am going to split this test into two part, e.g. a validation part for Draw and DrawIndexed, and a operation part for DrawIndirect and DrawIndexedIndirect. (Maybe keep a TODO and implement it later since DrawIndexedIndirect is disallowed before OT)
Thanks! @shrekshao

src/webgpu/api/validation/encoding/cmds/index_access.spec.ts
jzm-intel: Done.

src/webgpu/api/validation/encoding/cmds/index_access.spec.ts
jzm-intel: Done.

jzm-intel:
PTAL, thanks! @Kangz

Kangz:
LGTM, feel free to merge. Can you create a test file with a TODO for the drawIndexedIndirect aspect of robust index access?

shrekshao:
I see. This is for DrawIndexed. Thanks for the change!

kainino0x:
This is ready to land, so I'm skipping this unless you would like me to review it.

PR #663, opened 2021-07-22
New shader test: let duplication

munrocket:
I found that I can declare `let` duplicate in Firefox and all will be ok. It should throw an error.

### DIFF
--- a/src/webgpu/shader/validation/wgsl/duplicate-let-name-within-func.fail.wgsl
+++ b/src/webgpu/shader/validation/wgsl/duplicate-let-name-within-func.fail.wgsl
@@ -1,0 +1,11 @@
+// v-0014 - This fails because constant `a` is redeclared.
+
+[[stage(fragment)]]
+fn main() {
+  let a = 1u;
+  if (true) {
+    let a = 2u;
+  }
+  return;
+}
+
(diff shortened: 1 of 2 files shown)


src/webgpu/shader/validation/wgsl/duplicate-let-name-within-func.fail.wgsl
```diff
@@ -0,0 +1,11 @@
+// v-0014 - This fails because constant `a` is redeclared.
+
+[[stage(fragment)]]
+fn main() {
+  let a = 1u;
+  if (true) {
+    let a = 2u;
```
jrprice, line 7: This is perfectly valid in WGSL; inter-scope shadowing is supported.

The key spec language is this (emphasis mine):
> A declaration must not introduce a name when that identifier is already in scope **with the same end scope** as another instance of that name.

The end scope of the second `a` declaration is not the same as the first `a` declaration, and so this is valid.

munrocket: Thanks, seems that your clarification raised another problem with `var` tests. We can return to `let` little bit later.

PR #669, opened 2021-07-23
Fix var duplication wgsl test

munrocket:
After clarification https://github.com/gpuweb/cts/pull/663#discussion_r675259846 I found an error in current wgsl test.

It not pass in Chrome Canary with tint generator. But pass in Firefox Nightly (at least in MSL backend) because they don't checks duplications yet.

<hr>

**Author checklist for test code/plans:**

- [x] All outstanding work is tracked with "TODO" in a test/file description or `.unimplemented()` on a test.
- [x] New helpers, if any, are documented using `/** doc comments */` and can be found via `helper_index.txt`.
- [x] (Optional, sometimes not possible.) Tests pass (or partially pass without unexpected issues) in an implementation. (Add any extra details above.)

**[Reviewer sign-off](https://github.com/gpuweb/cts/blob/main/docs/reviews.md) for test code/plans:** (Note: feel free to pull in other reviewers at any time for any reason.)

- [ ] The test path is reasonable, the [description](https://github.com/gpuweb/cts/blob/main/docs/intro/plans.md) "describes the test, succinctly, but in enough detail that a reader can read only the test plans in a file or directory and evaluate the completeness of the test coverage."
- [ ] Tests appear to cover this area completely, except for outstanding TODOs. Validation tests use control cases.
    (This is critical for coverage. Assume anything without a TODO will be forgotten about forever.)
- [ ] Existing (or new) test helpers are used where they would reduce complexity.
- [ ] TypeScript code is readable and understandable (is unobtrusive, has reasonable type-safety/verbosity/dynamicity).

PR #670, opened 2021-07-23
Add 3 new var duplications wgsl tests

munrocket:
This is continues #669 according to [wgsl spec](https://www.w3.org/TR/WGSL/#declaration-and-scope).

This tests is passed in Chrome Canary but not in Firefox Nightly. I am tested it blindly by copy pasting, please be more careful.

<hr>

### DIFF
--- a/src/webgpu/shader/validation/wgsl/duplicate-var-in-one-scope-v2.fail.wgsl
+++ b/src/webgpu/shader/validation/wgsl/duplicate-var-in-one-scope-v2.fail.wgsl
@@ -1,0 +1,9 @@
+// v-0014 - This fails because variable `i` is redeclared
+
+[[stage(fragment)]]
+fn main() {
+  for (var i: i32 = 0; i < 10; i = i + 1) {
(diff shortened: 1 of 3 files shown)


src/webgpu/shader/validation/wgsl/duplicate-var-in-one-scope-v2.fail.wgsl
```diff
@@ -0,0 +1,9 @@
+// v-0014 - This fails because variable `i` is redeclared
```
jrprice, line 1: Nit: The WGSL spec no longer has validation error code, so please remove the "v-0014" from these tests.

jrprice (approved):
LGTM with one nit - thanks!

src/webgpu/shader/validation/wgsl/duplicate-var-in-one-scope-v2.fail.wgsl
munrocket: done

PR #671, opened 2021-07-23
More detailed plans for stress tests

krockot:
First pass at a more detailed set of plans for stress tests, as well as some basic placeholder tests. Tried to capture everything that was already in the top-level README.txt but in a more granular and/or concrete form, and added a bunch of other possible ideas.

<hr>


src/stress/device/buffer_allocation.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+g.test('concurrent_buffer_allocation')
+  .desc(`Tests allocation of many concurrent GPUBuffer objects.`)
+  .unimplemented();
+
+g.test('continuous_buffer_allocation_with_destroy')
+  .desc(
+    `Tests allocation and destruction of many GPUBuffer objects over time. Objects
+are sequentially created and destroyed over a very large number of iterations.`
+  )
+  .unimplemented();
```
kainino0x, line 19: Some of these stress tests shouldn't in theory put much load on the system, unless there are bugs. For example in this issue:
https://bugs.chromium.org/p/dawn/issues/detail?id=914
Sunny is looking at memory allocation tests that _could_ be classified as stress tests, but the hope was that while they'll stress the code, they won't stress the system. (TBD whether that works out, I guess.)

So on the one hand, maybe some of what you have in this PR could be conformance tests (src/webgpu/).
But on the other hand, we should probably try to keep the conformance tests responsive (under the WPT timeouts). So it occurs to me it might be best to categorize all long-running tests as "stress" tests, instead of the split I had previously been imagining.

src/stress/adapter/device_allocation.ts
```diff
@@ -0,0 +1,28 @@
+export const description = `
+Stress tests for GPUAdapter.requestDevice.
+`;
+
+import { makeTestGroup } from '../../common/framework/test_group.js';
+import { GPUTest } from '../../webgpu/gpu_test.js';
+
+export const g = makeTestGroup(GPUTest);
+
+g.test('concurrent_device_allocation')
```
kainino0x, line 10: We might want to pick a word other than "concurrent"? because at some point we will also need to write tests which create objects from different threads (workers), both on separate devices (which can be implemented now, actually) and on shared devices (which we don't support yet).

I don't know what, though. I think it's "concurrent" is also OK.

src/stress/device/buffer_allocation.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+
+g.test('continuous_buffer_allocation_with_gc')
```
kainino0x, line 21: FWIW: Our garbage collectors don't really understand the memory cost of these objects (since they're in GPU memory, not CPU memory), so these tests might not end up behaving the way you might expect. However you can use the `attemptGarbageCollection` helper to try to incur GC.

src/stress/adapter/device_allocation.ts
```diff
@@ -0,0 +1,28 @@
+export const description = `
+Stress tests for GPUAdapter.requestDevice.
+`;
+
+import { makeTestGroup } from '../../common/framework/test_group.js';
+import { GPUTest } from '../../webgpu/gpu_test.js';
+
+export const g = makeTestGroup(GPUTest);
```
kainino0x, line 8: (For better or worse,) GPUTests always provide their own GPUDevice object. This is fine, but since it won't specifically be used these could also use the base `Fixture` fixture or something.

src/stress/device/buffer_allocation.spec.ts
```diff
@@ -0,0 +1,27 @@
+export const description = `
+Stress tests for allocation of GPUBuffer objects through GPUDevice.
+`;
+
+import { makeTestGroup } from '../../common/framework/test_group.js';
+import { GPUTest } from '../../webgpu/gpu_test.js';
+
+export const g = makeTestGroup(GPUTest);
+
+g.test('concurrent_buffer_allocation')
```
kainino0x, line 10: nit: Redundancy with the filename isn't necessary, because the filename is part of the path. So e.g. these could be:

- concurrent
- continuous_with_destroy
- continuous_with_gc

Or using a comma-separator (which makes it tree structured):

- concurrent
- continuous,with_destroy
- continuous,with_gc

Same comment in the other files with this pattern.

src/stress/memory/churn.spec.ts
```diff
@@ -0,0 +1,17 @@
...
+import { makeTestGroup } from '../../common/framework/test_group.js';
+import { GPUTest } from '../../webgpu/gpu_test.js';
+
+export const g = makeTestGroup(GPUTest);
+
+g.test('churn')
+  .desc(
+    `Allocates and populates a huge number of buffers and textures over time,
+retaining some while dropping or explicitly destroying others. When finished,
+verifies the expected contents of any remaining buffers and textures.`
```
kainino0x, line 15: Actually, reading this description, I think @sunnyps's test probably is closer to this one than to the buffer_allocation one, since it checks contents at the end. Except this one does both buffers and textures.

src/stress/memory/oom.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+Stress tests covering robustness when available VRAM is exhausted.
+`;
+
+import { makeTestGroup } from '../../common/framework/test_group.js';
+import { GPUTest } from '../../webgpu/gpu_test.js';
+
+export const g = makeTestGroup(GPUTest);
+
+// Helper to exhaust VRAM until there is less than 64 MB of capacity. Returns
+// an opaque closure which can be called to free the allocated resources later.
```
kainino0x, line 11: nit: Use `/** tsdoc-style comments */` so this will be picked up by editors (and documentation generators, if this helper were in a different file).

src/stress/memory/oom.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+const exhaustVramUntilUnder64MB = async (device: GPUDevice) => {
```
kainino0x, line 12: This helper could be useful in other stress tests (or if at some point we decide to split this file into multiple), so this could go in another `.ts` file. If you do that, you should also add the file to the helper_index.

Also, normally we use `function` for helper functions (`async function` in this case). No real reason, but a little more readable, and consistent.

src/stress/memory/oom.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+    `Tests getMappedRange on a mappedAtCreation GPUBuffer that failed allocation due
+to OOM. This should throw a RangeError, but below a certain threshold may just
+crash the page.`
+  )
+  .unimplemented();
+
+g.test('map_after_vram_oom')
+  .desc(
+    `Allocates tons of buffers and textures with varying mapping states (unmappable,
+mappable, mapAtCreation, matAtCreation-then-unmapped) until OOM; then attempts
```
kainino0x, line 51:
```suggestion
mappable, mapAtCreation, mapAtCreation-then-unmapped) until OOM; then attempts
```

src/stress/shaders/entry_points.spec.ts
```diff
@@ -0,0 +1,12 @@
...
+Stress tests covering behavior around shader entry points.
+`;
+
+import { makeTestGroup } from '../../common/framework/test_group.js';
+import { GPUTest } from '../../webgpu/gpu_test.js';
+
+export const g = makeTestGroup(GPUTest);
+
+g.test('many_entry_points')
+  .desc(`Tests compilation and usage of shaders with a huge number of entry points.`)
```
kainino0x, line 11: There could end up being a normative limit to the number of entry points, in which case this would become a validation test instead. Let's leave a note (`TODO:`) so whoever looks at this knows to check.
cc @alan-baker

src/stress/texture/large.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+import { makeTestGroup } from '../../common/framework/test_group.js';
+import { GPUTest } from '../../webgpu/gpu_test.js';
+
+export const g = makeTestGroup(GPUTest);
+
+g.test('loading')
+  .desc(
+    `Tests execution of shaders loading values from very large (at least 8192x8192)
+textures. The texture size is selected according to the limit supported by the
+GPUDevice.`
```
kainino0x, line 14: Could potentially be expanded to very large 2d array or 3d textures as well (though they might run out of memory).

The base limit for 2d textures is 8192x8192 with 2048 layers.
The base limit for 3d textures is 2048x2048x2048.

src/stress/device/buffer_allocation.spec.ts
kainino0x: Ruminating on this some more I do think @sunnyps's test should become a stress test, as it would be best if we could run it for longer periods.

See below, though, it might actually belong more in `stress/memory/churn.spec.ts` depending on what distinction we want between this test and that test. (Maybe we don't want any? We could just dedup them.)

kainino0x (approved):
Lots of great details here, thanks! I had a few small comments throughout.

src/stress/device/buffer_allocation.spec.ts
krockot: For the memory tests, I think de-duping probably makes sense. It sounds like @sunnyps's work will cover the churn cases that I was imagining, and probably more.

Regarding the general non-stressness of some tests, I wasn't really able to gauge how likely each plan would be to put a ton of load on the system, so my heuristic so far has been a pretty simplistic, "make number big."

From there my plan was to build the most stressful version of each test I can come up with, and if it turns out to not have a long running time in practice, move it to a regular conformance test.

And I guess finally, if there turn out to only be a handful of useful, very-long-running tests, maybe we just make them regular conformance tests with longer timeouts, or some mechanism to make them optional in CI, etc?

Does that sound reasonable?

src/stress/adapter/device_allocation.ts
krockot: Makes sense. Went with "coexisting" to avoid the ambiguous connotation of "concurrent"

src/stress/device/buffer_allocation.spec.ts
krockot: Actually incurring GC would not be a hard expectation of these tests, the only expectation would be that the test doesn't blow up due to lack of explicit disposal. My intent was to reflect as accurately as possible the UA's behavior when confronted with an application that indiscriminately creates and drops lots of objects.

src/stress/adapter/device_allocation.ts
krockot: Done

src/stress/device/buffer_allocation.spec.ts
krockot: Done

src/stress/memory/churn.spec.ts
krockot: ack

src/stress/memory/oom.spec.ts
krockot: Done

src/stress/memory/oom.spec.ts
krockot: Done

src/stress/shaders/entry_points.spec.ts
krockot: Done

src/stress/texture/large.spec.ts
krockot: Done, added sub cases for 2d, 2d_array, and 3d

PR #675, opened 2021-07-26
Add ImportExternalTexture Video Test

bjjones:
<hr>


bjjones:
I'm a little confused on what test goes where for this. This file existed already and seemed to fit - but I've still got a couple more tests that seem to fit better elsewhere like:
- Ensure that external texture destruction occurs via microtask
- Ensure that using a destroyed external texture results in error
- Ensure that using an invalid HTMLVideoElement results in error

I feel like these^ tests belong in the normal /validation/importExternalTexture.spec.ts and /operation/importExternalTexture.spec.ts files. Or do you think I should just stick everything in this file?

src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -3,13 +3,150 @@ Tests for external textures from HTMLVideoElement (and other video-type sources?
...
+    const video = document.createElement('video');
+    video.src = videoUrl;
+    if (!video.canPlayType) {
+      throw new Error();
+    }
+
+    video.loop = true;
+    video.muted = true;
+    video.play();
+    await video.play();
```
kainino0x, line 50: Is it necessary to call this twice? One `await video.play()` seems enough, but if it's not, leave a comment about it.

src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -3,13 +3,150 @@ Tests for external textures from HTMLVideoElement (and other video-type sources?
...
+    `
+Test that importing different video formats into an external texture and sampling from them works correctly.
+`
+  )
+  .params(u =>
+    u //
+      .combine('videoSource', kVideoSources)
+  )
+  .fn(async t => {
+    const videoUrl = '../out/resources/' + t.params.videoSource;
```
kainino0x, line 40: I don't think this URL is going to work when we're in WPT. Fortunately, for web_platform, tests we don't have to worry about other runtimes (cmdline (node) and deno runtimes), but we still have both the standalone and wpt runtimes. And the wpt runtime is the one we use on automated testing, so we'll want it to work.

We may need some kind of base URL (pointing to the root of the src/ or out/ or out-wpt/ directory or whatever), passed in from the runtime, so that we know how to grab these resources. Happy to chat more about this, or look into it (but may take me some time to get to it).

src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -3,13 +3,150 @@ Tests for external textures from HTMLVideoElement (and other video-type sources?
...
+    const video = document.createElement('video');
+    video.src = videoUrl;
+    if (!video.canPlayType) {
+      throw new Error();
+    }
+
+    video.loop = true;
+    video.muted = true;
+    video.play();
```
kainino0x, line 49: The state of the art in getting videos to play for tests can usually be found in `startPlayingAndWaitForVideo`:
https://github.com/KhronosGroup/WebGL/blob/main/sdk/tests/js/webgl-test-utils.js
So it might be worth adding a link here to that, in case we ever need to reference it again.

I'm not sure whether we should port in everything that's there right now. It may be the safest bet.

src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -3,13 +3,150 @@ Tests for external textures from HTMLVideoElement (and other video-type sources?
...
+    passEncoder.endPass();
+    t.device.queue.submit([commandEncoder.finish()]);
+
+    // Top left corner should be red.
+    t.expectSinglePixelIn2DTexture(
+      colorAttachment,
+      kFormat,
+      { x: 2, y: 2 },
+      {
+        exp: new Uint8Array([0xff, 0x00, 0x00, 0xff]),
```
kainino0x, line 139: I hope these are the right colors once color management is taken into account! I don't actually have any clue...

src/webgpu/web_platform/external_texture/video.spec.ts
bjjones: This was a mistake. Only one is needed.

src/webgpu/web_platform/external_texture/video.spec.ts
bjjones: At least 3 of these tests pass (not sure about mp4...haven't tried with an mp4-enabled build). I'm actually sampling a few pixels away from the edge, which allows me to sample exactly red/green. Added a comment for this.

src/webgpu/web_platform/external_texture/video.spec.ts
bjjones: I think it makes sense to add in at least a portion of the helper - it seems like it's going to require running each test as a callback, so may as well start in that format to avoid churn later.

src/webgpu/web_platform/external_texture/video.spec.ts
bjjones: I see - so it looks like I can acquire `window.location.href` within common/runtime/wpt.ts and then pass it into `run()` - right? Alternatively is there any reason I can't just use `window.location.href` directly from the test?

src/webgpu/web_platform/external_texture/video.spec.ts
kainino0x: I'm thinking about color profiles/ICC profiles. I don't really know how they work with videos. I assume these test videos are "untagged" insofar as a video can be, so I think these expectations are correct.

(The comment is good though, thanks)

src/webgpu/web_platform/external_texture/video.spec.ts
kainino0x: Unfortunately window.location.href by itself won't let us figure out the location of the resources directory. The file paths look like this:

- In standalone: `/standalone/` -> `/out/resources/`
- In the actual WPT build: `/webgpu/cts.html` -> `/webgpu/resources/`
- In Chromium: `/wpt_internal/webgpu/cts.html` -> `/gen/third_party/webgpu-cts/src/`

In order to deal with this for the rest of the generated files, we point to the appropriate build directory from the template:
- https://github.com/gpuweb/cts/blob/main/src/common/templates/cts.html#L32
- https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/web_tests/webgpu/ctshtml-template.txt;l=24;drc=6b69efe99ef855e2ac953b37a31063d07ecfee9a

So I think we are going to have to add something like `loadWebGPUExpectations` to point to the root of the build directory.

src/webgpu/web_platform/external_texture/video.spec.ts
kainino0x: I think that makes sense. It could be a promise instead of a callback - but I suppose when requestVideoFrameCallback is supported, the callback is probably more likely to give us exactly the right frame (rather than a later one).

bjjones: I've been trying to figure out how use the their function directly - but I'm not very experienced with JS and having issues with async execution. I've updated the PR with startPlayingAndWaitForVideo, however it doesn't function correctly. I find that finalize() is being called before the test callback occurs, which is obviously wrong. Maybe take a look and see if you have a suggestion? If I had a way for startPlayingAndWaitForVideo to wait until maybeCallCallback has finished, we could use this.

src/webgpu/web_platform/external_texture/video.spec.ts
bjjones: Resolved.

src/webgpu/web_platform/external_texture/video.spec.ts
bjjones: Resolved.

bjjones:
Comments resolved. Thanks for the help.

shaoboyan:
@bjjones would it possible to cover the usage of external texture in compute shader too ?

bjjones:
> @bjjones would it possible to cover the usage of external texture in compute shader too ?

Added.

bjjones:
@kainino0x I added a check for microtask destruction in the sampling test. Lmk if you'd rather see it split into it's own test (it may also be more appropriate as a validation test?)

src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -3,13 +3,241 @@ Tests for external textures from HTMLVideoElement (and other video-type sources?
...
+  'red-green.theora.ogv',
+];
 
 export const g = makeTestGroup(GPUTest);
+
+g.test('importExternalTexture,sample')
+  .desc(
+    `
+Test that importing different video formats into an external texture and sampling from them works correctly.
+Ensure GPUExternalTexture
```
kainino0x, line 38: Incomplete?

src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -3,13 +3,241 @@ Tests for external textures from HTMLVideoElement (and other video-type sources?
...
+        colorAttachment,
+        kFormat,
+        { x: 2, y: 2 },
+        {
+          exp: new Uint8Array([0xff, 0x00, 0x00, 0xff]),
+        }
+      );
+
+      // Bottom right corner should be green. Sample a few pixels away from the edges to avoid color
+      // management artifacts.
```
kainino0x, line 144: nit: I think if there are corner artifacts it would be due to compression, not color management
```suggestion
      // Bottom right corner should be green. Sample a few pixels away from the edges to avoid
      // compression artifacts.
```

src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -3,13 +3,241 @@ Tests for external textures from HTMLVideoElement (and other video-type sources?
...
+                  width : i32;
+              };
+    
+              [[group(0), binding(0)]] var t : texture_external;
+              [[group(0), binding(1)]] var<storage, read_write> dst : Data;
+    
+              [[stage(compute), workgroup_size(1)]] fn main() {
+                var dims : vec2<i32> = textureDimensions(t);
+                dst.width = dims.x;
+                dst.height = dims.y;
```
kainino0x, line 214: Ideally copy the data into a storage texture and check the contents? That way we can be sure the sampling is working in compute shaders when we implement the 0-copy impl (which might differ slightly for compute shaders since iirc texture sampling of the underlying textures won't be possible in compute).

If not implementing this now, please leave a TODO before landing the PR.

src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -3,13 +3,241 @@ Tests for external textures from HTMLVideoElement (and other video-type sources?
...
+      const encoder = t.device.createCommandEncoder();
+      const pass = encoder.beginComputePass();
+      pass.setPipeline(pipeline);
+      pass.setBindGroup(0, bg);
+      pass.dispatch(1);
+      pass.endPass();
+      t.device.queue.submit([encoder.finish()]);
+
+      // Check that video dimension are returned correctly as 80x128
+      const resultData = new Uint32Array([0x0000005000000080]);
```
kainino0x, line 240: This number is too big for uint32array, is this meant to be:
```suggestion
      const resultData = new Uint32Array([80, 128]);
```
?

kainino0x (approved):
> @kainino0x I added a check for microtask destruction in the sampling test. Lmk if you'd rather see it split into it's own test (it may also be more appropriate as a validation test?)

Would probably be ideal to have it be a separate test if it's not too hard. Then browsers can skip it if their implementation or the test turns out to be flaky. (The timing of this is potentially hard to test reliably.)
I'm neutral on validation vs operation test. It fits equally well into both, I think.

kainino0x:
LGTM, but I'd like to review the diff if you split the destruction test or enhance the compute test. Or feel free to land this with TODOs and do that in a new PR.

src/webgpu/web_platform/external_texture/video.spec.ts
bjjones: Done.

src/webgpu/web_platform/external_texture/video.spec.ts
bjjones: Added TODO

src/webgpu/web_platform/external_texture/video.spec.ts
bjjones: Done.

src/webgpu/web_platform/external_texture/video.spec.ts
bjjones: Removed.

bjjones:
Will merge after this passes. Will do a follow  up PR to address TODOs.

PR #676, opened 2021-07-28
Add type validation tests for fragment outputs and attachments

shrekshao:
https://crbug.com/dawn/811

<hr>

### DIFF
--- a/src/webgpu/api/operation/render_pipeline/pipeline_output_targets.spec.ts
+++ b/src/webgpu/api/operation/render_pipeline/pipeline_output_targets.spec.ts
@@ -35,12 +35,13 @@ class F extends GPUTest {
   }
 
   getFragmentShaderCode(
-    v: string[],
+    output: readonly number[],
     sampleType: GPUTextureSampleType,
     componentCount: number
   ): string {
     let fragColorType;
     let suffix;
+    let fractionDigits = 0;
     switch (sampleType) {
       case 'sint':
         fragColorType = 'i32';
@@ -52,9 +53,12 @@ class F extends GPUTest {
         break;
       default:
         fragColorType = 'f32';
-        suffix = '.0';
+        suffix = '';
+        fractionDigits = 2;
         break;
     }
+
+    const v = output.map(n => n.toFixed(fractionDigits));
 
     let outputType;
     let result;
@@ -109,7 +113,7 @@ g.test('color,component_count')
 
     // expected RGBA values
     // extra channels are discarded
-    const v = [0, 1, 0, 1];
+    const result = [0, 1, 0, 1];
 
     const renderTarget = t.device.createTexture({
       format,
@@ -136,11 +140,7 @@ g.test('color,component_count')
       },
       fragment: {
         module: t.device.createShaderModule({
-          code: t.getFragmentShaderCode(
-            v.map(n => n.toString()),
-            expectedType,
-            componentCount
-          ),
+          code: t.getFragmentShaderCode(result, expectedType, componentCount),
         }),
         entryPoint: 'main',
         targets: [{ format }],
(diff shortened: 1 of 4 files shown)


src/webgpu/api/validation/render_pass_descriptor.spec.ts
```diff
@@ -192,6 +199,57 @@ g.test('attachments_must_match_whether_they_are_used_for_color_or_depth_stencil'
...
+      fragment: {
+        module: t.device.createShaderModule({
+          code: `
+          [[stage(fragment)]] fn main() -> [[location(0)]] vec4<f32> {
+            return vec4<f32>(1.0, 0.0, 0.0, 1.0);
+          }
+          `,
+        }),
+        entryPoint: 'main',
+        targets: [{ format: 'rgba8unorm' }],
```
kainino0x, line 237: Since the format is hardcoded in the pipeline here this test is testing the compatibility between the pipeline and the render pass. (Is that the intent?)
I think this validation is covered in `api,validation,attachment_compatibility:render_pass_or_bundle_and_pipeline,color_format`

However there's another piece of validation, which is compatibility between the fragment shader itself (the `vec4<f32>` in this test) and the `.fragments.targets`. (This validation occurs inside createRenderPipeline.)
Unfortunately the spec is not very specific about this validation yet:
https://gpuweb.github.io/gpuweb/#abstract-opdef-validating-gpufragmentstate
> The pipeline output type must be compatible with colorState.format.

But I'm pretty sure the validation is:
- The scalar type (f32, i32, or u32) must match the [sample type of the format](https://gpuweb.github.io/gpuweb/#plain-color-formats).
- The size of the vec (e.g. `f32`, `vec2<f32>`, `vec3<f32>`, `vec4<f32>`) doesn't matter. Extra components are discarded and missing components are filled to `0,0,1`.

But we should verify that (by testing, probably).

I don't _think_ we have any tests for that yet. Possible places:
- `api,validation,createRenderPipeline` doesn't seem to have it though it _does_ generate shaders with matching primitive types
- `api,validation,layout_shader_compat` is only for compatibility between the shader and the pipeline layout, not other parts of the pipeline
- `api,validation,attachment_compatibility` is only for compatibility between pipeline, pass, and bundle.

so I'd put that in either createRenderPipeline or a new file.

kainino0x: > (by testing, probably)

Validation testing will help here (it will make sure we don't pass anything to the drivers that the debug layers don't like), but really we'll need operation testing to make sure the results come through properly.

Maybe you can try to see if we need to add an operation test plan for that.

shrekshao:
Add a validation test for fragment outputs with different type/componentCounts <-> different color target formats. Add a description for operation test and will come later (at a different pull request likely)

shrekshao:
Oops. Turns out on Mac Metal having fragment output like `vec2<f32>` to a `rgba8unorm` is not allowed
```
Error creating pipeline statefragment shader color output does not have enough components for the pixel format (MTLPixelFormatRGBA8Unorm)
    at Initialize (../../third_party/dawn/src/dawn_native/metal/RenderPipelineMTL.mm:415)
    at Create (../../third_party/dawn/src/dawn_native/metal/RenderPipelineMTL.mm:326)
    at GetOrCreateRenderPipeline (../../third_party/dawn/src/dawn_native/Device.cpp:627)
```

Also on win D3D12 it seems to expand the result with `0` (0 for alpha channel as well)

Kangz:
On the other hand is Metal happy to go `vec4<f32>` -> double channel or single channel formats? If that's the case we could always translate to be a `vec4<f32>` padded with 0s.

shrekshao:
* `vec4<f32>` -> `r8unorm`  Metal ✅ 
* `vec2<f32>` -> `rgba8unorm` Metal ❌

shrekshao:
In terms of that, the rules are interpreted as

> The pipeline output type must be compatible with colorState.format.

* The scalar type (f32, i32, or u32) must match the sample type of the format.
* The componentCount of the fragment output (e.g. f32, vec2<f32>, vec3<f32>, vec4<f32>) should have no fewer channels than that of the color attachment texture formats. Missing components are filled with 0.

kainino0x:
We can alternatively add a validation rule that matches Metal's. However I like this approach, as it seems very straightforward - as long as we can make things work on all backends.

(Note the transform can be done in createShaderModule to always output vec4 - fortunately it doesn't have to be done late, in createRenderPipeline. If it had to be done late, the group would prefer to avoid it.)

src/webgpu/api/validation/createRenderPipeline.spec.ts
```diff
@@ -187,3 +238,34 @@ g.test('sample_count_must_be_valid')
 
     t.doCreateRenderPipelineTest(isAsync, _success, descriptor);
   });
+
+g.test('pipeline_fragment_output_type_must_be_compatible_with_target_color_state_format')
```
kainino0x, line 242: This can have a shorter name and have the details in the description. The really long test names are an artifact from before we had test descriptions.

src/webgpu/api/validation/createRenderPipeline.spec.ts
```diff
@@ -187,3 +238,34 @@ g.test('sample_count_must_be_valid')
...
+    u
+      .combine('isAsync', [false, true])
+      .combine('format', kRenderableColorTextureFormats)
+      .beginSubcases()
+      .combine('sampleType', [
+        'float',
+        'uint',
+        'sint',
+        'unfilterable-float',
+      ] as GPUTextureSampleType[])
```
kainino0x, line 253:
```suggestion
      ] as const)
```
should work, and is typo-proof

src/webgpu/api/operation/render_pipeline/pipeline_output_targets.spec.ts
```diff
@@ -0,0 +1,12 @@
+export const description = `
+TODO:
+- Test pipeline outputs with different color targets, depth/stencil targets.
+  - different scalar types in shader (f32, u32, i32) to targets with different format (f32 to unorm/float, u32 to uint)
```
kainino0x, line 4: I think this should be a validation error?

src/webgpu/api/validation/createRenderPipeline.spec.ts
```diff
@@ -187,3 +238,34 @@ g.test('sample_count_must_be_valid')
...
+  .params(u =>
+    u
+      .combine('isAsync', [false, true])
+      .combine('format', kRenderableColorTextureFormats)
+      .beginSubcases()
+      .combine('sampleType', [
+        'float',
+        'uint',
+        'sint',
+        'unfilterable-float',
```
kainino0x, line 252: I don't think this currently does anything differently from the 'float' case?

kainino0x (commented):
I know this probably isn't quite done yet, but so far lgtm

shrekshao:
> (Note the transform can be done in createShaderModule to always output vec4 - fortunately it doesn't have to be done late, in createRenderPipeline. If it had to be done late, the group would prefer to avoid it.)

We don't know targets format at `createShaderModule`. It is passed in to `GPUFragmentState`. So seems we cannot do the transform early? Shall I open an issue at gpuweb/gpuweb ?

shrekshao:
Add operation tests to check pipeline output value is correct when having more components than color target formats (e.g. `vec4<f32>` -> `r8unorm` target)

Change validation test to allow extra pipeline components but disallow fewer pipeline compoenents (as stated in [comments above](https://github.com/gpuweb/cts/pull/676#issuecomment-889418845) )
Current chrome fails the validation test for mac (device lost, metal error) and win (unexpectedly pass, no validation). So further work in dawn and webgpu spec is needed.

kainino0x:
> We don't know targets format at `createShaderModule`. It is passed in to `GPUFragmentState`. So seems we cannot do the transform early? Shall I open an issue at gpuweb/gpuweb ?

If it's valid to output too many components (e.g. vec4 into rg8unorm or vec2 into r8unorm), then I think the suggestion was that createShaderModule could *always* "upgrade" fragment outputs to vec4 and fill any remaining components to 0. Then the shader doesn't have to care how many components are in the GPUFragmentState.

shrekshao:
Fair point

shrekshao:
As discussed at https://github.com/gpuweb/gpuweb/issues/2013

Add validation tests for extra rules related to blending factors;
Add operation tests making sure blending behaves correctly when fragment output or attachment format doesn't have alpha channel.

src/webgpu/api/operation/render_pipeline/pipeline_output_targets.spec.ts
```diff
@@ -1,12 +1,399 @@
...
+    u
+      .combine('format', kRenderableColorTextureFormats)
+      .beginSubcases()
+      .combine('componentCount', [1, 2, 3, 4])
+  )
+  .fn(async t => {
+    const { format, componentCount } = t.params;
+    const { expectedType, expectedComponentCount } = t.getExpectedTypeAndComponentCount(format);
+    if (componentCount < expectedComponentCount) {
+      t.skip('componentCount of pipeline output must not be fewer than that of target format');
```
kainino0x, line 109: For anything that can be "statically" filtered (without having a webgpu device), prefer to use `.unless` or `.filter` in the params instead.

src/webgpu/api/operation/render_pipeline/pipeline_output_targets.spec.ts
```diff
@@ -1,12 +1,399 @@
...
+    t.expectSingleColor(renderTarget, format, {
+      size: [1, 1, 1],
+      exp: { R: result[0], G: result[1], B: result[2], A: result[3] },
+    });
+  });
+
+g.test('color,component_count,blend')
+  .desc(
+    `Test that blending behaves correctly when:
+- fragement output has no alpha, but the src alpha is not used for the blend operation indicated by blend factors
```
kainino0x, line 175:
```suggestion
- fragment output has no alpha, and the src alpha is not used for the blend operation indicated by blend factors
```

src/webgpu/api/operation/render_pipeline/pipeline_output_targets.spec.ts
```diff
@@ -1,12 +1,399 @@
...
+- attachment format has no alpha, and the dst alpha should be assumed as 1
```
kainino0x, line 176: this will be interesting to test out, I'm still not 100% sure it's true :)

src/webgpu/api/operation/render_pipeline/pipeline_output_targets.spec.ts
```diff
@@ -1,12 +1,399 @@
...
+      .combine('format', ['r8unorm', 'rg8unorm', 'rgba8unorm', 'bgra8unorm'] as const)
+      .beginSubcases()
+      // result is expected values in the color attachment (extra channels are discarded)
+      // output is the fragment shader output vector
+      // 0.498 -> 0x7f, 0.502 -> 0x80
+      .combineWithParams([
+        // fragment output has no alpha
+        {
+          result: [0, 1, 0, 0],
+          output: [0],
```
kainino0x, line 192: I think `result` and `output` are just expected values? so they can be "private" params by spelling them `_result` and `_output`. (or maybe only `result` should be private, didn't read enough of this test yet)

src/webgpu/api/operation/render_pipeline/pipeline_output_targets.spec.ts
```diff
@@ -1,12 +1,399 @@
...
+    pass.endPass();
+    t.device.queue.submit([encoder.finish()]);
+
+    t.expectSingleColor(renderTarget, format, {
+      size: [1, 1, 1],
+      exp: { R: result[0], G: result[1], B: result[2], A: result[3] },
+    });
+  });
+
+g.test('color,component_count,blend')
```
kainino0x, line 172: (todo for self: still need to review this test)

src/webgpu/api/validation/createRenderPipeline.spec.ts
```diff
@@ -263,9 +285,120 @@ g.test('pipeline_fragment_output_type_must_be_compatible_with_target_color_state
       fragmentShaderCode: t.getFragmentShaderCode(sampleType, componentCount),
     });
 
-    const expectedType = t.getExpectedType(format);
+    const _success = expectedType === sampleType && componentCount >= expectedComponentCount;
+    t.doCreateRenderPipelineTest(isAsync, _success, descriptor);
+  });
+
+g.test('pipeline_output_targets,blend')
```
kainino0x, line 292: (todo for self: still need to review this test)

kainino0x (commented):
partially reviewed, lgtm so far

src/webgpu/api/operation/render_pipeline/pipeline_output_targets.spec.ts
```diff
@@ -0,0 +1,399 @@
...
+      output,
+      colorSrcFactor,
+      colorDstFactor,
+      alphaSrcFactor,
+      alphaDstFactor,
+    } = t.params;
+    const { expectedType, expectedComponentCount } = t.getExpectedTypeAndComponentCount(format);
+    const componentCount = output.length;
+    if (componentCount < expectedComponentCount) {
+      t.skip('componentCount of pipeline output must not be fewer than that of target format');
```
kainino0x, line 326: (same as above)

src/webgpu/api/operation/render_pipeline/pipeline_output_targets.spec.ts
```diff
@@ -0,0 +1,399 @@
...
+    u
+      .combine('format', ['r8unorm', 'rg8unorm', 'rgba8unorm', 'bgra8unorm'] as const)
+      .beginSubcases()
+      // result is expected values in the color attachment (extra channels are discarded)
+      // output is the fragment shader output vector
+      // 0.498 -> 0x7f, 0.502 -> 0x80
+      .combineWithParams([
+        // fragment output has no alpha
+        {
+          result: [0, 1, 0, 0],
```
kainino0x, line 191: Having a hard time understanding this expected result. If the src value is [0] and the dst value is [1,0,0,1] where does the green=1 come from?

src/webgpu/api/validation/createRenderPipeline.spec.ts
```diff
@@ -187,3 +257,148 @@ g.test('sample_count_must_be_valid')
...
+              operation: 'add',
+            },
+          },
+        },
+      ],
+      fragmentShaderCode: t.getFragmentShaderCode(sampleType, componentCount),
+    });
+
+    const blendingReadAlpha =
+      colorSrcFactor.includes('src-alpha') ||
```
kainino0x, line 394: TIL about string.includes! Have been using .indexOf()===-1 like in old JS...

src/webgpu/api/validation/createRenderPipeline.spec.ts
```diff
@@ -187,3 +257,148 @@ g.test('sample_count_must_be_valid')
...
+              dstFactor: alphaDstFactor,
+              operation: 'add',
+            },
+          },
+        },
+      ],
+      fragmentShaderCode: t.getFragmentShaderCode(sampleType, componentCount),
+    });
+
+    const blendingReadAlpha =
```
kainino0x, line 393: nit: blendingReadSrcAlpha

kainino0x (commented):
Finish reviewing and just a few more comments

src/webgpu/api/operation/render_pipeline/pipeline_output_targets.spec.ts
```diff
@@ -1,12 +1,399 @@
...
 `;
 
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { unreachable } from '../../../../common/util/util.js';
+import { kRenderableColorTextureFormats, kTextureFormatInfo } from '../../../capability_info.js';
 import { GPUTest } from '../../../gpu_test.js';
 
-export const g = makeTestGroup(GPUTest);
+class F extends GPUTest {
+  getExpectedTypeAndComponentCount(
```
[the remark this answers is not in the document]
shrekshao: Add `sampleType` to table with `GPUTextureSampleType` values (`float`, `sint`, `uint`, `depth`) to avoid further confusion. (Formatting with full words doesn't turn out to be a problem)

src/webgpu/api/operation/render_pipeline/pipeline_output_targets.spec.ts
shrekshao: ah... The extra components of the results are discarded so for a `r8` the g channel is unrelated. I removed extra channels to avoid confusions.

shrekshao:
Addressed the review comments (some with additional comments details). Thanks for the review!

src/webgpu/api/operation/render_pipeline/pipeline_output_targets.spec.ts
kainino0x: Ohh gotcha, thanks for the explanation

PR #679, opened 2021-07-29
Add tests for 3D texture copies

Richard-Yunchao:
This change add tests for 3D texture copies on existing tests. It
mainly adds dimension as a new parameter into each test in order
to validate copies for 3D and 2D array textures in addition to 2D
textures, with some minor changes in the test body if needed.

<hr>


Richard-Yunchao:
> This change add tests for 3D texture copies on existing tests. It
> mainly adds dimension as a new parameter into each test in order
> to validate copies for 3D and 2D array textures in addition to 2D
> textures, with some minor changes in the test body if needed.
> 


The existing tests are for 2D textures only. This change simply expands the existing tests 3D and 2D array textures in order to test 3D texture copies. It seems a little redundant. But I think it should be fine considering that 1) full coverage is more important for CTS, and 2) different dimension types (2D non-array, 2D array, 3D, and 1D) are the most important factor for a texture.  

I will add new tests specific for 3D texture copies later if needed.

Kangz:
There's a couple `.swp` files that were added ^^

I found that using `git add -u` works better than `git add` to avoid including unnecessary files.

src/webgpu/api/validation/image_copy/texture_related.spec.ts
```diff
@@ -89,7 +105,9 @@ g.test('usage')
   });
 
 g.test('sample_count')
-  .desc(`Multisampled textures cannot be copied.`)
+  .desc(
+    `Multisampled textures cannot be copied. Note that we don't test 2D array and 3D textures because multisample is not supported for 2D array and 3D texture creation`
```
Kangz, line 109: I assume this is changed in the other PR.

Kangz (approved):

src/webgpu/api/validation/image_copy/texture_related.spec.ts
Richard-Yunchao: Yeah. But that PR attempted to address another issue. In addition, as you pointed out, that PR might be unneeded because sample_count can be 4 only if it is T2T copy but this test is for writeTexture, B2T and T2B copy.

Richard-Yunchao:
> There's a couple `.swp` files that were added ^^
> 
> I found that using `git add -u` works better than `git add` to avoid including unnecessary files.

Oops. they are removed now. I noticed those .swp file via `git status` but I guess I forgot that later. Yeah, `git add -u` is a good option.

src/webgpu/api/validation/image_copy/texture_related.spec.ts
```diff
@@ -2,7 +2,11 @@ export const description = '';
```
kainino0x, line 2: IIUC most of these tests could be expanded to 1d as well. Can you put a TODO in the file description for that?

(I know it's already mentioned in `image_copy/README.txt`, but here we can be more specific.
Also, this partially resolves the TODO in that readme, but I don't know if there's still more 3d texture work to do so it can probably be left as is.)

src/webgpu/api/validation/image_copy/texture_related.spec.ts
```diff
@@ -310,13 +351,15 @@ g.test('copy_rectangle')
         break;
       }
       case 2: {
-        textureSize.depthOrArrayLayers = textureSizeValue;
+        textureSize.depthOrArrayLayers =
+          dimension === '2d' ? textureSizeValue : textureSizeValue << mipLevel;
```
kainino0x, line 355: nit: flip this so it'll still be correct when 1d is added:
```suggestion
          dimension === '3d' ? textureSizeValue << mipLevel : textureSizeValue;
```

kainino0x (approved):

src/webgpu/api/validation/image_copy/texture_related.spec.ts
Richard-Yunchao: Done. 

For the TODO in README.txt, I will remove the `3d` if I feel like that validation tests for 3D texture copies are complete.

src/webgpu/api/validation/image_copy/texture_related.spec.ts
Richard-Yunchao: Done.

Richard-Yunchao:
I merged this PR since it got 2 approvals and comments have already been addressed.

PR #689, opened 2021-08-02
Add refttests for different size of canvas and swapchain

shaoboyan:
This tests covers the cases that user config webgpu back
buffer with larger than/smaller than/equal to canvas size.

WebGPU canvas should present things by doing
downscaling/upscaling/no-changes for the back buffer.

Bug: https://crbug.com/dawn/936

<hr>

### DIFF
--- a/src/webgpu/web_platform/reftests/canvas_size_different_with_back_buffer_size.html
+++ b/src/webgpu/web_platform/reftests/canvas_size_different_with_back_buffer_size.html
@@ -14,7 +14,8 @@
   <canvas id="cvs_smaller_than_back_buffer" width="3" height="4"></canvas>
   <canvas id="cvs_change_size_after_configure" width="3" height="4"></canvas>
   <canvas id="cvs_change_size_and_reconfigure" width="3" height="4"></canvas>
-  <canvas id="cvs_back_buffer_css_different_size" width="6" height="8" style="width: 12px; height: 16px;"></canvas>
+  <canvas id="back_buffer_smaller_than_cvs_and_css" width="6" height="8" style="width: 12px; height: 16px;"></canvas>
+  <canvas id="cvs_smaller_than_back_buffer_and_css" width="3" height="4" style="width: 12px; height: 16px;"></canvas>
   <script src="/common/reftest-wait.js"></script>
   <script type="module">
     import { run } from './canvas_size_different_with_back_buffer_size.html.js';
--- a/src/webgpu/web_platform/reftests/canvas_size_different_with_back_buffer_size.html
+++ b/src/webgpu/web_platform/reftests/canvas_size_different_with_back_buffer_size.html
@@ -1,0 +1,20 @@
+<html class="reftest-wait">
+  <title>WebGPU canvas_back_buffer_different_size</title>
+  <meta charset="utf-8" />
+  <link rel="help" href="https://gpuweb.github.io/gpuweb/" />
+  <meta
+    name=fuzzy
+    content="128;48"
+  />
+  <link rel="match" 
+        href="./ref/canvas_size_different_with_back_buffer_size-ref.html" />
+
+  <canvas id="cvs_larger_than_back_buffer" width="6" height="8"></canvas>
+  <canvas id="cvs_same_as_back_buffer" width="3" height="4"></canvas>
+  <canvas id="cvs_smaller_than_back_buffer" width="3" height="4"></canvas>
+  <script src="/common/reftest-wait.js"></script>
+  <script type="module">
+    import { run } from './canvas_size_different_with_back_buffer_size.html.js';
+    run();
+  </script>
+</html>
--- a/src/webgpu/web_platform/reftests/canvas_size_different_with_back_buffer_size.html
+++ b/src/webgpu/web_platform/reftests/canvas_size_different_with_back_buffer_size.html
@@ -3,8 +3,8 @@
   <meta charset="utf-8" />
   <link rel="help" href="https://gpuweb.github.io/gpuweb/" />
   <meta
-    name=fuzzy
-    content="128;48"
+    name="assert"
+    content="WebGPU canvas should present correctly with different size of back buffer"
   />
   <link rel="match" 
         href="./ref/canvas_size_different_with_back_buffer_size-ref.html" />
@@ -12,6 +12,9 @@
   <canvas id="cvs_larger_than_back_buffer" width="6" height="8"></canvas>
   <canvas id="cvs_same_as_back_buffer" width="3" height="4"></canvas>
   <canvas id="cvs_smaller_than_back_buffer" width="3" height="4"></canvas>
+  <canvas id="cvs_change_size_after_configure" width="3" height="4"></canvas>
+  <canvas id="cvs_change_size_and_reconfigure" width="3" height="4"></canvas>
+  <canvas id="cvs_back_buffer_css_different_size" width="6" height="8" style="width: 12px; height: 16px;"></canvas>
   <script src="/common/reftest-wait.js"></script>
   <script type="module">
     import { run } from './canvas_size_different_with_back_buffer_size.html.js';
--- a/src/webgpu/web_platform/reftests/canvas_size_different_with_back_buffer_size.html.ts
+++ b/src/webgpu/web_platform/reftests/canvas_size_different_with_back_buffer_size.html.ts
@@ -203,7 +203,7 @@ export function run() {
         [back_buffer_width, back_buffer_height],
         [
           [red, red, red, red, green, green],
-          [red, red, red, red, green, green],
+          [red, red, green, red, green, green],
           [red, red, red, red, green, green],
           [red, red, red, red, green, green],
           [blue, blue, blue, blue, yellow, yellow],
--- a/src/webgpu/web_platform/reftests/canvas_size_different_with_back_buffer_size.html.ts
+++ b/src/webgpu/web_platform/reftests/canvas_size_different_with_back_buffer_size.html.ts
@@ -6,6 +6,9 @@ import { runRefTest } from './gpu_ref_test.js';
 declare const cvs_larger_than_back_buffer: HTMLCanvasElement;
 declare const cvs_same_as_back_buffer: HTMLCanvasElement;
 declare const cvs_smaller_than_back_buffer: HTMLCanvasElement;
+declare const cvs_change_size_after_configure: HTMLCanvasElement;
+declare const cvs_change_size_and_reconfigure: HTMLCanvasElement;
+declare const cvs_back_buffer_css_different_size: HTMLCanvasElement;
 
 export function run() {
   runRefTest(async t => {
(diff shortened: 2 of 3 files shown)


shaoboyan:
This PR is actually in WIP status but I'd like to create this for some comments.
I find that the ref I write with WebGL/Canvas2d has slightly pixel difference than the WebGPU output. 
The canvas shows(from left to right) are: back buffer has smaller size than canvas(case 1) + back buffer has the same size as canvas(case 2) + back buffer has larger size than canvas(case 3).
It seems WebGPU has more clear output than 2d Canvas or WebGL on case 3. And for case 1, WebGPU seems has pixel difference than any of the 2d or WebGL outputs.

Here are snapshots(Scale my browser to maximum):
- Use 2D canvas to generate ref: 
2dcanvas
- Use WebGL canvas with LINEAR filter config:
ref-linear
- Use WebGL canvas with NEAREST config:
ref-NEAREST
- WebGPU output:
webgpu

@kainino0x  I'm not quite sure whether this is the same issue as [this chromium issue](https://bugs.chromium.org/p/chromium/issues/detail?id=1083478&q=1083478). 
Do you have any suggestions or comments on the output?

kainino0x:
I don't know offhand why there is a difference between WebGL and WebGPU but I'm confused about the test. Since you mention LINEAR and NEAREST it sounds like you are using a draw call with a WebGL texture source to draw into the canvas? But that means you're comparing actual rendering results between WebGL and WebGPU.

The goal here of this test is to verify how a WebGPU canvas with particular contents appears. It should be the same as _either_\* a 2d canvas or a WebGL canvas that has exactly the same drawing buffer contents.  So instead of using scissor+clear to set the contents of a framebuffer texture you can just use them on the default framebuffer directly.

\* Also I don't think a WebGL reference page is necessarily required. The reference can probably just be 2d canvas.

shaoboyan:
@kainino0x  Yes, the reftest will be updated to use 2D canvas as reference. Using WebGL try to understand the difference but seems not work.

shaoboyan:
@kainino0x  PTAL, thanks.
(BTW, current output, scale 500%):
- Ref:
2d-ref

- WebGPU output:
WebGPU-Output

src/webgpu/web_platform/reftests/canvas_size_different_with_back_buffer_size.html
```diff
@@ -0,0 +1,20 @@
+<html class="reftest-wait">
+  <title>WebGPU canvas_back_buffer_different_size</title>
+  <meta charset="utf-8" />
+  <link rel="help" href="https://gpuweb.github.io/gpuweb/" />
+  <meta
+    name=fuzzy
+    content="128;48"
+  />
```
kainino0x, line 8: Per https://web-platform-tests.org/writing-tests/reftests.html#fuzzy-matching
this are very lenient settings. It may be good enough, but do we really need it to be so lenient? Does it really need to be fuzzy at all?

src/webgpu/web_platform/reftests/canvas_size_different_with_back_buffer_size.html.ts
```diff
@@ -0,0 +1,208 @@
...
+      const bytesPerRow = 256;
+      const buffer = t.device.createBuffer({
+        mappedAtCreation: true,
+        size: rows * bytesPerRow,
+        usage: GPUBufferUsage.COPY_SRC,
+      });
+      const mapping = buffer.getMappedRange();
+      const data = new Uint8Array(mapping);
+      // 1st row
+      data.set(red, 0); // red
```
kainino0x, line 86: nit: these comments are now redundant, also several of them are wrong.

src/webgpu/web_platform/reftests/canvas_size_different_with_back_buffer_size.html.ts
```diff
@@ -0,0 +1,208 @@
...
+      data.set(red, 4); // red
+      data.set(green, 8); // green
+
+      // 2nd row
+      data.set(red, 256 + 0); // red
+      data.set(red, 256 + 4); // red
+      data.set(green, 256 + 8); // blue
```
kainino0x, line 93: I think you can factor out a small helper so each canvas can have something as simple as:
```ts
setRowColors(data, 256 * 0, [red, red, green]);
setRowColors(data, 256 * 1, [red, red, green]);
setRowColors(data, 256 * 2, [blue, blue, yellow]);
setRowColors(data, 256 * 3, [blue, blue, yellow]);
```

That way these three cases will be way less repetitive.

src/webgpu/web_platform/reftests/canvas_size_different_with_back_buffer_size.html.ts
```diff
@@ -0,0 +1,208 @@
...
+      data.set(blue, 256 * 3 + 4); // blue
+      data.set(yellow, 256 * 3 + 8); // yellow
+      buffer.unmap();
+
+      const texture = ctx.getCurrentTexture();
+
+      const encoder = t.device.createCommandEncoder();
+      encoder.copyBufferToTexture({ buffer, bytesPerRow }, { texture }, [3, 4, 1]);
+      t.device.queue.submit([encoder.finish()]);
+      await t.device.queue.onSubmittedWorkDone();
```
kainino0x, line 111: nit: the onSubmittedWorkDone calls shouldn't be needed

src/webgpu/web_platform/reftests/ref/canvas_size_different_with_back_buffer_size-ref.html
```diff
@@ -0,0 +1,45 @@
...
+    const context = cvs_larger_than_back_buffer_ref.getContext('2d');
+    context.fillStyle = "#FF0000";
+    context.fillRect(0, 0, 4, 4);
+    context.fillStyle = "#00FF00";
+    context.fillRect(4, 0, 2, 4);
+    context.fillStyle = "#0000FF";
+    context.fillRect(0, 4, 4, 4);
+    context.fillStyle = "#FFFF00";
+    context.fillRect(4, 4, 2, 4);
+    }
```
kainino0x, line 19: To strictly match how we expect WebGPU to work, this canvas's inherent size should actually be 3x4, but its _CSS_ size should be 6px by 8px.

src/webgpu/web_platform/reftests/ref/canvas_size_different_with_back_buffer_size-ref.html
```diff
@@ -0,0 +1,45 @@
+<html>
+  <title>WebGPU canvas_back_buffer_different_size (ref)</title>
+  <meta charset="utf-8" />
+  <link rel="help" href="https://gpuweb.github.io/gpuweb/" />
+  <canvas id="cvs_larger_than_back_buffer_ref" width="6" height="8"></canvas>
```
kainino0x, line 5: To strictly match how we expect WebGPU to work, this canvas's inherent size should actually be 3x4, but its _CSS_ size should be 6px by 8px. `width=6 height=8 style="width:6px;height:8px;"`

kainino0x (commented):
These look good so far but we also need to test:
- What happens when the `canvas.width` or `canvas.height` gets changed via JS? (The getCurrentTexture() textures should stay the same size as they were originally configured with, but the canvas's layout size should change.)
- Then what happens if the context is configured again (with implicit sizes)? (configure should use the new canvas size and new getCurrentTexture() textures should now have the new size.)
- And finally, what happens when the canvas has an explicit CSS size, and the CSS size, canvas size, and configure size are all different?

src/webgpu/web_platform/reftests/canvas_size_different_with_back_buffer_size.html
shaoboyan: You're right, we don't need be fuzzy at all after applying CSS!
(My state: Idiot on CSS -- -Level up! ---> rookie on CSS :P )

src/webgpu/web_platform/reftests/canvas_size_different_with_back_buffer_size.html.ts
shaoboyan: Thanks for the suggestion.

src/webgpu/web_platform/reftests/canvas_size_different_with_back_buffer_size.html.ts
shaoboyan: Adding this previously to ensure when the test achieve to the end, the canvas should present things. But agree to remove it.

shaoboyan:
@kainino0x Big thanks for the comments about using CSS, now we got the output like this:
-- WebGPU output:
WebGPU-Output

-- Ref:
webgpu-ref

One remain question:
When implementing the canvas size, back buffer size and css size are all different case, I tried to construct a case that back buffer size > canvas size but css size is the largest. I expect this will finally be back buffer --- downscale ---> canvas size -- upcale --> CSS size.
But not sure how to generate the 2D canvas to simulate the downscale and upscale progress. Is it possible to create such case?

src/webgpu/web_platform/reftests/canvas_size_different_with_back_buffer_size.html.ts
kainino0x: Any DOM changes in the test should appear on the screen in the reftest capture - and that includes drawing to canvases like we do here. (In practice I think the reftest frameworks wait a few frames before grabbing a capture.)

kainino0x (commented):
LGTM except for one new comment. Glad this was an educational test to work on!

> When implementing the canvas size, back buffer size and css size are all different case, I tried to construct a case that back buffer size > canvas size but css size is the largest. I expect this will finally be back buffer --- downscale ---> canvas size -- upcale --> CSS size.
> But not sure how to generate the 2D canvas to simulate the downscale and upscale progress. Is it possible to create such case?

There is no separate downscale step. Compositing doesn't care about the actual canvas size, it just displays the backbuffer at the layout size. (The layout size is set by CSS, but if there is no CSS it defaults to the canvas .width and .height. Just like how configure() takes an optional size, but defaults to the canvas .width and .height.)

So let's add the case you suggested like so:
- webgpu: configure 6x8, `<canvas>` 3x4, css 12x16
- ref: `<canvas>` 6x8, css 12x16

kainino0x:
> So let's add the case you suggested like so:
> 
> * webgpu: configure 6x8, `<canvas>` 3x4, css 12x16
> * ref: `<canvas>` 6x8, css 12x16

Also make sure the pixel pattern you use verifies that it's not getting downsampled before being upsampled.

shaoboyan:
@kainino0x  Ah, thanks for the explain about how compositor handle size! I think the comments has been address. PTAL again, thanks!

src/webgpu/web_platform/reftests/canvas_size_different_with_back_buffer_size.html.ts
```diff
@@ -163,19 +170,48 @@ export function run() {
...
+        [
+          [red, red, red, red, green, green],
+          [red, red, red, red, green, green],
+          [red, red, red, red, green, green],
+          [red, red, red, red, green, green],
+          [blue, blue, blue, blue, yellow, yellow],
+          [blue, blue, blue, blue, yellow, yellow],
+          [blue, blue, blue, blue, yellow, yellow],
+          [blue, blue, blue, blue, yellow, yellow],
+        ]
```
kainino0x, line 213: nit: I think we wouldn't be able to tell if this were downsampled to 3x4 before being upsampled to 12x16

kainino0x (approved):

src/webgpu/web_platform/reftests/canvas_size_different_with_back_buffer_size.html.ts
kainino0x: Fixed by adding one pixel

shaoboyan: That's clever, thanks and got the skill.

PR #690, opened 2021-08-03
Test buffer related validations for texture copies

Richard-Yunchao:
The new added cases are:
  - tests that buffer must be valid when buffer is valid, destroy, or error buffer
  - tests that bytesPerRow must be 256-bytes aligned when it is required

<hr>


src/webgpu/api/validation/image_copy/buffer_related.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+      size: 512 * 8 * 16,
+      usage: GPUBufferUsage.COPY_SRC | GPUBufferUsage.COPY_DST,
+    });
+
+    let success = false;
+    // writeTexture doesn't require bytersPerRow to be 256-byte aligned.
+    if (method === 'WriteTexture') success = true;
+    // If the copy height <= 1, bytesPerRow is not required.
+    if (copyHeight / info.blockHeight <= 1 && bytesPerRow === undefined) success = true;
+    // If bytesPerRow > 0 and it is a multiple of 256, it will succeeed if other parameters are valid.
```
Richard-Yunchao, line 101: If we are copying 0 or 1 row (copyHeight / info.blockHeight <= 1), we don't need to assign a value for rowsPerImage. But what happen if we assign a value for it? For example:
  1) bytesPerRow is 0 or a number that < bytes required for one row data for writeTexture (which doesn't require 256B aligned), is it valid? 
  2) bytesPerRow is 0 or a number that < bytes required for one row data and 256B aligned, it should be invalid, because the spec says "If specified, layout.bytesPerRow must be greater than or equal to bytesInLastRow." for B2T and T2B copies at 12.3.7 <Image Copies>.
 
  Considering the statement at 12.3.7 <Image Copies>, maybe we should add similar statement in order to clarify this edge case at 12.3.1 <GPUImageDataLayout>, WDYT?

Richard-Yunchao:
When I was adding tests to validate 3D textures for writeTexture, B2TCopy and T2BCopy, I found that the validations for 2D textures are not complete. So I'd like to add the missing tests. Some of them have already been added as TODOs.

src/webgpu/api/validation/image_copy/buffer_related.spec.ts
kainino0x: > * bytesPerRow is 0 or a number that < bytes required for one row data for writeTexture (which doesn't require 256B aligned), is it valid?

It should have the same rule as B2T and T2B; `writeTexture` also calls into the `validating linear texture data` algorithm - though when it fails, it throws an exception instead of a validation error, because the data is being read out of an ArrayBuffer instead of a GPUBuffer.

> Considering the statement at 12.3.7 , maybe we should add similar statement in order to clarify this edge case at 12.3.1 <GPUImageDataLayout>, WDYT?

Adding it here could be a useful reference, but it isn't strictly necessary. The validation rules need to be specified in algorithms anyway so it's clear what happens when they fail.

src/webgpu/api/validation/image_copy/buffer_related.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+        break;
+      }
+      case 'error': {
+        buffer = t.getErrorBuffer();
+        break;
+      }
+    }
+
+    const success = bufferState === 'valid';
+    const submit = bufferState === 'destroyed';
```
kainino0x, line 44: This might be simpler with `createEncoder` + `validateFinishAndSubmitGivenState`

src/webgpu/api/validation/image_copy/image_copy.ts
```diff
@@ -98,6 +98,72 @@ export class ImageCopyTest extends ValidationTest {
...
+        if (submit) {
+          const cmd = encoder.finish();
+          this.expectValidationError(() => {
+            this.device.queue.submit([cmd]);
+          }, !success);
+        } else {
+          this.expectValidationError(() => {
+            encoder.finish();
+          }, !success);
+        }
```
kainino0x, line 144: You can use `createEncoder` + `validateFinishAndSubmit` to simplify these a bit.

src/webgpu/api/validation/image_copy/buffer_related.spec.ts
```diff
@@ -0,0 +1,149 @@
...
+
+    let success = false;
+    // writeTexture doesn't require bytesPerRow to be 256-byte aligned.
+    if (method === 'WriteTexture') success = true;
+    // If the copy height <= 1, bytesPerRow is not required.
+    if (copyHeight / info.blockHeight <= 1 && bytesPerRow === undefined) success = true;
+    // If bytesPerRow > 0 and it is a multiple of 256, it will succeeed if other parameters are valid.
+    if (bytesPerRow !== undefined && bytesPerRow > 0 && bytesPerRow % 256 === 0) success = true;
+
+    const size = [info.blockWidth, !copyHeight ? info.blockHeight : copyHeight, 1];
```
kainino0x, line 136: nit: I think this is equivalent? and clearer; I initially wasn't sure whether copyHeight could be undefined or something.
```suggestion
    const size = [info.blockWidth, copyHeight === 0 ? info.blockHeight : copyHeight, 1];
```

src/webgpu/api/validation/image_copy/image_copy.ts
kainino0x: I don't understand why `validateFinishAndSubmit` needs an extra argument? It basically does two things, finish and submit, and it takes two boolean arguments for whether each of them succeed. `validateFinishAndSubmitGivenState` may be too specific for this use-case, but it's just a wrapper around `validateFinishAndSubmit` for convenience so you don't have to use it.

As for why the default argument didn't work, it's because you would have to add it to the function definition (in the constructor of CommandBufferMaker), not the type definition of the field. (That class is set up really weirdly so that destructuring its methods works, so it's confusing.)

src/webgpu/api/validation/image_copy/buffer_related.spec.ts
kainino0x: On second inspection I agree validateFinishAndSubmitGivenState won't work.

src/webgpu/api/validation/image_copy/buffer_related.spec.ts
```diff
@@ -136,25 +126,24 @@ g.test('bytes_per_row_alignment')
...
     });
 
-    t.testBuffer(
-      buffer,
-      texture,
-      { bytesPerRow },
-      { width: info.blockWidth, height: copyHeight, depthOrArrayLayers: 1 },
-      { dataSize: 512 * 8 * 16, method, success }
-    );
+    t.testBuffer(buffer, texture, { bytesPerRow }, size, {
```
kainino0x, line 144: AFAICT this no longer tests the case where copyHeight is 0. So it's just a duplicate of another case (the one where copyHeight == blockHeight)

kainino0x: (Because it was changed to pass `size` here instead of the original `copyHeight`)

src/webgpu/api/validation/image_copy/buffer_related.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+    switch (bufferState) {
+      case 'destroyed': {
+        buffer.destroy();
+        break;
+      }
+      case 'error': {
+        buffer = t.getErrorBuffer();
+        break;
+      }
+    }
```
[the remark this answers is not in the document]
kainino0x: Please mark conversations as "resolved" when they're done so it's possible to see which comments still need attention

src/webgpu/api/validation/image_copy/image_copy.ts
kainino0x: i.e.:
```ts
const { encoder, validateFinishAndSubmit } = this.createEncoder('non-pass');
encoder.copyBufferToTexture({ buffer, ...textureDataLayout }, textureCopyView, size);
validateFinishAndSubmit(success, submit);
```

src/webgpu/api/validation/image_copy/buffer_related.spec.ts
Richard-Yunchao: Agreed! I reset size.height because texture's height cannot be zero, otherwise it will fail to create a texture. However, copyHeight is 0 while texture's height >  0 is not allowed for depth/stencil formats, so I use texture's size as copySize in order to make sure it is a full copy. Then copyHeight == 0 is not useful, like you said. 

In the updated code, I filtered out copyHeight == 0 for depth/stencil formats but use 0 as copyheight for other formats.

src/webgpu/api/validation/image_copy/buffer_related.spec.ts
Richard-Yunchao: Done.

src/webgpu/api/validation/image_copy/image_copy.ts
Richard-Yunchao: Oops. I don't remember why I insisted on using validateFinishAndSubmitGivenState and didn't try to use validateFinishAndSubmit directly. Your suggestion is correct. Thanks!

src/webgpu/api/validation/image_copy/buffer_related.spec.ts
```diff
@@ -104,8 +104,15 @@ g.test('bytes_per_row_alignment')
       .beginSubcases()
       .combine('bytesPerRow', [undefined, 0, 1, 255, 256, 257, 512])
       .combine('copyHeight', [0, 1, 2, 4, 8])
-      // copyHeight must be aligned with blockHeight
-      .filter(({ format, copyHeight }) => copyHeight % kTextureFormatInfo[format].blockHeight === 0)
+      // copyHeight has some constraints:
+      //   - it must be aligned with blockHeight.
```
kainino0x, line 108: In many tests we solve this by having the test parameters provide e.g. `copyHeightInBlocks`. Could that be done here? This works now but would break down if the texel block height were 6 for example. (which happens with ASTC)

src/webgpu/api/validation/image_copy/buffer_related.spec.ts
```diff
@@ -104,8 +104,15 @@ g.test('bytes_per_row_alignment')
...
+      //   - if it is 0, the format cannot be a depth/stencil format. Because it must be a full copy if the format is a depth/stencil format but copied subresource's height can never be 0.
```
kainino0x, line 109: This restriction comes in because of how copyHeight=0 is specially treated (that it's rounded up to blockHeight). It's confusing here without reading the test even though it's explained by the comment.

That's ok, but maybe possible to do something like:

```ts
.combine('copyHeightInBlocks', [0, 1, 2, 4, 8])
.expand('_textureHeightInBlocks', p => p.copyHeightInBlocks === 0 ? 1 : p.copyHeightInBlocks)
// Depth/stencil format copies must copy the whole subresource.
.unless(p => {
  const info = kTextureFormatInfo[p.format];
  return (info.depth || info.stencil) && p.copyHeightInBlocks !== p._textureHeightInBlocks;
})
```

src/webgpu/api/validation/image_copy/buffer_related.spec.ts
```diff
@@ -133,15 +140,17 @@ g.test('bytes_per_row_alignment')
     // If bytesPerRow > 0 and it is a multiple of 256, it will succeeed if other parameters are valid.
     if (bytesPerRow !== undefined && bytesPerRow > 0 && bytesPerRow % 256 === 0) success = true;
 
-    const size = [info.blockWidth, !copyHeight ? info.blockHeight : copyHeight, 1];
+    const size = [info.blockWidth, copyHeight === 0 ? info.blockHeight : copyHeight, 1];
```
kainino0x, line 143: Height here would become `_textureHeightInBlocks * info.blockHeight`

kainino0x (approved):
Sorry for the delay, LGTM but I had a ideas for small improvements that should hopefully be easy.

src/webgpu/api/validation/image_copy/buffer_related.spec.ts
Richard-Yunchao: Done.

src/webgpu/api/validation/image_copy/buffer_related.spec.ts
Richard-Yunchao: Thanks for the suggestion, Kai. I tried the code, looks like `expand` requires the expanded parameter `_textureHeightInBlocks` to be iterable but it is not in the code snippet. So it failed to build. 

Any suggestion? Maybe we can just merge this PR?

kainino0x: Sorry, that was missing a pair of brackets.

`.expand('_textureHeightInBlocks', p => [p.copyHeightInBlocks === 0 ? 1 : p.copyHeightInBlocks])`

Richard-Yunchao: Done. I should have figured out this.

src/webgpu/api/validation/image_copy/buffer_related.spec.ts
Richard-Yunchao: Done.

kainino0x:
No problem, life always takes priority!
Sorry that you had to return to this PR after so much time due to small changes.

PR #691, opened 2021-08-03
Validate texture format for copies

Richard-Yunchao:
If the texture format is depth/stencil format, it must be a full
copy for copies resource(s).

<hr>


src/webgpu/api/validation/image_copy/texture_related.spec.ts
```diff
@@ -127,7 +127,7 @@ g.test('sample_count')
       size: { width: 4, height: 4, depthOrArrayLayers: 1 },
       sampleCount,
       format: 'rgba8unorm',
-      usage: GPUTextureUsage.COPY_SRC | GPUTextureUsage.COPY_DST | GPUTextureUsage.SAMPLED,
```
Richard-Yunchao, line 130: sample_count test doesn't mean that we need SAMPLED usage.

Richard-Yunchao:
Like I said at another PR, when I was adding tests to validate 3D textures for writeTexture/B2TCopy/T2BCopy, I found that the validations for 2D textures are not complete. So I'd like to add the missing tests. Some of them have already been added as TODOs, the others are not in TODOs.

src/webgpu/api/validation/image_copy/texture_related.spec.ts
```diff
@@ -175,6 +175,73 @@ g.test('mip_level')
...
+    u //
+      .combine('method', kImageCopyTypes)
+      .combineWithParams([
+        { depthOrArrayLayers: 1, dimension: '2d' },
+        { depthOrArrayLayers: 3, dimension: '2d' },
+        { depthOrArrayLayers: 32, dimension: '3d' },
+      ] as const)
+      .combine('format', kSizedTextureFormats)
+      .filter(({ dimension, format }) => textureDimensionAndFormatCompatible(dimension, format))
+      .filter(formatCopyableWithMethod)
```
Richard-Yunchao, line 190: BTW, this test actually can't test what I wanted to test, because copySrc/copyDst is marked as false in capability_info.js. As a result, depth/stencil formats won't be tested in this test. 

We discussed the issue about copySrc/copyDst of capability_info.js at https://github.com/gpuweb/cts/pull/481. But we didn't address that issue because it is not finalized in WebGPU spec. Now that the spec says that we can do full copy for depth/stencil format, I think it's time to change copySrc/copyDst in capability_info.js. WDYT?

kainino0x: I agree we should do it. I am not exactly sure how we should represent it in the table. Maybe one sub-structure for each aspect? `{ color: {...} | undefined, depth: {...} | undefined, stencil: {...} | undefined }`
but it would result in some churn in the rest of the project.

Another option, we could use the existing fields just for color, and add sub-structures (or just new fields) for depth/stencil?

We'd probably also need some helper that takes a GPUTextureAspect and fetches the appropriate aspect sub-structure for you.

src/webgpu/api/validation/image_copy/texture_related.spec.ts
```diff
@@ -175,6 +175,73 @@ g.test('mip_level')
...
+      success = false;
+    }
+
+    const copySize = [
+      (size.width >> mipLevel) + copyWidthModifier * info.blockWidth,
+      (size.height >> mipLevel) + copyHeightModifier * info.blockHeight,
+      // Note that compressed format is not supported for 3D textures yet, so there is no info.blockDepth.
+      dimension === '3d'
+        ? (size.depthOrArrayLayers >> mipLevel) + copyDepthModifier
+        : size.depthOrArrayLayers,
```
kainino0x, line 235: Shouldn't this still have the copyDepthModifier applied?

src/webgpu/api/validation/image_copy/texture_related.spec.ts
```diff
@@ -175,6 +175,73 @@ g.test('mip_level')
...
+    ];
```
kainino0x, line 236: The `virtualMipSize` might make slightly more straightforward:
```
const levelSize = virtualMipSize(/* ... */);
const copySize = [ levelSize[0] + copyWidthModifier * info.blockWidth, /* ... */ ];
```

src/webgpu/api/validation/image_copy/README.txt
```diff
@@ -9,6 +9,7 @@ Test coverage:
 	- texture_must_be_valid: for valid, destroyed, error textures.
 	- sample_count_must_be_1: for sample count 1 and 4.
 	- mip_level_must_be_in_range: for various combinations of mipLevel and mipLevelCount.
+	- full_copy_for_depth_stencil_format: for full and non-full copies on width, height, and depth.
```
kainino0x, line 12: Test name doesn't match. I think the others might not match anymore either, whoops.

FWIW it's not super necessary that we keep this README up to date. I'm supposed to dissolve this README out into proper per-test descriptions, but I haven't done it yet. But better to make this one thing match now so it'll be easier to dissolve later.

src/webgpu/api/validation/image_copy/texture_related.spec.ts
Richard-Yunchao: Hmm, I found that we have already had function depthStencilBufferTextureCopySupported and kDepthStencilFormatCapabilityInBufferTextureCopy. We can use that function.

Richard-Yunchao: I added tests for SizedDepthStencilFormats based on existing information/functions in capability_info.ts. I also added a couple more functions. UnsizedDepthStencilFormats will be the next step, whose BytesPerBlock should be set per aspect but it is set per format in capability_info.ts.

src/webgpu/api/validation/image_copy/texture_related.spec.ts
Richard-Yunchao: If it is 2D array texture, copyDepthModifier will always be 0 (I added some comments for it at copyDepthModifier). Adding it or not is the same.

src/webgpu/api/validation/image_copy/texture_related.spec.ts
Richard-Yunchao: Done!

src/webgpu/api/validation/image_copy/README.txt
Richard-Yunchao: Done. Now that the document is still there, I'd like to make it consistent with tests. We can remove the file (and move description to each test if we haven't) after all tests related to copy validation are done.

src/webgpu/api/validation/image_copy/texture_related.spec.ts
kainino0x: Ah I see, sorry I didn't understand this.

kainino0x (approved):
LGTM

PR #692, opened 2021-08-03
Implement External Texture Follow Up Test Changes

bjjones:
<hr>

### DIFF
--- a/src/webgpu/web_platform/external_texture/video.spec.ts
+++ b/src/webgpu/web_platform/external_texture/video.spec.ts
@@ -8,7 +8,6 @@ TODO: consider whether external_texture and copyToTexture video tests should be 
 
 import { getResourcePath } from '../../../common/framework/resources.js';
 import { makeTestGroup } from '../../../common/framework/test_group.js';
-import { resolveOnTimeout } from '../../../common/util/util.js';
 import { GPUTest } from '../../gpu_test.js';
 import { startPlayingAndWaitForVideo } from '../../web_platform/util.js';
 
@@ -101,7 +100,7 @@ g.test('importExternalTexture,sample')
     `
 Tests that we can import an HTMLVideoElement into a GPUExternalTexture, sample from it for all
 supported video formats {vp8, vp9, ogg, mp4}, and ensure the GPUExternalTexture is destroyed by
-a microtask. 
+a microtask.
 TODO: Multiplanar scenarios
 `
   )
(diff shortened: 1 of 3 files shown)


src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -151,26 +151,105 @@ Test that importing different video formats into an external texture and samplin
...
+    video.src = videoUrl;
+
+    await startPlayingAndWaitForVideo(video, () => {
+      const colorAttachment = t.device.createTexture({
+        format: kFormat,
+        size: { width: kWidth, height: kHeight, depthOrArrayLayers: 1 },
+        usage: GPUTextureUsage.COPY_SRC | GPUTextureUsage.RENDER_ATTACHMENT,
+      });
+
+      const pipeline = t.device.createRenderPipeline({
```
kainino0x, line 175: Are all these helper objects (pipeline, bind group, etc.) copied from above? They're not really material to this test, so maybe they can be split into a helper function and shared.

src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -151,26 +151,105 @@ Test that importing different video formats into an external texture and samplin
...
+      passEncoder.endPass();
 
-        t.expectGPUError('validation', () => t.device.queue.submit([commandEncoder2.finish()]));
+      // requestAnimationFrame should trigger a microtask that destroys the GPUExternalTexture,
+      // which will cause a validation error when we try to use the destroyed GPUExternalTexture
+      const ensureDestroyed = function () {
+        t.expectGPUError('validation', () => t.device.queue.submit([commandEncoder.finish()]));
       };
 
       requestAnimationFrame(ensureDestroyed);
```
kainino0x, line 255: This can be `timeout(ensureDestroyed, 0)`, which is a little bit stricter. (`timeout` is a util wrapper around `setTimeout`.)

If we want to be perfectly strict, we should really test that it gets scheduled as a microtask at exactly the right time. This gets kind of in the weeds, but what we can do is issue three microtasks by taking advantage of the fact that .then() on an already-resolved promise is always scheduled as a microtask:

```ts
let externalTexture;
Promise.resolve().then(() => { /* here, externalTexture should be alive */ });
externalTexture = device.importExternalTexture(/* ... */);
Promise.resolve().then(() => { /* here, externalTexture should have been destroyed */ });
await resolveOnTimeout(0); // [1]
```

(I'm not _100%_ sure this will work, so let me know if something seems wrong.)

[1] AFAICT, nothing is strictly ensuring that the test waits for this to happen before ending. Probably it works because `await startPlayingAndWaitForVideo` ends up waiting long enough after the callback for whatever reason.
With the changes above to make everything happen in microtasks, it should be enough to wait for a (non-micro) task like `setTimeout`, as microtasks always happen before tasks.

kainino0x (approved):
Compute test LGTM, left a comment about making the destroy test stricter

src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -10,10 +10,10 @@ Test Coverage:
   supported video formats {vp8, vp9, ogg, mp4}, and ensure the GPUExternalTexture is destroyed by
   a microtask. 
     TODO: Multiplanar scenarios
-    TODO: Pull microtask destruction into its own test
+  - Tests that a GPUExternalTexture is destroyed by a microtask and that using it after it has been
+  destroyed results in an error.
```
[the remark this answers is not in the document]
bjjones: Done.

src/webgpu/web_platform/external_texture/video.spec.ts
bjjones: _Seems_ to work.

src/webgpu/web_platform/external_texture/video.spec.ts
bjjones: Added GetPipeline, GetBG helpers.

src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -165,15 +201,19 @@ Test that importing different video formats into an external texture and samplin
...
+        passEncoder.draw(6);
+        passEncoder.endPass();
+      });
 
-        t.expectGPUError('validation', () => t.device.queue.submit([commandEncoder2.finish()]));
-      };
+      // This should trigger a microtask that destroys the GPUExternalTexture, which will cause
+      // a validation error when we try to use the destroyed GPUExternalTexture.
+      await Promise.resolve().then(() => {
+        t.expectGPUError('validation', () => t.device.queue.submit([commandEncoder.finish()]));
```
kainino0x, line 213: nit: This puts two operations (finish and submit) in the callback, so we don't know which it came from.

Anyway I don't think the micro/tasks are quite issuing correctly in this code, so I'm looking at it locally.

src/webgpu/web_platform/external_texture/video.spec.ts
kainino0x: fwiw I figured out something less dumb than resolveOnTimeout: just await on the two promises we created

src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -151,12 +162,37 @@ Test that importing different video formats into an external texture and samplin
...
+  )
+  .fn(async t => {
+    const videoUrl = getResourcePath('red-green.webmvp8.webm');
+    const video = document.createElement('video');
+    video.src = videoUrl;
+
+    let commandEncoder: GPUCommandEncoder;
+
+    await startPlayingAndWaitForVideo(video, async () => {
+      await Promise.resolve().then(() => {
```
kainino0x, line 183: Oops, while editing this test I realized this async function is not awaited by startPlayingAndWaitForVideo, which means the test was still not waiting for completion. (I had an error in the code that threw an exception, and it was on the console but the test was green.)

src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -151,12 +162,37 @@ Test that importing different video formats into an external texture and samplin
...
+`
+  )
+  .fn(async t => {
+    const videoUrl = getResourcePath('red-green.webmvp8.webm');
+    const video = document.createElement('video');
+    video.src = videoUrl;
+
+    let commandEncoder: GPUCommandEncoder;
+
+    await startPlayingAndWaitForVideo(video, async () => {
```
kainino0x, line 182: Oops, while editing this test I realized this async function is not awaited by startPlayingAndWaitForVideo, which means the test was still not waiting for completion. (I had an error in the code that threw an exception, and it was on the console but the test was green.)

kainino0x:
@bjjones OK, I pushed a commit which I think issues the microtasks correctly. Sorry for partially rewriting, it was too hard to try to explain the changes over review comments.
I also realized while I was editing that the pipeline and draw call weren't actually needed at all. (I didn't factor them back inline, though, I think the helper functions are good to keep.)
PTAL at the last commit!

This test still passes on Chrome on my Mac (and I manually tried several variants to make sure it was failing when expected, instead of having unhandled errors like before).

bjjones:
LGTM. Passes on Chrome/Win as well. Again, thanks for the help - I'm slowly starting to figure out JS's concurrency model.

kainino0x:
No worries, JS's execution model is super hard to figure out. Took me at least a few months...

PR #695, opened 2021-08-04
Add tests for requiredLimits in requestDevice

toji:
Adds some basic tests for various `requiredLimits` configurations being passed to `requestDevice`. The only limit not covered is `maxComputeWorkgroupSize` because there's currently some confusion about how non-numeric limits should work.

Some limits are currently unsupported in Chrome, so several of the subcases fail there, but I've tested against a WIP Chrome patch that adds the missing limits and it works as expected.

<hr>

### DIFF
--- a/src/webgpu/api/operation/adapter/requestDevice_limits.spec.ts
+++ b/src/webgpu/api/operation/adapter/requestDevice_limits.spec.ts
@@ -1,0 +1,148 @@
+export const description = `
+Tests passing various requiredLimits to GPUAdapter.requestDevice.
+`;
+
+import { Fixture } from '../../../../common/framework/fixture.js';
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { keysOf } from '../../../../common/util/data_tables.js';
+import { assert } from '../../../../common/util/util.js';
+import { DefaultLimits } from '../../../constants.js';
+
+const kLimitTypes = keysOf(DefaultLimits);
+
+const kMaxUnsignedLongValue = 4294967295;
+/** Clamps a numeric value to the valid unsigned long range, as defined by WebIDL */
+function clampToUnsignedLong(value: number): number {
+  return Math.min(kMaxUnsignedLongValue, Math.max(0, value));
+}
+
+export const g = makeTestGroup(Fixture);
+
+g.test('unknown_limits')
+  .desc(
+    `
+    Test that specifiying limits that aren't part of the supported limit set causes
+    requestDevice to reject.`
+  )
+  .fn(async t => {
+    const adapter = await navigator.gpu.requestAdapter();
+    assert(adapter !== null);
+
+    const requiredLimits: Record<string, number> = { unknownLimitName: 9000 };
+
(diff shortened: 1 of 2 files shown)


src/webgpu/api/operation/adapter/requestDevice_limits.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+export const g = makeTestGroup(Fixture);
+
+g.test('unknown_limits')
+  .desc(
+    `
+    Test that specifiying limits that aren't part of the supported limit set causes
+    requestDevice to reject.`
+  )
+  .fn(async t => {
+    const adapter = await navigator.gpu.requestAdapter();
```
kainino0x, line 28: nit: use the `getGPU()` function instead of `navigator.gpu`, for eventually when these tests aren't running in a browser.

src/webgpu/api/operation/adapter/requestDevice_limits.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+    u.combine('limit', kLimitTypes).unless(p => typeof DefaultLimits[p.limit] !== 'number')
+  )
+  .fn(async t => {
+    const { limit } = t.params;
+
+    const adapter = await navigator.gpu.requestAdapter();
+    assert(adapter !== null);
+
+    const requiredLimits: Record<string, number> = {};
+    requiredLimits[limit] = DefaultLimits[limit];
```
kainino0x, line 53: uNit: I believe something like this would work:
```suggestion
    const requiredLimits = { [limit]: DefaultLimits[limit] } as const;
```

kainino0x (approved):

src/webgpu/api/operation/adapter/requestDevice_limits.spec.ts
toji: Done! Thanks for pointing that out.

src/webgpu/api/operation/adapter/requestDevice_limits.spec.ts
```diff
@@ -46,11 +47,13 @@ g.test('default_limits')
...
     const { limit } = t.params;
 
-    const adapter = await navigator.gpu.requestAdapter();
+    const gpu = getGPU();
+    const adapter = await gpu.requestAdapter();
     assert(adapter !== null);
 
-    const requiredLimits: Record<string, number> = {};
-    requiredLimits[limit] = DefaultLimits[limit];
+    if (adapter.limits) { return; }
```
kainino0x, line 54: this check seems wrong?

toji: That was an accidental leftover from some previous debugging. I've got it removed locally, but I'm kind of waiting to see what happens with https://github.com/gpuweb/gpuweb/pull/2020 before pushing this any futher, because if that lands then I can solve some of my problems with this CL by updating the webgpu types and updating the defaults in the CTS.

kainino0x (approved):

toji:
Bwa ha ha ha! After almost a week of yak shaving to get all the pieces in place so that these tests can pass, I can finally commit this!  😤

PR #696, opened 2021-08-06
Reftests expand with formats, write canvas methods, and compositingAlphaMode

shrekshao:
Add reftests for different formats (actually only bgra8unorm for now)
and different write to canvas methods
  * 'copyBufferToTexture'
  * 'copyTextureToTexture'
  * 'copyExternalImageToTexture'
  * 'DrawTextureSample'
  * 'DrawVertexColor'

~~Tests for different compositingAlphaMode to come.~~
compositingAlphaMode
* 'premultiplied'
* 'opaque'

Q: for `-srgb` context format, do we need a pre-curved texture/color source? Or the backend already handles the linear to srgb space conversion so we pass the same value into the context texture? (Maybe it's different for buffer to canvas and shader to canvas cuz I think it's always linear space in shader, but generally confused here)

<hr>

- [x] (Optional, sometimes not possible.) Tests pass (or partially pass without unexpected issues) in an implementation. (chrome only support `brga8unorm` now. other formats are untested.)


kainino0x:
> Q: for `-srgb` context format, do we need a pre-curved texture/color source? Or the backend already handles the linear to srgb space conversion so we pass the same value into the context texture? (Maybe it's different for buffer to canvas and shader to canvas cuz I think it's always linear space in shader, but generally confused here)

It is indeed different: No conversions occur inside copy commands, they operate on raw data. E.g. if you copyTextureToTexture from rgba8unorm to rgba8unorm-srgb, the bytes will be copied directly (and sampling the resulting texture from a shader will give you different values).

The places where the srgb transfer function is applied are where they are converted to/from "continuous" values (shader floats):
- reading the texture from a shader (note: iirc it's undefined whether the transfer function is applied before or after filtering)
- writing the texture from a render pass (note: iirc it's undefined whether the transfer function is applied before or after blending)

The `-srgb` should also be ignored by the compositor; i.e., if you put the same bytes into the texture, it doesn't matter whether the canvas format is `rgba8unorm` or `rgba8unorm-srgb`, it will appear on screen the same way - that way a shader can output linear light values into an `rgba8unorm-srgb`canvas texture, they'll get gamma encoded, and the screen result will be correct. (If the user wanted something else, they would control it using `GPUCanvasConfiguration.colorSpace` which (TBD) will have an option like `"srgb-linear"`).

src/webgpu/web_platform/reftests/canvas_complex_bgra8unorm_copy_buffer_to_texture.html
```diff
@@ -12,7 +12,6 @@
   <script src="/common/reftest-wait.js"></script>
   <script type="module">
     import { run } from './canvas_complex.html.js';
-    run('bgra8unorm');
-    // TODO: make a copy of this test for rgba16float
+    run('bgra8unorm', 'copyBufferToTexture');
```
kainino0x, line 15: _Option_ to consider: we could combine multiple tests into a single html by putting multiple canvases in the test (and multiple canvases in the ref). We did this in #689 - though in that test, the cases are all very similar and there wouldn't have been value in splitting them. In these tests, the different cases are meaningfully different, but we could still consider combining some (e.g. B2T+T2T+EI2T could be one, DrawTextureSample and DrawVertexColor could be another). This would at least reduce the explosion of html files when we add more formats.

Formats could be combined too - like bgra8unorm-srgb + rgba8unorm-srgb. TBH eventually I think we would put every format in every case, but today we don't support all formats so we can't do that.

src/webgpu/web_platform/reftests/canvas_composite_alpha.html.ts
```diff
@@ -0,0 +1,129 @@
+import { assert, unreachable } from '../../../common/util/util.js';
+
+import { runRefTest } from './gpu_ref_test.js';
+
+// <canvas> element from html page
+declare const cvs: HTMLCanvasElement;
+
+export function run(format: GPUTextureFormat, compositingAlphaMode: GPUCanvasCompositingAlphaMode) {
+  runRefTest(async t => {
+    const ctx = cvs.getContext('gpupresent');
```
kainino0x, line 10: use 'webgpu' now

src/webgpu/web_platform/reftests/canvas_composite_alpha.html.ts
```diff
@@ -0,0 +1,129 @@
...
+      case 'bgra8unorm':
+      case 'bgra8unorm-srgb':
+      case 'rgba8unorm':
+      case 'rgba8unorm-srgb':
+        break;
+      default:
+        unreachable();
+    }
+
+    const a = compositingAlphaMode === 'opaque' ? (1.0).toFixed(1) : (0.5).toFixed(1);
```
kainino0x, line 23: The `'opaque'` compositingAlphaMode is supposed to ignore the contents of the alpha channel and display the texture as if alpha were cleared to 1. So I don't think this would do anything.

Actually it would be best to test multiple different alpha values (e.g. on different pixels, output  = 0.0, 0.5, or 1.0) just to make sure all cases behave the same way.

src/webgpu/web_platform/reftests/canvas_composite_alpha.html.ts
```diff
@@ -0,0 +1,129 @@
...
+                      srcFactor: 'src-alpha',
+                      dstFactor: 'one-minus-src-alpha',
+                      operation: 'add',
+                    },
+                    alpha: {
+                      srcFactor: 'one',
+                      dstFactor: 'one-minus-src-alpha',
+                      operation: 'add',
+                    },
+                  },
```
kainino0x, line 99: Add comment about what this blend state is. IIUC it's converting unpremultiplied shader values to premultiplied texture values. Also can be shorter by taking advantage of defaults (maybe that's too confusing?)
```suggestion
                : {
                    color: { srcFactor: 'src-alpha' }, // Premultiply alpha into the color channels
                    alpha: {} // Pass through alpha
                  },
```

kainino0x (approved):

src/webgpu/web_platform/reftests/canvas_complex.html.ts
```diff
@@ -1,78 +1,342 @@
...
+    vec2<f32>( -0.5,  0.5),
+    vec2<f32>( 0.5, 0.5),
+    vec2<f32>(-0.5, -0.5),
+    vec2<f32>( 0.5,  -0.5));
+
+  var color = array<vec4<f32>, 4>(
+      vec4<f32>(0.49804, 0.0, 0.0, 1.0),
+      vec4<f32>(0.0, 0.49804, 0.0, 1.0),
+      vec4<f32>(0.0, 0.0, 0.49804, 1.0),
+      vec4<f32>(0.49804, 0.49804, 0.0, 1.0)); // 0.49804 -> 0x7f
```
kainino0x, line 271: FYI for the srgb cases: you can use the existing `gammaDecompress` helper to find the corresponding float value for srgb 0x7f: `0.21223`

(may need to be bumped up or down a bit? I don't know what the rounding behavior for unorm storage is)

src/webgpu/web_platform/reftests/canvas_composite_alpha.html.ts
shrekshao: There's blending inside canvas happening here. We are having 4 draw calls so each draw result still blends with each other inside the canvas. This is to mimic the behavior of canvas 2d ref to achieve the same in-canvas result. (although what we are really testing is canvas blending with other elements on page)

src/webgpu/web_platform/reftests/canvas_composite_alpha.html.ts
shrekshao: Same here as above. We are drawing 4 quads in 4 different draw calls so the content blending inside the canvas context still matters (to match the result of canvas 2d ref tests)

shrekshao: image
Canvas is 4x4
Drawing 3x3 quads with different color for each corner in order. (I should update with some comments as well)

shrekshao:
* "address comments" is the only new commit
* Path for srgb output format is untested.
* Having multiple canvases for some cases is not done yet.

kainino0x (approved):
LGTM, sorry for misunderstanding the tests previously!

shrekshao:
New commits: `address review comments` and `merge complex canvas 3 in 1`
Thanks for the review!

kainino0x (approved):
lgtm!

PR #699, opened 2021-08-06
Roll @webgpu/types to 0.1.6

toji:
Rolls the WebGPU types from 0.1.4 to 0.1.5 and fixes any (visible) issues that resulted.

### DIFF
--- a/src/webgpu/api/validation/validation_test.ts
+++ b/src/webgpu/api/validation/validation_test.ts
@@ -324,9 +324,9 @@ export class ValidationTest extends GPUTest {
       occlusionQuerySet?: GPUQuerySet;
     } = {}
   ): CommandBufferMaker<T> {
-    const fullAttachmentInfo: GPURenderBundleEncoderDescriptor = {
+    const fullAttachmentInfo = {
       // Defaults if not overridden:
-      colorFormats: ['rgba8unorm'] as GPUTextureFormat[],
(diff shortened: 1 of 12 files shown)


src/webgpu/api/validation/validation_test.ts
```diff
@@ -324,9 +324,9 @@ export class ValidationTest extends GPUTest {
       occlusionQuerySet?: GPUQuerySet;
     } = {}
   ): CommandBufferMaker<T> {
-    const fullAttachmentInfo = {
+    const fullAttachmentInfo: GPURenderBundleEncoderDescriptor = {
       // Defaults if not overridden:
-      colorFormats: ['rgba8unorm'],
+      colorFormats: ['rgba8unorm'] as GPUTextureFormat[],
```
kainino0x, line 329: Oops, this breakage is because the type changed from `Iterable<GPUTextureFormat>` to `Array<GPUTextureFormat>`. That should be fixed in the types.

kainino0x (approved):
LGTM but noticed an issue in the types that we may as well fix first

src/webgpu/api/validation/validation_test.ts
toji: Oops! Sorry! https://github.com/gpuweb/types/pull/79

kainino0x: No worries!

PR #704, opened 2021-08-09
Make reftest slightly fuzzy to account for apparent differences

kainino0x:
Failed here:
https://chromium-review.googlesource.com/c/chromium/src/+/3079142/3
https://ci.chromium.org/ui/p/chromium/builders/try/dawn-linux-x64-deps-rel/17573/overview

From manual inspection of the actual and expected pngs, there's a difference of up to 1 in each channel in the interpolated areas. Apparently there is a very slight difference in how they get interpolated, and that's fine.
Chromium's `image_diff` image seems to show diffs in the solid areas too, but I don't see it by manual inspection.

<hr>

### DIFF
--- a/src/webgpu/web_platform/reftests/README.txt
+++ b/src/webgpu/web_platform/reftests/README.txt
@@ -10,3 +10,6 @@ This tests things like:
 
 TODO: Test all possible output texture formats (currently only testing bgra8unorm).
 TODO: Test all possible ways to write into those formats (currently only testing B2T copy).
+
+TODO: Why is there sometimes a difference of 1 (e.g. 3f vs 40) in canvas_size_different_with_back_buffer_size?
+And why does chromium's image_diff show diffs on other pixels that don't seem to have diffs?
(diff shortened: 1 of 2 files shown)


kainino0x:
Landing TBR (to-be-reviewed) to get these tests running (hopefully)

kainino0x:
Come to think of it this is probably https://bugs.chromium.org/p/chromium/issues/detail?id=1083478#c16 .
I wonder if we could make those tests fuzzy as well and get them running.

(Not very important now though since this reftest is currently strictly better than that one!)

shaoboyan:
(Sry I miss this).
LGTM and I think it might be good to have 1 pixel tolerance for the reftest(Maybe all of the current ones) with TODOs. We'll finally get rid of it.

kainino0x:
OK, I just checked and unfortunately we would have to make the other tests a bit more fuzzy for them to work properly. I'm not exactly sure why they have a bigger diff. But it looks like they might just fit in under a threshold of maxDifference=8 or 10.

I think I'd prefer to leave them strict, and we'll try to fix our reftest infra first. We now have quite good coverage from `canvas_size_different_with_back_buffer_size.htm` so I'm not concerned about coverage.

PR #709, opened 2021-08-12
Implement some initial stress tests

krockot:
This adds a variety of tests from different categories. The limits in
place for most were chosen to keep running times generally under
~10-20 seconds -- and to avoid TDRs -- during local development,
but they may run (much) longer depending on environment.

<hr>

### DIFF
--- a/src/stress/compute/compute_pass.spec.ts
+++ b/src/stress/compute/compute_pass.spec.ts
@@ -3,6 +3,7 @@ Stress tests covering GPUComputePassEncoder usage.
 `;
 
 import { makeTestGroup } from '../../common/framework/test_group.js';
+import { iterRange } from '../../common/util/util.js';
 import { GPUTest } from '../../webgpu/gpu_test.js';
 
 export const g = makeTestGroup(GPUTest);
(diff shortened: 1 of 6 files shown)


src/stress/compute/compute_pass.spec.ts
```diff
@@ -12,26 +13,230 @@ g.test('many')
...
+        [[stage(compute), workgroup_size(1)]] fn main(
+            [[builtin(global_invocation_id)]] id: vec3<u32>) {
+          buffer.data[id.x] = buffer.data[id.x] + 2u;
+        }
+      `,
+    });
+    const kNumIterations = 250_000;
+    for (let i = 0; i < kNumIterations; ++i) {
+      const module = i % 2 === 0 ? module1 : module2;
+      const pipeline = t.device.createComputePipeline({ compute: { module, entryPoint: 'main' } });
```
austinEng, line 87: how different do you want the compute pipelines? Most implementations (I know at least Chrome / Firefox) are going to dedup these to point to exactly the same native object handle. If you really want them to be unique, change the shader contents.

maybe it can be a different test though - it would be a TON of pipelines

src/stress/compute/compute_pass.spec.ts
```diff
@@ -12,26 +13,230 @@ g.test('many')
...
+      buffer,
+      new Uint32Array([...iterRange(kNumElements, x => x + kNumIterations)])
+    );
+  });
 
 g.test('huge_dispatches')
   .desc(`Tests execution of compute passes with huge dispatch calls`)
-  .unimplemented();
+  .fn(async t => {
+    const kDimensions = [512, 512, 128];
```
austinEng, line 207: do we have constants somewhere where we can assert these are in the valid range?

austinEng (approved):

src/stress/compute/compute_pass.spec.ts
krockot: Ah thanks, that's very good to know. I will perturb the shaders.

src/stress/compute/compute_pass.spec.ts
krockot: Will check, will add if not

src/stress/compute/compute_pass.spec.ts
krockot: Settled for 10k pipelines, and I did end up differentiating them only by shader code. Pipeline creation itself ends up being the main bottleneck.

src/stress/compute/compute_pass.spec.ts
krockot: Added assertions against maxComputeWorkgroupsPerDimension.

PR #723, opened 2021-08-26
Introduce expectGPUBufferRepeatsSingleValue

krockot:
This helper offloads work onto the GPU if the buffer is large enough,
rather than the typical behavior of reading back values to validate on
the CPU. This makes it more suitable for checking the contents of large
buffers quickly.

This is then used by expectSingleColor to validate its buffer of
copied texture data. For an empirical comparison, verifying the contents
of a 4k texture could take ~7-10 seconds before this change, and with
this change it takes only a few milliseconds.

<hr>

### DIFF
--- a/src/webgpu/gpu_test.ts
+++ b/src/webgpu/gpu_test.ts
@@ -433,7 +433,7 @@ export class GPUTest extends Fixture {
       layout: pipeline.getBindGroupLayout(0),
       entries: [
         { binding: 0, resource: { buffer: expectedDataBuffer } },
-        { binding: 1, resource: { buffer: storageBuffer, } },
+        { binding: 1, resource: { buffer: storageBuffer } },
         { binding: 2, resource: { buffer: resultBuffer } },
       ],
     });


src/webgpu/gpu_test.ts
```diff
@@ -317,6 +317,140 @@ export class GPUTest extends Fixture {
...
+    // against them on the CPU.
+    const kMaxBufferSizeToCheckOnCpu = 256 * 1024;
+    const bufferSize = bytesPerRow * (numRows - 1) + minBytesPerRow;
+    if (bufferSize <= kMaxBufferSizeToCheckOnCpu) {
+      const valueBytes = Array.from(new Uint8Array(expectedValue));
+      const rowValues = new Array(minBytesPerRow / valueSize).fill(valueBytes);
+      const rowBytes = new Uint8Array([].concat(...rowValues));
+      const expectedContents = new Uint8Array(bufferSize);
+      range(numRows, row => expectedContents.set(rowBytes, row * bytesPerRow));
+      this.expectGPUBufferValuesEqual(buffer, expectedContents);
```
kainino0x, line 355: I didn't read this in detail, but if you can generate the data here, maybe you can generate it lazily, similar to:
https://github.com/gpuweb/cts/blob/44953ed6611238a7874ba2efc12ae0bb485850e4/src/webgpu/api/operation/compute/basic.spec.ts#L155

i.e. compare the elements against generated values instead of actually filling an ArrayBuffer to compare with?

src/webgpu/gpu_test.ts
```diff
@@ -317,6 +317,140 @@ export class GPUTest extends Fixture {
...
+      usage: GPUBufferUsage.STORAGE | GPUBufferUsage.COPY_SRC,
+    });
+    this.trackForCleanup(resultBuffer);
+
+    const readsPerRow = Math.ceil(minBytesPerRow / expectedDataSize);
+    const reducer = `
+    [[block]] struct Buffer { data: array<u32>; };
+    [[group(0), binding(0)]] var<storage, read> expected: Buffer;
+    [[group(0), binding(1)]] var<storage, read> in: Buffer;
+    [[group(0), binding(2)]] var<storage, read_write> out: Buffer;
```
kainino0x, line 407: nit: could storage mode just be `write`?

kainino0x (commented):
lgtm

PR #729, opened 2021-08-30
Add uniqueId() to class TestBuilder

sarahM0:
`.uniqueId(<id>)` is used to highlight the `id` assigned to a wgsl test plan. This`id` will be used for tracking changes of wgsl spec and keeping the test plan up to date.

A typical wgsl test plan looks like this:
```
g.test(<name>)
  .uniqueId(<id>)
  .desc(`<description>`)
  .params(u => u.combine('placeHolder1', ['placeHolder2', 'placeHolder3']))
  .unimplemented();
```
The test `description` is generated by paring the latest version of [wgsl specification](https://www.w3.org/TR/2021/WD-WGSL-20210826/). `id` is `sha1(<description>)[0:8]`. As a result if the next versions of spec changes the `description`, `id` becomes invalid.

`id` can also be used to check whether a test plan with a specific `description` is defined or implemented in cts. ie. `grep -rnw <id> /gpuweb/cts`.

<hr>

### DIFF
--- a/src/common/internal/test_group.ts
+++ b/src/common/internal/test_group.ts
@@ -120,7 +120,12 @@ export class TestGroup<F extends Fixture> implements TestGroupBuilder<F> {
 
 interface TestBuilderWithName<F extends Fixture> extends TestBuilderWithParams<F, {}> {
   desc(description: string): this;
-
+  /**
+   * A noop function with the purpose of highlighting value `id`.
+   *
+   * @param id a value uniquely assigned to this test.
+   */
+  uniqueId(id: number): this;
   /**
    * Parameterize the test, generating multiple cases, each possibly having subcases.
    *
@@ -182,6 +187,10 @@ class TestBuilder {
 
   desc(description: string): this {
     this.description = description.trim();
+    return this;
+  }
+
+  uniqueId(id: number): this {
     return this;
   }
 
--- a/src/webgpu/shader/execution/builtin/logical_built_in_functions.spec.ts
+++ b/src/webgpu/shader/execution/builtin/logical_built_in_functions.spec.ts
@@ -1,0 +1,55 @@
+export const description = `WGSL Logical built-in functions execution test`;
+
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { GPUTest } from '../../../gpu_test.js';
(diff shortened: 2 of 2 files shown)


src/webgpu/shader/execution/builtin/section16_01.spec.ts
```diff
@@ -0,0 +1,55 @@
+export const description = `WGSL section 16.1 execution test`;
```
kainino0x, line 1: WGSL section numbers probably aren't going to stay stable. For the same reason, URLs link to sections by name, e.g.:
https://gpuweb.github.io/gpuweb/wgsl/#logical-builtin-functions
That's probably necessary here too.

kainino0x (commented):
What's the goal of the hashes - to refer to the changes in bugs?

I wonder if instead you could refer to spec URLs, or something.
Unfortunately the WGSL spec doesn't currently have URLs for every single function definition. But eventually it probably should.

This change is OK from my perspective, but may need some discussion with other Tint folks on the logistics of how this will work in practice.

ben-clayton:
> What's the goal of the hashes - to refer to the changes in bugs?

Sarah has been working on tooling to extract out all the bits of the WGSL spec that must be tested. With this we hope to stub the bulk of the WGSL tests, and have engineers flesh these out by hand.
In order to associate the stubbed tests to the spec, we're attempting to hash the relevant bits of spec text.

Obviously text changes, and so the tool will attempt to collect up all the hashes already in the CTS and compare these to another parse of the spec. Mismatches are either a change to text, new rules or removed rules. In each case, we'll want to manually investigate what's changed and make the appropriate changes to the CTS tests.

That's the theory, anyway. If it doesn't work out, we can grep for this function and strip it.

kainino0x (approved):
I see, it detects changes in the spec. That sounds solid to me!

Am I understanding correctly that the test description is the same string used to hash? If so would it make sense to associate the tests by that string directly rather than having the extra hash to maintain? Doesn't have to be figured out now, but perhaps something to think about if the uniqueId turns out to be extra work.

LGTM except I still have the concern about the spec section numbers changing. Does it make sense to id them by the `#` URL in the spec?

PR #730, opened 2021-08-31
wgsl abs() built-in function execution test

sarahM0:
<hr>

**Author checklist for test code/plans:**

- [ ] All outstanding work is tracked with "TODO" in a test/file description or `.unimplemented()` on a test.
- [ ] New helpers, if any, are documented using `/** doc comments */` and can be found via `helper_index.txt`.
- [ ] (Optional, sometimes not possible.) Tests pass (or partially pass without unexpected issues) in an implementation. (Add any extra details above.)

**[Reviewer sign-off](https://github.com/gpuweb/cts/blob/main/docs/reviews.md) for test code/plans:** (Note: feel free to pull in other reviewers at any time for any reason.)

- [x] The test path is reasonable, the [description](https://github.com/gpuweb/cts/blob/main/docs/intro/plans.md) "describes the test, succinctly, but in enough detail that a reader can read only the test plans in a file or directory and evaluate the completeness of the test coverage."
- [x] Tests appear to cover this area completely, except for outstanding TODOs. Validation tests use control cases.
    (This is critical for coverage. Assume anything without a TODO will be forgotten about forever.)
- [x] Existing (or new) test helpers are used where they would reduce complexity.
- [x] TypeScript code is readable and understandable (is unobtrusive, has reasonable type-safety/verbosity/dynamicity).

PR #731, opened 2021-09-01
wgsl type parameter generator

sarahM0:
This PR uses the idea of a wgsl type iterator generator form [this indexable type generator](https://github.com/gpuweb/cts/blob/main/src/webgpu/shader/execution/robust_access.spec.ts#L118). I implements a similar one that can generates types based on the following, and asserts if the type is invalid (eg. 'bool' with storage storage class or as an atomic base type).
- type container (scalar, vector, matrix, array)
- the base type (bool, i32, u32, f32)
- whether its atomic

The motivation is having a type generator that other test can reuse. (eg. testing builtin functions such as [this](https://github.com/gpuweb/cts/pull/730/files#diff-d403b8a26ac9412452ae3eb1640fa5eb01ec4e483245553bb3cce05d206e0747R72) one that working on).

Here are two examples:

```
g.test('example1')
.desc(`generates these types: f32, vecN<f32>, matMxN<f32>`)
.params(u =>
  u
  .combineWithParams([{ storageClass: 'storage' }] as const)
    .combineWithParams([
      { containerType: 'scalar' },
      { containerType: 'vector' },
      { containerType: 'matrix' },
    ] as const)
    .combineWithParams([{ supportsAtomics: false }] as const)
    .combineWithParams([{ baseType: 'f32' }] as const)
    .beginSubcases()
    .expandWithParams(generateTypes)
).unimplemented();

g.test('example2')
.desc(`generates these types: atomic<u32>, array<atomic<u32>, 3>, array<atomic<32>>`)
.params(u =>
  u
  .combineWithParams([{ storageClass: 'storage' }] as const)
    .combineWithParams([
      { containerType: 'scalar' },
      { containerType: 'array' },
    ] as const)
    .combineWithParams([{ supportsAtomics: true }] as const)
    .combineWithParams([{ baseType: 'u32' }] as const)
    .beginSubcases()
    .expandWithParams(generateTypes)
).unimplemented();

```

<hr>

### DIFF
--- a/src/webgpu/shader/types.ts
+++ b/src/webgpu/shader/types.ts
@@ -50,7 +50,7 @@ export function* generateTypes({
   storageClass,
   baseType,
   containerType,
-  supportsAtomics = false,
+  isAtomic = false,
 }: {
   storageClass: string;
   /** Base scalar type (i32/u32/f32/bool). */
@@ -58,13 +58,13 @@ export function* generateTypes({
   /** Container type (scalar/vector/matrix/array) */
   containerType: ContainerType;
   /** Whether to wrap the baseType in `atomic<>`. */
-  supportsAtomics?: boolean;
+  isAtomic?: boolean;
 }) {
   const scalarInfo = kScalarTypeInfo[baseType];
-  if (supportsAtomics) {
+  if (isAtomic) {
     assert(scalarInfo.supportsAtomics, 'type does not support atomics');
   }
-  const scalarType = supportsAtomics ? `atomic<${baseType}>` : baseType;
+  const scalarType = isAtomic ? `atomic<${baseType}>` : baseType;
 
   // Storage and uniform require host-sharable types.
   if (storageClass === 'storage' || storageClass === 'uniform') {
@@ -171,12 +171,12 @@ export function supportsAtomics(p: {
 }
 
 /** Generates an iterator of supported base types (i32/u32/f32/bool) */
-export function* supportedScalarTypes(p: { supportsAtomics: boolean; storageClass: string }) {
+export function* supportedScalarTypes(p: { isAtomic: boolean; storageClass: string }) {
   for (const scalarType of kScalarTypes) {
     const info = kScalarTypeInfo[scalarType];
 
     // Test atomics only on supported scalar types.
-    if (p.supportsAtomics && !info.supportsAtomics) continue;
+    if (p.isAtomic && !info.supportsAtomics) continue;
 
     // Storage and uniform require host-sharable types.
     const isHostShared = p.storageClass === 'storage' || p.storageClass === 'uniform';
--- a/src/webgpu/shader/types.ts
+++ b/src/webgpu/shader/types.ts
@@ -1,7 +1,13 @@
 import { keysOf } from '../../common/util/data_tables.js';
+import { assert } from '../../common/util/util.js';
+import { align } from '../util/math.js';
 
-/** WGSL plain scalar type names. */
+const kArrayLength = 3;
+
+export type ContainerType = 'scalar' | 'vector' | 'matrix' | 'atomic' | 'array';
 export type ScalarType = 'i32' | 'u32' | 'f32' | 'bool';
+
+export const HostSharableTypes = ['i32', 'u32', 'f32'] as const;
 
 /** Info for each plain scalar type. */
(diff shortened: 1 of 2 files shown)


src/webgpu/shader/types.ts
```diff
@@ -36,3 +42,146 @@ export const kMatrixContainerTypeInfo = /* prettier-ignore */ {
...
+  containerType,
+  supportsAtomics = false,
+}: {
+  storageClass: string;
+  /** Base scalar type (i32/u32/f32/bool). */
+  baseType: ScalarType;
+  /** Container type (scalar/vector/matrix/array) */
+  containerType: ContainerType;
+  /** Whether to wrap the baseType in `atomic<>`. */
+  supportsAtomics?: boolean;
```
austinEng, line 61: nit: maybe just "atomic". We're indicating whether we want it to be atomic, not whether it is atomic

src/webgpu/shader/types.ts
```diff
@@ -36,3 +42,146 @@ export const kMatrixContainerTypeInfo = /* prettier-ignore */ {
...
+
+  // Storage and uniform require host-sharable types.
+  if (storageClass === 'storage' || storageClass === 'uniform') {
+    assert(isHostSharable(baseType), 'type ' + baseType.toString() + ' is not host sharable');
+  }
+
+  // Scalar types
+  if (containerType === 'scalar') {
+    yield {
+      type: `${scalarType}`,
```
austinEng, line 77: maybe "type" can be an input parameter to this function. I can see lots of tests wanting to use "type" to mean something else

src/webgpu/shader/types.ts
```diff
@@ -36,3 +42,146 @@ export const kMatrixContainerTypeInfo = /* prettier-ignore */ {
...
+      _kTypeInfo: {
+        elementBaseType: `${scalarType}`,
```
austinEng, line 79: what if elementBaseType were a full _kTypeInfo as well? could be useful if/when we have nested containers

austinEng (commented):
Really cool! I also found myself needing something similar as well. A couple of small comments, but I think Kai should review as well.

src/webgpu/shader/types.ts
sarahM0: I think "type" is the same as baseType here. Or maybe I don't know what you mean by "type".
Could you give an example? Or refer me to one the tests using "type".

austinEng: Sorry, I wasn't very clear. Also though, this is a small usability nit. I think it's OK if we leave it as-is for now.

At other times, "type" is:
"type: `array<${scalarType},${kArrayLength}>`"

If I understand correctly, `generateTypes` returns:
```
Generator<{
  type: string,
  _kTypeInfo: {
    elementBaseType: string,
    // other type info..
  }
}>
```

It could cause a conflict if the place where the helper is used wants to use "type" for something else:

```
const enum OtherType {
  Small, Medium, Big
}

.params(u => 
  u
    .combine('type', [OtherType.Small, OtherType.Medium, OtherType.Big])
    .beginSubcases()
    .expandWithParams(generateTypes)
)
```

This would be a problem because there are two params with the same name "type".

We could *maybe* make it look like:
```
  u
    .combine('type', [OtherType.Small, OtherType.Medium, OtherType.Big])
    .beginSubcases()
    .expandWithParams(generateTypes.bind('wgslType'))
    .expandWithParams(generateTypes.bind('otherWgslType'))
)
```
so it names the param "wgslType" instead.

src/webgpu/shader/types.ts
sarahM0: Yes, that makes sense. How about isAtomic?

src/webgpu/shader/types.ts
sarahM0: Let's leave this as-is. I'll fix it in my next CL.

src/webgpu/shader/types.ts
sarahM0: Let's leave this as-is. I'll add this utility in my next CL.
Although I think this would only benefit arrays. Since other baseTypes has to be a ScalarType. ie. vec4x4<vec2x2<f32>> is invalid.

austinEng (approved):

PR #733, opened 2021-09-02
Add an option to automatically batch large lists of subcases

austinEng:
This adds .batch(n) which adds a batch__ case param and splits
the subcases so that there are at most |n| subcases in a case.
This is helpful when a test will not realistically fit within
the test timeout for a single case.

<hr>


kainino0x:
This seems OK. Previously I had strayed away from batches because I didn't want there to be too much instability in which test ran in which place. But since this can only be used for subcases, it's less of a problem.

There's been at least one test before where we had to do a loop inside the test itself, because doing the full test setup for every subcase was too expensive - the inner loop was way, way faster. And there's also a recent test which does many shader loads inside one thread which is of course many orders of magnitude faster than individual subcases. Anyway, I don't know if an inner loop would be helpful here, but of note, it should be possible to loop over a CaseParamsBuilder using a for-loop if that's helpful.

austinEng:
> Anyway, I don't know if an inner loop would be helpful here

probably not - compiling distinct pipelines

> There's been at least one test before where we had to do a loop inside the test itself, because doing the full test setup for every subcase was too expensive - the inner loop was way, way faster. 

maybe we can have a separate thing which optionally allows memoizing some data (with a list of dependencies, ex. the device) and passing it into the parameterizations. 
Opt-in because it could cause instability.

kainino0x:
> maybe we can have a separate thing which optionally allows memoizing some data (with a list of dependencies, ex. the device) and passing it into the parameterizations.
> Opt-in because it could cause instability.

This seems like a good idea. We have some places where caching pipelines/etc would be helpful, though not sure whether it would have helped with the specific cases I'm thinking of.

We do have some places where we cache computed stuff, but not just pure-computations, not webgpu objects.

austinEng:
although with pipelines, it should probably hit the cache in the implementation and not be so much of a problem? Could be useful for other resources though like buffers - especially if they're used as read-only. Less risk of unstable test behavior.

kainino0x (approved):
True, we'd have to look at what cases we actually might need a caching mechanism for.

Anyway, this should lgtm after rebase

PR #734, opened 2021-09-03
Fix tools/deno

lucacasonato:
This commit moves runtime specific things functions reading files into a
common/runtime/helpers/sys.ts file.


kvark (approved):

tools/run_deno
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env -S deno run --unstable --allow-read --allow-write --allow-env --allow-net=deno.land --no-check
+
+import '../out/common/runtime/cmdline.js';
```
kainino0x, line 3:
```suggestion
import '../out/common/runtime/cmdline.js';

```

kainino0x (approved):

PR #742, opened 2021-09-09
Update destinationOffset alignment to 256 in resolveQuerySet tests

haoxli:
We have updated the destinationOffset alignment of resolveQuerySet
to 256 in SPEC. These tests need to be updated to align with the
change.

Also update the existing cases to use helper functions.

<hr>

### DIFF
--- a/src/webgpu/api/validation/encoding/queries/resolveQuerySet.spec.ts
+++ b/src/webgpu/api/validation/encoding/queries/resolveQuerySet.spec.ts
@@ -32,12 +32,9 @@ Tests that resolve query set with invalid object.
       usage: GPUBufferUsage.QUERY_RESOLVE,
     });
 
-    const encoder = t.device.createCommandEncoder();
-    encoder.resolveQuerySet(querySet, 0, 1, destination, 0);
-
-    t.expectValidationError(() => {
-      encoder.finish();
-    }, querySetState === 'invalid' || destinationState === 'invalid');
+    const encoder = t.createEncoder('non-pass');
+    encoder.encoder.resolveQuerySet(querySet, 0, 1, destination, 0);
+    encoder.validateFinish(querySetState === 'valid' && destinationState === 'valid');
   });
 
 g.test('resolveQuerySet,first_query_and_query_count')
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/encoding/queries/resolveQuerySet.spec.ts
```diff
@@ -91,37 +85,35 @@ Tests that resolve query set with invalid destinationBuffer:
...
     `
 Tests that resolve query set with invalid destinationOffset:
-- destinationOffset is not a multiple of 8
+- destinationOffset is not a multiple of 256
 - The size of destinationBuffer - destinationOffset < queryCount * 8
 - destinationOffset out of range
   `
   )
-  .paramsSubcasesOnly(u => u.combine('destinationOffset', [0, 6, 8, 16]))
+  .paramsSubcasesOnly(u => u.combine('destinationOffset', [0, 8, 256, 512, 768]))
```
kainino0x, line 102: nit: test a "minimally unaligned" value, i.e. a multiple of the largest power of 2 less than 256:
```suggestion
  .paramsSubcasesOnly(u => u.combine('destinationOffset', [0, 128, 256, 384, 512, 768]))
```

src/webgpu/api/validation/encoding/queries/resolveQuerySet.spec.ts
```diff
@@ -91,37 +85,35 @@ Tests that resolve query set with invalid destinationBuffer:
...
-    const encoder = t.device.createCommandEncoder();
-    encoder.resolveQuerySet(querySet, 0, kQueryCount, destination, t.params.destinationOffset);
-
-    t.expectValidationError(() => {
-      encoder.finish();
-    }, t.params.destinationOffset > 0);
+    const encoder = t.createEncoder('non-pass');
+    encoder.encoder.resolveQuerySet(querySet, 0, kQueryCount, destination, destinationOffset);
+    encoder.validateFinish(
+      destinationOffset % 256 === 0 && destinationOffset + kQueryCount * 8 <= size
```
kainino0x, line 115: This is now testing both the destination offset alignment and that the resolveQuerySet is in bounds of the destination buffer. This is not terrible, but it's hard to be sure that all cases are covered (because one validation rule could hide the an error in the other).

If we don't already have separate tests for these two rules, I think we should.

haoxli: Split offset alignment and oob into two tests.

kainino0x (approved):
LGTM!

PR #751, opened 2021-09-17
ShaderValidationTest: remove string overload

dneto0:
There are no standard error codes in the WGSL spec.
So remove support for checking them.
See https://github.com/gpuweb/gpuweb/issues/1813

<hr>

### DIFF
--- a/src/webgpu/shader/validation/shader_validation_test.ts
+++ b/src/webgpu/shader/validation/shader_validation_test.ts
@@ -12,10 +12,8 @@ export class ShaderValidationTest extends GPUTest {
    * ```ts
    * t.expectCompileResult(true, `wgsl code`); // Expect success
    * t.expectCompileResult(false, `wgsl code`); // Expect validation error with any error string
-   * t.expectCompileResult('substr', `wgsl code`); // Expect validation error containing 'substr'
    * ```
    *
-   * TODO(gpuweb/gpuweb#1813): Remove the "string" overload if there are no standard error codes.
    */
   expectCompileResult(expectedResult: boolean | string, code: string) {
     let shaderModule: GPUShaderModule;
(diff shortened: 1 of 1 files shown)


src/webgpu/shader/validation/shader_validation_test.ts
```diff
@@ -12,10 +12,8 @@ export class ShaderValidationTest extends GPUTest {
    * ```ts
    * t.expectCompileResult(true, `wgsl code`); // Expect success
    * t.expectCompileResult(false, `wgsl code`); // Expect validation error with any error string
-   * t.expectCompileResult('substr', `wgsl code`); // Expect validation error containing 'substr'
    * ```
-   *
-   * TODO(gpuweb/gpuweb#1813): Remove the "string" overload if there are no standard error codes.
+   * foobar
    */
   expectCompileResult(expectedResult: boolean | string, code: string) {
```
kainino0x, line 18: Type should change too:

```suggestion
  expectCompileResult(expectedResult: boolean, code: string) {
```

(This may result in type check errors - if any call sites are passing strings)

kainino0x: Committed this suggestion, will see if it passes checks.

kainino0x (approved):
Looks all OK!

src/webgpu/shader/validation/shader_validation_test.ts
```diff
@@ -12,12 +12,10 @@ export class ShaderValidationTest extends GPUTest {
    * ```ts
    * t.expectCompileResult(true, `wgsl code`); // Expect success
    * t.expectCompileResult(false, `wgsl code`); // Expect validation error with any error string
-   * t.expectCompileResult('substr', `wgsl code`); // Expect validation error containing 'substr'
    * ```
-   *
-   * TODO(gpuweb/gpuweb#1813): Remove the "string" overload if there are no standard error codes.
+   * foobar
```
kainino0x, line 16: er, missed this the first time
```suggestion
```

src/webgpu/shader/validation/shader_validation_test.ts
```diff
@@ -12,12 +12,10 @@ export class ShaderValidationTest extends GPUTest {
...
    */
-  expectCompileResult(expectedResult: boolean | string, code: string) {
+  expectCompileResult(expectedResult: boolean, code: string) {
```
sarahM0, line 18: I would suggest to  keep this as the string expectedResult can handy when want to check a test is failing with an expected error message (instead of a specific error code)?

Unless failures for wrong reasons is not a real concern or expecting a specific substring in the error message introduces the same problems as we had with error codes.

kainino0x: I think the WGSL spec doesn't specify particular strings that must appear in messages for particular errors, so we can't test them. Unless we added some way for each browser to customize the strings they expect for every test.

Without a way to test which error occurred, we have to instead test that a new error arises from a minimal code change between a control case and a test case.

dj2:
@dneto0 can we pick this up again and land? Seems like the right thing to do given there are no error strings in the spec.

dj2:
Closing. Took over in #1184

PR #753, opened 2021-09-22
Depth clip/clamp test

kainino0x:
See https://github.com/gpuweb/gpuweb/issues/2100

<hr>

### DIFF
--- a/src/webgpu/api/operation/rendering/depth_clip_clamp.spec.ts
+++ b/src/webgpu/api/operation/rendering/depth_clip_clamp.spec.ts
@@ -1,0 +1,510 @@
+export const description = `
+Tests for depth clipping, depth clamping (at various points in the pipeline), and maybe extended
+depth ranges as well.
+
+TODO: Based on documentation and experimental results, depth should actually always be clamped.
+The depth-clamping feature here is actually used to toggle depth _clipping_. These tests need to be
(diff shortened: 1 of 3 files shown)


kainino0x:
PTAL! @kvark @Jiawei-Shao since you've been following https://github.com/gpuweb/gpuweb/issues/2100, perhaps you could both review these (rather nitpicky) tests? Please see if there's aspect I've missed testing.

@Jiawei-Shao, perhaps additionally you could verify these tests against an implementation of the depth-clamping feature on D3D12, since I think you were working on that?

These tests are passing in Chromium (except for depth16unorm which we don't implement yet) on Metal (with both depthClamp values) and D3D12 (with depthClamp:false). They pass on Vulkan with clampDepth:true but fail with clampDepth:false, because for consistency with D3D12/Metal, Tint will need to inject a clamp to the viewport bounds at the end of fragment stages that write `frag_depth`.
I haven't tried them against Firefox because I don't know what the implementation status is.

These tests will need updating later because I think we need to change the spec from having a depthClamp toggle to having a depthClip toggle.

kainino0x:
Also I tried to write good comments, but please let me know if any of it can be clarified further.

Jiawei-Shao (approved):

kvark (approved):

kainino0x:
Thanks!

PR #755, opened 2021-09-24
draft for wgsl validation

munrocket:
Seems that just sharing /wgsl/ folder not enough, `fetch` works only in browsers and not defined in node.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -84,6 +84,24 @@ module.exports = function (grunt) {
           '--copy-files'
         ],
       },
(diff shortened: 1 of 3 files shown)


munrocket:
node-fetch library not fit too current build workflow and require new configs for esm support.

dj2:
Is this PR still valid? Or can it be closed?

dj2:
These `.wgsl` files have been removed and will get re-created as spec tests. I'm going to close this for now, please re-open if you think it's still needed.

PR #758, opened 2021-09-27
Do not require grunt to be installed globally.

greggman:
Deleted the checklist as none of it applied.


kainino0x (approved):
Whoops, I thought we had grunt-cli in here already! Thanks!

PR #760, opened 2021-09-28
cmdline: Add new flag to allow overriding of the GPU provider

ben-clayton:
Allows non-web implementations to provide a custom Node module that implements the GPU interface.

For additional context, see:
https://dawn-review.googlesource.com/c/dawn/+/64940/19/src/dawn_node/Module.cpp#38

### DIFF
--- a/src/common/runtime/cmdline.ts
+++ b/src/common/runtime/cmdline.ts
@@ -1,12 +1,12 @@
 /* eslint no-console: "off" */
 
-import { setGPUProvider } from '../../webgpu/util/navigator_gpu.js';
 import { DefaultTestFileLoader } from '../internal/file_loader.js';
 import { prettyPrintLog } from '../internal/logging/log_message.js';
 import { Logger } from '../internal/logging/logger.js';
 import { LiveTestCaseResult } from '../internal/logging/result.js';
 import { parseQuery } from '../internal/query/parseQuery.js';
 import { parseExpectationsForTestQuery } from '../internal/query/query.js';
+import { setGPUProvider } from '../util/navigator_gpu.js';
 import { assert, unreachable } from '../util/util.js';
 
 import sys from './helper/sys.js';
--- a/src/common/runtime/cmdline.ts
+++ b/src/common/runtime/cmdline.ts
@@ -1,5 +1,6 @@
 /* eslint no-console: "off" */
 
+import { setGPUProvider } from '../../webgpu/util/navigator_gpu.js';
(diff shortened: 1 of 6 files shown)


src/common/runtime/cmdline.ts
```diff
@@ -49,6 +51,11 @@ for (let i = 0; i < sys.args.length; ++i) {
     } else if (a === '--expectations') {
       const expectationsFile = new URL(sys.args[++i], `file://${sys.cwd()}`).pathname;
       loadWebGPUExpectations = import(expectationsFile).then(m => m.expectations);
+    } else if (a === '--gpu-provider') {
+      const modulePath = sys.args[++i];
+      setGPUProvider(() => {
+        return require(modulePath).gpu;
+      });
```
austinEng, line 58: nit: could be more concise with:
```suggestion
      setGPUProvider(() => require(modulePath).gpu);
```

austinEng (approved):
lgtm

ben-clayton:
```
/home/runner/work/cts/cts/src/common/runtime/cmdline.ts
Error:   3:32  error  Unexpected path "../../webgpu/util/navigator_gpu.js" imported in restricted zone. Non common/ code imported from common/  import/no-restricted-paths
```

So that's not going to work then. @austinEng , @kainino0x - any suggestions?

src/common/runtime/cmdline.ts
ben-clayton: Done

austinEng:
hmmm.. idk as I don't have context on why these import rules exist. like for organization? or because code breaks if we don't obey them?

- could we just move navigator_util.ts to common/ or common/framework/ ?
- or set the module path as some env var, and then in `navigator_util.ts` do `require(process.env.WEBGPU_MODULE_PATH)` ?
- if all else fails, there's always `//eslint-disable-next-line import/no-restricted-paths` but that's probably the wrong option

kainino0x:
They're mainly for code organization. I think how this file used to work is it would just take a different path by detecting whether it was running under Node. Moving it somewhere else probably makes sense, I'll checkout the change and think about it.

ben-clayton:
Thanks Kai!

kainino0x:
I think moving to src/common/util makes sense. Uploaded a commit to do that (just did a rename in vscode), and also added one assert.

kainino0x (approved):
Rest of the changes LGTM

ben-clayton:
Perfect. Thank you both!

PR #762, opened 2021-09-29
webgpu/shader/execution: Fix bad WGSL generation

ben-clayton:
Subcases of `webgpu:shader,execution,robust_access:linear_memory,*` tests were attempting to call `any()` on boolean scalars. No such overload exists.

Currently for these tests, only matrices, when indexed, produce a vector. So instead of checking fro `innerLength`, check the outer container type is a matrix.

Fixes 6 test cases.

### DIFF
--- a/src/webgpu/shader/execution/robust_access.spec.ts
+++ b/src/webgpu/shader/execution/robust_access.spec.ts
@@ -146,7 +146,16 @@ g.test('linear_memory')
       .expandWithParams(generateTypes)
   )
   .fn(async t => {
-    const { storageClass, storageMode, access, isAtomic, baseType, type, _kTypeInfo } = t.params;
+    const {
+      storageClass,
+      storageMode,
+      access,
+      isAtomic,
+      containerType,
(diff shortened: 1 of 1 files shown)


ben-clayton:
Note: Ideally, the `_kTypeInfo` would actually provide another TypeInfo for the element type, which could describe the kind of the element. I think this would simplify a lot of this kind of logic.

Kangz (approved):

sarahM0:
> Note: Ideally, the `_kTypeInfo` would actually provide another TypeInfo for the element type, which could describe the kind of the element. I think this would simplify a lot of this kind of logic.

@austinEng suggested a similar change in [#731-disscusion](https://github.com/gpuweb/cts/pull/731#discussion_r701231269). I should have added a todo for it. I created issue [#763](https://github.com/gpuweb/cts/issues/763) to fix this.

sarahM0 (approved):

kainino0x (approved):

PR #766, opened 2021-09-30
Add test for compute shader input builtin values

jrprice:
Launch a compute shader with a variety of workgroup and dispatch sizes, and test that the values received for each input builtin are correct for every invocation in the grid.

Covers parameters, structures, and a combination of both.

Tested on macOS with ToT Chromium.

I'm new to TypeScript, so apologies for anything that is non-idiomatic (and all feedback welcome of course).

<hr>

- [X] New helpers, if any, are documented using `/** doc comments */` and can be found via `helper_index.txt`.

### DIFF
--- a/src/webgpu/shader/execution/shader_io/compute_builtins.spec.ts
+++ b/src/webgpu/shader/execution/shader_io/compute_builtins.spec.ts
@@ -1,0 +1,278 @@
+export const description = `Test compute shader builtin variables`;
+
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { iterRange } from '../../../../common/util/util.js';
+import { GPUTest } from '../../../gpu_test.js';
+
+export const g = makeTestGroup(GPUTest);
+
+// Test that the values for each input builtin are correct.
+g.test('inputs')
+  .desc(`Test compute shader builtin inputs values`)
+  .params(u =>
+    u
+      .combine('method', ['param', 'struct', 'mixed'] as const)
+      .combineWithParams([
+        {
+          groupSize: { x: 1, y: 1, z: 1 }, numGroups: { x: 1, y: 1, z: 1 },
+        },
+        {
+          groupSize: { x: 8, y: 4, z: 2 }, numGroups: { x: 1, y: 1, z: 1 },
+        },
+        {
+          groupSize: { x: 1, y: 1, z: 1 }, numGroups: { x: 8, y: 4, z: 2 },
+        },
+        {
+          groupSize: { x: 3, y: 7, z: 5 }, numGroups: { x: 13, y: 9, z: 11 },
+        },
+      ] as const)
+      .beginSubcases()
+  )
+  .fn(async t => {
+    const invocationsPerGroup =
+      t.params.groupSize.x *
+      t.params.groupSize.y *
+      t.params.groupSize.z;
+    const totalInvocations =
+      invocationsPerGroup *
+      t.params.numGroups.x *
+      t.params.numGroups.y *
+      t.params.numGroups.z;
+
+    // Generate the structures, parameters, and builtin expressions used in the shader.
+    let params = ''
+    let structures = ''
+    let local_id = ''
+    let local_index = ''
+    let global_id = ''
+    let group_id = ''
+    let num_groups = ''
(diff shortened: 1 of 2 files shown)


kainino0x:
To fix the formatting issues you can use `npm run fix` locally.

kainino0x:
The formatter is fairly strict (unfortunately), but you can influence a few things, in particular you can have an object either on one line, or with one member per line.

src/webgpu/shader/execution/shader_io/compute_builtins.spec.ts
```diff
@@ -0,0 +1,278 @@
...
+  ) {
+  let group_index = ((${group_id}.z * ${num_groups}.y) + ${group_id}.y) * ${num_groups}.x + ${group_id}.x;
+  let global_index = group_index * ${invocationsPerGroup}u + ${local_index};
+  local_id_out.data[global_index] = ${local_id};
+  local_index_out.data[global_index] = ${local_index};
+  global_id_out.data[global_index] = ${global_id};
+  group_id_out.data[global_index] = ${group_id};
+  num_groups_out.data[global_index] = ${num_groups};
+}
+`;
```
kainino0x, line 132: nit: We usually indent wgsl strings to match the surrounding code for readability (it'll end up with unnecessary spaces inside the string but it doesn't really matter)

jrprice:
> To fix the formatting issues you can use `npm run fix` locally.

Done, thanks!

src/webgpu/shader/execution/shader_io/compute_builtins.spec.ts
```diff
@@ -0,0 +1,286 @@
...
+    // Check [[builtin(global_invocation_id)]] values.
+    const getGlobalId = (groupId: vec3, localId: vec3) => {
+      return {
+        x: groupId.x * t.params.groupSize.x + localId.x,
+        y: groupId.y * t.params.groupSize.y + localId.y,
+        z: groupId.z * t.params.groupSize.z + localId.z,
+      };
+    };
+    const checkGlobalId = (outputData: Uint32Array) => {
+      return checkEachIndex(outputData, 'global_invocation_id', getGlobalId);
```
kainino0x: These can be made slightly less unwieldy with type inference (removing `: Uint32Array`) and lambda-style functions (removing `return`). Pushed a commit for this since it's hard to explain.

src/webgpu/shader/execution/shader_io/compute_builtins.spec.ts
```diff
@@ -0,0 +1,286 @@
...
+    type vec3 = { x: number; y: number; z: number };
+
+    // Helper to check that the vec3<u32> value at each index of the provided `output` buffer
+    // matches the expected value for that invocation, as generated by the `getBuiltinValue`
+    // function. The `name` parameter is the builtin name, used for error messages.
+    const checkEachIndex = (
+      output: Uint32Array,
+      name: string,
+      getBuiltinValue: (groupId: vec3, localId: vec3) => vec3
+    ) => {
```
kainino0x, line 192: This looks good, though I wonder if it could be done with the `checkElementsEqualGenerated` helper in `check_contents.ts`?

If it's not straightforward don't worry about it. It may require recomputing gx/gy/gz/lx/ly/lz from the buffer index which would probably be a pain.

kainino0x (approved):

src/webgpu/shader/execution/shader_io/compute_builtins.spec.ts
jrprice: Done.

src/webgpu/shader/execution/shader_io/compute_builtins.spec.ts
jrprice: Thanks, looks much better now.

src/webgpu/shader/execution/shader_io/compute_builtins.spec.ts
jrprice: Yeah I initially tried this, but recomputing the group/local indices from the buffer index was awkward, and also I'm not sure how to handle the padding bytes since these buffers are arrays of `vec3` with a 16 byte stride.

PR #767, opened 2021-09-30
webgpu/api/validation: Fix vertex_shader_input_location_limit test

ben-clayton:
The WGSL spec states that the `[[location(n)]]` decoration must be a "non-negative i32 literal".

### DIFF
--- a/src/webgpu/api/validation/vertex_state.spec.ts
+++ b/src/webgpu/api/validation/vertex_state.spec.ts
@@ -83,11 +83,10 @@ class F extends ValidationTest {
   }
 
   testVertexState(
-    success: boolean,
+    success: { shader: boolean; pipeline: boolean },
     buffers: Iterable<GPUVertexBufferLayout>,
     vertexShader: string = VERTEX_SHADER_CODE_WITH_NO_INPUT
   ) {
-    const vsModule = this.device.createShaderModule({ code: vertexShader });
     const fsModule = this.device.createShaderModule({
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/vertex_state.spec.ts
```diff
@@ -396,7 +413,13 @@ g.test('vertex_shader_input_location_limit')
       },
     ];
 
-    const success = testLocation < kMaxVertexAttributes;
+    const success = {
+      // [[location(n)]] decoration must be a "non-negative i32 literal"
+      // TODO: Dawn raises a "Attribute location (N) over limits" error at
+      // shader creation time for N>15. Is this right?
```
ben-clayton, line 419: @Kangz - It seems to me that Dawn might be validating too early here. Please can you advise?

kainino0x: https://github.com/gpuweb/gpuweb/pull/1423

kainino0x: ... restricted the locations themselves to be under the limit after some discussion that that was most likely sufficient, and forward-compatible with relaxing the restriction.

ben-clayton: Thanks. Made both the `shader` and `pipeline` conditions both dependent on the location being [0..kMaxVertexAttributes)

PR #775, opened 2021-10-13
wgsl: Add built-in variable validation tests

jrprice:
Test that the use of a built-in variable is only accepted for the
correct combination of stage, in|out usage, and type.

Tested and all passing with top-of-tree Chromium on macOS.

<hr>

- [X] New helpers, if any, are documented using `/** doc comments */` and can be found via `helper_index.txt`.

### DIFF
--- a/src/webgpu/shader/validation/shader_io/builtins.spec.ts
+++ b/src/webgpu/shader/validation/shader_io/builtins.spec.ts
@@ -1,0 +1,135 @@
+export const description = `Validation tests for entry point built-in variables`;
+
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { ShaderValidationTest } from '../shader_validation_test.js';
+
+import { generateShader } from './util.js';
+
+export const g = makeTestGroup(ShaderValidationTest);
+
+// List of all built-in variables and their stage, in|out usage, and type.
+// Taken from table in Section 15:
+// https://www.w3.org/TR/2021/WD-WGSL-20211013/#builtin-variables
+const kBuiltins = [
+  { name: 'vertex_index', stage: 'vertex', io: 'in', type: 'u32' },
+  { name: 'instance_index', stage: 'vertex', io: 'in', type: 'u32' },
+  { name: 'position', stage: 'vertex', io: 'out', type: 'vec4<f32>' },
+  { name: 'position', stage: 'fragment', io: 'in', type: 'vec4<f32>' },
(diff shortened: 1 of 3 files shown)


src/webgpu/shader/validation/shader_io/util.ts
```diff
@@ -0,0 +1,78 @@
+/**
+ * Generate an entry point that uses an entry point IO variable.
+ *
+ * @param attribute The entry point IO attribute.
+ * @param type The type to use for the entry point IO variable.
+ * @param stage The shader stage.
+ * @param io An "in|out" string specifying whether the entry point IO is an input or an output.
+ * @param use_struct True to wrap the entry point IO in a struct.
```
kainino0x, line 8: nit: These param docs may need to move to the members of the args object type below in order to get associated properly. Or maybe there's a tsdoc syntax for this, but iirc there may not be.

src/webgpu/shader/validation/shader_io/builtins.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+  'atomic<i32>',
+  'array<bool,4>',
+  'array<u32,4>',
+  'array<i32,4>',
+  'array<f32,4>',
+  'MyStruct',
+] as const;
+
+g.test('stage_inout')
+  .desc(`Test validation of built-in variable stage and in/out usage`)
```
kainino0x, line 66: Go into slightly more detail here so it's possible to understand the test without reading it, e.g. "Test different [[builtin]] declarations are validated to match the stage and in/out usage of the specified builtin."

src/webgpu/shader/validation/shader_io/builtins.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+        x.name === t.params.name &&
+        x.stage === t.params.target_stage &&
+        x.io === t.params.target_io &&
+        x.type === t.params.type
+    );
+    t.expectCompileResult(expectation, code);
+  });
+
+g.test('type')
+  .desc(`Test validation of built-in variable types`)
```
kainino0x, line 96: same

kainino0x (approved):

src/webgpu/shader/validation/shader_io/util.ts
jrprice: Apparently there is syntax, which I've used. Don't really have a setup to check that it works though.

PR #782, opened 2021-10-18
Add src/common/utils/colors.ts

ben-clayton:
This is a thin wrapper around the 'ansi-colors' module, which will be used if it can be found.

This interface may also support HTML colorization in the future.

Will be used to make the mental parsing of expectation failures easier.

### DIFF
--- a/src/common/util/colors.ts
+++ b/src/common/util/colors.ts
@@ -1,0 +1,123 @@
+/**
+ * The interface used for formatting strings to contain color metadata.
+ *
+ * Use the interface properties to construct a style, then use the
+ * `(s: string): string` function to format the provided string with the given
+ * style.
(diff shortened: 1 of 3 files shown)


src/common/util/colors.ts
```diff
@@ -0,0 +1,123 @@
+/**
+ * The interface used for formatting strings to contain color metadata.
+ *
+ * Use the interface properties to construct a style, then use the
+ * `(s: string): string` function to format the provided string with the given
+ * style.
+ */
+export interface Colors {
```
austinEng, line 8: maybe it's possible to reuse types definitions from ansi-colors? not strictly necessary though

austinEng (approved):
LGTM

PR #785, opened 2021-10-18
src/common/util: Fix 5 second stall at end of test run

ben-clayton:
The `DeviceHolder.ensureRelease()` method calls `await raceWithRejectOnTimeout()` with a 5 second timeout. For the NodeJS command line test runner, this sets up a `setTimeout()` callback that the Node VM always has to wait on before quitting the process.

To fix this, cancel the `setTimeout()` if the other promise completes first.


austinEng (approved):
LGTM

PR #789, opened 2021-10-19
Add a commandline server test runner

ben-clayton:
This is similar to the cmdline tool, but acts a simple HTTP RPC test runner.

This is significantly faster than using the cmdline runner to run each test in isolation, as it remove the overhead of scanning for tests, and constructing a new device each test.

### DIFF
--- a/src/common/runtime/server.ts
+++ b/src/common/runtime/server.ts
@@ -44,11 +44,13 @@ let debug = false;
 for (let i = 0; i < sys.args.length; ++i) {
   const a = sys.args[i];
   if (a.startsWith('-')) {
-    if (a === '--verbose') {
-      debug = true;
-    } else if (a === '--gpu-provider') {
+    if (a === '--gpu-provider') {
       const modulePath = sys.args[++i];
       setGPUProvider(() => require(modulePath).gpu);
+    } else if (a === '--help') {
+      usage(1);
+    } else if (a === '--verbose') {
+      debug = true;
     }
   }
 }
@@ -56,7 +58,7 @@ for (let i = 0; i < sys.args.length; ++i) {
 (async () => {
   Logger.globalDebugMode = debug;
   const log = new Logger();
-  const testcases = await allWebGPUTestcases();
+  const testcases = allWebGPUTestcases();
 
   async function allWebGPUTestcases() {
     const webgpuQuery = parseQuery('webgpu:*');
@@ -82,11 +84,16 @@ for (let i = 0; i < sys.args.length; ++i) {
   const server = http.createServer(
     async (request: http.IncomingMessage, response: http.ServerResponse) => {
       if (request.url === undefined) {
+        response.end('invalid url');
         return;
       }
-      if (request.url.startsWith('/run?')) {
-        const name = request.url.substr(5);
-        const testcase = testcases.get(name);
+
+      const runPrefix = '/run?';
+      const terminatePrefix = '/terminate';
+
+      if (request.url.startsWith(runPrefix)) {
+        const name = request.url.substr(runPrefix.length);
+        const testcase = (await testcases).get(name);
         let status = 'fail';
         let message = '';
         if (testcase) {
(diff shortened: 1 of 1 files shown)


src/common/runtime/server.ts
```diff
@@ -0,0 +1,117 @@
...
+      }
+    }
+  );
+
+  server.listen(0, () => {
+    const address = server.address() as AddressInfo;
+    console.log(`Server listening at [[${address.port}]]`);
+  });
+})().catch(ex => {
+  console.log(ex.stack ?? ex.toString());
```
austinEng, line 115: console.error?
```suggestion
  console.error(ex.stack ?? ex.toString());
```

src/common/runtime/server.ts
```diff
@@ -0,0 +1,117 @@
...
+    return res;
+  }
+
+  const server = http.createServer(
+    async (request: http.IncomingMessage, response: http.ServerResponse) => {
+      if (request.url === undefined) {
+        return;
+      }
+      if (request.url.startsWith('/run?')) {
+        const name = request.url.substr(5);
```
austinEng, line 88: maybe:
```suggestion
      const runPrefix = '/run?';
      if (request.url.startsWith(runPrefix)) {
        const name = request.url.substr(runPrefix.length);
```

src/common/runtime/server.ts
```diff
@@ -0,0 +1,117 @@
...
+for (let i = 0; i < sys.args.length; ++i) {
+  const a = sys.args[i];
+  if (a.startsWith('-')) {
+    if (a === '--verbose') {
+      debug = true;
+    } else if (a === '--gpu-provider') {
+      const modulePath = sys.args[++i];
+      setGPUProvider(() => require(modulePath).gpu);
+    }
+  }
```
austinEng, line 53: check for `--help` and print the usage info?

src/common/runtime/server.ts
```diff
@@ -0,0 +1,117 @@
...
+}
+
+(async () => {
+  Logger.globalDebugMode = debug;
+  const log = new Logger();
+  const testcases = await allWebGPUTestcases();
```
austinEng, line 59: nit: if we just have

`const testcasesPormise = allWebGPUTestcases();`

and later on inside the request handler do:

`const testcase = (await testcasesPromise).get(name);`

then the server will get instantiated and start listening right away, and block later on if testcases aren't done loading yet.
I find it a minor annoyance when I start a tool, hit the web server, and then realize it hasn't started listening yet.

src/common/runtime/server.ts
```diff
@@ -0,0 +1,117 @@
...
+    const name = testcase.query.toString();
+    const [rec, res] = log.record(name);
+    await testcase.run(rec, expectations);
+    return res;
+  }
+
+  const server = http.createServer(
+    async (request: http.IncomingMessage, response: http.ServerResponse) => {
+      if (request.url === undefined) {
+        return;
```
austinEng, line 85: when does this happen? should we return an error status code?

src/common/runtime/server.ts
```diff
@@ -0,0 +1,117 @@
...
+          }
+        } else {
+          message = `test case '${name}' not found`;
+        }
+        const res: RunResult = { status, message };
+        response.end(JSON.stringify(res));
+      } else if (request.url.startsWith('/terminate')) {
+        server.close();
+        sys.exit(1);
+      }
```
austinEng, line 106: error response if it doesn't match /run or /terminate? I think these requests will hang around if they're never answered

src/common/runtime/server.ts
ben-clayton: Done

src/common/runtime/server.ts
ben-clayton: Done

src/common/runtime/server.ts
ben-clayton: Done

PR #791, opened 2021-10-20
common/util: Fix subtle bug introduced in #785

ben-clayton:
Apparently you cannot use a Promise that has had `.finally()` called on it for other `.then()` /  `.finally()` type uses, you *must* use the returned Promise.

Fixes weird failures that have been observed in the Chromium CTS roll.


jrprice (approved):
LGTM and can confirm that it fixes some weirdness I saw locally, but I'm not well versed in Javascript so should probably get a review from somebody that is.

PR #807, opened 2021-10-29
Serve directory listings to make it easier to find html files

kainino0x:
Apparently `express` no longer does this. Add a thing to make it work again.


austinEng (approved):

austinEng:
missing types though

PR #800, opened 2021-10-27
Test isnan

dneto0:
<hr>

**Author checklist for test code/plans:**

- [x] All outstanding work is tracked with "TODO" in a test/file description or `.unimplemented()` on a test.
- [x] New helpers, if any, are documented using `/** doc comments */` and can be found via `helper_index.txt`.
- [x] (Optional, sometimes not possible.) Tests pass (or partially pass without unexpected issues) in an implementation. (Add any extra details above.)

**[Reviewer sign-off](https://github.com/gpuweb/cts/blob/main/docs/reviews.md) for test code/plans:** (Note: feel free to pull in other reviewers at any time for any reason.)

- [ ] The test path is reasonable, the [description](https://github.com/gpuweb/cts/blob/main/docs/intro/plans.md) "describes the test, succinctly, but in enough detail that a reader can read only the test plans in a file or directory and evaluate the completeness of the test coverage."
- [ ] Tests appear to cover this area completely, except for outstanding TODOs. Validation tests use control cases.
    (This is critical for coverage. Assume anything without a TODO will be forgotten about forever.)
- [ ] Existing (or new) test helpers are used where they would reduce complexity.
- [ ] TypeScript code is readable and understandable (is unobtrusive, has reasonable type-safety/verbosity/dynamicity).

PR #810, opened 2021-11-01
shader/execution/builtin: Overhaul builtin tests

ben-clayton:
`NumberRepr` has been replaced with `Scalar`, `Vector`, `ScalarType` and `VectorType`, and a whole bunch of helpers.
These new types provide a cleaner set of primitives to construct test cases, and allows us to consolidate the test logic for all builtin types (instead of an entirely new test for each overload flavor).

Update all the existing builtin tests to use the new framework.

<hr>

### DIFF
--- a/src/webgpu/shader/execution/builtin/abs.spec.ts
+++ b/src/webgpu/shader/execution/builtin/abs.spec.ts
@@ -3,12 +3,18 @@ Execution Tests for the 'abs' builtin function
 `;
 
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
-import { assert } from '../../../../common/util/util.js';
 import { GPUTest } from '../../../gpu_test.js';
-import { NumberRepr } from '../../../util/conversion.js';
-import { generateTypes } from '../../types.js';
+import {
+  F32,
+  F32Bits,
+  I32Bits,
+  TypeF32,
+  TypeI32,
+  TypeU32,
+  U32Bits,
+} from '../../../util/conversion.js';
 
-import { kBit, kValue, runShaderTest } from './builtin.js';
+import { anyOf, kBit, kValue, run } from './builtin.js';
 
 export const g = makeTestGroup(GPUTest);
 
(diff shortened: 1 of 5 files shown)


src/webgpu/shader/execution/builtin/builtin.ts
```diff
@@ -73,94 +297,89 @@ export function runShaderTestImpl<F extends NumberType>(
...
+// Packs a list of scalar test cases into a smaller list of vector cases.
+export function packScalarsToVector(cases: CaseList, vectorWidth: number): CaseList {
+  const asScalar = function (val: Array<Value> | Value): Scalar {
+    if (val instanceof Scalar) {
+      return val;
+    }
+    if (val instanceof Array && val.length === 1) {
+      return asScalar(val[0]);
+    }
+    throw new Error('packScalarsToVector() requires all Case values to be a single scalar Value');
```
dneto0, line 340: So if I understand correctly, this is not intended to work with a builtin function that takes multiple arguments, because the input field would be an array of scalars.
Example:     `select` would have input example of [number, number, bool] which is an array of scalars.

That's ok, but not super clear.  Maybe an updated comment at 331?

src/webgpu/shader/execution/builtin/abs.spec.ts
```diff
@@ -26,61 +32,48 @@ Component-wise when T is a vector.
   )
   .params(u =>
     u
-      .combineWithParams([{ storageClass: 'storage', storageMode: 'read_write' }] as const)
-      .combine('containerType', ['scalar', 'vector'] as const)
-      .combine('isAtomic', [false])
-      .combine('baseType', ['u32'] as const)
-      .expandWithParams(generateTypes)
+      .combine('storageClass', ['uniform', 'storage_r', 'storage_rw'] as const)
```
dneto0, line 35: Future cleanup:  I don't think storage class parameterization is interesting for testing the 'abs' builtin, or any builtin which is a pure function (depending only on its parameters)

src/webgpu/shader/execution/builtin/builtin.ts
```diff
@@ -1,53 +1,277 @@
...
+  got: string; // The string representation of the 'got' value (possibly with markup)
+  expected: string; // The string representation of the 'expected' value (possibly with markup)
+}
+
+/** Comparator is a function that compares whether the provided value matches an expectation. */
+export interface Comparator {
+  (got: Value): Comparison;
+}
+
+// diff compares 'got' to 'expected, returning the Comparison information.
```
dneto0, line 26: 'diff' seems like it succeeds when the values are different....

Should this be 'same' ?  or 'equalValue' or 'comparesAsEqual'?

src/webgpu/shader/execution/builtin/builtin.ts
```diff
@@ -1,53 +1,277 @@
...
+/**
+ * Runs a builtin test
+ * @param t the GPUTest
+ * @param builtin the builtin being tested
+ * @param parameterTypes the list of builtin parameter types
+ * @param returnType the return type for the builtin overload
+ * @param cfg test configuration values
+ * @param cases list of test cases
+ */
+export function run(
```
dneto0, line 184: This makes the common case really nice.

One thing this doesn't handle is, for example, what I've done with the isNan tests https://github.com/gpuweb/cts/pull/800/files#diff-c4a58312f4e5e245645aaccd1702c362a848dd0e0096ee2bc9f79e903081ea9bR79 

where I construct specific bit patterns that map to NaNs, and handle them as Uints on the host side, but they are read and manipulated as f32 on the device side.

I think that can be handled with a special hybrid of F32 and U32.  I can do that as followup, because it's very niche.

ben-clayton: Does `F32Bits` not give you what you need?

src/webgpu/shader/execution/builtin/builtin.ts
ben-clayton: Define 'succeeds'? It does not throw or fail on difference, it simply returns a structure that says whether they match, along with a marked-up string representation of what was different.
So, I'm not sure I see why `diff` is not an appealing name.

src/webgpu/shader/execution/builtin/abs.spec.ts
ben-clayton: Mayyyyyybe. HLSL has a `RW` prefix on `ByteAddressBuffer` for mutable buffers. There is some potential for divergence here

src/webgpu/util/conversion.ts
```diff
@@ -223,124 +224,303 @@ export function uint32ToInt32(u32: number): number {
...
+    if (this.type.kind === 'bool') {
+      return Colors.bold(this.value.toString());
+    }
+    switch (this.value) {
+      case 0:
+      case Infinity:
+      case -Infinity:
+        return Colors.bold(this.value.toString());
+      default:
+        return Colors.bold(this.value.toString()) + ' (0x' + this.value.toString(16) + ')';
```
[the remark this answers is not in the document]
ben-clayton: Ack.

src/webgpu/shader/execution/builtin/builtin.ts
```diff
@@ -1,53 +1,277 @@
...
-    }
+  parameterTypes: Array<Type>,
+  returnType: Type,
+  cfg: Config = { storageClass: 'storage_r' },
+  cases: CaseList
+) {
+  // If the 'vectorize' config option was provided, pack the cases into vectors.
+  if (cfg.vectorize !== undefined) {
+    cases = packScalarsToVector(cases, cfg.vectorize);
+    parameterTypes = [TypeVec(cfg.vectorize, parameterTypes[0] as ScalarType)];
```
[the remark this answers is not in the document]
ben-clayton: Correct. Added a comment here

src/webgpu/shader/execution/builtin/builtin.ts
```diff
@@ -73,94 +297,89 @@ export function runShaderTestImpl<F extends NumberType>(
...
 
   t.expectGPUBufferValuesPassCheck(outputBuffer, checkExpectation, {
-    type: arrayType,
-    typedLength: inputData.length,
+    type: Uint8Array,
+    typedLength: outputBufferSize,
   });
 }
 
+// Packs a list of scalar test cases into a smaller list of vector cases.
```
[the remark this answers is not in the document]
ben-clayton: Done

src/webgpu/shader/execution/builtin/builtin.ts
ben-clayton: Done

src/webgpu/shader/execution/builtin/all.spec.ts
```diff
@@ -122,72 +21,63 @@ e: vecN<bool> all(e): bool Returns true if each component of e is true. (OpAll)
...
+          { input: vec3(True, True, False), expected: False },
+          { input: vec3(False, False, True), expected: False },
+          { input: vec3(True, False, True), expected: False },
+          { input: vec3(False, True, True), expected: False },
+          { input: vec3(True, True, True), expected: True },
+        ],
+      },
+      vec4: {
+        type: TypeVec(4, TypeBool),
+        cases: [
```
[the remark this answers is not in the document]
ben-clayton: Thanks :)

ben-clayton:
I'm going to land this with the two remaining open issues as I see you (@dneto0) are in meetings, and I want to unblock the team from using this stuff.
Will address comments if my responses aren't compelling.

dneto0:
Your updates for review look good.  Thanks!

Issue #792, opened 2021-10-20
How should accuracy of floating point builtins be evaluated?


zoddicus:
Currently the WGSL spec specifies the accuracy required for a variety of FP builtins, https://www.w3.org/TR/WGSL/#floating-point-accuracy

A large number of them resolve down to an accuracy in terms of ULP, which can be then converted into a concrete epsilon that values should be within to the true value.

The issue becomes, what is the true value?
Is it the JS/Typescript value for that specific operation?
Is it from a table of values generated offline via a precision C math library? Maybe generated as part of the build process for the CTS?
Is there some other source of truth?

Additionally, how many values do we test it against, just a set of specific interesting values, all the potential values, or maybe a random selection of values?

Instead of having the tests for each builtin figure this out for themselves, and potentially be inconsistent with each other, there should probably be a standardized way of doing this, and probably a related doc outlining it.

zoddicus:
Additionally how should tests test for accuracy when it is defined as inherited from another function?

For example:
`asin`'s accuracy is `Inherited from atan2(x, sqrt(1.0 - x * x))`
`atan2`'s accuracy is `4096 ULP`
`sqrt`'s accuracy is `Inherited from 1.0 / inverseSqrt(x)`
`inverseSqrt`'s accuracy is `2 ULP`

dneto0:
Let's talk about ECMAScript first:

- a Number is the same as IEEE-754 binary64 (double precision float).
- the Math object provides things like sin, cos, atan, etc.   All the interesting ones are declared to be [implementation-approximated](https://tc39.es/ecma262/#implementation-approximated):
   > An _implementation-approximated_ facility is one that defers its definition to an external source while recommending an ideal behaviour. While conforming implementations are free to choose any behaviour within the constraints put forth by this specification, they are encouraged to strive to approximate the ideal.
- basic arithmetic is provided by the Number object. It describes edge cases (infinities, +0, -0,NaNs).  It doesn't describe accuracy requirements, but defers to IEEE-754. 
    - I looked at IEEE-754.  I'm not 100% certain but it seems to imply requiring correctly rounded results for add, sub, mul, div, and sqrt.
    - C++14 says "Since this International Standard imposes no restrictions on the accuracy of floating-point operations..."  which goes entirely in the opposite direction.

So, in principal we can't guarantee consistency across implementations for JavaScript math. That's my reading.

[...]

PR #811, opened 2021-11-01
shader/execution/builtin: Add `cos()` test

ben-clayton:
And untitle the scalar constructor function names.
The CTS has a convention of untitled function names, and this matches the WGSL syntax better anyway.

<hr>


src/webgpu/shader/execution/builtin/builtin.ts
```diff
@@ -18,13 +19,50 @@ export interface Comparison {
   expected: string; // The string representation of the 'expected' value (possibly with markup)
 }
 
+/** FloatTest is a function that compares whether the two floating point numbers match. */
+export interface FloatTest {
```
dneto0, line 23: nit: Maybe this should be called FloatMatch ?

src/webgpu/shader/execution/builtin/builtin.ts
```diff
@@ -18,13 +19,50 @@ export interface Comparison {
...
+ */
+export function ulpThreshold(ulp: number): FloatTest {
+  return (got, expected) => {
+    if (got === expected) {
+      return true;
+    }
+    if (!Number.isFinite(got) || !Number.isFinite(expected)) {
+      return false;
+    }
+    return diffULP(got, expected) <= ulp;
```
dneto0, line 60: I wonder if diffULP should be updated to return Infinity when either argument is NaN or Infinity.
Then callers could be simplified, like here.

src/webgpu/shader/execution/builtin/builtin.ts
```diff
@@ -18,13 +19,50 @@ export interface Comparison {
...
+  };
 }
 
 // diff compares 'got' to 'expected, returning the Comparison information.
-function diff(got: Value, expected: Value): Comparison {
+function diff(got: Value, expected: Value, cmpFloats: FloatTest): Comparison {
```
dneto0, line 65: Suggest renaming 'diff' to valuesMatch

src/webgpu/shader/execution/builtin/cos.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+  )
+  .params(u =>
+    u
+      .combine('storageClass', ['uniform', 'storage_r', 'storage_rw'] as const)
+      .combine('vectorize', [undefined, 2, 3, 4] as const)
+  )
+  .fn(async t => {
+    const cases = new Array<Case>(1000);
+    for (let i = 0; i < cases.length; i++) {
+      const angle = -Math.PI + (2.0 * Math.PI * i) / cases.length;
```
dneto0, line 33: This is the easy range for cosine.  Yes, it's periodic, but sometimes the hard part is in doing the range reduction accurately.  Values nearest the largest positive and negative finite values would be interesting too.

Suggestions:
- divide by cases.length-1  so the +Math.PI case is tested too


Please reference https://github.com/gpuweb/cts/issues/792 so this sampling can be visited.  With a reference like that, I'm ok landing this for now.

dneto0 (changes requested):
Please add a link to #792.

Also, naming could be improved.

src/webgpu/shader/execution/builtin/cos.spec.ts
dneto0: The reference should be listed as TODO so it can be picked up automatically later.

ben-clayton: Done.

src/webgpu/shader/execution/builtin/builtin.ts
ben-clayton: Renamed to `compare`. Hope this works for you.

src/webgpu/shader/execution/builtin/builtin.ts
ben-clayton: Done

src/webgpu/shader/execution/builtin/builtin.ts
ben-clayton: Done

dneto0 (approved):
Great. Thanks!

PR #799, opened 2021-10-27
test isInf builtin function

dneto0:
Test path:
webgpu:shader,execution,builtin,value_testing_built_in_functions:value_testing_builtin_functions,isInf:*

This passes on Dawn / SwiftShader

<hr>

**Author checklist for test code/plans:**

- [x] All outstanding work is tracked with "TODO" in a test/file description or `.unimplemented()` on a test.
- [x] New helpers, if any, are documented using `/** doc comments */` and can be found via `helper_index.txt`.
- [x] (Optional, sometimes not possible.) Tests pass (or partially pass without unexpected issues) in an implementation. (Add any extra details above.)

**[Reviewer sign-off](https://github.com/gpuweb/cts/blob/main/docs/reviews.md) for test code/plans:** (Note: feel free to pull in other reviewers at any time for any reason.)

- [ ] The test path is reasonable, the [description](https://github.com/gpuweb/cts/blob/main/docs/intro/plans.md) "describes the test, succinctly, but in enough detail that a reader can read only the test plans in a file or directory and evaluate the completeness of the test coverage."
- [ ] Tests appear to cover this area completely, except for outstanding TODOs. Validation tests use control cases.
    (This is critical for coverage. Assume anything without a TODO will be forgotten about forever.)
- [ ] Existing (or new) test helpers are used where they would reduce complexity.
- [ ] TypeScript code is readable and understandable (is unobtrusive, has reasonable type-safety/verbosity/dynamicity).

PR #816, opened 2021-11-03
test isInf builtin function

ben-clayton:
Forked from https://github.com/gpuweb/cts/pull/799.

---
Test path:
webgpu:shader,execution,builtin,value_testing_built_in_functions:value_testing_builtin_functions,isInf:*

This passes on Dawn / SwiftShader

---

Also contains:
* a fix for a typo in `toStorage()` where all packed booleans were returning `false`.
* test batching to handle case lists that exceed the binding limits for the given storage class.

<hr>

### DIFF
--- a/src/webgpu/shader/execution/builtin/builtin.ts
+++ b/src/webgpu/shader/execution/builtin/builtin.ts
@@ -153,10 +153,13 @@ export type Case = {
 /** CaseList is a list of Cases */
 export type CaseList = Array<Case>;
 
+/** The storage class to use on test input buffers */
+export type StorageClass = 'uniform' | 'storage_r' | 'storage_rw';
+
 /** Configuration for running a builtin test */
 export type Config = {
   // Where the input values are read from
-  storageClass: 'uniform' | 'storage_r' | 'storage_rw';
+  storageClass: StorageClass;
   // If defined, scalar test cases will be packed into vectors of the given width.
   // Requires that all parameters of the builtin overload are of a scalar type,
   // and the return type of the builtin overload is also a scalar type.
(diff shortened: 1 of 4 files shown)


src/webgpu/shader/execution/builtin/builtin.ts
```diff
@@ -205,14 +208,20 @@ function toStorage(ty: Type, expr: string): string {
   }
   if (ty instanceof VectorType) {
     if (ty.elementType.kind === 'bool') {
-      return `select(vec${ty.width}<u32>(0u), vec${ty.width}<u32>(0u), ${expr})`;
+      return `select(vec${ty.width}<u32>(0u), vec${ty.width}<u32>(1u), ${expr})`;
```
dneto0, line 211: Ah. That explains it.
Sorry I didn't catch this on the earlier review.

src/webgpu/shader/execution/builtin/builtin.ts
```diff
@@ -211,8 +214,14 @@ function toStorage(ty: Type, expr: string): string {
...
 
+// Currently all values are packed into buffers of 16 byte strides
+const kValueStride = 16;
+
 /**
- * Runs a builtin test
+ * Runs the list of builtin tests, possibly splitting the tests into multiple
+ * dispatches to keep the input data within the buffer binding limits.
+ * run() will pack the scalar test cases into smaller set of vectorized tests
+ * if `cfg.vectorize` is defined.
```
dneto0, line 224: nit: Suggest adding:  

> If defined, cfg.vectorize is the vector width, one of 2,3, or 4.

dneto0 (approved):
LGTM with tiny documentation nit.

dneto0:
Note: https://gpuweb.github.io/gpuweb/wgsl/#floating-point-evaluation says

> Implementations may assume that NaNs, inifinities are not present

> Note: This means some functions (e.g. isInf, isNan, min and max) may not return the expected result due to optimizations about the presence of NaNs and infinities.

For now, I'd like to land the test assuming the more strict behaviour.  We can relax it later.

src/webgpu/shader/execution/builtin/builtin.ts
ben-clayton: There was already a comment on the `Config.vectorize` field, but I've expanded this to be explicit about the allowed values.

src/webgpu/shader/execution/builtin/builtin.ts
ben-clayton: No need to apologise, my stupid typo.

dneto0 (approved):

PR #796, opened 2021-10-26
shader/execution: Implement quantizeToF16() tests

ben-clayton:
<hr>

**Author checklist for test code/plans:**

- [x] All outstanding work is tracked with "TODO" in a test/file description or `.unimplemented()` on a test.
- [x] New helpers, if any, are documented using `/** doc comments */` and can be found via `helper_index.txt`.
- [x] (Optional, sometimes not possible.) Tests pass (or partially pass without unexpected issues) in an implementation. (Add any extra details above.)

**[Reviewer sign-off](https://github.com/gpuweb/cts/blob/main/docs/reviews.md) for test code/plans:** (Note: feel free to pull in other reviewers at any time for any reason.)

- [ ] The test path is reasonable, the [description](https://github.com/gpuweb/cts/blob/main/docs/intro/plans.md) "describes the test, succinctly, but in enough detail that a reader can read only the test plans in a file or directory and evaluate the completeness of the test coverage."
- [ ] Tests appear to cover this area completely, except for outstanding TODOs. Validation tests use control cases.
    (This is critical for coverage. Assume anything without a TODO will be forgotten about forever.)
- [ ] Existing (or new) test helpers are used where they would reduce complexity.
- [ ] TypeScript code is readable and understandable (is unobtrusive, has reasonable type-safety/verbosity/dynamicity).

PR #821, opened 2021-11-03
Implement `atan` tests

zoddicus:
<hr>

### DIFF
--- a/src/webgpu/shader/execution/builtin/atan.spec.ts
+++ b/src/webgpu/shader/execution/builtin/atan.spec.ts
@@ -4,9 +4,9 @@ Execution Tests for the 'atan' builtin function
 
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
 import { GPUTest } from '../../../gpu_test.js';
-import { f32, TypeF32 } from '../../../util/conversion.js';
+import { f32, f32Bits, TypeF32 } from '../../../util/conversion.js';
 
-import { ulpThreshold, Case, Config, run } from './builtin.js';
+import { Case, Config, kBit, run, ulpThreshold } from './builtin.js';
 
 export const g = makeTestGroup(GPUTest);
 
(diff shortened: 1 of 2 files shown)


src/webgpu/shader/execution/builtin/atan.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+`
+  )
+  .params(u =>
+    u
+      .combine('storageClass', ['uniform', 'storage_r', 'storage_rw'] as const)
+      .combine('vectorize', [undefined, 2, 3, 4] as const)
+  )
+  .fn(async t => {
+    // TODO(https://github.com/gpuweb/cts/issues/792): Decide what the ground-truth is for these tests.
+    const truth_func = (x: number): number => {
```
ben-clayton, line 29: nit: CTS style uses `camelCase`, not `snake_case`

src/webgpu/shader/execution/builtin/atan.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+      return Math.atan(x);
+    };
+    // Well defined/border cases
+    const manual_inputs: Array<number> = [
+      Number.NEGATIVE_INFINITY, // -Inf -> -90°
```
ben-clayton, line 34: We seem to be mixing how we declare infinities. [David has been using `Infinity` and `-Infinity`](https://github.com/gpuweb/cts/blob/8d46b8304c5978490b941c41b521b8ae35b2d7db/src/webgpu/shader/execution/builtin/isinf.spec.ts#L36-L37), [I've been using `f32Bits(0x7f800000)` and `f32Bits(0xff800000)`](https://github.com/gpuweb/cts/pull/796/files), and you're using `Number.POSITIVE_INFINITY` and `Number.NEGATIVE_INFINITY`, and I just stumbled across Sarah's [`kBit.f32.infinity.positive` and `kBit.f32.infinity.negative`](https://github.com/gpuweb/cts/blob/8d46b8304c5978490b941c41b521b8ae35b2d7db/src/webgpu/shader/execution/builtin/builtin.ts#L549-L552)

I've no particular strong feeling which one is better, but we should probably be consistent. Sarah's constants predate all of ours, so maybe we should switch to using them. @dneto0 - thoughts?

David

zoddicus: In a local patch I have converted the manual cases to use defined values, as requested in a related PR, i.e.
`{ input: f32(Number.NEGATIVE_INFINITY), expected: f32(-Math.PI / 2) },`

This passes, but if I change the input to `kBit.f32.infinity.negative`, the test starts failing because the return value is actually Pi/2 instead of -Pi/2.

I am unsure if this is an issue with the testing framework or TS/JS conversions.

ben-clayton: > This passes, but if I change the input to `kBit.f32.infinity.negative`, the test starts failing 

This is likely because these numbers are unsigned integer representations of the float values. You need to pass these to `f32Bits` not `f32`

zoddicus: Yup that fixed it.

src/webgpu/shader/execution/builtin/atan.spec.ts
zoddicus: Done

zoddicus:
@ben-clayton ptal

src/webgpu/shader/execution/builtin/atan.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+      { input: f32(0), expected: f32(0) },
+      { input: f32(Math.sqrt(3) / 3), expected: f32(Math.PI / 6) },
+      { input: f32(1), expected: f32(Math.PI / 4) },
+      { input: f32(Math.sqrt(3)), expected: f32(Math.PI / 3) },
+      { input: f32Bits(kBit.f32.infinity.positive), expected: f32(Math.PI / 2) },
+    ];
+
+    // Spread of cases over wide domain
+    const automatic = new Array<Case>(1000);
+    const increment = (Number.MAX_SAFE_INTEGER - Number.MIN_SAFE_INTEGER) / automatic.length;
```
ben-clayton, line 48: Not exactly sure what integers have to do with a `f32` value, but I guess this is a `Spread of cases over a wide domain`, so okay

src/webgpu/shader/execution/builtin/atan.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+    for (let i = 0; i < automatic.length; i++) {
+      const x = Number.MIN_SAFE_INTEGER + increment * i;
```
ben-clayton, line 50: I'm actually surprised that the ULP is not just guaranteed for a specific range, but this does appear to test the spec as it is currently written.

ben-clayton (approved):
Two slight surprises, but I think this is testing according to spec!
LGTM.

src/webgpu/shader/execution/builtin/atan.spec.ts
zoddicus: Good catch, this was meant to be a temp range, since TS doesn't specify float min/max. I have converted these to the values provided by kBit.

src/webgpu/shader/execution/builtin/atan.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+      { input: f32(0), expected: f32(0) },
+      { input: f32(Math.sqrt(3) / 3), expected: f32(Math.PI / 6) },
+      { input: f32(1), expected: f32(Math.PI / 4) },
+      { input: f32(Math.sqrt(3)), expected: f32(Math.PI / 3) },
+      { input: f32Bits(kBit.f32.infinity.positive), expected: f32(Math.PI / 2) },
+    ];
+
+    // Spread of cases over wide domain
+    const automatic = new Array<Case>(1000);
+    const increment = (kBit.f32.positive.max - kBit.f32.negative.min) / automatic.length;
```
ben-clayton, line 48: You're performing arithmetic in a `reinterpret_cast<uint32>(f32)` space, which isn't going to give you numbers that you expect.
You can use these constants, but you need to first convert them to true floats. You can do that with:

```ts
const f32Min = f32Bits(kBit.f32.positive.min).value as number
const f32Max = f32Bits(kBit.f32.positive.max).value as number
```

zoddicus: Thanks, I will write a follow up PR

PR #822, opened 2021-11-03
Implement `asin` tests

zoddicus:
<hr>

### DIFF
--- a/src/webgpu/shader/execution/builtin/asin.spec.ts
+++ b/src/webgpu/shader/execution/builtin/asin.spec.ts
@@ -1,0 +1,66 @@
+export const description = `
+Execution Tests for the 'asin' builtin function
+`;
+
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { GPUTest } from '../../../gpu_test.js';
+import { f32, TypeF32 } from '../../../util/conversion.js';
+
+import { ulpThreshold, Case, Config, run } from './builtin.js';
+
+export const g = makeTestGroup(GPUTest);
+
+g.test('float_builtin_functions,asin')
+  .uniqueId('322c7c5ba84c257a')
+  .specURL('https://www.w3.org/TR/2021/WD-WGSL-20210929/#float-builtin-functions')
+  .desc(
+    `
+asin:
+T is f32 or vecN<f32> asin(e: T ) -> T Returns the arc sine of e. Component-wise when T is a vector. (GLSLstd450Asin)
+`
+  )
+  .params(u =>
+    u
+      .combine('storageClass', ['uniform', 'storage_r', 'storage_rw'] as const)
+      .combine('vectorize', [undefined, 2, 3, 4] as const)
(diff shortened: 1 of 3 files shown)


src/webgpu/shader/execution/builtin/asin.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+    // Spread of cases over wide domain
+    const automatic_cases = new Array<Case>(1000);
+    const increment = 2 / automatic_cases.length;
+    for (let i = 0; i < automatic_cases.length; i++) {
+      const x = -1 + increment * i;
+      automatic_cases[i] = { input: f32(x), expected: f32(truth_func(x)) };
+    }
+
+    const cfg: Config = t.params;
+    cfg.cmpFloats = ulpThreshold(4096);
```
zoddicus, line 64: I am not sure if this is the correct value to use here.
`asin`'s accuracy is `Inherited from atan2(x, sqrt(1.0 - x * x))`
`atan2`'s accuracy is `4096 ULP`
`sqrt`'s accuracy is `Inherited from 1.0 / inverseSqrt(x)`
`inverseSqrt`'s accuracy is `2 ULP`

I am not sure if that means I should be going with tightest bounds or loosest bounds, or if there is a way to calculate the value between. Tests pass with both 4096 ULP and 2 ULP.

src/webgpu/shader/execution/builtin/builtin.ts
```diff
@@ -51,7 +51,7 @@ export function absThreshold(diff: number): FloatMatch {
  */
 export function ulpThreshold(ulp: number): FloatMatch {
   return (got, expected) => {
-    if (got === expected) {
+    if (got === expected || (Number.isNaN(got) && Number.isNaN(expected))) {
```
zoddicus, line 54: This is because comparing NaN to NaN is always false. I am not sure if this is behaviour we want all the time for tests, or if there should be a NaN-y and non-NaN-y version of ulpThreshold.

src/webgpu/shader/execution/builtin/asin.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+      .combine('vectorize', [undefined, 2, 3, 4] as const)
+  )
+  .fn(async t => {
+    // TODO(https://github.com/gpuweb/cts/issues/792): Decide what the ground-truth is for these tests.
+    const truth_func = (x: number): number => {
+      return Math.asin(x);
+    };
+
+    // Well defined/border cases
+    const manual_inputs: Array<number> = [
```
ben-clayton, line 34: For manual, well known values like this, I think it would be nicer to just explicitly declare the expected value (can be done using multiples / fractions of `PI`). Especially when you're relying on the JS engine's implementation around trig and DBZs / `NaN`s.

zoddicus:
I don't have the math chops either to figure out the finer points of the ULP calculation here.

PR #828, opened 2021-11-08
Test select with scalar condition

dneto0:
Also fixes uniqueness logic in TypeVec

<hr>

### DIFF
--- a/src/webgpu/shader/execution/builtin/select.spec.ts
+++ b/src/webgpu/shader/execution/builtin/select.spec.ts
@@ -1,0 +1,137 @@
+export const description = `WGSL execution test. Section: Logical built-in functions Function: select`;
+
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { GPUTest } from '../../../gpu_test.js';
+import {
+  Scalar,
+  TypeVec,
+  TypeBool,
+  TypeF32,
+  TypeI32,
+  TypeU32,
+  f32,
+  i32,
+  u32,
+  False,
+  True,
+  bool,
+  vec2,
+  vec3,
+  vec4,
+} from '../../../util/conversion.js';
+
+import { run } from './builtin.js';
+
+export const g = makeTestGroup(GPUTest);
+
+function makeBool(n: number) {
+  return bool((n & 1) === 1);
+}
+
+type scalarKind = 'b' | 'f' | 'i' | 'u';
+
+const dataType = {
+  b: {
+    type: TypeBool,
+    constructor: makeBool,
+  },
+  f: {
+    type: TypeF32,
+    constructor: f32,
+  },
+  i: {
+    type: TypeI32,
+    constructor: i32,
+  },
+  u: {
+    type: TypeU32,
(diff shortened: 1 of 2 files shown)


dneto0:
Passes on SwiftShader and Metal.

src/webgpu/shader/execution/builtin/select.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+function makeBool(n: number) {
+  return bool((n & 1) === 1);
+}
+
+type scalarKind = 'b' | 'f' | 'i' | 'u';
+
+const dataType = {
+  b: {
+    type: TypeBool,
+    constructor: makeBool,
```
ben-clayton, line 36: I was thinking of adding these to `ScalarType`, which would be useful for other tests. 

i.e. We could add a `create` field to `ScalarType`:
```ts
/** ScalarType describes the type of WGSL Scalar. */
export class ScalarType {
  readonly kind: ScalarKind; // The named type
  readonly size: number; // In bytes
  readonly create: (value: number) => Scalar; // Constructs a Scalar with the given value
  readonly read: (buf: Uint8Array, offset: number) => Scalar; // reads a scalar from a buffer
```

Then we'd pass these functions to the `ScalarType` constructors:

```ts
export const TypeI32 = new ScalarType('i32', 4, i32, (buf: Uint8Array, offset: number) =>
  i32(new Int32Array(buf.buffer, offset)[0])
);
```

To make this work, I think we'd have to change the `bool(value: boolean)` signature to `bool(value: boolean | number)`, so that bools don't have a different constructor signature.

If you're not keen on changing this, I can take a look at shuffling this to the `ScalarType`.

src/webgpu/util/conversion.ts
```diff
@@ -272,10 +272,11 @@ export class VectorType {
   }
 }
 
-const vectorTypes = new Map<{ width: number; elementType: ScalarType }, VectorType>();
+// Maps a string representation of a vector type to vector type.
+const vectorTypes = new Map<string, VectorType>();
 
 export function TypeVec(width: number, elementType: ScalarType): VectorType {
-  const key = { width, elementType };
```
dneto0, line 278: This code doesn't work because any time you create a new dictionary like this, it's a new object that never compares identically-equal to a previously created on.

ben-clayton (approved):
LGTM. I think the `constructor` stuff can be promoted to the `ScalarType` class, but I don't mind doing this if you'd prefer to land this as-is.

src/webgpu/util/conversion.ts
ben-clayton: TIL. Thanks for the fix!

src/webgpu/shader/execution/builtin/select.spec.ts
dneto0: I'll file this as an issue for followup.

dneto0: https://github.com/gpuweb/cts/issues/829

PR #835, opened 2021-11-11
Reland "Adds ETC2 and ASTC types to enable testing when relevant."

lokokung:
This is a reland of fd20759

Additional changes:
- Updates other tests using kAllTextureFormatInfo to be compatible with the new compressed texture types.
- Overhauls webgpu:api,validation,image_copy,layout_related:bound_on_bytes_per_row:* tests to reflect a more recent version of the testing for validation.

Original change's description:
> Adds ETC2 and ASTC types to enable testing when relevant.
>
> - Modifies texture sizes to be functions of the block sizes to generalize the tests.

<hr>

### DIFF
--- a/src/webgpu/api/validation/createTexture.spec.ts
+++ b/src/webgpu/api/validation/createTexture.spec.ts
@@ -130,11 +130,16 @@ g.test('mipLevelCount,format')
     const info = kTextureFormatInfo[format];
     await t.selectDeviceOrSkipTestCase(info.feature);
 
+    // Compute dimensions such that the dimensions are less than 2^6 (32) but also a multiple of the format block size.
+    const maxMipLevelCount = 6;
+    const textureWidth =
+      Math.floor(((1 << maxMipLevelCount) - 1) / info.blockWidth) * info.blockWidth;
+    const textureHeight =
+      Math.floor(((1 << maxMipLevelCount) - 1) / info.blockHeight) * info.blockHeight;
+
     // Note that compressed formats are not valid for 1D. They have already been filtered out for 1D in this test.
     // So there is no dilemma about size.width equals 1 vs size.width % info.blockHeight equals 0 for 1D compressed formats.
-    const size = dimension === '1d' ? [32, 1, 1] : [32, 32, 1];
-    assert(32 % info.blockWidth === 0 && 32 % info.blockHeight === 0);
-
+    const size = dimension === '1d' ? [textureWidth, 1, 1] : [textureWidth, textureHeight, 1];
     const descriptor = {
       size,
       mipLevelCount,
@@ -143,7 +148,7 @@ g.test('mipLevelCount,format')
       usage: GPUTextureUsage.TEXTURE_BINDING,
     };
 
-    const success = mipLevelCount <= 6;
+    const success = mipLevelCount <= maxMipLevelCount;
 
     t.expectValidationError(() => {
       t.device.createTexture(descriptor);
@@ -236,9 +241,10 @@ g.test('sampleCount,various_sampleCount_with_all_formats')
   .fn(async t => {
     const { dimension, sampleCount, format } = t.params;
     await t.selectDeviceOrSkipTestCase(kTextureFormatInfo[format].feature);
+    const { blockWidth, blockHeight } = kTextureFormatInfo[format];
 
     const descriptor = {
-      size: [32, 32, 1],
+      size: [32 * blockWidth, 32 * blockHeight, 1],
       sampleCount,
       dimension,
       format,
@@ -285,13 +291,14 @@ g.test('sampleCount,valid_sampleCount_with_other_parameter_varies')
   .fn(async t => {
     const { dimension, sampleCount, format, mipLevelCount, arrayLayerCount, usage } = t.params;
     await t.selectDeviceOrSkipTestCase(kTextureFormatInfo[format].feature);
+    const { blockWidth, blockHeight } = kTextureFormatInfo[format];
 
     const size =
       dimension === '1d'
-        ? [32, 1, 1]
+        ? [32 * blockWidth, 1 * blockHeight, 1]
         : dimension === '2d' || dimension === undefined
-        ? [32, 32, arrayLayerCount]
-        : [32, 32, 32];
+        ? [32 * blockWidth, 32 * blockHeight, arrayLayerCount]
+        : [32 * blockWidth, 32 * blockHeight, 32];
     const descriptor = {
       size,
       mipLevelCount,
@@ -505,11 +512,6 @@ g.test('texture_size,2d_texture,compressed_format')
     const { dimension, format, size } = t.params;
     const info = kTextureFormatInfo[format];
     await t.selectDeviceOrSkipTestCase(info.feature);
-
-    assert(
-      DefaultLimits.maxTextureDimension2D % info.blockWidth === 0 &&
-        DefaultLimits.maxTextureDimension2D % info.blockHeight === 0
-    );
 
     const descriptor: GPUTextureDescriptor = {
       size,
--- a/src/webgpu/api/validation/createTexture.spec.ts
+++ b/src/webgpu/api/validation/createTexture.spec.ts
@@ -130,7 +130,7 @@ g.test('mipLevelCount,format')
     const info = kTextureFormatInfo[format];
     await t.selectDeviceOrSkipTestCase(info.feature);
 
-    // Compute dimensions such that the dimensions are in range [17, 32] and alinged with the
+    // Compute dimensions such that the dimensions are in range [17, 32] and aligned with the
     // format block size so that there will be exactly 6 mip levels.
     const maxMipLevelCount = 5;
     const textureWidth =
--- a/src/webgpu/api/validation/createTexture.spec.ts
+++ b/src/webgpu/api/validation/createTexture.spec.ts
@@ -130,15 +130,18 @@ g.test('mipLevelCount,format')
     const info = kTextureFormatInfo[format];
     await t.selectDeviceOrSkipTestCase(info.feature);
 
-    // Compute dimensions such that the dimensions are less than 2^6 (32) but also a multiple of the format block size.
-    const maxMipLevelCount = 6;
-    const textureWidth =
-      Math.floor(((1 << maxMipLevelCount) - 1) / info.blockWidth) * info.blockWidth;
+    // Compute dimensions such that the dimensions are in range [17, 32] and alinged with the
+    // format block size so that there will be exactly 6 mip levels.
+    const maxMipLevelCount = 5;
+    const textureWidth = Math.floor((1 << maxMipLevelCount) / info.blockWidth) * info.blockWidth;
     const textureHeight =
       Math.floor(((1 << maxMipLevelCount) - 1) / info.blockHeight) * info.blockHeight;
+    assert(17 <= textureWidth && textureWidth <= 32);
+    assert(17 <= textureHeight && textureHeight <= 32);
 
-    // Note that compressed formats are not valid for 1D. They have already been filtered out for 1D in this test.
-    // So there is no dilemma about size.width equals 1 vs size.width % info.blockHeight equals 0 for 1D compressed formats.
+    // Note that compressed formats are not valid for 1D. They have already been filtered out for 1D
+    // in this test. So there is no dilemma about size.width equals 1 vs
+    // size.width % info.blockHeight equals 0 for 1D compressed formats.
     const size = dimension === '1d' ? [textureWidth, 1, 1] : [textureWidth, textureHeight, 1];
     const descriptor = {
       size,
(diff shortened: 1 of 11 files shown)


lokokung:
I merged the changes into Chromium and ran the tests locally with the harness and did not seem to see any regressions this time. I made a couple of smaller changes afterwards so I might re-run it one more time after revisions and before we merge.

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -130,11 +130,16 @@ g.test('mipLevelCount,format')
     const info = kTextureFormatInfo[format];
     await t.selectDeviceOrSkipTestCase(info.feature);
 
+    // Compute dimensions such that the dimensions are less than 2^6 (32) but also a multiple of the format block size.
```
kainino0x, line 133: nit: More precisely I think what you're trying to get is a size between 17 and 32 inclusive (so there will be exactly 6 mip levels). So perhaps comment that and assert afterward than the results are in that range.

src/webgpu/api/validation/image_copy/layout_related.spec.ts
```diff
@@ -250,68 +254,86 @@ g.test('offset_alignment')
...
-        { blocksPerRow: 1, additionalPaddingPerRow: 0, copyWidthInBlocks: 2 }, // failure, bytesPerRow > 0
-        { blocksPerRow: 0, additionalPaddingPerRow: 0, copyWidthInBlocks: 1 }, // failure, bytesPerRow = 0
-      ])
-      .combineWithParams([
-        { copyHeightInBlocks: 0, copyDepth: 1 }, // we don't have to check the bound
-        { copyHeightInBlocks: 1, copyDepth: 0 }, // we don't have to check the bound
-        { copyHeightInBlocks: 2, copyDepth: 1 }, // we have to check the bound
-        { copyHeightInBlocks: 0, copyDepth: 2 }, // we have to check the bound
-      ])
+      .expand('params', p => {
```
kainino0x, line 264: You can use `.expandWithParams(p => {` and then each of the members below (`bytesPerRow`, `blocksPerRow`, `copyWidthInBlocks`, `success`) will arrive as a separate parameter.

This makes the query strings a bit nicer, but more importantly it lets you change `success` to `_success` which makes it a "private" parameter which doesn't show up in the query string.

src/webgpu/api/validation/image_copy/layout_related.spec.ts
```diff
@@ -250,68 +254,86 @@ g.test('offset_alignment')
...
+        const info = kTextureFormatInfo[p.format];
+        return [
+          // Copying exact fit with aligned bytesPerRow should work.
+          {
+            bytesPerRow: 256,
+            blocksPerRow: 256 / info.bytesPerBlock,
```
kainino0x, line 270: Where is blocksPerRow used? I don't see it in the test body.

src/webgpu/api/validation/image_copy/layout_related.spec.ts
```diff
@@ -250,68 +254,86 @@ g.test('offset_alignment')
...
+            copyWidthInBlocks: 256 / info.bytesPerBlock,
```
kainino0x, line 271: These divisions can result in fractions, so using `Math.floor` would better express what the cases are. (truncation happens implicitly in conversions to integer in WebIDL)

src/webgpu/api/validation/image_copy/layout_related.spec.ts
```diff
@@ -250,68 +254,86 @@ g.test('offset_alignment')
...
+          // Copying larger texture when padding in bytesPerRow is enough should work.
+          {
+            bytesPerRow: 256,
+            blocksPerRow: 256 / info.bytesPerBlock - 1,
+            copyWidthInBlocks: 256 / info.bytesPerBlock,
+            success: true,
+          },
+          // Unaligned bytesPerRow should not work unless the method is 'WriteTexture'.
+          {
+            bytesPerRow: 257,
```
kainino0x, line 283: For alignment tests we try to test "minimally unaligned" values, i.e. 128 or 384 in this case, which is a multiple of the largest power of 2 without being a multiple of 256. That would be the most likely to erroneously pass if the alignment check was wrong. (Keeping a case for 257 also makes sense)

src/webgpu/api/validation/image_copy/layout_related.spec.ts
```diff
@@ -250,68 +254,86 @@ g.test('offset_alignment')
...
+            copyWidthInBlocks: 256 / info.bytesPerBlock + 1,
+            success: false,
+          },
+          // When bytesPerRow is smaller than bytesInLastRow copying should fail.
+          {
+            bytesPerRow: 256,
+            blocksPerRow: (2 * 256) / info.bytesPerBlock,
+            copyWidthInBlocks: (2 * 256) / info.bytesPerBlock,
+            success: false,
+          },
```
kainino0x, line 301: We should also have tests for `bytesPerRow: undefined` (unless those already exist somewhere else), which is valid only when height<=1 and depth<=1.

This could be a TODO but it would be good to add because I'm pretty sure it's a hole in our test coverage since we subtly changed the validation rules here a while back.

src/webgpu/api/validation/image_copy/layout_related.spec.ts
```diff
@@ -250,68 +254,86 @@ g.test('offset_alignment')
...
+            success: p.method === 'WriteTexture',
+          },
+          // Copying larger texture when padding in bytesPerRow is not enough should fail.
+          {
+            bytesPerRow: 256,
+            blocksPerRow: 256 / info.bytesPerBlock,
+            copyWidthInBlocks: 256 / info.bytesPerBlock + 1,
+            success: false,
+          },
+          // When bytesPerRow is smaller than bytesInLastRow copying should fail.
```
kainino0x, line 295: Don't this and the previous test hit the same validation rule? (That's fine, but the comments make it seem otherwise)

kainino0x (commented):
Mostly lgtm but had a few comments!

I found that it would have been easier to review with a clean reland at the base of the commit-stack - then I could ignore it since I previously reviewed it. I ended up doing this locally to look at the diffs.

```
79be3d8f 2021-11-11 Update more kAllTextureFormatInfo tests, overhaul bound_on_bytes_per_row (48 minutes ago) <Loko Kung>
22390511 2021-11-11 Update tests that use kAllTextureFormatInfo (48 minutes ago) <Loko Kung>
85fda638 2021-11-11 Revert "Revert "Adds ETC2 and ASTC types to enable testing when relevant. (#805)" (#831)" (49 minutes ago) <Kai Ninomiya>
```

(Anyway, now that I've reviewed it, you don't need to do that and force-push. Just push new changes as separate commits, as usual)

src/webgpu/api/validation/image_copy/layout_related.spec.ts
```diff
@@ -162,7 +162,11 @@ g.test('required_bytes_in_copy')
     const layout = { offset, bytesPerRow, rowsPerImage };
     const minDataSize = dataBytesForCopyOrFail({ layout, format, copySize, method });
 
-    const texture = t.createAlignedTexture(format, copySize);
+    const texture = t.createAlignedTexture(format, {
+      width: copyWidthInBlocks,
+      height: copyHeightInBlocks,
+      depthOrArrayLayers: copyDepth,
+    });
```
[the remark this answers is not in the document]
lokokung: I actually made createAlignedTexture take raw pixel size instead since taking sizes in block actually makes less sense. I also went and updated the other tests that were using it so that they make more sense now.

src/webgpu/api/validation/createTexture.spec.ts
lokokung: Done.

src/webgpu/api/validation/image_copy/layout_related.spec.ts
lokokung: Added test for it.

src/webgpu/api/validation/image_copy/layout_related.spec.ts
lokokung: I think they should be different? The first one checks that it fails because essentially out of bounds, and the second one should fail because bytesInLastRow = widthInBlocks * bytesPerBlock (512) > bytesPerRow (256).

src/webgpu/api/validation/image_copy/layout_related.spec.ts
lokokung: Added those cases and actually removed the 257 case because it didn't make sense with the other arguments.

src/webgpu/api/validation/image_copy/layout_related.spec.ts
lokokung: They should always (and need to) result in integers, so I added comment and assert before it to ensure that's the case instead.

src/webgpu/api/validation/image_copy/layout_related.spec.ts
lokokung: Updated and renamed. (It was wrong... good catch!)

src/webgpu/api/validation/image_copy/layout_related.spec.ts
lokokung: Done.

lokokung:
With the fixes, when I run the tests via the harness, I am getting time-out on webgpu:api,validation,image_copy,texture_related:origin_alignment:*. When I run the test locally though, it seems to pass fine. Is there a way to adjust the timeout value for these tests?

kainino0x:
In Chromium you'll want to add one line to https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/web_tests/webgpu/internal_cts_test_splits.pyl like:
```
  'wpt_internal/webgpu/cts.https.html?q=webgpu:api,validation,image_copy,texture_related:origin_alignment:method="WriteTexture";*',
```

and then rerun `regenerate_internal_cts_html.py`. You'll see in the generated `cts.https.html` that it split `origin_alignment` into several parts (in this case, one for each method). That should make each page-load take about 1/3 of the time which is most likely fast enough. If not you can do something like:

```
  'wpt_internal/webgpu/cts.https.html?q=webgpu:api,validation,image_copy,texture_related:origin_alignment:method="WriteTexture";format="r8unorm";*',
  'wpt_internal/webgpu/cts.https.html?q=webgpu:api,validation,image_copy,texture_related:origin_alignment:method="CopyB2T";format="r8unorm";*',
  'wpt_internal/webgpu/cts.https.html?q=webgpu:api,validation,image_copy,texture_related:origin_alignment:method="CopyT2B";format="r8unorm";*',
```
which will do one page load for each method+format combination (and be a lot slower).

kainino0x (approved):

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -130,15 +130,18 @@ g.test('mipLevelCount,format')
     const info = kTextureFormatInfo[format];
     await t.selectDeviceOrSkipTestCase(info.feature);
 
-    // Compute dimensions such that the dimensions are less than 2^6 (32) but also a multiple of the format block size.
-    const maxMipLevelCount = 6;
-    const textureWidth =
-      Math.floor(((1 << maxMipLevelCount) - 1) / info.blockWidth) * info.blockWidth;
+    // Compute dimensions such that the dimensions are in range [17, 32] and alinged with the
```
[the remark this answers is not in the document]
lokokung: Done

src/webgpu/api/validation/image_copy/texture_related.spec.ts
```diff
@@ -399,13 +399,10 @@ g.test('size_alignment')
...
-    const texture = t.createAlignedTexture(format, size, origin);
+    const texture = t.createAlignedTexture(format, size, origin, dimension);
 
-    const bytesPerRow = align(
-      (align(size.width, info.blockWidth) / info.blockWidth) * info.bytesPerBlock,
-      256
-    );
-    const rowsPerImage = align(size.height, info.blockHeight) / info.blockHeight;
+    const bytesPerRow = align(Math.max(1, size.width) * info.bytesPerBlock, 256);
+    const rowsPerImage = size.height;
```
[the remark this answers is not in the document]
lokokung: Oh! Good catch! Too many places where this was being confused...

src/webgpu/api/validation/image_copy/layout_related.spec.ts
```diff
@@ -261,79 +257,105 @@ g.test('bound_on_bytes_per_row')
...
+            widthInBlocks: 384 / info.bytesPerBlock,
+            copyWidthInBlocks: 384 / info.bytesPerBlock,
+            _success: p.method === 'WriteTexture',
           },
           // Copying larger texture when padding in bytesPerRow is not enough should fail.
           {
             bytesPerRow: 256,
-            blocksPerRow: 256 / info.bytesPerBlock,
+            widthInBlocks: 256 / info.bytesPerBlock,
             copyWidthInBlocks: 256 / info.bytesPerBlock + 1,
```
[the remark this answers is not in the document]
lokokung: Removed.

lokokung:
With all the changes above and https://chromium-review.googlesource.com/c/chromium/src/+/3286423, I was able to successfully run the tests and didn't see any new regression failures.

kainino0x (approved):
Nice work!

PR #836, opened 2021-11-16
Fix bug in testing copyExternalImageToTexture with *-srgb dst formats

shaoboyan:
In current cts implementation in webgpu:web_platform,copyToTexture,canvas:*,
*-srgb dst formats has been transferred to no -srgb formats to get the
expected results.

This CL fix this issue by remove the wrongly conversion.

<hr>


kainino0x (commented):
If I understand this change correctly it would make the *sample* values match between the source and the destination, rather than the *byte* values. I wrote in the spec that it should preserve the byte value. I think this choice makes sense (given the name), though after some consideration I think both choices would be valid.

shaoboyan:
After discussion with Kai, I think this PR is not correct. We should keep the bytes exactly same when copying to -srgb formats

PR #838, opened 2021-11-16
Handle zero-like cases with ULP accuracy constraints

### LINKED ISSUE #837: Gracefully handle ULP calculations when value is true 0
There are effectively multiple different ways that something zero-like may be expressed, negative zero `0x80000000, positive zero `0x00000000`, max negative `0xFF7FFFFF`, min positive `0x00800000`, etc.

Comparing something like negative zero to min positive can be tricky, so the calculations in `diffULP` need to check for things like negative zero and handle them correctly. The `atan` tests have hit issues related to this on some platforms.

zoddicus:
The added test cases for `atan` exercises this functionality

Fixes #837

<hr>

### DIFF
--- a/src/webgpu/util/math.ts
+++ b/src/webgpu/util/math.ts
@@ -1,4 +1,5 @@
 import { assert } from '../../common/util/util.js';
+import { kBit } from '../shader/execution/builtin/builtin.js';
 
 /**
  * A multiple of 8 guaranteed to be way too large to allocate (just under 8 pebibytes).
@@ -31,6 +32,20 @@ export function clamp(n: number, { min, max }: { min: number; max: number }): nu
 }
 
 /**
+ * @returns true if value is a zero like, otherwise false.
+ */
+function isZero(x: number): boolean {
+  return (
+    x === kBit.f32.positive.zero ||
+    x === 0x00000000 ||
+    x === kBit.f32.positive.min ||
+    x === kBit.f32.negative.zero ||
+    x === 0x80000000 ||
+    x === kBit.f32.negative.max
+  );
+}
+
+/**
  * @returns the Units of Last Place difference between the numbers a and b.
  * If either `a` or `b` are not finite numbers, then diffULP() returns Infinity.
  */
@@ -39,8 +54,20 @@ export function diffULP(a: number, b: number): number {
     return Infinity;
   }
   const arr = new Uint32Array(new Float32Array([a, b]).buffer);
-  const u32_a = arr[0];
-  const u32_b = arr[1];
+  let u32_a = arr[0];
+  let u32_b = arr[1];
+
+  // Handle zero value edge cases.
+  // If both values are a type of zero, then they are equal, so return 0.
+  // If only one value is a zero, adjust it to be the closest value to zero with
+  // the same sign as the other value, so that there doesn't need to be a bunch of special logic below.
+  if (isZero(u32_a) && isZero(u32_b)) {
+    return 0;
+  } else if (isZero(a) && !isZero(b)) {
+    u32_a = (b & 0x80000000) !== 0 ? kBit.f32.positive.min : kBit.f32.negative.max;
+  } else if (!isZero(a) && isZero(b)) {
+    u32_b = (a & 0x80000000) !== 0 ? kBit.f32.positive.min : kBit.f32.negative.max;
+  }
 
   const sign_a = (u32_a & 0x80000000) !== 0;
   const sign_b = (u32_b & 0x80000000) !== 0;
--- a/src/webgpu/shader/execution/builtin/atan.spec.ts
+++ b/src/webgpu/shader/execution/builtin/atan.spec.ts
@@ -1,3 +1,4 @@
+
 export const description = `
 Execution Tests for the 'atan' builtin function
 `;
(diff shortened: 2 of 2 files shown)


zoddicus:
Talked with @ben-clayton about this in chat, will be updating it

zoddicus:
PTAL, this should be ready to review now

zoddicus:
@dneto0 could you specifically review this, since this is ULP code that Ben and I came up with?

src/webgpu/util/math.ts
```diff
@@ -46,9 +47,16 @@ export function diffULP(a: number, b: number): number {
   const sign_b = (u32_b & 0x80000000) !== 0;
   const masked_a = u32_a & 0x7fffffff;
   const masked_b = u32_b & 0x7fffffff;
+  const subnormal_or_zero_a = (u32_a & 0x7f800000) === 0;
+  const subnormal_or_zero_b = (u32_b & 0x7f800000) === 0;
+
+  // If the number is not subnormal, then subtract 0x7fffff to remove the
+  // subnormal region from the ULP comparison.
+  const bits_a = masked_a - (subnormal_or_zero_a ? 0 : 0x7fffff);
```
dneto0, line 55: This seems wrong.  

From [wikipedia](https://en.wikipedia.org/wiki/Unit_in_the_last_place)

> Another definition, suggested by John Harrison, is slightly different: ulp(x)  is the distance between the two closest straddling floating-point numbers _a_ and _b_ (i.e., those with a &leq; x &leq; b  and  a &ne; b,  assuming that the exponent range is not upper-bounded.[3][4] 

So, for example when x = 0, then `a` and `b` can be selected as `a` = x = 0, and `b` is the smallest subnormal number. We may as well pick `b` having the same sign as the zero.

This avoids the divide-by-zero problem near zero.

Then diffULP(0,y), assuming y > 0, would be well defined.   If y is finite (either normal or subnormal), then the answer is going to be:
* 0 when y is zero
* masked_y  as if computed around line 48 and 49, when y != 0

As luck has it, both cases are the `masked_y` value.

The original issue #837 says negative zero was the problem.

WGSL says:

> Implementations may ignore the sign of a zero. That is, a zero with a positive sign may behave like a zero a with a negative sign, and vice versa.

Can we accommodate this in diffULP by patching up the sign when the magnitude is zero:

```js
if ( masked_a === 0 && masked_b === 0) { return 0; } 
if (masked_a === 0) { sign_a = sign_b; }
if (masked_b === 0) { sign_b = sign_a; }
```

dneto0 (changes requested):
Thanks for your patience.
This looks wrong.  See below.

src/webgpu/util/math.ts
ben-clayton: As discussed internally, I think we make the fix with just replacing these two lines:

```ts
const bits_a = masked_a - (subnormal_or_zero_a ? 0 : 0x7fffff);
const bits_b = masked_b - (subnormal_or_zero_b ? 0 : 0x7fffff);
```

with:

```ts
const bits_a = subnormal_or_zero_a ? 0 : (masked_a - 0x800000);
const bits_b = subnormal_or_zero_b ? 0 : (masked_b - 0x800000);
```

David would like more comments on this function to describe the subnormal behavior (it's just round-to-zero), and some more tests around these cases.

dneto0: Rationale:  The John Harrison definition of ULP(X) requires us to find the two floating point values a and b that straddle X.  When an implementation is allowed to flush subnormals to zero, then in then   {a,b} = {0, smallest_normal_number} (as sets).  And so ULP(0) and ULP(any subnormal)  will be equal to the smallest normal floating point number.

dneto0: Expanding this out:

Then diffULP(x,y) is "the number of ULP(min(|x|,|y|)) steps between x and y"

So diffULP(0,subnormal) = 0  (I think).  and diffULP(0,smallest_normal) = 1

zoddicus: I have updated the code and extended the tests. The change for `bits_*` that I actually used is this:

```
const bits_a = subnormal_or_zero_a ? 0 : (masked_a - 0x7fffff);
const bits_b = subnormal_or_zero_b ? 0 : (masked_b - 0x7fffff);
```

since using `0x800000` would push the smallest normal to 0, diffULP(0, smallest normal) was returning instead of 1.

src/webgpu/util/math.ts
```diff
@@ -46,9 +47,15 @@ export function diffULP(a: number, b: number): number {
   const sign_b = (u32_b & 0x80000000) !== 0;
   const masked_a = u32_a & 0x7fffffff;
   const masked_b = u32_b & 0x7fffffff;
+  const subnormal_or_zero_a = (u32_a & 0x7f800000) === 0;
+  const subnormal_or_zero_b = (u32_b & 0x7f800000) === 0;
+
+  // If the number is subnormal, then reduce it to 0 for ULP comparison.
+  const bits_a = subnormal_or_zero_a ? 0 : masked_a - 0x7fffff;
```
dneto0, line 54: I think I finally understand why the subtraction is the right thing to do here. It took me a *long* time to see it.

I think this needs another comment:

// If the number is normal then reduce its bits-representation so that we can pretend 
// that the subnormal numbers don't exist, for the purposes of counting ULP steps from
// zero (or any subnormal) to any of the normal numbers.

src/unittests/maths.spec.ts
```diff
@@ -38,8 +42,30 @@ g.test('test,math,diffULP')
...
+    { a: hexToF32(0x00000001), b: 0, ulp: 0 }, // Subnormals near 0
+    { a: 0, b: hexToF32(0x00000001), ulp: 0 }, // Subnormals near 0
+    { a: -hexToF32(0x00000001), b: 0, ulp: 0 }, // Subnormals near 0
+    { a: 0, b: -hexToF32(0x00000001), ulp: 0 }, // Subnormals near 0
+    { a: hexToF32(0x00000001), b: -hexToF32(0x00000001), ulp: 0 }, // Subnormals near 0
+    { a: -hexToF32(0x00000001), b: hexToF32(0x00000001), ulp: 0 }, // Subnormals near 0
+    { a: hexToF32(0x00000001), b: hexToF32(0x00800000), ulp: 1 }, // Normal/Subnormal boundary
+    { a: hexToF32(0x00800000), b: hexToF32(0x00000001), ulp: 1 }, // Normal/Subnormal boundary
+    { a: -hexToF32(0x00000001), b: -hexToF32(0x00800000), ulp: 1 }, // Normal/Subnormal boundary
+    { a: -hexToF32(0x00800000), b: -hexToF32(0x00000001), ulp: 1 }, // Normal/Subnormal boundary
```
dneto0, line 68: Please add cases that are slightly away from the edge cases.

I used these to convince myself (finally!) that the subtraction was the right thing to do:

    { a: hexToF32(0x00800001), b: hexToF32(0x00000000), ulp: 2 }, // Just-above-Normal/Subnormal boundary
    { a: hexToF32(0x00800001), b: hexToF32(0x00000001), ulp: 2 }, // Just-above-Normal/Subnormal boundary
    { a: hexToF32(0x00800005), b: hexToF32(0x00000001), ulp: 6 }, // Just-above-Normal/Subnormal boundary
    { a: hexToF32(0x00800005), b: hexToF32(0x00000111), ulp: 6 }, // Just-above-Normal/Subnormal boundary

dneto0 (changes requested):
I finally understand why the subtraction is the right thing to do, and not the shift.

But I'd be helped (and future readers too) if you added the suggested test cases and comments.

src/webgpu/util/math.ts
zoddicus: Done

src/unittests/maths.spec.ts
zoddicus: Done

dneto0 (approved):
Thank you!

PR #849, opened 2021-12-06
Roll to the lastest WebGPU type definitions

toji:
Won't be able to land as-is with the 0.1.8 types package. Requires https://github.com/gpuweb/types/pull/87 to land and be published first, at which point I'll update the packages.json to pull the correct version. Verified locally that it should passes the checks at that point.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "@types/offscreencanvas": "^2019.6.2",
     "@types/serve-index": "^1.9.1",
     "@typescript-eslint/parser": "^4.22.0",
-    "@webgpu/types": "0.1.7",
+    "@webgpu/types": "0.1.9",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -40,7 +40,7 @@
     "@types/offscreencanvas": "^2019.6.2",
     "@types/serve-index": "^1.9.1",
     "@typescript-eslint/parser": "^4.22.0",
-    "@webgpu/types": "^0.1.7",
+    "@webgpu/types": "^0.1.8",
```
kainino0x, line 43: Will need rebase to remove the carat `^`.

I just released 0.1.9 so you should be able to update now!

kainino0x (approved):

PR #851, opened 2021-12-07
Implement index buffer OOB validation test for drawIndexed/Indirect

jzm-intel:
Implement index buffer OOB validation test for drawIndexed and drawIndexedIndirect

Implement the index buffer OOB test to check calling drawIndexed with OOB index buffer access result in validation error when finish encoder, while drawIndexedIndirect will cause no CPU validation error.
This is a part of vertex access validation test for draw.

<hr>

### DIFF
--- a/src/webgpu/api/validation/encoding/cmds/render/draw.spec.ts
+++ b/src/webgpu/api/validation/encoding/cmds/render/draw.spec.ts
@@ -5,121 +5,49 @@ and parameters as expect.
 `;
 
 import { makeTestGroup } from '../../../../../../common/framework/test_group.js';
+import { unreachable } from '../../../../../../common/util/util.js';
 import { ValidationTest } from '../../../validation_test.js';
 
-// Class that indicate how to call a draw function
-class DrawCall {
-  test: ValidationTest;
-  drawType: 'draw' | 'drawIndexed' | 'drawIndirect' | 'drawIndexedIndirect' = 'draw';
-
-  // Draw
-  vertexCount: number = 4;
-  firstVertex: number = 0;
-
-  // DrawIndexed
-  indexCount: number = 4;
-  firstIndex: number = 0;
-  baseVertex: number = 0;
-
-  // Both Draw and DrawIndexed
-  instanceCount: number = 1;
-  firstInstance: number = 0;
-
-  constructor(test: ValidationTest) {
-    this.test = test;
-  }
-
-  callDraw(encoder: GPURenderEncoderBase) {
-    switch (this.drawType) {
-      case 'draw': {
-        encoder.draw(this.vertexCount, this.instanceCount, this.firstVertex, this.firstInstance);
-        break;
-      }
-      case 'drawIndexed': {
-        encoder.drawIndexed(
-          this.indexCount,
-          this.instanceCount,
-          this.firstIndex,
-          this.baseVertex,
-          this.firstInstance
-        );
-        break;
-      }
-      case 'drawIndirect': {
-        encoder.drawIndirect(this.generateIndirectBuffer(), 0);
-        break;
-      }
-      case 'drawIndexedIndirect': {
-        encoder.drawIndexedIndirect(this.generateIndexedIndirectBuffer(), 0);
-        break;
-      }
-    }
-  }
-
-  // Create an indirect buffer containing draw call values
-  private generateIndirectBuffer(): GPUBuffer {
-    const indirectArray = new Int32Array([
-      this.vertexCount,
-      this.instanceCount,
-      this.firstVertex,
-      this.firstInstance,
-    ]);
-    return this.test.makeBufferWithContents(indirectArray, GPUBufferUsage.INDIRECT);
-  }
-
-  // Create an indirect buffer containing indexed draw call values
-  private generateIndexedIndirectBuffer(): GPUBuffer {
-    const indirectArray = new Int32Array([
-      this.indexCount,
-      this.instanceCount,
-      this.firstIndex,
-      this.baseVertex,
-      this.firstInstance,
-    ]);
-    return this.test.makeBufferWithContents(indirectArray, GPUBufferUsage.INDIRECT);
-  }
+interface DrawIndexedParameter {
+  indexCount: number;
+  firstIndex?: number;
+  baseVertex?: number;
+  instanceCount?: number;
+  firstInstance?: number;
 }
 
 class F extends ValidationTest {
-  // Create a GPUBuffer that suitable for given index count and format
-  createIndexBufferWithElementCount(elementCount: number, indexFormat: GPUIndexFormat): GPUBuffer {
-    const elementSize = indexFormat === 'uint16' ? 2 : 4;
-    const bufferSize = elementCount * elementSize;
-    const desc: GPUBufferDescriptor = {
-      size: bufferSize,
-      usage: GPUBufferUsage.INDEX | GPUBufferUsage.COPY_DST,
-    };
-    return this.createBufferWithState('valid', desc);
-  }
-
-  // Create a render pipeline that has no vertex input
-  createNaiveRenderPipeline(): GPURenderPipeline {
-    const desc: GPURenderPipelineDescriptor = {
-      vertex: {
-        module: this.device.createShaderModule({
-          code: `
-          [[stage(vertex)]]
-          fn main() -> [[builtin(position)]] vec4<f32> {
-            return vec4<f32>(0.0, 1.0, 0.0, 1.0);
-          }
-          `,
-        }),
-        entryPoint: 'main',
-        buffers: [],
-      },
-      fragment: {
-        module: this.device.createShaderModule({
-          code: `
-          [[stage(fragment)]] fn main() -> [[location(0)]] vec4<f32> {
-            return vec4<f32>(1.0, 0.0, 0.0, 1.0);
-          }`,
-        }),
-        entryPoint: 'main',
-        targets: [{ format: 'rgba8unorm', writeMask: 0 }],
-      },
-      primitive: { topology: 'triangle-list' },
-    };
-    return this.device.createRenderPipeline(desc);
+  callDrawIndexed(
+    encoder: GPURenderEncoderBase,
+    drawType: 'drawIndexed' | 'drawIndexedIndirect',
+    param: DrawIndexedParameter
+  ) {
+    switch (drawType) {
+      case 'drawIndexed': {
+        encoder.drawIndexed(
+          param.indexCount,
+          param.instanceCount || 1,
+          param.firstIndex || 0,
+          param.baseVertex || 0,
+          param.firstInstance || 0
+        );
+        break;
+      }
+      case 'drawIndexedIndirect': {
+        const indirectArray = new Int32Array([
+          param.indexCount,
+          param.instanceCount || 1,
+          param.firstIndex || 0,
+          param.baseVertex || 0,
+          param.firstInstance || 0,
+        ]);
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/encoding/cmds/render/draw.spec.ts
```diff
@@ -7,8 +7,120 @@ and parameters as expect.
...
+    const bufferSize = elementCount * elementSize;
+    const desc: GPUBufferDescriptor = {
+      size: bufferSize,
+      usage: GPUBufferUsage.INDEX | GPUBufferUsage.COPY_DST,
+    };
+    return this.createBufferWithState('valid', desc);
+  }
+
+  // Create a render pipeline that has no vertex input
+  createNaiveRenderPipeline(): GPURenderPipeline {
```
kainino0x, line 96: Can you use ValidationTest.createNoOpRenderPipeline here?
https://gpuweb.github.io/cts/docs/tsdoc/classes/webgpu_api_validation_validation_test.validationtest.html#createnooprenderpipeline

src/webgpu/api/validation/encoding/cmds/render/draw.spec.ts
```diff
@@ -7,8 +7,120 @@ and parameters as expect.
...
+      this.baseVertex,
+      this.firstInstance,
+    ]);
+    return this.test.makeBufferWithContents(indirectArray, GPUBufferUsage.INDIRECT);
+  }
+}
+
 class F extends ValidationTest {
-  // TODO: Implement the helper functions
+  // Create a GPUBuffer that suitable for given index count and format
```
kainino0x, line 84: For documentation comments please use tsdoc/jsdoc syntax as in other files: `/** blah blah */`

src/webgpu/api/validation/encoding/cmds/render/draw.spec.ts
```diff
@@ -7,8 +7,120 @@ and parameters as expect.
...
+  createIndexBufferWithElementCount(elementCount: number, indexFormat: GPUIndexFormat): GPUBuffer {
+    const elementSize = indexFormat === 'uint16' ? 2 : 4;
+    const bufferSize = elementCount * elementSize;
+    const desc: GPUBufferDescriptor = {
+      size: bufferSize,
+      usage: GPUBufferUsage.INDEX | GPUBufferUsage.COPY_DST,
+    };
+    return this.createBufferWithState('valid', desc);
```
kainino0x, line 92: I think you can just use `this.device.createBuffer`.

src/webgpu/api/validation/encoding/cmds/render/draw.spec.ts
```diff
@@ -35,7 +147,53 @@ drawIndexedIndirect as it is GPU-validated.
...
+      bindingSizeInElements,
+      bufferSizeInElements,
+      drawIndexCount,
+      drawType,
+      useBundle,
+    } = t.params;
+
+    const indexBuffer = t.createIndexBufferWithElementCount(bufferSizeInElements, indexFormat);
+    const indexElementSize = indexFormat === 'uint16' ? 2 : 4;
+    const bindingSize = bindingSizeInElements * indexElementSize;
```
kainino0x, line 173: `createIndexBufferWithElementCount` helper doesn't seem to be helping here - half of the function is duplicated out here anyway. I think it should be inlined.

src/webgpu/api/validation/encoding/cmds/render/draw.spec.ts
```diff
@@ -7,8 +7,120 @@ and parameters as expect.
 import { makeTestGroup } from '../../../../../../common/framework/test_group.js';
 import { ValidationTest } from '../../../validation_test.js';
 
+// Class that indicate how to call a draw function
+class DrawCall {
```
kainino0x, line 11: This class is pretty big for pretty simple behavior, and it isn't well-encapsulated so it's not very useful to write as a class (as seen by the fact that the fields aren't even initialized from the constructor, this class is pretty much being used as an `interface` or simple `{ key: value }` type).

Please consider either:
- Moving the other definition of `DrawCall` (from `robust_access_vertex.spec.ts`) into a common file, if that would work well, or
- (probably better) just have a standalone helper that takes only the necessary arguments, something like:

```ts
function callDraw(
  t: GPUTest,
  encoder: GPURenderEncoderBase,
  drawType: 'drawIndexed' | 'drawIndexedIndirect',
  { indexCount }: { indexCount: number }
) {
```

The `generateIndexedIndirectBuffer` helper is sufficiently simple that I think it could be inline in this helper, but you could also consider deduplicating it with the one in `robust_access_vertex.spec.ts`, and moving both of those helpers to `GPUTest`.

src/webgpu/api/validation/encoding/cmds/render/draw.spec.ts
```diff
@@ -7,8 +7,120 @@ and parameters as expect.
...
+          this.firstIndex,
+          this.baseVertex,
+          this.firstInstance
+        );
+        break;
+      }
+      case 'drawIndirect': {
+        encoder.drawIndirect(this.generateIndirectBuffer(), 0);
+        break;
+      }
```
kainino0x, line 51: The 'draw' and 'drawIndirect' cases aren't used AFAICT. And therefore generateIndirectBuffer is also not used.

src/webgpu/api/validation/encoding/cmds/render/draw.spec.ts
```diff
@@ -35,7 +147,53 @@ drawIndexedIndirect as it is GPU-validated.
...
+    const commandBufferMaker = t.createEncoder(useBundle ? 'render bundle' : 'render pass');
+    const renderEncoder = commandBufferMaker.encoder;
+
+    renderEncoder.setIndexBuffer(indexBuffer, indexFormat, 0, bindingSize);
+
+    renderEncoder.setPipeline(renderPipeline);
+
+    drawCall.callDraw(renderEncoder);
+
+    commandBufferMaker.validateFinish(isFinishSuccess);
```
kainino0x, line 195: I think you should use validateFinishAndSubmit so we can also just be sure no validation fails in submit().

kainino0x (commented):
Some comments below. Please also complete the "Author checklist for test code/plans" section.

src/webgpu/api/validation/encoding/cmds/render/draw.spec.ts
jzm-intel: Done.

src/webgpu/api/validation/encoding/cmds/render/draw.spec.ts
jzm-intel: Done.

src/webgpu/api/validation/encoding/cmds/render/draw.spec.ts
jzm-intel: Done.

src/webgpu/api/validation/encoding/cmds/render/draw.spec.ts
jzm-intel: Currently I have a callDrawIndexed to deal with drawIndexed and drawIndexedIndirect, and in the next step I will write something like callDraw to deal with draw and drawIndirect.  Test for vertex buffer OOB may need to test all these four draw calls.

src/webgpu/api/validation/encoding/cmds/render/draw.spec.ts
jzm-intel: Done.

src/webgpu/api/validation/encoding/cmds/render/draw.spec.ts
jzm-intel: Done. Sorry for forgetting the success case.

src/webgpu/api/validation/encoding/cmds/render/draw.spec.ts
```diff
@@ -7,120 +7,45 @@ and parameters as expect.
...
-        break;
-      }
       case 'drawIndexedIndirect': {
-        encoder.drawIndexedIndirect(this.generateIndexedIndirectBuffer(), 0);
+        const indirectArray = new Int32Array([
+          param.indexCount,
+          param.instanceCount || 1,
+          param.firstIndex || 0,
+          param.baseVertex || 0,
+          param.firstInstance || 0,
```
kainino0x, line 41: same

src/webgpu/api/validation/encoding/cmds/render/draw.spec.ts
```diff
@@ -7,120 +7,45 @@ and parameters as expect.
...
-
-  constructor(test: ValidationTest) {
-    this.test = test;
-  }
+interface DrawIndexedParameter {
+  indexCount: number;
+  firstIndex?: number;
+  baseVertex?: number;
+  instanceCount?: number;
+  firstInstance?: number;
```
kainino0x, line 15: uNit: please order these the same way as in the drawIndexed call.

src/webgpu/api/validation/encoding/cmds/render/draw.spec.ts
```diff
@@ -7,120 +7,45 @@ and parameters as expect.
...
-          this.indexCount,
-          this.instanceCount,
-          this.firstIndex,
-          this.baseVertex,
-          this.firstInstance
+          param.indexCount,
+          param.instanceCount || 1,
+          param.firstIndex || 0,
+          param.baseVertex || 0,
+          param.firstInstance || 0
```
kainino0x, line 31: Use `??` instead of `||` because, e.g. if instanceCount is 0, `0 || 1` is `1`, but `0 ?? 1` is `0` which is more correct.

src/webgpu/api/validation/encoding/cmds/render/draw.spec.ts
```diff
@@ -7,120 +7,45 @@ and parameters as expect.
...
+}
 
-  callDraw(encoder: GPURenderEncoderBase) {
-    switch (this.drawType) {
-      case 'draw': {
-        encoder.draw(this.vertexCount, this.instanceCount, this.firstVertex, this.firstInstance);
-        break;
-      }
+class F extends ValidationTest {
+  callDrawIndexed(
```
kainino0x, line 19: nit: I (personally) think this should be a standalone function (`function callDrawIndexed`) rather than a member of the fixture. It makes it a bit lighter weight and if you do this you won't have to customize the fixture.

kainino0x (approved):
LGTM with a few small comments, merge when ready

src/webgpu/api/validation/encoding/cmds/render/draw.spec.ts
kainino0x: > Use `??` instead of `||` because, e.g. if instanceCount is 0, `0 || 1` is `1`, but `0 ?? 1` is `0` which is more correct.

jzm-intel: Done.

src/webgpu/api/validation/encoding/cmds/render/draw.spec.ts
jzm-intel: Done.

src/webgpu/api/validation/encoding/cmds/render/draw.spec.ts
jzm-intel: Done.

src/webgpu/api/validation/encoding/cmds/render/draw.spec.ts
jzm-intel: Done.

jzm-intel:
I am going to merging this PR, thanks for reviewing!

PR #856, opened 2021-12-10
Fix handling of magic values in parseQuery.

### LINKED ISSUE #855: We need additional unittests for parseQuery
ParseQuery is never tested on its own, but always through more of integration tests etc. We should test it better because it is something subtle and can break easily.

Kangz:
JSON.parse's reviver modifies the key, not the value like
parseParamValueReviver was doing. Also this commit simplifies a bit the
handling of -0 in json_param_value.ts.

Adds tests for magic values and also add parseQuery tests because it had
only indirect test coverage.

Fixes #855

<hr>

### DIFF
--- a/src/common/internal/query/json_param_value.ts
+++ b/src/common/internal/query/json_param_value.ts
@@ -8,11 +8,17 @@ const jsNaNMagicValue = '_nan_';
 const jsPositiveInfinityMagicValue = '_posinfinity_';
 const jsNegativeInfinityMagicValue = '_neginfinity_';
 
+// -0 needs to be handled separately, because -0 === +0 returns true. Not
+// special casing +0/0, since it behaves intuitively. Assuming that if -0 is
+// being used, the differentiation from +0 is desired.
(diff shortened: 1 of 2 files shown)


src/common/internal/query/json_param_value.ts
```diff
@@ -68,12 +71,8 @@ export function stringifyParamValueUniquely(value: JSONWithUndefined): string {
 // 'any' is part of the JSON.parse reviver interface, so cannot be avoided.
 // eslint-disable-next-line  @typescript-eslint/no-explicit-any
 function parseParamValueReviver(k: string, v: any): any {
-  if (k === jsNegativeZeroMagicValue) {
-    return -0;
-  }
-
-  if (fromStringMagicValue.has(k)) {
-    return fromStringMagicValue.get(k);
+  if (fromStringMagicValue.has(v)) {
```
Kangz, line 74: The fix is the switch from `k` to `v` here and in the return.

ben-clayton (approved):

Kangz:
Thanks!

PR #859, opened 2021-12-23
Remove the parameter 'opts' in makeBufferWithContents

Jiawei-Shao:
This patch removes the parameter 'opts' in makeBufferWithContents to
indicate whether the buffer size should be aligned to 4 because in
this function the returned buffer is always created with
'mappedAtCreation' set to true, so the size of the returned buffer
should always be a multiple of 4.

<hr>

### DIFF
--- a/src/webgpu/api/operation/command_buffer/image_copy.spec.ts
+++ b/src/webgpu/api/operation/command_buffer/image_copy.spec.ts
@@ -351,10 +351,7 @@ class ImageCopyTest extends GPUTest {
         break;
       }
       case 'CopyB2T': {
-        const buffer = this.makeBufferWithContents(partialData, GPUBufferUsage.COPY_SRC, {
-          padToMultipleOf4: true,
-        });
-
+        const buffer = this.makeBufferWithContents(partialData, GPUBufferUsage.COPY_SRC);
         const encoder = this.device.createCommandEncoder();
         encoder.copyBufferToTexture(
           { buffer, ...appliedDataLayout },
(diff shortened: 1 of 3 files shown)


Kangz (approved):

kainino0x (approved):
This might change the behavior of a lot of tests, because almost every usage of makeBufferWithContents doesn't specify padToMultipleOf4. In order to make this change we need to audit all of them and make sure none of them will break subtly with this change.

(If Dawn already implements https://github.com/gpuweb/gpuweb/pull/899 then any _passing_ tests are probably fine, but we could have broken tests that this would introduce bugs into.)

I looked at the call sites. Most of them use `*32Array`, so those are fine. To find the rest, I changed the signature of GPUTest.makeBufferWithContents to only take `Int32Array | Uint32Array | Float32Array` and looked at the call sites that errored.
- api,operation,command_buffer,copyBufferToBuffer: shouldn't rely on validation so OK
- api,operation,command_buffer,image_copy copyPartialTextureToBufferAndCheckContents: complicated but shouldn't rely on validation or robustness so OK
- api,operation,queue,writeBuffer: looks OK
- api,operation,rendering,draw: Actually only Uint32Array or Float32Array
- api,operation,sampling,anisotropy: Only used to initialize texture so OK
- api,operation,vertex_state,index_format: Used to create an index buffer, and not used  to test drawIndirect robustness, so OK
- shader/execution/builtin/builtin.ts: looks OK
- shader,execution,robust_buffer_access: fixed size 512, OK

PR #863, opened 2021-12-31
Improve PR template

kainino0x:
Trying to make it slightly more effective. We also need to start enforcing these more.


Kangz (approved):
LGTM. Agreed that we should consistently use this template to make sure the CTS stays in a good state.

PR #869, opened 2022-01-06
Implement `fract` tests

zoddicus:
<hr>

### DIFF
--- a/src/unittests/maths.spec.ts
+++ b/src/unittests/maths.spec.ts
@@ -32,7 +32,7 @@ function hexToFloat64(h32: number, l32: number): number {
 }
 
 g.test('test,math,diffULP')
-  .paramsSimple<DiffULPCase>([
+  .paramsSubcasesOnly<DiffULPCase>([
     { a: 0, b: 0, ulp: 0 },
     { a: 1, b: 2, ulp: 2 ** 23 }, // Single exponent step
     { a: 2, b: 1, ulp: 2 ** 23 }, // Single exponent step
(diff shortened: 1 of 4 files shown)


src/webgpu/shader/execution/builtin/fract.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+
+      // Min and Max f32
+      { input: f32Bits(kBit.f32.positive.min), expected: f32Bits(kBit.f32.positive.min) },
+      { input: f32Bits(kBit.f32.positive.max), expected: f32(0) },
+      { input: f32Bits(kBit.f32.negative.max), expected: f32Bits(0x3f7fffff) },
+      { input: f32Bits(kBit.f32.negative.min), expected: f32(0) },
+
+      // Subnormal f32
+      // prettier-ignore
+      { input: f32Bits(kBit.f32.subnormal.positive.max), expected: f32Bits(kBit.f32.subnormal.positive.max) },
```
ben-clayton, line 57: https://www.w3.org/TR/WGSL/#floating-point-evaluation states that:

> Implementations may flush denormalized value on the input and/or output of any operation listed in [§ 12.5.1 Floating Point Accuracy](https://www.w3.org/TR/WGSL/#floating-point-accuracy).
> Other operations are required to preserve denormalized numbers.

I assume "any operation listed in [§ 12.5.1 Floating Point Accuracy](https://www.w3.org/TR/WGSL/#floating-point-accuracy)." includes the contents of the table _"Accuracy of built-in functions"_, and so `floor` is allowed to flush denormalized values to zero. @dneto0 or @alan-baker to confirm†.

I don't see any handling of denormals in `correctlyRounded()`, so I assume that this isn't handled by the tests. You could change `correctlyRounded()`, or change `expected` to `[f32Bits(kBit.f32.subnormal.positive.max), f32(0)]`

---
† Also: would it be clearer to change:
> Implementations may flush denormalized value...

to

> Implementations may **flush-to-zero** denormalized value...

As not everyone reading may know what the value is flushed to?

alan-baker: Correct, floor can flush denormalized values.

zoddicus: I should probably update correctlyRounded to flush denormals, so that subsequent tests don't have to be away of this subtleness.

Related to that, currently nextAfter, which is used be correctlyRounded(0) will return the subnormal nearest to 0 that is in the appropriate direction. Is it correct to assume that nextAfter should also flush subnormals? thus this should be changed to the nearest normal to 0, otherwise we will get into a potential situation of getting trapped at 0.

zoddicus:
I have added support for flush to zero to correctlyRounded

src/webgpu/shader/execution/builtin/fract.spec.ts
```diff
@@ -49,15 +49,17 @@ T is f32 or vecN<f32> fract(e: T ) -> T Returns the fractional bits of e (e.g. e
       // Min and Max f32
       { input: f32Bits(kBit.f32.positive.min), expected: f32Bits(kBit.f32.positive.min) },
       { input: f32Bits(kBit.f32.positive.max), expected: f32(0) },
-      { input: f32Bits(kBit.f32.negative.max), expected: f32Bits(0x3f7fffff) },
+      { input: f32Bits(kBit.f32.negative.max), expected: anyOf(f32Bits(0x3f7fffff), f32(1)) },
```
zoddicus, line 52: Here and below running the tests on NVIDIA vs SwiftShader they disagree on 1 vs very close to 1. I think what is occurring is that if the calculation is internally done in f32, i.e. rounding/flushing each step, you end up with one value, but all of the math is done in a higher precision, and then converted at the end, you end up with a different value.

I am not sure what we want to about this, or if this is fine, so I encoded the most permissive form of the check for now.

src/unittests/maths.spec.ts
```diff
@@ -210,68 +273,68 @@ g.test('test,math,correctlyRounded')
     // Zeros
     { test_val: f32Bits(kBit.f32.positive.zero), target: 0, is_correct: true },
     { test_val: f32Bits(kBit.f32.negative.zero), target: 0, is_correct: true },
-    { test_val: f32Bits(kBit.f32.subnormal.positive.min), target: 0, is_correct: false },
-    { test_val: f32Bits(kBit.f32.subnormal.positive.max), target: 0, is_correct: false },
-    { test_val: f32Bits(kBit.f32.subnormal.negative.max), target: 0, is_correct: false },
-    { test_val: f32Bits(kBit.f32.subnormal.negative.min), target: 0, is_correct: false },
+    { test_val: f32Bits(kBit.f32.subnormal.positive.min), target: 0, is_correct: true },
```
ben-clayton, line 276: It's possible that there might be tests we'll need to write that'll want to use `correctlyRounded` without thet FTZ behaviour.
For this reason, I'd prefer it if exposed this as a parameter flag (which can happily default to testing with and without FTZ), but make sure that we have tests for both behaviours. I think you can just copy the parent-change test's as another `g.test('test,math,correctlyRoundedNoFlushToZero)` set.

src/unittests/maths.spec.ts
```diff
@@ -32,7 +32,7 @@ function hexToFloat64(h32: number, l32: number): number {
 }
 
 g.test('test,math,diffULP')
-  .paramsSimple<DiffULPCase>([
+  .paramsSubcasesOnly<DiffULPCase>([
```
ben-clayton, line 35: It's not immediately clear to me why this set of changes were made.

ben-clayton (commented):
Implementation looks good to me, although it's fiddly logic and I'd like a second set of eyes on this.

One request to test against the non-FTZ behaviour for `correctlyRounded`. I think we might need this one day.

src/unittests/maths.spec.ts
zoddicus: I have added control flags and additional testing

zoddicus:
@ben-clayton @dneto0 PTAL

src/unittests/maths.spec.ts
```diff
@@ -94,65 +94,128 @@ g.test('test,math,diffULP')
...
-  .paramsSimple<nextAfterCase>([
+  .paramsSubcasesOnly<nextAfterCase>([
     // Edge Cases
-    { val: Number.NaN, dir: true, result: f32Bits(0x7fffffff) },
-    { val: Number.NaN, dir: false, result: f32Bits(0x7fffffff) },
-    { val: Number.POSITIVE_INFINITY, dir: true, result: f32Bits(kBit.f32.infinity.positive) },
-    { val: Number.POSITIVE_INFINITY, dir: false, result: f32Bits(kBit.f32.infinity.positive) },
-    { val: Number.NEGATIVE_INFINITY, dir: true, result: f32Bits(kBit.f32.infinity.negative) },
-    { val: Number.NEGATIVE_INFINITY, dir: false, result: f32Bits(kBit.f32.infinity.negative) },
+    { val: Number.NaN, dir: true, flush: true, result: f32Bits(0x7fffffff) },
```
ben-clayton, line 104: I think it would have been cleaner to have the flush cases split into two `g.test()` blocks, but meh.

ben-clayton (approved):
LGTM - I'd still like someone else to double check. @dneto0 ?

src/unittests/maths.spec.ts
zoddicus: Done

src/webgpu/util/math.ts
```diff
@@ -66,14 +66,46 @@ export function diffULP(a: number, b: number): number {
...
+  const u32_val = new Uint32Array(new Float32Array([val.value.valueOf() as number]).buffer)[0];
+  return (u32_val & 0x7f800000) === 0 ? f32(0) : val;
+}
+
 /**
  * @returns the next single precision floating point value after |val|,
  * towards +inf if |dir| is true, otherwise towards -inf.
- * For -/+0 the nextAfter will be the closest subnormal in the correct
- * direction, since -0 === +0.
+ * If |flush_| is true, all subnormal values will be flushed to 0,
```
dneto0, line 101: nit:   flush_   ->  flush

src/webgpu/shader/execution/builtin/fract.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+  .params(u =>
+    u
+      .combine('storageClass', ['uniform', 'storage_r', 'storage_rw'] as const)
+      .combine('vectorize', [undefined, 2, 3, 4] as const)
+  )
+  .fn(async t => {
+    const cfg: Config = t.params;
+    cfg.cmpFloats = correctlyRoundedThreshold();
+    run(t, 'fract', [TypeF32], TypeF32, cfg, [
+      // Positive numbers
```
dneto0, line 31: Please add the case for 0.

dneto0 (changes requested):
Looking really good.

src/webgpu/util/math.ts
zoddicus: Done

src/webgpu/shader/execution/builtin/fract.spec.ts
zoddicus: Done

dneto0 (approved):
Thanks!

Issue #882, opened 2022-01-13
val: state,device_lost


kainino0x:
Tests of behavior while the device is lost.

It's probably ideal to try to test this thoroughly (for full coverage across the whole API surface), but it's a big surface. We need at least basic tests.

May want to distribute this across the test suite like device_mismatch. It's also possible these can be folded together with some other tests (like the device_mismatched, destroyed object, or other tests).

lokokung:
Can probably close this now since tests have since been added for operations after the device is lost.

PR #932, opened 2022-01-19
Adds some device.destroy and lost tests.

lokokung:
- Changes the tests to save the entire lost info interface instead of just the message so that we can identify the reason for device lost and suppress failures when tests expect devices to be destroyed.

- Tested in local chromium build with changes:
   - https://dawn-review.googlesource.com/c/dawn/+/77200
   - https://chromium-review.googlesource.com/c/chromium/src/+/3388269

Issue: https://github.com/gpuweb/cts/issues/882

### DIFF
--- a/src/webgpu/api/operation/device/lost.spec.ts
+++ b/src/webgpu/api/operation/device/lost.spec.ts
@@ -2,28 +2,24 @@ export const description = `
 Tests for GPUDevice.lost.
 `;
 
+import { Fixture } from '../../../../common/framework/fixture.js';
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
 import { attemptGarbageCollection } from '../../../../common/util/collect_garbage.js';
 import { getGPU } from '../../../../common/util/navigator_gpu.js';
 import { assert, raceWithRejectOnTimeout } from '../../../../common/util/util.js';
-import { GPUTest } from '../../../gpu_test.js';
 
-class DeviceLostTests extends GPUTest {
+class DeviceLostTests extends Fixture {
   // Default timeout for waiting for device lost is 2 seconds.
   readonly kDeviceLostTimeoutMS = 2000;
 
-  getDeviceLostWithTimeout(device: GPUDevice = this.device): Promise<GPUDeviceLostInfo> {
-    return raceWithRejectOnTimeout(
-      this.device.lost,
-      this.kDeviceLostTimeoutMS,
-      'device was not lost'
-    );
+  getDeviceLostWithTimeout(lost: Promise<GPUDeviceLostInfo>): Promise<GPUDeviceLostInfo> {
+    return raceWithRejectOnTimeout(lost, this.kDeviceLostTimeoutMS, 'device was not lost');
   }
 
-  expectDeviceDestroyed(): void {
+  expectDeviceDestroyed(device: GPUDevice): void {
     this.eventualAsyncExpectation(async niceStack => {
       try {
-        const lost = await this.getDeviceLostWithTimeout();
+        const lost = await this.getDeviceLostWithTimeout(device.lost);
         this.expect(lost.reason === 'destroyed', 'device was lost from destroy');
       } catch (ex) {
         niceStack.message = 'device was not lost';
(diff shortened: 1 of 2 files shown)


src/webgpu/util/device_pool.ts
```diff
@@ -243,10 +250,11 @@ type DeviceHolderState = 'free' | 'reserved' | 'acquired';
 class DeviceHolder implements DeviceProvider {
   readonly device: GPUDevice;
   state: DeviceHolderState = 'free';
-  lostReason?: string; // initially undefined; becomes set when the device is lost
+  // initially undefined; becomes set when the device is lost
+  lost?: GPUDeviceLostInfo;
```
kainino0x, line 254: nit: maybe `lostInfo` just to disambiguate from `GPUDevice.lost` which is a promise

src/webgpu/api/operation/device/lost.spec.ts
```diff
@@ -2,24 +2,82 @@ export const description = `
...
   .desc(
     `'lost' is never resolved by GPUDevice being garbage collected (with attemptGarbageCollection).`
   )
-  .unimplemented();
+  .fn(async t => {
+    // Create a new device for this because it is hard to lose all refs to the test fixture device.
+    const adapter = await getGPU().requestAdapter();
+    assert(adapter !== null);
+    let device: GPUDevice | undefined = await adapter.requestDevice();
+    assert(device !== null);
```
kainino0x, line 47: It shouldn't be possible for device to be null (types affirm this because there's no `null` in `GPUDevice | undefined`)
```suggestion
```

src/webgpu/api/operation/device/lost.spec.ts
```diff
@@ -2,24 +2,82 @@ export const description = `
...
-  .unimplemented();
+  .desc(`'lost' provides the same Promise and GPUDeviceLostInfo objects each time it's accessed.`)
+  .fn(async t => {
+    // The promises should be the same promise object.
+    const lostPromise1 = t.device.lost;
+    const lostPromise2 = t.device.lost;
+    t.expect(lostPromise1 === lostPromise2);
+
+    // The results should also be the same result object.
+    t.device.destroy();
```
kainino0x, line 71: Would be good to also access device.lost after destroy() and make sure it's still the same.
And then again after the first .lost resolved.

src/webgpu/api/operation/device/lost.spec.ts
```diff
@@ -2,24 +2,82 @@ export const description = `
...
+      lostPromise1,
+      t.kDeviceLostTimeoutMS,
+      'device was not lost'
+    );
+    const lost2 = await raceWithRejectOnTimeout(
+      lostPromise2,
+      t.kDeviceLostTimeoutMS,
+      'device was not lost'
+    );
+    t.expect(lost1 === lost2);
```
kainino0x, line 82: As a side note, if the Promise object is the same, then its resolved value is also guaranteed to stay the same forever. It doesn't hurt to test this once though.

src/webgpu/api/operation/device/lost.spec.ts
```diff
@@ -2,24 +2,82 @@ export const description = `
...
+    device = undefined;
+    await attemptGarbageCollection();
+  });
 
 g.test('lost_on_destroy')
   .desc(`'lost' is resolved, with reason='destroyed', on GPUDevice.destroy().`)
-  .unimplemented();
+  .fn(async t => {
+    t.expectDeviceDestroyed();
+    t.device.destroy();
```
kainino0x, line 59: I think it's strictly speaking OK to destroy the device provided by the GPUTest, but for the sake of locality in these tests I think it would be better for the test to own its own GPUDevice (and not use the GPUTest fixture at all). This way the DevicePool can also just assume tests never destroy devices.

src/webgpu/util/device_pool.ts
```diff
@@ -60,10 +60,17 @@ export class DevicePool {
...
+      if (lost !== undefined) {
+        if (lost.reason !== 'destroyed') {
+          // Fail the current test.
+          unreachable(`Device was unexpectedly lost: ${lost.message}`);
+        }
+        if (holder === this.defaultHolder) {
+          this.defaultHolder = 'uninitialized';
+        } else {
+          this.nonDefaultHolders.deleteByDevice(holder.device);
+        }
```
kainino0x, line 73: see above, this can basically just assert `lostInfo === undefined` (but with different messages like "Device was unexpectedly destroyed during the test")

src/webgpu/util/device_pool.ts
```diff
@@ -60,10 +60,17 @@ export class DevicePool {
...
       }
     } catch (ex) {
       // Any error that isn't explicitly TestFailedButDeviceReusable forces a new device to be
```
kainino0x, line 76: Below on line 84 there's a MAINTENANCE_TODO to device.destroy() after a device is discarded, to clean up resources. I think we should add that now.

But guard it behind `if ('destroy' in device)` just so we don't break the entire CTS for implementations that haven't implemented device.destroy() yet.

src/webgpu/api/operation/device/lost.spec.ts
```diff
@@ -2,24 +2,82 @@ export const description = `
...
+    // Create a new device for this because it is hard to lose all refs to the test fixture device.
+    const adapter = await getGPU().requestAdapter();
+    assert(adapter !== null);
+    let device: GPUDevice | undefined = await adapter.requestDevice();
+    assert(device !== null);
+
+    t.shouldReject('Error', t.getDeviceLostWithTimeout(device), 'device was unexpectedly lost');
+
+    device = undefined;
+    await attemptGarbageCollection();
```
kainino0x, line 52: Tracking `device` to make sure it actually goes out of scope is a little difficult, since it gets passed getDeviceLostWithTimeout - and the escape analysis of JS VMs might have the same problem (and keep it alive even though it doesn't need to). To try to make as much stuff go directly go out of scope, how about tightly constraining them to a function scope something like:

```ts
const lost = await (async () => {
  const adapter = await getGPU().requestAdapter();
  assert(adapter !== null);
  return (await adapter.requestDevice()).lost;
})();
```

It's inherently impossible to guarantee things get garbage collected in JS but I think this at least makes it more likely.

It would be interesting to try to detect GC using WeakRef - this could just be a TODO for now if you don't want to get into it. We could have another copy of this test (that skips itself if WeakRef isn't available) that uses WeakRef to detect if the GPUDevice got GC'd, and produces a warning if it didn't. Right now warnings don't count as failure, but eventually I think we will start running a warning-as-error mode in Chromium.

kainino0x (commented):
Overall LGTM, had several suggestions :)

src/webgpu/api/operation/device/lost.spec.ts
kainino0x:
```ts
const { lost } = await (async () => {
  const adapter = await getGPU().requestAdapter();
  assert(adapter !== null);
  const lost = (await adapter.requestDevice()).lost;
  return { lost };
})();
```

src/webgpu/api/operation/device/lost.spec.ts
lokokung: Done

src/webgpu/api/operation/device/lost.spec.ts
lokokung: Done

src/webgpu/api/operation/device/lost.spec.ts
lokokung: Done

src/webgpu/api/operation/device/lost.spec.ts
lokokung: Done

src/webgpu/api/operation/device/lost.spec.ts
lokokung: Ack.

src/webgpu/util/device_pool.ts
lokokung: Done

src/webgpu/util/device_pool.ts
lokokung: Done

src/webgpu/util/device_pool.ts
lokokung: Done

src/webgpu/api/operation/device/lost.spec.ts
```diff
@@ -40,44 +36,49 @@ g.test('not_lost_on_gc')
...
-    );
-    t.expect(lost1 === lost2);
+    device.destroy();
+    const lostPromise3 = device.lost;
+    t.expect(lostPromise1 === lostPromise3);
+
+    const lost1 = await t.getDeviceLostWithTimeout(lostPromise1);
+    const lost2 = await t.getDeviceLostWithTimeout(lostPromise2);
+    const lost3 = await t.getDeviceLostWithTimeout(lostPromise3);
+    const lost4 = await t.getDeviceLostWithTimeout(device.lost);
```
kainino0x, line 82:
```suggestion
    // Promise object should still be the same after we've been notified about device loss.
    const lostPromise4 = device.lost;
    t.expect(lostPromise1 === lostPromise4);
    const lost4 = await t.getDeviceLostWithTimeout(lostPromise4);
```

kainino0x (approved):

PR #950, opened 2022-01-24
Build out/{stress,manual}/listing.js so they can be run from online

kainino0x:
e.g. https://gpuweb.github.io/cts/standalone/?runnow=0&worker=0&debug=0&q=stress:*
or local non-dev builds.


lokokung (approved):
Ah! This explains it :D (FWIW it works locally, just not the one on the web). Thanks for the quick fix!

PR #953, opened 2022-01-26
Implementing `min` tests

zoddicus:
These tests are for the builtin float function `min`.
<hr>

### DIFF
--- a/src/webgpu/util/math.ts
+++ b/src/webgpu/util/math.ts
@@ -87,12 +87,24 @@ function flushSubnormalBits(val: number): number {
  * @returns 0 if |val| is a subnormal f32 number, otherwise returns |val|
  */
 function flushSubnormalScalar(val: Scalar): Scalar {
+  return isSubnormalScalar(val) ? f32(0) : val;
+}
+
+/**
+ * @returns true if |val| is a subnormal f32 number, otherwise returns false
+ * 0 is considered a non-subnormal number by this function.
+ */
+export function isSubnormalScalar(val: Scalar): boolean {
   if (val.type.kind !== 'f32') {
-    return val;
+    return false;
+  }
+
+  if (val === f32(0)) {
+    return false;
   }
 
   const u32_val = new Uint32Array(new Float32Array([val.value.valueOf() as number]).buffer)[0];
-  return (u32_val & 0x7f800000) === 0 ? f32(0) : val;
+  return (u32_val & 0x7f800000) === 0;
 }
 
 /**
--- a/src/webgpu/shader/execution/builtin/float_built_functions.spec.ts
+++ b/src/webgpu/shader/execution/builtin/float_built_functions.spec.ts
@@ -292,21 +292,6 @@ https://github.com/gpuweb/cts/blob/main/docs/plan_autogen.md
   .params(u => u.combine('placeHolder1', ['placeHolder2', 'placeHolder3']))
   .unimplemented();
 
-g.test('float_builtin_functions,min')
-  .uniqueId('53efc46faad0f380')
-  .specURL('https://www.w3.org/TR/2021/WD-WGSL-20210929/#float-builtin-functions')
-  .desc(
-    `
-min:
-T is f32 or vecN<f32> min(e1: T ,e2: T ) -> T Returns e2 if e2 is less than e1, and e1 otherwise. If one operand is a NaN, the other is returned. If both operands are NaNs, a NaN is returned. Component-wise when T is a vector. (GLSLstd450NMin)
-
-Please read the following guidelines before contributing:
-https://github.com/gpuweb/cts/blob/main/docs/plan_autogen.md
-`
-  )
-  .params(u => u.combine('placeHolder1', ['placeHolder2', 'placeHolder3']))
-  .unimplemented();
-
 g.test('float_builtin_functions,mix_all_same_type_operands')
   .uniqueId('f17861e71386bb59')
   .specURL('https://www.w3.org/TR/2021/WD-WGSL-20210929/#float-builtin-functions')
--- a/src/webgpu/shader/execution/builtin/min.spec.ts
+++ b/src/webgpu/shader/execution/builtin/min.spec.ts
@@ -4,9 +4,20 @@ Execution Tests for the 'min' builtin function
 
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
 import { GPUTest } from '../../../gpu_test.js';
-import { i32, i32Bits, TypeI32, TypeU32, u32 } from '../../../util/conversion.js';
+import {
+  f32,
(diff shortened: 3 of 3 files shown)


src/unittests/maths.spec.ts
```diff
@@ -3,20 +3,19 @@ Util math unit tests.
 `;
 
 import { makeTestGroup } from '../common/framework/test_group.js';
-import { kBit } from '../webgpu/shader/execution/builtin/builtin.js';
+import { anyOf, CaseList, deepEquals, kBit } from '../webgpu/shader/execution/builtin/builtin.js';
```
ben-clayton, line 6: It doesn't seem right that we have `src/unittests` test depending on `../webgpu/shader/execution/builtin`

src/webgpu/shader/execution/builtin/builtin.ts
```diff
@@ -845,3 +847,80 @@ export const kValue = {
...
+
+/**
+ * Checks for equality between two variable, doing a deep comparison if needed.
+ *
+ * |funcs_equal| controls how functions are compared. If |funcs_equal| is true,
+ * then functions are not deep inspected, and lhs and rhs both being a function
+ * is sufficient to pass.
+ * If false, then JSON.stringify is called to confirm they have the same body.
+ * */
+export function deepEquals(lhs: any, rhs: any, funcs_equal: boolean = false): boolean {
```
ben-clayton, line 859: I'm not familiar enough with TS or the testing framework to properly review this. My gut response is that I'm surprised that this is needed, and I'm not confident it'll work for all inputs. @kainino0x for feedback.

It appears that it cannot handle cyclic references. Probably worth a mention at least.

src/unittests/maths.spec.ts
kainino0x: It's a little odd, but I think unittests/ (which runs on presubmit) is the right place for testing helper code, and webgpu/ is the right place for the helper code. We have several other webgpu/ helpers tested in unittests/ so it's at least consistent.

We do enforce limitations on imports between different parts of the source tree: `.eslintrc.json` has a few specific lints set up. If we wanted to change it, we could add rules there.

src/webgpu/shader/execution/builtin/builtin.ts
kainino0x: There's nothing standard AFAIK, but we do have `objectEquals` in `src/common/util/util.ts` already. (Maybe not the easiest-to-find name.)

It doesn't handle cycles and most implementations of deep equality out there don't. It's also slightly simpler at least in function comparisons and probably some other things, so if you need them you can improve that implementation.

src/unittests/maths.spec.ts
zoddicus: I agree, there is an issue with this. I have redone this code to isolate the builtin specific stuff to that file and only implement a generic cartesian pair functions in maths.spec.ts.

src/webgpu/shader/execution/builtin/builtin.ts
zoddicus: I no longer need this, so removed it. There is also an implementation in src/common/util/util.ts that I should have been using.

zoddicus (commented):
PTAL, significant rewrite.

zoddicus:
Apparently there was a race between me and Kai on comments, I wrote my review comments without seeing Kai's

src/webgpu/util/math.ts
```diff
@@ -295,3 +307,16 @@ export function biasedRange(min: Scalar, max: Scalar, num_steps: Scalar): Array<
     lerp(f32_min, f32_max, Math.pow(lerp(0, 1, i / (u32_num_steps - 1)), c))
   );
 }
+
+/** Generates all of the pairs created by taking the cartesian product of an array with itself. */
+export function cartesianPairs<T>(values: T[]): [T, T][] {
+  const result = new Array<[T, T]>();
+  for (let lhs_index = 0; lhs_index < values.length; lhs_index++) {
```
dneto0, line 314: I think you could do this instead of the nested loop:

```js
values.forEach((e,ei) => {
  values.forEach((f,fi) => {
    result.push([e,f]);
  })
});
```

src/webgpu/shader/execution/builtin/builtin.ts
```diff
@@ -845,3 +851,39 @@ export const kValue = {
...
+ *   {[-1.0, 10.0], -1.0}
+ *   {[10.0, 11.0], -1.0}
+ *   {[0.0, 0.0], anyOf(x, 0.0)}
+ *   {[0.0, 10.0], anyOf(x, 0.0)}
+ *   {[10.0, 0.0], anyOf(x, 0.0)}
+ *   {[10.0, 10.0], 10.0}
+ **/
+export function generateCartesianPairCases(values: Array<Scalar>): CaseList {
+  const cases: CaseList = [];
+  cartesianPairs(values).forEach(inputs => {
```
dneto0, line 878: I know you're trying to be DRY here by reusing the cartesian product.  
But the algorithm is a bit obscure, and has a built-in inefficiency in having to look up the element every time.

Open-coding the cartesian product would be clearer, using the form of [forEach](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach) where the index is given to the lambda as a parameter:

```js
values.forEach((e,ei) => {
  values.forEach((f,fi) => {
    let precise_expected = (ei <= fi ? e : f);
    let expected = isSubnormalScalar(precise_expected ? anyOf(precise_expected,f32(0.0)) : precise_expected);
    result.push([e,f, expected ]);
  })
});
```

src/webgpu/util/math.ts
```diff
@@ -87,12 +87,24 @@ function flushSubnormalBits(val: number): number {
...
  */
 function flushSubnormalScalar(val: Scalar): Scalar {
+  return isSubnormalScalar(val) ? f32(0) : val;
+}
+
+/**
+ * @returns true if |val| is a subnormal f32 number, otherwise returns false
+ * 0 is considered a non-subnormal number by this function.
+ */
+export function isSubnormalScalar(val: Scalar): boolean {
```
dneto0, line 97: This is a nice tidiness fixup.

dneto0 (changes requested):
Using JS idioms / facilities will make this code clearer.

src/webgpu/util/math.ts
dneto0: whoops. In this part of the code you don't need the ei and fi arguments...
(And you may not need this function at all)

zoddicus (commented):
PTAL

zoddicus:
PTAL, I have updated this to integrate David's suggestions

dneto0 (approved):
Very nice!

Issue #949, opened 2022-01-24
stress: device allocation


lokokung:
Ensure device allocation and destruction is handled properly and resources are not held up.

kainino0x:
Looks like this is complete

PR #961, opened 2022-01-28
Adds stress tests for device destroy.

lokokung:
Issue: #949

<hr>


lokokung:
Note that this is currently blocked because of missing forceFallbackAdapter flag in types. I created a PR there for the patch: https://github.com/gpuweb/types/pull/88

The fallback adapters tests may also start off as failing since those changes were only recently landed. I will be trying to run these on the bots ASAP, but they are WAI on my local Mac and wanted to get reviews going asap.

kainino0x:
Sorry, I couldn't get to this before vacation. Hopefully someone else can review, or I can have it on my backlog for when I get back.

src/webgpu/capability_info.ts
```diff
@@ -12,6 +12,19 @@ import { ImageCopyType } from './util/texture/layout.js';
 
 // Base device limits can be found in constants.ts.
 
+// Adapters
```
austinEng, line 15: nit: maybe these should just be in the device_allocation.ts for now. Not clear where else they will be used (yet), and it's not exactly "capabilities" like much of the info in this file.

src/stress/adapter/device_allocation.spec.ts
```diff
@@ -0,0 +1,264 @@
...
+      device.destroy();
+    }
+  });
+
+g.test('continuous,no_destroy')
+  .desc(
+    `Tests allocation and implicit GC of many GPUDevice objects over time. Objects are
+sequentially requested and dropped for GC over a very large number of iterations. Note
+that without destroy, we do not create device allocated objects because that will
+implicitly keep the device in scope.`
```
austinEng, line 246: is this true? if we just do: 
```
await kFunctions[i % kFunctions.length](adapter);
```

without assigning to anything on the left hand side, they should all still be GC'ed

src/stress/adapter/device_allocation.spec.ts
```diff
@@ -0,0 +1,264 @@
...
+    const adapter = await getGPU().requestAdapter(kAdapterTypeOptions[adapterType]);
+    assert(adapter !== null, 'Failed to get adapter.');
+
+    // Since devices are being destroyed, we should be able to create many devices.
+    const kNumDevices = 100;
+    const kFunctions = [
+      t.createDeviceAndBuffers,
+      t.createDeviceAndComputeCommands,
+      t.createDeviceAndRenderCommands,
+    ];
```
austinEng, line 228: nit: probably best to put these each as a param, or a subcase param. Overall, I think this test and the following test can have params deviceOnly, deviceAndBuffers, deviceAndComputeCommands, deviceAndRenderCommands

src/stress/adapter/device_allocation.spec.ts
```diff
@@ -0,0 +1,264 @@
...
+import { Fixture } from '../../common/framework/fixture.js';
+import { makeTestGroup } from '../../common/framework/test_group.js';
+import { attemptGarbageCollection } from '../../common/util/collect_garbage.js';
+import { getGPU } from '../../common/util/navigator_gpu.js';
+import { assert, iterRange } from '../../common/util/util.js';
+import { kAdapterTypes, kAdapterTypeOptions } from '../../webgpu/capability_info.js';
+import { DefaultLimits } from '../../webgpu/constants.js';
+
+class DeviceAllocationTests extends Fixture {
+  async createDeviceAndComputeCommands(adapter: GPUAdapter) {
```
austinEng, line 14: nit: could be helpful to comment that we're only trying to make a valid pipeline, but we're never going to actually run it.

except.. maybe it's also worth having a .destroy test that does queue.submit([]) of everything, immediately followed by destroy? i.e. destroy while GPU executing

austinEng (approved):

src/webgpu/capability_info.ts
lokokung: Done, but we will probably want to eventually move them somewhere more common since I think they will be useful for future coverage testing.

src/stress/adapter/device_allocation.spec.ts
lokokung: Yea, but in the previous test, we are purposely keeping track of all the objects which acts as a test control to ensure that the device destroy is the one recycling memory. In this case, what I wanted to do was only keep track of the objects, but not the device and test the same behavior. Unfortunately, though, that doesn't work because by keeping the objects, the device will be kept alive and hence won't be GC-ed.

If we don't assign anything to the left side, it's not guaranteed that the recycling of resources is due to the device being GC-ed since the objects are being GC-ed as well. (I could still include that, but I'm not sure what value it brings atm, assuming we are going to eventually implement the other destroy stress tests for the sub-device objects.)

src/stress/adapter/device_allocation.spec.ts
lokokung: Added comments for each of the helpers.

src/stress/adapter/device_allocation.spec.ts
lokokung: Hmm not entirely sure what you mean, but I am hesitant to make them params because the test itself takes on the order of 3-5 minutes already, so having a combination of things could make the suite take a long time. At the same time though, I don't want to have to decrease the number of iterations per run since the large number of iterations is important for testing OOM situations.

I might leave them as is for now?

src/stress/adapter/device_allocation.spec.ts
lokokung: Resolved offline.

PR #977, opened 2022-02-03
Adds destroy validation tests for some APIs.

lokokung:
- Adds destroy validation tests for all device dependent webgpu APIs.
- Updates device_pool.ts to allow for destroyed devices to not fail and still mark device as needing to be reset
- Adds type/copy info params into capability_info.ts to make testing simpler but still have a decent combinatoric coverage

Issue: #882 

<hr>


src/webgpu/capability_info.ts
```diff
@@ -38,6 +38,34 @@ export const kQueryTypes = keysOf(kQueryTypeInfo);
 /** Required alignment of a GPUBuffer size, by spec. */
 export const kBufferSizeAlignment = 4;
 
+/** Per-GPUBufferUsage type info. */
+export const kBufferUsageTypesInfo: {
+  readonly [name: string]: number;
```
austinEng, line 43: could we do this?

```suggestion
  readonly [name: string]: GPUBufferUsageFlags;
```

src/webgpu/api/validation/state/device_lost/destroy.spec.ts
```diff
@@ -0,0 +1,184 @@
...
+  kRegularTextureFormats,
+  kCompressedTextureFormats,
+  kTextureFormatInfo,
+} from '../../../../capability_info.js';
+import { ValidationTest } from '../../validation_test.js';
+
+export const g = makeTestGroup(ValidationTest);
+
+g.test('createBuffer')
+  .desc(`A destroyed device should not be able to create any valid buffers.`)
```
austinEng, line 22: I think the description should have more detail about what we're testing and why. See https://github.com/gpuweb/cts/blob/main/docs/intro/plans.md

ex.) for this one, I imagine we combne different usages because we expect it to change how the implementaton allocates the resources.

src/webgpu/api/validation/state/device_lost/destroy.spec.ts
```diff
@@ -0,0 +1,184 @@
...
+  kTextureUsageCopy,
+  kTextureUsageCopyInfo,
+  kRegularTextureFormats,
+  kCompressedTextureFormats,
+  kTextureFormatInfo,
+} from '../../../../capability_info.js';
+import { ValidationTest } from '../../validation_test.js';
+
+export const g = makeTestGroup(ValidationTest);
+
```
austinEng, line 20: how did we decide what resources to test creation of?
if there's other stuff to add testing for, we need to plan it out with TODOs or no one will know what remains.

src/webgpu/api/validation/state/device_lost/destroy.spec.ts
```diff
@@ -0,0 +1,184 @@
...
+      texture.createView({ format });
+    });
+  });
+
+g.test('createSampler')
+  .desc(`A destroyed device should not be able to create any valid samplers.`)
+  .fn(async t => {
+    t.device.destroy();
+    t.expectValidationError(() => {
+      t.device.createSampler({});
```
austinEng, line 182: I think the descriptor should be optional

```suggestion
      t.device.createSampler();
```

src/webgpu/capability_info.ts
```diff
@@ -38,6 +38,34 @@ export const kQueryTypes = keysOf(kQueryTypeInfo);
...
+  'map-write':     Number(GPUConst.BufferUsage.MAP_WRITE),
+  'index':         Number(GPUConst.BufferUsage.INDEX),
+  'vertex':        Number(GPUConst.BufferUsage.VERTEX),
+  'uniform':       Number(GPUConst.BufferUsage.UNIFORM),
+  'storage':       Number(GPUConst.BufferUsage.STORAGE),
+  'indirect':      Number(GPUConst.BufferUsage.INDIRECT),
+  'query-resolve': Number(GPUConst.BufferUsage.QUERY_RESOLVE),
+};
+/** List of all GPUBufferUsage type values. */
+export const kBufferUsageTypes = keysOf(kBufferUsageTypesInfo);
```
austinEng, line 55: mm maybe we can do something like...
```typescript
export const kBufferUsageTypes = Object.keys(GPUConst.BufferUsage) as (keyof typeof GPUConst.BufferUsage)[];

// I think this should work
for (const t of kBufferUsageType) {
  const usage = GPUConst.BufferUsage[t];
}
```

but maybe name it somethng like kBufferUsageKeys or something like that so it's clear it's a key into GPUConst.BufferUsage

src/webgpu/util/device_pool.ts
```diff
@@ -69,7 +69,10 @@ export class DevicePool {
           holder.device.destroy();
         }
       }
-      throw ex;
+      // Supress failures if the device was explicitly destroyed.
+      if (holder.lostInfo !== undefined && holder.lostInfo.reason !== 'destroyed') {
+        throw ex;
+      }
```
austinEng, line 75: which failures happen? isn't it caught in expectValidationError ?

src/webgpu/capability_info.ts
lokokung: Done.

src/webgpu/api/validation/state/device_lost/destroy.spec.ts
lokokung: Done.

src/webgpu/api/validation/state/device_lost/destroy.spec.ts
lokokung: Ah... I added more tests for now. The plan is to test all the API. I decided I will just implement them all in this PR, so it will be a WIP. Feel free to make suggestions as I add commits, or once at the end if you prefer. Converting the PR to a draft for now.

src/webgpu/capability_info.ts
lokokung: Played around with it a bit and ended up with this. PTAL and see if there's any better way to do it. (I'm still quite new the this TS typing and the errors I saw are rather cryptic sometimes so not sure if this is the best way.)

Ultimately, the goal is so that we can use a string as the param, and use some sort of mapping in the test to get the flags. A string in the param is a lot easier to read in the CTS format instead of opaque integers.

src/webgpu/util/device_pool.ts
lokokung: If device is lost, there is a separate check, not just the validation check. Before the code re-threw the error. Now we only re-throw the error if the device wasn't explicitly destroyed.

src/webgpu/api/validation/state/device_lost/destroy.spec.ts
lokokung: Done.

src/webgpu/util/device_pool.ts
austinEng: Ah I see - however, my concern with this is that it will not throw for *any* device.destroy() call in *any test*.
I think it would be better if we can explicitly tell the harness in the test that we're expecting the lostInfo.reason to be "destroyed"

src/webgpu/capability_info.ts
```diff
@@ -38,23 +38,32 @@ export const kQueryTypes = keysOf(kQueryTypeInfo);
...
+  'none':     0,
+  'src':      Number(GPUConst.BufferUsage.COPY_SRC),
+  'dst':      Number(GPUConst.BufferUsage.COPY_DST),
+  'src-dest': Number(GPUConst.BufferUsage.COPY_SRC) | Number(GPUConst.BufferUsage.COPY_DST),
 };
+/** List of all GPUBufferUsage copy values. */
+export const kBufferUsageCopy = keysOf(kBufferUsageCopyInfo);
+
+/** Per-GPUBufferUsage keys and info. */
+export const kBufferUsageKeys = Object.keys(GPUConst.BufferUsage);
```
austinEng, line 54: the type of this is just `string[]` because Object.keys doesn't do more magic stuff.

So we need `as` to narrow it more. `GPUConst.BufferUsage` is an object, not a type. So `typeof` gives us the type of it:
```typescript
// typeof GPUConst.BufferUsage
type BufferUsageT = {
    readonly MAP_READ: 1;
    readonly MAP_WRITE: 2;
    readonly COPY_SRC: 4;
    readonly COPY_DST: 8;
    readonly INDEX: 16;
    readonly VERTEX: 32;
    readonly UNIFORM: 64;
    readonly STORAGE: 128;
    readonly INDIRECT: 256;
    readonly QUERY_RESOLVE: 512;
}
```
`keyof` will give you a type that is a union of the possible keys of the type.

```typescript
// keyof BufferUsageT
type BufferUsageKeys = "MAP_READ" | "MAP_WRITE" | "COPY_SRC" | "COPY_DST" | "INDEX" | "VERTEX" | "UNIFORM" | "STORAGE" | "INDIRECT" | "QUERY_RESOLVE";
```


```suggestion
type BufferUsageKey = keyof typeof GPUConst.BufferUsage;

export const kBufferUsageKeys = Object.keys(GPUConst.BufferUsage) as BufferUsageKey[];
```

src/webgpu/util/device_pool.ts
lokokung: Changed it so that it requires an explicit call.

src/webgpu/capability_info.ts
lokokung: Ah... this is so much nicer. Thanks for the suggestions!

src/webgpu/capability_info.ts
```diff
@@ -456,6 +465,29 @@ export function textureDimensionAndFormatCompatible(
...
+
+/** Per-GPUTextureUsage copy info. */
+export const kTextureUsageCopyInfo: {
+  readonly [name: string]: number;
+} = /* prettier-ignore */ {
+  'none':     0,
+  'src':      Number(GPUConst.TextureUsage.COPY_SRC),
+  'dst':      Number(GPUConst.TextureUsage.COPY_DST),
+  'src-dest': Number(GPUConst.TextureUsage.COPY_SRC) | Number(GPUConst.TextureUsage.COPY_DST),
+};
```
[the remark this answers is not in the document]
lokokung: Done.

src/webgpu/capability_info.ts
```diff
@@ -38,23 +38,32 @@ export const kQueryTypes = keysOf(kQueryTypeInfo);
...
+export const kBufferUsageKeys = Object.keys(GPUConst.BufferUsage);
+export const kBufferUsageInfo: {
+  readonly [k in typeof kBufferUsageKeys[number]]: GPUBufferUsageFlags;
+} = (() => {
+  const result: { [k in typeof kBufferUsageKeys[number]]: GPUBufferUsageFlags } = {};
+  for (const k in kBufferUsageKeys) {
+    result[k] = GPUConst.BufferUsage[k as keyof typeof GPUConst.BufferUsage];
+  }
+  return result;
+})();
```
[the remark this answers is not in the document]
lokokung: Yes! This is much better thanks!

src/webgpu/capability_info.ts
```diff
@@ -38,23 +38,32 @@ export const kQueryTypes = keysOf(kQueryTypeInfo);
...
+  'dst':      Number(GPUConst.BufferUsage.COPY_DST),
+  'src-dest': Number(GPUConst.BufferUsage.COPY_SRC) | Number(GPUConst.BufferUsage.COPY_DST),
 };
+/** List of all GPUBufferUsage copy values. */
+export const kBufferUsageCopy = keysOf(kBufferUsageCopyInfo);
+
+/** Per-GPUBufferUsage keys and info. */
+export const kBufferUsageKeys = Object.keys(GPUConst.BufferUsage);
+export const kBufferUsageInfo: {
+  readonly [k in typeof kBufferUsageKeys[number]]: GPUBufferUsageFlags;
```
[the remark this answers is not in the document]
lokokung: N/A anymore.

lokokung:
Added the TODOs for the other tests to be implemented in a follow up PR.

src/webgpu/util/device_pool.ts
```diff
@@ -238,6 +252,9 @@ class DeviceHolder implements DeviceProvider {
   state: DeviceHolderState = 'free';
   // initially undefined; becomes set when the device is lost
   lostInfo?: GPUDeviceLostInfo;
+  // Set if the device is expected to be lost. The null value is used to indicate that the lost
+  // reason can be anything.
```
austinEng, line 256: Why do we need to allow it to be anything? I would think null means that we''re not expecting device lost to happen. And if it's non-null, then we expect the reason to match exactly.

src/webgpu/capability_info.ts
```diff
@@ -38,23 +38,29 @@ export const kQueryTypes = keysOf(kQueryTypeInfo);
...
+  'COPY_SRC':     Number(GPUConst.BufferUsage.COPY_SRC),
+  'COPY_DST':     Number(GPUConst.BufferUsage.COPY_DST),
+  'COPY_SRC_DST': Number(GPUConst.BufferUsage.COPY_SRC) | Number(GPUConst.BufferUsage.COPY_DST),
+};
+/** List of all GPUBufferUsage copy values. */
+export const kBufferUsageCopy = keysOf(kBufferUsageCopyInfo);
+
+/** Per-GPUBufferUsage keys and info. */
+type BufferUsageKey = keyof typeof GPUConst.BufferUsage;
+export const kBufferUsageKeys = Object.keys(GPUConst.BufferUsage) as BufferUsageKey[];
```
austinEng, line 55: I'm sorry, I realized this is already what `keysOf` does for us.
```suggestion
export const kBufferUsageKeys = keysOf(GPUConst.BufferUsage);
```

src/webgpu/capability_info.ts
```diff
@@ -38,23 +38,29 @@ export const kQueryTypes = keysOf(kQueryTypeInfo);
...
-  [GPUConst.BufferUsage.INDEX]:         {},
-  [GPUConst.BufferUsage.VERTEX]:        {},
-  [GPUConst.BufferUsage.UNIFORM]:       {},
-  [GPUConst.BufferUsage.STORAGE]:       {},
-  [GPUConst.BufferUsage.INDIRECT]:      {},
-  [GPUConst.BufferUsage.QUERY_RESOLVE]: {},
+  'COPY_NONE':    0,
+  'COPY_SRC':     Number(GPUConst.BufferUsage.COPY_SRC),
+  'COPY_DST':     Number(GPUConst.BufferUsage.COPY_DST),
+  'COPY_SRC_DST': Number(GPUConst.BufferUsage.COPY_SRC) | Number(GPUConst.BufferUsage.COPY_DST),
```
austinEng, line 48: Do we really need `Number(...)` ?
```suggestion
  'COPY_SRC':     GPUConst.BufferUsage.COPY_SRC,
  'COPY_DST':     GPUConst.BufferUsage.COPY_DST,
  'COPY_SRC_DST': GPUConst.BufferUsage.COPY_SRC) | GPUConst.BufferUsage.COPY_DST,
```

src/webgpu/api/validation/state/device_lost/destroy.spec.ts
```diff
@@ -0,0 +1,543 @@
...
+  - Binds valid pipeline and bindgroups, then draws
+  - Tests that finishing encoding fails on destroyed device
+  - Tests that submitting command fails on destroyed device
+  `
+  )
+  .unimplemented();
+
+g.test('command,renderPass,renderBundle')
+  .desc(
+    `
```
austinEng, line 536: maybe we need queue.writeBuffer, queue.writeTextue, and queue.copyExternalImageToTexture as well

austinEng (commented):
test plans look good. Not sure if there's a good way to make sure we catch everything should new resources / commands be added. Just some comments on how we do the device lost expectation.

src/webgpu/api/validation/state/device_lost/destroy.spec.ts
```diff
@@ -406,3 +408,136 @@ Tests creating query sets on destroyed device fails.
...
+
+g.test('command,copyBufferToTexture,uncompressed_format')
+  .desc(
+    `
+Tests copyBufferToTexture command with various uncompressed formats on destroyed device fails
+  - Tests that finishing encoding fails on destroyed device
+  - Tests that submitting command fails on destroyed device
+  `
+  )
+  .unimplemented();
```
kainino0x, line 430: I don't think we need any variations on the formats; it would be pretty weird for an implementation to pass for one format and fail for another format. Testing with just one format would be OK.

src/webgpu/api/validation/state/device_lost/destroy.spec.ts
```diff
@@ -406,3 +408,136 @@ Tests creating query sets on destroyed device fails.
       t.device.createQuerySet({ type, count: 4 });
     });
   });
+
+g.test('command,copyBufferToBuffer')
+  .desc(
+    `
+Tests copyBufferToBuffer command on destroyed device fails.
+  - Tests that finishing encoding fails on destroyed device
+  - Tests that submitting command fails on destroyed device
```
kainino0x, line 417: How are you going to actually check this though? Lost devices don't provide error scope.

also: It may be that testing submit in each test is trivial, but if not then we can skip it. I don't think we need submit tests for command buffers containing different commands. (They're all invalid command buffers anyway.)
There's a difference between [finish, destroy, submit] and [destroy, finish, submit]. I think we only really need tests for:
- [finish, destroy, submit]
- [destroy, finish]

src/webgpu/util/device_pool.ts
```diff
@@ -69,7 +70,20 @@ export class DevicePool {
...
+      // `holder.expectedLostReason` is set to null or equal to `holder.lostInfo.reason`. The
+      // condition below is just the negation to avoid extra branching.
+      if (
+        !(
+          holder.lostInfo !== undefined &&
+          holder.expectedLostReason !== undefined &&
+          (holder.expectedLostReason === null ||
+            holder.expectedLostReason === holder.lostInfo.reason)
+        )
+      ) {
```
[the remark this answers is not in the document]
lokokung: Applied readability change. Note that the reason I am doing the assert is because the catch block marks the device for reinitialization for future tests. Otherwise, I would need to duplicate the device marking code. Since a destroyed device always needs to be reinitialized, I handle the special cases for suppressing the error in the catch block.

src/webgpu/capability_info.ts
lokokung: Done

src/webgpu/capability_info.ts
lokokung: Done

src/webgpu/api/validation/state/device_lost/destroy.spec.ts
lokokung: Oops, I actually had those but I accidentally deleted them in my splitting. Good catch!

src/webgpu/api/validation/state/device_lost/destroy.spec.ts
lokokung: Done.

src/webgpu/util/device_pool.ts
lokokung: Done.

src/webgpu/api/validation/state/device_lost/destroy.spec.ts
kainino0x: SGTM. What about the error checking problem though? popErrorScope rejects on lost devices so you can't see if a validation error occurred.

lokokung:
Added subcases for awaitLost as per offline discussion with Kai. PR is also currently blocked by decision in https://github.com/gpuweb/gpuweb/issues/2582 since these CTS tests are implicitly using the fact that pop/push is currently still valid (at least on Chrome), even when the device is lost. If that changes, these tests will need to be updated as well.

src/webgpu/api/validation/state/device_lost/destroy.spec.ts
lokokung: Resolving this comment for now as it is blocked by https://github.com/gpuweb/gpuweb/issues/2582. Depending on the decision there, I will re-work the tests as necessary.

lokokung:
Updated to reflect the changes for PopErrorScope! PTAL again. I will add explicit tests separately to test PopErrorScope and I have some follow up changes for the other tests introduced in this PR as well.

austinEng (approved):

Issue #872, opened 2022-01-11
Expand tests for 1D textures.


Kangz:
While implementing 1D textures in Chromium, I'm finding there's a bunch of missing coverage or incorrect tests. Opening this issue to track so that PRs improving 1D coverage can be linked to it.

PR #978, opened 2022-02-03
Extend copyB2T/copyT2B/WriteTexture tests for 1D/3D textures

Kangz:
Issue: #872

<hr>

### DIFF
--- a/src/webgpu/api/operation/command_buffer/image_copy.spec.ts
+++ b/src/webgpu/api/operation/command_buffer/image_copy.spec.ts
@@ -1377,6 +1377,7 @@ works for every format with 2d and 2d-array textures.
       textureHeight = info.blockHeight;
       rowsPerImage = 1;
     }
+    const textureSize = [4 * info.blockWidth, textureHeight, copyDepth] as const;
 
     const minDataSize = dataBytesForCopyOrFail({
       layout: { offset, bytesPerRow, rowsPerImage },
@@ -1392,7 +1393,7 @@ works for every format with 2d and 2d-array textures.
       textureDataLayout: { offset, bytesPerRow, rowsPerImage },
       copySize,
       dataSize,
-      textureSize: [4 * info.blockWidth, textureHeight, copyDepth],
+      textureSize,
       format,
       dimension,
       initMethod,
@@ -1512,6 +1513,7 @@ function* generateTestTextureSizes({
   mipLevel: number;
   _mipSizeInBlocks: Required<GPUExtent3DDict>;
 }): Generator<[number, number, number]> {
+  assert(dimension !== '1d'); // textureSize[1] would be wrong for 1D mipped textures.
   const info = kTextureFormatInfo[format];
 
   const widthAtThisLevel = _mipSizeInBlocks.width * info.blockWidth;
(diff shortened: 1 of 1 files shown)


src/webgpu/api/operation/command_buffer/image_copy.spec.ts
```diff
@@ -1358,8 +1392,9 @@ works for every format with 2d and 2d-array textures.
       textureDataLayout: { offset, bytesPerRow, rowsPerImage },
       copySize,
       dataSize,
-      textureSize: [4 * info.blockWidth, 4 * info.blockHeight, copyDepth],
+      textureSize: [4 * info.blockWidth, textureHeight, copyDepth],
```
austinEng, line 1395: maybe should pull the width out above as well so all the size computations are colocated.

src/webgpu/api/operation/command_buffer/image_copy.spec.ts
```diff
@@ -1514,6 +1565,8 @@ g.test('mip_levels')
       .combineWithParams(kMethodsToTest)
       .combine('format', kWorkingTextureFormats)
       .filter(formatCanBeTested)
+      .combine('dimension', ['2d', '3d'] as const)
```
austinEng, line 1568: comment why we don't do 1d. Or better, comment about what dimensinos are tested in the description

src/webgpu/api/operation/command_buffer/image_copy.spec.ts
```diff
@@ -1472,7 +1519,7 @@ function* generateTestTextureSizes({
   const textureSize: [number, number, number] = [
     widthAtThisLevel << mipLevel,
     heightAtThisLevel << mipLevel,
-    _mipSizeInBlocks.depthOrArrayLayers,
+    _mipSizeInBlocks.depthOrArrayLayers << (dimension === '3d' ? mipLevel : 0),
   ];
```
austinEng, line 1523: this could use helper `virtualMipSize`. textureSize[1] would be wrong for 1d if this generator shouldn't be used for 1d, add an assert about that

src/webgpu/api/operation/command_buffer/image_copy.spec.ts
```diff
@@ -1500,6 +1547,10 @@ function* generateTestTextureSizes({
   if (modifyWidth && modifyHeight) {
     yield [modifiedWidth, modifiedHeight, textureSize[2]];
   }
+
+  if (dimension === '3d') {
+    yield [textureSize[0], textureSize[1], textureSize[2] + 1];
+  }
```
austinEng, line 1553: not clear to me where we're only testing [w, h, d+1]. I guess that given the function comment, 3D here might not really matter much since the purpose of this generator is to generate differences in width+height ?

src/webgpu/api/operation/command_buffer/image_copy.spec.ts
Kangz: This is a bit hard because `uploadTextureAndVerifyCopy` expects a `[number, number, number]` and that's only possible if I use `as const`, which prevents modifying `textureSize`. I moved `textureSize` right after the block  that computes `textureHeight` though.

src/webgpu/api/operation/command_buffer/image_copy.spec.ts
Kangz: What this function is doing is actually the inverse of `virtualMipSize`, given the physical size of a level (in blocks), it computes the size of the texture at level 0.

src/webgpu/api/operation/command_buffer/image_copy.spec.ts
Kangz: We're also testing `[w, h, d]`, I thought it would be some extra coverage for partial-depth copies to 3D texture mip levels. WDYT?

Updated the description accordingly.

austinEng: ah okay, sounds good

austinEng (approved):

PR #981, opened 2022-02-05
Rewrite the standalone test page with React

austinEng:
Biggest feature is that the page can render incrementally
while loading. This brings time to Largest Contentful Paint
from 4 seconds down to 500ms.


austinEng:
Friday hackday! - review in your own time @kainino0x 
I can clean this up and comment some more on this code, but probably not until we decide this rewrite is worth doing / in the right direction.

PR preview works [https://gpuweb-cts--cts-prs-981-d9abc7347f4e104436306c6756f460-ta2w5lhq.web.app/standalone/?runnow=0&worker=0&debug=0&q=webgpu:*](https://gpuweb-cts--cts-prs-981-d9abc7347f4e104436306c6756f460-ta2w5lhq.web.app/standalone/?runnow=0&worker=0&debug=0&q=webgpu:*)

kainino0x:
Looks nice on the build preview page. (Convenient!)
I've been wanting to move to something like React for ages and I think it's the right direction. Haven't reviewed yet really though.

WDYT of using [htm](https://github.com/developit/htm) rather than JSX? It's slightly more verbose but I used it over my vacation and found it pretty nice. It would be nice to simplify the build by not having to deal with JSX at all. I'm not sure how much of the new build-related stuff it would eliminate, though.

(I used htm with Preact, but it works with React too.)

austinEng:
Interesting, I didn't know about htm. Don't feel strongly, but perhaps slight -1 from me since Typescript already has built-in jsx support - it's just wrapped up by Babel for the rest of the translations we do.

austinEng:
Also, how do you feel about using a bundler? We can load all the framework code in one shot, single request which should be much much faster than the serial loading going on. Browser doesn't know the file dependencies ahead of time, so if we do it all together, loading these tiny files will be much faster. Maybe we can also alleviate it with a prefetch, but I feel like bundling is a better guarantee.
Also, probably can load React / Preact (haven't tried it, but small size is attractive) from a CDN.

cts-loading

kainino0x:
Bundling only some of the code (the framework code but not the test code) would put us in a weird hybrid situation. Can any of the bundlers take JS modules as input and also produce JS modules as output?

Also, since a bunch of util code would move to the bundle, there would be circular dependencies from bundle -> test -> bundle (maybe fine). And that's assuming the build can rewrite all of the dependencies in the test files to point at the bundle, which seems iffy.

Maybe there are circles we can draw around the file graph to bundle certain parts, but that seems like it would create quite a bit of complexity.

PR #989, opened 2022-02-10
Make cmdline.ts runnable when compiled as cmdline.js as well

austinEng:
Issue: #

<hr>

### DIFF
--- a/src/common/runtime/cmdline.ts
+++ b/src/common/runtime/cmdline.ts
@@ -28,7 +28,15 @@ function usage(rc: number): never {
   return sys.exit(rc);
 }
 
-if (!sys.existsSync('src/common/runtime/cmdline.ts')) {
+function selfFileExtension() {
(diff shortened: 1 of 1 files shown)


austinEng:
If you compile everything into a path like `path/to/compiled-js-out/src` and then run it with plain `node`, without  `ts-node`

but, it might be better to automatically change directories to where the script wants to be run from

kainino0x:
We have that in the form of `out-node/` but it doesn't have `src/` in it; `src/*` corresponds to `out-node/*`.

Also how can the script know where it should `cd` to? would that be a command line parameter?

It occurs to me I have no idea why this script actually needs to be run from the root, since all of the imports are relative. There's probably a reason, but I should have documented it...

austinEng:
maybe cd to `dir(dir(dir(dir(__filename))))` ? or if we don't actually need this check, then that's great too

kainino0x:
Found another solution

PR #993, opened 2022-02-14
Revert "Implement validation tests on pipeline_output_targets,format_blendable"

kainino0x:
Reverts gpuweb/cts#990

> hitting an unreachable() inside getFragmentShaderCode
> https://chromium-review.googlesource.com/c/chromium/src/+/3462090


kainino0x:
Two tests failed:
webgpu:api,operation,render_pipeline,pipeline_output_targets:color,component_count:*
webgpu:api,validation,createRenderPipeline:pipeline_output_targets:*

Likely due to the changes in capability_info - please test both when relanding.
(No problem that we didn't catch it beforehand - it's expected because capability_info changes always have wide reaching consequences that can be hard to find before roll.)

PR #994, opened 2022-02-15
Remove CWD check from cmdline.ts

kainino0x:
Alternative fix to #989


austinEng (approved):
thanks for finding this better fix!

PR #1023, opened 2022-02-28
Implement `atan2` test

zoddicus:
Also extends `linearRange` and `biasedRange` to support arbitrary
inputs instead of requiring them to be in order.

<hr>


src/webgpu/util/math.ts
```diff
@@ -267,43 +267,42 @@ export function lerp(a: number, b: number, t: number) {
...
- * @returns a non-linear increasing range of numbers, with a bias towards min.
+ * @returns a non-linear increasing range of numbers, with a bias towards the beginning.
  *
  * Generates a linear range on [0,1] with |num_steps|, then squares all the values to make the curve be quadratic,
  * thus biasing towards 0, but remaining on the [0, 1] range.
  * This biased range is then scaled to the desired range using lerp.
  * Different curves could be generated by changing c, where greater values of c will bias more towards 0.
  * */
-export function biasedRange(min: Scalar, max: Scalar, num_steps: Scalar): Array<number> {
+export function biasedRange(a: Scalar, b: Scalar, num_steps: Scalar): Array<number> {
```
ben-clayton, line 301: Why do these function use `Scalar` instead of just `number`?

zoddicus: Probably legacy from me having rewritten it multiple times originally. Converted the code to use `number`

src/webgpu/shader/execution/builtin/atan2.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+  )
+  .fn(async t => {
+    // [1]: Need to decide what the ground-truth is.
+    const truthFunc = (y: number, x: number): Case => {
+      assert(x !== 0, 'atan2 is undefined for x = 0');
+      return { input: [f32(y), f32(x)], expected: f32(Math.atan2(y, x)) };
+    };
+
+    let numeric_range: Array<number> = [];
+    // //  -2^32 < x <= -1, biased towards -1
```
ben-clayton, line 39: `// //` ?

src/webgpu/shader/execution/builtin/atan2.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+    numeric_range = numeric_range.concat(biasedRange(-1, -(2 ** 32), 50));
+    // -1 <= x < 0, linearly spread
+    numeric_range = numeric_range.concat(
+      linearRange(-1, f32Bits(kBit.f32.negative.max).value as number, 20)
+    );
+    // 0 < x < -1, linearly spread
+    numeric_range = numeric_range.concat(
+      linearRange(f32Bits(kBit.f32.positive.min).value as number, 1, 20)
+    );
+    // // 1 <= x < 2^32, biased towards 1
```
ben-clayton, line 49: ditto

src/webgpu/shader/execution/builtin/atan2.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+      assert(x !== 0, 'atan2 is undefined for x = 0');
+      return { input: [f32(y), f32(x)], expected: f32(Math.atan2(y, x)) };
+    };
+
+    let numeric_range: Array<number> = [];
+    // //  -2^32 < x <= -1, biased towards -1
+    numeric_range = numeric_range.concat(biasedRange(-1, -(2 ** 32), 50));
+    // -1 <= x < 0, linearly spread
+    numeric_range = numeric_range.concat(
+      linearRange(-1, f32Bits(kBit.f32.negative.max).value as number, 20)
```
ben-clayton, line 43: `float32ToUint32` ? Here and elsewhere

src/webgpu/shader/execution/builtin/atan2.spec.ts
zoddicus: artifact of commenting/uncommenting regions in my editor, fixed

src/webgpu/shader/execution/builtin/atan2.spec.ts
zoddicus: fixed

src/webgpu/shader/execution/builtin/atan2.spec.ts
zoddicus: done

zoddicus: I think this is correct, we want the smallest floating point number to 0, but is not 0, since atan2 internally going to perform `y / x`, so isn't defined when `x = 0`.

`float32ToUint32` on `kBit.f32.negative.max` would give us 0x80800000 as a value, since it is treating the input as a number and not a bitfield, and calling it on `f32Bits(kBit.f32.negative.max)` would round to 0.

I don't think the CTS has the f32 limits defined as numeric constants, just as bit field constants, hence the `f32Bits` call.

zoddicus: I could add using f32Bits to convert the bitfield into a value to the kValue table, so that we don't have to do the conversion all over the place.

zoddicus: I actually ended up doing that, so we don't need to do the `.value as number` dance in the test code.

zoddicus:
I am not sure why the bot thinks i32Bits isn't defined. I will take a look to try to figure that out.

zoddicus:
> I am not sure why the bot thinks i32Bits isn't defined. I will take a look to try to figure that out.

*sigh* commenting out the offending lines causes the compile to bail on the u32Bits lines. I am assuming that the issue is that there is a circular dependency somewhere, so we cannot use these imported functions to define these exported values.

ben-clayton (approved):
Looks great. Thank you for all the changes!

PR #1024, opened 2022-03-01
Ensure lost device messages output the error message and reason.

toji:
Found that lost device messages in the CTS were simply printing:

`Device was unexpectedly lost: [object GPUDeviceLostInfo]`

This updates the error to include the lost info reason and message, which makes debugging failures easier.


austinEng (approved):
If there's no reason/message it'll say "reason: undefined" or "message: undefined".

Not sure if that's your desired result or if you want to remove the reason/message completely. LGTM though.

austinEng:
fyi: it appears the formatter is upset

toji:
> If there's no reason/message it'll say "reason: undefined" or "message: undefined".

That's fine by me, personally, as it makes it pretty explicit that there was nothing to report.

> fyi: it appears the formatter is upset

Oops! Fixed.

Issue #917, opened 2022-01-13
reftests: test canvas compositingAlphaMode


kainino0x:
Note Chromium doesn't implement this as of 2022-01, so this is soft-blocked for Chromium folks.

Note also, more testing might be needed if we do something about https://github.com/gpuweb/gpuweb/issues/1988.

kainino0x:
I think this is done between #1667 and past PRs.

PR #1028, opened 2022-03-02
Add variants of canvas_composite_alpha reftests that use a copy

austinEng:
This is to add coverage where the implementation may use a
render pass to clear alpha to 1.0 to implement "opaque". So
internally, the canvas texture should be made renderable.

Issue: #917

<hr>

### DIFF
--- a/src/webgpu/web_platform/reftests/README.txt
+++ b/src/webgpu/web_platform/reftests/README.txt
@@ -10,7 +10,6 @@ This tests things like:
 
 TODO(#915): canvas_complex: test rgba8unorm and rgba16float
 TODO(#916): canvas_complex: Test all ways to write into textures (currently only testing copy methods).
-TODO(#917): Test compositingAlphaMode options
 TODO(#918): Test all possible color spaces (once we have more than 1)
 
 TODO(#921): Why is there sometimes a difference of 1 (e.g. 3f vs 40) in canvas_size_different_with_back_buffer_size?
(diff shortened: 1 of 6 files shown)


austinEng:
Kai, what other tests were you thinking of for compositingAlphaMode (for what's currently specified). I think the reftests we have cover the behavior, unless of course we want to test additional formats.

Also, question on the existing code. Why do we scale up like this:?
```html
<canvas id="cvs" width="4" height="4" style="width: 20px; height: 20px;"></canvas>
```

I'm wondering if this is causing some of the pixel differences on CQ.

austinEng:
> Also, question on the existing code. Why do we scale up like this:?

probably to test the scaling behavior. In any case, looking at the Chromium bug - it really does seem like a Chrome-specific infra issue since local screenshots don't have the half pixel offset.

kainino0x (approved):

kainino0x:
> Kai, what other tests were you thinking of for compositingAlphaMode (for what's currently specified). I think the reftests we have cover the behavior, unless of course we want to test additional formats.

I think we just forgot to remove the TODO from README.txt in #696 (speaking of which, please do so in this PR), so when I went to file all the tasks I filed it.

> probably to test the scaling behavior. In any case, looking at the Chromium bug - it really does seem like a Chrome-specific infra issue since local screenshots don't have the half pixel offset.

Yeah, pretty much, and because it makes it easier for a human to inspect the test results. Eventually we want tests for CSS `image-rendering`, too, and ResizeObserver/devicePixelRatio, which I wrote down here yesterdayish: https://github.com/orgs/gpuweb/projects/3/views/15?filterQuery=is%3Aissue+is%3Aopen+ResizeObserver

And yes, agree it looks like a Chrome-specific problem. Come to think of it, if we _didn't_ scale up the canvas, wouldn't a half-pixel offset be really bad? Like, cause the whole image to become blurry?

Issue #913, opened 2022-01-13
web_platform: copyExternalImageToTexture with more formats, extended-srgb and other color spaces


kainino0x:
https://crbug.com/dawn/1138 adds new formats to copyExternalImageToTexture.

Some of the new formats are float, which means they go outside [0,1]. Color conversion behavior is now specified (colorspaces like srgb are extended). Find a way to test that no clamping occurs for float formats, e.g. upload a Display P3 source image - if possible.

---

Here's the single minimal test we would need in order to make sure we're implementing this correctly. It's quite simple. This is what we need to start implementing with some confidence.

- Create a Display P3 (1,0,0,1) image source. This is a red that's outside the srgb range.
  It probably has to be a PNG because it's going to have the best browser compatibility and least room for extra bugs.
- Use copyExternalImageToTexture to put it in an r(gba)16float or r(gba)32float texture with the default options.
- Test that the red channel of that texture is > 1.

Here are some test images. I took the WebKit test image and cropped it to 1x1 in GIMP.
Source: https://webkit.org/blog-files/color-gamut/

- red1x1-displayp3.png: https://webkit.org/blog-files/color-gamut/Webkit-logo-P3.png cropped to 1x1 and saved with the same color profile.
- red1x1-untagged.png: https://webkit.org/blog-files/color-gamut/Webkit-logo-P3.png cropped to 1x1 and saved with no color profile.
- red1x1-srgb.png: https://webkit.org/blog-files/color-gamut/Webkit-logo-sRGB.png cropped to 1x1 and saved with the same color profile.

Verified they display as expected on my MacBook (displayp3 is brighter than the other two).

We can test all three of these. The displayp3 red should be > 1, the others should be == 1.

Some test files are here: https://bugs.chromium.org/p/dawn/issues/detail?id=1139#c8

shaoboyan:
This issue is fixed by above PRs

kainino0x:
Thanks for the work on this! When closing an issue, if verifying it's complete please change the status in the project tracker to "Complete" (you can do this in the right-hand sidebar on the github issue).

Though, I noticed we never actually added any test image files to the CTS, so I guess we are only testing 2d canvas sources. I think we still need to expand these to more sources (image files, and [ImageData with colorSpace](https://developer.mozilla.org/en-US/docs/Web/API/ImageData/ImageData)), so I filed #1476 for that.

PR #1043, opened 2022-03-07
CTS: Color space conversion operation test for CopyToTexture

shaoboyan:
This PR add cts to test copyExternalImageToTexture() color space
conversion ability.
It creates canvas with color space attr and copy to dst texture in
user defined color space through CopyExternalImageToTexture().

Issue: #913
<hr>

### DIFF
--- a/src/webgpu/util/color_space_conversion.ts
+++ b/src/webgpu/util/color_space_conversion.ts
@@ -2,7 +2,7 @@ import { assert } from '../../common/util/util.js';
 
 import { multiplyMatrices } from './math.js';
 
-// The color space conversion function definations copies directly from
+// These color space conversion function definitions are copied directly from
 // CSS Color Module Level 4 Sample Code: https://drafts.csswg.org/css-color/#color-conversion-code
 
 // Sample code for color conversions
--- a/src/webgpu/util/color_space_conversion.ts
+++ b/src/webgpu/util/color_space_conversion.ts
@@ -1,45 +1,30 @@
 import { assert } from '../../common/util/util.js';
 
+import { multiplyMatrices } from './math.js';
+
 /**
  * The color space conversion function definations are fully aign with
  * CSS Color Module Level 4 Sample Code: https://drafts.csswg.org/css-color/#color-conversion-code
  */
 
 /**
- * Simple matrix (and vector) multiplication
- * Warning: No error handling for incompatible dimensions!
- * @author Lea Verou 2020 MIT License
+ *  Sample code for color conversions
+ * Conversion can also be done using ICC profiles and a Color Management System
+ * For clarity, a library is used for matrix multiplication (multiply-matrices.js)
  */
-// A is m x n. B is n x p. product is m x p.
-function multiplyMatrices(A: Array<Array<number>>, B: Array<Array<number>>) {
-  const B_cols = B[0].map((_, i) => B.map(x => x[i])); // transpose B
-  const product = A.map(row =>
-    B_cols.map(col => {
-      if (!Array.isArray(row)) {
-        return col.reduce((a, c) => a + c * row, 0);
-      }
-
-      return row.reduce((a, c, i) => a + c * (col[i] || 0), 0);
-    })
-  );
-
-  return product;
-}
-
-// Sample code for color conversions
-// Conversion can also be done using ICC profiles and a Color Management System
-// For clarity, a library is used for matrix multiplication (multiply-matrices.js)
 
 // sRGB-related functions
 
+/**
+ * convert an array of sRGB values
+ * where in-gamut values are in the range [0 - 1]
+ * to linear light (un-companded) form.
+ * https://en.wikipedia.org/wiki/SRGB
+ * Extended transfer function:
+ * for negative values,  linear portion is extended on reflection of axis,
+ * then reflected power function is used.
+ */
 function lin_sRGB(RGB: Array<number>) {
-  // convert an array of sRGB values
-  // where in-gamut values are in the range [0 - 1]
-  // to linear light (un-companded) form.
-  // https://en.wikipedia.org/wiki/SRGB
-  // Extended transfer function:
-  // for negative values,  linear portion is extended on reflection of axis,
-  // then reflected power function is used.
   return RGB.map(val => {
     const sign = val < 0 ? -1 : 1;
     const abs = Math.abs(val);
@@ -52,13 +37,15 @@ function lin_sRGB(RGB: Array<number>) {
   });
 }
 
+/**
+ * convert an array of linear-light sRGB values in the range 0.0-1.0
+ * to gamma corrected form
+ * https://en.wikipedia.org/wiki/SRGB
+ * Extended transfer function:
+ * For negative values, linear portion extends on reflection
+ * of axis, then uses reflected pow below that
+ */
 function gam_sRGB(RGB: Array<number>) {
-  // convert an array of linear-light sRGB values in the range 0.0-1.0
-  // to gamma corrected form
-  // https://en.wikipedia.org/wiki/SRGB
-  // Extended transfer function:
-  // For negative values, linear portion extends on reflection
-  // of axis, then uses reflected pow below that
   return RGB.map(val => {
     const sign = val < 0 ? -1 : 1;
     const abs = Math.abs(val);
@@ -71,10 +58,11 @@ function gam_sRGB(RGB: Array<number>) {
   });
 }
 
+/**
+ * convert an array of linear-light sRGB values to CIE XYZ
+ * using sRGB's own white, D65 (no chromatic adaptation)
+ */
 function lin_sRGB_to_XYZ(rgb: Array<Array<number>>) {
-  // convert an array of linear-light sRGB values to CIE XYZ
-  // using sRGB's own white, D65 (no chromatic adaptation)
-
   const M = [
     [0.41239079926595934, 0.357584339383878, 0.1804807884018343],
     [0.21263900587151027, 0.715168678767756, 0.07219231536073371],
@@ -83,9 +71,10 @@ function lin_sRGB_to_XYZ(rgb: Array<Array<number>>) {
   return multiplyMatrices(M, rgb);
 }
 
+/**
+ * convert XYZ to linear-light sRGB
+ */
 function XYZ_to_lin_sRGB(XYZ: Array<Array<number>>) {
-  // convert XYZ to linear-light sRGB
-
   const M = [
     [3.2409699419045226, -1.537383177570094, -0.4986107602930034],
     [-0.9692436362808796, 1.8759675015077202, 0.04155505740717559],
@@ -97,24 +86,28 @@ function XYZ_to_lin_sRGB(XYZ: Array<Array<number>>) {
 
 //  display-p3-related functions
 
+/**
+ * convert an array of display-p3 RGB values in the range 0.0 - 1.0
+ * to linear light (un-companded) form.
+ */
 function lin_P3(RGB: Array<number>) {
-  // convert an array of display-p3 RGB values in the range 0.0 - 1.0
-  // to linear light (un-companded) form.
-
   return lin_sRGB(RGB); // same as sRGB
 }
 
+/**
+ * convert an array of linear-light display-p3 RGB  in the range 0.0-1.0
+ * to gamma corrected form
+ */
 function gam_P3(RGB: Array<number>) {
-  // convert an array of linear-light display-p3 RGB  in the range 0.0-1.0
-  // to gamma corrected form
-
   return gam_sRGB(RGB); // same as sRGB
 }
 
+/**
+ * convert an array of linear-light display-p3 values to CIE XYZ
+ * using  D65 (no chromatic adaptation)
+ * http://www.brucelindbloom.com/index.html?Eqn_RGB_XYZ_Matrix.html
+ */
 function lin_P3_to_XYZ(rgb: Array<Array<number>>) {
-  // convert an array of linear-light display-p3 values to CIE XYZ
-  // using  D65 (no chromatic adaptation)
-  // http://www.brucelindbloom.com/index.html?Eqn_RGB_XYZ_Matrix.html
   const M = [
     [0.4865709486482162, 0.26566769316909306, 0.1982172852343625],
     [0.2289745640697488, 0.6917385218365064, 0.079286914093745],
@@ -125,8 +118,10 @@ function lin_P3_to_XYZ(rgb: Array<Array<number>>) {
   return multiplyMatrices(M, rgb);
 }
 
+/**
+ * convert XYZ to linear-light P3
+ */
 function XYZ_to_lin_P3(XYZ: Array<Array<number>>) {
-  // convert XYZ to linear-light P3
   const M = [
     [2.493496911941425, -0.9313836179191239, -0.40271078445071684],
     [-0.8294889695615747, 1.7626640603183463, 0.023624685841943577],
@@ -136,9 +131,13 @@ function XYZ_to_lin_P3(XYZ: Array<Array<number>>) {
   return multiplyMatrices(M, XYZ);
 }
 
-// Follow conversion steps in CSS Color Module Level 4
-// https://drafts.csswg.org/css-color/#predefined-to-predefined
-// display-p3 and sRGB share the same white points.
+/**
+ * @returns the converted pixels in {R: number, G: number, B: number, A: number}.
+ *
+ * Follow conversion steps in CSS Color Module Level 4
+ * https://drafts.csswg.org/css-color/#predefined-to-predefined
+ * display-p3 and sRGB share the same white points.
+ */
 export function displayP3ToSrgb(pixel: {
   R: number;
   G: number;
@@ -163,7 +162,13 @@ export function displayP3ToSrgb(pixel: {
 
   return pixel;
 }
-
+/**
+ * @returns the converted pixels in {R: number, G: number, B: number, A: number}.
+ *
+ * Follow conversion steps in CSS Color Module Level 4
+ * https://drafts.csswg.org/css-color/#predefined-to-predefined
+ * display-p3 and sRGB share the same white points.
+ */
 export function srgbToDisplayP3(pixel: {
   R: number;
   G: number;
--- a/src/webgpu/util/color_space_conversion.ts
+++ b/src/webgpu/util/color_space_conversion.ts
@@ -2,16 +2,12 @@ import { assert } from '../../common/util/util.js';
 
 import { multiplyMatrices } from './math.js';
 
-/**
- * The color space conversion function definations are fully aign with
- * CSS Color Module Level 4 Sample Code: https://drafts.csswg.org/css-color/#color-conversion-code
- */
+// The color space conversion function definations copies directly from
+// CSS Color Module Level 4 Sample Code: https://drafts.csswg.org/css-color/#color-conversion-code
 
-/**
- *  Sample code for color conversions
- * Conversion can also be done using ICC profiles and a Color Management System
- * For clarity, a library is used for matrix multiplication (multiply-matrices.js)
- */
+// Sample code for color conversions
+// Conversion can also be done using ICC profiles and a Color Management System
+// For clarity, a library is used for matrix multiplication (multiply-matrices.js)
 
 // sRGB-related functions
 
@@ -73,6 +69,7 @@ function lin_sRGB_to_XYZ(rgb: Array<Array<number>>) {
 
 /**
  * convert XYZ to linear-light sRGB
+ * using sRGB's own white, D65 (no chromatic adaptation)
  */
 function XYZ_to_lin_sRGB(XYZ: Array<Array<number>>) {
   const M = [
@@ -104,7 +101,7 @@ function gam_P3(RGB: Array<number>) {
 
 /**
  * convert an array of linear-light display-p3 values to CIE XYZ
- * using  D65 (no chromatic adaptation)
+ * using display-p3's D65 (no chromatic adaptation)
  * http://www.brucelindbloom.com/index.html?Eqn_RGB_XYZ_Matrix.html
  */
 function lin_P3_to_XYZ(rgb: Array<Array<number>>) {
@@ -120,6 +117,7 @@ function lin_P3_to_XYZ(rgb: Array<Array<number>>) {
 
 /**
  * convert XYZ to linear-light P3
+ * using display-p3's own white, D65 (no chromatic adaptation)
  */
 function XYZ_to_lin_P3(XYZ: Array<Array<number>>) {
   const M = [
--- a/src/webgpu/util/color_space_conversion.ts
+++ b/src/webgpu/util/color_space_conversion.ts
@@ -1,0 +1,190 @@
+import { assert } from '../../common/util/util.js';
+
+/**
+ * The color space conversion function definations are fully aign with
+ * CSS Color Module Level 4 Sample Code: https://drafts.csswg.org/css-color/#color-conversion-code
+ */
+
+/**
+ * Simple matrix (and vector) multiplication
+ * Warning: No error handling for incompatible dimensions!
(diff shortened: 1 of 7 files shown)


src/webgpu/web_platform/copyToTexture/canvas.spec.ts
```diff
@@ -4,16 +4,58 @@ copyToTexture with HTMLCanvasElement and OffscreenCanvas sources.
...
+
+    let canvasContext = null;
+    canvasContext = canvas.getContext('2d', { colorSpace }) as CanvasRenderingContext2D | null;
+
+    if (canvasContext === null) {
+      this.skip('onscreen canvas 2d context not available');
+    }
+
+    if (
+      typeof (canvasContext as any).getContextAttributes === 'undefined' ||
```
shaoboyan, line 43: I'm using `typeof canvasContext.getContextAttributes === 'undefined'` but always got compiler error said: No property named getContextAttributes in CanvasRenderingContext2D. Do you have any suggestions?

src/webgpu/web_platform/copyToTexture/canvas.spec.ts
```diff
@@ -239,7 +290,8 @@ class F extends CopyToTextureUtils {
     width: number,
     height: number
   ): Uint8ClampedArray {
-    return context.getImageData(0, 0, width, height).data;
+    // Always read back the raw data from canvas
+    return (context as any).getImageData(0, 0, width, height, { colorSpace: 'srgb' }).data;
```
shaoboyan, line 294: It always said "Expect 4 parameters but got 5". Do you have any suggestions?

shaoboyan:
The compatible Blink CL is here https://chromium-review.googlesource.com/c/chromium/src/+/3501779

src/webgpu/util/color_space_conversion.ts
```diff
@@ -0,0 +1,190 @@
...
+ * CSS Color Module Level 4 Sample Code: https://drafts.csswg.org/css-color/#color-conversion-code
+ */
+
+/**
+ * Simple matrix (and vector) multiplication
+ * Warning: No error handling for incompatible dimensions!
+ * @author Lea Verou 2020 MIT License
+ */
+// A is m x n. B is n x p. product is m x p.
+function multiplyMatrices(A: Array<Array<number>>, B: Array<Array<number>>) {
```
austinEng, line 14: multiplyMatrices is probably generally more useful. We can put it somewhere else like utils/math.ts ?

Please also add all exported helpers to helper_index.txt

src/webgpu/util/color_space_conversion.ts
```diff
@@ -0,0 +1,190 @@
...
+
+    if (abs > 0.0031308) {
+      return sign * (1.055 * Math.pow(abs, 1 / 2.4) - 0.055);
+    }
+
+    return 12.92 * val;
+  });
+}
+
+function lin_sRGB_to_XYZ(rgb: Array<Array<number>>) {
```
austinEng, line 74: use JS doc syntax, same for the rest of the helpers
/**
 *
 */

src/webgpu/web_platform/copyToTexture/canvas.spec.ts
austinEng: Both of these seem to be OK if we update tsc in package.json to 4.6.2 (latest) and run `npm install` again. There are several new typecheck errors after doing that though.

austinEng (commented):
Some comments - though I did not actually review the test content. I think Kai would be a better reviewer here.

shaoboyan:
@austinEng  I try to upgrade tsc but got lots of errors. Maybe we should use a seperate PR to do the upgrading.

kainino0x:
I'm working on a tsc upgrade, should be able to get it done today.

kainino0x:
TypeScript update landed, please merge `main` into your branch.

src/webgpu/util/color_space_conversion.ts
```diff
@@ -0,0 +1,195 @@
...
+ * where in-gamut values are in the range [0 - 1]
+ * to linear light (un-companded) form.
+ * https://en.wikipedia.org/wiki/SRGB
+ * Extended transfer function:
+ * for negative values,  linear portion is extended on reflection of axis,
+ * then reflected power function is used.
+ */
+function lin_sRGB(RGB: Array<number>) {
+  return RGB.map(val => {
+    const sign = val < 0 ? -1 : 1;
```
kainino0x, line 29: nit: This can be just `Math.sign(val)`, and can be moved after the `if` block.

src/webgpu/util/color_space_conversion.ts
```diff
@@ -0,0 +1,195 @@
...
+/**
+ * convert an array of sRGB values
+ * where in-gamut values are in the range [0 - 1]
+ * to linear light (un-companded) form.
+ * https://en.wikipedia.org/wiki/SRGB
+ * Extended transfer function:
+ * for negative values,  linear portion is extended on reflection of axis,
+ * then reflected power function is used.
+ */
+function lin_sRGB(RGB: Array<number>) {
```
kainino0x, line 27: `conversion.ts` already has `gammaCompress` and `gammaDecompress`, though unfortunately they behave somewhat differently (they're for srgb formats, not real color management), so they're not necessarily easy to clean up. We should at least leave a MAINTENANCE_TODO in both places pointing to the other, so we can maybe clean it up later.

src/webgpu/util/color_space_conversion.ts
```diff
@@ -0,0 +1,195 @@
...
+ * convert an array of linear-light sRGB values in the range 0.0-1.0
+ * to gamma corrected form
+ * https://en.wikipedia.org/wiki/SRGB
+ * Extended transfer function:
+ * For negative values, linear portion extends on reflection
+ * of axis, then uses reflected pow below that
+ */
+function gam_sRGB(RGB: Array<number>) {
+  return RGB.map(val => {
+    const sign = val < 0 ? -1 : 1;
```
kainino0x, line 50: same

src/webgpu/util/color_space_conversion.ts
```diff
@@ -0,0 +1,195 @@
...
+    const abs = Math.abs(val);
+
+    if (abs > 0.0031308) {
```
kainino0x, line 53: nit: I'd invert this `if` so it's parallel with the code in `lin_sRGB`.

src/webgpu/util/color_space_conversion.ts
kainino0x, line 27: nit: The name tells you what the output is, but isn't clear about the input. How about `extendedGammaCompressArray`, or just `extendedGammaCompress` and use `.map` with it later? Also then you don't need lin_P3/gam_P3 too.

nit: `RGB` should be lowercase since it's a local variable

src/webgpu/util/color_space_conversion.ts
```diff
@@ -0,0 +1,195 @@
...
+      return sign * (1.055 * Math.pow(abs, 1 / 2.4) - 0.055);
+    }
+
+    return 12.92 * val;
+  });
+}
+
+/**
+ * convert an array of linear-light sRGB values to CIE XYZ
+ * using sRGB's own white, D65 (no chromatic adaptation)
```
kainino0x, line 63: this is a very useful clarification, thanks for including it!

src/webgpu/util/color_space_conversion.ts
```diff
@@ -0,0 +1,195 @@
...
+  const M = [
+    [0.41239079926595934, 0.357584339383878, 0.1804807884018343],
+    [0.21263900587151027, 0.715168678767756, 0.07219231536073371],
+    [0.01933081871559182, 0.11919477979462598, 0.9505321522496607],
+  ];
+  return multiplyMatrices(M, rgb);
+}
+
+/**
+ * convert XYZ to linear-light sRGB
```
kainino0x, line 75: nit: Include the detail about D65 here too.

src/webgpu/util/copy_to_texture.ts
```diff
@@ -81,7 +109,9 @@ export class CopyToTextureUtils extends GPUTest {
     format: RegularTextureFormat,
     srcPremultiplied: boolean,
     dstPremultiplied: boolean,
-    isFlipY: boolean
+    isFlipY: boolean,
+    srcColorSpace: 'display-p3' | 'srgb' = 'srgb',
+    dstColorSpace: 'srgb' = 'srgb'
```
kainino0x, line 114: I think this would more precisely be:
```suggestion
    dstColorSpace: GPUPredefinedColorSpace = 'srgb'
```

src/webgpu/util/copy_to_texture.ts
```diff
@@ -104,17 +139,44 @@ export class CopyToTextureUtils extends GPUTest {
           A: orientedPixels[pixelPos * 4 + 3] / divide,
         };
 
-        if (!srcPremultiplied && dstPremultiplied) {
+        if (requireUnpremultiplyAlpha && rgba.A !== 0.0) {
```
kainino0x, line 142: Rather than silently skipping when A=0, should probably assert that R=G=B=0 as well.

(Normally I would say we should assert R<=A && G<=A && B<=A, but it occurs to me that would be valid if we ever have extended-gamut premultiplied values.)

src/webgpu/util/copy_to_texture.ts
```diff
@@ -165,31 +227,88 @@ export class CopyToTextureUtils extends GPUTest {
...
+          bytesPerRow / check.BYTES_PER_ELEMENT
+        );
+        const expRowF32 = new Float32Array(expRow.length);
+        const checkRowF32 = new Float32Array(checkRow.length);
+        checkRow.forEach((v: number, i: number) => {
+          checkRowF32[i] = float16BitsToFloat32(v);
+        });
+        expRow.forEach((v: number, i: number) => {
+          expRowF32[i] = float16BitsToFloat32(v);
+        });
```
kainino0x, line 271: optional: I usually use `for (const [i, v] of expRow.entries())` or a C-style loop, though this is fine too

src/webgpu/util/copy_to_texture.ts
```diff
@@ -165,31 +227,88 @@ export class CopyToTextureUtils extends GPUTest {
...
+        const checkResult = checkElementsBetween(checkRowF32, [
+          i => expRowF32[i] - 0.001,
+          i => expRow[i] + 0.001,
```
kainino0x, line 274:
```suggestion
          i => expRowF32[i] + 0.001,
```
I think. Perhaps worth renaming `expRow` to `expRowF16Bits` so this error would be more apparent.

src/webgpu/web_platform/copyToTexture/canvas.spec.ts
```diff
@@ -653,6 +770,109 @@ g.test('copy_contents_from_gpu_context_canvas')
...
+  
+    It creates HTMLCanvasElement/OffscreenCanvas with '2d' and 'colorSpace' attributes.
+    Use fillRect(2d context) to render red rect for top-left,
+    green rect for top-right, blue rect for bottom-left and white for bottom-right.
+  
+    Then call copyExternalImageToTexture() to do a full copy to the 0 mipLevel
+    of dst texture, and read the contents out to compare with the canvas contents.
+  
+    Do premultiply alpha in advance if 'premultipliedAlpha' in 'GPUImageCopyTextureTagged'
+    is set to 'ture' and do unpremultiply alpha if it is set to 'false'.
```
kainino0x, line 792:
```suggestion
    Provide premultiplied input if 'premultipliedAlpha' in 'GPUImageCopyTextureTagged'
    is set to 'true' and unpremultiplied input if it is set to 'false'.
```

kainino0x (approved):
LGTM w/ a few nits

src/webgpu/util/color_space_conversion.ts
shaoboyan: Ok, the origin plan here is that "we move the code from the CSS spec and that's why it is trusted". So I don't change anything(considering the spec is in draft and might have changes in that part).

kainino0x: Oh, sorry, I didn't realize this was copied directly from that spec. In that case, I'm happy to keep all this code exactly as it was in the source.

kainino0x: (except still best to have doc comments)

src/webgpu/util/color_space_conversion.ts
```diff
@@ -0,0 +1,195 @@
+import { assert } from '../../common/util/util.js';
+
+import { multiplyMatrices } from './math.js';
+
+/**
+ * The color space conversion function definations are fully aign with
```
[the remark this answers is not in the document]
kainino0x: Actually, should edit this comment to make it clearer that this is copied directly from that link.

src/webgpu/web_platform/copyToTexture/canvas.spec.ts
shaoboyan: And we don't need this anymore! :)

src/webgpu/util/color_space_conversion.ts
shaoboyan: Keep the form because the code is directly copy from CSS4 spec(Simple to update if CSS4 spec update the algorithm)

src/webgpu/util/color_space_conversion.ts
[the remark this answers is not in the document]
shaoboyan: Done.

src/webgpu/util/color_space_conversion.ts
shaoboyan: Done.

src/webgpu/util/color_space_conversion.ts
shaoboyan: Make sense, added it! Done.

src/webgpu/util/copy_to_texture.ts
shaoboyan: Unfortunately, current GPUPredefinedColorSpace only have 'srgb', add a TODO for this.

src/webgpu/util/copy_to_texture.ts
shaoboyan: Done.

src/webgpu/util/copy_to_texture.ts
shaoboyan: Update the name and fix this bug. Thanks!

src/webgpu/web_platform/copyToTexture/canvas.spec.ts
shaoboyan: Done for other places too.

src/webgpu/util/copy_to_texture.ts
shaoboyan: Updated!

src/webgpu/util/color_space_conversion.ts
```diff
@@ -0,0 +1,195 @@
...
+ * CSS Color Module Level 4 Sample Code: https://drafts.csswg.org/css-color/#color-conversion-code
+ */
```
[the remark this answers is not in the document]
shaoboyan: Ack, Thanks!

src/webgpu/util/math.ts
```diff
@@ -289,3 +289,26 @@ export function biasedRange(a: number, b: number, num_steps: number): Array<numb
     lerp(a, b, Math.pow(lerp(0, 1, i / (num_steps - 1)), c))
   );
 }
+
+/**
+ * @returns the result matrix in Array<Array<number>> type.
+ *
+ * Simple matrix (and vector) multiplication
+ * Warning: No error handling for incompatible dimensions!
+ * @author Lea Verou 2020 MIT License
```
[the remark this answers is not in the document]
shaoboyan: Change the implementation to the basic C-style loop version. Done.

shaoboyan:
One problem here:
The test runs super slow(I think ~ a minute) with f16 formats.

src/webgpu/util/copy_to_texture.ts
kainino0x: Right, which is exactly why it makes sense to use it here. Then when we update GPUPredefinedColorSpace this will get updated automatically.

src/webgpu/util/color_space_conversion.ts
```diff
@@ -0,0 +1,193 @@
+import { assert } from '../../common/util/util.js';
+
+import { multiplyMatrices } from './math.js';
+
+// The color space conversion function definations copies directly from
```
kainino0x, line 5:
```suggestion
// These color space conversion function definitions are copied directly from
```

src/webgpu/util/copy_to_texture.ts
```diff
@@ -74,14 +102,18 @@ export class CopyToTextureUtils extends GPUTest {
...
     sourcePixels: Uint8ClampedArray,
     width: number,
     height: number,
     format: RegularTextureFormat,
     srcPremultiplied: boolean,
     dstPremultiplied: boolean,
-    isFlipY: boolean
+    isFlipY: boolean,
+    srcColorSpace: GPUPredefinedColorSpace | 'display-p3' = 'srgb',
+    dstColorSpace: GPUPredefinedColorSpace = 'srgb'
```
kainino0x, line 116: I only meant to suggest changing dstColorSpace, not srcColorSpace, since srcColorSpace is not in principle tied to GPUPredefinedColorSpace (it is actually tied to the canvas enum `PredefinedColorSpace` but I'm guessing that's not defined in TypeScript yet).
```suggestion
    srcColorSpace: 'srgb' | 'display-p3' = 'srgb',
    dstColorSpace: GPUPredefinedColorSpace = 'srgb'
```

src/webgpu/util/copy_to_texture.ts
```diff
@@ -74,14 +102,18 @@ export class CopyToTextureUtils extends GPUTest {
...
   getExpectedPixels(
     sourcePixels: Uint8ClampedArray,
     width: number,
     height: number,
     format: RegularTextureFormat,
     srcPremultiplied: boolean,
     dstPremultiplied: boolean,
-    isFlipY: boolean
+    isFlipY: boolean,
+    srcColorSpace: GPUPredefinedColorSpace | 'display-p3' = 'srgb',
```
kainino0x, line 115: I only meant to suggest changing dstColorSpace, not srcColorSpace, since srcColorSpace is not in principle tied to GPUPredefinedColorSpace. It is actually tied to the canvas enum `PredefinedColorSpace` but I was assuming that's not defined in TypeScript yet. But it appears it is!

```suggestion
    srcColorSpace: PredefinedColorSpace = 'srgb',
```

kainino0x (approved):

kainino0x:
> The test runs super slow(I think ~ a minute) with f16 formats.

Do you know what causes this? If it's slowness in the implementation, we can mark the test as slow in our expectations (hopefully - though we should consider doing so before landing this PR so that the Chrome WebGPU Gardener won't have to debug it).
If it's slowness in the test we should see if we can optimize it. I'm guessing that the float16BitsToFloat32 helper is a bottleneck.

src/webgpu/util/copy_to_texture.ts
shaoboyan: Ah, thanks! I mixed these two `GPUPredefinedColorSpace` and `PredefinedColorSpace`. Thanks for fixing it

shaoboyan:
> I'm guessing that the float16BitsToFloat32 helper is a bottleneck.
Yes, and the `actual` and `expected` arrays needs to reinterpret to Uin16Array first and then call float16ToFloat32 for each entries. 
Since the other format works fine, I think this is the bottleneck.

src/webgpu/util/copy_to_texture.ts
```diff
@@ -165,31 +236,93 @@ export class CopyToTextureUtils extends GPUTest {
...
+          bytesPerRow / expF16bits.BYTES_PER_ELEMENT
+        );
+
+        const checkRowF16bits = checkF16bits.subarray(
+          (y * bytesPerRow) / checkF16bits.BYTES_PER_ELEMENT,
+          bytesPerRow / checkF16bits.BYTES_PER_ELEMENT
+        );
+
+        const expRowF32bits = new Float32Array(expRowF16bits.length);
+        const checkRowF32bits = new Float32Array(checkRowF16bits.length);
```
kainino0x, line 276: These aren't the bits for f32 values, they're actual f32 values. Should just be expRowF32/checkRowF32.

src/webgpu/util/copy_to_texture.ts
```diff
@@ -74,14 +102,18 @@ export class CopyToTextureUtils extends GPUTest {
     );
   }
 
+  // MAINTENANCE_TODO: type GPUPredefinedColorSpace only have 'srgb', remove the
+  // 'display-p3' type after it is included in GPUPredefinedColorSpace.
```
kainino0x, line 106:
```suggestion
```

src/webgpu/util/copy_to_texture.ts
```diff
@@ -165,31 +236,93 @@ export class CopyToTextureUtils extends GPUTest {
...
-          i => (expRow[i] > 0 ? expRow[i] - 1 : expRow[i]),
-          i => expRow[i] + 1,
+        const expRowF32bits = expF32bits.subarray(
+          (y * bytesPerRow) / expF32bits.BYTES_PER_ELEMENT,
+          bytesPerRow / expF32bits.BYTES_PER_ELEMENT
+        );
+        const checkRowF32bits = checkF32bits.subarray(
+          (y * bytesPerRow) / checkF32bits.BYTES_PER_ELEMENT,
+          bytesPerRow / checkF32bits.BYTES_PER_ELEMENT
+        );
```
kainino0x, line 312: Same here

src/webgpu/util/copy_to_texture.ts
```diff
@@ -130,13 +201,13 @@ export class CopyToTextureUtils extends GPUTest {
...
-    expected: ArrayBufferView,
+    expected: TypedArrayBufferView,
     width: number,
     height: number,
     bytesPerPixel: number,
-    isFp16: boolean
+    dstFormat: RegularTextureFormat
   ): void {
-    const exp = new Uint8Array(expected.buffer, expected.byteOffset, expected.byteLength);
+    //const exp = new Uint8Array(expected.buffer, expected.byteOffset, expected.byteLength);
```
kainino0x, line 210: Stray commented code

src/webgpu/util/copy_to_texture.ts
```diff
@@ -104,17 +141,51 @@ export class CopyToTextureUtils extends GPUTest {
           A: orientedPixels[pixelPos * 4 + 3] / divide,
         };
 
-        if (!srcPremultiplied && dstPremultiplied) {
+        if (requireUnpremultiplyAlpha && rgba.A !== 0.0) {
```
kainino0x, line 144:
```suggestion
        if (requireUnpremultiplyAlpha) {
```

kainino0x (commented):
Apparently I forgot to complete my re-review, sorry.

> Since the other format works fine, I think this is the bottleneck.

OK. Since we know it's slow, do you think we could disable the f16 tests in this PR so we can land it, and re-enable them (hopefully with optimizations) in a separate PR?

src/webgpu/util/copy_to_texture.ts
shaoboyan: It seems I mixed f32 and f32bits to the same meaning as the bit length. Thanks !

shaoboyan:
>OK. Since we know it's slow, do you think we could disable the f16 tests in this PR so we can land it, and re-enable them (hopefully with optimizations) in a separate PR?

The f16 expect logic will cover more than color_space conversion cases. So I try to fix it by comparing the f16bits and handling ~0 values. PTAL again, thanks!

shaoboyan:
After this fixing, f16 format cases could finished in ~4s on my local machine.

src/webgpu/util/copy_to_texture.ts
kainino0x: Ah, you're right it's very confusing!

Issue #881, opened 2022-01-13
Consolidate helpers for texture content checking


kainino0x:
Search CTS for ~`crbug.com/dawn/868`~ `MAINTENANCE_TODO(#881)` for some places that could use helpers.

kainino0x:
WIP is at https://github.com/gpuweb/cts/compare/main...kainino0x:texture-checking

kainino0x:
Filed three draft issues as followups.

PR #1068, opened 2022-03-16
Assorted helpers used for texture checking

kainino0x:
These are some assorted helpers that are used in the texture checking helpers in #1055 (WIP).
The commits are separate changes and may be reviewed separately if desired.

Issue: #881

<hr>

### DIFF
--- a/src/webgpu/util/conversion.ts
+++ b/src/webgpu/util/conversion.ts
@@ -111,6 +111,20 @@ export const kFloat32Format = { signed: 1, exponentBits: 8, mantissaBits: 23, bi
 /** FloatFormat defining IEEE754 16-bit float. */
 export const kFloat16Format = { signed: 1, exponentBits: 5, mantissaBits: 10, bias: 15 } as const;
 
+const workingData = new ArrayBuffer(4);
+const workingDataU32 = new Uint32Array(workingData);
+const workingDataF32 = new Float32Array(workingData);
+/** Bitcast u32 (represented as integer Number) to f32 (represented as floating-point Number). */
+export function float32BitsToNumber(bits: number): number {
+  workingDataU32[0] = bits;
+  return workingDataF32[0];
+}
+/** Bitcast f32 (represented as floating-point Number) to u32 (represented as integer Number). */
+export function numberToFloat32Bits(number: number): number {
+  workingDataF32[0] = number;
+  return workingDataU32[0];
+}
+
 /**
  * Decodes an IEEE754 float with the supplied format specification into a JS number.
  *
--- a/src/webgpu/util/conversion.ts
+++ b/src/webgpu/util/conversion.ts
@@ -101,11 +101,62 @@ export function float32ToFloat16Bits(n: number) {
  * Decodes an IEEE754 16 bit floating point number into a JS `number` and returns.
  */
 export function float16BitsToFloat32(float16Bits: number): number {
-  const buf = new DataView(new ArrayBuffer(Float32Array.BYTES_PER_ELEMENT));
-  // shift exponent and mantissa bits and fill with 0 on right, shift sign bit
-  buf.setUint32(0, ((float16Bits & 0x7fff) << 13) | ((float16Bits & 0x8000) << 16), true);
-  // shifting for bias different: f16 uses a bias of 15, f32 uses a bias of 127
-  return buf.getFloat32(0, true) * 2 ** (127 - 15);
+  return floatBitsToNumber(float16Bits, kFloat16Format);
+}
+
+type FloatFormat = { signed: 0 | 1; exponentBits: number; mantissaBits: number; bias: number };
+
+/** FloatFormat defining IEEE754 32-bit float. */
+export const kFloat32Format = { signed: 1, exponentBits: 8, mantissaBits: 23, bias: 127 } as const;
+/** FloatFormat defining IEEE754 16-bit float. */
+export const kFloat16Format = { signed: 1, exponentBits: 5, mantissaBits: 10, bias: 15 } as const;
+
+/**
+ * Decodes an IEEE754 float with the supplied format specification into a JS number.
+ *
+ * The format MUST be no larger than a 32-bit float.
+ */
+export function floatBitsToNumber(bits: number, fmt: FloatFormat): number {
+  // Pad the provided bits out to f32, then convert to a `number` with the wrong bias.
+  // E.g. for f16 to f32:
+  // - f16: S    EEEEE MMMMMMMMMM
+  //        ^ 000^^^^^ ^^^^^^^^^^0000000000000
+  // - f32: S eeeEEEEE MMMMMMMMMMmmmmmmmmmmmmm
+
+  const kNonSignBits = fmt.exponentBits + fmt.mantissaBits;
+  const kNonSignBitsMask = (1 << kNonSignBits) - 1;
+  const expAndMantBits = bits & kNonSignBitsMask;
+  let f32BitsWithWrongBias = expAndMantBits << (kFloat32Format.mantissaBits - fmt.mantissaBits);
+  f32BitsWithWrongBias |= (bits << (31 - kNonSignBits)) & 0x8000_0000;
+  const numberWithWrongBias = float32BitsToNumber(f32BitsWithWrongBias);
+  return numberWithWrongBias * 2 ** (kFloat32Format.bias - fmt.bias);
+}
+
+/**
+ * Given a floating point number (as an integer representing its bits), computes how many ULPs it is
+ * from zero.
+ *
+ * Subnormal numbers are skipped, so that 0 is one ULP from the minimum normal number.
+ * Subnormal values are flushed to 0.
+ * Positive and negative 0 are both considered to be 0 ULPs from 0.
+ */
+export function floatBitsToNormalULPFromZero(bits: number, fmt: FloatFormat): number {
+  const mask_sign = fmt.signed << (fmt.exponentBits + fmt.mantissaBits);
+  const mask_expt = ((1 << fmt.exponentBits) - 1) << fmt.mantissaBits;
+  const mask_mant = (1 << fmt.mantissaBits) - 1;
+  const mask_rest = mask_expt | mask_mant;
+
+  assert(fmt.exponentBits + fmt.mantissaBits <= 31);
+
+  const sign = bits & mask_sign ? -1 : 1;
+  const rest = bits & mask_rest;
+  const subnormal_or_zero = (bits & mask_expt) === 0;
+  const infinity_or_nan = (bits & mask_expt) === mask_expt;
+  assert(!infinity_or_nan, 'no ulp representation for infinity/nan');
+
+  // The first normal number is mask_mant+1, so subtract mask_mant to make min_normal - zero = 1ULP.
+  const abs_ulp_from_zero = subnormal_or_zero ? 0 : rest - mask_mant;
+  return sign * abs_ulp_from_zero;
 }
 
 /**
--- a/src/webgpu/util/math.ts
+++ b/src/webgpu/util/math.ts
@@ -34,8 +34,11 @@ export function clamp(n: number, { min, max }: { min: number; max: number }): nu
 }
 
 /**
- * @returns the Units of Last Place difference between the numbers a and b.
- * If either `a` or `b` are not finite numbers, then diffULP() returns Infinity.
+ * @returns the (absolute) Units of Last Place difference between the float32 numbers a and b, taken
+ * as JS doubles. If either `a` or `b` are not finite numbers, then diffULP() returns Infinity.
+ *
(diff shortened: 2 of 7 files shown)


kainino0x:
@shaoboyan PTAL at this first round of helpers.

> The commits are separate changes and may be reviewed separately if desired.

I'm going to add some tests for floatBitsToNumber since that's difficult to be sure is correct.

src/webgpu/util/conversion.ts
```diff
@@ -101,11 +101,66 @@ export function float32ToFloat16Bits(n: number) {
...
+  const mask_expt = ((1 << fmt.exponentBits) - 1) << fmt.mantissaBits;
+  const mask_mant = (1 << fmt.mantissaBits) - 1;
+  const mask_rest = mask_expt | mask_mant;
+
+  assert(fmt.exponentBits + fmt.mantissaBits <= 31);
+
+  const sign = bits & mask_sign ? -1 : 1;
+  const rest = bits & mask_rest;
+  const subnormal_or_zero = (bits & mask_expt) === 0;
+  const infinity_or_nan = (bits & mask_expt) === (-1 & mask_expt);
```
shaoboyan, line 158: (bits & mask_expt) === mask_expt ?

src/webgpu/util/conversion.ts
```diff
@@ -101,11 +101,66 @@ export function float32ToFloat16Bits(n: number) {
...
+
+/**
+ * Given a floating point number (as an integer representing its bits), computes how many ULPs it is
+ * from zero.
+ *
+ * Subnormal numbers are skipped, so that 0 is one ULP from the minimum normal number.
+ * Subnormal values are flushed to 0.
+ * Positive and negative 0 are both considered to be 0 ULPs from 0.
+ */
+export function floatBitsToNormalULPFromZero(bits: number, fmt: FloatFormat): number {
```
shaoboyan, line 147: For copy_to_texture 16-bit float(and 32-bit float) result comparasion. With this helper function, it seems that we could check the result by:
```
// Assume expect is larger than actual
floatBitsToNormalULPFromZero(Uint16(expected) - Uint16(actual), kFloat16Format) < constant?
```
Am I right?

shaoboyan: If this is the direction, I'm a bit worry about the case running time.
As you may know, current compare logic is [simple and hack](https://github.com/gpuweb/cts/pull/1043/files#diff-0781362fbbe96f1c4922e18e4ccead52e6c15573ed9bfa091c5b039c940e4d08R268)
But it still took longer time because it requires a [buffer view reinterpretation](https://github.com/gpuweb/cts/pull/1043/files#diff-0781362fbbe96f1c4922e18e4ccead52e6c15573ed9bfa091c5b039c940e4d08R246).
If we add an extra ops `floatBitsToNormalULPFromZero`, I think it took longer time.

So maybe a bit hack but do you think it is possible that we took everything as Uint8 as input and do the bit ops? It will save the buffer view reinterpretation.

shaoboyan: And another option is to save time on the other place rather than the float compare.

src/webgpu/util/math.ts
```diff
@@ -317,3 +320,9 @@ export function multiplyMatrices(
 
   return product;
 }
+
+/** Sign-extend the `bits`-bit number `n` to a 32-bit signed integer. */
+export function signExtend(n: number, bits: number): number {
```
shaoboyan, line 325: Why we need this?

shaoboyan (commented):
I've reviewed float helpers and `getTextureSubCopyLayout` implementations. Looks great!

src/webgpu/util/math.ts
kainino0x: This is used in the next PR, but I included it here since it's easy to review on its own.

src/webgpu/util/conversion.ts
kainino0x: Performance is definitely a potential issue and I haven't investigated it enough yet, thanks for highlighting it.
It's even worse than your example code, because it's more like `floatBitsToNormalULPFromZero(expected) - floatBitsToNormalULPFromZero(actual)`.

We have a `diffULP` helper already for directly determining the ULPs between two values without computing them relative to zero. I'll investigate the performance and see what can be done.

kainino0x: As a point of comparison,
`webgpu:web_platform,copyToTexture,ImageBitmap:from_ImageData:alpha="none";orientation="none";srcDoFlipYDuringCopy=true;dstColorFormat="rgba16float";*`
before: 2500ms each
after: 4300ms each

Not quite as bad as I expected, but could probably be better

kainino0x: > We have a diffULP helper already for directly determining the ULPs between two values without computing them relative to zero.

Reviewing, I don't think this would help. It's doing almost the same stuff.

> But it still took longer time because it requires a [buffer view reinterpretation](https://github.com/gpuweb/cts/pull/1043/files#diff-0781362fbbe96f1c4922e18e4ccead52e6c15573ed9bfa091c5b039c940e4d08R246).

And reinterpretation would still be happening if I used diffULP. That said, I wouldn't expect reinterpretation to be slow, so not sure yet where exactly the slowdown comes from.

kainino0x: Looking at https://github.com/gpuweb/cts/pull/1043/commits/74816819e165625a094f6060d8134281f21ecb46, I'm guessing it was `float16BitsToFloat32`/`float16BitsToFloat32`. Which is probably a little more expensive than `floatBitsToNormalULPFromZero` though I wouldn't expect it to be that much worse.

shaoboyan: >Looking at https://github.com/gpuweb/cts/commit/74816819e165625a094f6060d8134281f21ecb46, I'm guessing it was float16BitsToFloat32/float16BitsToFloat32. Which is probably a little more expensive than floatBitsToNormalULPFromZero though I wouldn't expect it to be that much worse.

Yes, removing these two helper functions help accelerated the tests a lot but it is still worse than the [Uint8 comparation](https://github.com/gpuweb/cts/commit/74816819e165625a094f6060d8134281f21ecb46#diff-0781362fbbe96f1c4922e18e4ccead52e6c15573ed9bfa091c5b039c940e4d08L321) a lot (on my machine) but the same performance as [float32 comparation](https://github.com/gpuweb/cts/commit/74816819e165625a094f6060d8134281f21ecb46#diff-0781362fbbe96f1c4922e18e4ccead52e6c15573ed9bfa091c5b039c940e4d08L309). So I suspect this is due to the reinterpretation (But I think it shouldn't took long time).

 >webgpu:web_platform,copyToTexture,ImageBitmap:from_ImageData:alpha="none";orientation="none";srcDoFlipYDuringCopy=true;dstColorFormat="rgba16float";*
before: 2500ms each
after: 4300ms each

Thanks for testing! I understand that 4300ms is the time that applying `diffULP`, right?

src/webgpu/util/math.ts
shaoboyan: Ack.

src/webgpu/util/conversion.ts
kainino0x: I dug down into the performance of the ImageBitmap:from_ImageData test and found that it was a simple matter of implementing this optimization I had left for myself (in #1055):
```ts
    // MAINTENANCE_TODO: Could be faster to actually implement numberToBits directly.
    numberToBits: (components: PerTexelComponent<number>) =>
      ret.unpackBits(new Uint8Array(ret.pack(encode(components)))),
```
before: 2500ms
draft: 4300ms
after: 2030ms!

kainino0x: For another test case:
https://gpuweb.github.io/cts/standalone/?runnow=1&worker=0&debug=1&q=webgpu:web_platform,copyToTexture,canvas:copy_contents_from_gpu_context_canvas:canvasType=%22onscreen%22;srcAndDstInSameGPUDevice=true;dstColorFormat=%22rgba16float%22;srcPremultiplied=true;dstPremultiplied=true;srcDoFlipYDuringCopy=true

before: 1490ms
after: 1410ms

kainino0x:
A few more perf numbers:
Optimizing floatBitsToNumber brought one testcase's total runtime from 4250ms to 3500ms. This was before I did other optimizations. The performance was extremely close to the original float16BitsToFloat32 even though that was specialized to float16, so I didn't add back the specialization.

src/webgpu/util/conversion.ts
shaoboyan: Cool!

src/webgpu/util/conversion.ts
```diff
@@ -111,6 +111,18 @@ export const kFloat32Format = { signed: 1, exponentBits: 8, mantissaBits: 23, bi
 /** FloatFormat defining IEEE754 16-bit float. */
 export const kFloat16Format = { signed: 1, exponentBits: 5, mantissaBits: 10, bias: 15 } as const;
 
+const workingData = new ArrayBuffer(4);
+const workingDataU32 = new Uint32Array(workingData);
+const workingDataF32 = new Float32Array(workingData);
+export function float32BitsToNumber(bits: number): number {
+  workingDataU32[0] = bits;
```
shaoboyan, line 118: I think the take away here is don't create temporary TypedArrayBuffer for reinterpretation.

shaoboyan (commented):
LGTM, thanks for the iteration!

shaoboyan (approved):

src/webgpu/util/conversion.ts
kainino0x: I measured the performance of this briefly while testing a bunch of other things. It didn't have a large effect, but it was enough that it seemed worth using.

However I tried measuring it against just now (by just moving `workingData*` inside these functions) and I wasn't able to measure a difference... ~2020ms either way. Maybe it got optimized better somehow when written this way?

kainino0x: Oh, the test case I was using is no longer bottlenecked on this function. I tested a different test case (rgba32float) which is, and the results are good.
`webgpu:web_platform,copyToTexture,ImageBitmap:from_ImageData:alpha="premultiply";orientation="flipY";srcDoFlipYDuringCopy=false;dstColorFormat="rgba32float";dstPremultiplied=true`

preallocated (this PR): 1640ms
late allocated (same but workingData moved inside the function): 2130ms
array-initialized (`new Float32Array(new Uint32Array([bits]).buffer)[0]`): 2260ms

incidentally I realized one of these functions is implemented wrong, so fixing that.

shaoboyan: Ok, so it seems that the take away is still correct!

austinEng:
@kainino0x after this PR, webgpu:api,operation,command_buffer,image_copy:mip_levels: tests are hitting an `assert`

1 | + This is a testharness.js-based test.
2 | + FAIL ;dimension="2d" assert_unreached:
3 | +   - INFO: subcase: copySizeInBlocks={"width":5,"height":4,"depthOrArrayLayers":1};originInBlocks={"x":3,"y":2,"z":0};mipLevel=1;textureSize=[64,48,1]
4 | +   - INFO: subcase: copySizeInBlocks={"width":5,"height":4,"depthOrArrayLayers":1};originInBlocks={"x":3,"y":2,"z":0};mipLevel=1;textureSize=[60,48,1]
5 | +   - EXCEPTION: copySize must be a multiple of the block size
6 | +         at assert (https://web-platform.test:8444/gen/third_party/webgpu-cts/src/common/util/util.js:21:15)
7 | +         at getTextureSubCopyLayout (https://web-platform.test:8444/gen/third_party/webgpu-cts/src/webgpu/util/texture/layout.js:32:5)
8 | +         at getTextureCopyLayout (https://web-platform.test:8444/gen/third_party/webgpu-cts/src/webgpu/util/texture/layout.js:19:20)
9 | +         at ImageCopyTest.uploadTextureAndVerifyCopy (https://web-platform.test:8444/gen/third_party/webgpu-cts/src/webgpu/api/operation/command_buffer/image_copy.spec.js:298:47)
10 | +         at RunCaseSpecific.fn (https://web-platform.test:8444/gen/third_party/webgpu-cts/src/webgpu/api/operation/command_buffer/image_copy.spec.js:1095:7)

austinEng:
Example https://gpuweb.github.io/cts/standalone/?runnow=1&worker=0&debug=0&q=webgpu:api,operation,command_buffer,image_copy:mip_levels:initMethod=%22WriteTexture%22;checkMethod=%22FullCopyT2B%22;format=%22bc1-rgba-unorm%22;dimension=%222d%22

kainino0x:
> * [x]  Test behaves as expected in a WebGPU implementation. (If not passing, explain above.)

I tested a few tests but not all. This is a clear fix, but I didn't test everything else so I'm not sure this fixes all of the regressions in #1068.

kainino0x:
Fix for at least that bug in https://github.com/gpuweb/cts/pull/1077

Issue #1066, opened 2022-03-15
op: explicitly test WebGPU is available on DedicatedWorkers


greggman:
Right now the whole CTS can be switched to run on a worker, but we need one test that makes sure some basic functionality works on DedicatedWorkers regardless of that switch, because we don't have a good story yet for WPT itself actually requiring any other tests to pass on workers.

PR #1075, opened 2022-03-16
Implement Basic Worker WebGPU Test

greggman:
A Basic Worker Test

Issue: #1066

Notes: There's a bunch of issues with finding the path to the worker. The issue is the code is mostly converted to forms that do not support `import.meta` when run in node via ts-node. This works for test_worker.js because it is excluded from conversion via node.tsconfig.json. Trying to exclude worker_launcher.ts failed because tsc sees what was a static import and imports it anyway. It works for `test_worker.js` because `standalone.js` is also excluded in node.tsconfig.json

Other solutions tried:

Tried solutions listed here: https://github.com/TypeStrong/ts-node/discussions/1290, No luck but maybe I just did something wrong.

Tried switching ts-node to use esm instead of commonjs. That included writing a custom resolver and/or using ts-node's resolver as in `node --loader ts-node/esm`. Failed for various reasons.

Tried putting the contents of `worker.js` in a string and creating a dataUrl or Blob url. It got further than the previous solutions but both failed because node doesn't support Blob and Worker as globals. Tried making them globals via setup-ts-in-node.js but it still failed because node's Worker doesn't support dataUrls and/or Blobs as input unless you pass in special flags. Tried setting the special flags but typescript complains that those special flags are invalid.

Rather than spending a bunch of time trying to make cross platform worker wrappers I just made it so the test is skipped in node.

<hr>

### DIFF
--- a/node.tsconfig.json
+++ b/node.tsconfig.json
@@ -11,6 +11,7 @@
   "exclude": [
     "src/common/runtime/wpt.ts",
     "src/common/runtime/standalone.ts",
-    "src/common/runtime/helper/test_worker.ts"
+    "src/common/runtime/helper/test_worker.ts",
+    "src/webgpu/web_platform/worker/worker_launcher.ts"
   ]
 }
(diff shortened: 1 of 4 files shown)


austinEng (approved):
LGTM! Makes sense to not expose it to node

kainino0x (approved):

kainino0x:
lgtm, sorry about the weird build issues! Hope we can move off of the commonjs target someday.

PR #1076, opened 2022-03-16
wgsl: Add validation tests for tokenization

jrprice:
Covers the following cases:
- Empty module
- Null characters
- Whitespace (delimiters)
- Comments
- Identifiers

Tested and mostly passing with ToT Dawn+Tint, with the following exceptions:
- Null characters are terminating the source string rather than being rejected
- The `enable` keyword is not yet reserved by Tint

<hr>

### DIFF
--- a/src/webgpu/shader/validation/tokens.spec.ts
+++ b/src/webgpu/shader/validation/tokens.spec.ts
@@ -92,7 +92,7 @@ const kValidIdentifiers = new Set(['foo', 'Foo', '_foo0', '_0foo', 'foo__0']);
 const kInvalidIdentifiers = new Set([
   '_', // Single underscore is a syntactic token for phony assignment.
   '__foo', // Leading double underscore is reserved.
-  '0foo', // Must start with single underscore or alpha-numeric.
+  '0foo', // Must start with single underscore or a letter.
   // No punctuation:
   'foo.bar',
   'foo-bar',
--- a/src/webgpu/shader/validation/tokens.spec.ts
+++ b/src/webgpu/shader/validation/tokens.spec.ts
@@ -1,0 +1,187 @@
+export const description = `Validation tests for tokenization`;
+
+import { makeTestGroup } from '../../../common/framework/test_group.js';
+
+import { ShaderValidationTest } from './shader_validation_test.js';
+
+export const g = makeTestGroup(ShaderValidationTest);
+
+g.test('empty')
+  .desc(`Test that an empty source file is consumed successfully.`)
+  .fn(t => {
+    t.expectCompileResult(true, '');
+  });
+
+g.test('null_characters')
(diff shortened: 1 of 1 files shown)


src/webgpu/shader/validation/tokens.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+  )
+  .fn(t => {
+    let code = '';
+    if (t.params.placement === 'comment') {
+      code = `// Here is a ${t.params.contains_null ? '\0' : 'Z'} character`;
+    } else if (t.params.placement === 'delimiter') {
+      code = `let${t.params.contains_null ? '\0' : ' '}name : i32 = 0;`;
+    } else if (t.params.placement === 'eol') {
+      code = `let name : i32 = 0;${t.params.contains_null ? '\0' : ''}`;
+    }
```
jrprice, line 31: @kainino0x or somebody else that understands Javascript: I'm not actually sure whether this is correct or not. Should these JS strings containing null characters get passed through the WebGPU implementation verbatim, or does something else cause them to be null terminated?

kainino0x: Not 100% sure if this answers the question, but:

JS strings are not null-terminated, internally they're essentially a `vector<uint16_t>`.
When we get a string in Blink for example, it arrives as WTF::String or something like that, and we can convert to UTF-8, and NULs inside the string get carried through.
(NUL is the 0 codepoint, encoded as 0x0000 in utf-16 and 0x00 in utf-8. The UTF-8 string resulting from that conversion should include any NULs and any text after them, and if it's std::string, there's also an extra NUL past the end for null-termination).

jrprice: Resolved offline - the tests are fine, there's an issue with our stack that we can address later.

src/webgpu/shader/validation/tokens.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+ ${t.params.terminated ? '*/' : ''}`;
+
+    t.expectCompileResult(t.params.terminated, code);
+  });
+
+const kValidIdentifiers = new Set(['foo', 'Foo', '_foo0', '_0foo', 'foo__0']);
+const kInvalidIdentifiers = new Set([
+  '_', // Single underscore is a syntactic token for phony assignment.
+  '__foo', // Leading double underscore is reserved.
+  '0foo', // Must start with single underscore or alpha-numeric.
```
zoddicus, line 95: I am unclear why this invalid. `0foo` looks like it starts with an alpha-numeric character, `0`

jrprice: The comment was wrong - the first character has to be underscore or a letter, not alphanumeric. I've fixed the comment.

zoddicus (approved):

PR #1089, opened 2022-03-18
color_space_conversion: use exact rational matrix formulations

kainino0x:
On a few tests, the conversion result was exactly the same down to the ULP. But these make a nice reference implementation since they can be copied to other contexts with arbitrary precisions.

Unfortunately, since the results didn't change, this didn't fix relatively large deltas I saw in the G channel of these tests (not huge, 0.0007). It's unclear whether we can fix that or if we'll keep those thresholds (in #1055).

Issue: #913

<hr>

### DIFF
--- a/src/webgpu/util/color_space_conversion.ts
+++ b/src/webgpu/util/color_space_conversion.ts
@@ -4,10 +4,11 @@ import { multiplyMatrices } from './math.js';
 
 // These color space conversion function definitions are copied directly from
 // CSS Color Module Level 4 Sample Code: https://drafts.csswg.org/css-color/#color-conversion-code
-// *EXCEPT* the conversion matrices are replaced with exact rational forms computed using
-// this Rust crate: https://crates.io/crates/rgb_derivation
-// as described for sRGB on this page: https://mina86.com/2019/srgb-xyz-matrix/
-// but using the numbers from the CSS spec: https://github.com/kainino0x/exact_css_xyz_matrices
+// *EXCEPT* the conversion matrices are replaced with exact rational forms computed here:
+// https://github.com/kainino0x/exact_css_xyz_matrices
+//   using this Rust crate: https://crates.io/crates/rgb_derivation
+//   as described for sRGB on this page: https://mina86.com/2019/srgb-xyz-matrix/
+//   but using the numbers from the CSS spec: https://www.w3.org/TR/css-color-4/#predefined
 
 // Sample code for color conversions
 // Conversion can also be done using ICC profiles and a Color Management System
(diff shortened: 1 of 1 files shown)


shaoboyan:
Recalled that the diff might also generated by the chromium implementation. It always use XYZ D50 color space as intermediate color space to calculate the conversion matrix. This might introduce D65 -> D50 conversion with Bradford chromatic adaptation(Not 100% sure uses this method).

shaoboyan:
[Bradford chromatic adaptation](http://www.brucelindbloom.com/index.html?Eqn_ChromAdapt.html) which is also used by CSS4 spec. ([More info](http://www.brucelindbloom.com/index.html?ChromAdaptEval.html)).
```
function D65_to_D50(XYZ) {
	// Bradford chromatic adaptation from D65 to D50
	// The matrix below is the result of three operations:
	// - convert from XYZ to retinal cone domain
	// - scale components from one reference white to another
	// - convert back to XYZ
	// http://www.brucelindbloom.com/index.html?Eqn_ChromAdapt.html
```

shaoboyan (approved):
LGTM!

src/webgpu/util/color_space_conversion.ts
```diff
@@ -4,6 +4,10 @@ import { multiplyMatrices } from './math.js';
 
 // These color space conversion function definitions are copied directly from
 // CSS Color Module Level 4 Sample Code: https://drafts.csswg.org/css-color/#color-conversion-code
+// *EXCEPT* the conversion matrices are replaced with exact rational forms computed using
+// this Rust crate: https://crates.io/crates/rgb_derivation
+// as described for sRGB on this page: https://mina86.com/2019/srgb-xyz-matrix/
+// but using the numbers from the CSS spec: https://github.com/kainino0x/exact_css_xyz_matrices
```
kainino0x, line 10:
```suggestion
// *EXCEPT* the conversion matrices are replaced with exact rational forms computed here:
// https://github.com/kainino0x/exact_css_xyz_matrices
//   using this Rust crate: https://crates.io/crates/rgb_derivation
//   as described for sRGB on this page: https://mina86.com/2019/srgb-xyz-matrix/
//   but using the numbers from the CSS spec: https://www.w3.org/TR/css-color-4/#predefined
```

kainino0x:
Your idea about D50 makes sense! Do you think it's just due to precision loss in the D50/D65 conversions or an actual error in the computation?

shaoboyan:
> Do you think it's just due to precision loss in the D50/D65 conversions
I think it is due to precision loss considering the big diff happens around ~0 values. And the method to convert from D65/D50 are various. We cannot control the implementation in browser and I think chromium chooses the recommended one.

PR #1090, opened 2022-03-18
Add checkbox to auto-close subtrees when they pass

kainino0x:
Small change, when a node gets turned green, this collapses it. Makes it easier to run some tests and filter down to what's failing. (Not perfect - have to have the box checked while running the tests, it doesn't apply afterward. But easier to have a tiny change.)


toji (approved):
LGTM! This will definitely be nice when running large blocks of tests.

Issue #912, opened 2022-01-13
val: device_mismatch throughout CTS


kainino0x:

kainino0x:
Reopening this issue which was accidentally auto-closed by a commit (hence why it still has status "Open"). @Gyuyoung added some more of these tests recently, so they may be done now, but we still need to review and make sure.

Gyuyoung:
Let me check if there are more tests that the `device_mismatch` validation tests need

Gyuyoung:
It's likely we've handled most of `device_mismatch` validation. It's difficult to find missing cases.

PR #1094, opened 2022-03-21
Implement api,validation,state,device_mismatched: Part II

haoxli:
- Remove helpers for getting device mismatched buffer/texture/queryset.
- Implement remaining device mismatch tests under
  validation/encoding/cmds and validation/encoding/queue.

Issue: #912

<hr>

### DIFF
--- a/src/webgpu/api/validation/encoding/cmds/clearBuffer.spec.ts
+++ b/src/webgpu/api/validation/encoding/cmds/clearBuffer.spec.ts
@@ -50,19 +50,19 @@ g.test('buffer,device_mismatch')
       await t.selectMismatchedDeviceOrSkipTestCase(undefined);
     }
 
-    const bufferSize = 8;
-    const descriptor: GPUBufferDescriptor = {
-      size: bufferSize,
+    const device = mismatched ? t.mismatchedDevice : t.device;
+    const size = 8;
+
+    const buffer = device.createBuffer({
+      size,
       usage: GPUBufferUsage.COPY_DST,
-    };
-    const buffer = mismatched
-      ? t.getDeviceMismatchedBuffer(descriptor)
-      : t.createBufferWithState('valid', descriptor);
+    });
+    t.trackForCleanup(buffer);
 
     t.TestClearBuffer({
       buffer,
       offset: 0,
-      size: bufferSize,
+      size,
       isSuccess: !mismatched,
     });
   });
--- a/src/webgpu/api/validation/encoding/cmds/clearBuffer.spec.ts
+++ b/src/webgpu/api/validation/encoding/cmds/clearBuffer.spec.ts
@@ -40,6 +40,29 @@ g.test('invalid_buffer')
     });
   });
 
+g.test('buffer,device_mismatch')
+  .desc(`Tests clearBuffer cannot be called with buffer created from another device.`)
+  .paramsSubcasesOnly(u => u.combine('mismatched', [true, false]))
+  .fn(async t => {
+    const { mismatched } = t.params;
+
+    const bufferSize = 8;
+    const descriptor: GPUBufferDescriptor = {
+      size: bufferSize,
+      usage: GPUBufferUsage.COPY_DST,
+    };
+    const buffer = mismatched
+      ? t.getDeviceMismatchedBuffer(descriptor)
+      : t.createBufferWithState('valid', descriptor);
+
+    t.TestClearBuffer({
+      buffer,
+      offset: 0,
+      size: bufferSize,
+      isSuccess: !mismatched,
+    });
+  });
+
 g.test('default_args')
   .desc(`Test that calling clearBuffer with a default offset and size is valid.`)
   .paramsSubcasesOnly([
--- a/src/webgpu/api/validation/encoding/cmds/clearBuffer.spec.ts
+++ b/src/webgpu/api/validation/encoding/cmds/clearBuffer.spec.ts
@@ -46,6 +46,10 @@ g.test('buffer,device_mismatch')
   .fn(async t => {
     const { mismatched } = t.params;
 
+    if (mismatched) {
+      await t.selectMismatchedDeviceOrSkipTestCase(undefined);
+    }
+
     const bufferSize = 8;
     const descriptor: GPUBufferDescriptor = {
       size: bufferSize,
--- a/src/webgpu/api/validation/encoding/cmds/compute_pass.spec.ts
+++ b/src/webgpu/api/validation/encoding/cmds/compute_pass.spec.ts
@@ -18,17 +18,6 @@ class F extends ValidationTest {
     return this.createErrorComputePipeline();
   }
 
-  createComputePipelineForMismatch(device: GPUDevice) {
-    return device.createComputePipeline({
-      compute: {
-        module: device.createShaderModule({
-          code: '@stage(compute) @workgroup_size(1) fn main() {}',
-        }),
-        entryPoint: 'main',
-      },
-    });
-  }
-
   createIndirectBuffer(state: ResourceState, data: Uint32Array): GPUBuffer {
     const descriptor: GPUBufferDescriptor = {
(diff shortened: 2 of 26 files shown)


src/webgpu/util/device_pool.ts
```diff
@@ -234,42 +237,55 @@ function supportsFeature(
 type DeviceHolderState = 'free' | 'reserved' | 'acquired';
 
 /**
- * Holds a GPUDevice and tracks its state (free/reserved/acquired) and handles device loss.
+ * Holds GPUDevice with same descriptor and tracks device state (free/reserved/acquired) and handles device loss.
+ * All the devices must be free/reserved/acquired together.
```
kainino0x, line 241: The idea of getting multiple devices from the device pool was to increase the pool's size so it could have more than 1 default device holder, rather than having multiple devices in each device holder. The device holder is designed to hold just one device (e.g. it has only one lostInfo). Extending it to hold 2 works for these tests, but seems overspecialized for this exact use case, I don't think it can be used for much else.

This new code _is_ overall nicer even though it's weird in a different way, but it also makes *every test* use 2 devices. That's probably not super costly, but it is surprising and I'd prefer not to.

Possible to split new test coverage and these framework changes? I'd like to land the new tests but I'm wary of the 2-device thing which could have effects on all the tests.

src/webgpu/api/validation/encoding/cmds/clearBuffer.spec.ts
```diff
@@ -40,6 +40,29 @@ g.test('invalid_buffer')
...
+    const { mismatched } = t.params;
+
+    const bufferSize = 8;
+    const descriptor: GPUBufferDescriptor = {
+      size: bufferSize,
+      usage: GPUBufferUsage.COPY_DST,
+    };
+    const buffer = mismatched
+      ? t.getDeviceMismatchedBuffer(descriptor)
+      : t.createBufferWithState('valid', descriptor);
```
kainino0x, line 56: nit: don't need to use createBufferWithState if the state is constant `'valid'`. Just directly use `t.device.createBuffer` which doesn't hide the API calls we're testing.

Same for the many other uses of createBufferWithState, createTextureWithState, and createQuerySetWithState in this PR.

src/webgpu/api/validation/encoding/cmds/compute_pass.spec.ts
```diff
@@ -65,7 +76,17 @@ setPipeline should generate an error iff using an 'invalid' pipeline.
 g.test('pipeline,device_mismatch')
   .desc('Tests setPipeline cannot be called with a compute pipeline created from another device')
   .paramsSubcasesOnly(u => u.combine('mismatched', [true, false]))
-  .unimplemented();
+  .fn(async t => {
+    const { mismatched } = t.params;
+
+    const pipeline = mismatched
+      ? t.createComputePipelineForMismatch(t.mismatchedDevice)
+      : t.createComputePipelineForMismatch(t.device);
```
kainino0x, line 84: nit: Since createComputePipelineForMismatch is only used in this one place, I don't think it needs to be in a shared location. I would either define it as a function inside this function, or just do it inline like

```
const pipelineDevice = mismatched ? t.mismatchedDevice : t.device;
const pipeline = pipelineDevice.createComputePipeline( // ...
```

src/webgpu/api/validation/encoding/cmds/render/setPipeline.spec.ts
```diff
@@ -31,4 +52,14 @@ Tests setPipeline should generate an error iff using an 'invalid' pipeline.
 g.test('pipeline,device_mismatch')
   .desc('Tests setPipeline cannot be called with a render pipeline created from another device')
   .paramsSubcasesOnly(kRenderEncodeTypeParams.combine('mismatched', [true, false]))
-  .unimplemented();
+  .fn(async t => {
+    const { encoderType, mismatched } = t.params;
+
+    const pipeline = mismatched
+      ? t.createRenderPipelineForMismatch(t.mismatchedDevice)
+      : t.createRenderPipelineForMismatch(t.device);
```
kainino0x, line 60: nit: same, createRenderPipelineForMismatch can be inline if it's only used in one place.

src/webgpu/api/validation/encoding/cmds/setBindGroup.spec.ts
```diff
@@ -140,7 +171,21 @@ g.test('bind_group,device_mismatch')
       .combine('useU32Array', [true, false])
       .combine('mismatched', [true, false])
   )
-  .unimplemented();
+  .fn(async t => {
+    const { encoderType, useU32Array, mismatched } = t.params;
+
+    const bindGroup = mismatched
+      ? t.createBindGroupForMismatch(t.mismatchedDevice, encoderType, useU32Array)
+      : t.createBindGroupForMismatch(t.device, encoderType, useU32Array);
```
kainino0x, line 179: same

src/webgpu/api/validation/validation_test.ts
```diff
@@ -214,18 +214,19 @@ export class ValidationTest extends GPUTest {
     }
   }
 
+  /** Return a GPUBuffer with descriptor from mismatched device. */
+  getDeviceMismatchedBuffer(descriptor: GPUBufferDescriptor): GPUBuffer {
+    return this.trackForCleanup(this.mismatchedDevice.createBuffer(descriptor));
+  }
```
kainino0x, line 220: Actually, even these helpers are not helping much. Doing just `this.mismatchedDevice.createBuffer` is not worse than `this.getDeviceMismatchedBuffer` (except you also want the trackForCleanup call).

src/webgpu/api/validation/queue/submit.spec.ts
```diff
@@ -24,4 +24,19 @@ g.test('command_buffer,device_mismatch')
...
+
+    const encoder0 = cb0Mismatched
+      ? t.mismatchedDevice.createCommandEncoder()
+      : t.device.createCommandEncoder();
+    const encoder1 = cb1Mismatched
+      ? t.mismatchedDevice.createCommandEncoder()
+      : t.device.createCommandEncoder();
+
+    t.expectValidationError(() => {
+      t.device.queue.submit([encoder0.finish(), encoder1.finish()]);
```
kainino0x, line 40: expectValidationError captures three calls: submit, finish, and finish. Be careful to only capture one API call in expectValidationError.

kainino0x (commented):
Nice coverage! Can we split the new tests off from the framework changes though?

I'll be out the rest of this week, you'll need to find another reviewer when you're ready.

src/webgpu/util/device_pool.ts
haoxli: OK, I will split the new tests to a separate PR, we can land it first. About getting multiple devices, I have different thought, let's dicuss it in another PR when you get back.

src/webgpu/api/validation/validation_test.ts
haoxli: Yes, using `getDeviceMismatchedBuffer` and `createBufferWithState` is for the trackForCleanup call. In addition, we can easily modify the `mismatchedDevice` without updating case by case.

src/webgpu/api/validation/encoding/cmds/clearBuffer.spec.ts
haoxli: Here I want the buffer/texture/queryset could be tracked for cleanup by using `createBufferWithState`, `createTextureWithState`, and `createQuerySetWithState`.

src/webgpu/api/validation/validation_test.ts
haoxli: Use `device.createBuffer` directly, and add `trackForCleanup`.

haoxli:
@austinEng Could you help to review the changes? Thanks!

austinEng (approved):
LGTM!

PR #1096, opened 2022-03-21
npm update, upgrade lockfileVersion, remove git conflict marker from lockfile

kainino0x:
Updated my local copy of node (`nvm install node`) and then just ran `npm update`, which did all these things (including fixing the conflict markers).

Also pins `prettier` because the reformatting is annoying and not particularly useful.


greggman (approved):
lgtm

greggman:
I don't know if this would be useful but I have some code like this in some of my Gruntfile.js

```
const requiredNodeVersion = 14;
if (parseInt((/^v(\d+)\./).exec(process.version)[1]) < requiredNodeVersion) {
  throw Error(`requires at least node: ${requiredNodeVersion}`);
}
```

Just to help people forking get a useful error message. Also, I know I could use semver to parse the version but I'm trying to avoid dependencies when the solution is simple (and crossing my fingers I don't screw up something :P)

PR #1106, opened 2022-03-24
Implements remaining device.destroy validation tests.

lokokung:
- Moves some canvas utilities around to be in central location for reuse, and update related tests.
- Moves some array typing around for reuse.
- Adds remaining tests for device destroy.

Issue: #882 

<hr>

### DIFF
--- a/src/common/util/util.ts
+++ b/src/common/util/util.ts
@@ -165,7 +165,7 @@ export function* iterRange<T>(n: number, fn: (i: number) => T): Iterable<T> {
   }
 }
 
-const TypedArrayBufferViewTypes = [
+const TypedArrayBufferViewInstances = [
   new Uint8Array(),
   new Uint8ClampedArray(),
   new Uint16Array(),
@@ -177,7 +177,7 @@ const TypedArrayBufferViewTypes = [
   new Float64Array(),
 ] as const;
 
-export type TypedArrayBufferView = typeof TypedArrayBufferViewTypes[number];
+export type TypedArrayBufferView = typeof TypedArrayBufferViewInstances[number];
 
 export type TypedArrayBufferViewConstructor<
   A extends TypedArrayBufferView = TypedArrayBufferView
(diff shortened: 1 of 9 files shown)


src/common/util/util.ts
```diff
@@ -164,16 +165,19 @@ export function* iterRange<T>(n: number, fn: (i: number) => T): Iterable<T> {
...
+  new Uint8Array(),
+  new Uint8ClampedArray(),
+  new Uint16Array(),
+  new Uint32Array(),
+  new Int8Array(),
+  new Int16Array(),
+  new Int32Array(),
+  new Float32Array(),
+  new Float64Array(),
+] as const;
```
austinEng, line 178: do we need to actually construct them? I think you can juts list them without new and (), and below, say result[v.name] = v;

src/webgpu/api/validation/state/device_lost/destroy.spec.ts
```diff
@@ -443,126 +497,446 @@ Tests creating query sets on destroyed device.
...
 g.test('queue,writeBuffer')
   .desc(
     `
-Tests writeBuffer on queue on destroyed device fails.
+Tests writeBuffer on queue on destroyed device.
+  `
+  )
+  .params(u =>
+    u
+      .combine('arrayType', kTypedArrayBufferViewKeys)
```
austinEng, line 798: now I see why you did the typed array changes.
we can keep them, but TBH not sure if it's worth testing writeBuffer with different types of typed arrays. 
The other writeBuffer test is a validation test, so it makes sense that we need to test with all the data types. For here, the destroy behavior shouldn't depend on the data type. 
Testing several different sizes would still be helpful though in case that puts the memory allocators in different states (I guess different types implicitly results in different sizes)

austinEng (approved):

src/common/util/util.ts
lokokung: So I originally had that, but if I don't construct them, then

`export type TypedArrayBufferView = typeof TypedArrayBufferViewTypes[number];`

is a union of constructor function types instead of just the types. Since generating the table was easier to manipulate than modifying the `typeof`, I just changed the array to be instances so that the inferred types were correct. Not sure if there's a better way though.

austinEng: ok, idk a better way so I guess it's fine.

let's call it kTypedArrayBufferViewInstances since it's a value not a type

src/webgpu/api/validation/state/device_lost/destroy.spec.ts
austinEng: but I think we should still remove the arrayType param since it shouldn't be important for the destroy test.

austinEng (approved):

PR #1126, opened 2022-03-31
Test view format reinterpretation

austinEng:
Issue: #

<hr>

### DIFF
--- a/src/webgpu/api/operation/texture_view/format_reinterpretation.spec.ts
+++ b/src/webgpu/api/operation/texture_view/format_reinterpretation.spec.ts
@@ -5,7 +5,6 @@ Test texture views can reinterpret the format of the original texture.
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
 import {
   EncodableTextureFormat,
-  kTextureFormatInfo,
   kRenderableColorTextureFormats,
   kRegularTextureFormats,
   viewCompatible,
@@ -110,32 +109,12 @@ g.test('texture_binding')
     // Make an input texel view.
     const inputTexelView = makeInputTexelView(format);
 
-    // Write data out to bytes.
-    const texelSize = kTextureFormatInfo[format].bytesPerBlock;
-    const textureData = new Uint8ClampedArray(texelSize * kTextureSize * kTextureSize);
-    inputTexelView.writeTextureData(textureData, {
-      bytesPerRow: texelSize * kTextureSize,
-      rowsPerImage: kTextureSize,
-      subrectOrigin: [0, 0],
-      subrectSize: [kTextureSize, kTextureSize],
-    });
-
-    // Create and upload data to the texture.
-    const texture = t.device.createTexture({
-      format,
+    // Create the initial texture with the contents if the input texel view.
+    const texture = t.makeTextureWithContents(inputTexelView, {
       size: [kTextureSize, kTextureSize],
-      usage: GPUTextureUsage.COPY_DST | GPUTextureUsage.TEXTURE_BINDING,
+      usage: GPUTextureUsage.TEXTURE_BINDING,
       viewFormats: [viewFormat],
     });
-
-    t.device.queue.writeTexture(
-      { texture },
-      textureData,
-      {
-        bytesPerRow: texelSize * kTextureSize,
-      },
-      [kTextureSize, kTextureSize]
-    );
 
     // Reinterepret the texture as the view format.
     // Make a texel view of the format that also reinterprets the data.
@@ -161,11 +140,13 @@ g.test('texture_binding')
     });
 
     // Create an rgba8unorm output texture.
-    const outputTexture = t.device.createTexture({
-      format: 'rgba8unorm',
-      size: [kTextureSize, kTextureSize],
-      usage: GPUTextureUsage.STORAGE_BINDING | GPUTextureUsage.COPY_SRC,
-    });
+    const outputTexture = t.trackForCleanup(
+      t.device.createTexture({
+        format: 'rgba8unorm',
+        size: [kTextureSize, kTextureSize],
+        usage: GPUTextureUsage.STORAGE_BINDING | GPUTextureUsage.COPY_SRC,
+      })
+    );
 
     // Execute a compute pass to load data from the reinterpreted view and
     // write out to the rgba8unorm texture.
@@ -209,7 +190,10 @@ g.test('texture_binding')
 g.test('render_and_resolve_attachment')
   .desc(
     `Test that a color render attachment allocated as 'format' is correctly rendered to as 'viewFormat',
-and resolved to an attachment allocated as 'format' viewed as 'viewFormat'.`
+and resolved to an attachment allocated as 'format' viewed as 'viewFormat'.
+
+Other combinations aren't possible because the render and resolve targets must both match
+in view format and match in base format.`
   )
   .params(u =>
     u //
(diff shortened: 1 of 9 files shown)


src/webgpu/api/operation/texture_view/format_reinterpretation.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+        ({ renderFormat, renderViewFormat }) =>
+          // Allow formats to be the same since reinterpretation may instead occur for the resolve format.
+          renderFormat === renderViewFormat ||
+          supportsReinterpretation(renderFormat, renderViewFormat)
+      )
+      .combine('resolveFormat', kRenderableColorTextureFormats)
+      .combine('resolveViewFormat', kRenderableColorTextureFormats)
+      .filter(
+        ({ renderFormat, renderViewFormat, resolveFormat, resolveViewFormat }) =>
+          renderViewFormat === resolveViewFormat && // Required by validation to match.
```
kainino0x, line 60: In that case do we really need both renderViewFormat and resolveViewFormat? Maybe just have one viewFormat.

kainino0x (approved):

src/webgpu/api/operation/texture_view/format_reinterpretation.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+import { GPUTest } from '../../../gpu_test.js';
+
+export const g = makeTestGroup(GPUTest);
+
+function supportsReinterpretation(a: GPUTextureFormat, b: GPUTextureFormat) {
+  return a + '-srgb' === b || b + '-srgb' === a;
+}
+
+g.test('texture_binding')
+  .desc(`Test that a regular texture allocated as 'format' may be sampled as 'viewFormat'.`)
```
kainino0x, line 19: nit: since these are operation tests, "may be" should be something like "are correctly"

kainino0x: > For all valid 'format'/'viewFormat' pairs, test that any regular texture allocated as 'format' is correctly sampled as 'viewFormat'.

similar below

src/webgpu/api/operation/texture_view/format_reinterpretation.spec.ts
```diff
@@ -0,0 +1,68 @@
+export const description = `
+Test texture views can reinterpret the format of the original texture.
+`;
```
kainino0x, line 3: oh just realized this overlaps with test descriptions in the neighboring `read.spec.ts` and `write.spec.ts` e.g.:

```
g.test('format')
  .desc(
    `Views of every allowed format.
- x= every texture format
- x= every valid view format (including same)
```

kainino0x: not sure if it makes more sense for these tests to be separate (in which case update the test description there) or combined with those.

austinEng: I'm keeping them separate and removed it from those files since I would think those tests are more about reading/writing to various subresources in different ways and checking the correct subresource specified by the view are written/read.

Testing the interpretation of the contents themselves feels like it should be separate.

src/webgpu/api/operation/texture_view/format_reinterpretation.spec.ts
austinEng: No longer applicable, given current spec issue I opened about this.

austinEng:
I ended up writing the tests for these as well as part of testing our implementation. @kainino0x PTAL again, thanks!

src/webgpu/api/operation/texture_view/format_reinterpretation.spec.ts
```diff
@@ -0,0 +1,390 @@
...
+    });
+
+    t.device.queue.writeTexture(
+      { texture },
+      textureData,
+      {
+        bytesPerRow: texelSize * kTextureSize,
+      },
+      [kTextureSize, kTextureSize]
+    );
```
kainino0x, line 138: maybe factor out to a `makeTextureWithContents` like `makeBufferWithContents`. (Can just put it in gpu_test; I don't think we actually need the separate `makeBufferWithContents` function that we have now)

Generating data straight into a mappedAtCreation buffer and copying to the texture might also be a little more optimal, though it's an extra copy vs an extra (non-staging) buffer allocation so idk.

src/webgpu/api/operation/texture_view/format_reinterpretation.spec.ts
```diff
@@ -0,0 +1,390 @@
...
+
+    // Create an rgba8unorm output texture.
+    const outputTexture = t.device.createTexture({
+      format: 'rgba8unorm',
+      size: [kTextureSize, kTextureSize],
+      usage: GPUTextureUsage.STORAGE_BINDING | GPUTextureUsage.COPY_SRC,
+    });
+
+    // Execute a compute pass to load data from the reinterpreted view and
+    // write out to the rgba8unorm texture.
```
kainino0x, line 171: Use rgba32float as the output format? Has better precision and range - not that it matters in practice right now.

src/webgpu/api/operation/texture_view/format_reinterpretation.spec.ts
```diff
@@ -0,0 +1,390 @@
...
+    pass.end();
+    t.device.queue.submit([commandEncoder.finish()]);
+
+    const result = await textureContentIsOKByT2B(
+      t,
+      { texture: outputTexture },
+      [kTextureSize, kTextureSize],
+      {
+        expTexelView: TexelView.fromTexelsAsColors('rgba8unorm', reinterpretedTexelView.color, {
+          clampToFormatRange: true,
```
kainino0x, line 201: wouldn't be needed

src/webgpu/api/operation/texture_view/format_reinterpretation.spec.ts
```diff
@@ -0,0 +1,390 @@
...
+        }),
+      },
+      { maxDiffULPsForNormFormat: 1 }
```
kainino0x, line 204: would change to float format

src/webgpu/api/operation/texture_view/format_reinterpretation.spec.ts
```diff
@@ -0,0 +1,390 @@
...
+    );
+    t.expectOK(result);
+  });
+
+g.test('render_and_resolve_attachment')
+  .desc(
+    `Test that a color render attachment allocated as 'format' is correctly rendered to as 'viewFormat',
+and resolved to an attachment allocated as 'format' viewed as 'viewFormat'.`
```
kainino0x, line 212:
```suggestion
and resolved to an attachment allocated as 'format' viewed as 'viewFormat'.

Other combinations aren't possible because the render and resolve targets must both match
in view format and match in base format.`
```

src/webgpu/api/operation/texture_view/format_reinterpretation.spec.ts
```diff
@@ -0,0 +1,390 @@
...
+
+    // Upload into the sample source.
+    t.device.queue.writeTexture(
+      { texture: sampleSource },
+      inputTextureData,
+      {
+        bytesPerRow: inputTexelSize * kTextureSize,
+      },
+      [kTextureSize, kTextureSize]
+    );
```
kainino0x, line 276: makeTextureWithContents

src/webgpu/api/operation/texture_view/format_reinterpretation.spec.ts
```diff
@@ -0,0 +1,390 @@
...
+    const renderViewTexels = TexelView.fromTexelsAsColors(viewFormat, inputTexelView.color, {
+      clampToFormatRange: true,
+    });
+    t.expectOK(
+      await textureContentIsOKByT2B(
+        t,
+        { texture: singleSampleRenderTexture },
+        [kTextureSize, kTextureSize],
+        { expTexelView: renderViewTexels },
+        { maxDiffULPsForNormFormat: 2 }
```
kainino0x, line 371: curious, do you happen to know why 2 is needed here?

kainino0x (approved):

src/webgpu/api/operation/texture_view/format_reinterpretation.spec.ts
austinEng: because the resolve operation blits into the resolve texture. On Vulkan, the exact values here can vary.

src/webgpu/api/operation/texture_view/format_reinterpretation.spec.ts
austinEng: It has better precision but then we need to convert our ULP difference of 1 or 2 in unorm format to float. I think that makes the test less clear.

austinEng:
@kainino0x PTAL again, thanks!

src/webgpu/api/operation/texture_view/format_reinterpretation.spec.ts
kainino0x: That's something to think about - you have one format but need ULPs in terms of another format.

kainino0x (approved):

PR #1134, opened 2022-04-04
val/copyT2T,format_compat: Don't assume compressed is always 4x4

Kangz:
The ASTC texture compression format has a lot of different block sizes
so using 16x16 for the test might not work. Instead use the LCM of the
block size.

(and add GCD and LCD util functions)

Issue: #

<hr>

### DIFF
--- a/src/webgpu/api/validation/encoding/cmds/copyTextureToTexture.spec.ts
+++ b/src/webgpu/api/validation/encoding/cmds/copyTextureToTexture.spec.ts
@@ -13,7 +13,7 @@ import {
   kTextureDimensions,
 } from '../../../../capability_info.js';
 import { kResourceStates } from '../../../../gpu_test.js';
-import { align } from '../../../../util/math.js';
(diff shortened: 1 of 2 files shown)


lokokung:
When I wrote some similar tests in the Dawn e2e suite, I think I just used the `block_size * modifiers` for the sizes instead of doing LCM. Might be easier to do that here also, then the params could just be the modifiers? WDYT?

Kangz:
Here the problem is that we want to test the compatibility of every pair of formats, so we could want to create an ETC2 texture (4x4) and try to copy it to an ASTC 6x5 texture. So we need the test texture sizes to be a multiple of both (so a multiple of 12x20). We could just multiply them but LCM is a tighter bound, I can change to the multiplication of block sizes if you'd like. Modifiers wouldn't work because we don't know a priori what modifiers to use for each test (asides from just the block size).

lokokung (approved):

lokokung:
Ahh, I see, I misunderstood the test sorry! LGTM then!

PR #1140, opened 2022-04-05
val,buffer,create: Buffer sizes don't have to be multiple of 4.

Kangz:
Issue: #

<hr>

### DIFF
--- a/src/webgpu/api/validation/buffer/create.spec.ts
+++ b/src/webgpu/api/validation/buffer/create.spec.ts
@@ -94,10 +94,6 @@ hidden behind the "more severe" validation error.`
       { usage: BufferUsage.MAP_READ | BufferUsage.UNIFORM, size: 16 },
       { usage: BufferUsage.MAP_READ | BufferUsage.UNIFORM, size: kMaxSafeMultipleOf8 },
       { usage: BufferUsage.MAP_READ | BufferUsage.UNIFORM, size: 0x20_0000_0000 }, // 128 GiB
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/buffer/create.spec.ts
```diff
@@ -94,10 +94,6 @@ hidden behind the "more severe" validation error.`
       { usage: BufferUsage.MAP_READ | BufferUsage.UNIFORM, size: 16 },
       { usage: BufferUsage.MAP_READ | BufferUsage.UNIFORM, size: kMaxSafeMultipleOf8 },
       { usage: BufferUsage.MAP_READ | BufferUsage.UNIFORM, size: 0x20_0000_0000 }, // 128 GiB
-      // Invalid because size is not aligned to 4 bytes.
-      { usage: BufferUsage.STORAGE, size: 15 },
-      { usage: BufferUsage.STORAGE, size: kMaxSafeMultipleOf8 - 1 },
-      { usage: BufferUsage.STORAGE, size: 0x20_0000_0000 - 1 }, // 128 GiB - 1
```
kainino0x, line 100: If these are valid, and not tested elsewhere, we should keep the cases and mark them as `_valid: true`

Kangz: This is already tested at the top of this file. The two latter cases are not valid because they produce OOMs, so I think we don't need to add these cases back?

PR #741, opened 2021-09-09
Implement api,validation,state,device_mismatched: Part Ⅰ

haoxli:
Complete *.spec.ts in src/webgpu/api/validation and src/webgpu/api/validation/encoding

<hr>

**Author checklist for test code/plans:**

- [x] All outstanding work is tracked with "TODO" in a test/file description or `.unimplemented()` on a test.
- [x] New helpers, if any, are documented using `/** doc comments */` and can be found via `helper_index.txt`.
- [x] (Optional, sometimes not possible.) Tests pass (or partially pass without unexpected issues) in an implementation. (Add any extra details above.)

**[Reviewer sign-off](https://github.com/gpuweb/cts/blob/main/docs/reviews.md) for test code/plans:** (Note: feel free to pull in other reviewers at any time for any reason.)

- [x] The test path is reasonable, the [description](https://github.com/gpuweb/cts/blob/main/docs/intro/plans.md) "describes the test, succinctly, but in enough detail that a reader can read only the test plans in a file or directory and evaluate the completeness of the test coverage."
- [x] Tests appear to cover this area completely, except for outstanding TODOs. Validation tests use control cases.
    (This is critical for coverage. Assume anything without a TODO will be forgotten about forever.)
- [ ] Existing (or new) test helpers are used where they would reduce complexity.
- [ ] TypeScript code is readable and understandable (is unobtrusive, has reasonable type-safety/verbosity/dynamicity).

PR #1142, opened 2022-04-05
Update DevicePool to provide multiple devices

haoxli:
- Remove mismatched device pool, make the device pool could provide
  multiple devices at once, and reserve a default mismatched device
  at test initialization.
- Cancel selecting the default mismatched device in tests. For
  requesting mismatched device with particular features, we still need
  to call selectMismatchedDeviceOrSkipTestCase.

Issue: #912

<hr>

### DIFF
--- a/src/webgpu/util/device_pool.ts
+++ b/src/webgpu/util/device_pool.ts
@@ -17,31 +17,47 @@ export class DevicePool {
   private defaultHolder: DeviceHolder | 'uninitialized' | 'failed' = 'uninitialized';
   /** Devices with descriptors. */
   private nonDefaultHolders = new DescriptorToHolderMap();
+  /** Device with no descriptor for device mismatch validation. */
+  private mismatchedDefaultHolder: DeviceHolder | 'uninitialized' | 'failed' = 'uninitialized';
+  /** Devices with descriptors for device mismatch validation. */
+  private mismatchedNonDefaultHolders = new DescriptorToHolderMap();
 
   /** Request a device from the pool. */
-  async reserve(descriptor?: UncanonicalizedDeviceDescriptor): Promise<DeviceProvider> {
+  async reserve(
+    descriptor?: UncanonicalizedDeviceDescriptor,
+    mismatched: boolean = false
+  ): Promise<DeviceProvider> {
     // Always attempt to initialize default device, to see if it succeeds.
     let errorMessage = '';
-    if (this.defaultHolder === 'uninitialized') {
+    let defaultHolder = mismatched ? this.mismatchedDefaultHolder : this.defaultHolder;
+    if (defaultHolder === 'uninitialized') {
       try {
-        this.defaultHolder = await DeviceHolder.create(undefined);
+        defaultHolder = await DeviceHolder.create(undefined);
       } catch (ex) {
-        this.defaultHolder = 'failed';
+        defaultHolder = 'failed';
         if (ex instanceof Error) {
           errorMessage = ` with ${ex.name} "${ex.message}"`;
         }
       }
     }
+
+    if (mismatched) {
+      this.mismatchedDefaultHolder = defaultHolder;
+    } else {
+      this.defaultHolder = defaultHolder;
+    }
     assert(
-      this.defaultHolder !== 'failed',
+      defaultHolder !== 'failed',
       `WebGPU device failed to initialize${errorMessage}; not retrying`
     );
 
     let holder;
     if (descriptor === undefined) {
-      holder = this.defaultHolder;
+      holder = defaultHolder;
     } else {
-      holder = await this.nonDefaultHolders.getOrCreate(descriptor);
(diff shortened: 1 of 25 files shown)


Jiawei-Shao:
Hi @kainino0x , we'd like to invite you to review this PR as it changes the test framework. Could you take a look?

src/webgpu/util/device_pool.ts
```diff
@@ -17,31 +17,47 @@ export class DevicePool {
   private defaultHolder: DeviceHolder | 'uninitialized' | 'failed' = 'uninitialized';
   /** Devices with descriptors. */
   private nonDefaultHolders = new DescriptorToHolderMap();
+  /** Device with no descriptor for device mismatch validation. */
+  private mismatchedDefaultHolder: DeviceHolder | 'uninitialized' | 'failed' = 'uninitialized';
+  /** Devices with descriptors for device mismatch validation. */
+  private mismatchedNonDefaultHolders = new DescriptorToHolderMap();
```
kainino0x, line 23: This looks essentially just like having two DevicePool objects (like before), but they've been merged into one object. What's the benefit to that?

src/webgpu/util/device_pool.ts
```diff
@@ -17,31 +17,47 @@ export class DevicePool {
...
   /** Request a device from the pool. */
-  async reserve(descriptor?: UncanonicalizedDeviceDescriptor): Promise<DeviceProvider> {
+  async reserve(
+    descriptor?: UncanonicalizedDeviceDescriptor,
+    mismatched: boolean = false
+  ): Promise<DeviceProvider> {
     // Always attempt to initialize default device, to see if it succeeds.
     let errorMessage = '';
-    if (this.defaultHolder === 'uninitialized') {
+    let defaultHolder = mismatched ? this.mismatchedDefaultHolder : this.defaultHolder;
```
kainino0x, line 32: nit: this isn't "default holder", should just be called `holder`

kainino0x (commented):
It's late so I didn't have time for a full review of the framework code, but one key question.

src/webgpu/util/device_pool.ts
haoxli: The reason for adding new properties, rather than expanding the existing holders into an array, is to make it easier to know where to delete the device when releasing a device, otherwise we need to go through all devices in the holders to check if it's default device or mismatched device. 

And if change `defaultHolder` and holders in `nonDefaultHolders` to array,  we use the first solt of DeviceHolders[] as the default device and the second solt as the mismatched device (so I wanted to change the `mismatchedDevice` to `secondaryDevice` before), but in terms of code reading, I think it is not very friendly to know their usages.

kainino0x: I understand why you do it as separate properties rather than an array, it's definitely a simpler change. But why do we need this change at all? It seems functionally identical to just having two DevicePools, which are more cleanly separated from each other.

I think you also made every GPUTest have two devices (primary and mismatched). That seems like it can be done just fine with two DevicePools. That said, I still don't think it's good to have every test in the entire test suite use two devices. Only the device-mismatch tests need them, and I don't think one line of setup in those tests is a problem. (Right now it's three, because of the `if (mismatched)` check, but that's not really necessary.)

haoxli: Oh! Do you meams why remove the these lines: 
```
if (mismatched) {
  await t.selectMismatchedDeviceOrSkipTestCase(undefined);
}
```

I remember that you had a comment at https://github.com/gpuweb/cts/pull/741, said 
> Come to think of it, it's also unfortunate that selectMismatchedDeviceOrSkipTestCase needs to be called in every test separately. It would be ideal if it happened implicitly, and if it wasn't necessary to pass in a device descriptor at all - the mismatched device should always have the same device descriptor as the original device

If we want to have a mismatched device implicitly, we need to initialize the device in every test, otherwise `selectMismatchedDeviceOrSkipTestCase` is needed in mismatched validation tests.

kainino0x: > I remember that you had a comment at https://github.com/gpuweb/cts/pull/741, said
> 
> > Come to think of it, it's also unfortunate that selectMismatchedDeviceOrSkipTestCase needs to be called in every test separately. It would be ideal if it happened implicitly, and if it wasn't necessary to pass in a device descriptor at all - the mismatched device should always have the same device descriptor as the original device

Ah yes, I remember that. I think the implementation I had in mind was, if a test uses the .mismatchedDevice property (or .getMismatchedDevice()) then we would get one automatically, rather than having to request one. I don't think we should do it for all tests regardless of whether they need one.

> I'm not sure if I misunderstood the DevicePool support multiple devices here.

The intent of this MAINTENANCE_TODO was to generalize the DevicePool so we can run multiple tests in parallel (since tests are async). But it hasn't been important so we haven't done anything about it.

Issue #901, opened 2022-01-13
Update tests for sparse render attachments if added


kainino0x:
https://github.com/gpuweb/gpuweb/issues/1250

webgpu:api,validation,attachment_compatibility:render_pass_and_bundle,color_count:*
webgpu:api,validation,attachment_compatibility:render_pass_or_bundle_and_pipeline,color_count:*

PR #1149, opened 2022-04-07
Sparse color attachments validation tests

shrekshao:
Issue: #901

WIP: need to land chromium change and npm publish updated types first

<hr>

### DIFF
--- a/src/webgpu/api/validation/attachment_compatibility.spec.ts
+++ b/src/webgpu/api/validation/attachment_compatibility.spec.ts
@@ -1,7 +1,5 @@
 export const description = `
 Validation for attachment compatibility between render passes, bundles, and pipelines
-
-TODO: Add sparse color attachment compatibility test when defined by specification
 `;
 
 import { makeTestGroup } from '../../../common/framework/test_group.js';
@@ -92,11 +90,11 @@ class F extends ValidationTest {
   }
 
   createColorAttachment(
-    format: GPUTextureFormat | undefined,
+    format: GPUTextureFormat | null,
     sampleCount?: number
-  ): GPURenderPassColorAttachment | undefined {
-    return format === undefined
-      ? undefined
+  ): GPURenderPassColorAttachment | null {
+    return format === null
+      ? null
       : {
           view: this.createAttachmentTextureView(format, sampleCount),
           clearValue: [0, 0, 0, 0],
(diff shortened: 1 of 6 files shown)


shrekshao:
A question for `flat` on `number[][][]`. Searching on internet shows it could be resolved by adding "es2019" to tsconfig.json `compilerOptions.lib`. Not sure if we have a specific reason to keep it "es2016" (I saw a commit changing it back from "es2017")

kainino0x:
I think it should be fine to change. This function, at least, has had support for a good while in all browsers and node/deno:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat#browser_compatibility

src/webgpu/api/validation/attachment_compatibility.spec.ts
```diff
@@ -157,6 +209,48 @@ g.test('render_pass_and_bundle,color_count')
...
+      .combine('attachmentCount', kColorAttachmentCounts)
+      .beginSubcases()
+      .combine('passAttachments', kColorAttachments)
+      .combine('bundleAttachments', kColorAttachments)
+      .filter(
+        p =>
+          p.attachmentCount === p.passAttachments.length &&
+          p.attachmentCount === p.bundleAttachments.length
+      )
+  )
```
austinEng, line 230: discussed offline: this is a *lot* of subcases. We could trim it down so emptySlots of 2 only goes up to maybe 4? color attachments. and emptySlots of 1 for the rest.

Or, perhaps generate a subset of the cases for emptySlots = 2

src/webgpu/api/validation/attachment_compatibility.spec.ts
```diff
@@ -18,6 +16,60 @@ import {
...
+          cases.push(r);
+        }
+
+        return cases;
+      }).flat()
+    );
+
+    return result;
+  })
+  .flat() as number[][];
```
austinEng, line 71: maybe make this boolean[][] ?

then later, `map` can just do `i ? 'rgba8unorm' : null`

or, since we always use rgba8unorm, we could just output that now, instead of translating later

src/webgpu/api/validation/attachment_compatibility.spec.ts
shrekshao: Fixed by doing this: "emptySlots of 2 only goes up to maybe 4"

src/webgpu/api/validation/attachment_compatibility.spec.ts
shrekshao: use boolean (still needs to translate to color formats, color targets, or color attachments)

austinEng (approved):
on Chrome Canary, the longest time is 300ms, so LGTM

PR #1155, opened 2022-04-10
Only 2D textures can be marked as render targets

litherum:
https://gpuweb.github.io/gpuweb/#dom-gpudevice-createtexture says:

> If descriptor.usage includes the RENDER_ATTACHMENT bit:
> ...
> descriptor.dimension must be "2d".

### DIFF
--- a/src/webgpu/api/validation/createTexture.spec.ts
+++ b/src/webgpu/api/validation/createTexture.spec.ts
@@ -294,7 +294,8 @@ g.test('sampleCount,valid_sampleCount_with_other_parameter_varies')
       .unless(({ sampleCount, usage, format, mipLevelCount, dimension }) => {
         const info = kTextureFormatInfo[format];
         return (
-          ((usage & GPUConst.TextureUsage.RENDER_ATTACHMENT) !== 0 && (!info.renderable || dimension !== '2d')) ||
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -294,7 +294,7 @@ g.test('sampleCount,valid_sampleCount_with_other_parameter_varies')
       .unless(({ sampleCount, usage, format, mipLevelCount, dimension }) => {
         const info = kTextureFormatInfo[format];
         return (
-          ((usage & GPUConst.TextureUsage.RENDER_ATTACHMENT) !== 0 && !info.renderable) ||
+          ((usage & GPUConst.TextureUsage.RENDER_ATTACHMENT) !== 0 && (!info.renderable || dimension !== '2d')) ||
```
Kangz, line 297: FYI you can fix the warning that was added by Github actions here with `npm run fix`.

Kangz (approved):

src/webgpu/api/validation/createTexture.spec.ts
Kangz: Also I think we should update the `texture_usage` test in the same file (around line 700) so that it checks that validation rule.

kainino0x (approved):

PR #1158, opened 2022-04-12
Make things work on Node 12.x, and test it on presubmit.

kainino0x:
Events aren't implemented on Node 12.x, so just stub out the necessary interfaces and skip the relevant tests.

Issue: none

<hr>

- [n/a] Tests are properly located in the test tree.
- [n/a] [Test descriptions](https://github.com/gpuweb/cts/blob/main/docs/intro/plans.md) allow a reader to "read only the test plans and evaluate coverage completeness", and accurately reflect the test code.
- [n/a] Tests provide complete coverage (including validation control cases). **Missing coverage MUST be covered by TODOs.**

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -16,7 +16,7 @@ jobs:
           git checkout ${{ github.event.pull_request.head.sha }}
       - uses: actions/setup-node@v2-beta
         with:
-          node-version: "15.x"
+          node-version: "12.x"
       - run: npm ci
       - run: npm test
       - run: |
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -13,7 +13,7 @@ jobs:
           persist-credentials: false
       - uses: actions/setup-node@v2-beta
         with:
-          node-version: "15.x"
+          node-version: "12.x"
       - run: npm ci
       - run: |
           npm test
--- a/src/common/internal/file_loader.ts
+++ b/src/common/internal/file_loader.ts
@@ -28,7 +28,20 @@ interface TestFileLoaderEventMap {
   finish: MessageEvent<void>;
 }
 
-export interface TestFileLoader extends EventTarget {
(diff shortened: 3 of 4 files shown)


kainino0x:
Node 12 because that's the version in [in Debian stable](https://packages.debian.org/search?keywords=nodejs&searchon=names&exact=1&suite=testing&section=all)

greggman (approved):
lgtm

A little worried that people are still using 12. 12's support ends this month

https://nodejs.org/en/about/releases/

Further, the default npm for 12 uses v1 package-lock.json where was 14/16 use v2 (I run into this because I have lots of 12 projects and if I happen to run `npm i` in them and I'm on 14 or 16 the default versions of npm for those version of node end up converting the project-lock.json to v2. Plan to switch all my personal stuff to v16

kainino0x:
Gregg pointed out that we use some syntax like `?.` which isn't accepted by Node 12. I'm now confused why I didn't catch this when running `npm test` against Node 12 locally, and why it wasn't caught by the checks on this PR.

kainino0x:
> Gregg pointed out that we use some syntax like `?.` which isn't accepted by Node 12. I'm now confused why I didn't catch this when running `npm test` against Node 12 locally, and why it wasn't caught by the checks on this PR.

Figured it out, the difference is ts-node is transpiling new syntaxes into old syntaxes (using tsc), but the `out/` build is using babel which is configured not to do that.

kainino0x:
It would be nice to get rid of that split. Looks like we could stop using ts-node and use babel-node:
https://www.jonathancreamer.com/running-programs-with-babel-node/

Then I think(?) we would never use tsc's code generation (only its type checking) which should make things more understandable.

greggman:
Want to point out, github actions is warning node 12 is no longer supported.

https://github.com/gpuweb/cts/actions/runs/3856103212

kainino0x:
With node 12 being 9 months older than it was when I opened this, I'm happy to close this.

PR #1169, opened 2022-04-14
Add an option to run subcases in parallel

austinEng:
Setting parallelSubcases=true in the standalone runner or passing
--parallel-subcases on the command line will cause subcases to not
wait for each other to complete before continuing to the next
subcases. When there is a lot of async behavior, this should yield
better test throughput.

Currently defaults to false.

Issue: #

<hr>

**Requirements for PR author:**

- [ ] All missing test coverage is tracked with "TODO" or `.unimplemented()`.
- [ ] New helpers are `/** documented */` and new helper files are found in `helper_index.txt`.
- [ ] Test behaves as expected in a WebGPU implementation. (If not passing, explain above.)

**Requirements for [reviewer sign-off](https://github.com/gpuweb/cts/blob/main/docs/reviews.md):**

- [ ] Tests are properly located in the test tree.
- [ ] [Test descriptions](https://github.com/gpuweb/cts/blob/main/docs/intro/plans.md) allow a reader to "read only the test plans and evaluate coverage completeness", and accurately reflect the test code.
- [ ] Tests provide complete coverage (including validation control cases). **Missing coverage MUST be covered by TODOs.**
- [ ] Helpers and types promote readability and maintainability.

When landing this PR, be sure to make any necessary issue status updates.

PR #1172, opened 2022-04-15
Pipeline gathering subcase results

austinEng:
Pipeline gathering subcase results
    
Makes some tests run 10x faster. On average, a 10-35% total
runtime reduction.

Running a test group is changed so that a FixtureSharedState is
shared between subcases. Creation of the state can be overriden
by the test fixture.

Subcase finalization is non-blocking until the end of the
test group run. This means that resolution of eventual async
expectations don't block subsequent subcases from starting.
All eventually expectations are finally waited on at the end of
the testcase.

This results in a bit worse device lost granularity as it
may not be possible to tell exactly where a device lost error
occured. Also, unexpected validation and OOM errors are caught
at the whole testcase level instead of the subcase level. This is
probably fine since subcases are intended to be like a "for-loop"
so exact reporting is probably not crucial.

Test finalization and releasing the device back to the
device pool occurs in the same step. This is necessary since we
shouldn't start reusing the device until eventual expectations
which operate on the same device are complete. This means a single
testcase shouldn't have complex interactions with the device pool.
To maintain this invariant, we require move methods for selecting
a device with optional formats/features to occur in the `before`
stage before any of the subcases run using the fixture shared state.
Then, all subcases run using the same device.

Issue: #

<hr>

### DIFF
--- a/src/common/internal/test_group.ts
+++ b/src/common/internal/test_group.ts
@@ -1,4 +1,10 @@
-import { Fixture, SkipTestCase, TestParams, UnexpectedPassError } from '../framework/fixture.js';
+import {
+  Fixture,
+  FixtureSharedState,
+  SkipTestCase,
+  TestParams,
+  UnexpectedPassError,
+} from '../framework/fixture.js';
 import {
   CaseParamsBuilder,
   builderIterateCasesWithSubcases,
@@ -7,7 +13,7 @@ import {
   SubcaseParamsBuilder,
 } from '../framework/params_builder.js';
 import { Expectation } from '../internal/logging/result.js';
-import { TestCaseRecorder } from '../internal/logging/test_case_recorder.js';
+import { CaseRecorder, TestCaseRecorder } from '../internal/logging/test_case_recorder.js';
 import { extractPublicParams, Merged, mergeParams } from '../internal/params_utils.js';
 import { compareQueries, Ordering } from '../internal/query/compare.js';
 import { TestQuerySingleCase, TestQueryWithExpectation } from '../internal/query/query.js';
@@ -40,11 +46,13 @@ export interface RunCase {
 }
 
 // Interface for defining tests
-export interface TestGroupBuilder<F extends Fixture> {
-  test(name: string): TestBuilderWithName<F>;
+export interface TestGroupBuilder<S extends FixtureSharedState, F extends Fixture<S>> {
+  test(name: string): TestBuilderWithName<S, F>;
 }
-export function makeTestGroup<F extends Fixture>(fixture: FixtureClass<F>): TestGroupBuilder<F> {
-  return new TestGroup(fixture);
+export function makeTestGroup<S extends FixtureSharedState, F extends Fixture<S>>(
+  fixture: FixtureClass<S, F>
+): TestGroupBuilder<S, F> {
+  return new TestGroup((fixture as unknown) as FixtureClass);
 }
 
 // Interfaces for running tests
@@ -60,22 +68,27 @@ export interface IterableTest {
 }
 
 export function makeTestGroupForUnitTesting<F extends Fixture>(
-  fixture: FixtureClass<F>
-): TestGroup<F> {
+  fixture: FixtureClass<FixtureSharedState, F>
+): TestGroup<FixtureSharedState, F> {
   return new TestGroup(fixture);
 }
 
-export type FixtureClass<F extends Fixture = Fixture> = new (
-  log: TestCaseRecorder,
-  params: TestParams
-) => F;
+export type FixtureClass<
+  S extends FixtureSharedState = FixtureSharedState,
+  F extends Fixture<S> = Fixture<S>
+> = (new (sharedState: S, log: CaseRecorder, params: TestParams) => F) & {
+  MakeSharedState: (params: TestParams) => S;
+};
 type TestFn<F extends Fixture, P extends {}> = (t: F & { params: P }) => Promise<void> | void;
-type BeforeFn<F extends Fixture, P extends {}> = (t: F & { params: P }) => Promise<void> | void;
+type BeforeFn<S extends FixtureSharedState, P extends {}> = (
+  s: S & { params: P }
+) => Promise<void> | void;
 
-export class TestGroup<F extends Fixture> implements TestGroupBuilder<F> {
+export class TestGroup<S extends FixtureSharedState, F extends Fixture<S>>
+  implements TestGroupBuilder<S, F> {
   private fixture: FixtureClass;
   private seen: Set<string> = new Set();
-  private tests: Array<TestBuilder<F>> = [];
+  private tests: Array<TestBuilder<S, F>> = [];
 
   constructor(fixture: FixtureClass) {
     this.fixture = fixture;
@@ -97,7 +110,7 @@ export class TestGroup<F extends Fixture> implements TestGroupBuilder<F> {
     this.seen.add(name);
   }
 
-  test(name: string): TestBuilderWithName<F> {
+  test(name: string): TestBuilderWithName<S, F> {
     const testCreationStack = new Error(`Test created: ${name}`);
 
     this.checkName(name);
@@ -109,7 +122,7 @@ export class TestGroup<F extends Fixture> implements TestGroupBuilder<F> {
 
     const test = new TestBuilder(parts, this.fixture, testCreationStack);
     this.tests.push(test);
-    return (test as unknown) as TestBuilderWithName<F>;
+    return (test as unknown) as TestBuilderWithName<S, F>;
   }
 
   validate(): void {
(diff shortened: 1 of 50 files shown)


austinEng:
Alternative version to #1169. Overall, not as fast in terms of test throughput, but executes subcases serially instead of interleaved. Interleaved execution might lead to confusing results and is blocked anyway on some device_pool work.

@Kangz could you take an early review? There's two commits, which might be helpful to look at separately.
If the overall direction makes sense, I'll need to refactor the remaining tests (there's a lot) to use the new setup behavior, and then I'll test it against Chrome in full to verify.

src/unittests/maths.spec.ts
```diff
@@ -649,7 +649,7 @@ g.test('test,math,correctlyRoundedFlushToZeroOnly')
   ]
   )
   .fn(t => {
-    const test_val = t.params.test_val;
+    const test_val = t.params.test_val as Scalar;
```
Kangz, line 652: Why is this needed?

src/webgpu/gpu_test.ts
```diff
@@ -62,6 +62,15 @@ export function initUncanonicalizedDeviceDescriptor(
   }
 }
 
+type DisallowedMethodsInTest = {
+  // Disallow tests to use these methods. They must be called in
```
Kangz, line 66: nit: This reads a bit weird.

src/common/internal/test_group.ts
```diff
@@ -446,40 +434,66 @@ class RunCaseSpecific implements RunCase {
       return didSeeFail ? 'fail' : 'pass';
```
Kangz, line 434: nit: getExptectedStatus seems like it could be a helper function on the side.

src/common/internal/test_group.ts
```diff
@@ -446,40 +434,66 @@ class RunCaseSpecific implements RunCase {
...
-            // Convert SkipTestCase to info messages
-            ex.message = 'subcase skipped: ' + ex.message;
-            rec.info(ex);
-            ++skipCount;
-          } else {
-            // Since we are catching all error inside runTest(), this should never happen
-            rec.threw(ex);
+    try {
+      rec.start();
+      const inst = new this.fixture(rec, this.params);
```
Kangz, line 439: Here we are reusing the same fixture many times. Don't we have a concern that tests could be storing state on the fixture that would break stuff because it is not expected?

src/common/internal/test_group.ts
```diff
@@ -446,40 +434,66 @@ class RunCaseSpecific implements RunCase {
...
+        await inst.doInit();
+        if (this.beforeFn) {
+          await this.beforeFn(inst as Fixture & { params: {} });
+        }
+
+        if (this.subcases) {
+          let totalCount = 0;
+          let skipCount = 0;
+          for (const subParams of this.subcases) {
+            rec.info(new Error('subcase: ' + stringifyPublicParams(subParams)));
```
Kangz, line 450: IMHO we could have `runAllSubTests` be a helper function as well. It would make things slightly easier to read.

src/common/internal/test_group.ts
```diff
@@ -446,40 +434,66 @@ class RunCaseSpecific implements RunCase {
...
+            getExpectedStatus(selfQuery)
+          );
         }
-        ++totalCount;
-      }
-      if (skipCount === totalCount) {
-        rec.skipped(new SkipTestCase('all subcases were skipped'));
+      } finally {
+        // Runs as long as the fixture constructor succeeded, even if initialization or the test failed.
+        await inst.doFinalize();
```
Kangz, line 487: I was thinking we would keep only the eventual expectations for a certain amount of cases in flight. But that may be difficult to do without extension changes to the fixture.

Kangz (commented):
General approach looks good, I'm just worried about how difficult to understand the code is, and how many eventual expectations we might be buffering.

src/common/internal/test_group.ts
austinEng: Yea, definitely. I've been working on a bigger refactor to let the fixture class perform some setup operations whose results then get passed into the fixture constructor.

It restructures things quite a bit so I'm still working out the details and making it not look like a mess.

austinEng:
@kainino0x high level thoughts on this PR? - it's going to be some non-trivial additional complexity since if subcase finalization is deferred, then recording those results will also be deferred and occur out-of-order with respect to other `rec.xyz` calls. And... other steps like destroy(), etc. need to also occur after expectations resolve, I've got this mostly working (I think) but it'll still need some polish.

austinEng:
I'm testing this on our infrastructure in https://dawn-review.googlesource.com/c/dawn/+/88081

Preliminary results just on Linux NVIDIA show that total shard time is about 10% improved - much less than I was hoping for.
The number of actual tests has ballooned from 27,482 to 56,489. Not clear at this time if the large number of tests is what's diminishing the benefit of pipelining, or if the pipelining isn't that advantageous for a large number of the tests. It could be both of these that are true. Locally, pipelining helped some tests by 8-10x, but smaller tests didn't benefit very much at all.

The more than doubling in test count is due to the fact that selecting a device for the test or skipping if it's not supported - must be done in the `before` clause. That transforms a bunch of tests to use cases instead of subcases. This is particularly bad for textureFormat x textureFormat parameterizations, adding at least 95*95=9025 testcases every time I need to expand this.

This huge increase in testcase count also makes the standalone UI rather unusable, so something else will need to be done about that. I'll try to get a baseline for how much the test times change with just the testcase count increase, without any pipelining of results in order to get a good idea of how much pipelining actually helps.

austinEng:
Some super back-of-the-envelope estimations:

Most of the additional testcases are coming from places where I transformed textureFormat from a subcase to a case param. This changed a test like `webgpu:api,validation,createView:format:textureFormat="bgra8unorm"` which used to run in 200ms into 95 tests - one for each viewFormat pair - which each take about 60ms. So that's about 95x60-200 = 5500ms of added overhead for each of the 95 formats, bringing us to 522,500 milliseconds of added overhead for all format x format pairs.. which is 8.7 minutes. If we consider that this happened three times (27k extra tests, and 8.7 minutes is just for 95*95=9025 of them), we have about 26 minutes of extra overhead.

Which means, if this overhead didn't exist, we'd have a shard time reduced from 69m down to 35m (instead of the original 61m). That would be an estimated 50% total reduction in test run time.

I need to verify these estimates a bit, but if it's true, it means that promoting some of the subcase params to case params is hurting us far too much. We could consider:
  1. Lifting the restriction that the device needs to be selected in the `before` clause. This makes the expectation resolution and test finalization rather complicated so I'd like not to do that.
  2. reducing the number of parameterizations. We really don't need all 9,025 parameterizations for these tests. I think it would be good enough to select a representative subset for the second format in a format x format combination.

Kangz:
Thanks for sharing all these numbers. A 50% reduction could be worth pursuing.

What's the difficulty with doing 1? Is it that it would require a device pool?

austinEng:
Testing idea 3 from https://github.com/gpuweb/cts/pull/1172#issuecomment-1111177491 in https://dawn-review.googlesource.com/c/dawn/+/88081/3

This brings the number of tests down to about 29k.  45m total shard time compared to 69m total shard time - so that's about 35% reduction in test time. This is not as good as the theoretical 50% but still a great improvement, and would be necessary anyway to make the standalone UI usable. Going to move forward with this approach for now and we can improve other parts as needed.

austinEng:
Update: it appears these speedups are platform dependent.

Linux NVIDIA 45m total shard time compared to 69m total shard time - so that's about 35% reduction in test time.
Windows Intel 187m total shard time compared to 216m - 15% reduction
Windows NVIDIA - 115m total shared time compared to 126m - 8% reduction

My reading of this is that we see less of a speedup on Windows here since the tests overall take much longer than on Linux. So most of the time is spent in the actual test and pipelining results isn't as helpful.

src/common/internal/test_group.ts
```diff
@@ -428,40 +476,83 @@ class RunCaseSpecific implements RunCase {
...
+                rec.threw(ex);
+              }
+            }
+            ++totalCount;
+
+            // Wait to ensure there aren't too many in-flight subcases.
+            const kMaxPendingEventualExpectations = 500;
+            while (sharedState.numPendingEventualExpectations > kMaxPendingEventualExpectations) {
+              await sharedState.makeEventualExpectationDecrementedPromisePromise();
+            }
```
kainino0x, line 527: Can we just do a maximum number of pending subcases instead? Even just batch them into groups of 500 (run 500 in parallel, wait for them to finalize, run another 500, and so on). Seems a bit simpler than counting EventualExpectations.

kainino0x (commented):
After some back and forth on chat, I think this looks like a good idea. The complexity from the SharedState stuff is a little unfortunate but it's not that bad for what it gets us. before() is pretty nice.

Biggest source of complexity seems to be the subcase logs. I haven't looked closely at them yet but wondering if we can simplify by just collecting up the logs (maybe by creating a temporary recorder for each subcase?), flattening, and reporting them all at the end.

austinEng:
This PR was split into smaller ones which are now merged.
CTS rolled successfully in https://dawn-review.googlesource.com/c/dawn/+/88720

Results look pretty good!

| Platform | Total Shard Time Before | Total Shard Time After | Decrease | % Decrease |
| ---------| --------------------- | ---------------------| ------- | -------- |
| Linux NVIDIA | 62 m | 45 m | 17 m | 27.4%
| Linux Intel |  106 m | 93 m | 13 m | 12.3% 
| Mac AMD |  159 m | 122 m | 37 m | 23.3%
| Mac Intel | 144 m | 109 m | 35 m | 24.3%
| Win NVIDIA | 118 m | 102 m | 16 m | 13.6%
| Win Intel | 220 m | 189 m | 31 m | 14.1%

Issue #1135, opened 2022-04-04
Issue with interpolate wgsl test


dj2:
Reading though https://github.com/gpuweb/cts/blob/main/src/webgpu/shader/validation/shader_io/interpolate.spec.ts#L44 it looks like the test will never validate the `@interpolate(flat)`, `@interpolate(perspective)` or `@interpolate(linear)` cases.

This is because the `type` is always applied with a `,` at the end, so the test generates `@interpolate(flat,)`, `@interpolate(perspective,)`, and `@interpolate(linear,)`. Those will then not match the `kValidInterpolationAttributes` set so the validation will fail (as it should). But, that means the non-failing versions never get tested.

(The test should probably also generate the `@interpolate(,center)` variants as well to verify they fail)

dj2:
@jrprice

dj2:
Fixed

PR #1189, opened 2022-04-22
Extend WGSL interpolate tests.

dj2:
The current interpolation tests will not attempt to validate an
interpolation with just the type (e.g. `@interpolate(flat)`). This is
due to the `,` always being appended so `@interpolate(flat, )` is
generated which is not a valid value.

This PR updates the generation to only add the `,` if there is a
sampling value to be appended. 

Issue: #1135

<hr>

### DIFF
--- a/src/webgpu/shader/validation/shader_io/interpolate.spec.ts
+++ b/src/webgpu/shader/validation/shader_io/interpolate.spec.ts
@@ -15,10 +15,16 @@ const kValidInterpolationAttributes = new Set([
   '@interpolate(perspective, center)',
   '@interpolate(perspective, centroid)',
   '@interpolate(perspective, sample)',
+  '@interpolate(perspective,center)',
+  '@interpolate(perspective,centroid)',
+  '@interpolate(perspective,sample)',
   '@interpolate(linear)',
   '@interpolate(linear, center)',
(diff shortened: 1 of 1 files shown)


jrprice (approved):
This change looks fine, but note that `@interpolate(flat,)` is actually valid. WGSL now [allows optional trailing commas in attribute lists](https://github.com/gpuweb/gpuweb/pull/2563), to be consistent with other comma-separated lists. We [haven't yet implemented that change in Tint](crbug.com/tint/1423), which is why we reject that code.

So maybe worth adding a TODO to test trailing commas as well once they're implemented, unless we want to just test that somewhere else.

dj2:
Oh, cool, didn't realize that was valid. I'll update this PR to test both the `@interpolate(flat,)` and `@interpolate(perspective,center,)` variants.

dj2:
Actually, thinking about this some more, I'm going to break the bit with adding/removing the space out of this PR and will skip the bit with the trailing `,`. Both of those will go into the parsing tests for `interpolate`. This test is specifically about `shader_io` and should focus on the combinations of values instead of the parsing of those values.

Will push an updated PR.

jrprice (approved):
LGTM!

PR #1190, opened 2022-04-23
Add a quiet option to the test runner.

dj2:
This CL adds an option `--quiet` to the test runner to suppress the
summary output. If the harness is being driven by an external program
then it's possible it's getting fed a single test at a time and the
summary adds a lot of noise to the output.


dj2:
`quiet` is probably the wrong name for this maybe `no-summary` ?

ben-clayton (approved):

toji (approved):
`--quiet` is fine IMO, but `--no-summary` would work too. Approved either way.

PR #1195, opened 2022-04-24
Add more interpolate tests.

dj2:
This PR validates that the interpolate sampling parameter can not come
before the type parameter. It also tests that the `interpolate` attribute can
only be applied once.

<hr>

### DIFF
--- a/src/webgpu/shader/validation/shader_io/interpolate.spec.ts
+++ b/src/webgpu/shader/validation/shader_io/interpolate.spec.ts
@@ -28,8 +28,24 @@ g.test('type_and_sampling')
       .combine('stage', ['vertex', 'fragment'] as const)
       .combine('io', ['in', 'out'] as const)
       .combine('use_struct', [true, false] as const)
-      .combine('type', ['', 'flat', 'perspective', 'linear'] as const)
-      .combine('sampling', ['', 'center', 'centroid', 'sample'] as const)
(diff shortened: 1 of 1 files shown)


src/webgpu/shader/validation/shader_io/interpolate.spec.ts
```diff
@@ -59,6 +59,47 @@ g.test('type_and_sampling')
...
+    }
+    const code = generateShader({
+      attribute: '@location(0)' + interpolate,
+      type: 'f32',
+      stage: t.params.stage,
+      io: t.params.io,
+      use_struct: t.params.use_struct,
+    });
+
+    t.expectCompileResult(false, code);
```
jrprice, line 100: I have a slight reservation about validation tests that are expected to always fail. These tests have no way of checking whether they're failing for the right reasons, which means that they have a good chance of becoming stale (e.g. WGSL syntax changes, the test now fails because the syntax is wrong, but we don't notice because fail means pass).

I've been striving to always have control cases for validation tests to avoid this issue, which I think is what the API tests tend to do as well. Is that something that is possible here (and below)?

dj2: Done. Removed this test and added the invalid values into the above test case.

jrprice (approved):
LGTM, thanks!

PR #1214, opened 2022-04-25
cmdline: Add --list-unimplemented

ben-clayton:
Prints all the unimplemented tests. Useful for tooling.

### DIFF
--- a/src/common/internal/test_group.ts
+++ b/src/common/internal/test_group.ts
@@ -31,6 +31,7 @@ export interface TestCaseID {
 
 export interface RunCase {
   readonly id: TestCaseID;
+  readonly isUnimplemented: boolean;
   run(
     rec: TestCaseRecorder,
     selfQuery: TestQuerySingleCase,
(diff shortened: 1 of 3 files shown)


ben-clayton:
You might be aware of a much nicer way to do this. This was just me inserting fields until I had a path from the command line runner to the `unimplemented()` calls.

kainino0x (approved):
lgtm, no problem with this approach.

The alternative would basically be to search for the string "TODO: unimplemented" or "TODO" in the description, which is how nodesWithTODO works:
https://github.com/gpuweb/cts/blob/8c2c490f50a24b2de0900a51194756044fcacaf2/src/common/internal/tree.ts#L366

That might be useful if you want to distinguish tests that are done vs partial vs unimplemented.

ben-clayton:
Thanks. I considered searching for `TODO`, but I was concerned `TODO` might crop up in descriptions for things like `TODO: Add more sub-cases`, which we'd incorrectly count.

Let's go with this for now then.

PR #1335, opened 2022-04-29
Don't block on subcases until the end of a testcase

austinEng:
This commit changes execution of a testcase such that await'ing
of all subcases is await'ed at the end of a testcase. This allows
better test throughput by allowing subsequent subcases to begin
without waiting on previous eventual expectations.

Because eventual expectations now record logs for a subcase
asynchronously after other subcases may have started running,
subcase logs are recorded in a deferred manner through a Proxy.
This allows logs to appear as if subcases were executed serially.

Separated out from PR #1172

<hr>

### DIFF
--- a/src/common/internal/test_group.ts
+++ b/src/common/internal/test_group.ts
@@ -498,39 +498,100 @@ class RunCaseSpecific implements RunCase {
           await this.beforeFn(sharedState);
         }
 
+        let allPreviousSubcasesFinalizedPromise: Promise<void> = Promise.resolve();
         if (this.subcases) {
           let totalCount = 0;
           let skipCount = 0;
-          for (const subParams of this.subcases) {
-            rec.info(new Error('subcase: ' + stringifyPublicParams(subParams)));
-            try {
-              const params = mergeParams(this.params, subParams);
-              const subcaseQuery = new TestQuerySingleCase(
-                selfQuery.suite,
-                selfQuery.filePathParts,
-                selfQuery.testPathParts,
-                params
-              );
-              await this.runTest(
-                rec,
-                sharedState,
-                params,
-                /* throwSkip */ true,
-                getExpectedStatus(subcaseQuery)
-              );
-            } catch (ex) {
-              if (ex instanceof SkipTestCase) {
-                // Convert SkipTestCase to info messages
-                ex.message = 'subcase skipped: ' + ex.message;
-                rec.info(ex);
-                ++skipCount;
-              } else {
-                // Since we are catching all error inside runTest(), this should never happen
-                rec.threw(ex);
-              }
+
+          // Maximum number of subcases in flight. If there are too many in flight,
+          // starting the next subcase will register `resolvePromiseBlockingSubcase`
+          // and wait until `subcaseFinishedCallback` is called.
+          const kMaxSubcasesInFlight = 500;
+          let subcasesInFlight = 0;
+          let resolvePromiseBlockingSubcase: (() => void) | undefined = undefined;
+          const subcaseFinishedCallback = () => {
+            subcasesInFlight -= 1;
+            // If there is any subcase waiting on a previous subcase to finish,
+            // unblock it now, and clear the resolve callback.
+            if (resolvePromiseBlockingSubcase) {
+              resolvePromiseBlockingSubcase();
+              resolvePromiseBlockingSubcase = undefined;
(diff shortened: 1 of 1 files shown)


austinEng:
PTAL only the last commit - to be rebased on #1333.

austinEng:
From the review on the other PR:
> collecting up the logs (maybe by creating a temporary recorder for each subcase?), flattening, and reporting them all at the end

I didn't do this because I found the proxy approach to be simpler and wouldn't require re-implementing the logic for computing severity levels, stack deduplication, and hiding log messages below a particular severity, etc.

src/common/internal/test_group.ts
```diff
@@ -411,7 +412,9 @@ class RunCaseSpecific implements RunCase {
         await this.fn(inst as Fixture & { params: {} });
       } finally {
         // Runs as long as constructor succeeded, even if initialization or the test failed.
-        await inst.doFinalize();
+        // Runs async without waiting for results (yet) so that subsequent subcases can start.
+        // All finalization steps will be waited on at the end of the testcase.
+        finalizePromise = inst.doFinalize();
```
kainino0x, line 417: I see now we're only pipelining the finalize steps, not the actual test bodies. Shouldn't it be possible pipeline the init and test functions too? By basically reverting these changes to runTest() and storing off the whole runTest() promise instead of just the finalizePromise. Guessing there's something that makes that harder than it seems?

src/common/internal/test_group.ts
```diff
@@ -472,11 +480,50 @@ class RunCaseSpecific implements RunCase {
...
+                      const rv = (prop as any).apply(target, args);
+                      // Because this proxy executes functions in a deferred manner,
+                      // it should never be used for functions that need to return a value.
+                      assert(rv === undefined);
+                    });
+                  };
+                }
+                return prop;
+              },
+            });
```
kainino0x, line 524: The proxy is cool but a bit hard to wrap my head around - would it be messy to implement the same thing directly on the TestCaseRecorder or a new class that wraps or subclasses it directly (just like the Proxy but as an actual class)? Other than having to manually list out all of the functions that are getting proxied here.

src/common/internal/test_group.ts
austinEng: We could certainly do that - do we want to?

if subcases have `await` inside the test body, then we'll have interleaved subcase execution. Maybe this might be surprising or unexpected for test authors.

kainino0x: It should be fine unless they share any state, which tests aren't supposed to (but that's hard to guarantee). I don't know how much of a performance gain we'd get, but there are some tests that do readbacks in the middle that should go faster.

We can start with just the finalizers if we know there's a concrete benefit from it.

src/common/internal/test_group.ts
```diff
@@ -423,15 +426,20 @@ class RunCaseSpecific implements RunCase {
...
+        .finally(() => {
+          try {
+            rec.endSubCase(expectedStatus);
+          } catch (ex) {
+            assert(ex instanceof UnexpectedPassError);
+            ex.message = `Testcase passed unexpectedly.`;
+            ex.stack = this.testCreationStack.stack;
+            rec.warn(ex);
+          }
+        });
```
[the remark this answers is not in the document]
austinEng: I think we still need the finally block to catch errors from constructor `new this.fixture(...` as well as performing `rec.endSubCase(...` if any error occurs before `inst.doFinalize()`

src/common/internal/test_group.ts
austinEng: Yea I'd probably just extend TestCaseRecorder and call it DeferredTestCaseRecorder and override the methods. I can do that if you'd like. I don't think it would work as nicely as the proxy though since I'd need to rebind `allPreviousSubcasesFinalizedPromise` each time it changes, or put it in a "box".

src/common/internal/test_group.ts
austinEng: Well, the implementation is simpler if we go fully async, so I'll do that for now then

src/common/internal/test_group.ts
kainino0x: Box seems like it would work pretty well, but leaving the proxy is probably fine, since it works.

kainino0x (approved):

PR #1338, opened 2022-05-03
Add --colors flag for enabling color output

ben-clayton:
Default to no-colors.

Allows users to write to logs without the ANSI escape codes.

### DIFF
--- a/src/common/runtime/cmdline.ts
+++ b/src/common/runtime/cmdline.ts
@@ -6,6 +6,7 @@ import { Logger } from '../internal/logging/logger.js';
 import { LiveTestCaseResult } from '../internal/logging/result.js';
 import { parseQuery } from '../internal/query/parseQuery.js';
 import { parseExpectationsForTestQuery } from '../internal/query/query.js';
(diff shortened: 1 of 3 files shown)


src/common/runtime/cmdline.ts
```diff
@@ -16,6 +17,7 @@ function usage(rc: number): never {
   console.log(`  tools/run_${sys.type} [OPTIONS...] QUERIES...`);
   console.log(`  tools/run_${sys.type} 'unittests:*' 'webgpu:buffers,*'`);
   console.log('Options:');
+  console.log('  --colors             Enable ANSI colors in output.');
```
dj2, line 20: Can we invert this to `--no-colors`? I'm going to forget to type it more on the command line then you would in a script I think?

ben-clayton: If you're using the `run-cts` script, then this will default enable it based on terminal detection logic.

dj2: That works.

dj2 (approved):

PR #1368, opened 2022-05-04
Allow eventualExpectOk to accept Promise.all

shrekshao:
Followup of #1344

This should avoid awaiting on each expectations when we have multiple parallel ones.


kainino0x (approved):
Nice convenience, LGTM

PR #1369, opened 2022-05-05
wgsl: Add support for f64 to Scalar

### LINKED ISSUE #1332: correctlyRoundedThreshold does not correctly handling some cases
While working on tests for degrees I have been seeing failures due to `57.2957763671875` not being considered correctly rounded for  `57.295780181884766`, which is correct, but is the end of test reported value, not necessarily what I passed in.

But after digging into it a bit the value is 1ULP below the f32 representation, but when doing correctly rounded this is not a match because the match is for either side of the f64 value and quantizing to f32 is rounding down.

It seemed suspicious that SwiftShader would be 2ULP off of the correct value, especially cause this was happening in a few test cases.

Digging into how I implemented correctlyRoundedThreshold, I realized that the target number is getting quantized for f32 as part of being passed made into a test case, which is wrong, since the internal implementation needs to know if rounding up or down was needed to quantize.

I need to rewrite how correctlyRounded is used, probably via making a comparator like the recent function based ulp comparator I added, so that the raw number can be passed into the implementation instead of quantizing it through a Scalar on the way.

zoddicus:
This is needed so that test expectation can be specified in a higher
precision than what WGSL is operating on (64-bits for 32-bits). This
allows for fluent specification of test conditions like correctly
rounded, instead of having to implement a per case Comparator.

This PR also renames some of the related helpers with similar names so
it is clearer what they return.

Fixes #1332

<hr>

### DIFF
--- a/src/webgpu/util/compare.ts
+++ b/src/webgpu/util/compare.ts
@@ -72,14 +72,13 @@ export function compare(got: Value, expected: Value, cmpFloats: FloatMatch): Com
     // Check types
     const gTy = got.type;
     const eTy = expected.type;
-    if (gTy !== eTy) {
-      if (!isFloatValue(got) || !isFloatValue(expected)) {
-        return {
-          matched: false,
-          got: `${Colors.red(gTy.toString())}(${got})`,
-          expected: `${Colors.red(eTy.toString())}(${expected})`,
-        };
-      }
+    const bothFloatTypes = isFloatValue(got) && isFloatValue(expected);
+    if (gTy !== eTy && !bothFloatTypes) {
+      return {
+        matched: false,
+        got: `${Colors.red(gTy.toString())}(${got})`,
+        expected: `${Colors.red(eTy.toString())}(${expected})`,
+      };
     }
   }
 
--- a/src/webgpu/util/compare.ts
+++ b/src/webgpu/util/compare.ts
@@ -1,6 +1,6 @@
 import { Colors } from '../../common/util/colors.js';
 
-import { f32, Scalar, Value, Vector } from './conversion.js';
+import { f32, isFloatValue, Scalar, Value, Vector } from './conversion.js';
 import { correctlyRounded, oneULP, withinULP } from './math.js';
 
 /** Comparison describes the result of a Comparator function. */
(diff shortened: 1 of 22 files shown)


src/webgpu/util/compare.ts
```diff
@@ -73,18 +73,20 @@ export function compare(got: Value, expected: Value, cmpFloats: FloatMatch): Com
     const gTy = got.type;
     const eTy = expected.type;
     if (gTy !== eTy) {
-      return {
-        matched: false,
-        got: `${Colors.red(gTy.toString())}(${got})`,
-        expected: `${Colors.red(eTy.toString())}(${expected})`,
-      };
+      if (!isFloatValue(got) || !isFloatValue(expected)) {
```
dj2, line 76: Not sure I understand this, previously if the numbers were not equal we'd fail, now we only fail if one of the things isn't a float?

src/webgpu/util/conversion.ts
```diff
@@ -666,3 +686,12 @@ export function vec4(x: Scalar, y: Scalar, z: Scalar, w: Scalar) {
 
 /** Value is a Scalar or Vector value. */
 export type Value = Scalar | Vector;
+
+/** @returns if the Value is a float scalar type */
+export function isFloatValue(v: Value): boolean {
+  if (v instanceof Scalar) {
+    const s = v as Scalar;
+    return s.type.kind === s.type.kind || s.type.kind === 'f32' || s.type.kind === 'f16';
```
dj2, line 694: Should this first one be `f64` instead of `s.type.kind`?

src/webgpu/util/compare.ts
zoddicus: I am loosening the restriction so that different float types will pass. I will take another look at if I can express this more clearly, since I agree it is a little obtuse

src/webgpu/util/conversion.ts
zoddicus: Yup, straight up typo

zoddicus (commented):
I need to take another pass at this, since in my degrees PR, which follows on, I am still seeing subtly wrong behaviours, just less of them.

src/webgpu/util/compare.ts
zoddicus: Pulling inner check into a variable, that allows for a more readable single check

zoddicus (commented):
The degree test issues I was seeing are related to how I was implementing derived/inherited accuracy, not this PR, so I think this is good to go.

dj2 (approved):

PR #1397, opened 2022-05-11
Fix unary tests with subnormal inputs

amaiorano:
Fixes:
webgpu:shader,execution,expression,call,builtin,ceil:*
webgpu:shader,execution,expression,call,builtin,floor:*
webgpu:shader,execution,expression,call,builtin,fract:*

Bug: tint:1546

Issue: crbug.com/tint/1546

### DIFF
--- a/src/webgpu/shader/execution/expression/call/builtin/fract.spec.ts
+++ b/src/webgpu/shader/execution/expression/call/builtin/fract.spec.ts
@@ -7,6 +7,7 @@ import { GPUTest } from '../../../../../gpu_test.js';
 import { correctlyRoundedMatch, anyOf } from '../../../../../util/compare.js';
 import { kBit, kValue } from '../../../../../util/constants.js';
 import { f32, f32Bits, TypeF32 } from '../../../../../util/conversion.js';
+import { isSubnormalNumber } from '../../../../../util/math.js';
 import { Case, Config, makeUnaryF32Case, run } from '../../expression.js';
 
 import { builtin } from './builtin.js';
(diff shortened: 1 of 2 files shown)


src/webgpu/shader/execution/expression/call/builtin/fract.spec.ts
```diff
@@ -51,7 +52,7 @@ Component-wise when T is a vector.
 
     const makeCase = (x: number): Case => {
       const result = x - Math.floor(x);
-      if (result < 1.0) {
+      if (result < 1.0 || flushSubnormalNumber(x) === 0.0) {
```
ben-clayton, line 55: `isSubnormalNumber()`?

ben-clayton (commented):
LGTM. Ryan for approvals

src/webgpu/shader/execution/expression/call/builtin/fract.spec.ts
amaiorano: Yeah, we could do, this is just the pattern established in other places. @zoddicus ?

zoddicus: I would prefer `isSubnormalNumber()`, other cases doing `flushSubnormalNumber(x)=== 0.0` are probably old code

zoddicus (approved):
lgtm
Thanks for fixing this, I am surprised this issue was present. I could have sworn I handled unary with subnormal inputs when originally refactoring all of this, but either I am mis-remembering or it got lost when I was porting things between branches.

src/webgpu/shader/execution/expression/call/builtin/fract.spec.ts
amaiorano: Will update with that change.

PR #1401, opened 2022-05-12
Remove dupes in anyOf to clean up error messages

amaiorano:
Many tests call calculateFlushedResults() that was returning a
Set<Scalar> of usually the same two Scalars (most values to not flush to
0). However, the set would always contain both values because Scalar is
a reference type, so two Scalar instances of the same value do not
compare equal (see https://stackoverflow.com/a/29759699 for why we can't
use Sets this way right now). As a result, most test failures would
output the same value for expectations, for example:

    (2.0584980561877986e+38(0x9add3b00000000000000000000000000, subnormal: false), 8.421775697122027e+37(0x3f5bc000000000000000000000000000, subnormal: false))
        returned: 1.5707963705062866(0x1.921fb6, subnormal: false)
        expected: 1.1824506521224976(0x1.2eb516, subnormal: false) or 1.1824506521224976(0x1.2eb516, subnormal: false)

With this change, calculateFlushedResults returns an Array<Scalar>
instead to make it obvious that Scalars of the same value are not being
dropped (and Array is likely more efficient than Set), and anyOf uses a
Set<String> and expectation.toString() to avoid comparing against the
same expectation twice. Note that this also works for Comparators, not
just Value.

Issue: #

<hr>


src/webgpu/shader/execution/expression/expression.ts
```diff
@@ -379,8 +379,8 @@ function packScalarsToVector(
 }
 
 /** @returns a set of flushed and non-flushed floating point results for a given number. */
-function calculateFlushedResults(value: number): Set<Scalar> {
-  return new Set([f64(value), f64(flushSubnormalNumber(value))]);
```
zoddicus, line 383: Would it make more sense to filter out the duplicate entries here? so that they don't propagate up and cause unneeded comparisons elsewhere.
Something like 
```
const subnormal_value = flushSubnormaNumber(value);
if (subnormal_value === value) {
  return  [f64(value)];
} else {
  return  [f64(value), f64(subnormal_value)];
}
```

src/webgpu/shader/execution/expression/expression.ts
```diff
@@ -379,8 +379,8 @@ function packScalarsToVector(
...
+function calculateFlushedResults(value: number): Array<Scalar> {
+  return new Array<Scalar>(f64(value), f64(flushSubnormalNumber(value)));
```
zoddicus, line 383: I think you can just use the short form `[ f64(value), f64(flushSubnormalNumber(value)) ]` here

src/webgpu/util/compare.ts
```diff
@@ -132,7 +132,13 @@ export function compare(got: Value, expected: Value, cmpFloats: FloatMatch): Com
 export function anyOf(...expectations: (Value | Comparator)[]): Comparator {
   return (got, cmpFloats) => {
     const failed: Array<string> = [];
+    const s = new Set<String>();
     for (const e of expectations) {
+      if (s.has(e.toString())) {
```
zoddicus, line 137: I am a touch leery of using string matching to do filtering on numbers, since I am not sure we guarantee that Value.toString() is unique for every possible value, since displayed digits might being truncated. If you could confirm that we are not truncating information that would alay my concern.

Though as I think about it, I think there is a bigger issue. Comparators don't define a .toString(), so I think you will get undefined for any e that is a Comparator, so you might be skipping later Comparators. I would move the check into the e as Value branch below to avoid this.

zoddicus (changes requested):

src/webgpu/shader/execution/expression/expression.ts
amaiorano: We could add this as an optimization, but it's not sufficient on its own because we can add more values via anyOf().

src/webgpu/util/compare.ts
amaiorano: > I am a touch leery of using string matching to do filtering on numbers, since I am not sure we guarantee that Value.toString() is unique for every possible value, since displayed digits might being truncated. If you could confirm that we are not truncating information that would alay my concern.

We defined toString() for Scalar which outputs (among other things) the 64-bit hex value. We also define toString for Vector in terms of the Scalar one. So every value will have a unique string.

> Though as I think about it, I think there is a bigger issue. Comparators don't define a .toString(), so I think you will get undefined for any e that is a Comparator, so you might be skipping later Comparators. I would move the check into the e as Value branch below to avoid this.

Comparators are functions, and their toString() actually outputs the definition of the function itself. I tested this locally by doing this in makeCase:

```
    const makeCase = (y: number, x: number): Case => {
      const c = makeBinaryF32Case(y, x, Math.atan2, true);
      const c2 = makeBinaryF32Case(y, x, Math.atan2, true);
      c.expected = anyOf(c.expected, c2.expected);
      return c;
    };
```

When it failed, it only output one value. However, I will admit that it's a bit confusing with the comparator in there. How about this instead: instead of skipping based on the `toString()` of `expectations` (which includes comparators), we skip the adding of `cmp.expected` to the `failed` array. This way, we're already dealing with strings, and just avoiding adding duplicate strings to the failed list. I'll do that, it's simpler and easier to understand.

src/webgpu/shader/execution/expression/expression.ts
amaiorano: Done.

src/webgpu/util/compare.ts
amaiorano: Updated with a new version that just avoids adding the same expected values to `failed`, which is now a `Set<string>` instead of an `Array<string>`. Note that the order of expected values should be in order of insertion, according to the [documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set).

zoddicus (approved):

PR #1402, opened 2022-05-12
DNS: Prototype for how to implement inherited FP accuracy tests

zoddicus:
I have re-expressed our fundamental accuracies, i.e. correctly
rounded, absolute, and ULP, in terms of intervals. Then built up a
framework to express other functions' accuracies in terms of these
building blocks.

This allows for composition of various blocks to express complex
accuracies like inherited accuracies.

I have shown how `cos`, `sin`, and `/` tests can be rewritten using
this framework, and implemented `tan` tests which our current
framework does not support.

As is, all of the cases I have implemented are monotonic. I believe
this framework can be extended to support non-monotonic functions via
custom impl/range implementations in a straightforward manner.

There is a bunch of duplicated patterns that I could refactor, and the
specific usage of inheritance can be revisited in a productionized
version of this.

A production version of this would probably involve a significant revamp
of the existing test runner/comparator code, since this supplants a lot
of the functionality in there and I just kinda bodge them together.

This does not handle accepting both rounding modes for quantization, I
have just forced quantization before sending it off, but our current
implementation doesn't actually handle this consistently either
as-is. I think adding this handling into my prototype would be easier,
since it would just be expanding out how normal vs subnormal is
handled.


src/webgpu/util/fp_interval.ts
```diff
@@ -0,0 +1,238 @@
...
+    });
+    return new FPInterval(Math.min(...mins), Math.max(...maxs));
+  }
+
+  public toString(): string {
+    return `[${this.begin}, ${this.end}]`;
+  }
+}
+
+export interface FPUnaryIntervalBuilder {
```
zoddicus, line 52: In a production version these will probably be super classes, so that I can roll up the boilerplate from below into it.

src/webgpu/util/fp_interval.ts
```diff
@@ -0,0 +1,238 @@
...
+  singular(x: number): FPInterval;
+  range?(i: FPInterval): FPInterval;
```
zoddicus, line 54: range is very intentionally optional, since it requires handling extremas in the interval correctly, and is only used if another function inherits accuracy from it. For monotonic functions implementing this is trivial, but needs custom code for non-monotonic functions, so I would only implement it for functions that we actually use it.

src/webgpu/util/fp_interval.ts
```diff
@@ -0,0 +1,238 @@
...
+      return new FPInterval(otherside, converted);
+    } else {
+      // x_32 rounded towards -inf, so is before x
+      const otherside = nextAfter(x_32, false, false).value as number;
+      return new FPInterval(converted, otherside);
+    }
+  }
+}
+
+class ULPFPIntervalBuilder implements FPUnaryIntervalBuilder {
```
zoddicus, line 129: ULP will still need the test case generator to calculate n for certain cases based off of the inputs into the tests, but we already do that, so there is no increase in complexity.

src/webgpu/util/fp_interval.ts
```diff
@@ -0,0 +1,238 @@
...
+    const ulp_flush = oneULP(x, true);
+    const ulp_noflush = oneULP(x, false);
+    const ulp = Math.max(ulp_flush, ulp_noflush);
+
+    return new FPInterval(x - this.n * ulp, x + this.n * ulp);
+  }
+}
+
+export class CosFPIntervalBuilder implements FPUnaryIntervalBuilder {
+  private readonly builder;
```
zoddicus, line 158: I am choosing to use composition here over inheritance, so that Builders that don't need a range() implementation don't inherit one from their parent

src/webgpu/util/fp_interval.ts
```diff
@@ -0,0 +1,238 @@
...
+  }
+}
+
+export class TanFPIntervalBuilder implements FPUnaryIntervalBuilder {
+  private readonly cos_builder;
+  private readonly sin_builder;
+  private readonly divide_builder;
+
+  public constructor() {
+    this.cos_builder = new CosFPIntervalBuilder();
```
zoddicus, line 224: In a production version, since these builders are always the same, I will probably figure out a way to properly cache them or make a table of them, instead of having each one creating instances. Possibly just use static classes/methods

src/webgpu/util/fp_interval.ts
```diff
@@ -0,0 +1,238 @@
...
+    } else {
+      // x_32 rounded towards -inf, so is before x
+      const otherside = nextAfter(x_32, false, false).value as number;
+      return new FPInterval(converted, otherside);
+    }
+  }
+}
+
+class ULPFPIntervalBuilder implements FPUnaryIntervalBuilder {
+  protected readonly n: number;
```
dneto0, line 130: n is the number of ULPs?  like for atan it would be 4096 for f32. ?

src/webgpu/util/fp_interval.ts
```diff
@@ -0,0 +1,238 @@
...
+  private impl(x: number): FPInterval {
+    return this.builder.singular(Math.sin(x));
+  }
+}
+
+export class DivisionFPIntervalBuilder implements FPBinaryIntervalBuilder {
+  private readonly builder;
+
+  public constructor() {
+    this.builder = new ULPFPIntervalBuilder(2.5);
```
dneto0, line 193: It's a detail, but the ULP only applies in a certain range. (I think it's the range of normal values?)

src/webgpu/util/compare.ts
```diff
@@ -174,3 +181,18 @@ export function ulpComparator(x: number, target: Scalar, n: (x: number) => numbe
     };
   };
 }
+
+export function intervalComparator(i: FPInterval): Comparator {
+  const match = intervalMatch(i);
+  return (got, _) => {
+    const cmp = compare(got, f64(0.0), match);
```
dneto0, line 188: This invocation actually ends up ignoring the 2nd arg (f64(0.0))?   Because the comparisons are actually being done by the matcher.  That's......a little too cute for production.   OR add a prominent comment.  :-)

dneto0 (commented):
I think this hangs together really well.  It's a nice framing of the problem: separating concerns and making good abstractions.

I was slowed down by lack of comments but I think I pieced it together.

I'm confused by the Abs builder's  impl implementation.... Seems wrong.  I think the others look good/ reasonable.

@ben-clayton should comment on how this replaces existing float comparator infra. I'm less familiar (i think it's evolved a lot).

src/webgpu/util/fp_interval.ts
```diff
@@ -0,0 +1,238 @@
+import { assert } from '../../common/util/util.js';
+
+import { kValue } from './constants.js';
+import { flushSubnormalNumber, nextAfter, oneULP } from './math.js';
+
+export class FPInterval {
```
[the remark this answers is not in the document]
zoddicus: For productionization, I think I will rename all of the FP* things F32* to make it clear that they are for f32, and when we start working on f16 we can address how we want to parameterize things.

zoddicus (commented):
Thank for the comments @dneto0 
Given that overall approach appears to be sound, I will start working on a more formal design doc.

zoddicus:
I have done a major rewrite of this, integrating in Ben's change to functional coding as well as other things that we discussed like handling inputs where accuracy isn't defined. PTAL

src/webgpu/shader/execution/expression/binary/f32_arithmetic.spec.ts
```diff
@@ -123,38 +125,43 @@ Expression: x / y
 Accuracy: 2.5 ULP for |y| in the range [2^-126, 2^126]
 `
   )
-  .params(u =>
-    u
-      .combine('storageClass', ['uniform', 'storage_r', 'storage_rw'] as const)
-      .combine('vectorize', [undefined, 2, 3, 4] as const)
+  .params(
+    u => u.combine('storageClass', ['uniform'] as const)
```
zoddicus, line 129: ugh, had this commented out for testing

src/webgpu/util/compare.ts
zoddicus: I agree this is too cute, I intend to take a pass at reworking how the whole runner/comparator thing is done once we are happy with how the intervals side of things is expressed.

src/webgpu/util/fp_interval.ts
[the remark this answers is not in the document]
zoddicus: I agree that your functional version is better and have merged it into my prototype.

src/webgpu/util/compare.ts
```diff
@@ -60,6 +61,12 @@ export function correctlyRoundedMatch(): FloatMatch {
   };
 }
 
+export function intervalMatch(i: F32Interval): FloatMatch {
```
ben-clayton, line 64: Do we still need `FloatMatch` ? Can we not entirely replace `FloatMatch` with the `intervalComparator` ? It would be a worthwhile goal, as it would let us entirely strip out the `cmpFloats` smell.

src/webgpu/util/f32_interval.ts
```diff
@@ -0,0 +1,213 @@
...
+      return this.begin === Number.NEGATIVE_INFINITY && this.end === Number.POSITIVE_INFINITY;
+    }
+    const i = toInterval(x);
+    return this.begin <= i.begin && this.end >= i.end;
+  }
+
+  /** Create interval with the tightest bounds that includes all provided intervals */
+  static span(...intervals: F32Interval[]): F32Interval {
+    assert(intervals.length > 0, `span of an empty list of F32Intervals is not allowed`);
+    const mins = new Set<number>();
```
ben-clayton, line 52: nit: Using sets here is seems unnecessarily expensive. Can't you just loop over the intervals and record the smallest of each interval's `.begin` and largest of each interval's `.end` ?

ben-clayton (commented):
Design is looking much nicer! Thank you for the changes!

zoddicus:
PTAL, I have done a major update of this prototype, porting over all of the existing FP tests to use the interval mechanism and refactoring/reorganizing the code.

I am going to start working on converting the test runner/float match code to be orientated around intervals, so we can get rid of  the 'cute' interval comparator hacks. That I expect to be the last update for this prototype, other than maybe responses to review comments.

For actually productionizing/landing this giant blob, I plan to break this up into a series of PRs and have them review separately. Specifically I am expecting to first land a patch with the core framework code + unit tests. Then a series of patches that convert tests and add unittests to the helpers, then finally a patch that removes the old framework code and replaces it with the interval-specific code. Then I can work on adding new tests.

src/webgpu/util/f32_interval.ts
ben-clayton: I wasn't suggesting brute-force testing all permutations as a replacement for Sets - removing duplicates is important for the reasons you gave. Just here, in `span`, which looks like very hot code, a single loop over the intervals and tracking min / max is clearly going to be faster than any implementation using sets.

Your latest implementation is exactly what I had imagined - looks great!

zoddicus: Sorry about this, I confounded two different areas I was working on and was responding to something else I had been thinking about vs what you had actually asked.

src/webgpu/util/f32_interval.ts
```diff
@@ -0,0 +1,629 @@
...
+/** Convert a point to an interval correctly rounded around the point */
+export function correctlyRoundedInterval(n: number): F32Interval {
+  return roundAndFlushPointToInterval(n, (n: number) => {
+    assert(!Number.isNaN(n), `absolute not defined for NaN`);
+    return toInterval(n);
+  });
+}
+
+/** Convert a point to an interval of absolute error around the point */
+export function absoluteErrorInterval(n: number, rng: number): F32Interval {
```
dneto0, line 180: nit: Please explain the 'rng'.  I don't understand the mnemonic.
It's the absolute error.

src/webgpu/util/f32_interval.ts
```diff
@@ -0,0 +1,629 @@
...
+export function ulpInterval(n: number, numULP: number): F32Interval {
+  numULP = Math.abs(numULP);
+  const ulp_flush = oneULP(n, true);
+  const ulp_noflush = oneULP(n, false);
+  const ulp = Math.max(ulp_flush, ulp_noflush);
+  return new F32Interval(n - numULP * ulp, n + numULP * ulp);
+}
+
+/** For a point operation run its impl over the input that may be an interval */
+function runPointImpl(x: F32Interval, impl: PointToInterval): F32Interval {
```
dneto0, line 200: It's worth noting that using this function on an interval only makes sense when the impl function is monotonic over the input interval.

zoddicus:
I removed FloatMatch and related complexity, still need to still promote intervals as a direct expectation, so that the IntervalComparator can be removed.

dneto0 (commented):
I looked in detail at f32_interval.ts, compare.ts, and spot checked a bunch of other things.

Overall this looks very good to me.

+1

src/webgpu/util/f32_interval.ts
```diff
@@ -0,0 +1,629 @@
...
+  return new F32Interval(n, n);
+}
+
+/** Does a number fit into the range of f32 [min, max]? */
+function isF32Finite(n: number) {
+  return n >= kValue.f32.negative.min && n <= kValue.f32.positive.max;
+}
+
+/** Operation that converts a point to an interval of valid results  */
+export interface PointToInterval {
```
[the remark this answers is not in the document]
zoddicus: I have flipped this a couple of times while working on this code, and I think I prefer Point over Unary. It reads slightly nicer to me. If someone has a specific objection to the terminology I can change it, I don't feel strongly about it.

src/webgpu/util/f32_interval.ts
zoddicus: Will do when productionizing

src/webgpu/util/f32_interval.ts
zoddicus: Will do when productionizing

zoddicus:
PTA(final)L, I have moved intervals up to being a type of expectation, so that they don't need to be wrapped in a comparator.

This I think is more or less the final form that I expect the production version of this to look like before I start adding in new tests and handling of non-monotonic functions.

I am going to start breaking this up into reviewable/landable PRs, unless there is any significant objections.

ben-clayton (commented):
Done a few skims over this, and I like it. Good job!
One question about quantizing.

src/webgpu/shader/execution/expression/call/builtin/ceil.spec.ts
```diff
@@ -39,39 +38,27 @@ g.test('f32')
       .combine('vectorize', [undefined, 2, 3, 4] as const)
   )
   .fn(async t => {
-    const cfg: Config = t.params;
-    cfg.cmpFloats = correctlyRoundedMatch();
-
     const makeCase = (x: number): Case => {
-      return makeUnaryF32Case(x, Math.ceil);
+      return makeUnaryF32IntervalCase(x, ceilInterval);
```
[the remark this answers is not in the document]
zoddicus: No, makeUnaryF32IntervalCase is responsible for quantizing to f32. I moved the quantization to the builders, since it needs to be done for every case for all of the test types, and made it explicit in the framework where/how the conversion is done.
This is also done right before calculating the acceptance interval, so we get rid of situations where we were actually calculating the interval for the non-quanitized input, which implementation will never actually see and getting subtle bugs.

zoddicus:
Added one last change that adds support for non-monotonic functions

src/webgpu/util/f32_interval.ts
```diff
@@ -98,32 +98,68 @@ function isF32Finite(n: number) {
...
 export interface PointToInterval {
   (x: number): F32Interval;
 }
 
-/** Operation that converts a pair of points to an interval of valid results  */
+/** Operation used to implement a PointToInterval  */
+interface PointToIntervalOp {
+  /** @returns acceptance interval at x */
+  impl: (x: number) => F32Interval;
+  /** Calculates where in x the min/max extrema of impl occur and returns the span of them.
```
dneto0, line 110: It's not clear what the "them" refers to:   The span of the domain values or the range values.
That is, the inputs to impl, or the outputs of impl?

From the usage below, namely `x = op.extrema(x);` I guess it's the *domain* of values.

src/webgpu/util/f32_interval.ts
```diff
@@ -197,47 +237,53 @@ export function ulpInterval(n: number, numULP: number): F32Interval {
...
+    roundAndFlushPointToInterval(x.begin, op),
+    roundAndFlushPointToInterval(x.end, op)
   );
 }
 
 /** For a binary operation run its impl over the inputs  that may be intervals */
-function runBinaryImpl(x: F32Interval, y: F32Interval, impl: BinaryToInterval): F32Interval {
+function runBinaryOp(x: F32Interval, y: F32Interval, op: BinaryToIntervalOp): F32Interval {
+  if (op.extrema !== undefined) {
+    [x, y] = op.extrema(x, y);
```
dneto0, line 257: I guess I'm confused.  For atan2, when the passed-in y interval contains 0, the resulting y interval is just [0,0].  
I'm not sure how that percolates through the rest of the calculation. Do we get the span over undefined values?

src/webgpu/util/f32_interval.ts
zoddicus: I agree, I will clarify these comments

src/webgpu/util/f32_interval.ts
zoddicus: {[0,0], x} will be passed into atan2, which will hit the special case for y === 0, which will create a span of {4096 * ulp(π), 4096 * ulp(0), 4096 * ulp(π)} (The zero term could be dropped since the other terms are on either side)

So the acceptance interval will be [-π, π].

Honestly atan2 is a hot mess of a trig function.

Taking a look at what Vulkan does. It looks like they are differentiating between the different quadrants based on the signs of y & x, and considering [0, 0] as undefined.

I think I need to rewrite this a bit given what Vulkan is doing and looking at the definition of atan2, I will put up another patch with corrected calculations that we can discuss in detail.

zoddicus: I think I have this cleaned up now. I have ended up implementing:
y != 0 && x != 0 -> 4096 * ulp(Math.atan2(y, x))
y == 0 && x == 0 -> Infinite (undefined)
y == 0 && x != 0 -> [-π, π]

I think in theory we should be able to use x > 0 vs x < 0 to distinguish between positive/negative π cases, but everytime I try that either SwiftShader or Nvidia will start failing tests due to having the other sign for some x values.

zoddicus:
This has been fully implemented in `main`, so no longer needed for reference.

PR #1412, opened 2022-05-16
Add all "recommended" typescript-eslint lints

kainino0x:
Enabling new lints, in 4 parts.

https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/README.md

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -28,28 +28,41 @@
 
     // Plugin: @typescript-eslint
     "@typescript-eslint/no-inferrable-types": "off",
-    "@typescript-eslint/no-unused-vars": ["warn", { "vars": "all", "args": "none" }],
     "@typescript-eslint/consistent-type-assertions": "warn",
+    // Recommended lints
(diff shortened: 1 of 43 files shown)


austinEng (commented):
the floating void syntax looks really weird - how much do we need this warning? Seem to be a lot of locations where we're suppressing it with `void`

kainino0x:
I think that warning is pretty valuable. In the past I've specifically commented in a number of places marking them as "noawait", i.e., I intentionally did not await this promise.

This lint actually did catch one test bug:
https://github.com/gpuweb/cts/pull/1412/commits/188093388d33ab43d93129b4a788be1f50aabc84#diff-2969bacf9f5413559e84acf16ca3771f3837005efc956f72cf70e6f4c2558275
```diff
-    t.testMapAsyncCall(true, null, buffer, mapMode);
+    await t.testMapAsyncCall(true, null, buffer, mapMode);
```

and also cleaned up some stuff that was unnecessarily async (not important)

austinEng (approved):
OK - seems worth it then

PR #1413, opened 2022-05-17
Add skipLibCheck to tsconfig.json

dj2:
This PR adds the `skipLibCheck` option into the `tsconfig.json` file and
turns it on. Adding the flag takes a local full rebuild of the CTS
using the dawn CTS runner from 14-16s down to 8-9s.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -16,7 +16,8 @@
     "allowUnreachableCode": true,
     /* Module Options */
     "moduleResolution": "node",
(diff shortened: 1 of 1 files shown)


kainino0x:
Nice! Seems safe to enable, except maybe when we're testing out changes to https://github.com/gpuweb/types
Hopefully that's not a problem.

kainino0x (approved):

PR #1424, opened 2022-05-17
Duplicate params: catch in ParamsBuilder, improve error in loadTreeForQuery

### LINKED ISSUE #1409: `insertLeaf` not setting `key` correctly.
In the `insertLeaf` code we create a `key` (https://github.com/gpuweb/cts/blob/9f4227be09ee268c508e9d5841c4f76772036264/src/common/internal/tree.ts#L530) which is set initialized to blank. The key is then used when inserting into the `parent.children` object.

As far as I can tell, the key is always blank. This causes issues as sometimes the same parent tries to insert two children with the same blank key and triggers they assert about a duplicate key.

kainino0x:
The check in ParamsBuilder is a bit more local than the one in
loadTreeForQuery, so hopefully easier to debug.

Only combine()/combineWithParams() can catch duplicate params
efficiently (just once, instead of once per parent parameterization), so
other methods don't check for them.
In those cases, the assertion in loadTreeForQuery will be hit, so
improve the error for that assertion, too.

Issue: fixes #1409

<hr>

### DIFF
--- a/src/common/framework/params_builder.ts
+++ b/src/common/framework/params_builder.ts
@@ -1,4 +1,6 @@
 import { Merged, mergeParams } from '../internal/params_utils.js';
+import { stringifyPublicParams } from '../internal/query/stringify_params.js';
+import { assert, mapLazy } from '../util/util.js';
 
 // ================================================================
 // "Public" ParamsBuilder API / Documentation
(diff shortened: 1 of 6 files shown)


src/common/framework/params_builder.ts
```diff
@@ -158,9 +160,17 @@ export class CaseParamsBuilder<CaseP extends {}>
...
     return this.expandWithParams(function* (p) {
+      const seenValues = new Set();
       for (const value of expander(p)) {
         // TypeScript doesn't know here that NewPKey is always a single literal string type.
-        yield { [key]: value } as { [name in NewPKey]: NewPValue };
+        const entry = { [key]: value } as { readonly [name in NewPKey]: NewPValue };
+        assert(
+          !seenValues.has(value),
+          `Duplicate value for param "${key}": ${stringifyPublicParams({ ...p, ...entry })}`
+        );
```
kainino0x, line 170: Oh, this checks for duplicates on every iteration. I was going to fix that and only do it once.

dj2 (approved):

dj2:
Maybe set Issue #1409 ?

kainino0x:
Oh yes, thanks. I just pushed a refactor to fix the inefficiency, so PTAL again (sorry)

dj2 (approved):
still lgtm

Issue #910, opened 2022-01-13
web_platform: test copies and imports from videos with color profiles, rotation/crop, etc.


kainino0x:

kainino0x:
Tentatively assigning @bjjones since you've been looking at implementation. We can land tests for copies and imports separately as needed.

Issue #1121, opened 2022-03-30
helpers: use texture_ok helpers in more tests


kainino0x:
Followup to #881

Need to look around for other tests that should use these helpers instead of custom stuff. 
Main candidates are `expectSinglePixelIn2DTexture` and sibling helpers.

PR #1471, opened 2022-05-25
Revise ExternalTexture Video Test Expectations

bjjones:
Revises external texture video test expectations to test for accurate
color space conversions. Adds BT.709 and BT.2020 video samples.

Issue: #910

<hr>


src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -4,6 +4,7 @@ Tests for external textures from HTMLVideoElement (and other video-type sources?
 - videos with various encodings, color spaces, metadata
 
 TODO: consider whether external_texture and copyToTexture video tests should be in the same file
+TODO: add a test configuration for red-green.mp4 to the importExternalTexture Sample test once the infrastructure can load it succesfully.
```
kainino0x, line 7: What infrastructure can't load it?

If it's in Chromium, we should just add the test and add a suppression in Dawn's `expectations.txt`. (Though this can be a separate PR if it's nontrivial.)

src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -14,11 +15,38 @@ import { startPlayingAndWaitForVideo } from '../../web_platform/util.js';
...
+    video: 'red-green.bt709.vp9.webm',
+    redExpectation: new Uint8Array([0xff, 0x00, 0x00, 0xff]),
+    greenExpectation: new Uint8Array([0x00, 0xff, 0x00, 0xff]),
+  },
+  {
+    video: 'red-green.bt2020.vp9.webm',
+    redExpectation: new Uint8Array([0xff, 0x00, 0x00, 0xff]),
+    greenExpectation: new Uint8Array([0x00, 0xff, 0x00, 0xff]),
+  },
 ];
```
kainino0x: If you use `as const` here you probably won't need to define the `VideoExpectations` type.
```
] as const;
```

src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -99,17 +127,16 @@ g.test('importExternalTexture,sample')
...
-a microtask.
-TODO: Multiplanar scenarios
+supported video formats {vp8, vp9, ogg}, common source colorspaces {bt.601, bt.709, bt.2020}, 
+and to ensure the GPUExternalTexture is destroyed by a microtask.
 `
   )
   .params(u =>
     u //
-      .combine('videoSource', kVideoSources)
+      .combine('videoExpectations', kVideoExpectations)
```
kainino0x: This adds the expectations to the query string, which is unnecessary. You can get them passed through with a "private" param which is a param that starts with `_`. I'll do this in a commit that I'm pushing to your branch.

src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -310,7 +337,7 @@ Tests that we can import an HTMLVideoElement into a GPUExternalTexture and use i
         kFormat,
         { x: 1, y: 0 },
         {
-          exp: new Uint8Array([0x00, 0xff, 0x00, 0xff]),
+          exp: kVideoExpectations[0].greenExpectation,
         }
```
kainino0x, line 340: FYI, if you run into any issues with precision (needing a threshold because there is a range of possible results): #1121

src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -242,7 +269,7 @@ Tests that we can import an HTMLVideoElement into a GPUExternalTexture and use i
 `
   )
   .fn(async t => {
-    const videoUrl = getResourcePath('red-green.webmvp8.webm');
+    const videoUrl = getResourcePath(kVideoExpectations[0].video);
```
kainino0x: Any reason this test can only run this case and not the others?

kainino0x (approved):

kainino0x:
I pushed some changes to your branch so you'll need to pull from it before making any further revisions.

src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -99,14 +122,13 @@ g.test('importExternalTexture,sample')
   .desc(
     `
 Tests that we can import an HTMLVideoElement into a GPUExternalTexture, sample from it for all
-supported video formats {vp8, vp9, ogg, mp4}, and ensure the GPUExternalTexture is destroyed by
-a microtask.
-TODO: Multiplanar scenarios
+supported video formats {vp8, vp9, ogg}, common source colorspaces {bt.601, bt.709, bt.2020},
+and to ensure the GPUExternalTexture is destroyed by a microtask.
```
shaoboyan, line 126: Note that we don't need to ensure GPUExternalTexture is destroyed by a microtask. It is not true for current spec any more.

kainino0x: Good catch, I didn't think about this test description very closely.

I assume this test still tests that it's destroyed in a microtask, so please add a TODO here and then we should actually fix the test separately.

bjjones: This test doesn't actually test the destroy behavior - my guess is that I decided to split the test out in a prior commit. It looks like Shaobo is already revising it, so I'll just remove it from the description and leave the destroy test as is.

src/webgpu/web_platform/external_texture/video.spec.ts
bjjones: I don't believe there's any restriction. I went ahead and added all of them.

src/webgpu/web_platform/external_texture/video.spec.ts
bjjones: ack. It's consistent for me on my test platform, but I will not be surprised if a tolerance is needed when it is running on all platforms.

src/webgpu/web_platform/external_texture/video.spec.ts
bjjones: Removed the TODO. Not sure if this is a cts issue or a Chromium issue, but I've noted it in a bug here:  https://bugs.chromium.org/p/dawn/issues/detail?id=1432

At the least, running it the standalone cts with Chromium will always fail to load. It hits this error: https://github.com/gpuweb/cts/blob/main/src/webgpu/web_platform/util.ts#L36

It's already disabled in chromium here: https://source.chromium.org/chromium/chromium/src/+/main:content/test/gpu/gpu_tests/test_expectations/webgpu_cts_expectation_queries.js;l=1618

Note that all external texture tests have been disabled in Chromium for awhile for some unknown reason, which I intend to turn back on soon here: https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/webgpu-cts/expectations.txt;l=242?q=expectations.txt

src/webgpu/web_platform/external_texture/video.spec.ts
kainino0x: Should we dry run this? I forget where the instructions are, but they've changed anyway, so I may as well try it out and document them.

src/webgpu/web_platform/external_texture/video.spec.ts
kainino0x: Ah, cool, I see now the mp4 was already there. I think the parameters match now so the existing expectation should still work. (A dry run will double check this, too.)

src/webgpu/web_platform/external_texture/video.spec.ts
kainino0x: Turns out the dry run process hasn't been simplified the way I thought it might be yet, but here's one with the old method:
https://dawn-review.googlesource.com/c/dawn/+/91760

It has changed slightly so hopefully I got it right.

- check the current CTS revision in Dawn's DEPS file
- rebase the candidate change on that revision
- push to a branch in gpuweb/cts
- change Dawn's DEPS file to the revision of that branch
- wait for the branch to get mirrored to https://chromium.googlesource.com/external/github.com/gpuweb/cts/
- run `gen_ts_dep_lists.py`
- commit and upload a dawn PR
- run `dawn-*-deps-rel` tryjobs

bjjones: I'm not sure. I added a Mac tryjob to your CL, which is the one I'm worried about because of this comment: https://source.chromium.org/chromium/chromium/src/+/main:ui/gfx/color_space.h;l=91

I don't know if this enum is even used in the path we're using, and it may not have any impact anyway since we're testing primaries.

kainino0x: Thanks. I only didn't add Mac at first because I wanted to make sure it was building at all (which failed the first time around).

That comment sounds like it applies only to the final display of the overlays, and not the data we would get by reading out of textures, but it's vague so hard to be sure. Good to test.

kainino0x: (BTW, once those pass, feel free to land this PR.)

bjjones: Passed. Merging

PR #1506, opened 2022-06-03
Roll @webgpu/types to 0.1.18

kainino0x:
I just rolled the spec into the types (to get the alphaMode update) and it turns out we haven't rolled the types here in a while. The changes aren't too substantial, but each different change is separated into a different commit.

Issue: None

<hr>


kainino0x:
@toji PTAL (specifically because I wanted to ask - the layout:'auto' changes are implemented, right, so this should be no problem when rolling into Chromium?)

toji (approved):
LGTM! And yes, the 'auto' change is in Chromium at this point, so all those updates should be fine.

PR #1510, opened 2022-06-06
Change Scalar.toString() to include hex bits

zoddicus:
Currently this is includes the value of the Scalar in base 16, which
isn't that useful. For unsigned integers this maps directly to the
bits, but for pretty much any other numeric type it is just making a
harder to read version of toString().

I am pretty sure the intent here was to include the bits in a readable
format so that one could easily evaluate what is being stored.

This changes the return of toString() from:
```
-1.00048828125(0x-1.002, subnormal: false)
```
to this:
```
-1.00048828125 (hex: 0xbf801000, subnormal: false)
```
<hr>

### DIFF
--- a/src/webgpu/util/conversion.ts
+++ b/src/webgpu/util/conversion.ts
@@ -498,19 +498,24 @@ export class Scalar {
       return Colors.bold(this.value.toString());
     }
     switch (this.value) {
-      case 0:
       case Infinity:
       case -Infinity:
         return Colors.bold(this.value.toString());
       default: {
+        // Uint8Array.map returns a Uint8Array, so cannot use .map directly
+        const hex = Array.from(this.bits)
+          .reverse()
+          .map(x => x.toString(16).padStart(2, '0'))
+          .join('');
         const n = this.value as Number;
-        if (n !== null) {
-          return (
-            Colors.bold(this.value.toString()) +
-            `(0x${this.value.toString(16)}, subnormal: ${isSubnormalNumber(n.valueOf())})`
-          );
+        if (n !== null && isFloatValue(this)) {
+          let str = this.value.toString();
+          str = str.indexOf('.') > 0 || str.indexOf('e') > 0 ? str : `${str}.0`;
+          return isSubnormalNumber(n.valueOf())
+            ? `${Colors.bold(str)} (0x${hex} subnormal)`
+            : `${Colors.bold(str)} (0x${hex})`;
         }
-        return Colors.bold(this.value.toString()) + `(0x${this.value.toString(16)})`;
+        return `${Colors.bold(this.value.toString())} (0x${hex})`;
       }
     }
   }
@@ -699,7 +704,7 @@ export type Value = Scalar | Vector;
 export function isFloatValue(v: Value): boolean {
   if (v instanceof Scalar) {
     const s = v;
-    return s.type.kind === s.type.kind || s.type.kind === 'f32' || s.type.kind === 'f16';
+    return s.type.kind === 'f64' || s.type.kind === 'f32' || s.type.kind === 'f16';
   }
   return false;
 }


src/webgpu/util/conversion.ts
```diff
@@ -503,14 +503,19 @@ export class Scalar {
...
+        const hex = Array.from(this.bits)
+          .reverse()
+          .map(x => x.toString(16).padStart(2, '0'))
+          .join('');
         const n = this.value as Number;
         if (n !== null) {
           return (
             Colors.bold(this.value.toString()) +
-            `(0x${this.value.toString(16)}, subnormal: ${isSubnormalNumber(n.valueOf())})`
+            ` (hex: 0x${hex}, subnormal: ${isSubnormalNumber(n.valueOf())})`
```
amaiorano, line 515: I think the "hex: " prefix is redundant here.

ben-clayton: +1

Also:

* How can you tell the difference between an integer and float without looking at the hex value? `1 (0x00000001)` vs `1 (0x3F800000)`? What do we think about always printing a `.0` if the number has no fractional part?

* I'm a bit concerned about the verbosity of always printing `subnormal: false`. Isn't this used to print the scalars of a vector?
What do we think about only printing `subnormal` if the number... is subnormal? e.g. `1 (0x3F800000)` vs `5.877472e-39 (0x00400000 subnormal)`

amaiorano: > What do we think about only printing subnormal if the number... is subnormal?
I added this feature, and my initial implementation did exactly as you suggest. Ryan felt it was less surprising to have it always output something. This doesn't bother me too much either way.

zoddicus: I will drop the `hex:`

For Scalars, to tell the difference `1 (0x00000001)` vs `1 (0x3F800000)`, you would have to check the value in this.type. 

I think I am missing something wrt 'always printing `.0`', wouldn't that just cause more confusion? because both integer adn float one would print as 1.0? or is the intent for them to be the same and use the hex to differentiate?

I don't think I properly considered vectors of scalars when reviewing @amaiorano's original subnormal patch. I buy the verbosity argument, so will change it.

ben-clayton: Sorry, I mistyped. I meant always print a `.0` **if the type is floating-point**. You're right - this would involve looking at `this.type.kind`. That or we add some `isFloatingPoint()` helper on `Scalar`.

zoddicus:
@ben-clayton  PTAL

src/webgpu/util/conversion.ts
```diff
@@ -498,19 +498,26 @@ export class Scalar {
...
+          .map(x => x.toString(16).padStart(2, '0'))
+          .join('');
         const n = this.value as Number;
         if (n !== null) {
-          return (
-            Colors.bold(this.value.toString()) +
-            `(0x${this.value.toString(16)}, subnormal: ${isSubnormalNumber(n.valueOf())})`
-          );
+          const str =
+            isFloatValue(this) && Number.isInteger(n)
```
ben-clayton, line 513: nit: `1.0e+20` is an integer, but shouldn't be printed as `1.0e+20.0`.

For #1515, I used the helper:

```
    const valueWithPoint = (x: number) => {
      const str = `${x}`;
      return str.indexOf('.') > 0 ? str : `${str}.0`;
    };
```

ben-clayton (approved):
One nit, otherwise LGTM

src/webgpu/util/conversion.ts
zoddicus: I agree that my version is not correct either, but I think your version still adds in unexpected `.0`s

For `1e+21` I am seeing `1e+21.0`

https://www.typescriptlang.org/play?#code/MYewdgzgLgBAZjAvDAjABg5r2cG4BQokIANgKYB0JIA5gBQAGAJAN5wC+DAlAUdDADcAhiQCuZAOoBLKAAsACiClhYyOgA8AXDDCiAtgCMyAJy5IAfDBb4YtmH1jRjSGMxbrOBOzGNkoo4zAYJwplABMydQB5ODoAcgo4s0s0GAB+YKhnbTcndgo0BgJ2XnAIUkpqejdhMUkZBSUVOjguTh4gA

ben-clayton: Humm, good spot. I thought it always printed the `.0` when it's going to emit the `e`. This seems to work:

```
const valueWithPoint = (x: number) => {
    const str = `${x}`;
    return (str.indexOf('.') > 0) || (str.indexOf('e') > 0) ? str : `${str}.0`;
};
```

zoddicus: That looks like it catches all of the cases I can quickly think of :+1:

PR #1514, opened 2022-06-07
[wgsl] use subcases in texture tests

dj2:
This CL uses subcases to break the texture execution tests into logical chunks based on
the type of texture being tested.

<hr>

### DIFF
--- a/src/webgpu/shader/execution/expression/call/builtin/textureGather.spec.ts
+++ b/src/webgpu/shader/execution/expression/call/builtin/textureGather.spec.ts
@@ -57,6 +57,7 @@ Parameters:
   )
   .params(u =>
     u
+      .beginSubcases()
       .combine('T', ['f32', 'i32', 'u32'] as const)
       .combine('S', ['clamp-to-edge', 'repeat', 'mirror-repeat'])
       .combine('C', ['i32', 'u32'] as const)
@@ -87,6 +88,7 @@ Parameters:
   )
   .params(u =>
     u
+      .beginSubcases()
       .combine('T', ['f32', 'i32', 'u32'] as const)
       .combine('S', ['clamp-to-edge', 'repeat', 'mirror-repeat'])
       .combine('C', ['i32', 'u32'] as const)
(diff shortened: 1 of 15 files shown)


src/webgpu/shader/execution/expression/call/builtin/textureGather.spec.ts
```diff
@@ -57,6 +57,7 @@ Parameters:
   )
   .params(u =>
     u
+      .beginSubcases()
```
kainino0x, line 60: note: you can alternatively use `.paramsSubcasesOnly(u =>` instead of `.beginSubcases()` at the start, but it's equivalent so either is fine.

src/webgpu/shader/execution/expression/call/builtin/textureStore.spec.ts
```diff
@@ -38,6 +38,7 @@ Parameters:
   )
   .params(u =>
     u
+      .beginSubcases()
       .combine('F', TexelFormats)
```
kainino0x, line 42: You may want to have texel format be a non-subcase parameter, since it would probably be relatively common for an implementation to be missing support for a texel format (e.g. if we later add a new format that can be used for storage and the test gets implemented before all the browsers implement it). We do this for texture formats in most tests for this reason.

---

As a side note, the "shader representation" part of TexelFormats isn't a necessary part of the identifying info of the testcase. So in this case we would usually either write something like:

- use an underscore to make it not part of the "public"/"identifying" parameters
```
export const kTexelFormatParams = [
  { format: 'rgba8unorm', _shaderType: 'f32' },
  // ...
];
```

`.params(u => u.combineWithParams(kTexelFormatParams)`

- look up the shader type separately (in roughly the style of `capability_info.ts`)
```
export const kStorageTextureFormatInfo = {
  'rgba8unorm': { shaderType: 'f32' },
  // ...
];
export const kStorageTextureFormats = keysOf(kStorageTextureFormatInfo);
```

`.params(u => u.combine('F', kStorageTextureFormats)`

`const { shaderType } = kStorageTextureFormatInfo[t.params.F];`

kainino0x (approved):

src/webgpu/shader/execution/expression/call/builtin/textureGather.spec.ts
dj2: done

src/webgpu/shader/execution/expression/call/builtin/textureStore.spec.ts
dj2: Perfect, thanks. Done.

ben-clayton (approved):

PR #1518, opened 2022-06-09
Add wrapper around ULP calculation to handle flush vs no flush

### LINKED ISSUE #1516: Refactor `oneULP` utility into `oneULP` + `oneULPImpl`
Currently `oneULP` takes in a boolean that indicates the flushing behaviour.
For uses what is wanted is the max ULP at a point, so this patterns shows up `Math.max(oneULP(x, ..., false), oneULP(x, ..., true)` at many call sites.

This is error prone cause you have to remember when calling that you need to check both results, and generally leads to duplicate code. `oneULP` should be refactored so that the current function becomes `oneULPImpl`, and `oneULP`, does the max of flushing behaviour via calling `oneULPImpl` twice. If there are specific call sites that a specific flushing behaviour is desired, `oneULPImpl` can be called directly.

zoddicus:
Fixes #1516

<hr>

### DIFF
--- a/src/unittests/maths.spec.ts
+++ b/src/unittests/maths.spec.ts
@@ -17,6 +17,7 @@ import {
   linearRange,
   nextAfter,
   oneULP,
+  oneULPImpl,
   withinULP,
 } from '../webgpu/util/math.js';
 
@@ -182,7 +183,7 @@ interface OneULPCase {
   expect: number;
 }
 
-g.test('oneULPFlushToZero')
+g.test('oneULPImplFlushToZero')
   .paramsSimple<OneULPCase>([
     // Edge Cases
     { target: Number.NaN, expect: Number.NaN },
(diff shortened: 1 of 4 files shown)


src/webgpu/util/math.ts
```diff
@@ -147,17 +147,16 @@ export function nextAfter(val: number, dir: boolean = true, flush: boolean): Sca
...
- * for a more detailed/nuanced discussion of the definition of ulp(x).
+ * This is the main implementation of oneULP, which is normally what should be
+ * used. This should only be called directly if a specific flushing mode is
+ * required.
  *
  * @param target number to calculate ULP for
  * @param flush should subnormals be flushed to zero
  */
-export function oneULP(target: number, flush: boolean): number {
+export function oneULPImpl(target: number, flush: boolean): number {
```
ben-clayton, line 159: I'm really nitpicking here, but I'm not a massive fan of implementation details leaking into public APIs, even if it's exclusively used for testing.
I'd probably opt for the interface:
`oneULP(target: number, flush?: boolean)`

Where omitting `flush` behaves like `oneULP` and providing it acts like `oneULPImpl`.

src/unittests/maths.spec.ts
```diff
@@ -244,35 +245,88 @@ g.test('oneULPNoFlush')
...
     { target: -1e40, expect: hexToF32(0x73800000) }, // negative out of range
   ])
   .fn(t => {
     const target = t.params.target;
-    const got = oneULP(target, false);
+    const got = oneULPImpl(target, false);
+    const expect = t.params.expect;
+    t.expect(
+      got === expect || (Number.isNaN(got) && Number.isNaN(expect)),
+      `oneULPImpl(${target}, true) returned ${got}. Expected ${expect}`
```
ben-clayton, line 280: s/`true`/`false`

ben-clayton (approved):
Approval with nits.

PR #1520, opened 2022-06-10
Review createRenderPipeline.spec.ts

shrekshao:
First batch of `createRenderPipeline.spec.ts` review:
  - createRenderPipeline validation tests should be complete now (with unimplemented parts listed).
  - Did some file regroup. (hopefully cts roll script handles the rename)

Tests failed in Chrome:
  - fragment_state,max_color_attachments_limit (https://dawn-review.googlesource.com/c/dawn/+/93307, https://chromium-review.googlesource.com/c/chromium/src/+/3704453)
  - primitive_state,unclipped_depth:isAsync=true;unclippedDepth=true (https://dawn-review.googlesource.com/c/dawn/+/93309)
  - multisample_state,alpha_to_coverage_enabled:isAsync=true;alphaToCoverageEnabled=true;hasSampleMaskOutput=true;mask=4294967295
  - shader_module,entry_point (known issue suppressed in `createComputePipeline.spec.ts`)

----
Spec validation text

- ☑ vertex state {valid, invalid} (`vertex_state.spec.ts`) 
- ☑ fragment state {valid, invalid}
    - ☑ targets length {0, 1, 2, 8, 16, 1024} must be <= device.limits.maxColorAttachments
    - For each non-null color state |target|
        - ☑ target.format must has render attachment capability
        - If target.blend is not undefined
            - ☑ target.format must be filterable
            - ☑ target.blend.color {valid, invalid}
            - ☑ target.blend.alpha {valid, invalid}
        - ☑ target.writeMask must be < 16
        - ☑ pipeline output and format compatibility (pipeline_output_targets)
 - ☑ primitive state
 - ☑ depth stencil state
 - ☑ rasterizationState various values (dynamic_state.spec.ts)
 - ☑ multisample state
   - ☑ count {0, 1, 3, 4, 8, 16, 1024}
   - ☑ sample mask {0, 0xFFFFFFFF}
   - ☑ alphaToCoverage:
     - ☑ alphaToCoverageEnabled is { true, false } and sampleCount { = 1, = 4 }.
       The only failing case is (true, 1).
     - ☑ sample_mask builtin is a pipeline output of fragment and
       alphaToCoverageEnabled is { true (fails), false (passes) }.
     - ☑ sampleMask is being used and alphaToCoverageEnabled is { true (fails), false (passes) }.
 - TODO: inter-stage validating

Questions:
- The file naming convention seems a bit inconsistent here, do we want to make them consistent using `a_b_c` style?
[...]

PR #1536, opened 2022-06-15
Remove redundant entry point tests

shrekshao:
Follow up of #1520
(As these tests are already splited into `validation/shader_module/entry_point.spec.ts`)


kainino0x (approved):
lgtm ~but needs a buildfix~

Issue #919, opened 2022-01-13
val: capability_checks,features errors should be exceptions


kainino0x:
Need to update query_types tests

kainino0x:
Fixed by https://github.com/gpuweb/cts/pull/1499

kainino0x:
Wait, no, #1499 only fixed the texture cases. Seems like there may also be fixes needed in other places.

jchen10:
@kainino0x I am working on the optimization of 0-copy video import on ChromeOS. It's unfortunately no better than WebGL yet. I may still need a couple of weeks on that. If this task is urgent, please assign it to someone else. Otherwise I would take case of it as soon as I've done my current work.

kainino0x:
No worries, this isn't urgent. I just was looking at it because Brandon was fixing some of these.

@toji if you *happen* to want to take the rest of these then you would be welcome to.

kainino0x:
Assigning @Gyuyoung who started working on this.

#1571 fixes the existing test for createQuerySet, but it looks like we are still missing a test for writeTimestamp, which should also throw a TypeError when the feature is not enabled.

Gyuyoung:
@kainino0x I added the test to check if writeTimestamp test throws an exception in #1584. So, I would like to mark this task as `completed`. If we have more tasks to complete, please let me know.

kainino0x:
I don't recall for sure, but according to my last comment it sounds like it's complete!

PR #1571, opened 2022-06-23
Update the createQuerySet test of query_types.spec.ts to expect an exception

Gyuyoung:
This PR updates the existing createQuerySet test of query_types.spec.ts
in order to expect exceptions rather than validation errors when the device
features don't contain 'timestamp-query'.

Issue: #919 

<hr>

### DIFF
--- a/src/webgpu/api/validation/capability_checks/features/query_types.spec.ts
+++ b/src/webgpu/api/validation/capability_checks/features/query_types.spec.ts
@@ -10,38 +10,35 @@ export const g = makeTestGroup(ValidationTest);
 g.test('createQuerySet')
   .desc(
     `
-Tests that creating query set shouldn't be valid without the required feature enabled.
-- createQuerySet
-  - type {occlusion, timestamp}
-  - x= {pipeline statistics, timestamp} query {enable, disable}
-
-TODO: This test should expect *synchronous* exceptions, not validation errors, per
-<https://github.com/gpuweb/gpuweb/blob/main/design/ErrorConventions.md>.
-As of this writing, the spec needs to be fixed as well.
+  Tests that creating a query set throws a type error exception if the features don't contain
+  'timestamp-query'.
+    - createQuerySet
+      - type {occlusion, timestamp}
+      - x= {pipeline statistics, timestamp} query {enable, disable}
   `
   )
   .params(u =>
     u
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/capability_checks/features/query_types.spec.ts
```diff
@@ -10,38 +10,35 @@ export const g = makeTestGroup(ValidationTest);
...
   .fn(async t => {
-    const { type, timestampQueryEnable } = t.params;
+    const { type, featureContainsTimestampQuery } = t.params;
 
     const count = 1;
-    const shouldError = type === 'timestamp' && !timestampQueryEnable;
+    const shouldException = type === 'timestamp' && !featureContainsTimestampQuery;
 
-    t.expectValidationError(() => {
+    t.shouldThrow(shouldException ? 'TypeError' : false, () => {
```
Gyuyoung, line 41: When type="timestamp" and featureContainsTimestampQuery=false, the test fails because a validation error is generated instead of a type error exception. I guess that Chromium doesn't generate an exception yet. Anyway, I update the code to expect an exception since the specification says it should throw a type error exception in such a case. But, I wonder if changing to expect an exception makes sense here now. wdyt?

kainino0x: This test looks correct. We should always update the CTS, even when Chromium is outdated, as we are creating a conformance target for Chromium and all other browsers to aim for.

(We always want CTS to match the latest agreed-upon standard, but unfortunately sometimes the spec has errors or is outdated relative to the standard. So when you find a difference between the spec and Chromium, it's good to double check which one is wrong, by checking for open issues on crbug.com, crbug.com/dawn, or gpuweb/gpuweb.)

kainino0x (approved):

src/webgpu/api/validation/capability_checks/features/query_types.spec.ts
Gyuyoung: ok, I will keep working on the task like this. Thanks!

toji: In support of what Kai said, as we get closer to releasing the API the gaps between the spec and the implementations will become less obvious. Having visible failures in the CTS helps guide our implementation efforts.

Thank you!

Issue #923, opened 2022-01-13
val: resource_usages,texture,in_pass_encoder


kainino0x:
I don't think it needs a full review, just need to split up the test plan into descriptions for each test, and move comments into descriptions.

Gyuyoung:
I think we achieved the goal of this issue in #1573 and #1598. Mark the status to `Complete`.

kainino0x:
Thanks! In that case we can also close it.

PR #1573, opened 2022-06-24
Describe each test of the validation tests in render|compute pass encoder

Gyuyoung:
This PR adds a description to each test of the validation tests written
in texture/in_pass_encoder.spec.ts file.

Issue: #923 

<hr>

### DIFF
--- a/src/webgpu/api/validation/resource_usages/texture/in_pass_encoder.spec.ts
+++ b/src/webgpu/api/validation/resource_usages/texture/in_pass_encoder.spec.ts
@@ -2,44 +2,6 @@ export const description = `
 Texture Usages Validation Tests in Render Pass and Compute Pass.
 
 TODO: update for new binding structure.
-
-TODO: description per test
-
-Test Coverage:
-  - For each combination of two texture usages:
-    - For various subresource ranges (different mip levels or array layers) that overlap a given
-      subresources or not for color formats:
-      - For various places that resources are used, for example, used in bundle or used in render
-        pass directly.
-        - Check that an error is generated when read-write or write-write usages are binding to the
-          same texture subresource. Otherwise, no error should be generated. One exception is race
-          condition upon two writeonly-storage-texture usages, which is valid.
-
-  - For each combination of two texture usages:
-    - For various aspects (all, depth-only, stencil-only) that overlap a given subresources or not
-      for depth/stencil formats:
-      - Check that an error is generated when read-write or write-write usages are binding to the
-        same aspect. Otherwise, no error should be generated.
-
-  - Test combinations of two shader stages:
-    - Texture usages in bindings with invisible shader stages should be validated. Invisible shader
-      stages include shader stage with visibility none, compute shader stage in render pass, and
-      vertex/fragment shader stage in compute pass.
-
-  - Tests replaced bindings:
-    - Texture usages via bindings replaced by another setBindGroup() upon the same bindGroup index
-      in render pass should be validated. However, replaced bindings should not be validated in
-      compute pass.
-
-  - Test texture usages in bundle:
-    - Texture usages in bundle should be validated if that bundle is executed in the current scope.
-
-  - Test texture usages with unused bindings:
-    - Texture usages should be validated even its bindings is not used in pipeline.
-
-  - Test texture usages validation scope:
-    - Texture usages should be validated per each render pass. And they should be validated per each
-      dispatch call in compute.
 `;
 
 import { makeTestGroup } from '../../../../../common/framework/test_group.js';
@@ -253,8 +215,14 @@ const BASE_LAYER = 1;
 const TOTAL_LAYERS = 6;
 const SLICE_COUNT = 2;
 
-// For all tests below, we test compute pass if 'compute' is true, and test render pass otherwise.
 g.test('subresources_and_binding_types_combination_for_color')
+  .desc(
+    `
+    Test the resource usage rules by using two views of the same GPUTexture in a usage scope. Tests
+    various combinations of {sampled, storage, render target} usages, mip-level ranges, and
+    array-layer ranges, in {compute pass, render pass, render pass via bundle}.
+  `
+  )
   .params(u =>
     u
       .combine('compute', [false, true])
@@ -316,7 +284,8 @@ g.test('subresources_and_binding_types_combination_for_color')
           _resourceSuccess: true,
         },
 
-        // The second texture usage contains the whole mip chain where the first texture usage is using.
+        // The second texture usage contains the whole mip chain where the first texture usage is
+        // using.
         {
           levelCount0: 1,
           layerCount0: 1,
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/resource_usages/texture/in_pass_encoder.spec.ts
```diff
@@ -255,6 +253,13 @@ const SLICE_COUNT = 2;
 
 // For all tests below, we test compute pass if 'compute' is true, and test render pass otherwise.
```
kainino0x, line 254: This explanation should be moved into the test descriptions instead of being a code comment. Actually, it's almost self-documenting, so with tweaks to the test descriptions, I think this can just be deleted.

src/webgpu/api/validation/resource_usages/texture/in_pass_encoder.spec.ts
```diff
@@ -255,6 +253,13 @@ const SLICE_COUNT = 2;
...
 g.test('subresources_and_binding_types_combination_for_color')
+  .desc(
+    `
+    Test that the commands recording of a given pass that has a bindGroup in pass directly or in
+    bundle succeeds if two texture are binding to various subresource ranges (different mip levels
+    or array layers) that overlap a given subresource or not for color formats.
```
kainino0x, line 260: Describing parameterized tests can result in a bit of a "word soup" that's hard to read. I think right now most of these new descriptions are too difficult to read. A reader should be able to "easily" understand the coverage provided by the test.

One way is how it's written at the top of this file: with a pseudocode `for`-loop. It would probably be fine to just move those down here directly, as long as they're still accurate.

---

Below, I'll talk a bit about how we can write these well. It's not necessary for you to actually rewrite all of the descriptions in this style, just suggestions for your continued work.

In some cases, the for-loop (or cartesian product) style writing can be overly precise, and in danger of not matching the code exactly. So it's more useful to summarize a bit in descriptions. I like to grammatically separate the explanation of the parameterization from the high-level overview. Sometimes, we used this bash-style {x,y,z} syntax which I think makes it easier to read, mentioned at the very bottom here: https://github.com/gpuweb/cts/blob/main/docs/intro/plans.md#conventions-used-in-test-plans

An example for this test would be:

> Test the resource usage rules by using two views of the same GPUTexture in a usage scope. Tests various combinations of {sampled, storage, render target} usages, mip-level ranges, and array-layer ranges, in {compute pass, render pass, render pass via bundle}.

kainino0x (changes requested):

src/webgpu/api/validation/resource_usages/texture/in_pass_encoder.spec.ts
Gyuyoung: ok, removed.

src/webgpu/api/validation/resource_usages/texture/in_pass_encoder.spec.ts
Gyuyoung: Sorry for the unclear descriptions. I'm not still good at writing the description well. So, I've updated the descriptions again to make readers understand more easily. But, additional updates may be required.

src/webgpu/api/validation/resource_usages/texture/in_pass_encoder.spec.ts
```diff
@@ -1030,6 +1032,14 @@ g.test('bindings_in_bundle')
   });
 
 g.test('unused_bindings_in_pipeline')
+  .desc(
+    `
+    Test the texture usages with unused bindings in the pipeline.
+      - Test the recording succeeds in the compute pass even though it's bindings is not used in the
+        pipeline.
+      - Test the recording fails in the render pass.
```
kainino0x, line 1040:
```suggestion
    Test that for compute pipelines with 'auto' layout, only bindings used by the pipeline count
    toward the usage scope. For render passes, test the pipeline doesn't matter because only the
    calls to setBindGroup count toward the usage scope.
```

src/webgpu/api/validation/resource_usages/texture/in_pass_encoder.spec.ts
```diff
@@ -1150,6 +1160,12 @@ g.test('unused_bindings_in_pipeline')
   });
 
 g.test('validation_scope,no_draw_or_dispatch')
+  .desc(
+    `
+    Test the texture usages without the draw or dispatch on the same pass that has two bind groups.
+      - In compute pass, no validation error should be generated.
```
kainino0x, line 1166:
```suggestion
    Test usage scope validation with two conflicting bind groups but no draw/dispatch call.
      - In compute pass, no validation error should be generated.
```

src/webgpu/api/validation/resource_usages/texture/in_pass_encoder.spec.ts
```diff
@@ -1168,6 +1184,12 @@ g.test('validation_scope,no_draw_or_dispatch')
   });
 
 g.test('validation_scope,same_draw_or_dispatch')
+  .desc(
+    `
+    Test the texture usages with the same draw or dispatch on the same pass that has two bind
+    groups. A validation error should be generated.
```
kainino0x, line 1190:
```suggestion
    Test usage scope validation with two conflicting bind groups both used in the same draw/dispatch.
```

src/webgpu/api/validation/resource_usages/texture/in_pass_encoder.spec.ts
```diff
@@ -1185,6 +1207,13 @@ g.test('validation_scope,same_draw_or_dispatch')
   });
 
 g.test('validation_scope,different_draws_or_dispatches')
+  .desc(
+    `
+    Test the texture usages with the different draw or dispatch on the same pass that has two bind
+    groups. A validation error should be generated even though the pass draws or dispatches per each
+    bindGroup.
```
kainino0x, line 1214: Upon review of this test, I think it's broken. I will open a change after this lands to fix some issues here.

```suggestion
    Test usage scope validation with two conflicting bind groups used in two different draw/dispatch calls.

    TODO: This test is failing validation due to a different thing that it intends to test.
```

src/webgpu/api/validation/resource_usages/texture/in_pass_encoder.spec.ts
```diff
@@ -1206,6 +1235,12 @@ g.test('validation_scope,different_draws_or_dispatches')
   });
 
 g.test('validation_scope,different_passes')
+  .desc(
+    `
+    Test the texture usages with different passes that have their own bind group. No validation
+    error should be generated.
```
kainino0x, line 1241:
```suggestion
    Test usage scope validation with two conflicting bind groups used in two entirely different passes.
    No validation error should be generated.
```

kainino0x (approved):
There is some more re-review work to do here later. I reviewed most of the tests but marked the ones I couldn't fully review in new TODOs.

LGTM with my suggestions landed (you can apply them from github using the "add suggestion to batch" feature)

Issue #1162, opened 2022-04-13
textureContentIsOKByT2B is slow


Kangz:
The `textureContentIsOKByT2B` helper function is very slow because it calls a lot of helper functions left and right for every pixel, both to generate the expected data (with the `TexelView`) and to compare data with ulp diffs.

This is causing the following test `webgpu:web_platform,copyToTexture,canvas:copy_contents_from_2d_context_canvas:canvasType="offscreen";dstColorFormat="rgba16float";dstPremultiplied=false;srcDoFlipYDuringCopy=true` to run in 2.5s seconds on my machine when it compares only ~20 textures (a couple are medium-sized at 256*256 pixels).

We should invest in having helpers to efficiently generate expected data and check diffs with ULPs. We could even port the ULP checks to compute shaders or other fancy stuff. The creation of Texel views could also be done by having large arrays of data with metadata instead of chains of callbacks for everything. Looking at performance profiles, it doesn't seem like callbacks get inlined very well.

FYI @shaoboyan @kainino0x

PR #1583, opened 2022-06-28
Optimized several expensive methods in texture tests

toji:
In my local testing I found that running tests under
`webgpu:web_platform,copyToTexture,ImageBitmap:*` was approx.
**62% faster** with these optimizations applied. Your milage may
vary. I did not do a comprehensive evaluation of tests outside that
group to determine the performance impact on them.

Should at least partially address #1162, though there's still more
work to be done.

Optimization done primarily by running different subsets of
`webgpu:web_platform,copyToTexture,ImageBitmap:*`, profiling with
Chrome dev tools, and examining the methods that consistently
showed up as the most expensive.

Almost all of the optimizations ended up being finding ways to
avoid repeated allocations of `TypedArrays`, `ArrayBuffers`, or
`DataViews`, as those tend to be reasonably expensive and these code
paths were allocating lots and lots of small ones.

Issue: #1162

<hr>

### DIFF
--- a/src/common/util/util.ts
+++ b/src/common/util/util.ts
@@ -240,12 +240,22 @@ export const kTypedArrayBufferViewConstructors = Object.values(kTypedArrayBuffer
 function subarrayAsU8(
   buf: ArrayBuffer | TypedArrayBufferView,
   { start = 0, length }: { start?: number; length?: number }
-): Uint8Array {
+): Uint8Array | Uint8ClampedArray {
   if (buf instanceof ArrayBuffer) {
     return new Uint8Array(buf, start, length);
+  } else if (buf instanceof Uint8Array || buf instanceof Uint8ClampedArray) {
+    // Don't wrap in new views if we don't need to.
+    if (start === 0 && (length === undefined || length === buf.byteLength)) {
+      return buf;
+    }
+    return buf.subarray(start, length !== undefined ? start + length : undefined);
   } else {
-    const sub = buf.subarray(start, length !== undefined ? start + length : undefined);
-    return new Uint8Array(sub.buffer, sub.byteOffset, sub.byteLength);
+    const byteOffset = buf.byteOffset + start * buf.BYTES_PER_ELEMENT;
+    const byteLength =
+      length !== undefined
+        ? byteOffset + length * buf.BYTES_PER_ELEMENT
+        : buf.byteLength - (byteOffset - buf.byteOffset);
+    return new Uint8Array(buf.buffer, byteOffset, byteLength);
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/webgpu/util/texture/texel_data.ts
```diff
@@ -122,6 +122,21 @@ export function makeClampToRange(format: EncodableTextureFormat): ComponentMapFn
   return applyEach(x => clamp(x, repr.numericRange!), repr.componentOrder);
 }
 
+const smallComponentDataViews = new Map();
```
austinEng, line 125: perhaps have a MAINTENANCE_TODO about exposing this map somehow to the test fixture so we can GC its contents at the end of a test group.

Then, we could potentially consider storing > 32-byte buffers in it as well
(not sure how helpful that would be)

src/common/util/util.ts
```diff
@@ -240,12 +240,22 @@ export const kTypedArrayBufferViewConstructors = Object.values(kTypedArrayBuffer
...
-): Uint8Array {
+): Uint8Array | Uint8ClampedArray {
   if (buf instanceof ArrayBuffer) {
     return new Uint8Array(buf, start, length);
+  } else if (buf instanceof Uint8Array || buf instanceof Uint8ClampedArray) {
+    // Don't wrap in new views if we don't need to.
+    if (start === 0 && (length === undefined || length === buf.byteLength)) {
+      return buf;
+    }
+    return buf.subarray(start, length !== undefined ? start + length : undefined);
```
austinEng, line 251: nit: is there value in using `.subarray` vs `new Uint8Array` here?
IIUC, they will both give you a new view, without a new buffer allocation.
So, `new Uint8Array` might be *slightly* more concise since we can put the length arg in as-is without the ternary

austinEng (approved):

src/common/util/util.ts
kainino0x: IIUC `new Uint8Array` can be used to make a view onto an ArrayBuffer, but not onto another `Uint8Array` without some complexity.

src/webgpu/util/texture/texel_data.ts
toji: Added a comment.

src/common/util/util.ts
toji: If we use `new Uint8Array()` and don't want it to do a copy, we have to pass in the source array's buffer. The buffer may have a longer length than the `Uint8Array` view, so we'd have to guard against an undefined length extending past the bounds of the original view.

Turns out this is what the `else` statement does in this case, so I just removed the `.subarray()` path here and had it fall though to that common case.

toji:
Given Austin's previous approval and lack of further feedback, I'm merging this now. I'll keep a careful eye on the next CTS->Dawn roll just in case any new failures crop up that might be a result of this.

kainino0x:
Sorry, I think I saw that Austin had approved it already and ignored the review request accidentally. :)

PR #1601, opened 2022-07-01
Guard against writeTimestamp() calls without the proper feature

toji:
writeTimestamp is specced to throw immediately if called when the
'timestamp-query' feature is not enabled. This change ensures that
any tests that call writeTimestamp() are skipped unless the device
enables 'timestamp-query'.

These tests started failing in Chrome once we began throwing
exceptions when the feature wasn't enabled, as the spec requires.

<hr>

### DIFF
--- a/src/webgpu/api/validation/encoding/queries/general.spec.ts
+++ b/src/webgpu/api/validation/encoding/queries/general.spec.ts
@@ -89,8 +89,15 @@ Tests that write timestamp to all types of query set on all possible encoders:
       .expand('queryIndex', p => (p.type === 'timestamp' ? [0, 2] : [0]))
   )
   .beforeAllSubcases(t => {
+    const { type } = t.params;
+
     // writeTimestamp is only available for devices that enable the 'timestamp-query' feature.
-    t.selectDeviceForQueryTypeOrSkipTestCase('timestamp');
+    const queryTypes: GPUQueryType[] = ['timestamp'];
+    if (type !== 'timestamp') {
+      queryTypes.push(type);
+    }
+
+    t.selectDeviceForQueryTypeOrSkipTestCase(queryTypes);
   })
   .fn(async t => {
     const { type, queryIndex } = t.params;
--- a/src/webgpu/api/validation/encoding/queries/general.spec.ts
+++ b/src/webgpu/api/validation/encoding/queries/general.spec.ts
@@ -89,10 +89,8 @@ Tests that write timestamp to all types of query set on all possible encoders:
       .expand('queryIndex', p => (p.type === 'timestamp' ? [0, 2] : [0]))
   )
   .beforeAllSubcases(t => {
-    const { type } = t.params;
-    if (type) {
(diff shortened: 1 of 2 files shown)


src/webgpu/api/validation/encoding/queries/general.spec.ts
```diff
@@ -89,10 +89,8 @@ Tests that write timestamp to all types of query set on all possible encoders:
       .expand('queryIndex', p => (p.type === 'timestamp' ? [0, 2] : [0]))
   )
   .beforeAllSubcases(t => {
-    const { type } = t.params;
-    if (type) {
-      t.selectDeviceForQueryTypeOrSkipTestCase(type);
-    }
+    // writeTimestamp is only available for devices that enable the 'timestamp-query' feature.
+    t.selectDeviceForQueryTypeOrSkipTestCase('timestamp');
```
austinEng, line 93: why not `t.params.type`?

src/webgpu/api/validation/state/device_lost/destroy.spec.ts
```diff
@@ -673,8 +673,8 @@ Tests encoding and finishing a writeTimestamp command on destroyed device.
       .combine('awaitLost', [true, false])
   )
   .beforeAllSubcases(t => {
-    const { type } = t.params;
-    t.selectDeviceForQueryTypeOrSkipTestCase(type);
+    // writeTimestamp is only available for devices that enable the 'timestamp-query' feature.
+    t.selectDeviceForQueryTypeOrSkipTestCase('timestamp');
```
austinEng, line 677: ditto

src/webgpu/api/validation/encoding/queries/general.spec.ts
austinEng: I'm silly - that's what it already was; why had it not properly skipped?

toji: Because then `writeTimestamp()` will be called and throw an exception when the query type is `occlusion` because `timestamp-query` wasn't enabled. We can wrap the `writeTimestamp()` in a `t.shouldThrow()` but then you've removed the entire thing you're supposed to be testing and replaced it with a test for "Does this function throw if the feature isn't enabled?" (Which is already being tested in `webgpu:api,validation,capability_checks,features,query_types:writeTimestamp:*`)

austinEng: OH it's because it is *always* doing writeTimestamp, with potentially non-timestamp queries.

Sounds like we need to skip if either t.params.type OR 'timestamp' is unsupported?

toji: Ah, good point! We might have future query types that require their own independent features. I'll make that adjustment.

toji: Okay, updated. Should be more future-proof now. Thanks!

austinEng (approved):
LGTM

PR #1608, opened 2022-07-04
Minor clean up in life_of.md

takahirox:
I think `"Status: New (not in repo)"` would be more straightforward than `"Status" "New (not in repo)"`? And `is` can be dropped, if I'm right?

### DIFF
--- a/docs/intro/life_of.md
+++ b/docs/intro/life_of.md
@@ -10,7 +10,7 @@ Filing new draft issues in the CTS project tracker is very lightweight.
 Anyone with access should do this eagerly, to ensure no testing ideas are forgotten.
 (And if you don't have access, just file a regular issue.)
 
-1. Enter a draft issue with "Status" "New (not in repo)" is in the
+1. Enter a draft issue with "Status: New (not in repo)" in the
     [CTS project tracker](https://github.com/orgs/gpuweb/projects/3) with any available info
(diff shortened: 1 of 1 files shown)


docs/intro/life_of.md
```diff
@@ -10,7 +10,7 @@ Filing new draft issues in the CTS project tracker is very lightweight.
 Anyone with access should do this eagerly, to ensure no testing ideas are forgotten.
 (And if you don't have access, just file a regular issue.)
 
-1. Enter a draft issue with "Status" "New (not in repo)" is in the
+1. Enter a draft issue with "Status: New (not in repo)" in the
     [CTS project tracker](https://github.com/orgs/gpuweb/projects/3) with any available info
```
kainino0x, line 14: Since the string "Status: New (not in repo)" doesn't show up anywhere else I wouldn't quote it all together. What do you think of:

```suggestion
1. Enter a [draft issue](https://github.com/orgs/gpuweb/projects/3), with the Status
    set to "New (not in repo)", and any available info included in the issue description
```

takahirox: Looking nice, thanks for the suggestion

takahirox:
Good to go with this change?

kainino0x (approved):
Yes, sorry I forgot about this!

Issue #1618, opened 2022-07-05
val: render pass descriptor


Gyuyoung:

kainino0x:
Closing because it was marked complete, which I think is accurate.

PR #1630, opened 2022-07-06
Implement val:render_pass_descriptor:timestamp_writes_location

Gyuyoung:
No two entries in timestampWrites have the same location. So,
this PR adds a test to ensure that two timestampWrite don't
have the same location.

Issue: #1618

<hr>

### DIFF
--- a/src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts
+++ b/src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts
@@ -744,7 +744,7 @@ g.test('timestamp_writes_location')
       location: locationB,
     };
 
-    const isValid = locationA === locationB ? false : true;
+    const isValid = locationA !== locationB;
 
     const colorTexture = t.createTexture({ format: 'rgba8unorm' });
     const descriptor = {
--- a/src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts
+++ b/src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts
@@ -713,3 +713,44 @@ g.test('multisample_render_target_formats_support_resolve')
       colorAttachments: [colorAttachment],
     });
   });
+
+g.test('timestamp_writes_location')
+  .desc('Test that entries in timestampWrites do not have the same location.')
+  .params(u =>
+    u //
+      .combine('locationA', ['beginning', 'end'] as const)
+      .combine('locationB', ['beginning', 'end'] as const)
+  )
+  .beforeAllSubcases(t => {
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts
```diff
@@ -713,3 +713,35 @@ g.test('multisample_render_target_formats_support_resolve')
...
+      location: 'beginning' as GPURenderPassTimestampLocation,
+    };
+
+    const colorTexture = t.createTexture({ format: 'rgba8unorm' });
+    const descriptor = {
+      colorAttachments: [t.getColorAttachment(colorTexture)],
+      timestampWrites: [timestampWriteA, timestampWriteB],
+    };
+
+    t.tryRenderPass(false, descriptor);
```
kainino0x, line 746: Right now this always fails so it's hard to tell that the error is due to this specific error. The test should be parameterized so that it has at least one "control case" (where the validation passes), and it should also test what happens when both are `'end'`. So I'd test all 4 combinations:
```ts
  .paramsSubcasesOnly(u =>
    u //
      .combine('locationA', ['beginning', 'end'] as const)
      .combine('locationB', ['beginning', 'end'] as const)
  )
```
updating the rest of the test as necessary.

kainino0x (changes requested):

src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts
Gyuyoung: The test is passing even though the locations are the same. I guess that Chromium doesn't generate a validation error.  

I add the combinations to the test. BTW, can we use .param instead of  .paramsSubcasesOnly in order to know what combinations fail more clear.

kainino0x: Yup, that would be fine!

src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts
```diff
@@ -713,3 +713,44 @@ g.test('multisample_render_target_formats_support_resolve')
...
+      location: locationA,
+    };
+
+    const timestampWriteB = {
+      querySet,
+      queryIndex: 1,
+      location: locationB,
+    };
+
+    const isValid = locationA === locationB ? false : true;
```
kainino0x, line 747: µNit:

```suggestion
    const isValid = locationA !== locationB;
```

kainino0x (approved):

PR #1653, opened 2022-07-12
Implement val:render_pass_descriptor:timestamp_writes_query_set_type

Gyuyoung:
The queryType of each timestampWrite in timestampWrites should be
`timestamp`. So, this PR adds a test to check if a validation error
is generated if all query set types aren't `timestamp`.

Issue: #1618

<hr>

### DIFF
--- a/src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts
+++ b/src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts
@@ -848,6 +848,47 @@ g.test('resolveTarget,format_supports_resolve')
     });
   });
 
+g.test('timestampWrites,query_set_type')
+  .desc(
+    `
+  Test that all entries of the timestampWrites must have type 'timestamp'. If all query types are
+  not 'timestamp', a validation error should be generated.
+  `
+  )
+  .params(u =>
+    u //
+      .combine('queryTypeA', kQueryTypes)
+      .combine('queryTypeB', kQueryTypes)
+  )
+  .beforeAllSubcases(t => {
+    t.selectDeviceOrSkipTestCase(['timestamp-query']);
+  })
+  .fn(async t => {
+    const { queryTypeA, queryTypeB } = t.params;
+
+    const timestampWriteA = {
+      querySet: t.device.createQuerySet({ type: queryTypeA, count: 1 }),
+      queryIndex: 0,
+      location: 'beginning' as const,
+    };
+
+    const timestampWriteB = {
+      querySet: t.device.createQuerySet({ type: queryTypeB, count: 1 }),
+      queryIndex: 0,
+      location: 'end' as const,
+    };
+
+    const isValid = queryTypeA === 'timestamp' && queryTypeB === 'timestamp';
+
+    const colorTexture = t.createTexture();
+    const descriptor = {
+      colorAttachments: [t.getColorAttachment(colorTexture)],
+      timestampWrites: [timestampWriteA, timestampWriteB],
+    };
+
+    t.tryRenderPass(isValid, descriptor);
+  });
+
 g.test('timestamp_writes_location')
   .desc('Test that entries in timestampWrites do not have the same location.')
   .params(u =>


src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts
```diff
@@ -848,6 +848,47 @@ g.test('resolveTarget,format_supports_resolve')
...
+    const { queryTypeA, queryTypeB } = t.params;
+
+    const timestampWriteA = {
+      querySet: t.device.createQuerySet({ type: queryTypeA, count: 1 }),
+      queryIndex: 0,
+      location: 'beginning' as GPURenderPassTimestampLocation,
+    };
+
+    const timestampWriteB = {
+      querySet: t.device.createQuerySet({ type: queryTypeA, count: 1 }),
```
kainino0x, line 876: Looks like a typo here. Did the test pass with this typo?

```suggestion
      querySet: t.device.createQuerySet({ type: queryTypeB, count: 1 }),
```

src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts
```diff
@@ -848,6 +848,47 @@ g.test('resolveTarget,format_supports_resolve')
...
+  .beforeAllSubcases(t => {
+    t.selectDeviceOrSkipTestCase(['timestamp-query']);
+  })
+  .fn(async t => {
+    const { queryTypeA, queryTypeB } = t.params;
+
+    const timestampWriteA = {
+      querySet: t.device.createQuerySet({ type: queryTypeA, count: 1 }),
+      queryIndex: 0,
+      location: 'beginning' as GPURenderPassTimestampLocation,
```
kainino0x, line 872: µNit: use `as const` instead of `as GPURenderPassTimestampLocation`

src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts
```diff
@@ -848,6 +848,47 @@ g.test('resolveTarget,format_supports_resolve')
...
+    };
+
+    const timestampWriteB = {
+      querySet: t.device.createQuerySet({ type: queryTypeA, count: 1 }),
+      queryIndex: 0,
+      location: 'beginning' as GPURenderPassTimestampLocation,
```
kainino0x, line 878: The locations need to be different, or it's invalid (as tested by the test just below this one): https://gpuweb.github.io/gpuweb/#abstract-opdef-gpucomputepassdescriptor-valid-usage

(I don't know if that test passes in chromium or not)

src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts
Gyuyoung: Oops, sorry for the typo. Fixed.

src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts
Gyuyoung: Ah, right. But, Chromium is passing the test even if the locations are equal. Fixed.

src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts
kainino0x: LGTM but if the test passed with this typo we should understand why - it seems like a bug either in this test or in Chromium, I think.

kainino0x (approved):

src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts
Gyuyoung: yes, the test has been passed even though the query type is `occlusion`. It looks like Chromium doesn't generate a validation error when all types are not `timestamp`.

kainino0x: Oh, is this just because the timestamp-query feature is not available? If you're running in the browser with `npm start` then you should see those cases as skipped.

I'm not sure whether we support it but it's possible that it's supported on some platforms, possibly requiring `chrome://flags/#enable-webgpu-developer-features` to be enabled.

kainino0x: If the feature is available but the check is actually missing, that's quite bad because I don't know what is going to happen in the backends if we allow that through. In that case please file a bug on https://bugs.chromium.org/p/dawn/issues/entry

Gyuyoung: > Oh, is this just because the timestamp-query feature is not available? If you're running in the browser with `npm start` then you should see those cases as skipped.

When I run the tests in the browser with `npm start`, the tests were skipped. It seems to me that the timestamp-query feature is not available.

`SKIP: GPUDeviceDescriptor not supported: {"requiredFeatures":["timestamp-query"],"requiredLimits":{},"defaultQueue":{}}
One or more features are not supported`


> I'm not sure whether we support it but it's possible that it's supported on some platforms, possibly requiring `chrome://flags/#enable-webgpu-developer-features` to be enabled.

Although I enabled #enable-webgpu-developer-features in the browser, the tests were still skipped. I'm not sure if the feature could support `timestamp-query`. Do you know how to check if the feature can support `timestamp-query` in the browser?

kainino0x: I think timestamp-query support is just not done yet. I saw a Chromium CL recently for implementing it. I don't know if it's complete or whether it's in Canary yet.

You can check feature support here https://webgpureport.org/
But I suspect you'll find that timestamp-query is not listed.

Gyuyoung: Thank you for letting me know the URL. yes, it's likely timestamp-query is not supported yet.

PR #1655, opened 2022-07-12
Update f32 tests to handle returning undefined for infinity

### LINKED ISSUE #1642: Handle infinite results of operations being potentially undefined
When I originally prototyped/started implementing the new FP testing framework, the behaviour of infinite values was that back ends that did not support them would clamp to max/min f32, effectively treating them as very large numbers for most purposes.

Since then, the following text has been added:
```
In such an implementation, when an evaluation would produce an infinity or a NaN, an undefined value of the target type is produced instead.
```
This effectively means that whenever an infinite value is returned (or NaN) from an operation the acceptance interval becomes the infinite interval and that will propegate up. I need to revisit how this is handled in the framework and operation implementations.

zoddicus:
The WGSL spec has been clarified to allow implementations to return
undefined values when they do not implement handling infinity. For CTS
tests this means that anytime infinity is encountered it may
potentially turn into an undefined value. In general terms this means
that acceptance intervals for certains cases need to loosened, and a
bunch of corner cases related to clamping infinities can be removed.

This is a big monolithic change, because the alternate would be
creating a parallel framework infrastructure with the new behaviour,
and rewriting tests piecemeal to use the new framework. Which I am
currently already in the process of doing wrt migration to the
interval framework. Having 3 different testing frameworks live at the
same time seems like a recipe for disaster.

Highlights of this change:
- F32Interval.infinity() has been renamed F32Interval.undefined()
- Infinity clamping behaviour in the F32Interval constructor is gone
- *ToIntervalOp.impl is no longer responsible for checking if inputs
  are valid
- *ToIntervalOp.domain? property has been added to allow for operation
  definitions to specify additional constraints on the inputs, run*Op
  actually tests these limits
- run*Op is now responsible for checking that inputs/outputs of
  operations are finite, otherwise returning undefined
- Usages of Number.POSITIVE/NEGATIVE_INFINITY on the input side of
  unit tests related to generating intervals have been change to
  kValue.f32.infinity.positive/negative, for consistency and since
  Number.* is a f64 value

Fixes #1642

<hr>

### DIFF
--- a/src/unittests/f32_interval.spec.ts
+++ b/src/unittests/f32_interval.spec.ts
@@ -3,7 +3,7 @@ F32Interval unit tests.
 `;
 
 import { makeTestGroup } from '../common/framework/test_group.js';
-import { assert, objectEquals } from '../common/util/util.js';
+import { objectEquals } from '../common/util/util.js';
 import { kValue } from '../webgpu/util/constants.js';
 import {
   absInterval,
@@ -40,15 +40,8 @@ import { UnitTest } from './unit_test.js';
 
 export const g = makeTestGroup(UnitTest);
 
-/** Convert a pair of numbers in an array to a F32Interval
- *
- * Used for fluently specifying test params as `[a, b]` instead of
- * `new F32Interval(a, b)`
- */
-function arrayToInterval(bounds: [number, number]): F32Interval {
-  const [begin, end] = bounds;
-  return new F32Interval(begin, end);
-}
+/** Bounds indicating an expectation of an interval of all possible values */
+const kAny: [number, number] = [Number.NEGATIVE_INFINITY, Number.POSITIVE_INFINITY];
 
 /** @returns a number N * ULP greater than the provided number */
 function plusNULP(x: number, n: number): number {
@@ -100,7 +93,7 @@ g.test('constructor')
       // Infinities
       { input: [0, kValue.f32.infinity.positive], expected: [0, Number.POSITIVE_INFINITY]},
       { input: [kValue.f32.infinity.negative, 0], expected: [Number.NEGATIVE_INFINITY, 0]},
-      { input: [kValue.f32.infinity.negative, kValue.f32.infinity.positive], expected: [Number.NEGATIVE_INFINITY, Number.POSITIVE_INFINITY]},
+      { input: [kValue.f32.infinity.negative, kValue.f32.infinity.positive], expected: kAny},
     ]
   )
   .fn(t => {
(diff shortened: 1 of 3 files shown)


src/unittests/f32_interval.spec.ts
```diff
@@ -379,8 +379,8 @@ g.test('correctlyRoundedInterval')
     // prettier-ignore
     [
       // Edge Cases
-      { value: kValue.f32.infinity.positive, expected: [kValue.f32.positive.max, Number.POSITIVE_INFINITY] },
-      { value: kValue.f32.infinity.negative, expected: [Number.NEGATIVE_INFINITY, kValue.f32.negative.min] },
+      { value: kValue.f32.infinity.positive, expected: [Number.NEGATIVE_INFINITY, Number.POSITIVE_INFINITY] },
```
ben-clayton, line 382: As `[Number.NEGATIVE_INFINITY, Number.POSITIVE_INFINITY]` is used so heavily now, maybe it would make sense to pre-declare an `Any` interval which can be used instead.

src/unittests/f32_interval.spec.ts
```diff
@@ -913,7 +913,7 @@ g.test('inverseSqrtInterval')
       { input: 1, expected: [1, 1] },
       { input: 100, expected: [minusOneULP(hexToF32(0x3dcccccd)), hexToF32(0x3dcccccd)] },  // ~0.1
       { input: kValue.f32.positive.max, expected: [hexToF32(0x1f800000), plusNULP(hexToF32(0x1f800000), 2)] },  // ~5.421...e-20, i.e. 1/√max f32
-      { input: Number.POSITIVE_INFINITY, expected: [0, plusNULP(hexToF32(0x1f800000), 2)] },
+      { input: kValue.f32.infinity.positive, expected: arrayToInterval([Number.NEGATIVE_INFINITY, Number.POSITIVE_INFINITY]) },
```
ben-clayton, line 916: As far as I can tell, `arrayToInterval` now only exists to prevent the `[Number.NEGATIVE_INFINITY, Number.POSITIVE_INFINITY]` interval being expanded by error thresholds. If you were to adopt the `Any` suggestion, then the test body could check whether the interval is equal to the `Any` array (object equality would work), and if so, skip the error expansion logic.

If you did all this then `PointToIntervalCase.expected` could just become `expected: [number, number]`, and you could drop the `arrayToInterval` smell in the test cases.

WDYT?

src/webgpu/util/f32_interval.ts
```diff
@@ -85,15 +73,15 @@ export class F32Interval {
...
 
-  /** @returns a singleton for the infinite interval
+  /** @returns a singleton for the undefined interval
    * This interval is used in situations where accuracy is not defined, so any
    * result is valid.
    */
-  public static infinite(): F32Interval {
-    if (this._infinite === undefined) {
-      this._infinite = new F32Interval(Number.NEGATIVE_INFINITY, Number.POSITIVE_INFINITY);
+  public static undefined(): F32Interval {
```
ben-clayton, line 80: A little bit troubling that this may collide with [Javascript's `undefined` property](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined).
This is the sort of thing that I'd expect to go horribly and subtly wrong with different JS engines, or emission with a transpiler like TypeScript.
Suggest renaming (`any()`?)

src/webgpu/util/f32_interval.ts
```diff
@@ -115,25 +103,28 @@ export interface PointToInterval {
...
+   * This is only implemented for subclasses that meet all of the following
    * criteria:
    *   a) non-monotonic
    *   b) used in inherited accuracy calculations
    *   c) need to take in an interval for b)
    *      i.e. fooInterval takes in x: number | F32Interval, not x: number
    */
   extrema?: (x: F32Interval) => F32Interval;
+
+  /** @returns an additional interval that inputs need to be within. If not defined, they are only tested for being finite. */
```
ben-clayton, line 126: Comment is a little confusing given the interface.
What happens if the input is not in the `domain`? Who's doing the `tested` ?

src/webgpu/util/f32_interval.ts
```diff
@@ -115,25 +103,28 @@ export interface PointToInterval {
...
-  impl: (x: number) => F32Interval;
+  impl: PointToInterval;
 
   /**
    * Calculates where in the domain defined by x the min/max extrema of impl
    * occur and returns a span of those points to be used as the domain instead.
    *
-   * If not defined the ends of the existing domain are assumed to be the
-   * extrema.
+   * If the result is not defined the ends of the existing domain are assumed to
```
ben-clayton, line 114: Comment is confusing.
What `result`?
Also: add comma after `defined`.

src/webgpu/util/f32_interval.ts
```diff
@@ -181,13 +175,17 @@ interface TernaryToIntervalOp {
   // All current ternary operations that are used in inheritance (clamp*) are
   // monotonic, so extrema calculation isn't needed. Re-using the Op interface
   // pattern for symmetry with the other operations
+
+  // All current ternary operations have a domain of the full f32 space, so a domain? override is not currently
```
ben-clayton, line 179: nit: Not sure this is technically an `override`. Suggest:

`// All current ternary operations have a domain of the full f32 space, so a domain is not currently provided.`

src/webgpu/util/f32_interval.ts
ben-clayton, line 126: Why is this external to the implementation? Couldn't it be part of the implementation of `impl` ? Seems fairly easy to create a helper to return a `PointToIntervalOp` that enforces this interval check before calling into the inner `PointToIntervalOp`.

ben-clayton (commented):
Looks good!
A bunch of minor comments and suggestions.

zoddicus (commented):
Apparently actually sending you responses/comments helps people see them...

src/unittests/f32_interval.spec.ts
ben-clayton: Looks great, thank you!

src/unittests/f32_interval.spec.ts
ben-clayton: Nice. I like it.

src/webgpu/util/f32_interval.ts
ben-clayton: Better thank you. Still not sure you need this exposed though :)

src/webgpu/util/f32_interval.ts
ben-clayton: Much nicer, thanks.

src/webgpu/util/f32_interval.ts
zoddicus: Thanks for the clarification, I was _overthinking this_ :tm:, trying to come up with a solution that rolled in the general must be finite requirement, whereas you were talking about just the specific cases that have tighter requirements.

I will implement this change and update the PR.

zoddicus: Done

zoddicus:
@ben-clayton PTAL, I have implemented the limit wrapper

src/webgpu/util/f32_interval.ts
```diff
@@ -93,6 +93,18 @@ function toInterval(n: number | F32Interval): F32Interval {
   return new F32Interval(n, n);
 }
 
+/** F32Interval of [-π, π] */
+const kNegPiToPiInterval = new F32Interval(
+  kValue.f32.negative.pi.whole,
+  kValue.f32.positive.pi.whole
+);
+
+/** F32Interval of values greater than 0 */
```
ben-clayton, line 96: I'd add: `and less than positive infinity`

src/webgpu/util/f32_interval.ts
```diff
@@ -335,12 +368,6 @@ function runBinaryOp(x: F32Interval, y: F32Interval, op: BinaryToIntervalOp): F3
     [x, y] = op.extrema(x, y);
   }
 
-  if (op.domain !== undefined) {
```
zoddicus: Changing this is why that one unit test is failing. Specifically division has an extrema @ 0, which isn't being handled, but the use of intervals at this point was catching the issue. I will fix the PR to add an extrema? to division

ben-clayton: Ack

ben-clayton:
Looking much cleaner, thank you!

src/webgpu/util/f32_interval.ts
zoddicus: Done

src/webgpu/util/f32_interval.ts
ben-clayton: 👍🏼

Issue #1561, opened 2022-06-21
canvas read as an image source alpha mode behavior update


shrekshao:
https://github.com/gpuweb/gpuweb/pull/2905

shrekshao:
https://github.com/gpuweb/gpuweb/pull/2905

shrekshao:
https://www.w3.org/TR/webgpu/#GPUCanvasAlphaMode

> (GPUCanvasAlphaMode) selects how the contents of the canvas will be interpreted when read, when [rendered to the screen](https://www.w3.org/TR/webgpu/#abstract-opdef-update-the-rendering-of-the-webgpu-canvas) or [used as an image source](https://www.w3.org/TR/webgpu/#abstract-opdef-get-a-copy-of-the-image-contents-of-a-context) (in drawImage, toDataURL, etc.)

We have reftests for "canvas read when rendered to the screen", we still need reftests for "canvas read when used as an image source."

Operations include drawImage, toDataURL, texImage2D, toBlob etc.

shrekshao:
Now it seems best to expand the `readbackFromWebGPUCanvas.spec.ts` with

- x= src canvas
  - type=
    - canvas element not in dom
    - canvas element in dom
    - OffscreenCanvas from transferControlToOffscreen from canvas not in dom
    - OffscreenCanvas from transferControlToOffscreen from canvas in dom
    - OffscreenCanvas from new OffscreenCanvas
  - alphaMode= {opaque, premultiplied}
- x= dst canvas
  - method=
    - 2d drawImage,
    - toDataURL,
    - toBlob,
    - createImageBitmap,
    - WebGL texImage2D,
    - WebGL texSubImage2D,
    - transferToImageBitmap
    - convertToBlob
    - ... and probably some others.
  - alphaMode= {opaque, premultiplied}

kainino0x:
Finally tested and filed https://github.com/gpuweb/gpuweb/issues/3295

kainino0x:
@shrekshao this issue is still marked "started". Do you remember if it was done or was there still some work to do?

shrekshao:
Seems that the TODOs on `kAllCanvasTypes` are still missing
- canvas element not in dom
- canvas element in dom
- offscreen canvas from transferControlToOffscreen from canvas not in dom
- offscreen canvas from transferControlToOffscreen from canvas in dom
- offscreen canvas from new OffscreenCanvas

kainino0x:
[...]

PR #1638, opened 2022-07-06
Add tests for getCurrentTexture and getPreferredCanvasFormat

toji:
Majority of tests are around whether or not getCurrentTexture
returns a new texture or not given a variety of circumstances.

Also tests the only thing that I think CAN be tested about 
getPreferredCanvasFormat, which is that it returns one of the
two allowed formats.

Currently the multiple_frames test fails in Chrome for onscreen
canvases, since we're not reliably compositing with each rAF
cycle when the canvas isn't visible. Not sure how to fix that
or if the test should be altered to stress that behavior some
other way.

<hr>

**Requirements for PR author:**

- [x] All missing test coverage is tracked with "TODO" or `.unimplemented()`.
- [x] New helpers are `/** documented */` and new helper files are found in `helper_index.txt`.
- [x] Test behaves as expected in a WebGPU implementation. (If not passing, explain above.)

**Requirements for [reviewer sign-off](https://github.com/gpuweb/cts/blob/main/docs/reviews.md):**

- [ ] Tests are properly located in the test tree.
- [ ] [Test descriptions](https://github.com/gpuweb/cts/blob/main/docs/intro/plans.md) allow a reader to "read only the test plans and evaluate coverage completeness", and accurately reflect the test code.
- [ ] Tests provide complete coverage (including validation control cases). **Missing coverage MUST be covered by TODOs.**
- [ ] Helpers and types promote readability and maintainability.

When landing this PR, be sure to make any necessary issue status updates.

PR #1657, opened 2022-07-13
Add reftests: canvas use as image source with opaque/premultiplied alpha mode

shrekshao:
Issue: #1561 

Test results in current chrome canary (Win 105.0.5176.0 (Official Build) canary (64-bit)):

* canvas_use_as_image_source_opaque
image

* canvas_use_as_image_source_opaque-ref
image

Alpha channel clear issue is working at: https://chromium-review.googlesource.com/c/chromium/src/+/3749085

WebGPU to WebGL image copy with incorrect flipY is tracked at: https://bugs.chromium.org/p/chromium/issues/detail?id=1344276

<hr>

### DIFF
--- a/src/webgpu/web_platform/reftests/canvas_use_as_image_source.html.ts
+++ b/src/webgpu/web_platform/reftests/canvas_use_as_image_source.html.ts
@@ -1,0 +1,98 @@
+import { assert, unreachable } from '../../../common/util/util.js';
+import { kTextureFormatInfo } from '../../capability_info.js';
+import { align } from '../../util/math.js';
+
+import { runRefTest } from './gpu_ref_test.js';
+import {
+  useAsImageSource2dDrawImage,
+  useAsImageSourceToDataURL,
+  useAsImageSourceToBlob,
+  useAsImageSourceCreateImageBitmap,
+  useAsImageSourceWebGLTexImage2D,
+  useAsImageSourceWebGLTexSubImage2D,
+} from './ref/canvas_use_as_image_source_utils.html.js';
+
+const alphaValue = 0x66;
+
+export function run(
+  alphaMode: GPUCanvasAlphaMode,
+  src: HTMLCanvasElement,
+  dst_draw_image: HTMLCanvasElement,
+  dst_data_url: HTMLCanvasElement,
+  dst_blob: HTMLCanvasElement,
+  dst_create_image_bitmap: HTMLCanvasElement,
+  dst_tex_image_2d: HTMLCanvasElement,
+  dst_tex_sub_image_2d: HTMLCanvasElement
+) {
+  runRefTest(async t => {
+    const ctx = src.getContext('webgpu');
+    assert(ctx !== null, 'Failed to get WebGPU context from canvas');
+
+    const format = 'bgra8unorm';
+    ctx.configure({
+      device: t.device,
+      format,
+      usage: GPUTextureUsage.COPY_DST,
+      alphaMode,
+    });
+
+    // Write to src canvas
+    const rows = ctx.canvas.height;
+    const bytesPerPixel = kTextureFormatInfo[format].bytesPerBlock;
+    if (bytesPerPixel === undefined) {
+      unreachable();
+    }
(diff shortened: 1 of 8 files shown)


src/webgpu/web_platform/reftests/ref/canvas_use_as_image_source_utils.html.ts
```diff
@@ -0,0 +1,138 @@
+import { assert } from '../../../../common/util/util.js';
```
kainino0x, line 1: filename nit: doesn't need to be `.html.ts` unless it corresponds with a particular `.html` file with that name

src/webgpu/web_platform/reftests/canvas_use_as_image_source_opaque.https.html
```diff
@@ -0,0 +1,32 @@
+<html class="reftest-wait">
```
kainino0x, line 1: nit: perhaps put all of these in their own folder, and also clarify the name since these are all copying from webgpu canvases specifically. (`web_platform/reftests/webgpu_canvas_as_image_source/opaque.https.html` for example)

src/webgpu/web_platform/reftests/ref/canvas_use_as_image_source_utils.html.ts
```diff
@@ -0,0 +1,138 @@
...
+varying vec2 v_Texcoord;
+uniform sampler2D u_Texture2D;
+void main() {
+  gl_FragColor = texture2D(u_Texture2D, v_Texcoord);
+}
+`;
+
+const kVertices = new Float32Array([1, 1, -1, 1, -1, -1, 1, 1, -1, -1, 1, -1]);
+
+function WebGLDrawTexQuad(gl: WebGLRenderingContext) {
```
kainino0x, line 59: note: If we switch to non-reftests - might be able to use readPixels instead of drawing a quad.

src/webgpu/web_platform/reftests/canvas_use_as_image_source.html.ts
```diff
@@ -0,0 +1,98 @@
...
+    }
+    buffer.unmap();
+
+    const encoder = t.device.createCommandEncoder();
+    encoder.copyBufferToTexture({ buffer, bytesPerRow }, { texture: ctx.getCurrentTexture() }, [
+      ctx.canvas.width,
+      ctx.canvas.height,
+      1,
+    ]);
+    t.device.queue.submit([encoder.finish()]);
```
kainino0x, line 88: nit: I think this could more simply use `writeTexture`

kainino0x (commented):
The semantics have changed since most/all of these tests were written - from compositingAlphaMode which doesn't affect use as an image source, to alphaMode which does. So for all the existing tests that we're updating the naming in, we need to also make sure they're still correct.

I know it's only a few things, but can you split updating of existing tests into a separate PR so it's easier to make sure we review them?

As for the `canvas_use_as_image_source` tests - I'm wondering if we should use reftests for this or not. We only *need* reftests to test the actual browser output for behavior that isn't directly observable to JS - but when using a canvas as an image source you can read back the result and check it directly. We already have some tests that do this, but maybe only into WebGPU canvases, which means it would take some effort to write readback paths for the other canvas types (which we don't have to do if we use a reftest).

But on the other hand, we also need to expand these tests to other "types" of webgpu source canvases, in particular offscreen canvases (but also others; see: https://github.com/gpuweb/cts/pull/1638#discussion_r916322274). If we do that then the number of combinations gets quite huge, which is much easier with the combinatorial framework than with reftests. In the end I think we would have:

- x= source is each the 5 canvas types from that link
- x= dst uses each of the 6 upload methods you have here, plus `transferToImageBitmap`, `convertToBlob`, and probably some others.

Of course we still need reftests for the compositing behavior of alphaMode (if we don't already have any), but those will be much smaller tests.

shrekshao:
Add alpha mode to `readbackFromWebGPUCanvas.spec.ts` at https://github.com/gpuweb/cts/pull/1673.
Closing this update on reftests side.

Issue #1662, opened 2022-07-18
val: Enhance `ranges` test in writeBuffer.spec.ts


Gyuyoung:
Add more checks regarding the ranges to writeBuffer:`ranges` test.

PR #1663, opened 2022-07-18
Add an out-of-range check to the ranges test in writeBuffer.spec.ts

Gyuyoung:
This PR adds a new check to ensure that writing with a data offset
that is out of range of data size with undefined size value generates
an operation error.

Issue: #1662

<hr>

### DIFF
--- a/src/webgpu/api/validation/queue/writeBuffer.spec.ts
+++ b/src/webgpu/api/validation/queue/writeBuffer.spec.ts
@@ -133,6 +133,9 @@ g.test('ranges')
       // Writing with a data offset that is out of range of data size.
       t.shouldThrow('OperationError', () => queue.writeBuffer(buffer, 0, arraySm, 9, 0));
 
+      // Writing with a data offset that is out of range of data size with undefined size.
+      t.shouldThrow('OperationError', () => queue.writeBuffer(buffer, 0, arraySm, 9, undefined));
+
       // A data offset of undefined should be treated as 0.
       queue.writeBuffer(buffer, 0, arraySm, undefined, 8);
       t.shouldThrow('OperationError', () => queue.writeBuffer(buffer, 0, arraySm, undefined, 12));
--- a/src/webgpu/api/validation/queue/writeBuffer.spec.ts
+++ b/src/webgpu/api/validation/queue/writeBuffer.spec.ts
@@ -133,7 +133,7 @@ g.test('ranges')
       // Writing with a data offset that is out of range of data size.
       t.shouldThrow('OperationError', () => queue.writeBuffer(buffer, 0, arraySm, 9, 0));
 
-      // Writing with a data offset that is out of range of data size with undefined size.
+      // Writing with a data offset that is out of range of data size with implicit copy size.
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/queue/writeBuffer.spec.ts
```diff
@@ -141,6 +137,9 @@ Also verifies that the specified data range:
       // A data offset of undefined should be treated as 0
       queue.writeBuffer(buffer, 0, arraySm, undefined, 8);
       t.shouldThrow('OperationError', () => queue.writeBuffer(buffer, 0, arraySm, undefined, 12));
+
+      // Writing with a size that is a negative value.
+      t.shouldThrow('OperationError', () => queue.writeBuffer(buffer, 0, arraySm, 0, -4));
```
Gyuyoung, line 142: When running this check, `TypeError' happens instead of `OperationError`. It seems to me that Chromium generates `TypeError`. Do you think I need to find a bug with this fail in the Chromium Gerrit?

kainino0x: TypeError is correct. The WebIDL specifies that the `writeBuffer` 5th parameter, `size`, has type `GPUSize64` = `[EnforceRange] unsigned long long`. The check for `contentsSize ≥ 0` is only relevant when `size` is `undefined` and so `contentsSize` has been computed as `dataSize − dataOffset`. It could alternatively be expressed like, "dataOffset must be ≤ dataSize", which perhaps would be clearer. (I don't remember why the spec doesn't word it that way.)

kainino0x: EnforceRange defined here: https://webidl.spec.whatwg.org/#EnforceRange

Gyuyoung: Ah, I misunderstood the meaning of the description. Thanks for correcting me. It looks like the 135 lines test the case that `dataOffset > dataSize` with 0 as a `size` valie like below,

      // Writing with a data offset that is out of range of data size
      t.shouldThrow('OperationError', () => queue.writeBuffer(buffer, 0, arraySm, 9, 0));

Do you think that it would be good to add a check when `size` is not defined?

      // Writing with a data offset that is out of range of data size
      t.shouldThrow('OperationError', () => queue.writeBuffer(buffer, 0, arraySm, 15));

If it's redundant, let me close this PR.

kainino0x: It would be good to add that. We should generally have more tests here that test a `size` of `undefined`, I think a lot of the existing cases could be duplicated in that way.

It's possible these cases are missing because when we wrote these tests originally, `size` defaulted to `0`, whereas now it has no default so `undefined` (or omitted) is different from `0`.

Gyuyoung: ok, then let me add below to the ranges test.

    `t.shouldThrow('OperationError', () => queue.writeBuffer(buffer, 0, arraySm, 9, undefined));`

PR #1640, opened 2022-07-07
Implement val:render_pass_descriptor:depth_stencil_attachment_depth_clear_value

Gyuyoung:
If depthLoadOp of depthStencilAttachment is "clear", depthClearValue
must be between 0.0 and 1.0. So, this PR adds a test to check if
depthClearValue is invalid if the value is out of the range.

Issue: #1618

<hr>

**Requirements for PR author:**

- [x] All missing test coverage is tracked with "TODO" or `.unimplemented()`.
- [x] New helpers are `/** documented */` and new helper files are found in `helper_index.txt`.
- [x] Test behaves as expected in a WebGPU implementation. (If not passing, explain above.)

**Requirements for [reviewer sign-off](https://github.com/gpuweb/cts/blob/main/docs/reviews.md):**

- [ ] Tests are properly located in the test tree.
- [ ] [Test descriptions](https://github.com/gpuweb/cts/blob/main/docs/intro/plans.md) allow a reader to "read only the test plans and evaluate coverage completeness", and accurately reflect the test code.
- [ ] Tests provide complete coverage (including validation control cases). **Missing coverage MUST be covered by TODOs.**
- [ ] Helpers and types promote readability and maintainability.

When landing this PR, be sure to make any necessary issue status updates.

PR #1674, opened 2022-07-19
Enhance val:render_pass_descriptor:depth_stencil_attachment_depth_clear_value

Gyuyoung:
This PR updates render_pass_descriptor:depth_stencil_attachment_depth_clear_value
to check if out of range of depthClearValue is only invalid when depthLoadOp is
`clear`. If depthLoadOp is not `clear`, depthClearValue should be valid even
though the value is out of range.

Issue: #1618

<hr>

### DIFF
--- a/src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts
+++ b/src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts
@@ -803,30 +803,42 @@ g.test('depth_stencil_attachment')
 g.test('depth_stencil_attachment,depth_clear_value')
   .desc(
     `
-  Test that depthClearValue is invalid if the value is out of the range(0.0 and 1.0) when
+  Test that depthClearValue is invalid if the value is out of the range(0.0 and 1.0) only when
   depthLoadOp is 'clear'.
   `
   )
-  .paramsSimple([
-    { depthClearValue: -1.0 },
-    { depthClearValue: 0.0 },
-    { depthClearValue: 0.5 },
-    { depthClearValue: 1.0 },
-    { depthClearValue: 1.5 },
-  ])
+  .params(u =>
+    u
+      .combine('depthLoadOp', ['load', 'clear', undefined] as const)
+      .combineWithParams([
+        { depthClearValue: -1.0 },
+        { depthClearValue: 0.0 },
+        { depthClearValue: 0.5 },
+        { depthClearValue: 1.0 },
+        { depthClearValue: 1.5 },
+      ])
+  )
   .fn(t => {
-    const { depthClearValue } = t.params;
+    const { depthLoadOp, depthClearValue } = t.params;
 
-    const depthStencilTexture = t.createTexture({ format: 'depth24plus-stencil8' });
+    const depthStencilTexture = t.createTexture({
+      format: depthLoadOp === undefined ? 'stencil8' : 'depth24plus-stencil8',
+    });
     const depthStencilAttachment = t.getDepthStencilAttachment(depthStencilTexture);
     depthStencilAttachment.depthClearValue = depthClearValue;
+    depthStencilAttachment.depthLoadOp = depthLoadOp;
+    if (depthLoadOp === undefined) {
+      depthStencilAttachment.depthStoreOp = undefined;
+    }
 
     const descriptor = {
       colorAttachments: [t.getColorAttachment(t.createTexture())],
       depthStencilAttachment,
     };
 
-    t.tryRenderPass(depthClearValue >= 0.0 && depthClearValue <= 1.0, descriptor);
+    const isValid = !(depthLoadOp === 'clear' && (depthClearValue < 0.0 || depthClearValue > 1.0));
+
+    t.tryRenderPass(isValid, descriptor);
   });
 
 g.test('resolveTarget,format_supports_resolve')


src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts
```diff
@@ -803,30 +803,38 @@ g.test('depth_stencil_attachment')
...
+    depthStencilAttachment.depthLoadOp = depthLoadOp;
 
     const descriptor = {
       colorAttachments: [t.getColorAttachment(t.createTexture())],
       depthStencilAttachment,
     };
 
-    t.tryRenderPass(depthClearValue >= 0.0 && depthClearValue <= 1.0, descriptor);
+    const isValid =
+      depthLoadOp === 'clear' && (depthClearValue < 0.0 || depthClearValue > 1.0) ? false : true;
```
kainino0x, line 835: nit: `x ? false : true` is the same as `!x`

src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts
```diff
@@ -803,30 +803,38 @@ g.test('depth_stencil_attachment')
...
+        { depthClearValue: 0.5 },
+        { depthClearValue: 1.0 },
+        { depthClearValue: 1.5 },
+      ])
+  )
   .fn(t => {
-    const { depthClearValue } = t.params;
+    const { depthLoadOp, depthClearValue } = t.params;
 
     const depthStencilTexture = t.createTexture({ format: 'depth24plus-stencil8' });
```
kainino0x, line 824: [1]

src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts
```diff
@@ -803,30 +803,38 @@ g.test('depth_stencil_attachment')
...
-  .paramsSimple([
-    { depthClearValue: -1.0 },
-    { depthClearValue: 0.0 },
-    { depthClearValue: 0.5 },
-    { depthClearValue: 1.0 },
-    { depthClearValue: 1.5 },
-  ])
+  .params(u =>
+    u
+      .combine('depthLoadOp', ['load', 'clear'] as const)
```
kainino0x, line 812: re: https://github.com/gpuweb/cts/pull/1640#discussion_r924015954

> It seems to me that `depth_stencil_attachment` test case can also check the case that `depthLoadOp` is `undefined`. Do you mean that we need to test it in `depth_stencil_attachment,depth_clear_value` again?
> 
> Please correct me if `depth_stencil_attachment` doesn't test the case that you want to test.

`depthLoadOp` has three possible values: `'load'`, `'clear'`, and `undefined`. I want to test that depthClearValue isn't validated when `depthLoadOp` is `undefined`. This is only possible when there is no depth attachment channel, so I think the easiest would be to write the format below ([1]) as `depthLoadOp === undefined ? 'stencil8' : 'depth24plus-stencil8'`.

I don't think this particular thing is covered by other tests but please correct me if I'm wrong.

Gyuyoung: Thank you for explaining it. Yes, it looks like there is no existing test that covers depthClearValue when depthLoadOp is undefined. Let me add `undefined` to `.combine('depthLoadOp', ['load', 'clear'] as const)`.

src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts
Gyuyoung: ok, and, it's likely we need to add `!(depthLoadOp === undefined)` because it should be invalid when depthLoadOp is `undefined`

kainino0x: If it's invalid in that case then we're not successfully testing that the depthClearValue validation is skipped in this case (because we can't tell). There must be some other thing that needs to be different (other than the depth/stencil attachment format).

kainino0x: Ah, I ran the test and it says:

```
EXCEPTION: Unexpected validation error occurred: depthStoreOp (StoreOp::Store) must not be set if the attachment ([TextureView]) has no depth aspect or depthReadOnly (0) is true.
 - While validating depthStencilAttachment.
 - While encoding [CommandEncoder].BeginRenderPass([RenderPassDescriptor]).
```

So if we're omitting the depth aspect we must also omit the depthStoreOp.

src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts
```diff
@@ -803,30 +803,41 @@ g.test('depth_stencil_attachment')
...
-    const { depthClearValue } = t.params;
+    const { depthLoadOp, depthClearValue } = t.params;
 
-    const depthStencilTexture = t.createTexture({ format: 'depth24plus-stencil8' });
+    const depthStencilTexture = t.createTexture({
+      format: depthLoadOp === undefined ? 'stencil8' : 'depth24plus-stencil8',
+    });
     const depthStencilAttachment = t.getDepthStencilAttachment(depthStencilTexture);
     depthStencilAttachment.depthClearValue = depthClearValue;
+    depthStencilAttachment.depthLoadOp = depthLoadOp;
```
kainino0x, line 829:
```suggestion
    depthStencilAttachment.depthLoadOp = depthLoadOp;
    if (depthLoadOp === undefined) {
      depthStencilAttachment.depthStoreOp = undefined;
    }
```

src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts
```diff
@@ -803,30 +803,41 @@ g.test('depth_stencil_attachment')
...
 
     const descriptor = {
       colorAttachments: [t.getColorAttachment(t.createTexture())],
       depthStencilAttachment,
     };
 
-    t.tryRenderPass(depthClearValue >= 0.0 && depthClearValue <= 1.0, descriptor);
+    const isValid =
+      !(depthLoadOp === 'clear' && (depthClearValue < 0.0 || depthClearValue > 1.0)) &&
+      !(depthLoadOp === undefined);
```
kainino0x, line 838:
```suggestion
    const isValid = !(depthLoadOp === 'clear' && (depthClearValue < 0.0 || depthClearValue > 1.0));
```

kainino0x (commented):
I think these suggestions will make it work

src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts
Gyuyoung: Ah, thank you for finding it. Fixed.

kainino0x (approved):

Issue #1225, opened 2022-04-25
wgsl: execution test for mix builtin


dj2:
Add an execution test for the `mix` builtin.

```
T is AbstractFloat, f32, f16, vecN<AbstractFloat>, vecN<f32>, or vecN<f16>

@const fn mix(e1: T ,e2: T ,e3: T) -> T

Returns the linear blend of e1 and e2 (e.g. e1*(1-e3)+e2*e3). Component-wise when T is a vector.
```

See `mix` in https://www.w3.org/TR/WGSL/#float-builtin-functions

dj2:
Also:

```
T is AbstractFloat, f32, or f16
T2 is vecN<T>

@const fn mix(e1: T2 ,e2: T2 ,e3: f32 ) -> T

Returns the component-wise linear blend of e1 and e2, using scalar blending factor e3 for each component.

Same as mix(e1,e2,T2(e3)).
```

zoddicus:
f32 work completed, splitting [AbstractFloat](https://github.com/gpuweb/cts/issues/2566) and [f16](https://github.com/gpuweb/cts/issues/2567) into their own issues.

PR #1719, opened 2022-07-27
Implement `mix` f32 tests

zoddicus:
Issue #1225

<hr>

### DIFF
--- a/src/unittests/f32_interval.spec.ts
+++ b/src/unittests/f32_interval.spec.ts
@@ -29,10 +29,13 @@ import {
   log2Interval,
   maxInterval,
   minInterval,
+  mixImpreciseInterval,
+  mixPreciseInterval,
   multiplicationInterval,
   negationInterval,
   powInterval,
   radiansInterval,
+  roundInterval,
   saturateInterval,
   signInterval,
   sinInterval,
@@ -40,7 +43,6 @@ import {
   subtractionInterval,
   tanInterval,
   ulpInterval,
-  roundInterval,
 } from '../webgpu/util/f32_interval.js';
 import { hexToF32, hexToF64, oneULP } from '../webgpu/util/math.js';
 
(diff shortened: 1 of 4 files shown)


zoddicus:
This is currently broken, uploading for others to see what is going wrong. Will update it with a corrected version before adding reviewers.

dj2:
So, does this need the spec to be updated before landing? Is the intention that a conformant implementation can pass _one_ of the two cases?

zoddicus:
My understanding from the discussion is that a conformant implement has to pass atleast one of the two options, since that is what Vulkan's CTS enforces.

Previously with PRs where the spec needs to change we have landed them with a comment in the could about the spec issue, if they are less restrictive than what currently is in the spec. i.e. no current spec compliant would fail the new version of the test. 

I am fine with leaving this pending until the spec change lands if you prefer, I mostly want it shared so there was a non-local version of the diff for people to look at

dj2:
I'm not sure how to encode the must pass one of the two, having a set of tests that a conformant implementation can _fail_ seems wrong.

Could we define two intervals for the answer and say it has to be in one of the two?

zoddicus:
> I'm not sure how to encode the must pass one of the two, having a set of tests that a conformant implementation can _fail_ seems wrong.
> 
> Could we define two intervals for the answer and say it has to be in one of the two?

I think this is already done?
Instead of having mixInterval return [F32Interval], there is a mixPreciseInterval and mixImpreciseInterval that return F32Interval, and the actual CTS running code in makeCase is responsible for including both of those expectations.

Specifically in `return makeTernaryF32IntervalCase(x, y, z, mixPreciseInterval, mixImpreciseInterval);`
The last param of makeTernaryF32IntervalCase is a spread of F32Intervals, if the result is in any of those intervals it will pass.

This is very similar to how we handle the clamp being either implemented via median of 3 or MinMax. There is two interval generator operations defined, and the actual CTS test uses the spread behaviour to check at least one works.
Each of this implementations has separate unit tests to ensure they behave correctly independently.

The interval array/spread handling is in the make*F32IntervalCase layer of the framework, so that all of the code in f32_interval.ts doesn't need to be aware of there being potentially array of intervals being passed around. I originally attempted, when working on clamp, to have a single clampInterval that returned a [F32Interval], but this introduced a bunch of complexity to interfaces, which bled out into other tests that didn't really need to be aware of this type of return.  The actual test running code could never truly avoid this complexity, since it either needs to allow passing in multiple expectation or take in an array of expectation so all of the complexity has been isolated there.

I really need to get on writing up a doc explaining how all of this hangs together and the reasoning for it.

dj2 (approved):

Issue #1743, opened 2022-08-10
val: Test beginComputePass and GPUComputePassDescriptor validation


Gyuyoung:
SSIA.

Gyuyoung:
I couldn't find more tasks that need to be tested in this issue. So, I would like to mark the status of CTS from `Started` to `Complete`. If I find more tasks, let me change the status again.

kainino0x:
Closing because it was marked complete, which I think is accurate.

PR #1744, opened 2022-08-10
Add beginComputePass validation test

Gyuyoung:
This PR adds beginComputePass.spec.ts file, then add
compute_pass_descriptor_location test in order to test
the validation of beginComputePass and its GPUComputePassDescriptor
as a parameter.

Issue: #1743

<hr>


src/webgpu/api/validation/encoding/beginComputePass.spec.ts
```diff
@@ -0,0 +1,57 @@
+export const description = `
+Tests for validation in beginComputePass and GPUComputePassDescriptor as its optional descriptor.
+`;
+
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { ValidationTest } from '../validation_test.js';
+
+export const g = makeTestGroup(ValidationTest);
+
+g.test('compute_pass_descriptor_location')
```
kainino0x, line 10: since the file is called `beginComputePass` it's not that necessary to also say `compute_pass_descriptor`. Shorter is nice for navigability as long as it's still easy to understand.

I would call this "timestamp_writes_location" to match the equivalent `createRenderPass` test, or `timestampWrites,same_location` to use the name of the descriptor field and a slightly more descriptive name for the test.

src/webgpu/api/validation/encoding/beginComputePass.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+  )
+  .beforeAllSubcases(t => {
+    t.selectDeviceOrSkipTestCase(['timestamp-query']);
+  })
+  .fn(async t => {
+    const { locationA, locationB } = t.params;
+
+    const querySet = t.device.createQuerySet({
+      type: 'timestamp',
+      count: 1,
```
kainino0x, line 29: Hmm... this seems wrong, how can we use queryIndex 1 if there's only one entry in the query set? Shouldn't this cause the test to fail because `beginComputePass( 0 beginning, 1 end )` is expected to be valid but won't be? Are we missing validation in `beginComputePass`?

Looks like the render pass version has the same problem - sorry I totally missed that before. You can fix both in this PR, I think.

src/webgpu/api/validation/encoding/beginComputePass.spec.ts
Gyuyoung: Ah, I see. Then, let me use `timestampWrites,same_location` for the test. Thanks.

src/webgpu/api/validation/encoding/beginComputePass.spec.ts
Gyuyoung: Ah, it's likely I missed catching this problem because the test has not generated a validation error when the locations are the same. I thought that Chromium doesn't generate a validation error when the locations are the same. Ok, I update the problem for both beginComputePass and render_pass_descriptor.

kainino0x: Does Chromium pass the tests now or is it still missing validation for duplicate locations?

Anyway, yeah, it's tricky to write tests when the implementation isn't correct either. It's good to look closely at the error messages and which cases fail to make sure they're the ones you expect given any known implementation bugs. I think that would have caught this because you should have seen an error about writing to a query set index that's out of bounds. At least, hopefully.

kainino0x (approved):

PR #1758, opened 2022-08-17
Pass a dummy pipeline layout to createNoOpComputePipeline

Gyuyoung:
The tests that use createNoOpComputePipeline have been failing because
it could not convert `layout`. To avoid the test failures, this PR
passes a dummy pipeline layout to createNoOpComputePipeline.

Additionally, this PR updates the indentations of the descriptions
in the tests.

<hr>

### DIFF
--- a/src/webgpu/api/validation/encoding/cmds/compute_pass.spec.ts
+++ b/src/webgpu/api/validation/encoding/cmds/compute_pass.spec.ts
@@ -12,7 +12,8 @@ import { ValidationTest } from '../../validation_test.js';
 class F extends ValidationTest {
   createComputePipeline(state: 'valid' | 'invalid'): GPUComputePipeline {
     if (state === 'valid') {
-      return this.createNoOpComputePipeline();
+      const layout = this.device.createPipelineLayout({ bindGroupLayouts: [] });
+      return this.createNoOpComputePipeline(layout);
     }
 
     return this.createErrorComputePipeline();
@@ -49,8 +50,8 @@ export const g = makeTestGroup(F);
 g.test('set_pipeline')
   .desc(
     `
-setPipeline should generate an error iff using an 'invalid' pipeline.
-`
+  setPipeline should generate an error iff using an 'invalid' pipeline.
+  `
   )
   .params(u => u.beginSubcases().combine('state', ['valid', 'invalid'] as const))
   .fn(t => {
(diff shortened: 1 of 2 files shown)


src/webgpu/api/validation/encoding/cmds/compute_pass.spec.ts
```diff
@@ -189,7 +195,8 @@ g.test('indirect_dispatch_buffer,device_mismatch')
   .fn(async t => {
     const { mismatched } = t.params;
 
-    const pipeline = t.createNoOpComputePipeline();
+    const layout = t.device.createPipelineLayout({ bindGroupLayouts: [] });
+    const pipeline = t.createNoOpComputePipeline(layout);
```
Gyuyoung, line 199: When I ran the tests that call createNoOpComputePipeline without a parameter, the tests failed with the below error. Though I'm not sure if calling createNoOpComputePipeline without a parameter was intended, it looks like the tests succeed if we pass a dummy pipeline layout to createNoOpComputePipeline. Please correct me if we keep the existing createNoOpComputePipeline invocation.

```
  - EXCEPTION: no overload matched for createComputePipeline:
    value is not object
    while converting member 'layout'
    TypeError: no overload matched for createComputePipeline:
    value is not object
    while converting member 'layout'
        at F.createNoOpComputePipeline (/home/gyuyoung/github/cts/out-node/webgpu/api/validation/validation_test.js:301:28)
        at RunCaseSpecific.fn (/home/gyuyoung/github/cts/out-node/webgpu/api/validation/encoding/cmds/compute_pass.spec.js:94:24)
```

kainino0x: This sounds like a bug of the dawn-node bindings rejecting the string `'auto'` as a layout. This would be because we changed the WebGPU API from implicitly selecting a default layout, to requiring you to explicitly request it by saying `layout: 'auto'`.

If you agree this sounds like what is happening could you please file a bug? https://crbug.com/dawn/new

Gyuyoung: ok, I just filed a bug there - https://bugs.chromium.org/p/dawn/issues/detail?id=1515. Then, do you think this PR is unnecessary, right?

Gyuyoung: enga@ answered on the issue. According to enga@, Chromium doesn't support `auto` yet.

kainino0x: Austin didn't realize the bug was filed against dawn-node. I replied there.

Yes, I think we should keep the current invocations and close this issue.

Gyuyoung: ok, thanks for clarifying it. Closed this PR.

PR #1768, opened 2022-08-19
Change .origin in GPUImageCopyExternalImage from 3D to 2D

shaoboyan:
This PR changes sourceOrigin in GPUImageCopyTexture from 3D to 2D
which follows the spec.

Issue: #

<hr>

### DIFF
--- a/src/webgpu/api/validation/queue/copyToTexture/CopyExternalImageToTexture.spec.ts
+++ b/src/webgpu/api/validation/queue/copyToTexture/CopyExternalImageToTexture.spec.ts
@@ -45,16 +45,16 @@ interface WithDstOriginMipLevel extends WithMipLevel {
 }
 
 // Helper function to generate copySize for src OOB test
-function generateCopySizeForSrcOOB({ srcOrigin }: { srcOrigin: Required<GPUOrigin3DDict> }) {
+function generateCopySizeForSrcOOB({ srcOrigin }: { srcOrigin: Required<GPUOrigin2DDict> }) {
   // OOB origin fails even with no-op copy.
-  if (srcOrigin.x > kDefaultWidth || srcOrigin.y > kDefaultHeight || srcOrigin.z > kDefaultDepth) {
+  if (srcOrigin.x > kDefaultWidth || srcOrigin.y > kDefaultHeight) {
     return [{ width: 0, height: 0, depthOrArrayLayers: 0 }];
   }
 
   const justFitCopySize = {
     width: kDefaultWidth - srcOrigin.x,
     height: kDefaultHeight - srcOrigin.y,
-    depthOrArrayLayers: kDefaultDepth - srcOrigin.z,
+    depthOrArrayLayers: 1,
   };
 
   return [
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/queue/copyToTexture/CopyExternalImageToTexture.spec.ts
```diff
@@ -858,7 +856,7 @@ g.test('OOB,source')
     if (
       srcOrigin.x + copySize.width > kDefaultWidth ||
       srcOrigin.y + copySize.height > kDefaultHeight ||
-      srcOrigin.z + copySize.depthOrArrayLayers > 1
+      copySize.depthOrArrayLayers > 1
```
Gyuyoung, line 859: I updated the test to meet the below requirement in the specification. But, you mean the requirement was already changed, right?

  `source.origin.z + copySize.depthOrArrayLayers must be ≤ 1`

shaoboyan: No, what I mean is that, sourceOrigin is a 2D one which don't have source.origin.z.

Gyuyoung: ok, I see. LGTM.

Gyuyoung (approved):

src/webgpu/api/validation/queue/copyToTexture/CopyExternalImageToTexture.spec.ts
kainino0x: @Gyuyoung did these tests pass when you wrote them? If so I guess it was just by luck that the tests didn't fail due to this.

I filed https://github.com/gpuweb/types/issues/104 to enhance the types to catch this.

kainino0x (approved):

src/webgpu/api/validation/queue/copyToTexture/CopyExternalImageToTexture.spec.ts
Gyuyoung: Ah, I remember that the test was skipped. So it looks like it wasn't detected.

shaoboyan: @Gyuyoung @kainino0x If this is the only expectation file (https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/webgpu-cts/expectations.txt), I'm a bit worry about there might be a hidden bug. I don't think this case has been skipped in this expectation file but I think the bots still works fine? (We find this issue by click the cts page manually in canary).
So I'm a bit hard to understand why bots works. Do you have any clues?

kainino0x: > Ah, I remember that the test was skipped. So it looks like it wasn't detected.

Do you mean it was skipped due to running it in dawn-node? That would make sense, since this is a web platform interop test.

> So I'm a bit hard to understand why bots works. Do you have any clues?

I think it is simply that there has not been a CTS roll since this regression landed.
Dawn currently is using ed11c6ba35570aea10ca3fb5245bbee362915f36 ([seen here](https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/DEPS;l=158;drc=cf2ac51987c49d10f1d5da210a3ed369f649d10d)), which is before 242105385aff3f7088a04b939c8697def26899da.

Gyuyoung: > > Ah, I remember that the test was skipped. So it looks like it wasn't detected.
> 
> Do you mean it was skipped due to running it in dawn-node? That would make sense, since this is a web platform interop test.

Yes, I supposed that the dawn skipped the test. I've tested 
CTS tests using the dawn and I know that the dawn is only the way to skip tests, isn't it?

kainino0x: Many tests will skip themselves if the system they're trying to interoperate with doesn't exist. I'm guessing that's what happened here if you ran using dawn-node rather than Chromium.

For example, if a browser doesn't implement transferControlToOffscreen(), we want to skip tests that use that function, rather than fail them. Similarly, we want them to be able to test a Node bindings implementation which doesn't have the rest of the Web platform (like HTMLVideoElement which would require an entire DOM implementation).

Gyuyoung: Thank you for explaining when the dawn skips tests. And, it looks like it would be good to mention that a new test is skipped, passed/failed in the PR together next time.

Issue #885, opened 2022-01-13
val: createBindGroupLayout


kainino0x:
Mostly done in 
https://github.com/gpuweb/cts/pull/719

kainino0x:
@Gyuyoung assigning you and marking the status as "started". Please do this when you start an issue so that other contributors can see that the issue is taken. (We have at least one more person working on CTS right now, so it's a bit more important than before.)

Gyuyoung:
@kainino0x Ah, I'm sorry for missing it. Let me keep it in mind. Thanks for assigning this issue to me.

Gyuyoung:
I couldn't find untested validation check for createBindGroupLayout now. So, I would like to mark this issue from `Started` to `Complete`. If there is untested validation check, let me reopen this issue.

PR #1770, opened 2022-08-19
Add visibility,VERTEX_shader_stage_storage_texture_access test to createBindGroupLayout.spec.ts

Gyuyoung:
According to the specification, the storageTexture.access of the entry must not be
"write-only" if the entry's visibility includes VERTEX. So this PR adds a new test
to ensure that a validation error is generated if the entry's visibility is VERTEX
and the storageTexture.access is "write-only".

Issue: #885

<hr>

### DIFF
--- a/src/webgpu/api/validation/createBindGroupLayout.spec.ts
+++ b/src/webgpu/api/validation/createBindGroupLayout.spec.ts
@@ -9,6 +9,7 @@ import { makeTestGroup } from '../../../common/framework/test_group.js';
 import {
   kShaderStages,
   kShaderStageCombinations,
+  kStorageTextureAccessValues,
   kTextureViewDimensions,
   allBindingEntries,
   bindingTypeInfo,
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/createBindGroupLayout.spec.ts
```diff
@@ -76,6 +77,37 @@ g.test('visibility')
...
+  .params(u =>
+    u //
+      .combine('shaderStage', kShaderStageCombinations)
+      .beginSubcases()
+      .combine('access', [undefined, ...kStorageTextureAccessValues])
+  )
+  .fn(async t => {
+    const { shaderStage, access } = t.params;
+
+    const success = !(access === 'write-only' && shaderStage & GPUShaderStage.VERTEX);
```
Gyuyoung, line 96: According to the spec, if entry.visibility includes VERTEX, entry.storageTexture?access must not be "write-only". So, that's why I set `success` with this condition. But, when I ran this test, the test generated a validation error with the below error message. It's likely storageTexture only could be used with ShaderStage::(Fragment|Compute), not ShaderStage::Vertex.

`  - EXCEPTION: Unexpected validation error occurred: storageTexture bindings cannot be used with a visibility of ShaderStage::Vertex. Only ShaderStage::(Fragment|Compute) are allowed.`

However, I couldn't find this behavior in the specification. I wonder if we need to update Chromium/dawn to meet this requirement. If I miss finding the current behavior(storageTexture bindings cannot be used with a visibility of ShaderStage::Vertex) in the specification, please correct me.

kainino0x: `"write-only"` is actually the *only* possible value of `access`. It's the default, so the case of `undefined` is actually also `"write-only"`. (The reason we still like to test both even when there's a default, is just so we can test the defaulting behavior.)
https://gpuweb.github.io/gpuweb/#dictdef-gpustoragetexturebindinglayout

So I think the implementation is correct.

kainino0x: If you implement the same defaulting here, like so:
```suggestion
    const success = !((access ?? 'write-only') === 'write-only' && shaderStage & GPUShaderStage.VERTEX);
```
then this test might pass.

Gyuyoung: Thank you for explaining it. And, I update the test with the suggestion.

kainino0x (approved):

PR #1779, opened 2022-08-23
Remove pushing lists of numbers through Set to guarantee uniqueness

### LINKED ISSUE #1778: wgsl: Don't use set when generating flushed/non-flushed number lists
Currently to generate a list of numbers that are flushed and not flushed the code naively adds a zero for every subnormal number, which means the list ends up with duplicates. This is resolved via pushing the list through a Set to remove duplicates. Instead we could use .some(isSubnormalNumber) to add only once, thuse avoid needing to do Array -> Set -> Array.

There are other usages of Set, when the FP builds up the results of running op.impl, that should remain.

zoddicus:
Removes a number of locations where the FP framework was transforming
Array -> Set -> Array to guarantee uniqueness, and replaces them with
a helper that generates lists without duplicates.

Fixes #1778

<hr>

### DIFF
--- a/src/webgpu/util/f32_interval.ts
+++ b/src/webgpu/util/f32_interval.ts
@@ -142,7 +142,10 @@ function toF32Vector(v: number[] | F32Vector): F32Vector {
   unreachable(`Cannot convert [${v}] to F32Vector`);
 }
 
-/** @returns the input + zero if any of the entries are subnormal, otherwise returns the input */
+/**
+ * @returns the input plus zero if any of the entries are subnormal, otherwise
+ * returns the input
+ */
 function addFlushedIfNeeded(values: number[]): number[] {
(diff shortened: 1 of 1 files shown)


src/webgpu/util/f32_interval.ts
```diff
@@ -141,6 +142,11 @@ function toF32Vector(v: number[] | F32Vector): F32Vector {
   unreachable(`Cannot convert [${v}] to F32Vector`);
 }
 
+/** @returns the input + zero if any of the entries are subnormal, otherwise returns the input */
```
ben-clayton, line 145: I'd avoid `+` here, as it looks like an addition instead of a concatenation.

ben-clayton (approved):

src/webgpu/util/f32_interval.ts
zoddicus: Done

PR #1780, opened 2022-08-23
webgpu/shader/execution: Check validation before testing results

ben-clayton:
Check that the shaders compiled, pipelines created, workgroups dispatched, etc, before attempting to inspect the shader output.

For shaders that are not currently compiling, this massively reduces the amount of error spam, as we no longer spew an error per failed expression case, or which some calls to `run()` contain thousands.

Issue: #

<hr>

### DIFF
--- a/src/webgpu/shader/execution/expression/expression.ts
+++ b/src/webgpu/shader/execution/expression/expression.ts
@@ -133,7 +133,7 @@ export interface ExpressionBuilder {
  * @param cfg test configuration values
  * @param cases list of test cases
  */
-export async function run(
+export function run(
   t: GPUTest,
   expressionBuilder: ExpressionBuilder,
   parameterTypes: Array<Type>,
@@ -167,7 +167,9 @@ export async function run(
     }
   })();
 
-  // Submit all the batches
+  // Submit all the batches inside an error scope.
+  // Note: there is no async work between "push" and "pop" so this is safe to
+  // run concurrently with itself (in other subcases).
   t.device.pushErrorScope('validation');
 
   const checkResults: Array<() => void> = [];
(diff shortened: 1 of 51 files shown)


src/webgpu/shader/execution/expression/expression.ts
```diff
@@ -167,21 +167,25 @@ export function run(
...
-      batchCases,
-      cfg.inputSource
+    checkResults.push(
+      submitBatch(t, expressionBuilder, parameterTypes, returnType, batchCases, cfg.inputSource)
     );
-    checkResults.push(checkResult);
   }
 
+  // Check GPU validation (shader compilation, pipeline creation, etc) before checking the results
+  const error = await t.device.popErrorScope();
```
kainino0x, line 182: This looks like it will work but the code is a little fragile: since subcases run concurrently, it's critical that there are no `await`s between pushErrorScope and popErrorScope. There aren't, but since the function is async, it would be easy to accidentally break later.

We can actually make it non-async by using what we call "eventualAsyncExpectation" which allows the test to continue doing stuff asynchronously after the end of the test function. I'll open a version that does that but also make a small suggestion to this PR that would also avoid that problem.

kainino0x: https://github.com/gpuweb/cts/pull/1784

kainino0x: And pushed a commit to this PR which does the small change.

ben-clayton: Looks great. This is exactly what I was going to do after seeing your other change. Thanks.

kainino0x (approved):
LGTM. The alternative in #1784 avoids putting await everywhere, which I kinda like but don't have a strong preferenc.

I think both are ready to land, so please land whichever one you want.

ben-clayton:
I think this is slightly more fluent, so let's go with this for now. Thank you for your help!

Issue #1757, opened 2022-08-16
Optimize input data generation


austinEng:
In profiles taken while running the CTS, I found that data generation can take a large majority of the test time - often 70-80% for some tests!
For example, methods like:
 - `GetInitialData` in `copyTextureToTexture.spec.ts`
 - `generate_data` in image_copy.spec.ts

These make ArrayBuffer allocations, so we should probably do a similar optimization that @toji did in #1583. 
The tests generally just need some arbitrary data, so we can have some allocator function which returns to the test an offset into some previously-allocated and initialized buffer of data.

cc @kainino0x

toji:
Worked on this today. WIP patch is in #1787 

Drops `copyTextureToTexture` tests from 71s to 20s on my machine.

PR #1787, opened 2022-08-24
Input data generation optimization

toji:
Issue: #1757

<hr>

WIP change, still need to apply the same optimization to `generate_data` in `image_copy.spec.ts` and do a bit more cleanup.

Initial results show that this change takes running the full suite of `copyTextureToTexture` tests **from 71s to 20s on my machine.**

### DIFF
--- a/src/webgpu/api/operation/command_buffer/data_generation.ts
+++ b/src/webgpu/api/operation/command_buffer/data_generation.ts
@@ -1,0 +1,69 @@
+// A helper function that generates ranges of dummy data for buffer or texture operations
+// efficiently. Tries to minimize allocations and data updates.
+export class DataArrayGenerator {
+  private dataBuffer = new Uint8Array(256);
+
+  private lastOffset = 0;
+  private lastStart = 0;
+  private lastByteSize = 0;
+
+  // Find the nearest power of two greater than or equal to the input value.
+  private nextPowerOfTwo(value: number) {
+    return 1 << 32 - Math.clz32(value-1);
+  }
+
+  private generateData(byteSize: number, start: number = 0, offset: number = 0) {
+    const prevSize = this.dataBuffer.length;
+
+    if (prevSize < byteSize) {
+      // If the requested data is larger than the allocated buffer, reallocate it to a buffer large
+      // enough to handle the new request.
+      const newData = new Uint8Array(this.nextPowerOfTwo(byteSize));
+
+      if (this.lastOffset == offset && this.lastStart == start && this.lastByteSize) {
+          // Do a fast copy of any previous data that was generated.
+          newData.set(this.dataBuffer);
+      }
+
+      this.dataBuffer = newData;
+    } else if (this.lastOffset < offset) {
+      // Ensure all values up to the offset are zeroed out.
+      this.dataBuffer.fill(0, this.lastOffset, offset);
+    }
+
+    // If the offset or start values have changed, the whole data range needs to be regenerated.
+    if (this.lastOffset != offset || this.lastStart != start) {
+      this.lastByteSize = 0;
+    }
+
+    // Generate any new values that are required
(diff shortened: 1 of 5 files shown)


toji:
Updated `image_copy.spec.ts` now as well. That one was a bit trickier, because one of it's uses of the generated data overlapped with another one, and as a result if the same underlying buffer was reused it's data returned by the first call would get corrupted by the second, but I addressed that by wrapping the generator in a class (shared by `copyTextureToTexture` and `image_copy`) and having a second instance for that use.

With this change in place both sets of tests now take **less than a quarter** of their previous times, and the data generation functions have practically disappeared from the profiler results.

Runtimes (different machine than the previous timing)

|                 |  copyTextureToTexture.spec.ts | image_copy.spec.ts |
|------------|---------------------------------|----------------------|
| Without Patch | 49s | 66s |
| With Patch| 12s | 25s |

src/webgpu/api/operation/command_buffer/data_generation.ts
```diff
@@ -0,0 +1,69 @@
...
+        this.dataBuffer[i + offset] = ((i ** 3 + i + start) % 251) + 1; // Ensure data is always non-zero
+      }
+
+      this.lastOffset = offset;
+      this.lastStart = start;
+      this.lastByteSize = byteSize;
+    }
+  }
+
+  // Returns a new into the generated data that's the correct length. Because this is a view any
```
kainino0x, line 51: new -> new view

src/webgpu/api/operation/command_buffer/data_generation.ts
```diff
@@ -0,0 +1,69 @@
...
+  private lastOffset = 0;
+  private lastStart = 0;
+  private lastByteSize = 0;
+
+  // Find the nearest power of two greater than or equal to the input value.
+  private nextPowerOfTwo(value: number) {
+    return 1 << (32 - Math.clz32(value - 1));
+  }
+
+  private generateData(byteSize: number, start: number = 0, offset: number = 0) {
```
kainino0x, line 15: `start` and `offset` need documentation, they're a bit hard to understand

src/webgpu/api/operation/command_buffer/image_copy.spec.ts
```diff
@@ -171,10 +176,13 @@ class ImageCopyTest extends GPUTest {
...
   generateData(byteSize: number, start: number = 0, offset: number = 0): Uint8Array {
+    return altDataGenerator.generateView(byteSize, start, offset);
+
     const arr = new Uint8Array(byteSize);
-    for (let i = 0; i < byteSize; ++i) {
+    arr.set(altDataGenerator.generateView(byteSize, start, offset));
+    /*for (let i = 0; i < byteSize; ++i) {
       arr[i + offset] = (i ** 3 + i + start) % 251;
-    }
+    }*/
```
kainino0x, line 185: dead code and commented out code here. Also I think this function is no longer used at all?

kainino0x (approved):

src/webgpu/api/operation/command_buffer/image_copy.spec.ts
toji: Done

src/webgpu/api/operation/command_buffer/data_generation.ts
toji: Done

src/webgpu/api/operation/command_buffer/data_generation.ts
toji: Done

austinEng:
Awesome stuff! I'm not going to have a chance to review this soon (lots to catch up on), so I defer to Kai's review.

toji:
CI failure is unrelated to this PR, so I'm going to go ahead and merge.

PR #1790, opened 2022-08-24
Add 'maximum_binding_limit' test to createBindGroupLayout.spec.ts

Gyuyoung:
The specification says that the binding of each entry must be less than
640. So this PR adds a new test to ensure that a validation error is
generated if the binding number exceeds the maximum binding limit.

Issue: #885

<hr>

### DIFF
--- a/src/webgpu/api/validation/createBindGroupLayout.spec.ts
+++ b/src/webgpu/api/validation/createBindGroupLayout.spec.ts
@@ -53,6 +53,39 @@ g.test('duplicate_bindings')
     }, !_valid);
   });
 
+// Need to use the maxBindingsPerBindGroup constant of GPUSupportedLimits instead of 640.
+const kMaxBindingNumber = 640;
+g.test('maximum_binding_limit')
+  .desc(
+    `
+  Test that a validation error is generated if the binding number exceeds the maximum binding limit.
+
+  TODO: Check if a higher maximum binding limit can be allowed. (e.g. 65535)
+  `
+  )
+  .paramsSubcasesOnly(u =>
+    u //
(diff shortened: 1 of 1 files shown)


Gyuyoung:
The test all passed.

src/webgpu/api/validation/createBindGroupLayout.spec.ts
```diff
@@ -53,6 +54,37 @@ g.test('duplicate_bindings')
     }, !_valid);
   });
 
+const kMaxBindingNumber = kLimitInfo.maxComputeWorkgroupsPerDimension.default;
```
kainino0x, line 57: `maxComputeWorkgroupsPerDimension` is not the correct limit here. There is currently no corresponding limit for max binding number... but we actually just agreed to make this a configurable limit, and lower its default value to 640 due to driver issues seen by some implementations. Spec change is here: https://github.com/gpuweb/gpuweb/pull/3318

Let's go ahead and change this test to use 640 now. Eventually, we will want to add a test case where we ask for a higher `maxComputeWorkgroupsPerDimension`, but we don't have a way to do that right now (because we need to know whether the adapter supports a higher limit). We can add a TODO for it.

Gyuyoung: Ah, I see. Then, let me change this test to use 640. But, please correct me if 640 should be allowed.

> Let's go ahead and change this test to use 640 now. Eventually, we will want to add a test case where we ask for a  higher maxComputeWorkgroupsPerDimension, but we don't have a way to do that right now (because we need to  know whether the adapter supports a higher limit). We can add a TODO for it.

ok, I simply add a TODO for it like this "TODO: Check if a higher maximum binding limit can be allowed."

kainino0x (approved):
LGTM, I fixed some nitpicks in a commit.

Issue #1781, opened 2022-08-23
op: requestDevice


lokokung:

lokokung:
Done now that the PR has landed.

kainino0x:
Should mark the project entry "Status" as "Complete" as well, doing that now.

PR #1798, opened 2022-08-25
Adds remaining adapter.requestDevice tests.

lokokung:
- Removes duplicated tests, and consolidates the limits tests

Issue: #1781

<hr>


lokokung:
Note that the features tests need https://chromium-review.googlesource.com/c/chromium/src/+/3853056 and the invalid adapter test needs https://chromium-review.googlesource.com/c/chromium/src/+/3858305.

src/webgpu/capability_info.ts
```diff
@@ -1048,6 +1048,19 @@ export const kLimitInfo = /* prettier-ignore */ makeTable(
...
+  "depth-clip-control",
+  "depth32float-stencil8",
+  "texture-compression-bc",
+  "texture-compression-etc2",
+  "texture-compression-astc",
+  "timestamp-query",
+  "indirect-first-instance",
+  "shader-f16",
+  "bgra8unorm-storage",
+];
```
kainino0x, line 1062: Best to use the object pattern, like this:
https://github.com/gpuweb/cts/blob/63c716bd354b19616360b6e7047d1508d40e7c89/src/webgpu/capability_info.ts#L70-L78

This looks really silly, but ensures the list here stays up to date with the list in the type definitions (which is generated from the spec).

src/webgpu/api/operation/adapter/requestDevice.spec.ts
```diff
@@ -1,16 +1,301 @@
...
+    `
+    Test requesting the default device.
+    - No features listed in default device
+    - Default limits`
+  )
+  .fn(async t => {
+    const gpu = getGPU();
+    const adapter = await gpu.requestAdapter();
+    assert(adapter !== null);
+    const device = await adapter.requestDevice(undefined);
```
kainino0x, line 28: This test seems to test the same thing as the next one?

src/webgpu/api/operation/adapter/requestDevice.spec.ts
```diff
@@ -1,16 +1,301 @@
...
+      const lostInfo = await device.lost;
+      t.expect(lostInfo.reason === 'destroyed');
+    }
+
+    // The adapter should now be invalid since a device was lost. Requesting another device should
+    // return an already lost device.
+    var kTimeoutMS = 1000;
+    const device = await adapter.requestDevice();
+    assert(device !== null);
+    t.shouldResolve(raceWithRejectOnTimeout(device.lost, kTimeoutMS, 'device was not lost'));
```
kainino0x, line 96: Add a check here that the `reason` is `undefined`.

src/webgpu/api/operation/adapter/requestDevice.spec.ts
```diff
@@ -1,16 +1,301 @@
...
+  .desc(
+    `
+    Test requesting device with an unknown feature.`
+  )
+  .fn(async t => {
+    const gpu = getGPU();
+    const adapter = await gpu.requestAdapter();
+    assert(adapter !== null);
+
+    let feature: any = "unknown-feature";
```
kainino0x, line 109: This probably will require an eslint suppression. Should come up if you run `npm run fix`.

I think you could use `'unknown-feature' as GPUFeatureName` instead though.

src/webgpu/api/operation/adapter/requestDevice.spec.ts
```diff
@@ -1,16 +1,301 @@
...
+    assert(adapter !== null);
+
+    if (adapter.features.has(feature)) {
+      const device = await adapter.requestDevice({ requiredFeatures: [feature] });
+      assert(device !== null);
+      t.expect(device.features.has(feature), 'Device should include the required feature');
+    } else {
+      t.shouldReject(
+        'TypeError',
+        adapter.requestDevice({ requiredFeatures: [feature] })
```
kainino0x, line 140: nit: can deduplicate this call outside of the if-else

src/webgpu/api/operation/adapter/requestDevice.spec.ts
```diff
@@ -1,16 +1,301 @@
...
+  .fn(async t => {
+    const { feature } = t.params;
+
+    const gpu = getGPU();
+    const adapter = await gpu.requestAdapter();
+    assert(adapter !== null);
+
+    if (adapter.features.has(feature)) {
+      const device = await adapter.requestDevice({ requiredFeatures: [feature] });
+      assert(device !== null);
```
kainino0x, line 135: nit: Assert shouldn't be necessary because device isn't supposed to be allowed to be null.
(If it is anyway, the test will crash on the next line on `device.features`.)

kainino0x (approved):

src/webgpu/api/operation/adapter/requestDevice.spec.ts
lokokung: Done

src/webgpu/api/operation/adapter/requestDevice.spec.ts
lokokung: Done.

src/webgpu/api/operation/adapter/requestDevice.spec.ts
lokokung: Done. Intention was to not pass anything in the default one.

src/webgpu/capability_info.ts
lokokung: Done.

src/webgpu/api/operation/adapter/requestDevice.spec.ts
kainino0x: Ah, gotcha. We don't usually test those two things separately since they're identical by the rules of WebIDL, but no harm in doing it for this core function.

src/webgpu/api/operation/adapter/requestDevice.spec.ts
```diff
@@ -1,16 +1,295 @@
...
+    `
+    Test requesting the default device.
+    - No features listed in default device
+    - Default limits`
+  )
+  .fn(async t => {
+    const gpu = getGPU();
+    const adapter = await gpu.requestAdapter();
+    assert(adapter !== null);
+    const device = await adapter.requestDevice();
```
kainino0x, line 28: nit: Since this is the same as the next test other than this line, how about making it a parameterization instead? Then we could also test some extra cases too. Something like this?

```ts
  .paramsSubcasesOnly(u =>
    u.combine('args', [
      [],
      [undefined],
      [{}],
      [{ requiredFeatures: [], requiredLimits: {} }],
    ] as const)
  )
  .fn(async t => {
    const { args } = t.params;

    const gpu = getGPU();
    const adapter = await gpu.requestAdapter();
    assert(adapter !== null);
    const device = await adapter.requestDevice(...args);
```

kainino0x (approved):
LGTM with or without nit

src/webgpu/api/operation/adapter/requestDevice.spec.ts
lokokung: Done

kainino0x (approved):

PR #1819, opened 2022-09-01
Fix a wrong bit operation to decide the validation in indirect_dispatch_buffer,usage

Gyuyoung:
The indirect_dispatch_buffer,usage test has been deciding the validation of
the buffer usage through OR operator between INDIRECT and |bufferUsage|.
The operation has decided that all usages are valid and it's not the purpose
of the test. So this PR replaces `|` operator with `&` operator to fix
the problem.

<hr>

### DIFF
--- a/src/webgpu/api/validation/encoding/cmds/compute_pass.spec.ts
+++ b/src/webgpu/api/validation/encoding/cmds/compute_pass.spec.ts
@@ -240,12 +240,11 @@ g.test('indirect_dispatch_buffer,usage')
     });
     t.trackForCleanup(buffer);
 
-    const isValid = GPUBufferUsage.INDIRECT | bufferUsage;
+    const success = (GPUBufferUsage.INDIRECT & bufferUsage) !== 0;
 
-    const { encoder } = t.createEncoder('compute pass');
+    const { encoder, validateFinish } = t.createEncoder('compute pass');
     encoder.setPipeline(pipeline);
 
-    t.expectValidationError(() => {
-      encoder.dispatchWorkgroupsIndirect(buffer, 0);
-    }, !isValid);
+    encoder.dispatchWorkgroupsIndirect(buffer, 0);
+    validateFinish(success);
   });


src/webgpu/api/validation/encoding/cmds/compute_pass.spec.ts
```diff
@@ -240,7 +240,7 @@ g.test('indirect_dispatch_buffer,usage')
     });
     t.trackForCleanup(buffer);
 
-    const isValid = GPUBufferUsage.INDIRECT | bufferUsage;
+    const isValid = GPUBufferUsage.INDIRECT & bufferUsage;
```
Gyuyoung, line 243: It looks like the original PR used the wrong bit operator. OR operator always sets true to |isValid|. But, dispatchWorkgroupsIndirect doesn't generate a validation error even when the buffer usage doesn't contain INDIRECT. So, it's likely that why the original PR did a mistake. IMO, we need to take a look at dawn or Chromium to find why dispatchWorkgroupsIndirect doesn't generate a validation error when using other usages.

kainino0x: Nice catch! Sorry, it looks like I didn't review this test closely enough.

Calling dispatchWorkgroupsIndirect never generates a validation error. Command encoders are special - when there is an error in a command, it doesn't generate an error; instead it marks the encoder as failed, and the validation error doesn't occur until `.finish()` is called.

Since you are using `createEncoder()` here, you can simply use `validateFinish`. Example here:
https://github.com/gpuweb/cts/blob/a912567d309f6bde49e7e0c3b0671a2e138ab96c/src/webgpu/api/validation/encoding/cmds/render/setVertexBuffer.spec.ts#L35-L37

Gyuyoung: No, it's my fault. I should have checked it. And, thank you for pointing the validateFinish out. I update the test to use validateFinish. Thanks!

kainino0x (approved):

Issue #884, opened 2022-01-13
val: createBindGroup


kainino0x:

kainino0x:
One known TODO: test compatibility between binding and BGL's sampleType ("unfilterable-float" etc)

Gyuyoung:
I would like to take this issue to add missing validation tests.

Gyuyoung:
It looks like there is no covered validation check in createBindGroup. If there is missing validation tests for createBindGroup, let me reopen this issue.

PR #1825, opened 2022-09-05
Add 'buffer,resource_offset' to createBindGroup.spec.ts

Gyuyoung:
The specification says that the resource offset should be a
multiple of limits.[minUniformBufferOffsetAlignment|minStorageBufferOffsetAlignment]
if the BindGroup entry defines buffer and the buffer's type is
[uniform|storage|read-only-storage].

Issue: #884

<hr>

### DIFF
--- a/src/webgpu/api/validation/createBindGroup.spec.ts
+++ b/src/webgpu/api/validation/createBindGroup.spec.ts
@@ -12,7 +12,9 @@ import {
   bufferBindingEntries,
   bufferBindingTypeInfo,
   kBindableResources,
+  kBufferBindingTypes,
   kBufferUsages,
+  kLimitInfo,
   kTextureUsages,
   kTextureViewDimensions,
   sampledAndStorageBindingEntries,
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/createBindGroup.spec.ts
```diff
@@ -793,3 +794,66 @@ g.test('buffer,usage')
...
+              kLimitInfo.minUniformBufferOffsetAlignment.default * 2,
+              kLimitInfo.minUniformBufferOffsetAlignment.default * 3,
+              kLimitInfo.minUniformBufferOffsetAlignment.default + 20,
+            ]
+          : [
+              kLimitInfo.minStorageBufferOffsetAlignment.default,
+              kLimitInfo.minStorageBufferOffsetAlignment.default * 2,
+              kLimitInfo.minStorageBufferOffsetAlignment.default * 3,
+              kLimitInfo.minStorageBufferOffsetAlignment.default + 20,
+            ]
```
kainino0x, line 823: We should test a value which is "minimally unaligned", that is, aligned to the largest power of 2 less than the required alignment. Which you can test with a case that does `* 0.5` or `* 1.5`.

Probably this is more useful than the `* 3` and `+ 20` cases so you could replace them, that is up to you.

kainino0x (approved):
LGTM with an extra case

src/webgpu/api/validation/createBindGroup.spec.ts
Gyuyoung: Ah, I didn't know it well. Thanks. I update the test to use `* 0.5` and `* 1.5`. Additionally, I also use `kBufferBindingTypes` from capability_info.ts here.

PR #1837, opened 2022-09-13
Update `atan2` F32 tests to account for atan based implementations

zoddicus:
Used atan(y/x) as the basis for atan2 accuracy, since this will be less accurate than having a direct atan2 instruction.

<hr>

### DIFF
--- a/src/webgpu/util/f32_interval.ts
+++ b/src/webgpu/util/f32_interval.ts
@@ -844,7 +844,7 @@ const AtanIntervalOp: PointToIntervalOp = {
 };
 
 /** Calculate an acceptance interval of atan(x) */
-export function atanInterval(n: number): F32Interval {
+export function atanInterval(n: number | F32Interval): F32Interval {
   return runPointToIntervalOp(toF32Interval(n), AtanIntervalOp);
 }
 
--- a/src/webgpu/util/f32_interval.ts
+++ b/src/webgpu/util/f32_interval.ts
@@ -850,20 +850,41 @@ export function atanInterval(n: number): F32Interval {
 
 const Atan2IntervalOp: BinaryToIntervalOp = {
   impl: (y: number, x: number): F32Interval => {
-    const numULP = 4096;
-    if (y === 0) {
-      if (x === 0) {
-        return F32Interval.any();
-      } else {
-        return F32Interval.span(
-          ulpInterval(kValue.f32.negative.pi.whole, numULP),
-          ulpInterval(kValue.f32.positive.pi.whole, numULP)
-        );
-      }
+    if (x === 0) {
+      // atan2 is not defined meaningfully here
+      return F32Interval.any();
     }
-    return ulpInterval(Math.atan2(y, x), numULP);
+
+    // y/x == 0, [-π, π]
+    if (y === 0) {
+      // For x > 0 atan2 is supposed to be 0, and for x < 0 atan2 is -π or π depending on the direction y/x approaches
+      // 0.
+      // Because implementations will be be calculating this via numeric approximations that will vary slightly around
+      // 0 and subnormals, it is very difficult to write a rule for accuracy here without just dictating a specific
+      // implementation. Instead if the value is at y/x = 0, then the result should be in [-π, π], plus 4096 ULP from
+      // atan.
+      return F32Interval.span(
+        ulpInterval(kValue.f32.negative.pi.whole, 4096),
+        ulpInterval(kValue.f32.positive.pi.whole, 4096)
+      );
+    }
+
+    const atan_yx = atanInterval(divisionInterval(y, x));
+    // x > 0, atan(y/x)
+    if (x > 0) {
+      return atan_yx;
+    }
+
+    // x < 0, y > 0, atan(y/x) + π
+    if (y > 0) {
+      return additionInterval(atan_yx, kValue.f32.positive.pi.whole);
+    }
+
+    // x < 0, y < 0, atan(y/x) - π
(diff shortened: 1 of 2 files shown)


ben-clayton (commented):
LGTM, but due to the subtleties around all of this, I'd like final approval from @dneto0.

src/webgpu/util/f32_interval.ts
```diff
@@ -746,26 +746,47 @@ const AtanIntervalOp: PointToIntervalOp = {
...
-          ulpInterval(kValue.f32.positive.pi.whole, numULP)
-        );
-      }
+      // For x > 0 atan2 is supposed to be 0, and for x < 0 atan2 is -π or π depending on the direction y/x approaches
+      // 0.
+      // Because implementations will be be calculating this via numeric approximations that will vary slightly around
+      // 0 and subnormals, it is very difficult to write a rule for accuracy here without just dictating a specific
+      // implementation. Instead if the value is at y/x = 0, then the result should be in [-π, π], plus 4096 ULP from
+      // atan.
+      return F32Interval.span(
```
dneto0, line 768: From the IEEE rules:

* If y is ±0 and x is negative or -0, ±π is returned
* If y is ±0 and x is positive or +0, ±0 is returned

Can we follow that here?   If  y === 0,  and X >0, then expect 0  within 4096 ULPs

Graphically, this follows the middle curve in this graph https://en.wikipedia.org/wiki/Atan2#/media/File:Arctangent2.svg to say that when x >0 then as y/x approaches zero, atan2(y,x) smoothly approaches 0.


Or do you already have evidence that drivers are erratic for  y == 0  and X> 0

src/webgpu/util/f32_interval.ts
```diff
@@ -746,26 +746,47 @@ const AtanIntervalOp: PointToIntervalOp = {
...
+        ulpInterval(kValue.f32.negative.pi.whole, 4096),
+        ulpInterval(kValue.f32.positive.pi.whole, 4096)
+      );
     }
-    return ulpInterval(Math.atan2(y, x), numULP);
+
+    const atan_yx = atanInterval(divisionInterval(y, x));
```
dneto0, line 774: These parts look fine.

dneto0 (changes requested):

src/webgpu/util/f32_interval.ts
```diff
@@ -746,26 +746,47 @@ const AtanIntervalOp: PointToIntervalOp = {
...
+export function atanInterval(n: number | F32Interval): F32Interval {
   return runPointOp(toF32Interval(n), AtanIntervalOp);
 }
 
 const Atan2IntervalOp: BinaryToIntervalOp = {
   impl: (y: number, x: number): F32Interval => {
-    const numULP = 4096;
+    if (x === 0) {
+      // atan2 is not defined meaningfully here
+      return F32Interval.any();
```
dneto0, line 757: DXC and cppreference both say that

x = 0   y <0 should yield  -pi/2
x = 0  y  > 0 should yield pi / 2

src/webgpu/util/f32_interval.ts
zoddicus: @amaiorano I believe there were Windows drivers that were returning ±π in unexpected cases, but I don't recall the details.  Do you recall?

Those cases may be covered by `If y is ±0 and x is negative or -0, ±π is returned`, so I will implement what you have sugggested and we can see if it is breaking anywhere.

zoddicus: Specifically this is becoming an issue when y is subnormal. Probably something related to y = 0 causing y/x = 0 and that is hard coded or something in the driver.

For example:
    (-1.1754942106924411e-38 (0x807fffff subnormal), 12.734692573547363 (0x414bc14d))
        returned: 3.1415927410125732 (0x40490fdb)
        expected: [-4.81482486096809e-35,4.81482486096809e-35]

zoddicus: Hmmm.... it appears that π is being returned unexpected when y is a negative subnormal, so maybe they are just doing -0 wrong?

src/webgpu/util/f32_interval.ts
zoddicus: I agree those specific implementations do that, but if we are supporting other implementations where atan2 ~ atan(y/x) then those others one will get an undefined value when x = 0, and that would be more permissive than pi/2 or -pi/2.

I should probably rewrite the comment to be clear that it is in the atan(y/x) case that this is undefined.

src/webgpu/util/f32_interval.ts
dneto0: Ok, it's very clarifying that it's the "y is subnormal" case.  The function is wildly approaching discontinuity from -pi to +pi as y transitions from small negative to small positive.
(Stare at the colour diagram at https://en.cppreference.com/w/cpp/numeric/math/atan2 )

Allowing for small errors in calculation it's not particularly meaningful to ask about the quadrant of a point near the origin.

Suggest we update the spec to say the accuracy of atan2 at most when y is normal.

zoddicus:
This needs to be updated based on https://github.com/gpuweb/gpuweb/pull/3474

zoddicus:
I have updated this PR to match the rules from https://github.com/gpuweb/gpuweb/pull/3474

dneto0 (approved):
Thanks for your patience.

I filed a Vulkan CTS issue internally about the fact that the y-is-subnormal case having out-of-spec results.  I pointed out your evidence for the conformant NVIDIA driver.
Vulkan CTS issue: #4022 (internal to Khronos).

Thanks!

Issue #1230, opened 2022-04-25
wgsl: f32 execution tests for `normalize` builtin


dj2:
Add f32 execution tests for the `normalize` builtin.
```
@const @must_use fn normalize(e: vecN<T> ) -> vecN<T>

T is f32
```
Returns a unit vector in the same direction as e.

Accuracy is `Inherited from x / length(x)`

See https://www.w3.org/TR/WGSL/#normalize-builtin

zoddicus:
Completed f32 work, moving [AbstractFloat](https://github.com/gpuweb/cts/issues/2570) and [f16](https://github.com/gpuweb/cts/issues/2571) to their own issues

PR #1841, opened 2022-09-14
Implement `normalize` f32 tests

zoddicus:
Adds support for `F32Interval[]` in the `isF32Vector` and `toF32Vector` functions.

Adds support for VectorToVector operations

Adds plumbing for Vector of floats comparisons

Also includes a renaming for some existing functions to include the `ToFoo` part to distinguish them.

Issue #1230

<hr>

### DIFF
--- a/src/unittests/f32_interval.spec.ts
+++ b/src/unittests/f32_interval.spec.ts
@@ -41,6 +41,7 @@ import {
   mixPreciseInterval,
   multiplicationInterval,
   negationInterval,
+  normalizeInterval,
   powInterval,
   quantizeToF16Interval,
   radiansInterval,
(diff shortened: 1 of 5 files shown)


src/unittests/f32_interval.spec.ts
```diff
@@ -2607,3 +2608,45 @@ g.test('dotInterval')
...
+  input: number[];
+  expected: IntervalBounds[];
+}
+
+g.test('normalizeInterval')
+  .paramsSubcasesOnly<VectorToVectorCase>(
+    // prettier-ignore
+    [
+      // vec2
+      {input: [1.0, 0.0], expected: [[hexToF64(0x3feffffe, 0x70000000), hexToF64(0x3ff00000, 0xb0000000)], [hexToF32(0x81200000), hexToF32(0x01200000)]] },  // [ ~1.0, ~0.0]
```
ben-clayton, line 2622: Bit confused by the `~0.0` here. Isn't `0 / y` always `0` for any finite `y`?

The WGSL spec says for `x / y` and f32:

> 2.5 ULP for `|y|` in the range <code>[2<sup>-126</sup>, 2<sup>126</sup>]</code>

So if `x` is zero, we get zero, no?

zoddicus: Unfortunately no :-(

y = length of the vector is in range, so it should have 2.5 ULP error

so 0/y does equal 0, but the acceptance interval is [0 - 2.5 * ULP(0), 0 + 2.5 * ULP(0)]
ULP(0) is larger then you would expect, because a platform may FTZ, so the nearest value is max negative or min positive, not the nearest subnormal.

ben-clayton: Ack. Thanks :)

PR #1842, opened 2022-09-14
Overrides: add operation tests for compute pipeline

shrekshao:
Issue: None

<hr>


src/webgpu/api/operation/compute/overrides.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+    expected: Uint32Array | Float32Array,
+    constants: Record<string, GPUPipelineConstantValue>,
+    code: string
+  ) {
+    const dst = this.device.createBuffer({
+      size: expected.byteLength,
+      usage: GPUBufferUsage.COPY_SRC | GPUBufferUsage.STORAGE,
+    });
+
+    const pipeline = this.device.createComputePipeline({
```
kainino0x, line 20: Would be good to add isAsync variants

src/webgpu/api/operation/compute/overrides.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+        }
+      `
+    );
+  });
+
+g.test('precision')
+  .desc(`Test that float number precision is preserved for constants.`)
+  .fn(async t => {
+    const c1 = 3.14159;
+    const c2 = 3.141592653589793238;
```
kainino0x, line 139: These values will get rounded to f32 by both `new Float32Array` and `createComputePipeline`. This is fine, but leave a comment about it since the values coming out of the shader won't actually match c1 and c2.

src/webgpu/api/operation/compute/overrides.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+  )
+  .fn(async t => {
+    const { x } = t.params;
+    t.ExpectShaderOutputWithConstants(
+      new Uint32Array([x]),
+      {
+        x,
+      },
+      `
+        override x: u32;
```
kainino0x, line 178: perhaps parameterize this a bit to test both i32 and u32, as well as y and z? Could be like

`override size: ${type}`
...
`@workgroup_size(${ /* 'size' or '1,size' or '1,1,size' */ })`

src/webgpu/api/operation/compute/overrides.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+                buf.data[0] = local_invocation_id.x + 1;
+            }
+        }
+      `
+    );
+  });
+
+g.test('shared_shader_module')
+  .desc(
+    `Test that when the same shader module is shared by different pipelines, the correct constant values are used.`
```
kainino0x, line 199: side note: when we do operation tests for render pipelines, would be good to have an override that's used by both the vertex and fragment entry points and override it to a different value in each stage. Should put this in a TODO somewhere or in a github issue to track it.

kainino0x (approved):

src/webgpu/api/operation/compute/overrides.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+    this.device.queue.submit([encoder.finish()]);
+
+    this.expectGPUBufferValuesEqual(dst, expected);
+  }
+}
+
+export const g = makeTestGroup(F);
+
+g.test('basic')
+  .desc(`Test that correct constants override values or default values are used.`)
```
kainino0x, line 51: I think all of the descriptions should have a little more detail on how the test works, so a reader can understand the coverage of the test better by reading the description.

src/webgpu/api/operation/compute/overrides.spec.ts
shrekshao: Add a `api/operation/render_pipeline/overrides.spec.ts` placeholder.

kainino0x (approved):

Issue #897, opened 2022-01-13
op: memory_sync,buffer,ww:multiple_buffers


kainino0x:

Gyuyoung:
It's likely this problem has been done.

PR #1857, opened 2022-09-22
op: Implement  test in multiple_buffers.spec.ts

Gyuyoung:
This PR implements `rw` test in multiple_buffers.spec.ts to ensure that a boundary can separate operations
in multiple buffers.
    
Issue: #897

<hr>

### DIFF
--- a/src/webgpu/api/operation/memory_sync/buffer/multiple_buffers.spec.ts
+++ b/src/webgpu/api/operation/memory_sync/buffer/multiple_buffers.spec.ts
@@ -16,8 +16,23 @@ TODO: Tests with more than one buffer to try to stress implementations a little 
 `;
 
 import { makeTestGroup } from '../../../../../common/framework/test_group.js';
+import {
+  kOperationBoundaries,
+  kBoundaryInfo,
+  OperationContextHelper,
+} from '../operation_context_helper.js';
 
-import { BufferSyncTest } from './buffer_sync_test.js';
+import {
+  kAllReadOps,
+  kAllWriteOps,
+  BufferSyncTest,
+  checkOpsValidForContext,
+} from './buffer_sync_test.js';
+
+// The src value is what stores in the src buffer before any operation.
+const kSrcValue = 0;
+// The op value is what the read/write operation write into the target buffer.
+const kOpValue = 1;
 
 export const g = makeTestGroup(BufferSyncTest);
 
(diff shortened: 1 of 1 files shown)


src/webgpu/api/operation/memory_sync/buffer/multiple_buffers.spec.ts
```diff
@@ -27,9 +42,57 @@ g.test('rw')
...
+
+      // The read op will read from src buffer and write to dst buffer based on what it reads.
+      // The write op will write the given op value into src buffer as well.
+      // The write op happens after read op. So we are expecting the src value to be in the dst
+      // buffer.
+      t.encodeReadOp(helper, readOp, readContext, srcBuffers[i], dstBuffers[i]);
+      helper.ensureBoundary(boundary);
+    }
+
+    t.encodeWriteOp(helper, writeOp, writeContext, srcBuffers[0], 0, kOpValue);
```
Gyuyoung, line 91: @kainino0x It seems to me that unimplemented tests in the multiple_buffers.spec.ts can be implemented by referring to single_buffer.spec.ts. So, I'm trying to implement `rw` test first.  But, I'm not sure if I understand the description of this test correctly. I understood that this test should perform a read operation on multiple buffers, then perform a write operation on each buffer, right? It's likely I don't find the right way to write this test. Any comment?

kainino0x: re: https://github.com/gpuweb/cts/pull/1857

The `memory_sync` tests are very difficult because they attempt to find implementations in memory synchronization primitives, like barriers issued to Vulkan to ensure a write completes before a read starts. This is the reason they've sat unimplemented for so long. Fortunately, I we have already written most of the `memory_sync,buffer` tests, and *think* the idea here was pretty straightforward - just add some complexity relative to the `single_buffer` tests to make the usage a little more robust, for example, to make sure barriers are placed on the correct resources.

I'll leave a separate comment about the test content.

BTW, the issue for this is https://github.com/gpuweb/cts/issues/897 (found by searching memory-sync in the project tracker: https://github.com/orgs/gpuweb/projects/3/views/1?filterQuery=is%3Aissue+-status%3AComplete%2CBlocked%2CStarted%2CPost-V1+memory_sync), so if you'll be working on this please assign yourself and mark it as started. This would have also given me a chance to look at the test and make sure it wasn't going to be too gnarly. For example, the `memory_sync,texture` tests have several extra layers of complexity over the `memory_sync,buffer` tests, and I want someone with some experience with Vulkan synchronization to write them (which excludes me, because I don't have much experience with it either).

src/webgpu/api/operation/memory_sync/buffer/multiple_buffers.spec.ts
```diff
@@ -27,9 +42,56 @@ g.test('rw')
...
+      await t.createIntermediateBuffersAndTexturesForWriteOp(writeOp, 0, kOpValue);
+
+      // The read op will read from src buffer and write to dst buffer based on what it reads.
+      // The write op will write the given op value into src buffer as well.
+      // The write op happens after read op. So we are expecting the src value to be in the dst
+      // buffer.
+      t.encodeReadOp(helper, readOp, readContext, srcBuffers[i], dstBuffers[i]);
+      helper.ensureBoundary(boundary);
+      t.encodeWriteOp(helper, writeOp, writeContext, srcBuffers[0], 0, kOpValue);
+      helper.ensureSubmit();
```
kainino0x, line 90: (r = read, w = write, c = check)
This will result in r,w,c,r,w,c,r,w,c,r,w,c. I think the intent of the test is to do r,r,r,r,w,w,w,w,c,c,c,c so that work is more interleaved. For example, an implementation might coalesce 4 barriers (one for each r-w pair) into 1 barrier to cover all 4, and we want to make sure that optimization is correct.

src/webgpu/api/operation/memory_sync/buffer/multiple_buffers.spec.ts
Gyuyoung: Thanks for explaining the `memory_sync` buffer tests. If this PR is ready to get reviews, let me take the issue(#897) because I was not sure if I can make this PR ready to get reviewed.

src/webgpu/api/operation/memory_sync/buffer/multiple_buffers.spec.ts
Gyuyoung: Thanks for the comment. So I update this PR to do `r,r,r,r -> a boundary -> w,w,w,w -> c,c,c,c` to ensure a boundary can separate operations in multiple buffers. Does it make sense?

kainino0x (approved):

src/webgpu/api/operation/memory_sync/buffer/multiple_buffers.spec.ts
kainino0x: Yes, lgtm!

PR #1860, opened 2022-09-22
webgpu/shader/execution: Reduce `const` case batch size

ben-clayton:
Reduce the batch size to avoid timeouts with Chromium's CTS runner.

We trigger a heartbeat between each batch, so by increasing the granularity of these batches, we should avoid whole-test timeouts.

Issue: #

<hr>

### DIFF
--- a/src/webgpu/shader/execution/expression/expression.ts
+++ b/src/webgpu/shader/execution/expression/expression.ts
@@ -162,7 +162,7 @@ export function run(
   const casesPerBatch = (function () {
     switch (cfg.inputSource) {
       case 'const':
-        return 256; // Arbitrary limit, to ensure shaders aren't too large
+        return 64; // Arbitrary limit, to ensure shaders aren't too large
       case 'uniform':
         return Math.floor(
           t.device.limits.maxUniformBufferBindingSize / (parameterTypes.length * kValueStride)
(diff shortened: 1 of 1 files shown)


src/webgpu/shader/execution/expression/expression.ts
```diff
@@ -162,7 +162,7 @@ export function run(
   const casesPerBatch = (function () {
     switch (cfg.inputSource) {
       case 'const':
-        return 256; // Arbitrary limit, to ensure shaders aren't too large
+        return 64; // Arbitrary limit, to ensure shaders aren't too large
```
austinEng, line 165: github won't let me comment lower down.

Looking at this again, we do:
```
pushErrorScope('validation');
submit all the batches
popErrorScope()
checkResults.forEach(f => f());
```
maybe instead we should do:
```
promises = [];
for each batch:
  pushErrorScope
  checkBatch = submitBatch
  promises.push(popErrorScope().then(checkBatch));
  
await Promise.all(promises);
```

the reason is that waiting for popErrorScope after ALL the batches will wait for them all to finish validating

austinEng (approved):
LGTM if you want to land this; but not sure if it'll actually fix the problem

also, consider carving out these tests in Chrome's harness to run without parallelization to make things more deterministic

src/webgpu/shader/execution/expression/expression.ts
austinEng: **wait for them all to run through validation in Dawn. the `f` that's returned by `submitBatch` will still run one-by-one with heartbeats in between in the current code.

ben-clayton: Done

PR #1877, opened 2022-09-29
Implement 'timestampWrites,invalid_query_set' test in beginComputePass.spec.ts

Gyuyoung:
~beginComputePass method should generate a validation error if device.feature doesn't contain "timestamp-query".~

Issue: #1743

<hr>

### DIFF
--- a/src/webgpu/api/validation/encoding/beginComputePass.spec.ts
+++ b/src/webgpu/api/validation/encoding/beginComputePass.spec.ts
@@ -106,6 +106,33 @@ g.test('timestampWrites,query_set_type')
     t.tryComputePass(isValid, descriptor);
   });
 
+g.test('timestampWrites,invalid_query_set')
+  .desc(`Tests that timestampWrite that has an invalid query set generates a validation error.`)
+  .params(u => u.combine('querySetState', ['valid', 'invalid'] as const))
+  .beforeAllSubcases(t => {
+    t.selectDeviceOrSkipTestCase(['timestamp-query']);
+  })
+  .fn(async t => {
+    const { querySetState } = t.params;
+
+    const querySet = t.createQuerySetWithState(querySetState, {
+      type: 'timestamp',
+      count: 1,
+    });
+
+    const timestampWrite = {
+      querySet,
+      queryIndex: 0,
+      location: 'beginning' as const,
+    };
+
+    const descriptor = {
+      timestampWrites: [timestampWrite],
+    };
+
+    t.tryComputePass(querySetState === 'valid', descriptor);
+  });
+
 g.test('timestampWrites,query_index_count')
   .desc(`Test that querySet.count should be greater than timestampWrite.queryIndex.`)
   .params(u => u.combine('queryIndex', [0, 1, 2, 3]))


src/webgpu/api/validation/encoding/beginComputePass.spec.ts
```diff
@@ -106,6 +106,40 @@ g.test('timestampWrites,query_set_type')
...
+  .params(u => u.combine('featureContainsTimestampQuery', [false, true]))
+  .beforeAllSubcases(t => {
+    const { featureContainsTimestampQuery } = t.params;
+
+    const requiredFeatures: GPUFeatureName[] = [];
+    if (featureContainsTimestampQuery) {
+      requiredFeatures.push('timestamp-query');
+    }
+
+    t.selectDeviceOrSkipTestCase({ requiredFeatures });
```
kainino0x, line 125: nit: This function has an overload so you can use simply
```ts
t.selectDeviceOrSkipTestCase(featureContainsTimestampQuery ? 'timestamp-query' : undefined);
```

src/webgpu/api/validation/encoding/beginComputePass.spec.ts
```diff
@@ -106,6 +106,40 @@ g.test('timestampWrites,query_set_type')
     t.tryComputePass(isValid, descriptor);
   });
 
+g.test('timestamp_query_feature')
+  .desc(
+    `
+  Test that a validation error is generated if the features doesn't contain 'timestamp-query'
+  feature.
```
kainino0x, line 113: nit: It's easier to read language which describes the intent to enable the feature, rather than the mechanism used to enable the feature.

```suggestion
  Test that a validation error if the 'timestamp-query' feature isn't enabled.
```

kainino0x, line 113: Description also needs to describe what call is generating the error; see below.

src/webgpu/api/validation/encoding/beginComputePass.spec.ts
```diff
@@ -106,6 +106,40 @@ g.test('timestampWrites,query_set_type')
...
+      requiredFeatures.push('timestamp-query');
+    }
+
+    t.selectDeviceOrSkipTestCase({ requiredFeatures });
+  })
+  .fn(async t => {
+    const { featureContainsTimestampQuery } = t.params;
+
+    const timestampWrite = {
+      querySet: t.device.createQuerySet({ type: 'timestamp', count: 1 }),
```
kainino0x, line 131: This `createQuerySet` call is supposed to fail if the feature timestamp-query feature is not enabled. That is already tested, in `api/validation/capability_checks/features/query_types.spec.ts`

Since it's impossible to create a timestamp query set without the feature, it's only possible to test `beginComputePass` with:
- `timestampWrites` containing a query set of the wrong type (tested in `timestampWrites,query_set_type`)
- `timestampWrites` containing an invalid query set (not sure if we have a test for this)

kainino0x (commented):
I don't think it's possible to test what you wanted to test here. However, there's a related thing we need to test. See comments.

src/webgpu/api/validation/encoding/beginComputePass.spec.ts
Gyuyoung: Ah, right. It's likely I missed checking them. Now I'm thinking that we've not tested the case that timestampWrites contain an invalid query set in calling of begingComputePass yet. So I update the test to check it. wdyt?

kainino0x (approved):
lgtm!

Issue #1891, opened 2022-10-04
Check the validation of getBindGroupLayout method


Gyuyoung:
SSIA.

Gyuyoung:
I'm not fully sure if we need to do something more about this issue. But, It's likely we've done most of things for this issue. So, I would like to mark this issue as 'complete' for now.

PR #1892, opened 2022-10-04
Implement 'index_range' test in getBindGroupLayout.spec.ts

Gyuyoung:
According to the specification, the index of getBindGroupLayout should not exceed the size of the bind group layouts. So this PR implements a new test to ensure it.

Issue: #1891

<hr>

### DIFF
--- a/src/webgpu/api/validation/getBindGroupLayout.spec.ts
+++ b/src/webgpu/api/validation/getBindGroupLayout.spec.ts
@@ -1,0 +1,70 @@
+export const description = `
+  getBindGroupLayout validation tests.
+`;
+
+import { makeTestGroup } from '../../../common/framework/test_group.js';
+
+import { ValidationTest } from './validation_test.js';
+
+export const g = makeTestGroup(ValidationTest);
+
+g.test('index_range,explicit_layout')
+  .desc(
+    `
+  Test that a validation error is generated if the index exceeds the size of the bind group layouts
+  using a pipeline with an explicit layout.
+  `
+  )
+  .params(u => u.combine('index', [0, 1, 2, 3, 4, 5]))
+  .fn(async t => {
+    const { index } = t.params;
+
+    const pipelineBindGroupLayouts = t.device.createBindGroupLayout({
+      entries: [],
+    });
+
+    const kBindGroupLayoutsSizeInPipelineLayout = 1;
+    const pipelineLayout = t.device.createPipelineLayout({
+      bindGroupLayouts: [pipelineBindGroupLayouts],
+    });
+
+    const pipeline = t.device.createRenderPipeline({
+      layout: pipelineLayout,
+      vertex: {
+        module: t.device.createShaderModule({
+          code: `
+            @vertex
+            fn main()-> @builtin(position) vec4<f32> {
+              return vec4<f32>(0.0, 0.0, 0.0, 1.0);
+            }`,
+        }),
+        entryPoint: 'main',
+      },
+      fragment: {
+        module: t.device.createShaderModule({
+          code: `
+            @fragment
+            fn main() -> @location(0) vec4<f32> {
+              return vec4<f32>(0.0, 1.0, 0.0, 1.0);
+            }`,
+        }),
+        entryPoint: 'main',
+        targets: [{ format: 'rgba8unorm' }],
+      },
+    });
+
+    const shouldError = index >= kBindGroupLayoutsSizeInPipelineLayout;
+
+    t.expectValidationError(() => {
+      pipeline.getBindGroupLayout(index);
+    }, shouldError);
+  });
+
+g.test('index_range,auto_layout')
+  .desc(
+    `
+  Test that a validation error is generated if the index exceeds the size of the bind group layouts
+  using a pipeline with an auto layout.
+  `
+  )
+  .unimplemented();


src/webgpu/api/validation/getBindGroupLayout.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+    const pipelineBindGroupLayouts = t.device.createBindGroupLayout({
+      entries: pipelineBufferBindings,
+    });
+
+    const pipelineLayout = t.device.createPipelineLayout({
+      bindGroupLayouts: [pipelineBindGroupLayouts],
+    });
+
+    const pipeline = t.device.createRenderPipeline({
+      layout: pipelineLayout,
```
kainino0x, line 41: This should be tested with both explicit layouts (as you have here) and the `"auto"` layout, because getBindGroupLayout would be likely to behave differently in both cases.

They'll be substantially different, so most likely two separate tests. The test with an `"auto"` layout would need bindings defined in its shaders. That can be a followup, we should just add an `unimplemented()` test now to track it.

src/webgpu/api/validation/getBindGroupLayout.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+        }),
+        entryPoint: 'main',
+        targets: [{ format: 'rgba8unorm' }],
+      },
+    });
+
+    const shouldError = index >= maxBindGroups;
+
+    t.expectValidationError(() => {
+      pipeline.getBindGroupLayout(index);
```
kainino0x, line 68: According to the spec, at least as currently written, this is supposed to throw an exception if `index >= maxBindGroups`. It's only a validation error if it's `< maxBindGroups` but `>=` the number of bind groups actually in the layout.

HOWEVER: I actually think we should change this in the spec. It doesn't seem consistent with other parts of the API. I'm going to open a spec proposal about it. Let's see if that gets resolved quickly.

Note, there's also an issue where `"auto"` layouts always fill the array up to `maxBindGroups`, instead of only up to the highest bind group index needed. I think this is a bug in the spec, so ignore this for now.

kainino0x (commented):
I have a few minor comments, but this also reminded me about some issues in the spec. Those will need to get resolved before we can complete this. See below for details, and sorry for the blocker!

src/webgpu/api/validation/getBindGroupLayout.spec.ts
kainino0x: https://github.com/gpuweb/gpuweb/pull/3505

https://github.com/gpuweb/gpuweb/pull/3506

Hoping both can be resolved tomorrow.

src/webgpu/api/validation/getBindGroupLayout.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+  .desc(
+    `
+  Test that a validation error is generated if the index exceeds the size of the bind group layouts.
+  `
+  )
+  .params(u => u.combine('index', [0, 1, 2, 3, 4, 5]))
+  .fn(async t => {
+    const { index } = t.params;
+
+    const maxBindGroups = t.device.limits?.maxBindGroups ?? 4;
```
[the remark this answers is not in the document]
Gyuyoung: Ah, I just copied & pasted it from the existing code. Fixed.

> 
Can you please update the other instance of t.device.limits?.maxBindGroups, in setBindGroup.spec.ts, too?

Sure, let me upload a new PR to update it.

src/webgpu/api/validation/getBindGroupLayout.spec.ts
Gyuyoung: Thanks for checking it. Once let me update this PR with some fixes you pointed out.

kainino0x: There have been no objections so far, so let's move forward assuming they'll land.

src/webgpu/api/validation/getBindGroupLayout.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+  `
+  )
+  .params(u => u.combine('index', [0, 1, 2, 3, 4, 5]))
+  .fn(async t => {
+    const { index } = t.params;
+
+    const maxBindGroups = t.device.limits.maxBindGroups;
+
+    const pipelineBufferBindings: GPUBindGroupLayoutEntry[] = [];
+    for (let binding = 0; binding < maxBindGroups; binding++) {
```
kainino0x, line 25: This `binding` is bindings within a bind group, not number of bind groups. So it doesn't make sense to go up to `maxBindGroups`. In fact, we don't really need any entries in the bind group layout. I think `entries` can just be `[]`.

src/webgpu/api/validation/getBindGroupLayout.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+            fn main() -> @location(0) vec4<f32> {
+              return vec4<f32>(0.0, 1.0, 0.0, 1.0);
+            }`,
+        }),
+        entryPoint: 'main',
+        targets: [{ format: 'rgba8unorm' }],
+      },
+    });
+
+    const shouldError = index >= maxBindGroups;
```
kainino0x, line 66: With #3506, this should error if index >= 1, since 1 is the number of bind group layouts in `pipelineLayout`.

kainino0x (commented):
LGTM with some small changes

src/webgpu/api/validation/getBindGroupLayout.spec.ts
Gyuyoung: yes, `pipelineLayout` only has `pipelineBindGroupLayouts` as for `bindGroupLayouts`. Thanks for correcting me!

src/webgpu/api/validation/getBindGroupLayout.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+export const g = makeTestGroup(ValidationTest);
+
+g.test('index_range,explicit_layout')
+  .desc(
+    `
+  Test that a validation error is generated if the index exceeds the size of the bind group layouts
+  using a pipeline with an explicit layout.
+  `
+  )
+  .params(u => u.combine('index', [0, 1, 2, 3, 4, 5]))
```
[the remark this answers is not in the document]
kainino0x: Thanks. I've filed https://bugs.chromium.org/p/dawn/issues/detail?id=1565 about it. We'll also see that the test is failing :)

kainino0x (approved):

Issue #1217, opened 2022-04-25
wgsl: f32 execution tests for `distance` builtin


dj2:
Add f32 execution tests for the `distance` builtin.
```
@const @must_use fn distance(e1: T,
                             e2: T) -> S

S is f32
T is S or vecN<S>
```
Returns the distance between e1 and e2 (e.g. length(e1 - e2)).

Accuracy is `Inherited from length(x - y)`

See https://www.w3.org/TR/WGSL/#distance-builtin

zoddicus:
f32 work completed, splitting [AbstractFloat](https://github.com/gpuweb/cts/issues/2556) and [f16](https://github.com/gpuweb/cts/issues/2557) off into their own issues.

PR #1894, opened 2022-10-04
Implement `distance` f32 tests

zoddicus:
This covers both scalar and vector inputs.

This also include an improvement for addFlushIfNeeded, which was flushing even when the value was 0. This was causing a lot of duplicate branches to be run for inputs. It doesn't functionally change the results, but significantly speeds them up.

Issue #1217

<hr>

### DIFF
--- a/src/unittests/f32_interval.spec.ts
+++ b/src/unittests/f32_interval.spec.ts
@@ -22,6 +22,7 @@ import {
   cosInterval,
   coshInterval,
   degreesInterval,
+  distanceInterval,
   divisionInterval,
   dotInterval,
   expInterval,
@@ -1087,7 +1088,9 @@ g.test('lengthIntervalScalar')
     [
       // Some of these are hard coded, since the error intervals are difficult to express in a closed human readable
       // form due to the inherited nature of the errors.
-      {input: 0, expected: kAny },  // ~0
+      //
+      // length(0) = kAny, because length uses sqrt, which is defined as 1/inversesqrt
+      {input: 0, expected: kAny },
       {input: 1.0, expected: [hexToF64(0x3fefffff, 0x70000000), hexToF64(0x3ff00000, 0x90000000)] },  // ~1
       {input: -1.0, expected: [hexToF64(0x3fefffff, 0x70000000), hexToF64(0x3ff00000, 0x90000000)] },  // ~1
       {input: 0.1, expected: [hexToF64(0x3fb99998, 0x90000000), hexToF64(0x3fb9999a, 0x70000000)] },  // ~0.1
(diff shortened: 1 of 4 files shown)


src/webgpu/util/f32_interval.ts
```diff
@@ -1631,15 +1701,9 @@ export function stepInterval(edge: number, x: number): F32Interval {
...
 
-const SubtractionInnerOp: BinaryToIntervalOp = {
-  impl: (x: number, y: number): F32Interval => {
-    return correctlyRoundedInterval(x - y);
-  },
-};
-
 const SubtractionIntervalOp: BinaryToIntervalOp = {
   impl: (x: number, y: number): F32Interval => {
-    return roundAndFlushBinaryToInterval(x, y, SubtractionInnerOp);
```
zoddicus, line 1642: This is a non-functional change, the inner op for subtraction is a remnant from before the framework was more built out.

src/webgpu/util/f32_interval.ts
```diff
@@ -1260,18 +1324,24 @@ export function ldexpInterval(e1: number, e2: number): F32Interval {
...
 };
 
 /** Calculate an acceptance interval of length(x) */
-export function lengthInterval(n: number | number[]): F32Interval {
+export function lengthInterval(n: number | F32Interval | number[] | F32Vector): F32Interval {
   if (n instanceof Array) {
-    return runVectorToIntervalOp(toF32Vector(n), LengthIntervalOp);
+    return runVectorToIntervalOp(toF32Vector(n), LengthIntervalVectorOp);
   } else {
-    return sqrtInterval(multiplicationInterval(n, n));
```
zoddicus, line 1274: This is just a small style cleanup.

dj2 (approved):

src/unittests/f32_interval.spec.ts
```diff
@@ -1087,7 +1088,9 @@ g.test('lengthIntervalScalar')
     [
       // Some of these are hard coded, since the error intervals are difficult to express in a closed human readable
       // form due to the inherited nature of the errors.
-      {input: 0, expected: kAny },  // ~0
+      //
+      // length(0) = kAny, because length uses sqrt, which is defined as 1/inversesqrt
```
ben-clayton, line 1092: I find the fact that `length(0)` isn't always `0` really surprising. Is this right @dneto0 ?

ben-clayton (approved):
LGTM, although I find `length(0)==any` really suspicious.

src/unittests/f32_interval.spec.ts
zoddicus: I agree it looks weird, I actually spent a non-trivial amount of time investigating this when I start writing this PR to make sure it was correct, I thought it was a bug in my code originally.

My understanding is that this is a consequence of the inherited accuracy.

length's accuracy is defined as `sqrt(x*x)`
square root's accuracy is defined as `1.0 / inverseSqrt(x)`
inverse square root's accuracy is defined as 2 ULP from truth, which is stated as `Returns the reciprocal of sqrt(e)`

So at x = 0 you have at the bottom of the inherited accuracy 2 ULP from 1/sqrt(0*0) which becomes 1/0, which is an undefined value, which is where `kAny` comes in and it propagates from there.

@dneto0 can you confirm that this is the intended interpretation? or correct me, since this affects a number of built-ins.

dneto0: Yes, that's the right interpretation.

Issue #1206, opened 2022-04-25
wgsl: f32 execution tests for `faceForward` builtin


dj2:
Add f32 execution tests for the `faceForward` builtin.
```
@const @must_use fn faceForward(e1: T,
                                e2: T,
                                e3: T) -> T
T is vecN<f32>
```
Returns e1 if dot(e2, e3) is negative, and -e1 otherwise.

Accuracy is `Inherited from select(-x, x, dot(z, y) < 0.0)`

[See https://www.w3.org/TR/WGSL/#exp2-builtin](https://www.w3.org/TR/WGSL/#faceForward-builtin)

zoddicus:
f32 work is done, splitting [AbstractFloat](https://github.com/gpuweb/cts/issues/2546) and [f16](https://github.com/gpuweb/cts/issues/2547) off into separate issues.

PR #1900, opened 2022-10-06
Implement `faceForward` f32 tests

zoddicus:
Issue #1206

<hr>

### DIFF
--- a/src/unittests/f32_interval.spec.ts
+++ b/src/unittests/f32_interval.spec.ts
@@ -23,6 +23,7 @@ import {
   correctlyRoundedInterval,
   cosInterval,
   coshInterval,
+  crossInterval,
   degreesInterval,
   distanceInterval,
   divisionInterval,
@@ -30,6 +31,7 @@ import {
   expInterval,
   exp2Interval,
   F32Interval,
+  faceForwardIntervals,
   floorInterval,
   fractInterval,
   IntervalBounds,
@@ -48,6 +50,7 @@ import {
   powInterval,
   quantizeToF16Interval,
   radiansInterval,
+  reflectInterval,
   remainderInterval,
   roundInterval,
   saturateInterval,
@@ -60,11 +63,9 @@ import {
   subtractionInterval,
   tanInterval,
   tanhInterval,
+  toF32Vector,
   truncInterval,
   ulpInterval,
-  crossInterval,
-  toF32Vector,
-  reflectInterval,
 } from '../webgpu/util/f32_interval.js';
 import { hexToF32, hexToF64, oneULP } from '../webgpu/util/math.js';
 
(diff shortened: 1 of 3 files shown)


zoddicus:
This PR depends on https://github.com/gpuweb/cts/pull/1894, so the only new code is in https://github.com/gpuweb/cts/pull/1900/commits/6400a381d8d7fde548e67fa63d36eaed34d05f09

dj2 (approved):
lgtm (which hopefully shows up, `an inherited accuracy`)

src/unittests/f32_interval.spec.ts
```diff
@@ -2817,3 +2818,69 @@ g.test('crossInterval')
...
+}
+
+g.test('faceForwardIntervals')
+  .paramsSubcasesOnly<FaceForwardCase>(
+    // prettier-ignore
+    [
+      // vec2
+      { input: [[1.0, 0.0], [1.0, 0.0], [1.0, 0.0]], expected: [[[-1.0], [0.0]]] },
+      { input: [[-1.0, 0.0], [1.0, 0.0], [1.0, 0.0]], expected: [[[1.0], [0.0]]] },
+      { input: [[1.0, 0.0], [-1.0, 1.0], [1.0, -1.0]], expected: [[[1.0], [0.0]]] },
```
ben-clayton, line 2834: Is this right? `[-1.0, 1.0]` and `[1.0, -1.0]` are at right angles to each other, so the dot is `0`.

The spec says:

> @const fn faceForward(e1: T, e2: T, e3: T) -> T
> Returns `e1` if `dot(e2, e3)` is negative, and `-e1` otherwise

As the dot is not negative, shouldn't this return `[-1.0, 0.0]`?
This is on the cusp of the fp-compare, so I wouldn't be surprised if you had to return either `-1.0` or `1.0` here.

zoddicus: They are in a straight line aren't they? [-1, 1] is the top-left quad and [1, -1] is the bottom-right.
Also `dot([-1, 1], [1, -1]) = (-1 * 1) + (1 * -1) = -2`

ben-clayton: Sorry, my bad. Right you are.

ben-clayton: Do we have any cases for the dot(v1, v2) == 0?

zoddicus: Having a 0 shouldn't be too much of an issue for normal inputs, because the calculation is using dot, which is mostly additions and multiplications, so composed of correctly rounded operations, so should be precise.

The last case, the one for subnormals, does have a spanning 0 situation, and demonstrates that both results are returned

ben-clayton: Just for completeness sake, please can we have a `[1.0, 0.0], [0.0, 1.0]` case?
I agree the subnormal case spans 0, but the fact they may be flushed means that these might not fully test the right-angle case.

zoddicus: Done

ben-clayton (approved):
Thanks!

PR #1911, opened 2022-10-12
Fix pack2x16float tests to correctly handle corner cases

zoddicus:
Previous version of these tests didn't correctly account for f32 and f16 subnormals.

Test generation code has been also refactored out of the test file into conversion.ts, so that unittests could be added.

<hr>

### DIFF
--- a/src/unittests/conversion.spec.ts
+++ b/src/unittests/conversion.spec.ts
@@ -236,42 +236,40 @@ expect: ${expect}`
   }
 });
 
-interface pack2x16floatCase {
-  inputs: [number, number];
-  result: (number | undefined)[];
-}
-
 g.test('pack2x16float')
-  .paramsSimple<pack2x16floatCase>([
+  .paramsSimple([
     // f16 normals
-    { inputs: [0, 0], result: [0x00000000] },
-    { inputs: [1, 0], result: [0x00003c00] },
+    { inputs: [0, 0], result: [0x00000000, 0x80000000, 0x00008000, 0x80008000] },
+    { inputs: [1, 0], result: [0x00003c00, 0x80003c00] },
     { inputs: [1, 1], result: [0x3c003c00] },
     { inputs: [-1, -1], result: [0xbc00bc00] },
-    { inputs: [10, 0], result: [0x00004900] },
-    { inputs: [-10, 0], result: [0x0000c900] },
+    { inputs: [10, 1], result: [0x3c004900] },
+    { inputs: [-10, 1], result: [0x3c00c900] },
 
     // f32 normal, but not f16 precise
-    { inputs: [1.00000011920928955078125, 0], result: [0x00003c00, 0x00003c01] },
+    { inputs: [1.00000011920928955078125, 1], result: [0x3c003c00, 0x3c003c01] },
 
     // f32 subnormals
-    { inputs: [kValue.f32.subnormal.positive.max, 0], result: [0x00000000, 0x00000001] },
     // prettier-ignore
-    { inputs: [kValue.f32.subnormal.negative.min, 0], result: [0x00008001, 0x00008000, 0x00000000] },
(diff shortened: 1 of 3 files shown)


src/unittests/conversion.spec.ts
```diff
@@ -232,3 +235,49 @@ expect: ${expect}`
...
+    // f16 subnormals
+    { inputs: [kValue.f16.subnormal.positive.max, 0], result: [0x000003ff, 0x00000000] },
+    { inputs: [kValue.f16.subnormal.negative.min, 0], result: [0x000083ff, 0x00000000] },
+
+    // f16 out of bounds
+    { inputs: [kValue.f16.positive.max + 1, 0], result: [undefined] },
+    { inputs: [kValue.f16.negative.min - 1, 0], result: [undefined] },
+    { inputs: [0, kValue.f16.positive.max + 1], result: [undefined] },
+    { inputs: [0, kValue.f16.negative.min - 1], result: [undefined] },
+  ])
```
kainino0x, line 271: nit: I believe if you use `as const` here then you can rely on type inference and not need the `interface` definition above.
```
  ] as const)
```

src/webgpu/util/conversion.ts
```diff
@@ -248,6 +255,62 @@ export function packRGB9E5UFloat(r: number, g: number, b: number): number {
...
+  // Generates all possible valid f16s from a given f32. Assumes FTZ for both the f32 and f16 value is allowed.
+  const generate_f16s = (n: number): number[] => {
+    const n_f32s = [n];
+    if (n !== 0 && isSubnormalNumberF32(n)) {
+      n_f32s.push(0);
+    }
+
+    const n_f16s = n_f32s.flatMap(correctlyRoundedF16);
+    if (!n_f16s.includes(0) && n_f16s.some(isSubnormalNumberF16)) {
+      n_f16s.push(0);
```
kainino0x, line 289: nit: Perhaps instead of having multiple places that push 0 but check if it's already there, just set a boolean flag variable and add 0 at the end if it's set?

I suspect this will be a bit faster, and this code might be a bit hot.

src/webgpu/util/conversion.ts
```diff
@@ -248,6 +255,62 @@ export function packRGB9E5UFloat(r: number, g: number, b: number): number {
...
+
+    return n_f16s;
+  };
+
+  const x_f16s = generate_f16s(x);
+  const y_f16s = generate_f16s(y);
+
+  const f16_pairs = cartesianProduct(x_f16s, y_f16s);
+  const results = new Array<number>();
+  f16_pairs.forEach(p => {
```
kainino0x, line 300: µNit: Generally we prefer `for (const p of f16_pairs)` for readability (looks more like a for loop than a map call)

src/webgpu/util/conversion.ts
```diff
@@ -248,6 +255,62 @@ export function packRGB9E5UFloat(r: number, g: number, b: number): number {
...
+    assert(p.length === 2, 'cartesianProduct of 2 arrays returned an entry with not 2 elements');
+
+    const buf = new ArrayBuffer(4);
+    const buf_f16 = new Float16Array(buf);
+    const buf_u32 = new Uint32Array(buf);
```
kainino0x, line 305: I'm unsure how hot this code is but ArrayBuffer/view allocations are notoriously expensive, so we should consider optimizing this. You can optimize it by just reusing these 3 objects for every call instead of making 3 new ones each time. We can wrap that into a small helper function that takes 2 f16s and returns one u32 and implements that optimization internally.

kainino0x (approved):

zoddicus:
@kainino0x can you take another pass on this, I had to add some significant amounts of logic to handle the fact that the bitfield of +0 and -0 are different, which actually matters here.

kainino0x (approved):

zoddicus:
@kainino0x I think this is good now, but my last round of changes was non-trivial

kainino0x (approved):

PR #1912, opened 2022-10-13
val: Implement 'pass_end_invalid_order' test

Gyuyoung:
This PR adds a new test to ensure that a validation error is generated if beginning
a second pass before ending the previous pass generates a validation error.

Issue: #1914

<hr>

### DIFF
--- a/src/webgpu/api/validation/encoding/encoder_state.spec.ts
+++ b/src/webgpu/api/validation/encoding/encoder_state.spec.ts
@@ -19,4 +19,62 @@ TODO:
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
 import { ValidationTest } from '../validation_test.js';
 
-export const g = makeTestGroup(ValidationTest);
+class F extends ValidationTest {
+  beginRenderPass(commandEncoder: GPUCommandEncoder): GPURenderPassEncoder {
+    const attachmentTexture = this.device.createTexture({
+      format: 'rgba8unorm',
+      size: { width: 16, height: 16, depthOrArrayLayers: 1 },
+      usage: GPUTextureUsage.RENDER_ATTACHMENT,
+    });
+    this.trackForCleanup(attachmentTexture);
+    return commandEncoder.beginRenderPass({
+      colorAttachments: [
+        {
+          view: attachmentTexture.createView(),
+          clearValue: { r: 1.0, g: 0.0, b: 0.0, a: 1.0 },
+          loadOp: 'clear',
+          storeOp: 'store',
+        },
+      ],
+    });
+  }
+}
+
+export const g = makeTestGroup(F);
+
+g.test('pass_end_invalid_order')
+  .desc(
+    `
+  Test that beginning a  {compute,render} pass before ending the previous {compute,render} pass
+  causes an error.
+
+  TODO: Need to add a control case to be sure a validation error happens because of ending order.
+  `
+  )
+  .paramsSubcasesOnly(u =>
+    u
+      .combine('pass0Type', ['compute', 'render'])
+      .combine('pass1Type', ['compute', 'render'])
+      .combine('endPasses', [[], [0], [1], [0, 1], [1, 0]])
+  )
+  .fn(async t => {
+    const { pass0Type, pass1Type, endPasses } = t.params;
+
+    const encoder = t.device.createCommandEncoder();
+
+    const firstPass =
+      pass0Type === 'compute' ? encoder.beginComputePass() : t.beginRenderPass(encoder);
+
+    // Begin a second pass before ending the previous pass.
+    const secondPass =
+      pass1Type === 'compute' ? encoder.beginComputePass() : t.beginRenderPass(encoder);
+
+    const passes = [firstPass, secondPass];
+    for (const index of endPasses) {
+      passes[index].end();
+    }
+
+    t.expectValidationError(() => {
+      encoder.finish();
+    }, true);
+  });


src/webgpu/api/validation/encoding/beginComputePass.spec.ts
```diff
@@ -191,3 +191,35 @@ g.test('timestamp_query_set,device_mismatch')
 
     t.tryComputePass(!mismatched, descriptor);
   });
+
+g.test('begin_compute_pass_before_ending_previous_pass')
```
Gyuyoung, line 195: I wrote this test referring to `CommandBufferValidationTest.BeginComputePassBeforeEndPreviousPass`[1] test. But, I'm not fully sure if this test was already implemented. wdyt?

[1] https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/src/dawn/tests/unittests/validation/CommandBufferValidationTests.cpp;l=145?q=CommandBufferValidationTests.cpp&ss=chromium

kainino0x: Took me a bit to find it but the file for this is `encoder_state.spec.ts`. It is currently empty. It has a very vague test plan though it's quite confusing. We had a draft issue open for it, so I just filed it and assigned you. #1914

Since we need many similar tests between render passes and compute passes, I do think it makes sense to put them all in one file, so I guess let's move this there.

src/webgpu/api/validation/encoding/beginComputePass.spec.ts
```diff
@@ -191,3 +191,36 @@ g.test('timestamp_query_set,device_mismatch')
...
+          storeOp: 'store',
+        },
+      ],
+    });
+
+    // Begin a compute pass before ending the previous render pass.
+    const computePass = encoder.beginComputePass();
+
+    renderPass.end();
+    computePass.end();
```
kainino0x, line 221: Because end() can also fail validation due to incorrect state, it's a bit tricky to test this robustly.
Let's add some extra subcases, maybe something like this:
```ts
    .combine('endPasses', [[], [0], [1], [0, 1], [1, 0]])
```
```ts
    const passes = [renderPass, computePass];
    for (const index of t.params.endPasses) {
        passes[index].end();
    }
```

to make sure we cover every state error we could reasonably detect, e.g. the implementation incorrectly does one of the following:
- implicitly closes the first pass
- ignores the second pass without setting an error
- allows the two passes to run in parallel
- allows passes to nest

src/webgpu/api/validation/encoding/beginComputePass.spec.ts
kainino0x: As mentioned in that issue there's possibility for parameterization by [compute, compute], [compute, render], [render, compute], etc. but we weren't sure if it would be best to just write separate tests for each one.

Gyuyoung: yes, I also considered using parameters to test the cases of the combination of `compute` and `render` passes. But, it seemed like it would get complicated.

src/webgpu/api/validation/encoding/beginComputePass.spec.ts
Gyuyoung: I apply the subcases to the test. And, it seems to me that separate tests should handle the error states.

src/webgpu/api/validation/encoding/encoder_state.spec.ts
```diff
@@ -19,4 +19,67 @@ TODO:
...
+g.test('pass_end_invalid_order')
+  .desc(
+    `
+  Test that beginning a [compute|render] pass before ending the previous pass causes an error.
+  `
+  )
+  .paramsSubcasesOnly(u =>
+    u
+      .combine(`IsRenderPassFirst`, [true, false])
+      .combine(`IsComputePassSecond`, [true, false])
```
kainino0x, line 54: nit: This naming is a little confusing, both the wording and why "true" means render for the first pass, but it means compute for the second pass. How about something like this?
```
      .combine('pass0Type', ['compute', 'render'])
      .combine('pass1Type', ['compute', 'render'])
```

src/webgpu/api/validation/encoding/encoder_state.spec.ts
```diff
@@ -19,4 +19,67 @@ TODO:
...
+    });
+  }
+}
+
+export const g = makeTestGroup(F);
+
+g.test('pass_end_invalid_order')
+  .desc(
+    `
+  Test that beginning a [compute|render] pass before ending the previous pass causes an error.
```
kainino0x, line 48:
```suggestion
  Test that beginning a {compute,render} pass before ending the previous {compute,render} pass causes an error.
```

src/webgpu/api/validation/encoding/encoder_state.spec.ts
```diff
@@ -19,4 +19,67 @@ TODO:
...
 import { ValidationTest } from '../validation_test.js';
 
-export const g = makeTestGroup(ValidationTest);
+class F extends ValidationTest {
+  beginRenderPass(commandEncoder: GPUCommandEncoder): GPURenderPassEncoder {
+    const attachmentTexture = this.device.createTexture({
+      format: 'rgba8unorm',
+      size: { width: 16, height: 16, depthOrArrayLayers: 1 },
+      usage: GPUTextureUsage.RENDER_ATTACHMENT,
+    });
```
kainino0x, line 28: nit: This texture is small but gets called quite a few times so let's make sure it gets cleaned up:
```suggestion
    });
    this.trackForCleanup(attachmentTexture);
```

kainino0x (approved):
LGTM w/ nits

src/webgpu/api/validation/encoding/encoder_state.spec.ts
```diff
@@ -19,4 +19,67 @@ TODO:
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
```
kainino0x, line 19: Also I know the TODO at the top of this file is confusing but we should try to update it as best as possible since we're adding some of the tests it describes.

kainino0x (approved):

src/webgpu/api/validation/encoding/encoder_state.spec.ts
```diff
@@ -19,4 +19,67 @@ TODO:
...
+    }
+
+    const passes = [firstPass, secondPass];
+    for (const index of endPasses) {
+      passes[index].end();
+    }
+
+    t.expectValidationError(() => {
+      encoder.finish();
+    }, true);
```
kainino0x, line 84: Maybe we should consider trying to have a control case here. Without a control case it's hard to be sure there isn't something wrong with the test other than what we're trying to test. However... I'm not really sure how to do it. It's probably fine if we can't think of a good way to add a control case.

kainino0x (approved):

src/webgpu/api/validation/encoding/encoder_state.spec.ts
Gyuyoung: Done.

src/webgpu/api/validation/encoding/encoder_state.spec.ts
Gyuyoung: Done.

src/webgpu/api/validation/encoding/encoder_state.spec.ts
Gyuyoung: ok, it seems more readable. Fixed.

src/webgpu/api/validation/encoding/encoder_state.spec.ts
Gyuyoung: yes, it was hard to add a control case from my side. That's why I just checked a validation error happens. Then, let me add a TODO for the control case. And, let's update this test if we have an idea to add a control case next time.

src/webgpu/api/validation/encoding/encoder_state.spec.ts
Gyuyoung: ok, let me update the description in the following CL because I feel like the updated description should be reviewed as well.

Issue #1914, opened 2022-10-13
val: encoding,encoder_state


kainino0x:
`webgpu:api,validation,encoding,encoder_state:*`
See code for TODO. Be sure to cover render bundles as well.

These probably don't need to be parameterized very much. The tests are pretty targeted.

Note the spec changed a bit here recently:
https://github.com/gpuweb/gpuweb/issues/2464
Implementations certainly don't match right now.

IIRC: It's an *exception* (not a validation error) to do anything on a closed pass or a closed encoder. Doing so does *not* invalidate the parent encoder.

kainino0x:
For testing the validation that the encoder is open when you issue a command, ideally we would test every command to make sure it implements the validation. I think we should have one file dedicated just to that. Here's a skeleton:

```ts
type EncoderCommands = keyof Omit<
  GPUCommandEncoder,
  '__brand' | 'label' | 'finish' | 'beginComputePass' | 'beginRenderPass'
>;
const kEncoderCommandInfo: {
  readonly [k in EncoderCommands]: {};
} = {
  clearBuffer: {},
  copyBufferToBuffer: {},
  copyBufferToTexture: {},
  copyTextureToBuffer: {},
  copyTextureToTexture: {},
  insertDebugMarker: {},
  popDebugGroup: {},
  pushDebugGroup: {},
  writeTimestamp: {},
  resolveQuerySet: {},
};
const kEncoderCommands = keysOf(kEncoderCommandInfo);

g.test('non_pass_commands')
  .params(u =>
    u
      .combine('command', kEncoderCommands)
      .beginSubcases()
      .combine('finishBeforeCommand', [false, true])
  )
  .beforeAllSubcases(t => {
    switch (t.params.command) {
      case 'writeTimestamp':
        t.selectDeviceOrSkipTestCase('timestamp-query');
        break;
    }
  })
  .fn(t => {
    const { command, finishBeforeCommand } = t.params;

    const encoder = t.device.createCommandEncoder();

    if (finishBeforeCommand) encoder.finish();

    t.expectValidationError(() => {
      switch (command) {
        case 'clearBuffer':
          // ...
          break;
        // ...
        default:
          unreachable();
      }
    }, finishBeforeCommand);

    if (!finishBeforeCommand) encoder.finish();
  });

// another test for compute pass commands

// another test for render pass commands
```

Gyuyoung:
[...]

PR #1917, opened 2022-10-15
val: Implement 'pass_end_none' and 'pass_end_twice' tests

Gyuyoung:
This PR adds new two tests to ensure that a validation error is generated if ending a pass twice or without ending a pass generates a validation error.

Issue: #1914

<hr>

### DIFF
--- a/src/webgpu/api/validation/encoding/encoder_state.spec.ts
+++ b/src/webgpu/api/validation/encoding/encoder_state.spec.ts
@@ -45,7 +45,7 @@ export const g = makeTestGroup(F);
 g.test('pass_end_invalid_order')
   .desc(
     `
-  Test that beginning a  {compute,render} pass before ending the previous {compute,render} pass
+  Test that beginning a {compute,render} pass before ending the previous {compute,render} pass
   causes an error.
 
   TODO: Need to add a control case to be sure a validation error happens because of ending order.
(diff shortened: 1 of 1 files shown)


kainino0x (commented):
This test reminded me that we actually changed this behavior in the spec. Now, end() itself is supposed to generate a validation error and *not* invalidate the encoder. This change was tracked here:
https://github.com/gpuweb/gpuweb/pull/2452

Filed https://crbug.com/dawn/1578 about it

Gyuyoung:
> This test reminded me that we actually changed this behavior in the spec. Now, end() itself is supposed to generate a validation error and _not_ invalidate the encoder. This change was tracked here: [gpuweb/gpuweb#2452](https://github.com/gpuweb/gpuweb/pull/2452)
> 
> Filed https://crbug.com/dawn/1578 about it

Then, I wonder if we should wait until .end() generates a validation error.

Gyuyoung:
@toji Could you please take a look at this PR instead of Kai?

src/webgpu/api/validation/encoding/encoder_state.spec.ts
```diff
@@ -114,3 +114,28 @@ g.test('call_after_successful_finish')
       encoder.finish();
     }
   });
+
+g.test('pass_end_none_or_twice')
```
kainino0x, line 118: This actually will be testing two separate validation errors, coming from different calls: one for when you finish() without ending the pass, and one for when you end() a pass that's already ended.

Let's make it two separate tests. The first test (finish without ending) should already pass. The second test (end after ending) will fail in Chromium, but we can write it anyway. It'll be very targeted so I think we'll be able to get it right.

Gyuyoung: ok, I separate the test into two tests like `pass_end_none`,  `pass_end_twice`. Please correct me if this is not what you want.

src/webgpu/api/validation/encoding/encoder_state.spec.ts
```diff
@@ -114,3 +114,45 @@ g.test('call_after_successful_finish')
...
+
+    const pass = passType === 'compute' ? encoder.beginComputePass() : t.beginRenderPass(encoder);
+
+    for (let i = 0; i < endCount; ++i) {
+      pass.end();
+    }
+
+    t.expectValidationError(() => {
+      encoder.finish();
+    }, endCount > 1);
```
kainino0x, line 157: There should be a validation error in the second pass.end() call, and then there shouldn't be one in finish.

```ts
pass.end();
if (endTwice) {
    pass.end();
  });
}

encoder.finish();
```

Chromium doesn't implement this yet, so the test will fail right now.

src/webgpu/api/validation/encoding/encoder_state.spec.ts
kainino0x: Split LGTM, left a comment about the second test

src/webgpu/api/validation/encoding/encoder_state.spec.ts
Gyuyoung: Ah, sorry for misunderstanding this. I apply this to the test and a fail happens as we expect.

kainino0x (approved):

PR #1921, opened 2022-10-18
val: Introduce encoder_open_state.spec.ts to test when an encoder is open

Gyuyoung:
This PR adds a new file to test all commands of GPUCommandEncoder, GPUComputePassEncoder, and GPURenderPassEncoder when the encoder is not finished.

This PR only tests all commands of GPUCommandEncoder first. Other commands of
GPUComputePassEncoder and GPURenderPassEncoder will be tested by the following
PRs.

Issue: #1914

<hr>

### DIFF
--- a/src/webgpu/api/validation/encoding/encoder_open_state.spec.ts
+++ b/src/webgpu/api/validation/encoding/encoder_open_state.spec.ts
@@ -2,7 +2,7 @@ export const description = `
 Validation tests to all commands of GPUCommandEncoder, GPUComputePassEncoder, and
 GPURenderPassEncoder when the encoder is not finished.
 
-- TODO: Test all commands of GPUComputePassEncoder and GPURenderPassEncoder.
+- TODO: Test all commands of GPUComputePassEncoder.
 `;
 
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
@@ -10,7 +10,30 @@ import { keysOf } from '../../../../common/util/data_tables.js';
 import { unreachable } from '../../../../common/util/util.js';
 import { ValidationTest } from '../validation_test.js';
 
-export const g = makeTestGroup(ValidationTest);
+import { beginRenderPassWithQuerySet } from './queries/common.js';
+
+class F extends ValidationTest {
+  beginRenderPass(commandEncoder: GPUCommandEncoder): GPURenderPassEncoder {
+    const attachmentTexture = this.device.createTexture({
+      format: 'rgba8unorm',
+      size: { width: 16, height: 16, depthOrArrayLayers: 1 },
+      usage: GPUTextureUsage.RENDER_ATTACHMENT,
+    });
+    this.trackForCleanup(attachmentTexture);
+    return commandEncoder.beginRenderPass({
+      colorAttachments: [
+        {
+          view: attachmentTexture.createView(),
+          clearValue: { r: 1.0, g: 0.0, b: 0.0, a: 1.0 },
+          loadOp: 'clear',
+          storeOp: 'store',
+        },
+      ],
+    });
+  }
+}
+
+export const g = makeTestGroup(F);
 
 type EncoderCommands = keyof Omit<
   GPUCommandEncoder,
@@ -31,6 +54,36 @@ const kEncoderCommandInfo: {
   resolveQuerySet: {},
 };
 const kEncoderCommands = keysOf(kEncoderCommandInfo);
+
+type RenderPassEncoderCommands = keyof Omit<
+  GPURenderPassEncoder,
+  | '__brand'
+  | 'label'
+  | 'end'
+  | 'setBindGroup'
+  | 'pushDebugGroup'
+  | 'popDebugGroup'
+  | 'setPipeline'
+  | 'insertDebugMarker'
+  | 'drawIndexedIndirect'
+  | 'draw'
+  | 'setIndexBuffer'
+  | 'setVertexBuffer'
+  | 'drawIndexed'
+  | 'drawIndirect'
+>;
+const kRenderPassEncoderCommandInfo: {
+  readonly [k in RenderPassEncoderCommands]: {};
+} = {
+  setViewport: {},
+  setScissorRect: {},
+  setBlendConstant: {},
+  setStencilReference: {},
+  beginOcclusionQuery: {},
+  endOcclusionQuery: {},
+  executeBundles: {},
+};
+const kRenderPassEncoderCommands = keysOf(kRenderPassEncoderCommandInfo);
 
 g.test('non_pass_commands')
   .params(u =>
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/encoding/encoder_open_state.spec.ts
```diff
@@ -0,0 +1,383 @@
...
+  .fn(t => {
+    const { command, finishBeforeCommand } = t.params;
+
+    const querySet = t.device.createQuerySet({ type: 'occlusion', count: 1 });
+    const encoder = t.device.createCommandEncoder();
+    const renderPass = beginRenderPassWithQuerySet(t, encoder, querySet);
+
+    if (finishBeforeCommand) {
+      renderPass.end();
+      encoder.finish();
```
kainino0x, line 237: This finish shouldn't be strictly necessary, commands should fail if the pass is ended even if the encoder is still open. But it's best to test both, which would be 3 cases: nothing before command, end before command, end+finish before command.

src/webgpu/api/validation/encoding/encoder_open_state.spec.ts
```diff
@@ -0,0 +1,383 @@
...
+            renderPass.beginOcclusionQuery(0);
+            renderPass.endOcclusionQuery();
+          }
+          break;
+        case 'executeBundles':
+          {
+            const bundleEncoder = t.device.createRenderBundleEncoder({
+              colorFormats: ['rgba8unorm'],
+            });
+            const bundle = bundleEncoder.finish();
```
kainino0x, line 277: We should avoid capturing createRenderBundleEncoder() and bundleEncoder.finish() in the expectValidationError call.

Maybe the best way to do this would be to have the switch create an anonymous function that executes the command. Like:

```ts
let encodeCommand: () => void;
switch (command) {
  // ...
  case 'executeBundles':
    {
      const bundle = ...;
      encodeCommand = () => renderPass.executeBundles([bundle]);
    }
    break;
  // ...
}

if (finishBeforeCommand) // ...
t.expectValidationError(encodeCommand, finishBeforeCommand);
```

src/webgpu/api/validation/encoding/encoder_open_state.spec.ts
```diff
@@ -0,0 +1,383 @@
...
+      renderPass.end();
+      encoder.finish();
+    }
+  });
+
+g.test('compute_pass_commands')
+  .params(u =>
+    u
+      .combine('command', kComputePassEncoderCommands)
+      // .beginSubcases()
```
kainino0x, line 296: nit: Should be uncommented?

src/webgpu/api/validation/encoding/encoder_open_state.spec.ts
```diff
@@ -0,0 +1,383 @@
...
+  )
+  .fn(t => {
+    const { command, finishBeforeCommand } = t.params;
+
+    const encoder = t.device.createCommandEncoder();
+    const computePass = encoder.beginComputePass();
+
+    if (finishBeforeCommand) {
+      computePass.end();
+      encoder.finish();
```
kainino0x, line 307: Same as comment above

src/webgpu/api/validation/encoding/encoder_open_state.spec.ts
```diff
@@ -0,0 +1,383 @@
...
+      switch (command) {
+        case 'setPipeline':
+          {
+            const computePipeline = t.createNoOpComputePipeline();
+            computePass.setPipeline(computePipeline);
+          }
+          break;
+        case 'dispatchWorkgroups':
+          {
+            const computePipeline = t.createNoOpComputePipeline();
```
kainino0x, line 320: Same as comment above

src/webgpu/api/validation/encoding/encoder_open_state.spec.ts
```diff
@@ -0,0 +1,383 @@
...
+              entries: [
+                {
+                  binding: 0,
+                  resource: outputTexture.createView(),
+                },
+              ],
+            });
+
+            computePass.setBindGroup(0, bindGroup);
+            computePass.setPipeline(computePipeline);
```
kainino0x, line 370: These commands would need to be encoded before ending the pass, so that we're testing validation inside dispatchWorkgroups only. However, I don't think we actually need to set the preconditions: if those fail, they'll just make the pass/encoder invalid. Here we are testing whether the encoding command *itself* generates an error. We don't even need to end the pass or finish the encoder at the end at all.

Same for other commands that have preconditions, like endOcclusionQuery.

src/webgpu/api/validation/encoding/encoder_open_state.spec.ts
```diff
@@ -0,0 +1,383 @@
...
+          break;
+        default:
+          unreachable();
+      }
+    }, finishBeforeCommand);
+
+    if (!finishBeforeCommand) {
+      renderPass.end();
+      encoder.finish();
+    }
```
kainino0x, line 289: Since we're just trying to test for validation errors coming directly from the encoding commands, I don't think we really need to end/finish at all here. See also comment on line 370

src/webgpu/api/validation/encoding/encoder_open_state.spec.ts
```diff
@@ -0,0 +1,383 @@
...
+      ],
+    });
+  }
+}
+
+export const g = makeTestGroup(F);
+
+type EncoderCommands = keyof Omit<
+  GPUCommandEncoder,
+  '__brand' | 'label' | 'finish' | 'beginComputePass' | 'beginRenderPass'
```
kainino0x, line 38: I think the validation in beginComputePass/beginRenderPass should be pretty similar, maybe they shouldn't be omitted here. We'll probably have other tests that cover it, but if it's trivial to test them here too, then why not. (Can just make this a TODO for now. Better to work on the other comments first)

src/webgpu/api/validation/encoding/encoder_open_state.spec.ts
```diff
@@ -0,0 +1,383 @@
...
+          break;
+        case 'setStencilReference':
+          {
+            renderPass.setStencilReference(0);
+          }
+          break;
+        case 'beginOcclusionQuery':
+        case 'endOcclusionQuery':
+          {
+            renderPass.beginOcclusionQuery(0);
```
kainino0x, line 268: Ideally we'd test beginOcclusionQuery and endOcclusionQuery separately. I think we can, see line 370.

kainino0x (commented):
Looks pretty good overall. I left some big comments but ultimately I don't think the code changes will be too huge.

src/webgpu/api/validation/encoding/encoder_open_state.spec.ts
Gyuyoung: Good point. I would like to test it in the following PR. I add a TODO for now.

src/webgpu/api/validation/encoding/encoder_open_state.spec.ts
Gyuyoung: Oops. Sorry for my mistake. Fixed.

src/webgpu/api/validation/encoding/encoder_open_state.spec.ts
Gyuyoung: Add a TODO here as well.

src/webgpu/api/validation/encoding/encoder_open_state.spec.ts
Gyuyoung: Ah, I see. I just thought that it would be good to end/finish the pass and encoder explicitly. Removed.

src/webgpu/api/validation/encoding/encoder_open_state.spec.ts
Gyuyoung: Removed them.

src/webgpu/api/validation/encoding/encoder_open_state.spec.ts
Gyuyoung: They needed to be called together because I called the end()/finish() explicitly. If we don't need to call the end()/finish() no longer, yes, we can separate them. Fixed.

src/webgpu/api/validation/encoding/encoder_open_state.spec.ts
Gyuyoung: Thank you for pointing it out. I remove unnecessary preconditions in the tests.

src/webgpu/api/validation/encoding/encoder_open_state.spec.ts
Gyuyoung: If we don't need to have the preconditions, how about just calling executeBundles without a bundle like `renderPass.executeBundles([])`?

src/webgpu/api/validation/encoding/encoder_open_state.spec.ts
Gyuyoung: Ok, I update the test to test begindComputePass/beginRenderPass.

src/webgpu/api/validation/encoding/encoder_open_state.spec.ts
```diff
@@ -0,0 +1,383 @@
...
+  | 'pushDebugGroup'
+  | 'popDebugGroup'
+  | 'setPipeline'
+  | 'insertDebugMarker'
+  | 'drawIndexedIndirect'
+  | 'draw'
+  | 'setIndexBuffer'
+  | 'setVertexBuffer'
+  | 'drawIndexed'
+  | 'drawIndirect'
```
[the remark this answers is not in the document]
Gyuyoung: I've only handled explicit functions of GPURenderPassEncoder and GPUComputePassEncoder except for including interfaces. But, I test them again :) Thanks!

> Which reminds me we should have similar tests for render bundles too.
Let me prepare the following PR for render bundles after merging these PRs.

Gyuyoung:
@kainino0x Thanks for reviewing these PRs! I've fixed all comments. But, please let me know if I'm not correct to fix them.

src/webgpu/api/validation/encoding/encoder_open_state.spec.ts
```diff
@@ -0,0 +1,468 @@
+export const description = `
+Validation tests to all commands of GPUCommandEncoder, GPUComputePassEncoder, and
+GPURenderPassEncoder when the encoder is not finished.
```
kainino0x, line 3:
```suggestion
GPURenderPassEncoder when the encoder is not finished.

TODO: Equivalent tests for GPURenderBundleEncoder.
```

kainino0x (approved):
Looks excellent! Just added a TODO and a minor MAINTENANCE_TODO.

Issue #1874, opened 2022-09-28
Add rg11b10ufloat-renderable feature test


takahirox:
Add `rg11b10ufloat-renderable` feature tests.

https://www.w3.org/TR/webgpu/#rg11b10ufloat-renderable

takahirox:
I made a PR for `requestDevice({requiredFeatures: ['rg11b10ufloat-renderable']})` test. #1888

I'm also thinking of writing texture validation/operation tests for `rg11b10ufloat-renderable`

kainino0x:
There are some issues where the current texture format tables in capability_info.ts can't represent things like this, as well as generally issues around testing optional features and limits.

We may be able to fix this at some point - I have some ideas - but right now, we should just write one-off tests for capabilities added by feature flags. They don't need to be perfectly represented in the capability table.

PR #1937, opened 2022-10-22
Add rg11b10ufloat-renderable validation test plan

takahirox:
Issue: #1874

This PR adds `rg11b10ufloat-renderable` validation test plan.

The validation test is for capabilities added by `rg11b10ufloat-renderable` feature. The `rg11b10ufloat-renderable` feature allows `rg11b10ufloat` texture format to be used with`RENDER_ATTACHMENT` usage and to be multisampled.

https://gpuweb.github.io/gpuweb/#rg11b10ufloat-renderable

So I picked validation tests related `RENDER_ATTACHMENT` usage and multisampling from the WebGPU spec. Let me know if I have missed anything.

Ideally `rg11b10ufloat-renderable` should be tested with `capability_info.ts` table but it has some problems now. Instead we should just write one-off tests and improve the test later.

So I added a new test file `src/webgpu/api/validation/texture/rg11b10ufloat_renderable.spec.ts`. We can just remove this file when we will imptove the test by integrating with `capability_info.ts` later.

Please also see: #1874

<hr>

### DIFF
--- a/src/webgpu/api/validation/texture/rg11b10ufloat_renderable.spec.ts
+++ b/src/webgpu/api/validation/texture/rg11b10ufloat_renderable.spec.ts
@@ -16,9 +16,7 @@ sampleCount > 1, iff rg11b10ufloat-renderable feature is enabled.
 `
   )
   .params(u =>
-    u
-      .combine('usage', [0, GPUConst.TextureUsage.RENDER_ATTACHMENT])
-      .combine('sampleCount', [1, 4])
+    u.combine('usage', [0, GPUConst.TextureUsage.RENDER_ATTACHMENT]).combine('sampleCount', [1, 4])
   )
   .unimplemented();
 
--- a/src/webgpu/api/validation/texture/rg11b10ufloat_renderable.spec.ts
+++ b/src/webgpu/api/validation/texture/rg11b10ufloat_renderable.spec.ts
@@ -47,7 +47,7 @@ g.test('create_render_pipeline')
   .desc(
     `
 Test that it is valid to create render pipeline with rg11b10ufloat texture format
-in descriptor.fragment.targets if rg11b10ufloat-renderable feature is enabled.
+in descriptor.fragment.targets iff rg11b10ufloat-renderable feature is enabled.
 `
   )
   .unimplemented();
--- a/src/webgpu/api/validation/texture/rg11b10ufloat_renderable.spec.ts
+++ b/src/webgpu/api/validation/texture/rg11b10ufloat_renderable.spec.ts
@@ -1,0 +1,53 @@
+export const description = `
+Tests for capabilities added by rg11b10ufloat-renderable flag.
+`;
+
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { ValidationTest } from '../validation_test.js';
+
+export const g = makeTestGroup(ValidationTest);
+
+g.test('create_texture_render_attachment')
+  .desc(
+    `
+Test that it is valid to create texture with rg11b10ufloat texture format and
+RENDER_ATTACHMENT usage is valid if rg11b10ufloat-renderable feature is enabled.
+`
+  )
+  .unimplemented();
+
+g.test('create_texture_multisampling')
+  .desc(
+    `
+Test that it is valid to create texture with rg11b10ufloat texture format and
+sampleCount > 1 if rg11b10ufloat-renderable feature is enabled.
+`
+  )
+  .unimplemented();
+
+g.test('begin_render_pass')
+  .desc(
+    `
+Test that it is valid to begin render pass with rg11b10ufloat texture format
+if rg11b10ufloat-renderable feature is enabled.
+`
+  )
+  .unimplemented();
+
+g.test('begin_render_bundle_encoder')
+  .desc(
+    `
+Test that it is valid to begin render bundle encoder with rg11b10ufloat texture
+format if rg11b10ufloat-renderable feature is enabled.
+`
+  )
+  .unimplemented();
+
+g.test('create_render_pipeline')
+  .desc(
+    `
+Test that it is valid to create render pipeline with rg11b10ufloat texture format
+in descriptor.fragment.targets if rg11b10ufloat-renderable feature is enabled.
+`
+  )
+  .unimplemented();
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/texture/rg11b10ufloat_renderable.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+  .unimplemented();
+
+g.test('create_texture_multisampling')
+  .desc(
+    `
+Test that it is valid to create texture with rg11b10ufloat texture format and
+sampleCount > 1 if rg11b10ufloat-renderable feature is enabled.
+`
+  )
+  .unimplemented();
```
kainino0x, line 26: Just for a little extra completeness, how about combining these two tests so they test 4 cases instead of just 3?

BTW when it's possible to write out the parameters with the plan, it makes it a bit easier to review test coverage completeness in the plan. For tests as simple as these, it's also totally fine to do both the plan and the implementation in one PR, though no harm in having a plan review!

```suggestion
g.test('create_texture')
  .desc(
    `
Test that it is valid to create rg11b10ufloat texture with RENDER_ATTACHMENT usage and/or
sampleCount > 1, iff rg11b10ufloat-renderable feature is enabled.
`
  )
  .params(u =>
    u
      .combine('enabled', [false, true])
      .combine('usage', [0, GPUConst.TextureUsage.RENDER_ATTACHMENT])
      .combine('sampleCount', [1, 4])
  )
  .unimplemented();
```

src/webgpu/api/validation/texture/rg11b10ufloat_renderable.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+if rg11b10ufloat-renderable feature is enabled.
+`
+  )
+  .unimplemented();
+
+g.test('begin_render_bundle_encoder')
+  .desc(
+    `
+Test that it is valid to begin render bundle encoder with rg11b10ufloat texture
+format if rg11b10ufloat-renderable feature is enabled.
```
kainino0x, line 41:
```suggestion
format iff rg11b10ufloat-renderable feature is enabled.
```

src/webgpu/api/validation/texture/rg11b10ufloat_renderable.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+`
+  )
+  .unimplemented();
+
+g.test('create_render_pipeline')
+  .desc(
+    `
+Test that it is valid to create render pipeline with rg11b10ufloat texture format
+in descriptor.fragment.targets if rg11b10ufloat-renderable feature is enabled.
```
kainino0x, line 50:
```suggestion
in descriptor.fragment.targets iff rg11b10ufloat-renderable feature is enabled.
```

src/webgpu/api/validation/texture/rg11b10ufloat_renderable.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+sampleCount > 1 if rg11b10ufloat-renderable feature is enabled.
+`
+  )
+  .unimplemented();
+
+g.test('begin_render_pass')
+  .desc(
+    `
+Test that it is valid to begin render pass with rg11b10ufloat texture format
+if rg11b10ufloat-renderable feature is enabled.
```
kainino0x, line 32: nit:

```suggestion
iff rg11b10ufloat-renderable feature is enabled.
```

src/webgpu/api/validation/texture/rg11b10ufloat_renderable.spec.ts
takahirox: > Just for a little extra completeness, how about combining these two tests so they test 4 cases instead of just 3?

That sounds good to me. Thanks for the suggestion.

>       .combine('enabled', [false, true])

If I'm right, disabled `rg11b10ufloat-renderable` feature test is not needed here beucause it's tested in https://github.com/gpuweb/cts/blob/main/src/webgpu/api/validation/createTexture.spec.ts

src/webgpu/api/validation/texture/rg11b10ufloat_renderable.spec.ts
takahirox: Thanks!

src/webgpu/api/validation/texture/rg11b10ufloat_renderable.spec.ts
takahirox: Thanks!

src/webgpu/api/validation/texture/rg11b10ufloat_renderable.spec.ts
takahirox: Thanks!

takahirox:
I have reflected the suggestions, thanks. I will make another PR for the implementation.

src/webgpu/api/validation/texture/rg11b10ufloat_renderable.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+import { GPUConst } from '../../../constants.js';
+import { ValidationTest } from '../validation_test.js';
+
+export const g = makeTestGroup(ValidationTest);
+
+g.test('create_texture')
+  .desc(
+    `
+Test that it is valid to create rg11b10ufloat texture with RENDER_ATTACHMENT usage and/or
+sampleCount > 1, iff rg11b10ufloat-renderable feature is enabled.
```
kainino0x, line 15:
```suggestion
sampleCount > 1, iff rg11b10ufloat-renderable feature is enabled.
Note, the createTexture tests cover these validation cases where this feature is not enabled.
```

kainino0x (approved):
Apologies for the delay, I lost track of this PR.

> If I'm right, disabled rg11b10ufloat-renderable feature test is not needed here

Good catch. It's true those tests will test that this is disallowed without the feature enabled. Since this test does have a control case (i.e. has both cases where validation passes and cases where validation fails), I think it's safe to leave out the enablement parameter. Let's add a note to the description though, so future readers don't have the same concern I had.

src/webgpu/api/validation/texture/rg11b10ufloat_renderable.spec.ts
kainino0x: Sigh, I forgot to commit this before merging (again)

PR #1942, opened 2022-10-25
Change from using `full` to `dense` for utility function names

zoddicus:
The utility function in question generate values across the entire range of a numeric type, but do not generate every value in the range, so it can be ambiguous when reading the name if it is going to produce all possible values or a sample of values.

`dense` is clearer that it is a larger selection of values than the corresponding `sparse` functions.

<hr>

### DIFF
--- a/src/unittests/maths.spec.ts
+++ b/src/unittests/maths.spec.ts
@@ -18,8 +18,8 @@ import {
   biasedRange,
   cartesianProduct,
   correctlyRoundedF32,
-  fullF32Range,
-  fullI32Range,
+  denseF32Range,
+  denseI32Range,
   hexToF32,
   hexToF64,
   lerp,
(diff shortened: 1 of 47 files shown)


ben-clayton:
`dense` implies to me that you've got a value for all the possible values of a set. That's not what it means here, though, right?

zoddicus:
> `dense` implies to me that you've got a value for all the possible values of a set. That's not what it means here, though, right?

Naming things is hard :upside_down_face: To me dense is clearer than full, but it sounds like you are in the opposite came, which implies there is still the ambiguity even with dense.

I will leave this as is, until I come up with an clear/unambiguous name.

ben-clayton:
I think my associations stem from `dense-matrix` vs `sparse-matrix`, where the former is usually value-for-each-cell / no gaps.
Sorry!

zoddicus:
> I think my associations stem from `dense-matrix` vs `sparse-matrix`, where the former is usually value-for-each-cell / no gaps. Sorry!

n/p, it is good to get different perspectives on changes like this, since I didn't view it like that.

Issue #1928, opened 2022-10-19
Debug marker validation and operation tests


Gyuyoung:
Need to test the validation of PushDebugGroup, PopDebugGroup, and InsertDebugMarker.

Gyuyoung:
I think this issue has done. Close.

PR #1945, opened 2022-10-27
op: Introduce debug_marker.spec.ts to test operations of debug APIs

Gyuyoung:
This PR introduces a new file to test the operation of debug APIs like pushDebugGroup, insertDebugMarker, and popDebugGroup methods.

As the first test, this PR adds a 'no_failure_without_debug_tool_attached' test to ensure calling a marker API without a debugging tool attached doesn't cause a failure.

Issue: #1928

<hr>

### DIFF
--- a/src/webgpu/api/operation/debug/debug_marker.spec.ts
+++ b/src/webgpu/api/operation/debug/debug_marker.spec.ts
@@ -1,0 +1,32 @@
+export const description = `
+Test operations of pushDebugGroup, popDebugGroup, and insertDebugMarker.
+`;
+
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { GPUTest } from '../../../gpu_test.js';
+import { kEncoderTypes } from '../../../util/command_buffer_maker.js';
+
+export const g = makeTestGroup(GPUTest);
+
+g.test('no_failure_without_debug_tool_attached')
+  .desc(
+    `
+  Test that calling a marker API without a debugging tool attached doesn't cause a failure.
+  `
+  )
+  .params(u =>
+    u //
+      .combine('encoderType', kEncoderTypes)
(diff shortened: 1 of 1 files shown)


src/webgpu/api/operation/debug/debug_marker.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+Test operations of pushDebugGroup, popDebugGroup, and insertDebugMarker.
+`;
+
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { GPUTest } from '../../../gpu_test.js';
+import { kEncoderTypes } from '../../../util/command_buffer_maker.js';
+
+export const g = makeTestGroup(GPUTest);
+
+g.test('no_failure_without_debug_tool_attached')
```
Gyuyoung, line 11: I write this test referring to the dawn end2end test[1]. However, I'm not fully sure if we need to add this test to ensure the operation since there are similar tests regarding this in some validation tests. wdyt?

[1] https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/src/dawn/tests/end2end/DebugMarkerTests.cpp;l=22?q=NoFailureWithoutDebugToolAttached&ss=chromium

kainino0x: As long as we have *some* test which calls the debug commands without expecting an error, I think that's good enough. I think the tests in #1921 will do this.

kainino0x: And we also have `api/validation/encoding/cmds/debug.spec.ts`

Gyuyoung: ok, then let me abandon this PR. Thanks.

PR #1951, opened 2022-11-01
Add support to load cross origin resources

shaoboyan:
This PR add ability to load cross origin resources from local if case running in local server. It uses '127.0.0.1' and 'localhost' trick to load resource from different origin in no network environment.

Issue: #

<hr>


shaoboyan:
@Kangz  I tried to verify this PR with network connection situation but hit some problem. When I build cts locally and try to access the build with local-ip/path/to/cts, it reprots `EXCEPTION: WebGPU device failed to initialize with Error "No WebGPU implementation found"; not retrying` in any case even I added `--enable-unsafe-webgpu` in command line.
But when I use `127.0.0.1/path/to/cts`, the cases works fine. Do you know how could I get the local-ip one work?

Kangz:
I don't really have any expertise here, sorry :x

kainino0x:
Sounds like `127.0.0.1` isn't a [SecureContext](https://webidl.spec.whatwg.org/#SecureContext) even though `localhost` is.

I think this [switch](https://peter.sh/experiments/chromium-command-line-switches/) will make it work, but we should probably find another solution where this isn't necessary. Unfortunate that the trick didn't work as well as we hoped :(
`--unsafely-treat-insecure-origin-as-secure=http://127.0.0.1`

kainino0x:
Sorry, I misread your error. Chrome does treat both `127.0.0.1` and `localhost` as SecureContext. Your error was on a local IP. *That* is not considered a SecureContext. You can still use `--unsafely-treat-insecure-origin-as-secure` to work around that if you need to.

In the case where we're running tests over the network, rather than on localhost, I think these tests will have to be skipped or fail, unless an alternate hostname is provided.

src/webgpu/api/validation/queue/copyToTexture/CopyExternalImageToTexture.spec.ts
```diff
@@ -136,6 +136,13 @@ function generateCopySizeForDstOOB({ mipLevel, dstOrigin }: WithDstOriginMipLeve
 }
 
 class CopyExternalImageToTextureTest extends ValidationTest {
+  onlineCrossOriginUrl = 'https://github.com/gpuweb/gpuweb/blob/main/logo/webgpu.png';
```
kainino0x, line 139: This goes to an HTML page, at least in a browser. Change `blob` to `raw`?

src/webgpu/api/validation/queue/copyToTexture/CopyExternalImageToTexture.spec.ts
```diff
@@ -323,9 +327,20 @@ g.test('source_image,crossOrigin')
...
-    const originCleanUrl = getResourcePath('Di-3d.png');
+    // Using 'localhost' and '127.0.0.1' trick to load cross origin resource. Set cross origin host name
+    // to 'localhost' if case is not running in 'localhost' domain. Otherwise, use '127.0.0.1'.
+    // host name to locahost unless the server running in
+    let crossOriginHostName = '';
+    if (location.hostname === 'localhost') {
+      crossOriginHostName = 'http://127.0.0.1';
+    } else {
+      crossOriginHostName = 'http://localhost';
+    }
```
kainino0x, line 338: I think this logic should be in `resources.ts`, because the cross-origin hostnames won't be specific to this test.

src/common/framework/resources.ts
```diff
@@ -5,11 +5,46 @@
...
+        break;
+      default:
+        relparts.push(pathparts[i]);
+        break;
+    }
+  }
+
+  return relparts.join('/');
+}
+
```
kainino0x, line 31: Is it possible to just concatenate the strings rather than manually parsing out `.`/`..`?
For example, this URL works:
`https://www.google.com/some/path/../../index.html`
because the browser resolves `..` for us.

shaoboyan: Unfortunately, not possible. I tried with ts logic but it seems not work. But as you said, when you type `./..` in url bar, the page could be access correctly. I think this might be url bar handles the relative path.

kainino0x: OK, makes sense, thanks.

shaoboyan:
> I don't really have any expertise here, sorry :x
No worries :P.

> Sorry, I misread your error. Chrome does treat both `127.0.0.1` and `localhost` as SecureContext. Your error was on a local IP. _That_ is not considered a SecureContext. You can still use `--unsafely-treat-insecure-origin-as-secure` to work around that if you need to.

Thanks! it works perfectly!

kainino0x (commented):
Mostly LGTM, just one last thing

src/common/framework/resources.ts
```diff
@@ -5,11 +5,81 @@
...
+function runningOnLocalHost(): boolean {
+  const hostname = window.location.hostname;
+  return hostname === 'localhost' || hostname === '127.0.0.1' || hostname === '::1';
```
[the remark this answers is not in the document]
shaoboyan: >All that to say, we'll need a way to override what cross-origin origin will be used, which can be overridden by the wpt runtime, for example. We don't need to figure out the logic to do that right now, but we do need to leave a MAINTENANCE_TODO at least in the wpt runtime file, for whoever ends up trying to run these tests and wants to stop them from accessing the internet.

Thanks for the details and that's right. One of the case I think not works is the “local area network”. I'd like to do one small step ahead by providing a parameter `crossOriginHost` in `getCrossOriginResourcePath`. And left a `MAINTENANCE_TODO` for a better solution.

>We should check it at least looks like it's going to work in Chromium's telemetry,
I think @haoxli helps me to check the old PR with telemetry (@haoxli , pls correct me if I understand not correctly). And I think we could check the latest PR.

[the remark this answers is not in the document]
haoxli: With this PR, the crossOrigin tests in Chromium telemetry will get 3 passed (isOriginClean=true), 3 skipped (isOriginClean=false).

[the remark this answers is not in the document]
shaoboyan: @haoxli  thanks! The skip cases means problems, let's sync offline.

kainino0x (approved):
LGTM, but ideally we can solve the problem in Telemetry before landing. It may just be a matter of customizing the crossOriginHost, though I don't know offhand how to do this.

shaoboyan:
@kainino0x  Thanks.
I've fixed the Telemetry skip issue. The reason is that in Telemetry `resoruces/` and `src/ (which is out/ when build cts)` are in the same level. It also config the `baseResorucePath` with absolute path. My previous function `getAbsoluteBaseResourcePath()` treat all paths as relative. So it generate a wrong url in Telemetry situation (wrong url seems like `localhost:8080/src/resources` but it should be `localhost:8080/resources`).

The new commit fix this issue and Telemetry seems happy now (Also verified standalone again)!

Issue #1954, opened 2022-11-01
insertBits: don't test overflow cases for const input source


amaiorano:
insertBits cannot overflow at const eval time: https://github.com/gpuweb/gpuweb/pull/3560

amaiorano:
Also will need to write tests to make sure it errors when overflowing for const eval (validation).

PR #1955, opened 2022-11-01
Fix insertBits tests to not test overflow cases for const input source

amaiorano:
Issue #1954

### DIFF
--- a/src/webgpu/shader/execution/expression/call/builtin/insertBits.spec.ts
+++ b/src/webgpu/shader/execution/expression/call/builtin/insertBits.spec.ts
@@ -92,7 +92,7 @@ g.test('integer')
       0b01010101010101010101010101010101
     );
 
-    await run(t, builtin('insertBits'), [T, T, TypeU32, TypeU32], T, cfg, [
+    const cases = [
       { input: [all_0, all_0, u32(0), u32(32)], expected: all_0 },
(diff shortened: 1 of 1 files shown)


zoddicus:
nit: For the comment I think it should be `Issue #1954` so that github cross links things nicely

src/webgpu/shader/execution/expression/call/builtin/insertBits.spec.ts
```diff
@@ -348,31 +348,37 @@ g.test('integer')
...
       { input: [pattern, all_1, u32(1), u32(0)], expected: pattern },
       { input: [pattern, all_1, u32(2), u32(0)], expected: pattern },
       { input: [pattern, all_1, u32(31), u32(0)], expected: pattern },
       { input: [pattern, all_1, u32(32), u32(0)], expected: pattern },
       { input: [pattern, all_1, u32(0), u32(0)], expected: pattern },
-    ]);
+    ];
+
+    if (t.params.inputSource != 'const') {
+      cases = cases.concat([
```
zoddicus, line 361: If you use `cases.push(...[` here, then `cases` can be a const up above.

amaiorano:
> nit: For the comment I think it should be `Issue #1954` so that github cross links things nicely

Done.

zoddicus (approved):

PR #1963, opened 2022-11-03
Convert asin, acos and atanh to a filtered range when needed.

dj2:
When doing const-eval for the `asin` and `acos` builtins they are only defined in the `[-1, 1]` range. The `atanh` builtin is only defined in the `(-1, 1)` range. Tests outside that range fail with a validation error. When not using const-evaluation having values outside the range is useful for validation.

This CL adds a `sourceFilteredF32Range` which will clamp the F32 values if the source is `const`. The `asin`, `acos`, and `atanh` tests are updated to use the new filter.

<hr>

### DIFF
--- a/src/webgpu/shader/execution/expression/call/builtin/atanh.spec.ts
+++ b/src/webgpu/shader/execution/expression/call/builtin/atanh.spec.ts
@@ -45,8 +45,21 @@ g.test('f32')
       ...biasedRange(kValue.f32.negative.less_than_one, -0.9, 20), // discontinuity at x = -1
       ...biasedRange(kValue.f32.positive.less_than_one, 0.9, 20), // discontinuity at x = 1
       ...sourceFilteredF32Range(
-        t.params.inputSource, kValue.f32.negative.less_than_one, kValue.f32.positive.less_than_one),
+        t.params.inputSource,
+        kValue.f32.negative.less_than_one,
+        kValue.f32.positive.less_than_one
+      ),
     ].map(makeCase);
+
+    // Handle the edge case of -1 and 1 when not doing const-eval
+    if (t.params.inputSource !== 'const') {
+      const edgeCases = [
+        ...biasedRange(-1, kValue.f32.negative.less_than_one, 20),
+        ...biasedRange(1, kValue.f32.positive.less_than_one, 20),
+      ].map(makeCase);
+      cases.push(...edgeCases);
+    }
+
     await run(t, builtin('atanh'), [TypeF32], TypeF32, t.params, cases);
   });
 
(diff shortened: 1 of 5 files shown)


zoddicus (approved):

src/webgpu/shader/execution/expression/call/builtin/atanh.spec.ts
```diff
@@ -41,10 +42,24 @@ g.test('f32')
...
+        kValue.f32.negative.less_than_one,
+        kValue.f32.positive.less_than_one
+      ),
     ].map(makeCase);
+
+    // Handle the edge case of -1 and 1 when not doing const-eval
+    if (t.params.inputSource !== 'const') {
+      const edgeCases = [
+        ...biasedRange(-1, kValue.f32.negative.less_than_one, 20),
+        ...biasedRange(1, kValue.f32.positive.less_than_one, 20),
```
zoddicus, line 58: These should just be -1 and 1. There shouldn't be any f32 values between 1(-1) and positive.less_than_one(negative.less_than_one), because there should only be 1 ULP between them.

 number is a f64 in TS, so it will generate values, but those will get quantized to either or less_than_one when going to f32.

src/webgpu/util/math.ts
```diff
@@ -565,6 +566,23 @@ export function fullF32Range(
...
+ * on the `source` parameter. If the `source` is `const` then the numbers will be restricted to be
+ * in the range `[low, high]`. This allows filtering out a set of `f32` values which are invalid for
+ * const-evaluation but are needed to test the non-const implementation.
+ *
+ * @param source the input source for the test. If the `source` is `const` then the return will be filtered
+ * @param low the lowest f32 value to permit when filtered
+ * @param high the highest f32 value to permit when filtered
+ */
+export function sourceFilteredF32Range(source: String, low: number, high: number): Array<number> {
+  const interval = new F32Interval(low, high);
```
zoddicus, line 582: self-nit: I know I suggested this, but now seeing it in situ, I am leery of pulling in F32Interval here, since a bunch of stuff in f32_interval.js depends on math.js and I have previously hit issues with weird dependency issues.

Might be better to just replace interval.contains(x) with (x >= low && x <= high), which also avoids building a F32Interval and a function call

src/webgpu/shader/execution/expression/call/builtin/atanh.spec.ts
dj2: done

src/webgpu/util/math.ts
dj2: done

zoddicus (approved):

Issue #1835, opened 2022-09-13
op: rendering,blending


kainino0x:

kainino0x:
This item refers to `api/operation/rendering/blending.ts`.
That file is partly implemented, but has a number of "unimplemented" tests at the bottom. These need to be fleshed out and implemented:
https://github.com/gpuweb/cts/blob/main/src/webgpu/api/operation/rendering/blending.spec.ts#L277

Example: `formats` : "Test blending results works for all formats that support it, and that blending is not applied for formats that do not. Blending should be done in linear space for srgb formats."

This description needs to be updated - it's a validation error to attempt to use blending with a format that doesn't support it. (We should check we have a validation test for that before deleting it from this test description.)

This test would:

- Set up a basic render pipeline like this one, with some basic blend states, for example `{ operation: "add", srcFactor: "one", dstFactor: "one" }` which should simply add the output of the fragment shader to the value currently in the buffer. From the fragment shader, output some value, like (0.4, 0.4, 0.4, 0.4).
- Set up a simple render pass with `loadOp: "clear"` and `clearValue` set to some value, like (0.2, 0.2, 0.2, 0.2).
- Ensure that the buffer contains the expected value, in this case would be (0.6, 0.6, 0.6, 0.6). Use a helper like `textureContentIsOKByT2B` to test the result with some error tolerance, because different formats will have different precisions. To start, try something like `{ maxDiffULPsForNormFormat: 1, maxDiffULPsForFloatFormat: 1 }` but you will likely need to increase the tolerance, at least for `maxDiffULPsForFloatFormat`.

You can ignore the srgb formats thing for the moment, it should hopefully "just work" when you use `textureContentIsOKByT2B`.

Some documentation here https://gpuweb.github.io/cts/docs/tsdoc/functions/webgpu_util_texture_texture_ok.textureContentIsOKByT2B.html

Gyuyoung:
@kainino0x Regarding `clamp,blend_foo` tests, I'm not clear what the `fixed-point` formats are, and how to list them up like `kBlendableFormats`[1]. In other words, I couldn't find what formats are the 'fixed-point' formats because it looks like there are no hints about `fixed-point` format in the tests.

[1] https://github.com/gpuweb/cts/blob/main/src/webgpu/api/operation/rendering/blending.spec.ts#L284

kainino0x:
[...]

PR #1976, opened 2022-11-08
op: Implement 'blend_component_clamp' test in blending.spec.ts

Gyuyoung:
This PR implements the' blend_conmponent_clamp' test to ensure that the blending components
are clamped in the blend equation with the simple component values.

Issue: #1835

<hr>

### DIFF
--- a/src/webgpu/api/operation/rendering/color_target_state.spec.ts
+++ b/src/webgpu/api/operation/rendering/color_target_state.spec.ts
@@ -17,6 +17,7 @@ import {
 } from '../../../capability_info.js';
 import { GPUTest } from '../../../gpu_test.js';
 import { float32ToFloat16Bits } from '../../../util/conversion.js';
+import { clamp } from '../../../util/math.js';
 import { TexelView } from '../../../util/texture/texel_view.js';
 import { textureContentIsOKByT2B } from '../../../util/texture/texture_ok.js';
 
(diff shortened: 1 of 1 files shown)


src/webgpu/api/operation/rendering/blending.spec.ts
```diff
@@ -377,8 +377,109 @@ g.test('clamp,blend_factor')
   .unimplemented();
 
 g.test('clamp,blend_color')
-  .desc('For fixed-point formats, test that the blend color is clamped in the blend equation.')
-  .unimplemented();
+  .desc('For unorm/snorm formats, test that the blend color is clamped in the blend equation.')
```
Gyuyoung, line 380: I'm still unclear on how to implement this test according to the description. I implement a simple test to check if the expected result is ok with several 'unorm' formats in order to make a progress on the test. Could you give me a hint or teach me what should we test here?

kainino0x: After reading #1980 I finally realized I think this test is referring to the blend constant. At some point we renamed "blend color" in the spec to "blend constant".

So, I guess what this was saying was, whatever value you pass into [setBlendConstant](https://gpuweb.github.io/gpuweb/#dom-gpurenderpassencoder-setblendconstant), it's clamped to the range of the format (0..1 or -1..1) before computing the blend results.

However, I think this is slightly wrong.

Here is the relevant bit of the Vulkan spec (which, unlike WebGPU, explains this clamping behavior):

> If the color attachment is fixed-point, the components of the source and destination values and blend factors are each clamped to [0,1] or [-1,1] respectively for an unsigned normalized or signed normalized color attachment prior to evaluating the blend operations. If the color attachment is floating-point, no clamping occurs.
> https://registry.khronos.org/vulkan/specs/1.2/html/chap27.html#framebuffer-blendoperations

This effectively says three things are clamped. We should update our tests to cover those.
- src value (shader output)
- src factor
- dst factor
- (Implicitly, clamping also necessarily occurs on the result, when it is stored into the texture.)

(Note, Vulkan says the dst value is also clamped, but that can't be out of range anyway because it comes directly from the render target.)

Note the src factor and dst factor are clamped after they're computed. Those are where the blend constants come in. In the case of blend factor `"constant"`, it doesn't matter, but in the case of `"one-minus-constant"`, I believe we want the clamping to occur after the "one minus" operation. Example:
If blend constant is 2, and format is snorm (-1..1 range):
Clamping before gives `1 - clamp(2) = 1 - 1 = 0`.
Clamping after gives `clamp(1 - 2) = clamp(-1) = -1`.
Clamping before AND after gives `clamp(1 - clamp(2)) = clamp(0) = 0`.
I think `-1` is correct according to Vulkan (but if we run the test and get a different result on any or all platforms, then we can investigate further).

Additionally, we should test that they're *not* clamped when using a float format.

Practically speaking, I think these three "clamp" tests should all be written as one test, with the cases being used to cover all of the clamps. Parameters would be, to start:
- Format: one unorm, one snorm, one float format.
- Blend operation: Just "add" should be fine.
- A list of manually written combinations of src value, src factor, dst value, and dst factor. Each of the 4 clamps (src value, src factor, dst factor, final result) needs at least one case to check whether it's happening, but we can enhance beyond that with tests for specific scenarios like the example I gave above.

src/webgpu/api/operation/rendering/blending.spec.ts
```diff
@@ -377,8 +377,109 @@ g.test('clamp,blend_factor')
...
+    renderPass.end();
+    t.device.queue.submit([commandEncoder.finish()]);
+
+    const expectedColor = computeBlendOperation(
+      srcColor,
+      srcComputeFactor,
+      dstColor,
+      dstComputeFactor,
+      t.params.operation
+    );
```
kainino0x, line 462: Here, srcColor, srcComputeFactor, and dstComputeFactor need to be clamped according to the range of the format, before computing the blend operation. (either here or inside computeBlendOperation)

src/webgpu/api/operation/rendering/blending.spec.ts
Gyuyoung: I appreciate your detail guidance about this test. Let me request to review this PR as soon as I update the test according to this comment.

Gyuyoung: Sorry for the late update on this draft. I check the value of the expected color. In `unorm` format case, the color value should be clamped to 1 if the sum of the `srcValue` and `dstValue` exceeds 1. But, in the float format case, it should not be clamped.

But, in the `snorm` format case, I don't know if the `snorm` format can be clamped yet. So, I only test the `unorm` and `float` formats for now.

Gyuyoung: @kainino0x I wonder if the current test can be merged with TODOs to make a progress in the test.

src/webgpu/api/operation/rendering/color_target_state.spec.ts
```diff
@@ -683,14 +683,105 @@ g.test('color_write_mask,blending_disabled')
...
-g.test('blending,clamp,blend_factor')
-  .desc('For fixed-point formats, test that the blend factor is clamped in the blend equation.')
-  .unimplemented();
+g.test('blend_components_clamp')
+  .desc(
+    `
+  Test that the blend constant is clamped in the blend equation on the unorm and float.
+    - TODO: Need to test snorm formats.
+    - TODO: Need to test srcFactor and dstFactor.
+  `
```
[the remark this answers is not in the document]
Gyuyoung: Applied, and thank you!

src/webgpu/api/operation/rendering/color_target_state.spec.ts
```diff
@@ -683,14 +683,105 @@ g.test('color_write_mask,blending_disabled')
...
+    });
+    renderPass.setPipeline(pipeline);
+    renderPass.draw(1);
+    renderPass.end();
+    t.device.queue.submit([commandEncoder.finish()]);
+
+    let expValue: number;
+    switch (format) {
+      case 'rgba8unorm': // unorm types should clamp if the sum of srcValue and dstValue exceeds 1.
+        expValue = srcValue + dstValue > 1 ? 1.0 : srcValue + dstValue;
```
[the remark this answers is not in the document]
Gyuyoung: Thank you for suggesting using it. Fixed.

src/webgpu/api/operation/rendering/blending.spec.ts
Gyuyoung: Thank you for reviewing this PR. I will prepare more following PRs to remove the TODOs.

PR #1980, opened 2022-11-09
op: Implement 'default_blend_color,initial_blend_color' test in blending.spec.ts

Gyuyoung:
This PR implements a new test to ensure the initial blending color is correctly set at the beginning of every sub-pass.

Issue: #1835

<hr>

### DIFF
--- a/src/webgpu/api/operation/rendering/blending.spec.ts
+++ b/src/webgpu/api/operation/rendering/blending.spec.ts
@@ -8,7 +8,7 @@ TODO:
 `;
 
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
-import { assert, unreachable } from '../../../../common/util/util.js';
+import { assert, TypedArrayBufferView, unreachable } from '../../../../common/util/util.js';
 import {
   kBlendFactors,
   kBlendOperations,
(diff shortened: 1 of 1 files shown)


Gyuyoung:
I write the test referring to ColorStateTest.DefaultBlendColor[1]. If the test is acceptable, I will add more cases of the test.

[1] https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/src/dawn/tests/end2end/ColorStateTests.cpp;l=931?q=ColorState&ss=chromium

src/webgpu/api/operation/rendering/blending.spec.ts
```diff
@@ -20,7 +20,71 @@ import { float32ToFloat16Bits } from '../../../util/conversion.js';
...
+  ): GPURenderPipeline {
+    return this.device.createRenderPipeline({
+      layout: 'auto',
+      fragment: {
+        targets: [
+          {
+            format,
+            blend: {
+              color: blendComponent ?? {},
+              alpha: blendComponent ?? {},
```
kainino0x, line 36: This assumes blendComponent is the same for color and alpha, but we'll probably need tests where they're not the same. I think we should change this method to take a whole `GPUColorTargetState` instead of the `format` and `blendComponent` separately.

src/webgpu/api/operation/rendering/blending.spec.ts
```diff
@@ -372,6 +436,67 @@ g.test('formats')
     t.expectOK(result);
   });
 
+g.test('default_blend_color,initial_blend_color')
```
kainino0x, line 439: At some point we changed the terminology in the spec from "blend color" to "blend constant", which has left some old code a bit confusing (and now I realize what one of the tests in #1976 is about...). Let's apply that rename to the changes in this PR, so that it will be clearer what it's testing.

src/webgpu/api/operation/rendering/blending.spec.ts
```diff
@@ -372,6 +436,67 @@ g.test('formats')
...
+    const basePipeline = t.createRenderPipelineForTest(kFormat, undefined);
+    const testPipeline = t.createRenderPipelineForTest(kFormat, {
+      srcFactor: 'constant',
+      dstFactor: 'one',
+      operation: 'add',
+    });
+
+    const renderTarget = t.device.createTexture({
+      usage: GPUTextureUsage.RENDER_ATTACHMENT | GPUTextureUsage.COPY_SRC,
+      size: [kSize, kSize],
```
kainino0x, line 456: We only really need this to be 1x1. This will make the `textureContentIsOKByT2B` check much faster.

src/webgpu/api/operation/rendering/blending.spec.ts
```diff
@@ -372,6 +436,67 @@ g.test('formats')
...
+          storeOp: 'store',
+        },
+      ],
+    });
+    renderPass.setPipeline(basePipeline);
+    renderPass.setBindGroup(
+      0,
+      t.createBindGroupForTest(basePipeline.getBindGroupLayout(0), kBlackColorData)
+    );
+    renderPass.draw(3);
```
kainino0x, line 475: I don't think this draw call has any effect. The texture should already be filled with zeroes when it's created, and I think this just re-fills it with zeroes. It's possible that, when the Dawn test was written, we needed this because we didn't have proper resource zeroing yet, but we don't need that here, so let's remove this draw call.

src/webgpu/api/operation/rendering/blending.spec.ts
```diff
@@ -372,6 +436,67 @@ g.test('formats')
     t.expectOK(result);
   });
 
+g.test('default_blend_color,initial_blend_color')
+  .desc(`Test that the initial blend color is correctly set at the beginning of every sub passes.`)
```
kainino0x, line 440:
```suggestion
  .desc(`Test that the blend constant is set to [0,0,0,0] at the beginning of a pass.`)
```

src/webgpu/api/operation/rendering/blending.spec.ts
```diff
@@ -372,6 +436,67 @@ g.test('formats')
...
+      0,
+      t.createBindGroupForTest(basePipeline.getBindGroupLayout(0), kBlackColorData)
+    );
+    renderPass.draw(3);
+    renderPass.setPipeline(testPipeline);
+    renderPass.setBindGroup(
+      0,
+      t.createBindGroupForTest(testPipeline.getBindGroupLayout(0), kWhiteColorData)
+    );
+    renderPass.draw(3);
```
kainino0x, line 481: Comment to explain the progression of the test.
```suggestion
    renderPass.draw(3);
    // Draw [1,1,1,1] with `src * constant + dst * 1`.
    // The blend constant defaults to [0,0,0,0], so the result is
    // `[1,1,1,1] * [0,0,0,0] + [0,0,0,0] * 1` = [0,0,0,0].
```

kainino0x (approved):
LGTM w/ nits

src/webgpu/api/operation/rendering/blending.spec.ts
Gyuyoung: ok, I change the parameter to `GPUColorTargetState`.

src/webgpu/api/operation/rendering/blending.spec.ts
Gyuyoung: ok, I rename `blend_color` to `blend_constant`.

src/webgpu/api/operation/rendering/blending.spec.ts
Gyuyoung: Ah, I see. I change the value of `kSize` to `1`.

src/webgpu/api/operation/rendering/blending.spec.ts
Gyuyoung: Thank you for pointing it out. I called the draw referring to the end2end's test.  Ok, let me remove the call both.

src/webgpu/api/operation/rendering/blending.spec.ts
Gyuyoung: Added.

PR #1985, opened 2022-11-10
Optimize shader expression by caching the cases

ben-clayton:
Profiling has highlighted that the logic to build the floating-point shader expression test cases can often be more expensive than running the test.

This problem is compounded by the fact that the shader expression tests are often run with slight variations to the input source ('const', 'uniform', etc).

This change introduces a framework for caching the test cases. The `DataCache` is two level:
* A "L1" in-memory cache - backed by a `Map`.
* A "L2" file cache, which can be pre-generated with an "offline" tool.

This change introduces the new tool: `npm run gen_cache <dir> [suite...]`

The idea here is that CTS testing infrastructure can pre-generate the test cases, and then reuse this data between the concurrent shards.

With the shader expression tests updated to use the `CaseCache`, local macOS + chromium + `standalone` CTS testing has shown 2x+ speed improvement when running all the `webgpu:shader,execution,expression,*` tests.
These performance gains will likely be reduced with multi-process execution, as the number of cache-misses will scale with the number of instances.

<hr>

### DIFF
--- a/src/webgpu/shader/execution/expression/case_cache.ts
+++ b/src/webgpu/shader/execution/expression/case_cache.ts
@@ -1,0 +1,202 @@
+import { Cacheable, dataCache } from '../../../../common/framework/data_cache.js';
+import { SerializedComparator, deserializeComparator } from '../../../util/compare.js';
+import {
+  Scalar,
+  Vector,
+  serializeValue,
+  SerializedValue,
+  deserializeValue,
+} from '../../../util/conversion.js';
+import {
+  deserializeF32Interval,
+  F32Interval,
+  SerializedF32Interval,
+  serializeF32Interval,
+} from '../../../util/f32_interval.js';
+
+import { Case, CaseList, Expectation } from './expression.js';
+
+/**
+ * SerializedExpectationValue holds the serialized form of an Expectation when
+ * the Expectation is a Value
+ * This form can be safely encoded to JSON.
+ */
+type SerializedExpectationValue = {
+  kind: 'value';
+  value: SerializedValue;
+};
+
+/**
+ * SerializedExpectationValue holds the serialized form of an Expectation when
+ * the Expectation is an Interval
+ * This form can be safely encoded to JSON.
+ */
+type SerializedExpectationInterval = {
+  kind: 'interval';
+  value: SerializedF32Interval;
+};
+
+/**
+ * SerializedExpectationValue holds the serialized form of an Expectation when
+ * the Expectation is a list of Intervals
+ * This form can be safely encoded to JSON.
+ */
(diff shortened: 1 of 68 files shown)


zoddicus (commented):
This overall looks really good to me, this takes the vague idea of memoizing these tests while I was writing them and delivers something far more complete and advanced.

I spot checked a number of the of the changes to the test .spec.ts files and they look good to me. I really like how they seperate the case generation out from the running of tests. As an aside, I might have to go through later and clean up some of these tests because they were some of the first ones that I wrote, but that isn't an issue with how you refactored what exists, but more of the existing code is iffy.

I will leave full review/approval to Kai, since he is much more of a TS expert than I and there is a bunch of node integration code/etc I don't feel qualified to review.

src/webgpu/shader/execution/expression/case_cache.ts
```diff
@@ -0,0 +1,202 @@
...
+
+/**
+ * SerializedExpectationValue holds the serialized form of an Expectation when
+ * the Expectation is a Comparator
+ * This form can be safely encoded to JSON.
+ */
+type SerializedExpectationComparator = {
+  kind: 'comparator';
+  value: SerializedComparator;
+};
```
kainino0x, line 57: optional idea:
I'm not sure if it would make any meaningful performance difference, but the JSON output could be much shorter by using something like:
`{ comparator: SerializedComparator }`
or a tuple type:
`['comparator', SerializedComparator]`

Similarly e.g. `{begin, end}` could be `[begin, end]`

src/webgpu/shader/execution/expression/case_cache.ts
```diff
@@ -0,0 +1,202 @@
...
+   */
+  deserialize(serialized: string): Record<string, CaseList> {
+    const data = JSON.parse(serialized) as Record<string, SerializedCase[]>;
+    const casesByName: Record<string, CaseList> = {};
+    for (const name in data) {
+      const cases = data[name].map(caseData => deserializeCase(caseData));
+      casesByName[name] = cases;
+    }
+    return casesByName;
+  }
```
kainino0x, line 194: You may be able to simplify/optimize some of this with the following:
- If `JSON.stringify` tries to serialize something with a `toJSON` method, it will call that and serialize the result
- `JSON.stringify` has `replacer` and `JSON.parse` has `reviver` that give you a chance to customize the JSON representation of a thing

kainino0x (approved):

src/webgpu/shader/execution/expression/case_cache.ts
ben-clayton: ack

PR #1987, opened 2022-11-11
op: Implement 'simple_blend_color,blend_constant_non_inherited' test in blending.spec.ts

Gyuyoung:
This PR implements a new test to ensure the blending constant is not inherited between render passes.

Issue: #1835

<hr>

### DIFF
--- a/src/webgpu/api/operation/rendering/blending.spec.ts
+++ b/src/webgpu/api/operation/rendering/blending.spec.ts
@@ -544,6 +544,83 @@ g.test('simple_blend_constant,setting_blend_constant')
     t.expectOK(result);
   });
 
+g.test('simple_blend_constant,blend_constant_non_inherited')
+  .desc(`Test that the blending constant is not inherited between render passes.`)
+  .fn(async t => {
+    const format = 'rgba8unorm';
+    const kSize = 1;
+    const kWhiteColorData = new Float32Array([255, 255, 255, 255]);
+
+    const blendComponent = { srcFactor: 'constant', dstFactor: 'one', operation: 'add' } as const;
+    const testPipeline = t.createRenderPipelineForTest({
+      format,
+      blend: { color: blendComponent, alpha: blendComponent },
+    });
+
+    const renderTarget = t.device.createTexture({
+      usage: GPUTextureUsage.RENDER_ATTACHMENT | GPUTextureUsage.COPY_SRC,
+      size: [kSize, kSize],
+      format,
+    });
+
+    const commandEncoder = t.device.createCommandEncoder();
+    {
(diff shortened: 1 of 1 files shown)


src/webgpu/api/operation/rendering/blending.spec.ts
```diff
@@ -544,6 +544,100 @@ g.test('simple_blend_constant,setting_blend_constant')
...
+      );
+      renderPass.draw(3);
+      // Draw [1,1,1,1] with `src * constant + dst * 1`. The blend constant defaults to [0,0,0,0],
+      // so the result is `[1,1,1,1] * [0,0,0,0] + [0,0,0,0] * 1` = [0,0,0,0].
+      renderPass.end();
+    }
+    t.device.queue.submit([commandEncoder.finish()]);
+
+    // Check that the blend constant is not inherited from the first render pass.
+    const expColor = { R: 0, G: 0, B: 0, A: 0 };
```
Gyuyoung, line 628: I understand this test is to ensure that the expected color should be black drawn by the second render pass without being affected by the first render pass drawing. But, a failure happens if `basePipeline` is not set and drawn first. Any thoughts on why this test should set `basePipeline` and draw it?

src/webgpu/api/operation/rendering/blending.spec.ts
```diff
@@ -544,6 +544,100 @@ g.test('simple_blend_constant,setting_blend_constant')
...
+      const renderPass = commandEncoder.beginRenderPass({
+        colorAttachments: [
+          {
+            view: renderTarget.createView(),
+            loadOp: 'load',
+            storeOp: 'store',
+          },
+        ],
+      });
+      renderPass.setPipeline(basePipeline);
```
Gyuyoung, line 609: I understand this test is to ensure that the expected color should be black drawn by the second render pass without being affected by the first render pass's drawing. But, if `basePipeline` is not set and drawn in the second render pass drawing, a failure happens because the texture is not set to zeros. That's why I keep drawing `basePipeline` here. I wonder if this makes sense to you as well.

kainino0x: How about just changing the second render pass to use `loadOp: 'clear'` instead of `'load'`? That should test the blend constant state just as well, and then you won't need a draw call to clear (in either render pass).

kainino0x (approved):
(LGTM either way though)

src/webgpu/api/operation/rendering/blending.spec.ts
Gyuyoung: Ah, it works after changing 'load' to 'clear in the `loadOp`. Fixed and thanks.

PR #1988, opened 2022-11-14
Implement rg11b10ufloat-renderable validation test

takahirox:
Issue: #1874 
Plan: #1937

This PR adds rg11b10ufloat-renderable validation test implementation.

<hr>

### DIFF
--- a/src/webgpu/api/validation/texture/rg11b10ufloat_renderable.spec.ts
+++ b/src/webgpu/api/validation/texture/rg11b10ufloat_renderable.spec.ts
@@ -16,10 +16,20 @@ sampleCount > 1, iff rg11b10ufloat-renderable feature is enabled.
 Note, the createTexture tests cover these validation cases where this feature is not enabled.
 `
   )
-  .params(u =>
-    u.combine('usage', [0, GPUConst.TextureUsage.RENDER_ATTACHMENT]).combine('sampleCount', [1, 4])
-  )
-  .unimplemented();
+  .params(u => u.combine('sampleCount', [1, 4]))
+  .beforeAllSubcases(t => {
+    t.selectDeviceOrSkipTestCase('rg11b10ufloat-renderable');
+  })
+  .fn(async t => {
+    const { sampleCount } = t.params;
+    const descriptor = {
+      size: [1, 1, 1],
+      format: 'rg11b10ufloat' as const,
+      sampleCount,
+      usage: GPUConst.TextureUsage.RENDER_ATTACHMENT,
+    };
+    t.device.createTexture(descriptor);
+  });
 
 g.test('begin_render_pass')
   .desc(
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/texture/rg11b10ufloat_renderable.spec.ts
```diff
@@ -16,10 +16,20 @@ sampleCount > 1, iff rg11b10ufloat-renderable feature is enabled.
 Note, the createTexture tests cover these validation cases where this feature is not enabled.
 `
   )
-  .params(u =>
-    u.combine('usage', [0, GPUConst.TextureUsage.RENDER_ATTACHMENT]).combine('sampleCount', [1, 4])
-  )
-  .unimplemented();
+  .params(u => u.combine('sampleCount', [1, 4]))
```
takahirox, line 19: I noticed that `usage` must not be zero and `sampleCount` > 1 requires `RENDER_ATTACHMENT` usage so removed `0` usage test case because it's useless here. (I think `0` usage fail test should be done somewhere else.)

https://gpuweb.github.io/gpuweb/#abstract-opdef-validating-gputexturedescriptor

kainino0x: I checked to make sure we didn't have missing coverage. It looks like 0 usage is already covered:
https://github.com/gpuweb/cts/pull/1575
Not sure about sampleCount without RENDER_ATTACHMENT, so I added the `needs-cts-issue` label to
https://github.com/gpuweb/gpuweb/pull/2786
to make sure we cover it.

kainino0x, line 19: As a control case, we should also parameterize over whether the feature is enabled or not. Then, createTexture should succeed "iff" the feature is enabled.

Same comment on all 4 of the tests here.

takahirox: This test case seems to cover `sampleCount > 1` without `RENDER_ATTACHMENT` usage. https://github.com/gpuweb/cts/blob/9c96bdc381db9aee24a445eac15f577d012fd91a/src/webgpu/api/validation/createTexture.spec.ts#L309-L390

takahirox: Sorry if I misunderstand something but we decided not to add enable parameter in this test case because non-feature-enabled test is covered by other tests and we added a note about it as comment, didn't we?

kainino0x: Sorry, I forgot about that conversation.

I was thinking there's a problem since we've removed the other parameterizations, and now there is no control case at all. However, since this is just a positive validation test (it expects validation *succeeds*), it's fine. Control cases are only strictly necessary for negative validation tests, to make sure the error isn't coming from the wrong place.

kainino0x (approved):

PR #1990, opened 2022-11-16
Remove infinity cases from const eval ranges

amaiorano:
The spec no longer allows for infinity values in const eval expressions (formerly, it wasn't allowed for abstracts, but was allowed for concretes).

Issue: #1989

### DIFF
--- a/src/webgpu/util/math.ts
+++ b/src/webgpu/util/math.ts
@@ -656,6 +656,23 @@ const kInterestingF32Values: number[] = [
   Number.POSITIVE_INFINITY,
 ];
 
+/** Short list of finite-only f32 values of interest to test against */
+const kInterestingF32ValuesFinite: number[] = [
+  kValue.f32.negative.min,
+  -10.0,
+  -1.0,
+  kValue.f32.negative.max,
+  kValue.f32.subnormal.negative.min,
+  kValue.f32.subnormal.negative.max,
+  0.0,
+  kValue.f32.subnormal.positive.min,
+  kValue.f32.subnormal.positive.max,
+  kValue.f32.positive.min,
+  1.0,
+  10.0,
+  kValue.f32.positive.max,
+];
+
 /** @returns minimal f32 values that cover the entire range of f32 behaviours
  *
  * Has specially selected values that cover edge cases, normals, and subnormals.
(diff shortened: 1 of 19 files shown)


amaiorano:
I still need to create issues for each of the ops and builtins to add validation for these cases.

src/webgpu/util/math.ts
```diff
@@ -683,38 +701,44 @@ export function sparseF32Range(): Array<number> {
...
-    [1.0, f, 2.0, 3.0],
-    [1.0, 2.0, f, 3.0],
-    [1.0, 2.0, 3.0, f],
-    [f, -1.0, -2.0, -3.0],
-    [-1.0, f, -2.0, -3.0],
-    [-1.0, -2.0, f, -3.0],
-    [-1.0, -2.0, -3.0, f],
-  ]),
-};
+export function vectorTestValues(dimension: number, source: String): number[][] {
```
ben-clayton, line 704: I'm really not keen on the propagation of the `source` as a string. It's pushing details of one particular part of tests into core maths libraries.

This isn't something you've introduced, but something that's being cemented. 

This is about finite vs non-finite test cases, right? So let's call this what it is. Something like `finite_only: boolean`.

amaiorano: ACK and done.

src/webgpu/util/math.ts
```diff
@@ -668,12 +685,14 @@ const kInterestingF32Values: number[] = [
  * specific values of interest. If there are known values of interest they
  * should be appended to this list in the test generation code.
  */
-export function sparseF32Range(): Array<number> {
-  return kInterestingF32Values;
+export function sparseF32Range(finite_only: boolean): Array<number> {
```
ben-clayton, line 688: I now remember that typescript has string literals as enum-like types, so `boolean` could be `'full' | 'finite'`, but I can follow up with stylistic changes. No need to block.

ben-clayton (approved):

src/webgpu/util/math.ts
amaiorano: Thanks, maybe I'll do it tomorrow

amaiorano:
Created validation issues for each case I modified here:
https://github.com/gpuweb/cts/issues?q=is%3Aopen+is%3Aissue+author%3Aamaiorano+created%3A2022-11-18

Issue #2023, opened 2022-11-24
op: Add more tests for Depth/Stencil State


Gyuyoung:
Need to test for the depth/stencil state[1] 

[1] https://gpuweb.github.io/gpuweb/#dictdef-gpudepthstencilstate

PR #2024, opened 2022-11-24
op: Implement 'depth_write_disabled' test in depth.spec.ts

Gyuyoung:
This PR implements 'depth_write_disabled' in depth.spec.ts. To implement the test, this PR checks the depth value is not written to the destination texture by copyTextureToTexture() function when 'depthWriteEnabled' state is disabled.

Issue: #2023 

<hr>

### DIFF
--- a/src/webgpu/api/operation/rendering/depth.spec.ts
+++ b/src/webgpu/api/operation/rendering/depth.spec.ts
@@ -3,13 +3,150 @@ Test related to depth buffer, depth op, compare func, etc.
 `;
 
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { TypedArrayBufferView } from '../../../../common/util/util.js';
 import { kDepthStencilFormats, kTextureFormatInfo } from '../../../capability_info.js';
 import { GPUTest } from '../../../gpu_test.js';
+import { TexelView } from '../../../util/texture/texel_view.js';
+import { textureContentIsOKByT2B } from '../../../util/texture/texture_ok.js';
 
 const backgroundColor = [0x00, 0x00, 0x00, 0xff];
 const triangleColor = [0xff, 0xff, 0xff, 0xff];
 
-export const g = makeTestGroup(GPUTest);
+const kBaseColor = new Float32Array([1.0, 1.0, 1.0, 1.0]);
+const kRedStencilColor = new Float32Array([1.0, 0.0, 0.0, 1.0]);
+const kGreenStencilColor = new Float32Array([0.0, 1.0, 0.0, 1.0]);
+
+type TestStates = {
+  state: GPUDepthStencilState;
+  color: Float32Array;
+  depth: number;
+};
+
+class DepthTest extends GPUTest {
+  runDepthStateTest(testStates: TestStates[], expectedColor: Float32Array) {
+    const renderTargetFormat = 'rgba8unorm';
+
+    const renderTarget = this.device.createTexture({
+      format: renderTargetFormat,
+      size: { width: 1, height: 1, depthOrArrayLayers: 1 },
+      usage: GPUTextureUsage.COPY_SRC | GPUTextureUsage.RENDER_ATTACHMENT,
+    });
+
+    const depthStencilFormat: GPUTextureFormat = 'depth24plus-stencil8';
+    const depthTexture = this.device.createTexture({
+      size: { width: 1, height: 1, depthOrArrayLayers: 1 },
+      format: depthStencilFormat,
+      sampleCount: 1,
+      mipLevelCount: 1,
+      usage: GPUTextureUsage.RENDER_ATTACHMENT | GPUTextureUsage.COPY_DST,
+    });
+
+    const depthStencilAttachment: GPURenderPassDepthStencilAttachment = {
+      view: depthTexture.createView(),
+      depthLoadOp: 'load',
+      depthStoreOp: 'store',
+      stencilLoadOp: 'load',
+      stencilStoreOp: 'store',
+    };
+
+    const encoder = this.device.createCommandEncoder();
+    const pass = encoder.beginRenderPass({
+      colorAttachments: [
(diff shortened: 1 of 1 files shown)


Gyuyoung:
I implement `depth_write_disabled` test referring to the `copy_multisampled_depth` test[1] in order to check if the depth value is not written to the destination texture by 'copyTextureToTexture()' with multi-sampled depth formats if depthWriteEnabled is disabled. But, there may be a simpler way to test it. Please let me know if there is simpler ways to test it.

[1] https://github.com/gpuweb/cts/blob/main/src/webgpu/api/operation/command_buffer/copyTextureToTexture.spec.ts#L1401

Gyuyoung:
Convert to 'draft' again because it's likely I need to check this test further.

Gyuyoung:
> I implement `depth_write_disabled` test referring to the `copy_multisampled_depth` test[1] in order to check if the depth value is not written to the destination texture by 'copyTextureToTexture()' with multi-sampled depth formats if depthWriteEnabled is disabled. But, there may be a simpler way to test it. Please let me know if there is simpler ways to test it.
> 
> [1] https://github.com/gpuweb/cts/blob/main/src/webgpu/api/operation/command_buffer/copyTextureToTexture.spec.ts#L1401

Convert to 'draft' again because it's likely I need to check this test further.

kainino0x:
IIRC there are two ways to check the contents of a depth texture:
- Run depth tests against it to see whether the fragments are drawn or not
- Bind the depth texture to a `texture_depth_2d` bind point in a shader and use the shader to copy the data out.

The latter is a very useful helper operation. I don't think we've implemented it. How about adding a `createDepthTextureCopyForMapRead` like `createTextureCopyForMapRead`, and `depthTextureContentIsOK` like `textureContentIsOKByT2B`, and then using that?

kainino0x:
Using depth testing (with `'equal'`) may work fine for this test, since we know the exact depth value we expect to see, and there may not be precision/rounding issues. But it's not much simpler than a readback helper, because you still need to run a shader, and you need the shader to output the expected depth value to see whether the `'equal'` depth test passed or not.

Gyuyoung:
Thanks for the comments. I would like to re-write this test after getting the review of https://github.com/gpuweb/cts/pull/2032 with your comments.

Gyuyoung:
> Using depth testing (with `'equal'`) may work fine for this test, since we know the exact depth value we expect to see, and there may not be precision/rounding issues. But it's not much simpler than a readback helper, because you still need to run a shader, and you need the shader to output the expected depth value to see whether the `'equal'` depth test passed or not.

I found that there is a `DepthWriteDisabled` test[1] in the end2end. It seems that the test checks whether the `equal` depth passes or not depending on the depth write enabled. So I implement `depth_write_disabled` test based on the test again.

[1] https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/src/dawn/tests/end2end/DepthStencilStateTests.cpp;l=460?q=DepthStencilStateTest,%20DepthWriteDisabled&ss=chromium

src/webgpu/api/operation/rendering/depth.spec.ts
```diff
@@ -20,8 +159,62 @@ g.test('depth_bias')
 g.test('depth_disabled').desc(`Tests render results with depth test disabled`).unimplemented();
 
 g.test('depth_write_disabled')
-  .desc(`Tests render results with depth write disabled`)
-  .unimplemented();
+  .desc(
+    `
+  Test that disabling depth writes works and leaves the depth buffer unchanged.
+  `
+  )
```
kainino0x, line 166: Just as a small control case, let's add a case with depth write disabled, that expects the result to be red (because the last draw fails the depth test).

src/webgpu/api/operation/rendering/depth.spec.ts
```diff
@@ -3,13 +3,152 @@ Test related to depth buffer, depth op, compare func, etc.
...
+          view: renderTarget.createView(),
+          storeOp: 'store',
+          loadOp: 'load',
+        },
+      ],
+      depthStencilAttachment,
+    });
+
+    // Draw a triangle with the given stencil reference and the comparison function.
+    // The color will be kGreenStencilColor if the stencil test passes, and kBaseColor if not.
```
kainino0x, line 65: nit: Comment is about stencil, probably not relevant here?

src/webgpu/api/operation/rendering/depth.spec.ts
```diff
@@ -3,13 +3,152 @@ Test related to depth buffer, depth op, compare func, etc.
...
+    for (const test of testStates) {
+      const testPipeline = this.createRenderPipelineForTest(test.state, test.depth);
+      pass.setPipeline(testPipeline);
+      pass.setStencilReference(0);
```
kainino0x, line 69: nit: I don't think this is relevant to this test? (Also the default stencil reference is 0 anyway so it doesn't do anything)

kainino0x (approved):

src/webgpu/api/operation/rendering/depth.spec.ts
Gyuyoung: Ah, right. This comment is wrong here.  Fixed the comment.

src/webgpu/api/operation/rendering/depth.spec.ts
Gyuyoung: yes, we don't need to keep it here.

src/webgpu/api/operation/rendering/depth.spec.ts
Gyuyoung: ok, then let me add a case that expects the result to be red because it passes the depth test.

kainino0x: Sorry, I made a typo, I meant we should have another case with depth write enabled. I'll make a PR

PR #2032, opened 2022-11-25
op: Implement 'stencil_compare_func' test

Gyuyoung:
This PR adds a new test to ensure the stencil comparison functions work correctly as expected.
    
Issue: #2023

<hr>

### DIFF
--- a/src/webgpu/api/operation/rendering/stencil.spec.ts
+++ b/src/webgpu/api/operation/rendering/stencil.spec.ts
@@ -1,0 +1,232 @@
+export const description = `
+Test related to stencil states, stencil op, compare func, etc.
+`;
+
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { TypedArrayBufferView } from '../../../../common/util/util.js';
+import { GPUTest } from '../../../gpu_test.js';
+import { TexelView } from '../../../util/texture/texel_view.js';
+import { textureContentIsOKByT2B } from '../../../util/texture/texture_ok.js';
+
+const kBaseColor = new Float32Array([1.0, 1.0, 1.0, 1.0]);
+const kStencilColor = new Float32Array([0.0, 0.0, 0.0, 1.0]);
+
+class StencilTest extends GPUTest {
+  checkStencilCompareFunction(
+    compareFunction: GPUCompareFunction,
+    stencilRefValue: number,
+    expectedColor: Float32Array
+  ) {
+    const depthStencilFormat: GPUTextureFormat = 'depth24plus-stencil8';
+
+    const baseStencilState = {
+      compare: 'always',
+      failOp: 'keep',
+      passOp: 'replace',
+    } as const;
+
+    const stencilState = {
+      compare: compareFunction,
+      failOp: 'keep',
+      passOp: 'keep',
+    } as const;
+
+    const baseState = {
+      format: depthStencilFormat,
+      depthWriteEnabled: false,
+      depthCompare: 'always',
+      stencilFront: baseStencilState,
+      stencilBack: baseStencilState,
+    } as const;
+
+    const state = {
+      format: depthStencilFormat,
+      depthWriteEnabled: false,
+      depthCompare: 'always',
+      stencilFront: stencilState,
+      stencilBack: stencilState,
+    } as const;
+
+    this.runStencilStateTest(baseState, state, stencilRefValue, expectedColor);
+  }
+
+  runStencilStateTest(
+    baseState: GPUDepthStencilState,
+    state: GPUDepthStencilState,
+    stencilRefValue: number,
+    expectedColor: Float32Array
+  ) {
+    const renderTargetFormat = 'rgba8unorm';
+    const renderTarget = this.device.createTexture({
+      format: renderTargetFormat,
+      size: { width: 1, height: 1, depthOrArrayLayers: 1 },
+      usage: GPUTextureUsage.COPY_SRC | GPUTextureUsage.RENDER_ATTACHMENT,
+    });
+
+    const depthStencilFormat: GPUTextureFormat = 'depth24plus-stencil8';
+    const depthTexture = this.device.createTexture({
+      size: { width: 1, height: 1, depthOrArrayLayers: 1 },
+      format: depthStencilFormat,
+      sampleCount: 1,
+      mipLevelCount: 1,
+      usage: GPUTextureUsage.RENDER_ATTACHMENT | GPUTextureUsage.COPY_DST,
+    });
+
+    const depthStencilAttachment: GPURenderPassDepthStencilAttachment = {
+      view: depthTexture.createView(),
+      depthLoadOp: 'load',
+      depthStoreOp: 'store',
+      stencilLoadOp: 'load',
+      stencilStoreOp: 'store',
+    };
+
+    const encoder = this.device.createCommandEncoder();
+    const pass = encoder.beginRenderPass({
+      colorAttachments: [
+        {
+          view: renderTarget.createView(),
+          storeOp: 'store',
+          loadOp: 'load',
+        },
+      ],
+      depthStencilAttachment,
+    });
+
+    // Draw the base triangle with stencil reference 1.
+    // This clears the stencil buffer to 1.
+    {
+      const testPipeline = this.createRenderPipelineForTest(baseState);
+      pass.setPipeline(testPipeline);
+      pass.setStencilReference(1);
+      pass.setBindGroup(
+        0,
+        this.createBindGroupForTest(testPipeline.getBindGroupLayout(0), kBaseColor)
+      );
+      pass.draw(1);
+    }
+
+    // Draw a triangle with the given stencil reference and the comparison function.
+    // The color will be kStencilColor if the stencil test passes, and kBaseColor if not.
+    {
+      const testPipeline = this.createRenderPipelineForTest(state);
+      pass.setPipeline(testPipeline);
+      pass.setStencilReference(stencilRefValue);
+      pass.setBindGroup(
+        0,
+        this.createBindGroupForTest(testPipeline.getBindGroupLayout(0), kStencilColor)
+      );
+      pass.draw(1);
+    }
+
+    pass.end();
+    this.device.queue.submit([encoder.finish()]);
+
+    const expColor = {
+      R: expectedColor[0],
+      G: expectedColor[1],
+      B: expectedColor[2],
+      A: expectedColor[3],
+    };
+    const expTexelView = TexelView.fromTexelsAsColors(renderTargetFormat, coords => expColor);
+
+    const result = textureContentIsOKByT2B(
+      this,
+      { texture: renderTarget },
+      [1, 1],
+      { expTexelView },
+      { maxDiffULPsForNormFormat: 1 }
+    );
+    this.eventualExpectOK(result);
+    this.trackForCleanup(renderTarget);
+  }
+
+  createRenderPipelineForTest(depthStencil: GPUDepthStencilState): GPURenderPipeline {
+    return this.device.createRenderPipeline({
+      layout: 'auto',
+      vertex: {
+        module: this.device.createShaderModule({
+          code: `
+            @vertex
+            fn main(@builtin(vertex_index) VertexIndex : u32) -> @builtin(position) vec4<f32> {
+                return vec4<f32>(0.0, 0.0, 0.0, 1.0);
+            }
+            `,
+        }),
+        entryPoint: 'main',
+      },
+      fragment: {
+        targets: [{ format: 'rgba8unorm' }],
+        module: this.device.createShaderModule({
+          code: `
+            struct Params {
+              color : vec4<f32>
+            }
+            @group(0) @binding(0) var<uniform> params : Params;
+
+            @fragment fn main() -> @location(0) vec4<f32> {
+                return vec4<f32>(params.color);
+            }`,
+        }),
+        entryPoint: 'main',
+      },
+      primitive: { topology: 'point-list' },
+      depthStencil,
+    });
+  }
+
+  createBindGroupForTest(layout: GPUBindGroupLayout, data: TypedArrayBufferView): GPUBindGroup {
+    return this.device.createBindGroup({
+      layout,
+      entries: [
+        {
+          binding: 0,
+          resource: {
+            buffer: this.makeBufferWithContents(data, GPUBufferUsage.UNIFORM),
+          },
+        },
+      ],
+    });
+  }
+}
+
+export const g = makeTestGroup(StencilTest);
+
+g.test('stencil_compare_func')
+  .desc(
+    `
+  Tests that stencil comparison functions with the stencil reference value works as expected.
+  `
+  )
+  .params(u =>
+    u //
+      .combineWithParams([
+        { stencilCompare: 'always', stencilRefValue: 0, _expectedColor: kStencilColor },
+        { stencilCompare: 'always', stencilRefValue: 1, _expectedColor: kStencilColor },
+        { stencilCompare: 'always', stencilRefValue: 2, _expectedColor: kStencilColor },
+        { stencilCompare: 'equal', stencilRefValue: 0, _expectedColor: kBaseColor },
+        { stencilCompare: 'equal', stencilRefValue: 1, _expectedColor: kStencilColor },
+        { stencilCompare: 'equal', stencilRefValue: 2, _expectedColor: kBaseColor },
+        { stencilCompare: 'greater', stencilRefValue: 0, _expectedColor: kBaseColor },
+        { stencilCompare: 'greater', stencilRefValue: 1, _expectedColor: kBaseColor },
+        { stencilCompare: 'greater', stencilRefValue: 2, _expectedColor: kStencilColor },
+        { stencilCompare: 'greater-equal', stencilRefValue: 0, _expectedColor: kBaseColor },
+        { stencilCompare: 'greater-equal', stencilRefValue: 1, _expectedColor: kStencilColor },
+        { stencilCompare: 'greater-equal', stencilRefValue: 2, _expectedColor: kStencilColor },
+        { stencilCompare: 'less', stencilRefValue: 0, _expectedColor: kStencilColor },
+        { stencilCompare: 'less', stencilRefValue: 1, _expectedColor: kBaseColor },
+        { stencilCompare: 'less', stencilRefValue: 2, _expectedColor: kBaseColor },
+        { stencilCompare: 'less-equal', stencilRefValue: 0, _expectedColor: kStencilColor },
+        { stencilCompare: 'less-equal', stencilRefValue: 1, _expectedColor: kStencilColor },
+        { stencilCompare: 'less-equal', stencilRefValue: 2, _expectedColor: kBaseColor },
+        { stencilCompare: 'never', stencilRefValue: 0, _expectedColor: kBaseColor },
+        { stencilCompare: 'never', stencilRefValue: 1, _expectedColor: kBaseColor },
+        { stencilCompare: 'never', stencilRefValue: 2, _expectedColor: kBaseColor },
+        { stencilCompare: 'not-equal', stencilRefValue: 0, _expectedColor: kStencilColor },
+        { stencilCompare: 'not-equal', stencilRefValue: 1, _expectedColor: kBaseColor },
+        { stencilCompare: 'not-equal', stencilRefValue: 2, _expectedColor: kStencilColor },
+      ] as const)
+  )
+  .fn(async t => {
+    const { stencilCompare, stencilRefValue, _expectedColor } = t.params;
+    t.checkStencilCompareFunction(stencilCompare, stencilRefValue, _expectedColor);
+  });


Gyuyoung:
This PR implements the `stencil_compare_func` test referring to the `DepthStencilStateTest, StenciFoo` tests [1]. PTAL.

[1] https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/src/dawn/tests/end2end/DepthStencilStateTests.cpp;l=505?q=depthStencilState&ss=chromium

src/webgpu/api/operation/rendering/depth.spec.ts
```diff
@@ -3,13 +3,211 @@ Test related to depth buffer, depth op, compare func, etc.
...
+  ) {
+    const depthSpencilFormat: GPUTextureFormat = 'depth24plus-stencil8';
+    const baseColor = [255, 255, 255, 255];
+
+    const baseStencilState = {
+      compare: 'always' as GPUCompareFunction,
+      failOp: 'keep' as GPUStencilOperation,
+      depthFailOp: 'keep' as GPUStencilOperation,
+      passOp: 'replace' as GPUStencilOperation,
+    };
```
kainino0x, line 25: µNit: a single `as const` at the end here should work:
```ts
    const baseStencilState = {
      compare: 'always',
      failOp: 'keep',
      depthFailOp: 'keep',
      passOp: 'replace',
    } as const;
```

same on the others below

Also I don't think `depthFailOp` is relevant to this test so you could just omit it (it's optional).

src/webgpu/api/operation/rendering/depth.spec.ts
```diff
@@ -23,6 +221,167 @@ g.test('depth_write_disabled')
...
+          stencilRefValue: 1,
+          stencilColorExpected: false,
+        },
+        {
+          stencilCompare: 'not-equal',
+          stencilColor: greaterCompareColor,
+          stencilRefValue: 2,
+          stencilColorExpected: true,
+        },
+      ])
```
kainino0x, line 373: nit: `as const` here should avoid the need for `as GPUCompareFunction` below.
```suggestion
      ] as const)
```

src/webgpu/api/operation/rendering/depth.spec.ts
```diff
@@ -3,13 +3,211 @@ Test related to depth buffer, depth op, compare func, etc.
 `;
 
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { TypedArrayBufferView } from '../../../../common/util/util.js';
 import { kDepthStencilFormats, kTextureFormatInfo } from '../../../capability_info.js';
 import { GPUTest } from '../../../gpu_test.js';
 
+class DepthTest extends GPUTest {
+  checkStencilCompareFunction(
```
kainino0x, line 11: Stencil is mostly a separate feature from depth, so perhaps put this in a new file. There are some interactions, like `depthFailOp`, but those would probably fit best into the stencil file.

src/webgpu/api/operation/rendering/depth.spec.ts
```diff
@@ -3,13 +3,211 @@ Test related to depth buffer, depth op, compare func, etc.
...
+    stencilColor: number[]
+  ) {
+    const renderTargetFormat = 'rgba8unorm';
+    const renderTarget = this.device.createTexture({
+      format: renderTargetFormat,
+      size: { width: 1, height: 1, depthOrArrayLayers: 1 },
+      usage: GPUTextureUsage.COPY_SRC | GPUTextureUsage.RENDER_ATTACHMENT,
+    });
+
+    const depthSpencilFormat: GPUTextureFormat = 'depth24plus-stencil8';
```
kainino0x, line 70: µNit: Spencil -> Stencil

src/webgpu/api/operation/rendering/depth.spec.ts
```diff
@@ -3,13 +3,211 @@ Test related to depth buffer, depth op, compare func, etc.
...
+    });
+
+    const depthStencilAttachment: GPURenderPassDepthStencilAttachment = {
+      view: depthTexture.createView(),
+      depthLoadOp: 'load',
+      depthStoreOp: 'store',
+    };
+
+    depthStencilAttachment.stencilLoadOp = 'load';
+    depthStencilAttachment.stencilStoreOp = 'store';
```
kainino0x, line 86: nit: Why not just define these inline?

src/webgpu/api/operation/rendering/depth.spec.ts
```diff
@@ -3,13 +3,211 @@ Test related to depth buffer, depth op, compare func, etc.
...
 import { GPUTest } from '../../../gpu_test.js';
 
+class DepthTest extends GPUTest {
+  checkStencilCompareFunction(
+    compareFunction: GPUCompareFunction,
+    stencilColor: number[],
+    stencilRefValue: number,
+    stencilColorExpected: boolean
+  ) {
+    const depthSpencilFormat: GPUTextureFormat = 'depth24plus-stencil8';
```
kainino0x, line 17: nit: Spencil -> Stencil

src/webgpu/api/operation/rendering/depth.spec.ts
```diff
@@ -3,13 +3,211 @@ Test related to depth buffer, depth op, compare func, etc.
...
+        {
+          view: renderTarget.createView(),
+          storeOp: 'store',
+          loadOp: 'load',
+        },
+      ],
+      depthStencilAttachment,
+    });
+
+    // Draw the base triangle with stencil reference 1.
```
kainino0x, line 100:
```suggestion
    // Draw the base triangle with stencil reference 1.
    // This clears the stencil buffer to 1.
```

src/webgpu/api/operation/rendering/depth.spec.ts
```diff
@@ -3,13 +3,211 @@ Test related to depth buffer, depth op, compare func, etc.
...
+                depth : f32,
+            }
+            @group(0) @binding(0) var<uniform> params : Params;
+
+            @fragment fn main() -> @location(0) vec4<f32> {
+                return vec4<f32>(params.color, 1.0);
+            }`,
+        }),
+        entryPoint: 'main',
+      },
```
kainino0x, line 160: nit: please order `vertex` before `fragment` in this descriptor. That's the order they run in, so it's a bit easier to read through the shader code that way.

src/webgpu/api/operation/rendering/depth.spec.ts
```diff
@@ -3,13 +3,211 @@ Test related to depth buffer, depth op, compare func, etc.
...
+
+            @vertex
+            fn main(@builtin(vertex_index) VertexIndex : u32) -> @builtin(position) vec4<f32> {
+                var pos = array<vec2<f32>, 6>(
+                        vec2<f32>(-1.0,  1.0),
+                        vec2<f32>(-1.0, -1.0),
+                        vec2<f32>( 1.0, -1.0), // front-facing
+                        vec2<f32>(-1.0,  1.0),
+                        vec2<f32>( 1.0,  1.0),
+                        vec2<f32>( 1.0, -1.0)); // back-facing
```
kainino0x, line 178: nit: since the target is just 1x1, you could use `'point-list'` topology mode and just output a single point (`draw(1)`) at `0,0`

src/webgpu/api/operation/rendering/depth.spec.ts
```diff
@@ -3,13 +3,211 @@ Test related to depth buffer, depth op, compare func, etc.
...
+                return vec4<f32>(pos[VertexIndex], ${depth}, 1.0);
```
kainino0x, line 179: params.depth seems unused, maybe you can delete the binding definition?

also currently `params.depth` is getting initialized to `stencilColor[3]`, not `depth`.

src/webgpu/api/operation/rendering/depth.spec.ts
```diff
@@ -3,13 +3,211 @@ Test related to depth buffer, depth op, compare func, etc.
...
+        module: this.device.createShaderModule({
+          code: `
+            struct Params {
+                color : vec3<f32>,
+                depth : f32,
+            }
+            @group(0) @binding(0) var<uniform> params : Params;
+
+            @fragment fn main() -> @location(0) vec4<f32> {
+                return vec4<f32>(params.color, 1.0);
```
kainino0x, line 156: Like `${depth}` is templated below, I think you can do the same for `color` here `${color.join(',')}`.

Then you wouldn't need the `params` binding at all.

src/webgpu/api/operation/rendering/depth.spec.ts
```diff
@@ -3,13 +3,211 @@ Test related to depth buffer, depth op, compare func, etc.
...
+      pass.setPipeline(testPipeline);
+      pass.setStencilReference(1);
+      pass.setBindGroup(
+        0,
+        this.createBindGroupForTest(testPipeline.getBindGroupLayout(0), kBaseColor)
+      );
+      pass.draw(6);
+    }
+
+    // Draw a triangle with the given stencil reference and the comparison function.
```
kainino0x, line 113:
```suggestion
    // Draw a triangle with the given stencil reference and the comparison function.
    // The color will be stencilColor if the stencil test passes, and kBaseColor if not.
```

src/webgpu/api/operation/rendering/depth.spec.ts
```diff
@@ -23,6 +221,167 @@ g.test('depth_write_disabled')
...
   .unimplemented();
 
+g.test('stencil_compare_func')
+  .desc('Tests that stencil comparison functions works as expected.')
+  .params(u =>
+    u //
+      .combineWithParams([
+        {
+          stencilCompare: 'always',
+          stencilColor: lessCompareColor,
```
kainino0x, line 231: I think we only need to see whether the fragment passed or failed the stencil test, so we only need 2 colors. Why are different `stencilColor`s needed for different test cases?

src/webgpu/api/operation/rendering/depth.spec.ts
```diff
@@ -3,13 +3,211 @@ Test related to depth buffer, depth op, compare func, etc.
...
+          storeOp: 'store',
+          loadOp: 'load',
+        },
+      ],
+      depthStencilAttachment,
+    });
+
+    // Draw the base triangle with stencil reference 1.
+    {
+      const kBaseColor = new Float32Array([255, 255, 255, 255]);
```
kainino0x, line 102: There seem to be 3 definitions of this color (`baseColor`, `kBaseColor`, `kTriangleColor`), use just one?

src/webgpu/api/operation/rendering/depth.spec.ts
```diff
@@ -23,6 +221,167 @@ g.test('depth_write_disabled')
...
+          stencilColor: greaterCompareColor,
+          stencilRefValue: 2,
+          stencilColorExpected: true,
+        },
+      ])
+  )
+  .fn(async t => {
+    const { stencilCompare, stencilColor, stencilRefValue, stencilColorExpected } = t.params;
+    t.checkStencilCompareFunction(
+      stencilCompare as GPUCompareFunction,
```
kainino0x, line 378:
```suggestion
      stencilCompare,
```

src/webgpu/api/operation/rendering/depth.spec.ts
```diff
@@ -3,13 +3,211 @@ Test related to depth buffer, depth op, compare func, etc.
...
+      stencilRefValue,
+      stencilColorExpected ? stencilColor : baseColor
+    );
+  }
+
+  runStencilStateTest(
+    baseState: GPUDepthStencilState,
+    state: GPUDepthStencilState,
+    stencilRefValue: number,
+    stencilColor: number[]
```
kainino0x, line 61: I think `stencilColor` is mixed up with an `expectedColor`. If `stencilColorExpected` is `false` then this `stencilColor` will just be `baseColor` and then both draw calls below might draw the same color and you can't tell whether the stencil test passed or not.

src/webgpu/api/operation/rendering/depth.spec.ts
Gyuyoung: Ah, I see. I create stencil.spec.ts file, and then move this test to the file.

src/webgpu/api/operation/rendering/depth.spec.ts
Gyuyoung: Thanks for informing me of the use. Fixed and remove `depthFailOp`.

src/webgpu/api/operation/rendering/depth.spec.ts
Gyuyoung: Oops, fixed.

src/webgpu/api/operation/rendering/depth.spec.ts
Gyuyoung: Fixed.

src/webgpu/api/operation/rendering/depth.spec.ts
Gyuyoung: Ah, just copied from the references. Move them to inline.

src/webgpu/api/operation/rendering/depth.spec.ts
Gyuyoung: Ah, right. I make the test only use the `baseColor`.

src/webgpu/api/operation/rendering/depth.spec.ts
Gyuyoung: Ah, I see. Ordered `vertex` before `fragment`.

src/webgpu/api/operation/rendering/depth.spec.ts
Gyuyoung: I added this for the following tests. ok, let me add it again if it's necessary.

src/webgpu/api/operation/rendering/depth.spec.ts
Gyuyoung: Ah, thank you for pointing it out. Then, let me change this to `return vec4<f32>(0.0, 0.0, 0.0, 1.0);`

src/webgpu/api/operation/rendering/depth.spec.ts
Gyuyoung: I wrote this test referring to the end2end dawn reference test. So I needed to use the different `stencilColor` to cover the test. Do you think it's better to only use 2 colors?

[1] https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/src/dawn/tests/end2end/DepthStencilStateTests.cpp;l=505?q=depthStencilState&ss=chromium

src/webgpu/api/operation/rendering/depth.spec.ts
Gyuyoung: Done.

src/webgpu/api/operation/rendering/depth.spec.ts
Gyuyoung: Done.

src/webgpu/api/operation/rendering/depth.spec.ts
Gyuyoung: I thought that the result could be different because of the 'stencil state'. `baseState` has `always` compare function with 'replace' passOp. On the other hand, 'state' has the given compare function with 'keep' passOp. Can't the state difference affect the test result? I referred to the `CheckStencilCompareFunction`.
[1] https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/src/dawn/tests/end2end/DepthStencilStateTests.cpp;l=192;drc=3874c9832e2de7ebf55eb4cad2bf9683556fb5e9?q=depthStencilState&ss=chromium

src/webgpu/api/operation/rendering/depth.spec.ts
Gyuyoung: Hmm, when I tried to use `${color.join(',')}`, tests failed. Could you give me more information? Of, can't we just use this as is?

kainino0x: Yes, keeping the bind group is fine. Just seemed simpler if `depth` wasn't using it anyway, but may be useful later.

src/webgpu/api/operation/rendering/stencil.spec.ts
```diff
@@ -0,0 +1,230 @@
...
+    // This clears the stencil buffer to 1.
+    {
+      const testPipeline = this.createRenderPipelineForTest(baseState);
+      pass.setPipeline(testPipeline);
+      pass.setStencilReference(1);
+      pass.setBindGroup(
+        0,
+        this.createBindGroupForTest(
+          testPipeline.getBindGroupLayout(0),
+          new Float32Array(kBaseColor)
```
kainino0x, line 108: The members of this Float32Array are initialized to 0xff (255.0) and 0x00 (0.0). This technically works, they get clamped to the 0.0-1.0 range, but then the same values are used to initialize a Uint8Array later resulting in the unorm values 0xff (1.0) and 0x00 (0.0) which seems accidental.

src/webgpu/api/operation/rendering/stencil.spec.ts
```diff
@@ -0,0 +1,230 @@
...
+            `,
+        }),
+        entryPoint: 'main',
+      },
+      fragment: {
+        targets: [{ format: 'rgba8unorm' }],
+        module: this.device.createShaderModule({
+          code: `
+            struct Params {
+              color : vec3<f32>
```
kainino0x, line 160: The bind group is still getting initialized with 4 components (kBaseColor or kStencilColor) which means the alpha channel is dropped (and would end up in `depth` if you still had it here). 
How about just changing this to vec4 to avoid that?

src/webgpu/api/operation/rendering/depth.spec.ts
kainino0x: The problem is this `stencilColor` variable is actually getting set to `expectStencilColor ? kStencilColor : kBaseColor`. But then it's used as the draw color for the second draw call. If expectStencilColor is false, then both draw calls draw kBaseColor, and we expect kBaseColor, so the test passes regardless of which draw call's results we see.

This is my understanding of what this code means to do: https://github.com/kainino0x/cts/commit/c502adcb37a5d17c45f8f404b48eca4c4442f593
Or maybe I've misunderstood the intent of the original test? Since I didn't understand why it had so many colors, either.

kainino0x (commented):
Few more comments

src/webgpu/api/operation/rendering/stencil.spec.ts
Gyuyoung: ah right, it's looks like changing it to vec4 is more flexible for the the next tests.

src/webgpu/api/operation/rendering/stencil.spec.ts
Gyuyoung: Ah, thank you for pointing it out. Then, how about using textureContentIsOKByT2B to check the result instead of expectSinglePixelIn2DTexture?

src/webgpu/api/operation/rendering/depth.spec.ts
Gyuyoung: Ah, I see. I thought that the color could be affected by the stencil reference value and the stencil face state. But, it might be strange to draw the same color twice in the case that expecteStencilColor is false. I update the test passes `expectedColor` directly after making the second draw using kStencilColor. Then, check if the result color is `expectedColor`. How is this approach?

src/webgpu/api/operation/rendering/stencil.spec.ts
kainino0x: Sounds good.

src/webgpu/api/operation/rendering/stencil.spec.ts
```diff
@@ -0,0 +1,232 @@
...
+        { stencilCompare: 'less', stencilRefValue: 2, expectedColor: kBaseColor },
+        { stencilCompare: 'less-equal', stencilRefValue: 0, expectedColor: kStencilColor },
+        { stencilCompare: 'less-equal', stencilRefValue: 1, expectedColor: kStencilColor },
+        { stencilCompare: 'less-equal', stencilRefValue: 2, expectedColor: kBaseColor },
+        { stencilCompare: 'never', stencilRefValue: 0, expectedColor: kBaseColor },
+        { stencilCompare: 'never', stencilRefValue: 1, expectedColor: kBaseColor },
+        { stencilCompare: 'never', stencilRefValue: 2, expectedColor: kBaseColor },
+        { stencilCompare: 'not-equal', stencilRefValue: 0, expectedColor: kStencilColor },
+        { stencilCompare: 'not-equal', stencilRefValue: 1, expectedColor: kBaseColor },
+        { stencilCompare: 'not-equal', stencilRefValue: 2, expectedColor: kStencilColor },
```
kainino0x, line 226: nit: Make `expectedColor` a private param by renaming it to `_expectedColor` to shorten the test case path length.

kainino0x (approved):
LGTM! one small nit

src/webgpu/api/operation/rendering/stencil.spec.ts
Gyuyoung: ok, I rename `expectedColor` to `_expectedColor`.

PR #2046, opened 2022-12-02
expect command encoder validation error to be at beginRenderPass/beginComputePass time

shrekshao:
Issue: https://crbug.com/dawn/1602

<hr>

### DIFF
--- a/src/webgpu/api/validation/encoding/beginComputePass.spec.ts
+++ b/src/webgpu/api/validation/encoding/beginComputePass.spec.ts
@@ -17,7 +17,9 @@ class F extends ValidationTest {
     this.expectValidationError(() => {
       pass.end();
     }, !success);
-    encoder.finish();
+    this.expectValidationError(() => {
+      encoder.finish();
+    }, !success);
   }
 }
 
--- a/src/webgpu/api/validation/encoding/beginComputePass.spec.ts
+++ b/src/webgpu/api/validation/encoding/beginComputePass.spec.ts
@@ -10,9 +10,14 @@ class F extends ValidationTest {
   tryComputePass(success: boolean, descriptor: GPUComputePassDescriptor): void {
     const encoder = this.device.createCommandEncoder();
 
+    let pass: GPUComputePassEncoder;
     this.expectValidationError(() => {
-      encoder.beginComputePass(descriptor);
+      pass = encoder.beginComputePass(descriptor);
     }, !success);
+    this.expectValidationError(() => {
+      pass.end();
+    }, !success);
+    encoder.finish();
   }
 }
 
--- a/src/webgpu/api/validation/encoding/beginComputePass.spec.ts
+++ b/src/webgpu/api/validation/encoding/beginComputePass.spec.ts
@@ -9,11 +9,9 @@ import { ValidationTest } from '../validation_test.js';
 class F extends ValidationTest {
   tryComputePass(success: boolean, descriptor: GPUComputePassDescriptor): void {
     const encoder = this.device.createCommandEncoder();
-    const computePass = encoder.beginComputePass(descriptor);
-    computePass.end();
 
     this.expectValidationError(() => {
-      encoder.finish();
+      encoder.beginComputePass(descriptor);
     }, !success);
   }
 }
--- a/src/webgpu/api/validation/encoding/beginComputePass.spec.ts
+++ b/src/webgpu/api/validation/encoding/beginComputePass.spec.ts
@@ -14,9 +14,7 @@ class F extends ValidationTest {
     this.expectValidationError(() => {
       pass = encoder.beginComputePass(descriptor);
     }, !success);
-    this.expectValidationError(() => {
-      pass.end();
-    }, !success);
+    pass.end();
     this.expectValidationError(() => {
       encoder.finish();
     }, !success);
--- a/src/webgpu/api/validation/encoding/beginComputePass.spec.ts
+++ b/src/webgpu/api/validation/encoding/beginComputePass.spec.ts
@@ -14,7 +14,9 @@ class F extends ValidationTest {
     this.expectValidationError(() => {
       pass = encoder.beginComputePass(descriptor);
     }, !success);
-    pass.end();
+    this.expectValidationError(() => {
+      pass.end();
+    }, false);
     this.expectValidationError(() => {
       encoder.finish();
     }, !success);
--- a/src/webgpu/api/validation/encoding/beginRenderPass.spec.ts
+++ b/src/webgpu/api/validation/encoding/beginRenderPass.spec.ts
@@ -82,28 +82,28 @@ g.test('color_attachments,device_mismatch')
       ? t.getDeviceMismatchedRenderTexture()
       : t.getRenderTexture();
 
-    const encoder = t.createEncoder('non-pass');
-    const pass = encoder.encoder.beginRenderPass({
-      colorAttachments: [
-        {
-          view: view0Texture.createView(),
-          clearValue: { r: 1.0, g: 0.0, b: 0.0, a: 1.0 },
-          loadOp: 'clear',
-          storeOp: 'store',
-          resolveTarget: target0Texture.createView(),
-        },
-        {
-          view: view1Texture.createView(),
-          clearValue: { r: 1.0, g: 0.0, b: 0.0, a: 1.0 },
-          loadOp: 'clear',
-          storeOp: 'store',
-          resolveTarget: target1Texture.createView(),
-        },
-      ],
-    });
-    pass.end();
-
-    encoder.validateFinish(!mismatched);
+    t.expectValidationError(() => {
+      const encoder = t.createEncoder('non-pass');
+      const pass = encoder.encoder.beginRenderPass({
+        colorAttachments: [
+          {
+            view: view0Texture.createView(),
+            clearValue: { r: 1.0, g: 0.0, b: 0.0, a: 1.0 },
+            loadOp: 'clear',
+            storeOp: 'store',
+            resolveTarget: target0Texture.createView(),
+          },
+          {
+            view: view1Texture.createView(),
+            clearValue: { r: 1.0, g: 0.0, b: 0.0, a: 1.0 },
+            loadOp: 'clear',
+            storeOp: 'store',
+            resolveTarget: target1Texture.createView(),
+          },
+        ],
+      });
+      pass.end();
+    }, mismatched);
   });
 
 g.test('depth_stencil_attachment,device_mismatch')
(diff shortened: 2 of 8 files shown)


src/webgpu/api/validation/encoding/beginRenderPass.spec.ts
```diff
@@ -82,28 +82,28 @@ g.test('color_attachments,device_mismatch')
...
-          loadOp: 'clear',
-          storeOp: 'store',
-          resolveTarget: target1Texture.createView(),
-        },
-      ],
-    });
-    pass.end();
-
-    encoder.validateFinish(!mismatched);
+    t.expectValidationError(() => {
```
kainino0x, line 85: expectValidationError should only contain one WebGPU call so we can ensure the error is coming from the correct one

Same comment in several places below

src/webgpu/api/validation/encoding/beginComputePass.spec.ts
```diff
@@ -9,11 +9,9 @@ import { ValidationTest } from '../validation_test.js';
 class F extends ValidationTest {
   tryComputePass(success: boolean, descriptor: GPUComputePassDescriptor): void {
     const encoder = this.device.createCommandEncoder();
-    const computePass = encoder.beginComputePass(descriptor);
-    computePass.end();
 
     this.expectValidationError(() => {
-      encoder.finish();
+      encoder.beginComputePass(descriptor);
     }, !success);
```
kainino0x, line 15: If beginComputePass generates a validation error, I believe the spec says the parent encoder remains valid and so finish() should always succeed.

We can test that with something like this (untested)

```ts
    let pass: GPUComputePass;
      pass = encoder.beginComputePass(descriptor);
    }, !success);
      pass.end();
    }, !success);
    encoder.finish();

File: src/webgpu/api/validation/encoding/beginRenderPass.spec.ts [hunk 3]
```diff
@@ -163,8 +163,9 @@ g.test('occlusion_query_set,device_mismatch')
     });
     t.trackForCleanup(occlusionQuerySet);
 
-    const encoder = t.createEncoder('render pass', { occlusionQuerySet });
-    encoder.validateFinish(!mismatched);
+    t.expectValidationError(() => {
+      t.createEncoder('render pass', { occlusionQuerySet });
```
kainino0x, line 167: Internally this will make multiple WebGPU calls so I think it should be replaced with direct calls to createCommandEncoder and beginRenderPass

src/webgpu/api/validation/encoding/encoder_open_state.spec.ts
```diff
@@ -204,7 +204,20 @@ g.test('non_pass_commands')
...
-            encoder.beginRenderPass({ colorAttachments: [] });
+            const colorTexture = t.device.createTexture({
+              format: 'rgba8unorm',
+              size: { width: 4, height: 4, depthOrArrayLayers: 1 },
+              usage: GPUTextureUsage.RENDER_ATTACHMENT,
+            });
+            encoder.beginRenderPass({
+              colorAttachments: [
+                {
+                  view: colorTexture.createView(),
```
kainino0x, line 215: createTexture and createView calls should be moved outside of expectValidationError

src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts
```diff
@@ -76,13 +76,21 @@ class F extends ValidationTest {
...
+
   tryRenderPass(success: boolean, descriptor: GPURenderPassDescriptor): void {
     const commandEncoder = this.device.createCommandEncoder();
-    const renderPass = commandEncoder.beginRenderPass(descriptor);
-    renderPass.end();
 
     this.expectValidationError(() => {
-      commandEncoder.finish();
+      commandEncoder.beginRenderPass(descriptor);
     }, !success);
```
kainino0x, line 94: Same comment as for tryComputePass

src/webgpu/api/validation/encoding/queries/general.spec.ts
```diff
@@ -50,10 +52,9 @@ Tests that begin occlusion query with a invalid query set that failed during cre
   .fn(t => {
     const occlusionQuerySet = t.createQuerySetWithState(t.params.querySetState);
 
-    const encoder = t.createEncoder('render pass', { occlusionQuerySet });
-    encoder.encoder.beginOcclusionQuery(0);
-    encoder.encoder.endOcclusionQuery();
-    encoder.validateFinishAndSubmitGivenState(t.params.querySetState);
+    t.expectValidationError(() => {
+      t.createEncoder('render pass', { occlusionQuerySet });
```
kainino0x, line 56: same

src/webgpu/api/validation/encoding/beginComputePass.spec.ts
shrekshao: Looks like for current Chrome `encoder.finish()` fails for
```
at CheckCurrentEncoder (../../third_party/dawn/src/dawn/native/EncodingContext.h:93)
    at Finish (../../third_party/dawn/src/dawn/native/CommandEncoder.cpp:1473)
```
Does this look like a chrome implementation that needs to catch up?

and `pass.end()` generates no validation error on the invalid pass object.
[corresponding spec language](https://www.w3.org/TR/webgpu/#dom-gpurenderpassencoder-end):
> end()
  If any of the following requirements are unmet, make this [invalid](https://www.w3.org/TR/webgpu/#invalid) and stop.
  this must be [valid](https://www.w3.org/TR/webgpu/#valid).

This one seems to be an implementation update is needed

kainino0x: It looks like [something is wrong](https://github.com/gpuweb/gpuweb/issues/3662) with publishing to [TR](https://www.w3.org/TR/webgpu/). Please look at the github.io (webgpu.dev) version as it's up to date:
https://gpuweb.github.io/gpuweb/#dom-gpucommandencoder-beginrenderpass

Specifically it hasn't picked up this fix to `end()` yet:
https://github.com/gpuweb/gpuweb/pull/3626
It changes "make this invalid" to "make parentEncoder invalid" in `GPURenderPass.end()` (GPUComputePass was already correct).

kainino0x: As for the first thing - yes, it looks like chrome is behind the spec. However, it's also giving me pause due to its weird behavior. I'm filing an issue about this. https://github.com/gpuweb/gpuweb/issues/3663

kainino0x (approved):
LGTM % the open question about beginning passes

src/webgpu/api/validation/encoding/beginComputePass.spec.ts
shrekshao: Try to recap: with our latest spec (including https://github.com/gpuweb/gpuweb/issues/3663)
The expected behavior for an invalid descriptor will be
```
let pass: GPUComputePass;
  pass = encoder.beginComputePass(descriptor);
}, !success);
  pass.end();
}, !success);
    encoder.finish();
}, !success);
```

Because pass.end() will stop at pass.state not being "open", and never reached setting parentEncoder.state to "open".
Is the above statement correct?

kainino0x: I believe the following is how it works, but if you see discrepancies with the spec (after that PR) then it's probably an editorial error on my part.

- `encoder.beginComputePass(descriptor)`.
`encoder` is `"open"`, so no validation error.
Since `descriptor` is invalid, `encoder` and `pass` become invalid, but `pass.[[state]]` still has its initial value of `"open"`.
`encoder` is changed to `"locked"`.
- `pass.end()`.
`pass` is `"open"` and `encoder` is `"locked"`, so no validation error.
`encoder` is changed to `"open"`.
But since `pass` is invalid, `encoder` becomes invalid.\*
- `encoder.finish()`.
`encoder` is `"open"`, so no validation error.
`encoder` is changed to `"ended"`.
Since `encoder` is invalid, it returns an invalid GPUCommandBuffer.

\* Under ["as-if" rules](https://gpuweb.github.io/gpuweb/#w3c-conformant-algorithms), we don't have to implement this exactly. Since `finish()` is now guaranteed to fail regardless of whether `pass.end()` is called, we can simply make `encoder` invalid immediately (on this or any error inside the pass), which is how Dawn already worked.

shrekshao: Thanks for the explanation. It was me mixing up `pass.[[state]]` with `pass` being valid or not.

So we use 
```
let pass: GPUComputePass;
    pass = encoder.beginComputePass(descriptor);
}, !success);
pass.end();
encoder.finish();
```
for the test, which also aligns with current chrome/dawn implementation.

kainino0x: Sorry, I made an error in my list above. Finish doesn't error due to the encoder state, but it _does_ error because `encoder` is invalid. And `beginComputePass` does not error, because `encoder` is in the correct state, it's just the `descriptor` which is wrong. So I think the final code should be exactly as it was before this PR.

kainino0x: And I think that extends to the rest of this PR. I think if we land that spec change then none of these test changes will be needed. (Sorry. But thank you for uncovering the problems with my spec work.)

shrekshao: Sorry somehow fixed but didn't comment???

Now the test have

```
let pass: GPUComputePassEncoder;
      pass = encoder.beginComputePass(descriptor);
    }, !success);
      pass.end();
    }, false);
      encoder.finish();
    }, !success);
```

`beginComputePass` will fail for invalid descriptor (our original intent for making this PR)

`end` won't fail. I brace it with `expectValidationError` to avoid `pass` not assigned lint error.

`encoder.finish` still errors. (as it used to be)

This also pass current chrome implementation. I think this will be good to go? (If so please help me to squash merge, I will be OOO tomorrow. Thanks!)

kainino0x: Sorry I think we need to revert the chrome change and close this PR. I messed up when I was reshaping the validation here and I don't think beginComputePass should actually generate a validation error when you pass it an invalid encoder. See https://github.com/gpuweb/gpuweb/pull/3665

Though it's still up for change if need be.

shrekshao: I see. So we should close this and change our dawn implementation back to not throwing validation at beginCompute/RenderPass.

PR #2047, opened 2022-12-02
op: Implement 'stencil_basic' test

Gyuyoung:
This PR adds a new test to check if the basic stencil state work correctly.

Issue: #2023

<hr>

### DIFF
--- a/src/webgpu/api/operation/rendering/stencil.spec.ts
+++ b/src/webgpu/api/operation/rendering/stencil.spec.ts
@@ -191,6 +191,46 @@ class StencilTest extends GPUTest {
 
 export const g = makeTestGroup(StencilTest);
 
+g.test('stencil_basic')
+  .desc('Tests that the basic stencil state works as expected.')
+  .fn(async t => {
+    const depthSpencilFormat: GPUTextureFormat = 'depth24plus-stencil8';
+
+    const baseStencilState = {
+      compare: 'always',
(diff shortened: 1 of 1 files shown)


src/webgpu/api/operation/rendering/stencil.spec.ts
```diff
@@ -191,6 +191,46 @@ class StencilTest extends GPUTest {
 
 export const g = makeTestGroup(StencilTest);
 
+g.test('stencil_basic')
+  .desc('Tests that the basic stencil state works as expected.')
```
kainino0x, line 195: I'm not sure how much value this test has since it uses all of the default stencil test (and depth test) states. That's is the same as not setting the stencil state at all, and we have lots of other tests that render without a stencil state.

A less trivial version of this test would be to parameterize it. I'd consider just calling it a test of stencil compare modes and parameterizing like this:
- All (3) formats that contain a stencil aspect
- All (8) compare functions
- Stencil ref values of 0, 1, 2

... though I don't remember offhand if you already have written tests that cover that.

Gyuyoung: I added `stencil_compare_func` for the second. I am unsure if I can implement a test to cover all of them simultaneously. Let me check it. But, if it's challenging to cover all of them at once, let me add a test to cover them partially.

PR #2043, opened 2022-12-01
op: Implement stencil_read_write_mask test in stencil.spec.ts

Gyuyoung:
This PR adds a new test to check if the stencil read/write masks work correctly.

Issue: #2023

<hr>

**Requirements for PR author:**

- [x] All missing test coverage is tracked with "TODO" or `.unimplemented()`.
- [x] New helpers are `/** documented */` and new helper files are found in `helper_index.txt`.
- [x] Test behaves as expected in a WebGPU implementation. (If not passing, explain above.)

**Requirements for [reviewer sign-off](https://github.com/gpuweb/cts/blob/main/docs/reviews.md):**

- [x] Tests are properly located in the test tree.
- [x] [Test descriptions](https://github.com/gpuweb/cts/blob/main/docs/intro/plans.md) allow a reader to "read only the test plans and evaluate coverage completeness", and accurately reflect the test code.
- [x] Tests provide complete coverage (including validation control cases). **Missing coverage MUST be covered by TODOs.**
- [x] Helpers and types promote readability and maintainability.

When landing this PR, be sure to make any necessary issue status updates.

PR #2050, opened 2022-12-03
op: Implement 'depth_disabled' test in depth.spec.ts

Gyuyoung:
This PR adds a new test to check if rendering works correctly with depth testing disabled.

Issue: #2023

<hr>

### DIFF
--- a/src/webgpu/api/operation/rendering/depth.spec.ts
+++ b/src/webgpu/api/operation/rendering/depth.spec.ts
@@ -154,7 +154,28 @@ g.test('depth_bias')
   )
   .unimplemented();
 
-g.test('depth_disabled').desc(`Tests render results with depth test disabled`).unimplemented();
+g.test('depth_disabled')
+  .desc('Tests render results with depth test disabled.')
+  .fn(async t => {
+    const depthSpencilFormat: GPUTextureFormat = 'depth24plus-stencil8';
+    const state = { format: depthSpencilFormat };
+
+    const testStates = [
+      { state, color: kBaseColor, depth: 0.0 },
+      { state, color: kRedStencilColor, depth: 0.5 },
+      { state, color: kGreenStencilColor, depth: 1.0 },
+    ];
+
+    // Test that for all combinations and ensure the last triangle drawn is the one visible
+    // regardless of depth testing.
+    for (let last = 0; last < 3; ++last) {
+      const i = (last + 1) % 3;
+      const j = (last + 2) % 3;
+
+      t.runDepthStateTest([testStates[i], testStates[j], testStates[last]], testStates[last].color);
+      t.runDepthStateTest([testStates[j], testStates[i], testStates[last]], testStates[last].color);
+    }
+  });
 
 g.test('depth_write_disabled')
   .desc(


Gyuyoung:
This test was implemented by referring to the `DepthStencilStateTest.DepthStencilDisabled`[1] test.

[1] https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/src/dawn/tests/end2end/DepthStencilStateTests.cpp;l=395?q=depthStencilState&ss=chromium

Gyuyoung:
This PR needs to be rebased on https://github.com/gpuweb/cts/pull/2043 if it can be merged.

src/webgpu/api/operation/rendering/stencil.spec.ts
```diff
@@ -319,3 +320,64 @@ g.test('stencil_fail_operation')
     ];
     t.runStencilStateTest(baseState, testStates, 2, _expectedColor);
   });
+
+g.test('stencil_disabled')
+  .desc(
+    `
+  Test that for all combinations, the last triangle drawn is the one visible if the stencil is
+  disabled. We need to check against three triangles because the stencil test may modify results.
```
kainino0x, line 328: This description doesn't make much sense out of context. It needs to include an overview before talking about small details.

That said this test doesn't really cover anything about stencil testing. Without enabling some stencil state, there's no way to ever update the stencil buffer, so we wouldn't really be able to tell if there was a bug where stencil _testing_ state was wrong.

The original test is the same way, it wasn't testing anything about stencil, but it _was_ meaningfully testing something about depth: it draws triangles at 3 different depths in 3 different orders and ensures that depth testing is not taking place.

So I think if we want to keep this test it needs to be moved to the depth state tests. The stencil state test file has enough coverage, I think, that I'm not concerned about the default "disabled" states being wrong.

src/webgpu/api/operation/rendering/stencil.spec.ts
```diff
@@ -319,3 +320,64 @@ g.test('stencil_fail_operation')
...
+
+    const state = {
+      format: depthSpencilFormat,
+      depthWriteEnabled: false,
+      depthCompare: 'always',
+      stencilFront: stencilState,
+      stencilBack: stencilState,
+      stencilReadMask: 0xff,
+      stencilWriteMask: 0xff,
+    } as const;
```
kainino0x, line 358: This comment may not matter given the comment above. But leaving it here anyway.

Since this test is specifically testing the "disabled" state it should use all of the defaults, i.e. just not enabling depth/stencil testing, which is the "canonical" way of disabling it. The WebIDL defaulting should have the same results as what you've written.

```suggestion
    const baseState = { format: depthSpencilFormat };
    const state = { format: depthSpencilFormat };
```

I'm guessing the Dawn test is very old and was written before we had defaulting support in webgpu_cpp.h.

src/webgpu/api/operation/rendering/stencil.spec.ts
Gyuyoung: > The original test is the same way, it wasn't testing anything about stencil, but it was meaningfully testing something about depth: it draws triangles at 3 different depths in 3 different orders and ensures that depth testing is not taking place.

> So I think if we want to keep this test it needs to be moved to the depth state tests. The stencil state test file has enough coverage, I think, that I'm not concerned about the default "disabled" states being wrong.


It's likely I was not enough to understand the original test. I'm going to move this test to depth.spec.ts with some updates.

src/webgpu/api/operation/rendering/stencil.spec.ts
Gyuyoung: Thank you for informing me of it. Let me try to apply it to the next patchset.

Gyuyoung: I applied it to the last patchset.

kainino0x (approved):

PR #2045, opened 2022-12-01
WIP: Remove infinities from values for f32 tests

zoddicus:
This patch removes the use of infinities in inputs for all of the f32 tests. It is also introduces a framework for specifying filtering of inputs to const-eval tests, so that inputs that would cause validation errors can be avoided.

This patch currently does not have tests covering the new *LargestIntermediateValue functions, those will be added before final review and merge.

This does NOT remove the generation of NaN/Infinity via 1/0 when writing out testing values. That will be handled in a follow on PR, since that may have impact outside of the numeric tests.

Issue #2020

<hr>

**Requirements for PR author:**

- [ ] All missing test coverage is tracked with "TODO" or `.unimplemented()`.
- [ ] New helpers are `/** documented */` and new helper files are found in `helper_index.txt`.
- [ ] Test behaves as expected in a WebGPU implementation. (If not passing, explain above.)

**Requirements for [reviewer sign-off](https://github.com/gpuweb/cts/blob/main/docs/reviews.md):**

- [ ] Tests are properly located in the test tree.
- [ ] [Test descriptions](https://github.com/gpuweb/cts/blob/main/docs/intro/plans.md) allow a reader to "read only the test plans and evaluate coverage completeness", and accurately reflect the test code.
- [ ] Tests provide complete coverage (including validation control cases). **Missing coverage MUST be covered by TODOs.**
- [ ] Helpers and types promote readability and maintainability.

When landing this PR, be sure to make any necessary issue status updates.

Issue #2020, opened 2022-11-23
Remove infinity cases from numeric testing code


zoddicus:
Currently some of the numeric range generator code is including positive/negative infinities, and there is translation from Scalar using the `1/0` trick. This no longer works, because `1/0` is now a validation error. This is causing a bit of a whack-a-mole problem + some hacks to supposed toggling generating infinities on/off.

All of the existing infinity code should be ripped out to simplify things in the CTS now. And later a new exceptional testing path for just infinity introduced, since infinities will need to bt generated via something like `var one = 1.0f; var zero = 0.0f; var infinity = one/zero` to get a runtime infinite/arbitrary like value, assuming we can do that at all, since generating infinity is not guaranteed. I will file an additional bug for implementing infinity testing support.

PR #2053, opened 2022-12-05
Replace usage of vector inputs with fullF32Range for arithmetic tests

zoddicus:
During one of the refactorings of this code, the vectorTestValues were substituted in for fullF32Range when generating the inputs. This is not desired, so I am reverting this behavioural change.

There are more broken const-eval tests on some backends due to this change, though const-eval was failing on these backends already. This will be resolved in a pending change I have for filtering out bad const-eval inputs, https://github.com/gpuweb/cts/pull/2045.

I am sending up this patch early, so that coverage of division can be restored for investigating potential issues with the atan2 tests.

Issue #2020

<hr>

### DIFF
--- a/src/webgpu/shader/execution/expression/binary/f32_arithmetic.spec.ts
+++ b/src/webgpu/shader/execution/expression/binary/f32_arithmetic.spec.ts
@@ -12,7 +12,7 @@ import {
   remainderInterval,
   subtractionInterval,
 } from '../../../../util/f32_interval.js';
-import { vectorTestValues } from '../../../../util/math.js';
+import { fullF32Range } from '../../../../util/math.js';
 import { makeCaseCache } from '../case_cache.js';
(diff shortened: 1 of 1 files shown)


src/webgpu/shader/execution/expression/binary/f32_arithmetic.spec.ts
```diff
@@ -26,89 +26,109 @@ export const d = makeCaseCache('binary/f32_arithmetic', {
       return makeBinaryToF32IntervalCase(lhs, rhs, additionInterval);
     };
 
-    return vectorTestValues(2, false).map(v => {
```
zoddicus, line 29: I am aware that const and non_const versions are the same. They will be different in the future with my patch to resolve all of the const/non-const issues and leaving it like this makes the merges easier.

ben-clayton: ack

PR #2059, opened 2022-12-07
Remove attempting to emit infinities in WGSL

### LINKED ISSUE #2020: Remove infinity cases from numeric testing code

zoddicus:
This removes from the Scalar wgsl emission function attempting to emit infinities into shaders using the 1/0 trick, since that no longer works.

Tests are still able to store infinities in Scalars for comparison purposes and printing out to logs, just shader emission has been removed.

Attempting to write out an infinity via Scalar will now cause an error to be thrown in the test runner.

Previous patches have removed test cases that would be broken by this, local testing has shown no impact on test output.

Fixes #2020

<hr>

### DIFF
--- a/src/webgpu/util/conversion.ts
+++ b/src/webgpu/util/conversion.ts
@@ -721,20 +721,6 @@ export class Scalar {
         case 'bool':
           return `${this.value}`;
       }
-    } else if (this.value === Number.POSITIVE_INFINITY) {
-      switch (this.type.kind) {
-        case 'f32':
-          return `(1.f/0.f)`;
-        case 'f16':
-          return `(1.h/0.h)`;
-      }
-    } else if (this.value === Number.NEGATIVE_INFINITY) {
-      switch (this.type.kind) {
-        case 'f32':
-          return `(-1.f/0.f)`;
-        case 'f16':
-          return `(-1.h/0.h)`;
-      }
     }
     throw new Error(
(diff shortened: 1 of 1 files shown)


ben-clayton:
> Attempting to write out an infinity via Scalar will now cause an error to be thrown in the test runner.

Is this in place already? Could we not throw an exception in the `wgsl()` method? I fear that accidentally calling this with a non-finite value will print a string that will not compile, or worse still, print a symbol that _does_ compile by resolving to some unintended thing.

ben-clayton:
Sorry, github hid the code just below the deletions. I now see that we will error by falling through to the code below.
LGTM

ben-clayton (approved):

PR #2066, opened 2022-12-08
Filter oob const

zoddicus:
<hr>


src/webgpu/shader/execution/expression/expression.ts
```diff
@@ -563,159 +563,349 @@ function packScalarsToVector(
...
+ * Builds a Case for the param and unary interval generator provided.
  * The Case will use use an interval comparator for matching results.
  * @param param the param to pass into the unary operation
+ * @param check_intervals should intervals be checked if they are f32 finite
  * @param ops callbacks that implement generating an acceptance interval for a unary operation
  */
-export function makeUnaryToF32IntervalCase(param: number, ...ops: PointToInterval[]): Case {
+function makeUnaryToF32IntervalCase(
+  param: number,
+  check_intervals: boolean,
```
ben-clayton, line 574: instead of a `boolean` here, how about we use TypeScript's string enumerators so that the call sites are more readable (e.g. `'full' | 'finite_only'`)?  It would also be more future-proof, if we wanted to add more filter modes.
Alternatively, you could change this to be a predicate function, removing the embedded logic altogether.

zoddicus: Done

zoddicus:
PTAL

Rebased onto 'main' since the generate cases patch has landed.

Converted from using boolean to string enum

Fixed up corner cases for refract, pack2x16float, and faceForward.

Only failing tests for me are ones that Tint currently doesn't implement.

src/webgpu/shader/execution/expression/expression.ts
```diff
@@ -562,67 +562,92 @@ function packScalarsToVector(
   };
 }
 
+/** Indicates how strictly should intervals be checked. */
+export type IntervalCheck =
+  | 'f32' // Expected to be f32 finite
+  | 'none'; // No checks applied
```
ben-clayton, line 568: unfortunately, I'm still not able to look at `generateU32ToVectorCases(fullU32Range(), 'none', unpack4x8unormInterval);` and know what `'none'` means without looking up the parameter.

`'f32-only' | 'unfiltered'` or something would be much clearer when just reading the code that makes a call.

I don't want to block the change though, so feel free to land. We can always tidy this stuff up later.

ben-clayton (approved):

src/webgpu/shader/execution/expression/expression.ts
zoddicus: Done, hitting the patch with sed enough times made the change pretty mechanical

PR #2068, opened 2022-12-13
Add code coverage collection feature to the server

ben-clayton:
If the server was started with `--coverage` and the implementation exposes code coverage functionality, then the result of each test will include code coverage info.

Will be used by dawn/node to provide per-testcase code coverage information.

### DIFF
--- a/src/common/runtime/server.ts
+++ b/src/common/runtime/server.ts
@@ -8,7 +8,7 @@ import { dataCache } from '../framework/data_cache.js';
 import { DefaultTestFileLoader } from '../internal/file_loader.js';
 import { prettyPrintLog } from '../internal/logging/log_message.js';
 import { Logger } from '../internal/logging/logger.js';
-import { LiveTestCaseResult } from '../internal/logging/result.js';
+import { LiveTestCaseResult, Status } from '../internal/logging/result.js';
 import { parseQuery } from '../internal/query/parseQuery.js';
 import { TestQueryWithExpectation } from '../internal/query/query.js';
 import { TestTreeLeaf } from '../internal/tree.js';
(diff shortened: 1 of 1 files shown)


src/common/runtime/server.ts
```diff
@@ -38,10 +39,22 @@ function usage(rc: number): never {
...
+  // @returns a GPU with the given flags
   create(flags: string[]): GPU;
+  // An optional interface to a CodeCoverageProvider
+  coverage?: CodeCoverageProvider;
+}
+
+interface CodeCoverageProvider {
+  // Starts collecting code coverage
+  begin(): void;
+  // Ends collecting of code coverage, returning the coverage data
```
Kangz, line 56: If I read dawn.node's code correctly, it doesn't return the coverage data but the path to the coverage data.

src/common/runtime/server.ts
```diff
@@ -78,8 +94,21 @@ for (let i = 0; i < sys.args.length; ++i) {
   }
 }
 
+let codeCoverage: CodeCoverageProvider | undefined = undefined;
```
Kangz, line 97: Isn't that just `CodeCoverageProvider?`

src/common/runtime/server.ts
```diff
@@ -38,10 +39,22 @@ function usage(rc: number): never {
 interface RunResult {
   status: string;
   message: string;
+  coverageData?: string;
```
Kangz, line 42: Maybe explain that this is the path to the coverage data file and not the coverage data itself.

Kangz (approved):

src/common/runtime/server.ts
ben-clayton: Added a comment to make clear that it's opaque

src/common/runtime/server.ts
ben-clayton: Added a comment to make clear that it's opaque

src/common/runtime/server.ts
ben-clayton: Doesn't seem to work. ESLint wants to move the `?` to the LHS of the type, and then that fails to compile. _shrug_

PR #2073, opened 2022-12-14
Update copyTextureToTexture tests to allow both encodings of -1 for 8-byte snorm formats

austinEng:
Some drivers have been found to swap the encoding of -1 for snorm copies. They are probably internally performing a blit which changes the encoding.

Issue: https://crbug.com/dawn/1047

<hr>

### DIFF
--- a/src/webgpu/api/operation/command_buffer/copyTextureToTexture.spec.ts
+++ b/src/webgpu/api/operation/command_buffer/copyTextureToTexture.spec.ts
@@ -17,7 +17,7 @@ import {
 } from '../../../capability_info.js';
 import { GPUTest } from '../../../gpu_test.js';
 import { makeBufferWithContents } from '../../../util/buffer.js';
-import { checkElementsBetween } from '../../../util/check_contents.js';
+import { checkElementsEqual, checkElementsEqualEither } from '../../../util/check_contents.js';
 import { align } from '../../../util/math.js';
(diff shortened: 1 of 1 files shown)


austinEng:
I tried using textureContentIsOKByT2B but it wasn't obvious how since this test is also checking padding bytes of the copy.

Kangz:
> Some drivers have been found to have some values off by 1. This is probably because the copy is implemented as a blit and some rounding occurs.

Note that it shouldn't be that rounding occurs, but that the `MIN` and `MIN+1` values in snorm both represents -1, so if the blit samples `MIN` as -1, then writes it back as `MIN+1` then the value changes, but only this one.

kainino0x (approved):

kainino0x:
I think this is fine for now and can land. textureContentIsOKByT2B would have tighter behavior if it's really just the MIN vs MIN+1 problem, but we'll need to do something to be able to use it. Thinking either:
- enhance textureContentIsOKByT2B to be able to check the padding bytes
- use textureContentIsOKByT2B (or at least its mechanisms) but ALSO use a buffer check to check the padding bytes

austinEng:
I updated this instead to allow the alternative encoding swap only for 128/129 (0x80 / 0x81). PTAL again

Kangz (approved):

kainino0x (approved):

PR #2082, opened 2022-12-15
Add 'skipUndefined' Comparator

### LINKED ISSUE #2077: `pack2x16float`'s usage of custom Comparator prevents usage of CaseCache for these tests
Currently these cases use a custom comparator that does not have serialization primitives defined for it. Either they should be moved away from using a custom comparator, or serialization written.

zoddicus:
This adds a Comparator that if the Expectation is undefined, just returns true, so that tests can include an 'Any' test condition. This is used in 'pack2x16float' to replace its custom Comparator and allow for serialization/usage of the cache.

Fixes #2077

<hr>


src/webgpu/util/compare.ts
```diff
@@ -245,6 +272,12 @@ export function deserializeComparator(data: SerializedComparator): Comparator {
...
       return anyOf(...data.data.map(e => deserializeExpectation(e)));
     }
+    case 'skipUndefined': {
+      if (data.data.length === 0) {
+        return skipUndefined(undefined);
+      }
+      return skipUndefined(deserializeExpectation(data.data[0]));
+    }
   }
-  throw `unhandled comparator kind: ${data.kind}`;
```
zoddicus, line 249: The usage of data.kind is now invalid, because the type is 'anyOf' | 'skipUndefined', and not string. When 'anyOf' was the only option I think TS was able to infer the string type.

src/webgpu/util/compare.ts
```diff
@@ -226,14 +226,41 @@ export function anyOf(
...
 /** SerializedComparatorAnyOf is the serialized type of an `anyOf` comparator. */
 type SerializedComparatorAnyOf = {
   kind: 'anyOf';
   data: SerializedExpectation[];
 };
 
+/** SerializedComparatorSkipUndefined is the serialized type of an `skipUndefined` comparator. */
+type SerializedComparatorSkipUndefined = {
+  kind: 'skipUndefined';
+  data: [] | [SerializedExpectation];
```
ben-clayton, line 259: This is an odd type. Is this supposed to be an optional field?
If so, can't you use`data?: SerializedExpectation`?
If that doesn't work, `data: SerializedExpectation | undefined` ?

zoddicus: I did this because in src/webgpu/shader/execution/expression/case_cache.ts there is logic guarding that .kind and .data are not undefined. I assume that guard is there because we cannot serialize undefined.

zoddicus: In theory it could have been a SerializedExpectation[], but I wanted to make it clear that the can only be one SerializedException.

src/webgpu/shader/execution/expression/case_cache.ts
```diff
@@ -80,7 +80,14 @@ export function serializeExpectation(e: Expectation): SerializedExpectation {
   if (e instanceof Function) {
     const comp = (e as unknown) as SerializedComparator;
     if (comp.kind !== undefined && comp.data !== undefined) {
```
zoddicus, line 82: Here is the check for undefined entries

src/webgpu/util/compare.ts
ben-clayton: I think we can serialize undefined - it's just the absence of a JSON field. Feel free to fix the case_cache.ts logic

zoddicus: or probably just an optional

zoddicus: Done

ben-clayton (approved):

PR #2084, opened 2022-12-15
Add a powerPreference option

greggman:
Not sure how to add this here's an attempt.

You can pass `powerPreference=low-power` or
`power-preference=high-performance` in the query string.

Issue: #2083

<hr>

### DIFF
--- a/src/common/util/navigator_gpu.ts
+++ b/src/common/util/navigator_gpu.ts
@@ -32,6 +32,15 @@ export function setGPUProvider(provider: GPUProvider) {
 
 let impl: GPU | undefined = undefined;
 
+let defaultRequestAdapterOptions: GPURequestAdapterOptions | undefined;
+
+export function setDefaultRequestAdapterOptions(options: GPURequestAdapterOptions) {
+  if (impl) {
+    throw new Error('must call setDefaultRequestAdapterOptions before getGPU');
+  }
+  defaultRequestAdapterOptions = { ...options };
+}
+
 /**
  * Finds and returns the `navigator.gpu` object (or equivalent, for non-browser implementations).
  * Throws an exception if not found.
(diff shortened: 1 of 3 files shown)


src/common/util/navigator_gpu.ts
```diff
@@ -43,5 +52,23 @@ export function getGPU(): GPU {
 
   impl = gpuProvider();
 
+  if (defaultRequestAdapterOptions) {
+    // eslint-disable-next-line @typescript-eslint/unbound-method
+    const oldFn = impl.requestAdapter;
+    impl.requestAdapter = function (
```
kainino0x, line 58: Normally I'd avoid interposing on the API since it means tests aren't running against the implementation itself, but it's probably fine since it only happens when you explicitly enable it. There are some tests which this will have slightly surprising effects on though, specifically `device_allocation.spec.ts` and `requestAdapter.spec.ts`.

The other option would be to change the requestAdapter call in `device_pool.ts`, which would affect _most_ tests but not the few that call requestAdapter themselves.

kainino0x (approved):
LGTM, see comment about possible alternative

src/common/util/navigator_gpu.ts
greggman: So I looked into adding it to device_pool.ts and it wasn't clear to me how to make it work. First I tried adding a `requestAdapterOptions` to `DevicePool` and that needed to be passed down to `DescriptorToHolderMap` and then to `DeviceHolder`. Then I tried pass in a default at the top where it creates the device pool singleton. But, you obviously can't have a function to set the defualt in that file since the function needs to be called before the singleton is created.

I put the functions to set (and get) the default options in a separate file alongsite `device_pool.ts` but when I tried to call it from `standalone.ts` I got an error calling into that part of the tree is off-limits.

Maybe there's an easier way but for now I decided to leave it as is. I'm happy to refactor into `device_pool.ts` with some guidance on how to do it.

PR #2093, opened 2022-12-20
Test beginComputePass/RenderPass after encoder.finish() in call_after_successful_finish

shrekshao:
Issue: #1914

Latest spec says beginComputePass/RenderPass generate validation error when `encoder.state` is ended. Add calling cmd of beginComputePass/RenderPass to the existing `call_after_successful_finish` test which used to only call `copyBufferToBuffer`

<hr>

### DIFF
--- a/src/webgpu/api/validation/encoding/encoder_state.spec.ts
+++ b/src/webgpu/api/validation/encoding/encoder_state.spec.ts
@@ -111,31 +111,41 @@ g.test('call_after_successful_finish')
     const pass = passType === 'compute' ? encoder.beginComputePass() : t.beginRenderPass(encoder);
     pass.end();
 
-    const callFn = () => {
-      switch (callCmd) {
-        case 'beginComputePass':
-          encoder.beginComputePass();
-          break;
-        case 'beginRenderPass':
-          t.beginRenderPass(encoder);
-          break;
-        case 'insertDebugMarker':
-          encoder.insertDebugMarker('');
-          break;
-        default:
-          unreachable();
-      }
-    };
-
     if (IsEncoderFinished) {
       encoder.finish();
-      t.expectValidationError(() => {
-        callFn();
-      }, IsEncoderFinished);
-    } else {
-      t.expectValidationError(() => {
-        callFn();
-      }, IsEncoderFinished);
+    }
+
+    switch (callCmd) {
+      case 'beginComputePass':
+        {
+          let pass: GPUComputePassEncoder;
+          t.expectValidationError(() => {
+            pass = encoder.beginComputePass();
+          }, IsEncoderFinished);
+          t.expectValidationError(() => {
+            pass.end();
+          }, IsEncoderFinished);
+        }
+        break;
+      case 'beginRenderPass':
+        {
+          let pass: GPURenderPassEncoder;
+          t.expectValidationError(() => {
+            pass = t.beginRenderPass(encoder);
+          }, IsEncoderFinished);
+          t.expectValidationError(() => {
+            pass.end();
+          }, IsEncoderFinished);
+        }
+        break;
+      case 'insertDebugMarker':
+        t.expectValidationError(() => {
+          encoder.insertDebugMarker('');
+        }, IsEncoderFinished);
+        break;
+    }
+
+    if (!IsEncoderFinished) {
       encoder.finish();
     }
   });
--- a/src/webgpu/api/validation/encoding/encoder_state.spec.ts
+++ b/src/webgpu/api/validation/encoding/encoder_state.spec.ts
@@ -17,21 +17,28 @@ TODO:
 `;
 
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
-import { objectEquals } from '../../../../common/util/util.js';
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/encoding/encoder_state.spec.ts
```diff
@@ -17,21 +17,28 @@ TODO:
 `;
 
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
-import { objectEquals } from '../../../../common/util/util.js';
+import { assert, objectEquals, unreachable } from '../../../../common/util/util.js';
 import { ValidationTest } from '../validation_test.js';
 
 class F extends ValidationTest {
+  private static attachmentTexture: GPUTexture | null = null;
+  private static attachmentTextureView: GPUTextureView | null = null;
```
kainino0x, line 25: It's not guaranteed each test/case will use the same GPUDevice, so we should avoid caching like this. Is it measurably slower without the caching? Likely not, but it is maybe we could implement this with a WeakMap with GPUDevice keys.

src/webgpu/api/validation/encoding/encoder_state.spec.ts
```diff
@@ -17,21 +17,28 @@ TODO:
...
-    const attachmentTexture = this.device.createTexture({
-      format: 'rgba8unorm',
-      size: { width: 16, height: 16, depthOrArrayLayers: 1 },
-      usage: GPUTextureUsage.RENDER_ATTACHMENT,
-    });
-    this.trackForCleanup(attachmentTexture);
+    if (!F.attachmentTexture) {
+      F.attachmentTexture = this.device.createTexture({
+        format: 'rgba8unorm',
+        size: { width: 16, height: 16, depthOrArrayLayers: 1 },
```
kainino0x, line 31: nit: Since it's just for a validation test this can be 1x1.

src/webgpu/api/validation/encoding/encoder_state.spec.ts
```diff
@@ -17,21 +17,28 @@ TODO:
...
+
   beginRenderPass(commandEncoder: GPUCommandEncoder): GPURenderPassEncoder {
-    const attachmentTexture = this.device.createTexture({
-      format: 'rgba8unorm',
-      size: { width: 16, height: 16, depthOrArrayLayers: 1 },
-      usage: GPUTextureUsage.RENDER_ATTACHMENT,
-    });
-    this.trackForCleanup(attachmentTexture);
+    if (!F.attachmentTexture) {
+      F.attachmentTexture = this.device.createTexture({
```
kainino0x, line 29: This createTexture will be caught inside the expectValidationError call. Not a big deal since it should never fail, but in principle not ideal.

src/webgpu/api/validation/encoding/encoder_state.spec.ts
```diff
@@ -110,15 +121,27 @@ g.test('call_after_successful_finish')
...
-      t.expectValidationError(() => {
-        encoder.copyBufferToBuffer(srcBuffer, 0, dstBuffer, 0, 0);
-      }, IsEncoderFinished);
+      t.expectValidationError(callFn);
     } else {
-      t.expectValidationError(() => {
-        encoder.copyBufferToBuffer(srcBuffer, 0, dstBuffer, 0, 0);
-      }, IsEncoderFinished);
+      callFn();
       encoder.finish();
```
kainino0x, line 149: The finish call was failing because the pass is open. Pushed a commit which closes it before finishing.

src/webgpu/api/validation/encoding/encoder_state.spec.ts
```diff
@@ -89,36 +96,56 @@ g.test('pass_end_invalid_order')
...
+      .combine('IsEncoderFinished', [false, true])
   )
   .fn(async t => {
-    const { passType, IsEncoderFinished } = t.params;
+    const { passType, IsEncoderFinished, callCmd } = t.params;
 
     const encoder = t.device.createCommandEncoder();
 
     const pass = passType === 'compute' ? encoder.beginComputePass() : t.beginRenderPass(encoder);
     pass.end();
```
kainino0x, line 112: This pass isn't really doing anything, technically could affect internal state but wouldn't generally expect it to change anything.
We could either remove it entirely, or we could add a third `passType` option where there is no pre-pass if you think it might be valuable coverage.

also nit: the name passType is a little confusing, could call it prePassType

kainino0x (commented):
Changed a few things in commits (after experimenting locally to figure out what to suggest), and have a few more comments

src/webgpu/api/validation/encoding/encoder_state.spec.ts
shrekshao: The intention is to avoid calling `createView` inside trying to catch validation error in beginRenderPass (breaking the rule that `expectValidationError` should only contain one GPU API call)
But as pointed out in the later comment, the first call to beginRenderPass will still include calling `createView`.
Since performance is not the concern here, rethinking it a bit, I should better create the texture and textureView in overrided `init` for this test set F

kainino0x (approved):

PR #2094, opened 2022-12-21
op: Update 'stencil_compare_func' test in stencil.spec.ts

Gyuyoung:
This PR updates the 'stencil_compare_func' test to ensure that other stencil texture formats
can also work on the test.

Issue: #2023

<hr>

### DIFF
--- a/src/webgpu/api/operation/rendering/stencil.spec.ts
+++ b/src/webgpu/api/operation/rendering/stencil.spec.ts
@@ -4,6 +4,7 @@ Test related to stencil states, stencil op, compare func, etc.
 
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
 import { TypedArrayBufferView } from '../../../../common/util/util.js';
+import { kDepthStencilFormats, kTextureFormatInfo } from '../../../capability_info.js';
 import { GPUTest } from '../../../gpu_test.js';
 import { TexelView } from '../../../util/texture/texel_view.js';
 import { textureContentIsOKByT2B } from '../../../util/texture/texture_ok.js';
@@ -75,10 +76,9 @@ class StencilTest extends GPUTest {
   checkStencilCompareFunction(
     compareFunction: GPUCompareFunction,
     stencilRefValue: number,
-    expectedColor: Float32Array
+    expectedColor: Float32Array,
+    depthStencilFormat: GPUTextureFormat = 'depth24plus-stencil8'
   ) {
-    const depthStencilFormat: GPUTextureFormat = 'depth24plus-stencil8';
-
     const baseStencilState = {
       compare: 'always',
       failOp: 'keep',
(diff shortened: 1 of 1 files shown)


src/webgpu/api/operation/rendering/depth.spec.ts
```diff
@@ -547,3 +547,16 @@ g.test('reverse_depth')
       }
     );
   });
+
+g.test('create_pipeline_with_all_formats')
```
Gyuyoung, line 551: I wrote this test referring to `DepthStencilStateTest.CreatePipelineWithAllFormats` test[1]. But, I'm not fully sure if adding this test is useful. Please let me know if this test is useless here. 

[1] https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/src/dawn/tests/end2end/DepthStencilStateTests.cpp;l=763?q=depthstencil&ss=chromium

src/webgpu/api/operation/rendering/depth.spec.ts
```diff
@@ -547,3 +547,16 @@ g.test('reverse_depth')
...
+  .desc('Test that creating a render pipeline works with for all depth and combined formats.')
+  .params(u =>
+    u
+      .combine('depthStencilFormat', kDepthStencilFormats)
+      .filter(({ depthStencilFormat }) => depthStencilFormat !== 'depth32float-stencil8')
```
Gyuyoung, line 556: It looks like the 'depth32float-stencil8' format can't create a render pipeline yet.

kainino0x: It can be supported, it just requires a device feature flag to be enabled. Pushed a commit fixing this.

src/webgpu/api/operation/rendering/depth.spec.ts
kainino0x: It's useful on its own, but it doesn't do much since it doesn't actually use the pipeline, just creates it. I'd prefer to have coverage of this through tests that actually test rendering. `depth.spec.ts` already has a test with cases over all formats with a depth aspect. If we can parameterize one or more similar tests in `stencil.spec.ts` to all formats with a stencil aspect - which we should have anyway - then that'll be much stronger than this test.

Gyuyoung: ok, then it's likely it would be good to make the `stencil_compare_func` test on other stencil texture formats like the `depth_compare_func` test in depth.spec.ts.

kainino0x (commented):
Apologies, I forgot to check if you had done this and I did something similar in #2096 but for all of the stencil tests. Can you take a look?

kainino0x (approved):
Actually I'll land this and rebase that over it. LGTM

Gyuyoung:
> Actually I'll land this and rebase that over it. LGTM

yes, your PR looks like covering this PR. Oh, you already merged this PR. Thanks!

Issue #2097, opened 2022-12-23
Support Dark Mode


greggman:
working in a dark space it's nice to not have my laptop light up the room 😄

Issue #888, opened 2022-01-13
op: commmand_buffer,occlusion_queries


kainino0x:

greggman:
Just checking that there are zero tests that queries actually work in the CTS. I see lots of validation tests but no operation tests

kainino0x:
This seems likely, yes.

PR #2098, opened 2022-12-23
Support dark mode

greggman:
Here's a stab at adding support for dark mode.

I removed the images in the buttons and changed them to unicode characters. Unfortunately I didn't find a good icon for "log". There's a unicode "document" icon but apparently it's not implemented in many fonts. There's also a emoji icon for "document" but doesn't look good in light vs dark.

I'm happy to change them back to images but it will require different icons for light and dark (I think). I thought about changing the icons to white on transparent and then using blend-mode: difference but the buttons are `#888` so I don't think that would work.

Issue: #2097

<hr>

### DIFF
--- a/standalone/index.html
+++ b/standalone/index.html
@@ -17,6 +17,66 @@
     <link rel="stylesheet" href="third_party/normalize.min.css" />
     <script src="third_party/jquery/jquery-3.3.1.min.js"></script>
     <style>
+      :root {
+        color-scheme: light dark;
+
+        --fg-color: #000;
+        --bg-color: #fff;
+        --border-color: #888;
+
+        --results-fg-color: gray;
+        --node-description-fg-color: #gray;
+        --node-hover-bg-color: rgba(0, 0, 0, 0.1);
+
+        --testcaselogbtn-bg-color: #eee;
+        --subtree-border-color: #ddd;
+        --subtree-hover-left-border-color: #000;
+        --multicase-border-color: #55f;
+        --testcase-border-color: #bbf;
+        --testcase-bg-color: #bbb;
+
+        --testcase-data-status-fail-bg-color: #fdd;
+        --testcase-data-status-warn-bg-color: #ffb;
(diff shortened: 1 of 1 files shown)


kainino0x:
> I'm happy to change them back to images but it will require different icons for light and dark (I think). I thought about changing the icons to white on transparent and then using blend-mode: difference but the buttons are `#888` so I don't think that would work.

How about `filter: invert(1);`?

standalone/index.html
```diff
@@ -224,31 +290,31 @@
...
+        background: var(--testcase-log-odd-bg-color);
       }
       .testcaselog:nth-child(even) {
-        background: #f8f8f8;
+        background: var(--testcase-log-even-bg-color);
       }
-      .testcaselogbtn {
-        background-image: url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAwAAAAMAQMAAABsu86kAAAABlBMVEUAAAAAAAClZ7nPAAAAAXRSTlMAQObYZgAAACRJREFUCNdjYGBg+H+AwUGBwV+BQUGAQX0CiNQQYFABk8ogLgBsYQUt2gNKPwAAAABJRU5ErkJggg==);
+      .testcaselogbtn::before {
+        content: "🪵"
```
kainino0x, line 306: Maybe 🖨️ or 🧾

kainino0x (approved):
Thanks! I've just been relying on Dark Reader instead of actually fixing this 😅 

I like my pixel art buttons but I'm not strongly attached to them. The main reason I used them instead of emoji is that emoji render very inconsistently across systems and often end up too small or too big for the button. That could be fixed by loading an emoji font but I didn't want to do that.

If the CSS invert thing doesn't work well then emoji are fine. Added suggestions for the log button.

standalone/index.html
greggman: filter worked following [this](https://css-tricks.com/apply-a-filter-to-a-background-image/)

PR #2080, opened 2022-12-15
Buffer.mapAsync pending map error shouldn't be a validation error

takahirox:
From: https://github.com/gpuweb/cts/issues/2009#issuecomment-1350608028

The current `Buffer.mapAsync` pending map error (mapAsync request on non-null [[pending_map]] error) tests expect a validation error but it doesn't produce a validation error in the WebGPU spec.

Issue: #2009

<hr>

**Requirements for PR author:**

- [x] All missing test coverage is tracked with "TODO" or `.unimplemented()`.
- [x] New helpers are `/** documented */` and new helper files are found in `helper_index.txt`.
- [x] Test behaves as expected in a WebGPU implementation. (If not passing, explain above.)

**Requirements for [reviewer sign-off](https://github.com/gpuweb/cts/blob/main/docs/reviews.md):**

- [x] Tests are properly located in the test tree.
- [x] [Test descriptions](https://github.com/gpuweb/cts/blob/main/docs/intro/plans.md) allow a reader to "read only the test plans and evaluate coverage completeness", and accurately reflect the test code.
- [ ] Tests provide complete coverage (including validation control cases). **Missing coverage MUST be covered by TODOs.**
- [x] Helpers and types promote readability and maintainability.

When landing this PR, be sure to make any necessary issue status updates.

Issue #2009, opened 2022-11-20
Add CTS test for early-reject mapAsync


takahirox:
A CTS test for early-reject mapAsync() should be added.

https://github.com/gpuweb/gpuweb/pull/3348

kainino0x:
I rechecked the spec and it reminds me, immediate rejection only happens if a map is pending. We should also test that the promise *isn't* already-rejected if the buffer was already mapped. That is, the rejection should occur in a later task instead of in a microtask.

This is probably just a matter of adding a check to `testMapAsyncCall` that, in a microtask, the rejection hasn't happened yet. This can be a separate PR from #2033.

takahirox:
> immediate rejection only happens if a map is pending. We should also test that the promise isn't already-rejected if the buffer was already mapped. That is, the rejection should occur in a later task instead of in a microtask.

Good catch, thanks.

> This is probably just a matter of (conditially) adding a check to testMapAsyncCall that, in a microtask, the rejection hasn't happened yet. This can be a separate PR from https://github.com/gpuweb/cts/pull/2033.

I made #2052. With the change [this test](https://github.com/gpuweb/cts/blob/fdc34f66b091435efbf2902a0034ee0ac0f8e0cc/src/webgpu/api/validation/buffer/mapping.spec.ts#L176) will check that `mapAsync()` doesn't reject immediately on buffers already mapped.

takahirox:
I noticed that early-rejection test needs to be updated because pending map error is not a validation error although the current test expects pending map error as a validation error. Please refer to https://bugs.chromium.org/p/dawn/issues/detail?id=1613

I will make a PR.

takahirox:
I also noticed that these tests need to be updated
* webgpu:api,operation,buffers,map_oom:mapAsync
* webgpu:api,validation,buffer,mapping:getMappedRange,state,mapped
* webgpu:api,validation,buffer,mapping:getMappedRange,state,mappedAtCreation

For example `await buffer;` needs to be inserted at [here](https://github.com/gpuweb/cts/blob/82ec26d039e19c70f6fa979bbff03b82d4b6ff7f/src/webgpu/api/operation/buffers/map_oom.spec.ts#L57).

See https://dawn-review.googlesource.com/c/dawn/+/113607/comments/09b70269_59aa698e

I will work on it after #2052 and #2080 will be merged.

takahirox:
@kainino0x Friendly ping for the PRs

kainino0x:
[...]

PR #2101, opened 2022-12-26
Improve Buffer.mapAsync tests

takahirox:
This PR improves `Buffer.mapAsync()` tests.

* Better error timing check.
* Test rejection with whether validation error.
* Follow the latest WebGPU API specification. Unmap abort error should have precedence over validation error.

This PR resolves

* https://github.com/gpuweb/cts/pull/2080#discussion_r1054967344
* https://dawn-review.googlesource.com/c/dawn/+/113607/comments/09b70269_59aa698e

Issue: #2009 

<hr>

### DIFF
--- a/src/webgpu/api/operation/buffers/map_oom.spec.ts
+++ b/src/webgpu/api/operation/buffers/map_oom.spec.ts
@@ -32,10 +32,10 @@ g.test('mapAsync')
     oomAndSizeParams //
       .beginSubcases()
       .combine('write', [false, true])
-      .combine('abortError', [false, true])
+      .combine('unmapBeforeResolve', [false, true])
   )
   .fn(async t => {
-    const { oom, write, size, abortError } = t.params;
+    const { oom, write, size, unmapBeforeResolve } = t.params;
 
     const buffer = t.expectGPUError(
       'out-of-memory',
@@ -55,7 +55,7 @@ g.test('mapAsync')
     }, oom);
 
     if (oom) {
-      if (abortError) {
+      if (unmapBeforeResolve) {
         // Should reject with abort error because buffer will be unmapped
         // before validation check finishes.
         t.shouldReject('AbortError', promise!);
--- a/src/webgpu/api/operation/buffers/map_oom.spec.ts
+++ b/src/webgpu/api/operation/buffers/map_oom.spec.ts
@@ -32,9 +32,10 @@ g.test('mapAsync')
     oomAndSizeParams //
       .beginSubcases()
       .combine('write', [false, true])
+      .combine('abortError', [false, true])
   )
   .fn(async t => {
-    const { oom, write, size } = t.params;
+    const { oom, write, size, abortError } = t.params;
 
     const buffer = t.expectGPUError(
       'out-of-memory',
@@ -45,15 +46,27 @@ g.test('mapAsync')
         }),
       oom
     );
-    const promise = t.expectGPUError(
-      'validation', // Should be a validation error since the buffer is invalid.
-      () => buffer.mapAsync(write ? GPUMapMode.WRITE : GPUMapMode.READ),
-      oom
-    );
+
+    let promise: Promise<void>;
+    // Should be a validation error since the buffer is invalid.
+    // Unmap abort error shouldn't cause a validation error.
+    t.expectValidationError(() => {
+      promise = buffer.mapAsync(write ? GPUMapMode.WRITE : GPUMapMode.READ);
+    }, oom);
 
     if (oom) {
-      // Should also reject in addition to the validation error.
-      t.shouldReject('OperationError', promise);
+      if (abortError) {
+        // Should reject with abort error because buffer will be unmapped
+        // before validation check finishes.
+        t.shouldReject('AbortError', promise!);
+      } else {
+        // Should also reject in addition to the validation error.
+        t.shouldReject('OperationError', promise!);
+
+        // Wait for validation error before unmap to ensure validation check
+        // ends before unmap.
+        await promise!;
+      }
 
       // Should throw an OperationError because the buffer is not mapped.
       // Note: not a RangeError because the state of the buffer is checked first.
@@ -64,7 +77,7 @@ g.test('mapAsync')
       // Should't be a validation error even if the buffer failed to be mapped.
       buffer.unmap();
     } else {
-      await promise;
+      await promise!;
       const arraybuffer = buffer.getMappedRange();
       t.expect(arraybuffer.byteLength === size);
       buffer.unmap();
--- a/src/webgpu/api/operation/buffers/map_oom.spec.ts
+++ b/src/webgpu/api/operation/buffers/map_oom.spec.ts
@@ -65,7 +65,12 @@ g.test('mapAsync')
 
         // Wait for validation error before unmap to ensure validation check
         // ends before unmap.
-        await promise!;
+        try {
+          await promise!;
+          throw new Error('The promise should be rejected.');
+        } catch {
+          // Should cause an exception because the promise should be rejected.
+        }
       }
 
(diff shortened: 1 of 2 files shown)


src/webgpu/api/validation/buffer/mapping.spec.ts
```diff
@@ -311,10 +335,42 @@ g.test('mapAsync,earlyRejection')
...
+      // unmap abort error should have precedence over validation error
+      const pending = t.testMapAsyncCall(false, false, false, 'AbortError', buffer, mapMode);
+      buffer.unmap();
+      await pending;
+    } else {
+      // map on already mapped buffer should cause validation error
+      await t.testMapAsyncCall(false, true, false, 'OperationError', buffer, mapMode);
+      buffer.unmap();
+    }
+  });
```
takahirox, line 372: Do you think we should test that unmap should have precedence over not only "map on already mapped buffer" but also other invalidation errors eg: "map async request on invalid buffer", "offset is not a multiple of 8"? Or does it sound duplicated?

src/webgpu/api/operation/buffers/map_oom.spec.ts
```diff
@@ -45,15 +46,27 @@ g.test('mapAsync')
...
+        // Should reject with abort error because buffer will be unmapped
+        // before validation check finishes.
+        t.shouldReject('AbortError', promise!);
+      } else {
+        // Should also reject in addition to the validation error.
+        t.shouldReject('OperationError', promise!);
+
+        // Wait for validation error before unmap to ensure validation check
+        // ends before unmap.
+        await promise!;
```
kainino0x, line 68: Since promise rejects, this should throw an exception and fail the test...
and it does, if I change it to use cases instead of subcases (remove `.beginSubcases()`).
Turns out there is a harness error. I put up a fix here, PTAL:
https://github.com/gpuweb/cts/pull/2102

src/webgpu/api/validation/buffer/mapping.spec.ts
```diff
@@ -12,6 +12,7 @@ import { ValidationTest } from '../validation_test.js';
 class F extends ValidationTest {
   async testMapAsyncCall(
     success: boolean,
+    validationError: boolean,
     earlyRejection: boolean,
     rejectName: string | null,
```
kainino0x, line 17: This is becoming a bit of a "boolean soup" where it's hard to see what the arguments are at the call sites.
Can we replace it with something a bit richer? Example:

```suggestion
    expectation: 'success' | { validationError: boolean, earlyRejection: boolean, rejectName: string },
```

kainino0x, line 17: This is becoming a bit of a "boolean soup" where it's hard to see what the arguments are at the call sites.
Can we replace it with something a bit richer? Example:

```suggestion
    expectation: 'success' | { validationError: boolean, earlyRejection: boolean, rejectName: string },
```

src/webgpu/api/validation/buffer/mapping.spec.ts
kainino0x: I think we should. It's reasonable for implementation to lift some validation to the content process, for example "offset is not a multiple of 8" can be validated easily in the content process, whereas "invalid buffer" is impossible to validate in the content process.

kainino0x (commented):
I think these changes look about correct, but I want to do my full review after rebasing over #2102 and after the suggested change that will make this much easier to read. Thank you for the thorough tests!

src/webgpu/api/validation/buffer/mapping.spec.ts
takahirox: Thanks for the suggestion. Good idea. Updated.

src/webgpu/api/validation/buffer/mapping.spec.ts
takahirox: > I think we should. 

OK, I added a TODO for now.

src/webgpu/api/operation/buffers/map_oom.spec.ts
```diff
@@ -32,9 +32,10 @@ g.test('mapAsync')
     oomAndSizeParams //
       .beginSubcases()
       .combine('write', [false, true])
+      .combine('abortError', [false, true])
```
Kangz, line 35: nit: unmapBeforeResolve ? It makes it more clear what the test checks.

src/webgpu/api/operation/buffers/map_oom.spec.ts
```diff
@@ -45,15 +46,32 @@ g.test('mapAsync')
...
-    const promise = t.expectGPUError(
-      'validation', // Should be a validation error since the buffer is invalid.
-      () => buffer.mapAsync(write ? GPUMapMode.WRITE : GPUMapMode.READ),
-      oom
-    );
+
+    let promise: Promise<void>;
+    // Should be a validation error since the buffer is invalid.
+    // Unmap abort error shouldn't cause a validation error.
+    t.expectValidationError(() => {
```
Kangz, line 53: Maybe we could make `expectValidationError` generic on the return type of `fn` like `expectGPUError` so that we can do:

```
const promise = t.expectValidationError(()=> buffer.mapAsync(,,,));
```

and we don't need the addition of `!` below.

src/webgpu/api/validation/buffer/mapping.spec.ts
```diff
@@ -11,22 +11,22 @@ import { ValidationTest } from '../validation_test.js';
...
     offset?: number,
     size?: number
   ) {
-    if (success) {
+    if (expectation === 'success') {
       const p = buffer.mapAsync(mode, offset, size);
       await p;
     } else {
       let p: Promise<void>;
       this.expectValidationError(() => {
```
Kangz, line 27: ditto

src/webgpu/api/validation/buffer/mapping.spec.ts
```diff
@@ -311,10 +375,56 @@ g.test('mapAsync,earlyRejection')
...
 
+g.test('mapAsync,abort_over_invalid_error')
+  .desc(
+    `Test that unmap abort error should have precedence over validation error
+TODO
+  - Add other validation error test (eg. offset is not a multiple of 8)
+  `
+  )
+  .paramsSubcasesOnly(u =>
+    u.combine('mapMode', kMapModeOptions).combine('abortError', [true, false])
```
Kangz, line 400: ditto about abortError

Kangz (approved):

src/webgpu/api/operation/buffers/map_oom.spec.ts
kainino0x: I've considered this but avoided it because it breaks a sanity check we have right now that the expectValidationError callback doesn't return a Promise. This is because if you write an async function there it _looks_ like the scope is around the whole thing, but it's only up to the first await.

However I did have an idea when reviewing this PR the other day, we could change it to check whether the function itself is async instead of checking the return value. People online say not to do this because async function often get transpiled into non-async functions, but in our case our build steps never do that.

In any case it's be complicated enough I think it should be a followup.

src/webgpu/api/operation/buffers/map_oom.spec.ts
takahirox: Thanks for the suggestion. Renamed to clarify, thanks.

src/webgpu/api/operation/buffers/map_oom.spec.ts
takahirox: Thanks for the suggesion. Let's make another PR for improvement.

takahirox:
Merging this PR because it is approved by two reviewers and I reflected the suggestiong.

PR #2102, opened 2022-12-28
Fix silencing of DOMExceptions

kainino0x:
Quick fix, but I'll put up a suggested cleanup (just need Austin's review to make sure that will work).

The bug was that `rec.threw()` would never be called for `DOMException`s, causing the test to erroneously pass.

Issue: Regressed in #2086, found in #2101

<hr>

### DIFF
--- a/src/common/internal/test_group.ts
+++ b/src/common/internal/test_group.ts
@@ -536,7 +536,12 @@ class RunCaseSpecific implements RunCase {
                       // Prepend the subcase name to all error messages.
                       for (const arg of args) {
                         if (arg instanceof Error) {
-                          arg.message = subcasePrefix + '\n' + arg.message;
+                          try {
+                            arg.message = subcasePrefix + '\n' + arg.message;
+                          } catch {
+                            // Silence exception if the property isn't settable
+                            // (e.g. arg.message on DOMException).
+                          }
                         }
                       }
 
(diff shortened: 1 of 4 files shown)


kainino0x:
The bug was that `rec.threw()` would never be called for `DOMException`s, causing the test to erroneously pass.

austinEng (approved):
Intent was to improve messages since with subcases, the logs look like:

```
INFO: subcase 1
INFO: subcase 2
INFO: subcase 3
INFO: subcase 4

EXPECTATION FAILED:
reason
```

and you don't know which subcase the failure came from

I guess we can't have that for DOMExceptions though?

src/common/internal/test_group.ts
```diff
@@ -536,7 +536,12 @@ class RunCaseSpecific implements RunCase {
                       // Prepend the subcase name to all error messages.
                       for (const arg of args) {
                         if (arg instanceof Error) {
-                          arg.message = subcasePrefix + '\n' + arg.message;
+                          try {
+                            arg.message = subcasePrefix + '\n' + arg.message;
+                          } catch {
+                            // Silence exception if the property isn't settable
+                            // (e.g. arg.message on DOMException).
+                          }
```
takahirox, line 544: Honestly I'm not sure what this try-catch is for. Does setting non-settable (read-only) property throw an exception? But it doesn't although I tried the following example code on my Chrome console.

```
const e = new DOMException('aaa');
try {
  e.message = 'bbb' + e.message;
  console.log('no exception');
} catch {
  console.log('exception');
}
console.log(e.message);

Output:
no exception
aaa // read-only so no change
```

takahirox: Ah, ok TypeScript(?) seems to check it so the change makes sense to me.

image

takahirox (approved):

src/common/internal/test_group.ts
kainino0x: There is an exception when `"use strict";` is enabled. I got confused by this too :)

PR #2103, opened 2022-12-28
Put subcase in the stack instead of the message

kainino0x:
`DOMException.stack` is settable, while `.message` isn't. Guessing there are no errors where `message` is settable and `stack` isn't, so this seems like a better place to put custom info. (It's kind of "stack"ish anyway.)

Question is, will this be surfaced in the places we want to see it? I don't know what context prompted the original PR, and whether it prints the message or the stack.

Issue: See also #2102

<hr>

### DIFF
--- a/src/common/internal/test_group.ts
+++ b/src/common/internal/test_group.ts
@@ -536,8 +536,10 @@ class RunCaseSpecific implements RunCase {
                       // Prepend the subcase name to all error messages.
                       for (const arg of args) {
                         if (arg instanceof Error) {
+                          let stack = subcasePrefix;
+                          if (arg.stack) stack += '\n' + arg.stack;
                           try {
-                            arg.message = subcasePrefix + '\n' + arg.message;
+                            arg.stack = stack;
                           } catch {
                             // Silence exception if the property isn't settable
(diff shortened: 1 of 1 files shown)


austinEng:
hmm looks like you don't see the subcase names in the logs which was the intent of the other PR

This PR:

Screenshot 2023-01-03 at 10 44 39 AM

Current shows OK or EXPECTATION FAILED labeled with the subcase

Screenshot 2023-01-03 at 10 46 12 AM

kainino0x:
Ah... guessing it's because of the "(elided: below max severity)". Maybe we could just stick it in an expando property as long as we make sure to print it out properly? I'll think about it more.

kainino0x:
(thinking aloud) Could we just detangle the logs before printing them? Maybe this would prevent logs from getting printed before crashes which would be bad

austinEng:
Hmm.. could we append to LogMessageWithStack.name? or add another member to LogMessageWithStack?

kainino0x:
I'm not sure of a simple way to do this and I don't think it warrants a lot of complexity. Opening alternative #2144

PR #2123, opened 2023-01-05
Unclutter the URL

greggman:
The URL is getting long as new options are added. It's common to want to edit the q parameter but as more options are added this gets harder as it get pushed off the to the right.

So, only show options that are non-falsy.

I wasn't sure this was ok. I kind of assumed you wanted to see all the options. But, as more get added, it's getting too verbose. The docs should show the available options.

<hr>

### DIFF
--- a/src/common/runtime/standalone.ts
+++ b/src/common/runtime/standalone.ts
@@ -12,7 +12,7 @@ import { TestTreeNode, TestSubtree, TestTreeLeaf, TestTree } from '../internal/t
 import { setDefaultRequestAdapterOptions } from '../util/navigator_gpu.js';
 import { assert, ErrorWithExtra, unreachable } from '../util/util.js';
 
-import { optionEnabled } from './helper/options.js';
+import { optionEnabled, optionString } from './helper/options.js';
 import { TestWorker } from './helper/test_worker.js';
 
 window.onbeforeunload = () => {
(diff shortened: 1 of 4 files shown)


kainino0x:
I did this initially as a janky self-documenting alternative to checkboxes. If we remove it I'd like to add documenting text (or even real checkboxes) to the standalone html page. Then I think it would be OK to completely remove the URL canonicalization and rewriting, which would be nice because it would be less code instead of more. WDYT?

greggman:
SGTM. will refactor.

kainino0x:
Thanks!

greggman:
pushed a version with UI.

src/common/runtime/standalone.ts
```diff
@@ -22,18 +22,71 @@ window.onbeforeunload = () => {
...
+  }
+  return (optionValues as unknown) as StandaloneOptions;
+}
+
+// This is just a cast in one place.
+function optionsToRecord(options: StandaloneOptions) {
+  return (options as unknown) as Record<string, boolean | string>;
+}
+
+const options = getOptionsInfoFromSearchParameters(optionsInfo);
```
kainino0x, line 80: Looks like options are snapshotted on page load, so we should have a note that changing the options requires reload

standalone/index.html
```diff
@@ -367,6 +391,12 @@
   </head>
   <body>
     <h1><img class="logo" src="webgpu-logo-notext.svg">WebGPU Conformance Test Suite</h1>
+    <details>
+      <summary>options</summary>
```
kainino0x, line 395: maybe simply:
```suggestion
      <summary>options (requires reload!)</summary>
```

kainino0x (approved):
looks great, thanks!

src/common/runtime/standalone.ts
greggman: Doh! You're right. I kind of mindlessly added them. I'll change it to not be snapshotted.

standalone/index.html
greggman: done

src/common/runtime/standalone.ts
greggman: Ah, I see refactoring for them to be live is probably too much work. Just added the note as you suggested.

Issue #2124, opened 2023-01-06
op: Add Buffer.mapState test


takahirox:
We need tests for Buffer.mapState

https://github.com/gpuweb/gpuweb/pull/3014

takahirox:
As pointed in #2359, we need to add test read map + mapState cases. (Likely I was missing this case in #2130 and #2145.)

kainino0x:
Copying details from #2359 to the summary of this issue so it can be cleanly tracked here

PR #2130, opened 2023-01-09
Add Buffer.mapState test

takahirox:
Issue: #2124

This PR adds `Buffer.mapState` test.

https://github.com/gpuweb/gpuweb/pull/3014

I made new tests for `Buffer.mapState` rather than inserting `Buffer.mapState` check in the existing tests because I thought it can be simpler and cleaner. Please let me know if inserting in the existing tests is better.

And also please let me know if you have any simplification or better test coverage ideas.

<hr>

### DIFF
--- a/src/webgpu/api/operation/buffers/map.spec.ts
+++ b/src/webgpu/api/operation/buffers/map.spec.ts
@@ -386,20 +386,20 @@ g.test('mappedAtCreation,mapState')
     }, validationError);
 
     // mapState must be "mapped" regardless of validation error
-    assert(buffer!.mapState === 'mapped');
+    t.expect(buffer!.mapState === 'mapped');
 
     // getMappedRange must not change the map state
     buffer!.getMappedRange(...range);
-    assert(buffer!.mapState === 'mapped');
+    t.expect(buffer!.mapState === 'mapped');
 
     if (afterUnmap) {
       buffer!.unmap();
-      assert(buffer!.mapState === 'unmapped');
+      t.expect(buffer!.mapState === 'unmapped');
     }
 
     if (afterDestroy) {
       buffer!.destroy();
-      assert(buffer!.mapState === 'unmapped');
+      t.expect(buffer!.mapState === 'unmapped');
     }
   });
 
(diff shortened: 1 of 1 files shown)


src/webgpu/api/operation/buffers/map.spec.ts
```diff
@@ -362,3 +362,134 @@ still present in the mapped buffer.`
...
+
+g.test('mappedAtCreation,mapState')
+  .desc('Test that exposed map state of buffer created with mappedAtCreation has expected values.')
+  .params(u =>
+    u
+      .combine('validationError', [false, true])
+      .combine('afterUnmap', [false, true])
+      .combine('afterDestroy', [false, true])
+      .beginSubcases()
+      .combineWithParams(kSubcases)
```
kainino0x, line 374: Hm, I don't think the kSubcases are important to this test case. Feel free to keep if you think they might be useful (or if they're very fast), but I think we could omit them and make the test a little faster.

src/webgpu/api/operation/buffers/map.spec.ts
```diff
@@ -362,3 +362,134 @@ still present in the mapped buffer.`
...
+  .params(u =>
+    u
+      .combine('bufferCreationValidationError', [false, true])
+      .combine('mapAsyncValidationError', [false, true])
+      .combine('beforeUnmap', [false, true])
+      .combine('beforeDestroy', [false, true])
+      .combine('afterUnmap', [false, true])
+      .combine('afterDestroy', [false, true])
+      .beginSubcases()
+      .combineWithParams(kSubcases)
```
kainino0x, line 417: same

src/webgpu/api/operation/buffers/map.spec.ts
```diff
@@ -362,3 +362,134 @@ still present in the mapped buffer.`
...
+    // If buffer is already mapped test mapAsync on already mapped buffer
+    if (buffer!.mapState === 'mapped') {
+      // mapAsync on already mapped buffer won't change the map state
+      const promise = buffer!.mapAsync(GPUMapMode.WRITE);
+      assert(buffer!.mapState === 'mapped');
+
+      // mapAsync on already mapped buffer must reject and the map state must keep 'mapped'
+      try {
+        await promise;
+        throw new Error('mapAsync on already mapped buffer must not succeed.');
```
kainino0x, line 480: Here you can use the `t.expect(false)` shorthand `t.fail()`.

src/webgpu/api/operation/buffers/map.spec.ts
```diff
@@ -362,3 +362,134 @@ still present in the mapped buffer.`
...
+    assert(buffer!.mapState === 'unmapped');
+
+    {
+      const promise = buffer!.mapAsync(mapAsyncValidationError ? 0 : GPUMapMode.WRITE);
+      assert(buffer!.mapState === 'pending');
+
+      try {
+        if (beforeUnmap) {
+          buffer!.unmap();
+          assert(buffer!.mapState === 'unmapped');
```
kainino0x, line 451: If the assert fails, it will throw an exception, and that will be caught by the `catch` below. You can use `t.expect()`, which will log a failure immediately instead of relying on it to be caught by the harness.

Same for other assertions/throws. I think you can just replace all of them with `t.expect()`.

src/webgpu/api/operation/buffers/map.spec.ts
takahirox: Honestly I just followed the other test cases and have added `kSubcases` without deeply thinking. The suggestion sounds good to me. I don't think it's needed in these test cases. I have removed it, thanks.

src/webgpu/api/operation/buffers/map.spec.ts
takahirox: Thanks, updated.

src/webgpu/api/operation/buffers/map.spec.ts
takahirox: Thanks, updated.

kainino0x (approved):
LGTM!

src/webgpu/api/operation/buffers/map.spec.ts
```diff
@@ -362,3 +362,135 @@ still present in the mapped buffer.`
...
+        // unmapped before resolve, destroyed before resolve, or mapAsync validation error
+        // will end up with rejection and 'unmapped'
+        t.expect(buffer!.mapState === 'unmapped');
+      }
+    }
+
+    // If buffer is already mapped test mapAsync on already mapped buffer
+    if (buffer!.mapState === 'mapped') {
+      // mapAsync on already mapped buffer won't change the map state
+      const promise = buffer!.mapAsync(GPUMapMode.WRITE);
```
takahirox, line 475: I noticed that `expectValidationError()` should have been used to catch a validation error caused by mapping on already mapped buffer.

I made a PR to fix #2145

PR #2126, opened 2023-01-06
Test texture descriptor validation in GPUCanvasContext

Kangz:
The specification describes that the configure() and getCurrentTexture() both run GPUTextureDescriptor validation steps on the descriptor for the canvas textures. The only things in control of the application are:

 - the format (already tested)
 - the view formats
 - the size (through canvas.width/height)

This commit adds tests for the view formats and the size, checking for corner cases where the size is 0 before the configure() or becomes 0 after the configure.

Passing locally in Chromium with WIP CLs that I will put up for review shortly.

Issue: None, found opportunistically.

<hr>

**Requirements for PR author:**

- [X] All missing test coverage is tracked with "TODO" or `.unimplemented()`.
- [X] New helpers are `/** documented */` and new helper files are found in `helper_index.txt`.
- [X] Test behaves as expected in a WebGPU implementation. (If not passing, explain above.)

**Requirements for [reviewer sign-off](https://github.com/gpuweb/cts/blob/main/docs/reviews.md):**

- [ ] Tests are properly located in the test tree.
- [ ] [Test descriptions](https://github.com/gpuweb/cts/blob/main/docs/intro/plans.md) allow a reader to "read only the test plans and evaluate coverage completeness", and accurately reflect the test code.
- [ ] Tests provide complete coverage (including validation control cases). **Missing coverage MUST be covered by TODOs.**
- [ ] Helpers and types promote readability and maintainability.

When landing this PR, be sure to make any necessary issue status updates.

PR #2131, opened 2023-01-09
Transfer to image bitmap npe

Kangz:
Issue: #

<hr>

### DIFF
--- a/src/webgpu/web_platform/canvas/readbackFromWebGPUCanvas.spec.ts
+++ b/src/webgpu/web_platform/canvas/readbackFromWebGPUCanvas.spec.ts
@@ -329,3 +329,39 @@ g.test('drawTo2DCanvas')
 
     readPixelsFrom2DCanvasAndCompare(t, ctx, expect[t.params.alphaMode]);
   });
+
+g.test('transferToImageBitmap_unconfigured_nonzero_size')
+  .desc(
+    `Regression test for a crash when calling transferImageBitmap on an unconfigured. Case where the canvas is not empty`
+  )
+  .fn(t => {
+    const canvas = createCanvas(t, 'offscreen', 2, 3);
+    canvas.getContext('webgpu');
+
+    // Transferring gives an ImageBitmap of the correct size filled with transparent black.
+    const ib = canvas.transferToImageBitmap();
+    t.expect(ib.width === 2);
+    t.expect(ib.height === 3);
+
+    // TODO check the pixels are black somehow
+  });
+
+g.test('transferToImageBitmap_zero_size')
+  .desc(
+    `Regression test for a crash when calling transferImageBitmap on an unconfigured. Case where the canvas is empty.`
+  )
+  .params(u => u.combine('configure', [true, false]))
+  .fn(t => {
+    const { configure } = t.params;
+    const canvas = createCanvas(t, 'offscreen', 0, 1);
+    const ctx = canvas.getContext('webgpu')!;
+
+    if (configure) {
+      t.expectValidationError(() => ctx.configure({ device: t.device, format: 'bgra8unorm' }));
+    }
+
+    // Transferring would give an empty ImageBitmap which is not possible, so an Exception is thrown.
+    t.shouldThrow(true, () => {
+      canvas.transferToImageBitmap();
+    });
+  });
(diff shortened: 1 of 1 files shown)


src/webgpu/web_platform/canvas/readbackFromWebGPUCanvas.spec.ts
```diff
@@ -329,3 +329,13 @@ g.test('drawTo2DCanvas')
...
     readPixelsFrom2DCanvasAndCompare(t, ctx, expect[t.params.alphaMode]);
   });
+
+g.test('transferToImageBitmap_unconfigured')
+  .desc(`Regression test for a crash when calling transferImageBitmap on an unconfigured canvas`)
+  .fn(t => {
+    const canvas = createCanvas(t, 'offscreen', 2, 2);
+    canvas.getContext('webgpu');
+    t.shouldThrow(true, () => {
+      canvas.transferToImageBitmap();
```
kainino0x, line 339: I think this is supposed to return a blank image instead of throwing? See "When transferToImageBitmap() is called on a canvas with GPUCanvasContext context"

kainino0x (commented):
First commit is duplicated from #2126, can be rebased
Left comment on second commit

src/webgpu/web_platform/canvas/readbackFromWebGPUCanvas.spec.ts
Kangz: What should happen when the canvas size is empty? It looks it is no possible to create an empty image bitmap. Should we return null in that case?

Kangz: Tentatively assuming it should raise an exception.

Do you know the best way to check the contents of an ImageBitmap?

kainino0x: An exception seems correct for Wx0 or 0xH.

To check an imagebitmap I think you'll have to draw it into something, for example 2d canvas drawImage.

src/webgpu/web_platform/canvas/readbackFromWebGPUCanvas.spec.ts
```diff
@@ -329,3 +329,33 @@ g.test('drawTo2DCanvas')
...
+
+    // Transferring gives an ImageBitmap of the correct size filled with transparent black.
+    const ib = canvas.transferToImageBitmap();
+    t.expect(ib.width === 2);
+    t.expect(ib.height === 3);
+
+    // TODO check the pixels are black somehow
+  });
+
+g.test('transferToImageBitmap_unconfigured_zero_size')
```
kainino0x, line 349: Should fail regardless of whether it's configured. Maybe add a true-false param for whether to configure it before transferToImageBitmap?

src/webgpu/web_platform/canvas/readbackFromWebGPUCanvas.spec.ts
```diff
@@ -329,3 +329,33 @@ g.test('drawTo2DCanvas')
...
+
+g.test('transferToImageBitmap_unconfigured_nonzero_size')
+  .desc(
+    `Regression test for a crash when calling transferImageBitmap on an unconfigured. Case where the canvas is not empty`
+  )
+  .fn(t => {
+    const canvas = createCanvas(t, 'offscreen', 2, 3);
+    canvas.getContext('webgpu');
+
+    // Transferring gives an ImageBitmap of the correct size filled with transparent black.
```
kainino0x, line 341: If the `alphaMode` is `"opaque"` it should be opaque black - I think worth adding a test case

Kangz: How can there be an alphaMode since this has never been configured?

kainino0x: Right. didn't think very hard about it when reading the spec, will check it and make sure the spec makes sense here

kainino0x (approved):
nit: avoid force pushing rebases after initial reviews; I can't very easily see the [relative diff](https://github.com/gpuweb/cts/compare/ae7a71d76fd3db0bb125a1a7b237f8f098c98bff..2f76b8e7d71c0b026d63031c62f9702a47ee9bd6#diff-81b6f45e0dbdf222d34dc0d490575a0030480d65e83e554f2dcb6d25340115a3) from last time because github. 

LGTM when the TODO is done (or moved into the test description for later)

Kangz:
Thanks for the reminder, I need to switch from Gerrit habits :)

Added the readback if you'd like to check.

Issue #2132, opened 2023-01-09
reftest: test canvas with CSS background color


greggman:
There are tests for a canvas over a colored background but not for a canvas with it's own background color

PR #2133, opened 2023-01-09
Add more states to canvas composite ref tests.

greggman:
There are bugs related to setting the CSS background color of a canvas vs setting the background color of the body so I added a couple of test cases.

The 3 cases are now

1. A canvas on top of a colored body with imageRendering set at runtime by javascript

2. A canvas on top of a colored body with imageRendering set via CSS canvas rule

3. A canvas with background color set in CSS with imageRendering set via CSS canvas rule

This test fails in Chrome though AFAICT it was already failing

Issue: #2132

<hr>

### DIFF
--- a/src/webgpu/web_platform/reftests/canvas_composite_alpha.html.ts
+++ b/src/webgpu/web_platform/reftests/canvas_composite_alpha.html.ts
@@ -2,9 +2,6 @@ import { assert, unreachable } from '../../../common/util/util.js';
 
 import { runRefTest } from './gpu_ref_test.js';
 
-// <canvas> element from html page
-declare const cvs: HTMLCanvasElement;
-
 type WriteCanvasMethod = 'draw' | 'copy';
 
 export function run(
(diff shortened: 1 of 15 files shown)


kainino0x (commented):
I'm concerned about blocking these very core tests on correct implementation of the relatively-niche `image-rendering` property. We for example would probably block shipment on major failures of the original tests but not on fixing `image-rendering`.

The canvas background color is definitely core enough to include in these tests though.

Also I thought most of these tests were running and passing currently; expectations are here:
https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/web_tests/WebGPUExpectations

greggman:
They are passing by luck. 

Most likely the content-shell doesn't render the same as Chrome or the image comparision is fuzzy?

If run locally, zoom in, and refresh I see the canvas is clearly filtered

Screenshot 2023-01-10 at 1 25 09 PM

note: the tests were already relying on `image-rendering: pixelated`. 
See line 17 of every HTML file before this PR. 

Should we remove that?  And/or should we maybe make the display size 200x200 instead of 20x20 so it's easier to see

greggman:
interestingly the ref files don't have `image-rendering: pixelated`. I guess that's why it's passing?

kainino0x:
Oh, I forgot they were already using `pixelated`. I guess since the tests seemed to be passing we figured it was fine.
If we can remove it I think that would be good, especially since we know it's buggy.

At some point we specifically changed these tests so the canvas size and the css size would match. Previously it rendered 2x2 and scaled up to 20x20, and the scaling was offset by like half a pixel which made the test fail. I think we also tried fuzzy matching but it wasn't good at dealing with the offset.
Not sure why this doesn't raise issues on hidpi devices.

kainino0x:
200x200 would be fine, no opinion about that

greggman:
So I spend some time checking. 

The reason it passes is because the devicePixelRatio in content-shell is 1.0

So, we're rendering a non-filtered result into a non-scaled canvas so `pixelated` or not we get the same result. (so it's passing but luck that DPR = 1)

I'll remove the `image-rendering` since we have a separate test for that.

PR #2119, opened 2023-01-05
Add a canvas image-rendering wpt ref test

greggman:
Note: the test currently fails in Chrome.

Issue: #1116 and #1115 (not sure what the difference between those are)

<hr>

**Requirements for PR author:**

- [X] All missing test coverage is tracked with "TODO" or `.unimplemented()`.
- [X] New helpers are `/** documented */` and new helper files are found in `helper_index.txt`.
- [ ] Test behaves as expected in a WebGPU implementation. (If not passing, explain above.)

**Requirements for [reviewer sign-off](https://github.com/gpuweb/cts/blob/main/docs/reviews.md):**

- [ ] Tests are properly located in the test tree.
- [ ] [Test descriptions](https://github.com/gpuweb/cts/blob/main/docs/intro/plans.md) allow a reader to "read only the test plans and evaluate coverage completeness", and accurately reflect the test code.
- [ ] Tests provide complete coverage (including validation control cases). **Missing coverage MUST be covered by TODOs.**
- [ ] Helpers and types promote readability and maintainability.

When landing this PR, be sure to make any necessary issue status updates.

PR #2138, opened 2023-01-09
Remove class=reftest-wait from reftests

kainino0x:
As far as I can tell this shouldn't be necessary. Without this, the page snapshot is supposed to occur after the 'load' event fires, which is fine for us (that should be after `<script type=module>`s are loaded).
https://web-platform-tests.org/writing-tests/reftests.html#controlling-when-comparison-occurs

Will verify when it rolls into Chromium (will hold off on landing this until we've successfully rolled the other new reftest, #2119, and have understood why it was failing to roll)

Issue: None

<hr>

### DIFF
--- a/src/webgpu/web_platform/reftests/canvas_clear.https.html
+++ b/src/webgpu/web_platform/reftests/canvas_clear.https.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html class="reftest-wait">
+<html>
   <title>WebGPU canvas_clear</title>
   <meta charset="utf-8" />
   <link rel="help" href="https://gpuweb.github.io/gpuweb/" />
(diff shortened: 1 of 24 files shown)


greggman (approved):
lgtm

One question though, "after the page loads" .... webgpu is async so I'd expect some sychronization needed

kainino0x:
oh yes, true, I forgot about that for some reason. That's probably why we put reftest-wait here. I'll just close this.

kainino0x:
I realized I should look at the test code to verify. The tests end in `takeScreenshotDelayed`:
https://github.com/web-platform-tests/wpt/blob/master/common/reftest-wait.js

Which indeed does the thing specified in the docs, removes the `reftest-wait` class from the root element.

PR #2139, opened 2023-01-10
op: Add infinity bias slope scale params to 'depth_bias' test

Gyuyoung:
This PR adds more parameterizations to 'depth_bias' test for the infinity bias slope scale.

Issue: #2023

<hr>

### DIFF
--- a/src/webgpu/api/operation/rendering/depth_bias.spec.ts
+++ b/src/webgpu/api/operation/rendering/depth_bias.spec.ts
@@ -5,6 +5,7 @@ Tests render results with different depth bias values like 'positive', 'negative
 
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
 import { unreachable } from '../../../../common/util/util.js';
+import { kValue } from '../../../../webgpu/util/constants.js';
 import {
   DepthStencilFormat,
   EncodableTextureFormat,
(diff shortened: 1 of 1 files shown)


src/webgpu/api/operation/rendering/depth_bias.spec.ts
```diff
@@ -298,6 +299,20 @@ g.test('depth_bias')
...
+          quadAngle: QuadAngle.TiltedX,
+          bias: 0,
+          biasSlopeScale: kValue.f32.infinity.positive,
+          biasClamp: 0,
+          _expectedDepth: 1.0,
+        },
+        {
+          quadAngle: QuadAngle.TiltedX,
+          bias: 0,
+          biasSlopeScale: kValue.f32.infinity.negative,
```
kainino0x, line 312: Our WebIDL doesn't allow infinities (or NaNs) here, see definition here:
https://gpuweb.github.io/gpuweb/#depth-stencil-state
is `float`, not `unrestricted float`.
(It may be allowed by dawn-node - it doesn't implement WebIDL perfectly - but Chromium should reject it)

That said, if infinities _are_ meaningful values (I don't really know), we should change the IDL to allow `unrestricted float` here. Maybe you can file a gpuweb issue if so?

Gyuyoung: I referred to the 'DepthBiasTests.PositiveInfinitySlopeBiasOnFloat'[1] and 'DepthBiasTests.NegativeInfinityBiasOnFloat'[2]. So, I thought that infinity values could be allowed. If you can't be sure if the infinity values are allowed. I think it would be not to add this because I can't keep working on this.

[1] https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/src/dawn/tests/end2end/DepthBiasTests.cpp;l=248?q=DepthBiasTests.cpp&ss=chromium

[2] https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/src/dawn/tests/end2end/DepthBiasTests.cpp;l=261?q=DepthBiasTests.cpp&ss=chromium

kainino0x: No worries! They definitely shouldn't be possible in the current JS spec, so if we add them in the future we can write tests then.

PR #2147, opened 2023-01-11
operation,render_pipeline,sample_mask

shrekshao:
Issue: #2146

Tests that the final sample mask is the logical AND of all the relevant masks

Test by drawing a fullscreen quad (for rasterization mask = 0xf) or small sub quads (one color quad) like this on a 1x1 render target texture.
image
(white dot roughly shows the standard sample point pattern)

Look at the color in the render target texture (sampleCount = 1) or resolve target texture (sampleCount = 4) to determine if it matches with the final sample masks output.

Passing on Windows Nvidia and Mac Intel.

<hr>

### DIFF
--- a/src/webgpu/api/operation/render_pipeline/sample_mask.spec.ts
+++ b/src/webgpu/api/operation/render_pipeline/sample_mask.spec.ts
@@ -1,16 +1,443 @@
 export const description = `
-TODO:
-- for sampleCount = { 1, 4 } and various combinations of:
-    - rasterization mask = { 0, 1, 2, 3, 15 }
-    - sample mask = { 0, 1, 2, 3, 15, 30 }
-    - fragment shader output mask (SV_Coverage) = { 0, 1, 2, 3, 15, 30 }
-- test that final sample mask is the logical AND of all the
-  relevant masks -- meaning that the samples not included in the final mask are discarded
-  for all the { color outputs, depth tests, stencil operations } on any attachments.
-- [choosing 30 = 2 + 4 + 8 + 16 because the 5th bit should be ignored]
+Tests that the final sample mask is the logical AND of all the relevant masks.
 `;
 
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { assert, range } from '../../../../common/util/util.js';
 import { GPUTest } from '../../../gpu_test.js';
 
-export const g = makeTestGroup(GPUTest);
+const kColors = [
+  // Red
+  { R: 0xff, G: 0, B: 0, A: 0xff },
+  // Green
+  { R: 0, G: 0xff, B: 0, A: 0xff },
+  // Blue
+  { R: 0, G: 0, B: 0xff, A: 0xff },
+  // Yellow
+  { R: 0xff, G: 0xff, B: 0, A: 0xff },
+];
+
+const kEmptySample = { R: 0, G: 0, B: 0, A: 0 };
+
+// Format of the render target and resolve target
+const format = 'rgba8unorm';
+
+class F extends GPUTest {
+  async GetResolvedTargetTexture(
+    sampleCount: number,
+    rasterizationMask: number,
+    sampleMask: number,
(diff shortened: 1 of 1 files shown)


src/webgpu/api/operation/render_pipeline/sample_mask.spec.ts
```diff
@@ -1,16 +1,443 @@
...
+- Sample 3, Texel 3, top-left: Yellow
+
+The test checks which sample is passed by looking at the final color of the 1x1 resolve target texture, to see if that matches
+what is expected given by the rasterization mask, sample mask, and fragment shader output mask.
+
 - for sampleCount = { 1, 4 } and various combinations of:
-    - rasterization mask = { 0, 1, 2, 3, 15 }
+    - rasterization mask = { 0, 1, 2, 3, 4, 5, 15 }
     - sample mask = { 0, 1, 2, 3, 15, 30 }
     - fragment shader output mask (SV_Coverage) = { 0, 1, 2, 3, 15, 30 }
```
kainino0x, line 409: Can update the language here, this description dates back to when we thought we might do WHLSL.

```suggestion
    - fragment shader output @builtin(sample_mask) = { 0, 1, 2, 3, 15, 30 }
```

src/webgpu/api/operation/render_pipeline/sample_mask.spec.ts
```diff
@@ -1,16 +1,443 @@
...
+    sampleMask: number,
+    fragmentShaderOutputMask: number
+  ) {
+    const expected = {
+      R: 0,
+      G: 0,
+      B: 0,
+      A: 0,
+    };
+    range(4, (id: number) => {
```
kainino0x, line 351: nit: I'd write this as a (C-style) `for` loop, it's a little hard to read that this is a loop.

src/webgpu/api/operation/render_pipeline/sample_mask.spec.ts
```diff
@@ -1,16 +1,443 @@
...
-`;
+`
+  )
+  .params(u =>
+    u
+      .combine('sampleCount', [1, 4] as const)
+      .beginSubcases()
+      .combine('rasterizationMask', [0, 1, 2, 3, 4, 5, 8, 15] as const)
+      .combine('sampleMask', [0, 1, 2, 3, 15, 30] as const)
+      .combine('fragmentShaderOutputMask', [0, 1, 2, 3, 15, 30] as const)
```
kainino0x, line 419: nit: I know the original notes suggested this parameterization but it's probably unnecessarily big. On my machine the tests take about 500ms, which isn't terrible, but probably isn't providing much coverage. Maybe we could just do [0b0111, 0b1011, 0b1101, 0b1110]` for each parameter or something?

kainino0x (approved):
Great test! LGTM with a few comments

BTW it is possible to check the individual samples of a texture by reading them using `texture_multisampled_2d` in wgsl. I don't think we've written a helper yet to "copy" a multisampled texture into a buffer, but it's something we could do for future tests.

src/webgpu/api/operation/render_pipeline/sample_mask.spec.ts
shrekshao: I think mask = 0 is needed
mask = 0b1111 is needed.
fragmentShaderOutputMask, sampleMask = 30 is still needed to test if 5th bit is ignored.
mask = 0b0001 and 0b0010 are needed because they have different meaning for single/multisampled cases.

then we can have [0b0111, 0b1011, 0b1101, 0b1110]
which doesn't seem like we trimmed down many cases.

If the performance is a concern, as I'm still adding depth and stencil tests which will make tests running time bigger, What about putting rasterizationMask before beginSubcases. (On my machine each case now still takes 200~400ms though)

kainino0x: No worries, I think we can keep as is.

PR #2152, opened 2023-01-11
Remove phantom dependency on reftest-wait.js, copy code into CTS

kainino0x:
We were relying on takeScreenshotDelayed to be defined by WPT, but it's not present when running test pages outside WPT, which means there are always errors, which makes it hard to see when we've forgotten the script tag to import it.

The implementation is trivial, just removes a class from `<html>`. So better to just copy it in.

This should fix a failure in `resize_observer.https.html`:
```
CONSOLE ERROR: Uncaught (in promise) ReferenceError: takeScreenshotDelayed is not defined
```

I didn't test this inside WPT, but locally fixes the errors when opening the page standalone, e.g.
`http://localhost:8080/out/webgpu/web_platform/reftests/resize_observer.https.html`
And the removal of the class is working as expected.

Issue: None

<hr>


greggman (approved):
lgtm

One bug, you might want to fix here or we can do in another PR. `canvas_image_rendering-ref.html` is has `class="reftest-wait"` but it's not removing the class when ready so my guess is it's timing out to pass.

The fix would be something like

```
  <script>
    (async () => {
      const dataURL = 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAEAAAABACAYAAACqaXHeAAAAAXNSR0IArs4c6QAAAKdJREFUeF7t28kJwDAQA0Cp/6I3hJAqNK/kazDr2atJ7u7S9v3Z+76nnz18m7oBboAYIAZMRv//1fMKcAAHkCAJLuYAXoEv8ZMLcAAHcAAHcAAHjBZEOeBSDuAADuAADuAADtjrCqsIqQh98xAkSIIkSIIkSIIkSIKrYzJ6gyRIgiRIgiRIgiRIgiRoZ2hwYcp8gKqwSVF9AXuD9gbtDdobXGWw7nCbB5+MQQlHipKKAAAAAElFTkSuQmCC';
      await Promise.all([...document.querySelectorAll('img')].map(img => {
        img.src = dataURL;
        return img.decode();
      }));
      document.documentElement.classList.remove("reftest-wait");
    })();
  </script>
```

If you don't want to do it here I'll add a PR

PR #2154, opened 2023-01-12
Fix local ref-test runner

greggman:
Turns out playwright (or maybe Chrome) re-renders the page when it takes a screenshot. This meant several bugs were not being repoed.

Maybe we can patch playwright but a quick-ish hacky solution was to use an OS level window screenshot. ATM MacOS only though I think Windows would be pretty easy

Also emit an HTML file with the screenshots. Surprised about half are failing but they do appear to be different when I examined them up close (I used Krita). You can set a layer to XOR so drag in actual then drag ref on top and set to XOR.

Hope this is not turing into a rabbit hole of work :P 

Issue: None

<hr>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,6 +24,7 @@
         "@types/serve-index": "^1.9.1",
         "@typescript-eslint/parser": "^4.33.0",
         "@webgpu/types": "0.1.22",
+        "active-win": "^8.0.3",
         "ansi-colors": "4.1.1",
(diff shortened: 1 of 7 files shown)


kainino0x (approved):
LGTM. Too bad we have to do this, if it does become a rabbit hole it would probably be best to pick another approach. WPT does have its own multi-browser infra for running tests, I think including reftests, so maybe we could write a script to copy tests into a WPT checkout and run using that?
https://web-platform-tests.org/running-tests/from-local-system.html

PR #2160, opened 2023-01-12
op,render_pipeline,sample_mask: depth stencil attachment tests

shrekshao:
Issue: #2146

Add depth and stencil tests to op,render_pipeline,sample_mask

I'm using the old way to check depth and stencil attachment content: make a new render pass using the depthStencilTexture from the previous multisampled render pass, loadOp = 'load' to preserve the value, rely on depth test/stencil test and look at final color texture content to see if the new render pass draw anything to infer what depth/stencil value exists in the depthStencilTexture.

But I just found https://gpuweb.github.io/gpuweb/#copying-depth-stencil and the `* MAINTENANCE_TODO: Add check for values of depth/stencil, probably through sampling of shader` at https://github.com/gpuweb/cts/blob/main/src/webgpu/gpu_test.ts#L589

Turns out we could read from depth and stencil texture and write to storage buffer, and use our util functions to compare buffer value. That seems to be a more general way to check depth/stencil texture content.

So send this out to have discuss whether we'd better to add and use such a helper function to more directly test depth/stencil texture content.

<hr>

### DIFF
--- a/src/webgpu/api/operation/render_pipeline/sample_mask.spec.ts
+++ b/src/webgpu/api/operation/render_pipeline/sample_mask.spec.ts
@@ -21,8 +21,18 @@ const kColors = [
 
 const kEmptySample = { R: 0, G: 0, B: 0, A: 0 };
 
+// (0.2, 0.4, 0.6, 0.8)
+const kDepthStencilCheckColor = { R: 0x33, G: 0x66, B: 0x99, A: 0xcc };
+
+const kStencilReferenceValue = 255;
+
 // Format of the render target and resolve target
 const format = 'rgba8unorm';
+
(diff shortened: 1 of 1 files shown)


kainino0x:
I recently checked and textureContentIsOKByT2B does work on depth32float. It probably works on stencil as well.
https://github.com/gpuweb/cts/pull/2075#discussion_r1049267982
However this will only work for single-sample textures. There is also a `texture_depth_multisampled_2d` so it's also possible to read the data directly out of a multisampled depth texture.

shrekshao:
Agree with having straight-forward readback helper for depth and stencil. Let me close this one. I will fix with using the more general helper

Issue #918, opened 2022-01-13
reftests: test canvas colorSpace


kainino0x:
Will be blocked until additional colorspaces are actually allowed by the spec.

kainino0x:
Unblocked.

PR #2161, opened 2023-01-13
Test GPUCanvasContext.configure colorSpace

greggman:
Note: Unless I'm not understanding, everything except srgb->srgb is broken, even display-p3->display-p3

Here's something that shows results doing the same with canvas 2d
https://jsgist.org/?src=5b8a3a2e438b1850c02c77f19b25ffb1

Issue: #918

<hr>

### DIFF
--- a/src/webgpu/web_platform/reftests/canvas_colorspace.html.ts
+++ b/src/webgpu/web_platform/reftests/canvas_colorspace.html.ts
@@ -1,0 +1,82 @@
+import { kUnitCaseParamsBuilder } from '../../../common/framework/params_builder.js';
+import { Float16Array } from '../../../external/petamoriken/float16/float16.js';
+import { kCanvasAlphaModes, kCanvasColorSpaces } from '../../capability_info.js';
+
+import { runRefTest } from './gpu_ref_test.js';
+
+function bgra8UnormFromRgba8Unorm(rgba8Unorm: Uint8Array) {
+  const bgra8Unorm = rgba8Unorm.slice();
+  for (let i = 0; i < bgra8Unorm.length; i += 4) {
+    [bgra8Unorm[i], bgra8Unorm[i + 2]] = [bgra8Unorm[i + 2], bgra8Unorm[i]];
+  }
+  return bgra8Unorm;
+}
+
+function rgba16floatFromRgba8unorm(rgba8Unorm: Uint8Array) {
+  const rgba16Float = new Float16Array(rgba8Unorm.length);
+  for (let i = 0; i < rgba8Unorm.length; ++i) {
+    rgba16Float[i] = rgba8Unorm[i] / 255;
+  }
+  return rgba16Float;
+}
+
+export function runColorSpaceTest(format: GPUTextureFormat) {
+  runRefTest(async t => {
+    const device = t.device;
+
+    // prettier-ignore
+    const kRGBA8UnormData = new Uint8Array([
+      0, 255, 0, 255,
+      117, 251, 7, 255,
+      170, 35, 209, 255,
+      80, 150, 200, 255,
+    ]);
+    const kBGRA8UnormData = bgra8UnormFromRgba8Unorm(kRGBA8UnormData);
+    const kRGBA16FloatData = rgba16floatFromRgba8unorm(kRGBA8UnormData);
+    const width = kRGBA8UnormData.length / 4;
+
+    const testData: { [id: string]: Uint8Array | Float16Array } = {
+      rgba8unorm: kRGBA8UnormData,
+      bgra8unorm: kBGRA8UnormData,
+      rgba16float: kRGBA16FloatData,
+    };
+
+    function createCanvas(
+      alphaMode: GPUCanvasAlphaMode,
+      format: GPUTextureFormat,
+      colorSpace: PredefinedColorSpace
+    ) {
+      const canvas = document.createElement('canvas');
+      canvas.width = width;
+      canvas.height = 1;
+      const context = canvas.getContext('webgpu') as GPUCanvasContext;
+      context.configure({
+        device,
+        format,
+        usage: GPUTextureUsage.COPY_DST | GPUTextureUsage.RENDER_ATTACHMENT,
+        alphaMode,
+        colorSpace,
+      });
+
+      const textureData = testData[format];
+      const texture = context.getCurrentTexture();
+      device.queue.writeTexture(
+        { texture },
+        textureData.buffer,
+        {
+          bytesPerRow: textureData.byteLength,
+        },
+        { width: 4, height: 1 }
+      );
+      document.body.appendChild(canvas);
+    }
+
+    const u = kUnitCaseParamsBuilder
+      .combine('alphaMode', kCanvasAlphaModes)
+      .combine('colorSpace', kCanvasColorSpaces);
+
+    for (const { alphaMode, colorSpace } of u) {
+      createCanvas(alphaMode, format, colorSpace);
+    }
+  });
+}
--- a/src/webgpu/web_platform/reftests/ref/canvas_colorspace-ref.html
+++ b/src/webgpu/web_platform/reftests/ref/canvas_colorspace-ref.html
@@ -1,0 +1,17 @@
+<!DOCTYPE html>
+<html>
+  <title>WebGPU canvas_colorspace (ref)</title>
+  <meta charset="utf-8" />
+  <link rel="help" href="https://gpuweb.github.io/gpuweb/" />
+  <style>
+    canvas {
+      width: 128px;
+      height: 128px;
+      margin-right: 5px;
+      image-rendering: pixelated;
+      image-rendering: crisp-edges;
+    }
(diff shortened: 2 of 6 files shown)


src/webgpu/web_platform/canvas/configure.spec.ts
```diff
@@ -422,3 +446,160 @@ g.test('viewFormats')
...
     }, !compatible);
   });
+
+g.test('colorSpace')
+  .desc(
+    `
+    Test the colorSpace has an affect when using WebGPU
+    
+    Note: Effectively this is testing is that the colorSpace makes it through
+    to the canvas so when applied to another canvas the correct conversion happens.
```
kainino0x, line 456:
```suggestion
    Test copying from opaque, preferred-format WebGPU canvases of various color spaces to 2d
    canvases of various color spaces. The copy is expected to have the same result as a 2d-to-2d
    copy with the same source/destination color space.

    If the color spaces match, the data should be preserved exactly. If not, conversion and rounding
    is expected, and may be slightly different between WebGPU-to-2d and 2d-to-2d, so allow for minor
    differences.
```

kainino0x (commented):
Overall LGTM but a few suggestions

> Issue: https://github.com/gpuweb/cts/issues/918

This issue is for a reftest. We need both, because readback and presentation paths differ a lot.

greggman:
Ok, I changed these to ref tests

src/webgpu/web_platform/reftests/canvas_colorspace.html.ts
```diff
@@ -0,0 +1,68 @@
+import { runRefTest } from './gpu_ref_test.js';
+
+function bgraFromRgba(rgba: Uint8Array) {
+  const bgra = rgba.slice();
+  for (let i = 0; i < bgra.length; i += 4) {
```
kainino0x, line 5: `i` is not used. Fixing this makes the results look correct on my system (at least, the ones that work)

optional: Could be written `[bgra[i], bgra[i + 2]] = [bgra[i + 2], bgra[i]];` if you like that style

src/webgpu/web_platform/reftests/canvas_colorspace.html.ts
```diff
@@ -0,0 +1,68 @@
...
+  const width = kRGBAData.length / 4;
+
+  const testData: { [id: string]: Uint8Array } = {
+    rgba8unorm: kRGBAData,
+    bgra8unorm: kBGRAData,
+  };
+
+  function createCanvas(
+    alphaMode: GPUCanvasAlphaMode,
+    format: GPUTextureFormat,
```
kainino0x, line 33: optional suggestion
```suggestion
  const testData = {
    rgba8unorm: kRGBAData,
    bgra8unorm: kBGRAData,
  };

  function createCanvas(
    alphaMode: GPUCanvasAlphaMode,
    format: keyof typeof testData,
```

src/webgpu/web_platform/reftests/canvas_colorspace.html.ts
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  createCanvas('opaque', 'rgba8unorm', 'srgb');
+  createCanvas('opaque', 'rgba8unorm', 'display-p3');
+  createCanvas('opaque', 'bgra8unorm', 'srgb');
+  createCanvas('opaque', 'bgra8unorm', 'display-p3');
+  createCanvas('premultiplied', 'rgba8unorm', 'srgb');
+  createCanvas('premultiplied', 'rgba8unorm', 'display-p3');
+  createCanvas('premultiplied', 'bgra8unorm', 'srgb');
+  createCanvas('premultiplied', 'bgra8unorm', 'display-p3');
```
kainino0x, line 67: An additional thought when looking at #2176: probably good to do loops over that PR's `kCanvasColorSpaces` and a new `kCanvasAlphaModes`, to help keep the test up to date with future changes

src/webgpu/web_platform/reftests/canvas_colorspace.html.ts
greggman: doh! thanks for catching that!

src/webgpu/web_platform/reftests/ref/canvas_colorspace-ref.html
```diff
@@ -0,0 +1,49 @@
...
+  // prettier-ignore
+  const kRGBAData = new Uint8Array([
+    0, 255, 0, 255,
+    117, 251, 7, 255,
+    170, 35, 209, 255,
+    80, 150, 200, 255,
+  ]);
+  const width = kRGBAData.length / 4;
+
+  function createCanvas(alphaMode, format, colorSpace) {
```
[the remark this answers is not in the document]
greggman: done

src/webgpu/web_platform/reftests/canvas_colorspace.html.ts
greggman: done

src/webgpu/web_platform/reftests/canvas_colorspace.html.ts
```diff
@@ -0,0 +1,68 @@
...
+      },
+      { width: 4, height: 1 }
+    );
+    document.body.appendChild(canvas);
+  }
+
+  createCanvas('opaque', 'rgba8unorm', 'srgb');
+  createCanvas('opaque', 'rgba8unorm', 'display-p3');
+  createCanvas('opaque', 'bgra8unorm', 'srgb');
+  createCanvas('opaque', 'bgra8unorm', 'display-p3');
```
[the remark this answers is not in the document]
greggman: I don't mind splitting it but this seems at odds with the next comment. It feels like the only reason this is being split is because Chrome can't pass the `rgba8unorm` test right now. But, the same point that the test should use `kCanvasColorSpaces` and `kCanvaAlphaModes`, it should also use `kCanvasTextureFormats` so when a new format is added a bunch of tests don't have to be added, the existing tests just pick up the new format.

src/webgpu/web_platform/reftests/canvas_colorspace.html.ts
greggman: done

src/webgpu/web_platform/reftests/ref/canvas_colorspace-ref.html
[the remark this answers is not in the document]
greggman: done

src/webgpu/web_platform/reftests/canvas_colorspace.html.ts
```diff
@@ -0,0 +1,71 @@
...
+          bytesPerRow: textureData.byteLength,
+        },
+        { width: 4, height: 1 }
+      );
+      document.body.appendChild(canvas);
+    }
+
+    const u = kUnitCaseParamsBuilder
+      .combine('alphaMode', kCanvasAlphaModes)
+      .combine('colorSpace', kCanvasColorSpaces);
```
kainino0x, line 65: nit: how about moving this out of the function and exporting it? Then it can be imported from the ref and reused directly so it's clear they need to be the same.

also nit: by convention `u` usually refers to the `unit` builder, which this is no longer.

src/webgpu/web_platform/reftests/canvas_colorspace.html.ts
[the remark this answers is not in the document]
kainino0x: That's fair, and that is pretty much the only reason. It's unfortunate that we won't get the benefit of picking up the new format but for now this is more practical.

As an alternative we could have one file for bgra8unorm and one file for [everything except bgra8unorm]? It's a bit contrived but would solve both problems for now.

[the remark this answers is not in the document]
greggman: split

greggman:
I tried adding a rgba16float one. It's possible I need to pick better colors but basically I think we can't easily test rgba16float until canvas2d gets a float format? At least looking at the ref differences the pixels are being dithered.

test:

Screenshot 2023-01-18 at 6 50 08 PM


ref:

Screenshot 2023-01-18 at 6 50 15 PM


diff:

Screenshot 2023-01-18 at 6 48 58 PM

kainino0x (approved):
nit: avoid rebases after a PR has had an initial review, github can't do incremental diffs past rebases

kainino0x:
> I tried adding a rgba16float one. It's possible I need to pick better colors but basically I think we can't easily test rgba16float until canvas2d gets a float format? At least looking at the ref differences the pixels are being dithered.

Hmm, that result is unfortunate. I forgot about rgba16float for some reason even though I thought about it briefly.
If you're picking colors that are `(integer)/255` then I don't know why there would be dithering. The dithering should be in a very small range of values though, so can fuzzy matching take care of it for us?

greggman:
yea, I passing in (int)/255 via `Float16Array`

I forgot we could set fuzzy. That fixes it.

kainino0x (approved):
great!

Issue #1626, opened 2022-07-05
wgsl: Arithmetic Expression tests


dj2:
# Negation
* [ ] Negate `0.0` to get `-0.0` _or_ `0.0`
* [ ] Negate `-0.0` to get `0.0` _or_ `-0.0`
* [ ] Do the above for {AbstractFloat, f32 and f16}
* [ ] Negate `0` to get `0`
* [ ] Negate `-0` to get `0` **amaiorano: I don't think this is valid WGSL anymore**
* [x] Do the above for i32
* [ ] Do the above for AbstractInt
* [ ] Common case negation for {AbstractFloat, AbstractInt, i32, f32, f16}
* [ ] Negation of a u32 is a compile error
* [ ] Negation of `-9_223_372_036_854_775_808` is a shader creation error (max AbstractInt overflows)
* [ ] Negation of `-9_223_372_036_854_775_807` is `9_223_372_036_854_775_807`
* [ ] Negation of `-2_147_483_648i` gives `-2_147_483_648i` (Int min stays int min)
* [ ] Negation of `-2_147_483_647i` gives `2_147_483_647i`
* [x] Negation of float min should work (I think)?
* [x] Negation of float max returns ... something ... but isn't an error (I think)?
* [ ] Negation of a `vec<{AbstractFloat, AbstractFloat, i32, f32, f16}>`

# Addition/ Subtraction / Multiplication
* [ ] Overflow AbstractInt or AbstractFloat is compile error
* [x] Overflow i32 or u32 is modulo 2^32
* [ ] Overflow f32 or f16 isn't error, but undefined value
* [x] Typical case for each
* [x] Vector case for each

# Division
* [x] {i32, u32} division by 0 returns the numerator
* [ ] AbstractInt divided by 0 is a compile error
* [ ] AbstractInt most negative number divided by -1 is compile error
* [x] i32 most negative divided by -1 is most negative number
* [ ] AbstractFloat divided by 0 ~~returns a value~~ is a compile error
* [ ] {f32, f16} ... https://github.com/gpuweb/gpuweb/issues/2798
* [ ] `-2i/-3i` == `0`
* [ ] `-4i/3i` == `-1`
* [ ] `2i/3i`== `0`
* [ ] `-2/-3` == `0`   (AbstractInt)
* [ ] `2/3`== `0`  (AbstractInt)
* [ ] `2u/3u` == `0`
[...]

PR #2162, opened 2023-01-13
Add unary i32 execution tests

amaiorano:
Issue: #1626

### DIFF
--- a/src/webgpu/shader/execution/expression/unary/i32_arithmetic.spec.ts
+++ b/src/webgpu/shader/execution/expression/unary/i32_arithmetic.spec.ts
@@ -1,0 +1,37 @@
+export const description = `
+Execution Tests for the i32 arithmetic unary expression operations
+`;
+
+import { makeTestGroup } from '../../../../../common/framework/test_group.js';
+import { GPUTest } from '../../../../gpu_test.js';
+import { i32, TypeI32 } from '../../../../util/conversion.js';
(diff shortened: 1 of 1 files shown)


src/webgpu/shader/execution/expression/unary/arithmetic.spec.ts
```diff
@@ -36,6 +41,21 @@ Accuracy: Correctly rounded
     u.combine('inputSource', allInputSources).combine('vectorize', [undefined, 2, 3, 4] as const)
   )
   .fn(async t => {
-    const cases = await d.get('negation');
+    const cases = await d.get('negation_f32');
     await run(t, unary('-'), [TypeF32], TypeF32, t.params, cases);
   });
+
+g.test('negation_i32')
```
amaiorano, line 48: When I made this change, I decided to put the i32 and f32 tests in the same file, thinking it would be easier to add the other types as well (f16, u32, AFloat, AInt). However, after looking at other tests, I'm not sure this is a good idea. So if you want, I'll update this CL by splitting out the i32 tests into its own file.

zoddicus: I would prefer to split this up

amaiorano: Done. PTAL.

zoddicus (approved):

PR #2167, opened 2023-01-16
runtime: Add server '/load' RPC

ben-clayton:
Instead of loading *all* the test cases, on first case, make the loading of the test cases explicit.

This can dramatically reduce the latency of small test runs.

Note: this currently only affects the `dawn/node` test runner, which requires the corresponding change: https://dawn-review.googlesource.com/c/dawn/+/117200

### DIFF
--- a/src/common/runtime/server.ts
+++ b/src/common/runtime/server.ts
@@ -144,18 +144,7 @@ if (verbose) {
 (async () => {
   Logger.globalDebugMode = verbose;
   const log = new Logger();
-  const testcases = allWebGPUTestcases();
-
-  async function allWebGPUTestcases() {
-    const webgpuQuery = parseQuery('webgpu:*');
-    const loader = new DefaultTestFileLoader();
-    const map = new Map<string, TestTreeLeaf>();
-    for (const testcase of await loader.loadCases(webgpuQuery)) {
-      const name = testcase.query.toString();
-      map.set(name, testcase);
-    }
-    return map;
-  }
+  const testcases = new Map<string, TestTreeLeaf>();
 
   async function runTestcase(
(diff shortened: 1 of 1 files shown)


src/common/runtime/server.ts
```diff
@@ -174,13 +163,28 @@ if (verbose) {
...
+          response.end();
+        } catch (err) {
+          response.statusCode = 500;
+          response.end(`load failed with error: ${err}\n${(err as Error).stack}`);
+        }
+      } else if (request.url.startsWith(runPrefix)) {
         const name = request.url.substr(runPrefix.length);
         try {
-          const testcase = (await testcases).get(name);
+          const testcase = testcases.get(name);
```
dj2, line 187: Where does this get populated? Or can you only call `/run` after explicitly calling `/load`?

ben-clayton: > Or can you only call `/run` after explicitly calling `/load`?

Correct.

dj2: Can we maybe add an error handler for that case? We check if the test case is found, but maybe also check if the `testcases` is empty and say forgot to call `load`, or something?

ben-clayton: I could, but I'm not sure it's worth it. This isn't a human API, and I'm not attempting to build a stable interface here. This is essentially an extension of the `dawn/node` runner, which just happens to live in the CTS code-base. I'd rather not go through the effort of pretending this is a stable API, when I fully intend to make breaking changes.

dj2: Fair.

dj2 (approved):

PR #2171, opened 2023-01-17
Add additional unit testing for builtins that depend on `dot`

### LINKED ISSUE #2134: Differing ordering of operations for builtins, i.e. `dot`, need to be handled
See https://bugs.chromium.org/p/tint/issues/detail?id=1790 for real world hardware/implementation that is exposing this.

`dot(x, y)`'s accuracy is an inherited accuracy from `x[i] * y[i]`, which for a 3D vector which can be unfolded as:
`x[0] * y[0] + x[1] * y[1] + x[2] * y[2]`
= `a + b + c`, where `a = x[0] * y[0], b = x[1] * y[1], c = x[2] * y[2]`
which is what the CTS currently does

This become problematic for very large/small values, where the for example a & b are a very large positives, and c is a very large negative.
Because if a + b is calculated first it might go out of bounds, but if b + c is calculated first they might cancel out, so depending on the order of calculation the specific case may have wildly different acceptance intervals.

The solution to this is so calculate all valid computation orderings, and span the resultant intervals.
This is likely occurring other places than just `dot`

zoddicus:
All of the existing implementations appear to not require functional changes, so this just mostly adds some
additional testing for if `dot` goes out of bounds.

I have added an early exit to `refract` for when `dot` goes OOB.

Fixes #2134

<hr>


src/webgpu/util/f32_interval.ts
```diff
@@ -1806,7 +1806,7 @@ export function refractInterval(i: number[], s: number[], r: number): F32Vector
   const one_minus_dot_squared = subtractionInterval(1, dot_squared);
   const k = subtractionInterval(1.0, multiplicationInterval(r_squared, one_minus_dot_squared));
 
-  if (k.containsZeroOrSubnormals()) {
+  if (!k.isFinite() || k.containsZeroOrSubnormals()) {
```
ben-clayton, line 1809: This doesn't appear to be mentioned in the CL description.

PR #2179, opened 2023-01-18
op,render_pipeline,sample_mask depth stencil multisampled tests

shrekshao:
Issue: https://github.com/gpuweb/cts/issues/2146

Add depth and stencil tests to op,render_pipeline,sample_mask

Check is performed by using a compute shader to `textureLoad` each sample from the depth/stencil texture and write to a buffer and then compare the result.

I think we can integrate the functionality of `checking multisampled color/depth/stencil` into the texture_ok helper in a later pull request to:
- have a first review to align
- later dealing with different formats (sample type 'f32' ULP, 'u32'; aspect: depth only, stencil only;)
- avoid conflicting too much with current helper function refactoring

<hr>

### DIFF
--- a/src/webgpu/api/operation/render_pipeline/sample_mask.spec.ts
+++ b/src/webgpu/api/operation/render_pipeline/sample_mask.spec.ts
@@ -21,16 +21,26 @@ const kColors = [
 
 const kEmptySample = { R: 0, G: 0, B: 0, A: 0 };
 
+const kDepthClearValue = 1.0;
+const kDepthWriteValue = 0.0;
+const kStencilClearValue = 0;
+const kStencilReferenceValue = 0xff;
+
 // Format of the render target and resolve target
 const format = 'rgba8unorm';
 
+// Format of depth stencil attachment
+const depthStencilFormat = 'depth24plus-stencil8';
+
+const kRenderTargetSize = 1;
+
 class F extends GPUTest {
-  async GetResolvedTargetTexture(
+  async GetTargetTexture(
     sampleCount: number,
     rasterizationMask: number,
     sampleMask: number,
     fragmentShaderOutputMask: number
-  ): Promise<GPUTexture> {
+  ): Promise<{ color: GPUTexture; depthStencil: GPUTexture }> {
     // Create a 2x2 color texture to sample from
     // texel 0 - Red
     // texel 1 - Green
@@ -168,7 +178,7 @@ class F extends GPUTest {
               );
           
             var output : VertexOutput;
-            output.Position = vec4<f32>(pos[VertexIndex], 0.0, 1.0);
+            output.Position = vec4<f32>(pos[VertexIndex], ${kDepthWriteValue}, 1.0);
             output.fragUV = uv[VertexIndex];
             return output;
           }`,
@@ -201,6 +211,20 @@ class F extends GPUTest {
         mask: sampleMask,
         alphaToCoverageEnabled: false,
       },
+      depthStencil: {
+        format: depthStencilFormat,
+        depthWriteEnabled: true,
+        depthCompare: 'always',
+
+        stencilFront: {
+          compare: 'always',
+          passOp: 'replace',
+        },
+        stencilBack: {
+          compare: 'always',
+          passOp: 'replace',
+        },
+      },
     });
 
     const uniformBindGroup = this.device.createBindGroup({
@@ -223,7 +247,6 @@ class F extends GPUTest {
       ],
     });
 
-    const kRenderTargetSize = 1;
     const renderTargetTexture = this.device.createTexture({
       format,
       size: {
@@ -251,6 +274,16 @@ class F extends GPUTest {
             usage: GPUTextureUsage.COPY_SRC | GPUTextureUsage.RENDER_ATTACHMENT,
           });
 
+    const depthStencilTexture = this.device.createTexture({
+      size: {
+        width: kRenderTargetSize,
+        height: kRenderTargetSize,
+      },
+      format: depthStencilFormat,
+      sampleCount,
+      usage: GPUTextureUsage.RENDER_ATTACHMENT | GPUTextureUsage.TEXTURE_BINDING,
+    });
+
     const renderPassDescriptor: GPURenderPassDescriptor = {
       colorAttachments: [
         {
@@ -262,11 +295,21 @@ class F extends GPUTest {
           storeOp: 'store',
         },
       ],
+      depthStencilAttachment: {
+        view: depthStencilTexture.createView(),
+        depthClearValue: kDepthClearValue,
+        depthLoadOp: 'clear',
+        depthStoreOp: 'store',
+        stencilClearValue: kStencilClearValue,
+        stencilLoadOp: 'clear',
+        stencilStoreOp: 'store',
+      },
     };
     const commandEncoder = this.device.createCommandEncoder();
     const passEncoder = commandEncoder.beginRenderPass(renderPassDescriptor);
     passEncoder.setPipeline(pipeline);
     passEncoder.setBindGroup(0, uniformBindGroup);
+    passEncoder.setStencilReference(kStencilReferenceValue);
 
     if (rasterizationMask === 15) {
       // draw full screen quad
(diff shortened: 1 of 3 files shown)


src/webgpu/api/operation/render_pipeline/sample_mask.spec.ts
```diff
@@ -339,39 +356,168 @@ class F extends GPUTest {
       B: 0,
       A: 0,
     };
+    if (sampleCount === 1) {
+      if (
+        // rasterization needs to cover the center of the pixel
+        (rasterizationMask & 0b1111) >= 0b1111 &&
```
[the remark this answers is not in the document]
shrekshao: Agree with statement. Now logic of sampleCount === 1 and sampleCount > 1 should have merged mostly. The only branching now is when sampleCount === 1, the center is sampling from Texel[3] yellow, instead of Texel[0] red (sampleIndex === 0)

src/webgpu/util/conversion.ts
```diff
@@ -576,6 +580,10 @@ export class VectorType {
   public toString(): string {
     return `vec${this.width}<${this.elementType}>`;
   }
+
+  public getSize(): number {
+    return this.elementType.size * this.width;
+  }
```
kainino0x, line 586: nit: I think if you make this `public get size(): number` then it will be compatible with the `readonly size: number` defined in ScalarType and you can use `.size` instead of `.getSize()`

src/webgpu/gpu_test.ts
```diff
@@ -684,6 +685,110 @@ export class GPUTest extends Fixture<GPUTestSubcaseBatchState> {
...
+      `;
+    const computePipeline = this.device.createComputePipeline({
+      layout: 'auto',
+      compute: {
+        module: this.device.createShaderModule({
+          code:
+            sampleCount === 1
+              ? `
+            struct Buffer {
+              data: array<${type.toString()}>,
```
kainino0x, line 726: nit: I think .toString() is implicit and you can leave it out

src/webgpu/gpu_test.ts
```diff
@@ -684,6 +685,110 @@ export class GPUTest extends Fixture<GPUTestSubcaseBatchState> {
     );
   }
 
+  /**
+   * Emulate a texture to buffer copy by using a compute shader
+   * to load texture value of a single pixel and write to a storage buffer.
+   * For sample count == 1, the buffer contains only one value of the sample.
+   * For sample count > 1, the buffer contains (N = 2 ** sampleCount) values sorted
+   * in the order of their sample index [0, 2 ** sampleCount - 1]
```
kainino0x, line 693: isn't it just `sampleCount` values not `2 ** sampleCount` values?

src/webgpu/gpu_test.ts
```diff
@@ -684,6 +685,110 @@ export class GPUTest extends Fixture<GPUTestSubcaseBatchState> {
...
+   * This can be useful when the texture to buffer copy is not available to the texture format
+   * e.g. (depth24plus)
+   *
+   * MAINTENANCE_TODO: extend to read multiple pixels with given origin and size.
+   *
+   * @returns storage buffer containing the copied value from the texture.
+   */
+  copySinglePixelTextureToBufferUsingComputePass(
+    type: ScalarType,
+    numElements: number,
```
kainino0x, line 704: nit: "componentCount"

src/webgpu/gpu_test.ts
```diff
@@ -684,6 +685,110 @@ export class GPUTest extends Fixture<GPUTestSubcaseBatchState> {
...
+  ): GPUBuffer {
+    const textureLoadCode =
+      numElements === 1
+        ? `dst.data[sampleIndex] = textureLoad(src, coord, sampleIndex).x;`
+        : `let o = sampleIndex * ${numElements};
+      let v = textureLoad(src, coord, sampleIndex);
+      dst.data[o] = v.r;
+      dst.data[o + 1] = v.g;
+      dst.data[o + 2] = v.b;
+      dst.data[o + 3] = v.a;
```
kainino0x, line 716: I think this code is assuming 4 components. I think you can write a loop that works regardless of the number of components? (and put inline in the shader below)

```
let o = sampleIndex * ${numElements};
let v = textureLoad(src, coord, sampleIndex);
for (var component = 0; component < ${componentCount}; component = component + 1) {
  dst.data[o + component] = v[component];
}
```

src/webgpu/gpu_test.ts
```diff
@@ -684,6 +685,110 @@ export class GPUTest extends Fixture<GPUTestSubcaseBatchState> {
...
+      `;
+    const computePipeline = this.device.createComputePipeline({
+      layout: 'auto',
+      compute: {
+        module: this.device.createShaderModule({
```
kainino0x, line 721: nit: for readability maybe move this string out into a separate variable

src/webgpu/gpu_test.ts
```diff
@@ -684,6 +685,110 @@ export class GPUTest extends Fixture<GPUTestSubcaseBatchState> {
...
 
+  /**
+   * Emulate a texture to buffer copy by using a compute shader
+   * to load texture value of a single pixel and write to a storage buffer.
+   * For sample count == 1, the buffer contains only one value of the sample.
+   * For sample count > 1, the buffer contains (N = 2 ** sampleCount) values sorted
+   * in the order of their sample index [0, 2 ** sampleCount - 1]
+   *
+   * This can be useful when the texture to buffer copy is not available to the texture format
+   * e.g. (depth24plus)
```
kainino0x, line 696: or the texture is multisampled

src/webgpu/gpu_test.ts
```diff
@@ -684,6 +685,110 @@ export class GPUTest extends Fixture<GPUTestSubcaseBatchState> {
...
+      dst.data[o + 2] = v.b;
+      dst.data[o + 3] = v.a;
+      `;
+    const computePipeline = this.device.createComputePipeline({
+      layout: 'auto',
+      compute: {
+        module: this.device.createShaderModule({
+          code:
+            sampleCount === 1
+              ? `
```
kainino0x, line 724: possible to use one source regardless of sampleCount? it looks like the loop should work fine for the single-sample case, only the type of `src` needs to be different

src/webgpu/api/operation/render_pipeline/sample_mask.spec.ts
```diff
@@ -311,7 +317,7 @@ class F extends GPUTest {
     passEncoder.setBindGroup(0, uniformBindGroup);
     passEncoder.setStencilReference(kStencilReferenceValue);
 
-    if (rasterizationMask === 15) {
+    if (rasterizationMask === 2 ** sampleCount - 1) {
       // draw full screen quad
       passEncoder.draw(6);
     } else {
```
kainino0x, line 323: assert here sampleCount is 4

src/webgpu/api/operation/render_pipeline/sample_mask.spec.ts
```diff
@@ -311,7 +317,7 @@ class F extends GPUTest {
     passEncoder.setBindGroup(0, uniformBindGroup);
     passEncoder.setStencilReference(kStencilReferenceValue);
 
-    if (rasterizationMask === 15) {
+    if (rasterizationMask === 2 ** sampleCount - 1) {
```
kainino0x, line 320: Possible to make this conditional on sampleCount==1 instead, and maybe make the quad smaller so it covers only the center? There's no need to optimize by using a single quad for the all-four-samples case, the code below will cover it fine.

kainino0x (approved):
LGTM, but a few more comments sorry! Pretty minor so should be able to land without another re-review

kainino0x:
BTW it's very nice the helper can so simply abstract over depth/stencil/color. I thought at first there would be issues with the bind group layout, but thanks to a spec change from a while ago it's really simple!

src/webgpu/gpu_test.ts
shrekshao: I think special code for sampleCount === 1 is to use `texture_2d` instead of `texture_multisampled_2d`

PR #2180, opened 2023-01-18
[wgsl] Alignment attribute tests.

### LINKED ISSUE #1432: wgsl: Validate `@align` parsing
# On a struct member
* [x] `@malign(1)` -- misspelling
* [x] `@align()` -- missing value
* [x] `@align 1)`  -- missing (
* [x] `@align(1 ` -- missing )
* [x] `@align(1, 2)` -- multiple values
* [x] `@align(identifier)` -- must be integer literal
* [x] `@align(1.0)` -- must be integer literal
* [x] `@align(-1)` -- non-negative
* [x] `@align(3)` -- must be a multiple of 2
* [x] `@align(1)` -- valid
* [x] `@align(1i)` -- valid
* [x] `@align(1u)` -- invalid
* [x] `@align(1f)` -- invalid
* [x] `@align(1h)` -- invalid
* [x] `@align(1,)` -- valid trailing comma
* [x] `@align(0x10)` -- (validate hex is accepted)
* [x] `@align(2147483648)` -- valid?
* [x] `@align`
* [x] `@ \t align`
* [x] `@/* comment */align`
* [x] `@align(0)` - invalid must be positive

# Non-struct member tests
* [x] For each type, try to align without satisfying the required-alignment for the type
* [ ] Add align to non-struct type, should reject
   * [ ] align on vector
   * [ ] align on matrix
   * [ ] align of struct itself instead of member
   * [ ] align of array
   * [ ] align on const
   * [ ] align on scalar

# Misc
* [ ] Validate `@align` can only be provided once on a given member.

dj2:
This CL adds test for alignment attribute used outside as struct and multiple alignment attributes applied to an item.

Fixes #1432

<hr>

### DIFF
--- a/src/webgpu/shader/validation/parse/align.spec.ts
+++ b/src/webgpu/shader/validation/parse/align.spec.ts
@@ -179,44 +179,72 @@ g.test('align_required_alignment')
     t.expectCompileResult(!fails, code);
   });
 
-g.test('align_non_struct')
-  .desc('Test that alignment is only allowed on struct members')
+g.test('placement')
+  .desc('Tests the locations @align is allowed to appear')
   .params(u =>
-    u.combine('scope', ['private', 'function']).combine('with_align', [true, false]).beginSubcases()
+    u
+      .combine('scope', [
+        'private-var',
+        'storage-var',
+        'struct-member',
+        'fn-decl',
+        'fn-param',
+        'fn-var',
+        'fn-return',
+        'while-stmt',
+        undefined,
+      ] as const)
+      .combine('attribute', [
+        {
+          'private-var': false,
+          'storage-var': false,
+          'struct-member': true,
+          'fn-decl': false,
+          'fn-param': false,
+          'fn-var': false,
+          'fn-return': false,
+          'while-stmt': false,
+        },
+      ])
+      .beginSubcases()
   )
   .fn(t => {
-    let code = `struct A {
-      a: i32,
-    }
+    const scope = t.params.scope;
+
+    const attr = '@align(32)';
+    const code = `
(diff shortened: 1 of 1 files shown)


dj2:
I intially wrote `align_non_struct` with all of the possible types in there. The test got pretty unreadable when starting to deal with things that have to be in `read_write` storage, or can't be in `storage` at all. It seemed a lot simpler just to drop back and check a single case but we'll have to decide if we want to expand that.

zoddicus (commented):
This looks reasonable to me, but I would like @ben-clayton to weigh in on if this needs to be expanded, since he is more familiar with the spec alignment requirements.

src/webgpu/shader/validation/parse/align.spec.ts
```diff
@@ -178,3 +178,62 @@ g.test('align_required_alignment')
 
     t.expectCompileResult(!fails, code);
   });
+
+g.test('align_non_struct')
+  .desc('Test that alignment is only allowed on struct members')
```
ben-clayton, line 183: This doesn't seem very exhaustive in terms of where the attribute can go. 

Would it be better to parameterize a list of shaders, with insertion points for attributes, then have a big table of `attribute + location = valid` that we can check? That way, you can validate nearly all the attributes for all locations.

ben-clayton (approved):
Approving what you have, as a first step. As discussed, I think this would be better to parameterize.

src/webgpu/shader/validation/parse/align.spec.ts
dj2: Done. Made it specific to align at the moment. Want to get a few more in place to see if/how it generalizes.

PR #2181, opened 2023-01-18
Add tests for binary multiplication of scalars with vectors for u32

amaiorano:
Issue: #1626 

<hr>

### DIFF
--- a/src/webgpu/shader/execution/expression/binary/u32_arithmetic.spec.ts
+++ b/src/webgpu/shader/execution/expression/binary/u32_arithmetic.spec.ts
@@ -4,10 +4,16 @@ Execution Tests for the u32 arithmetic binary expression operations
 
 import { makeTestGroup } from '../../../../../common/framework/test_group.js';
 import { GPUTest } from '../../../../gpu_test.js';
-import { TypeU32 } from '../../../../util/conversion.js';
-import { fullU32Range } from '../../../../util/math.js';
+import { TypeU32, TypeVec } from '../../../../util/conversion.js';
+import { fullU32Range, sparseU32Range, vectorU32Range } from '../../../../util/math.js';
 import { makeCaseCache } from '../case_cache.js';
-import { allInputSources, generateBinaryToU32Cases, run } from '../expression.js';
+import {
+  allInputSources,
+  generateBinaryToU32Cases,
+  generateU32VectorBinaryToVectorCases,
+  generateVectorU32BinaryToVectorCases,
+  run,
+} from '../expression.js';
 
 import { binary } from './binary.js';
 
@@ -61,10 +67,40 @@ export const d = makeCaseCache('binary/u32_arithmetic', {
       return x % y;
     });
   },
+  multiplication_scalar_vector2: () => {
+    return generateU32VectorBinaryToVectorCases(sparseU32Range(), vectorU32Range(2), (x, y) => {
+      return Math.imul(x, y);
+    });
+  },
+  multiplication_scalar_vector3: () => {
+    return generateU32VectorBinaryToVectorCases(sparseU32Range(), vectorU32Range(3), (x, y) => {
+      return Math.imul(x, y);
+    });
+  },
+  multiplication_scalar_vector4: () => {
+    return generateU32VectorBinaryToVectorCases(sparseU32Range(), vectorU32Range(4), (x, y) => {
+      return Math.imul(x, y);
+    });
+  },
+  multiplication_vector2_scalar: () => {
+    return generateVectorU32BinaryToVectorCases(vectorU32Range(2), sparseU32Range(), (x, y) => {
+      return Math.imul(x, y);
+    });
+  },
+  multiplication_vector3_scalar: () => {
+    return generateVectorU32BinaryToVectorCases(vectorU32Range(3), sparseU32Range(), (x, y) => {
+      return Math.imul(x, y);
+    });
+  },
+  multiplication_vector4_scalar: () => {
+    return generateVectorU32BinaryToVectorCases(vectorU32Range(4), sparseU32Range(), (x, y) => {
+      return Math.imul(x, y);
+    });
+  },
 });
 
 g.test('addition')
-  .specURL('https://www.w3.org/TR/WGSL/#floating-point-evaluation')
+  .specURL('https://www.w3.org/TR/WGSL/#arithmetic-expr')
   .desc(
     `
 Expression: x + y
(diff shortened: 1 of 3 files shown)


amaiorano:
*This is WIP*. I wanted to see what you thought of this way of approaching writing tests for mixed types.

zoddicus (changes requested):
This feels like a lot of complexity for generating the values, that I am unclear on why it is needed.

My understanding is that this code is going to get the cached values for ScalarA * ScalarB = Expected, and turn that into something like ScalarA * [ScalarB, ScalarB, ScalarB] = [Expected, Expected, Expected], (with symmetry for the vec * scalar case).

I guess I see to fail to see the benefit on fetching the multiplication values from the cache, and introducing having to resolve Promises of Promises, instead of the just calling the underlying numeric generators directly and redoing the expected calculations.

I was expecting to see something more like a generate a range of numbers `x`, generate a range of vectors `y`, x.flatMap(x.map(...)) returning something like  `{ input: [u32(x), vec2(y)], expected: vec2(y.map(i=> u32(i * x)) }`

Also as-is I think the proposed implementation is only covering scalar-vector operations where the vector is made up of equal elements, i.e. {1, 1, 1}, so we are not getting coverage for cases where the vectors have heterogeneous values in them, {-10000, 23, 0}

amaiorano:
> This feels like a lot of complexity for generating the values, that I am unclear on why it is needed.
> 
> My understanding is that this code is going to get the cached values for ScalarA * ScalarB = Expected, and turn that into something like ScalarA * [ScalarB, ScalarB, ScalarB] = [Expected, Expected, Expected], (with symmetry for the vec * scalar case).

Yes, this is what it's doing. 

> I guess I see to fail to see the benefit on fetching the multiplication values from the cache, and introducing having to resolve Promises of Promises, instead of the just calling the underlying numeric generators directly and redoing the expected calculations.

The one advantage is that these functions can be used for all types, not just u32.

> I was expecting to see something more like a generate a range of numbers `x`, generate a range of vectors `y`, `x.flatMap(x.map(...))` returning something like `{ input: [u32(x), vec2(y)], expected: vec2(y.map(i=> u32(i * x)) }`

Right, so we'd need more "generate" functions for all permutations of types. 

> Also as-is I think the proposed implementation is only covering scalar-vector operations where the vector is made up of equal elements, i.e. {1, 1, 1}, so we are not getting coverage for cases where the vectors have heterogeneous values in them, {-10000, 23, 

Yes. I thought about this as well. As an initial wip change, I kept this simple, but I was thinking of producing vectors of sequential values from the original list.

Having said all that, I actually agree with you.  I'll do as you propose instead.

src/webgpu/shader/execution/expression/expression.ts
```diff
@@ -1059,3 +1059,79 @@ export function generateBinaryToU32Cases(
     return cases;
   }, new Array<Case>());
 }
+
+/**
+ * A function that performs a binary operation on x and y, and returns the expected
+ * result.
+ */
+export interface BinaryOp {
+  (x: number, y: number): number;
```
amaiorano, line 1068: For now, this only returns number. Will be changing this to return number|undefined when implementing division and modulo of scalar and vector.

src/webgpu/shader/execution/expression/binary/u32_arithmetic.spec.ts
```diff
@@ -4,10 +4,16 @@ Execution Tests for the u32 arithmetic binary expression operations
 
 import { makeTestGroup } from '../../../../../common/framework/test_group.js';
 import { GPUTest } from '../../../../gpu_test.js';
-import { TypeU32 } from '../../../../util/conversion.js';
-import { fullU32Range } from '../../../../util/math.js';
+import { TypeU32, TypeVec } from '../../../../util/conversion.js';
+import { fullU32Range, sparseF32Range, vectorU32Range } from '../../../../util/math.js';
```
zoddicus, line 8: I think you want sparseU32Range

src/webgpu/util/math.ts
```diff
@@ -626,6 +627,55 @@ export function fullI32Range(
   ].map(Math.trunc);
 }
 
+/** Short list of u32 values of interest to test against */
+const kInterestingU32Values: number[] = [0, 1, kValue.u32.max / 2, kValue.u32.max];
+
+/** @returns minimal u32 values that cover the entire range of u32 behaviours
+ *
+ * This is used instead of fullF32Range when the number of test cases being
```
zoddicus, line 635: s/fullF32Range/fullU32Range

src/webgpu/util/math.ts
```diff
@@ -626,6 +627,55 @@ export function fullI32Range(
...
+ * generated is a super linear function of the length of f32 values which is
```
zoddicus, line 636: s/f32/u32

src/webgpu/util/math.ts
```diff
@@ -790,6 +840,11 @@ export function quantizeToI32(num: number): number {
   return i32(num).value as number;
 }
 
+/** @returns the closest 32-bit signed integer value to the input */
```
zoddicus, line 843: In src/webgpu/shader/execution/expression/expression.ts there are probably other locations that are doing Math.trunc/Math.round for quantizing to u32 that should be converted to use this

src/webgpu/util/math.ts
```diff
@@ -626,6 +627,55 @@ export function fullI32Range(
...
+    [1, 2, 3, f],
+  ]),
+};
+
+/**
+ * Returns set of vectors, indexed by dimension containing interesting u32
+ * values.
+ *
+ * The tests do not do the simple option for coverage of computing the cartesian
+ * product of all of the interesting float values N times for vecN tests,
```
zoddicus, line 666: s/float/integer

zoddicus (approved):
Overall lgtm, just some cleanups

src/webgpu/shader/execution/expression/binary/u32_arithmetic.spec.ts
zoddicus: The callsites below will need to updated also.

amaiorano: lol, went through all that trouble to add sparseU32Range, then didn't even use it /facepalm

src/webgpu/util/math.ts
amaiorano: I'll follow up with that.

PR #2184, opened 2023-01-20
Texture utility refactoring.

lokokung:
Changes include:
- Implementation of new TextureTestMixin for texture related utilities
- Test modernizations to use the new mixin and helpers
- Added integer truncation for reifyOrigin/Extent helpers since those should always be integers
- Removed unnecessary float16 code paths in tests
- Some minor error message updates
- Added sparse pixel comparison in texture_ok.ts
- Fixed `expectSingleColor` helper for 2d array and 3d textures

A follow up refactoring will be done to finish off removing/modernizing `readSinglePixelFrom2DTexture`, `expectSinglePixelBetweenTwoValuesIn2DTexture`, and `expectSingleColor`

Now that many of the single pixel tests have been unified to do 1 texture copy instead of a bunch per pixel texture copies, I saw up to 10-15% speedups in tests that were testing many pixels!

FYI for reviewers: I tried to split the logical changes a bit in the separate commits, so reviewing them independently might be a bit easier w.r.t the diffs. I also ran each test that I modified locally against Chrome Canary on Mac to make sure that we shouldn't be seeing any regression.

<hr>


lokokung:
FYI I will fix the conflicts after the first round of reviews just to minimize moving parts.

shrekshao (commented):
LGTM with the fixture/gpu_test/texture part. Kai would have more ideas for sure.

The conflict for `sample_mask` is trivial.
The conflict for `gpu_test` is also trivial. In my eariler change I basically add a `copySinglePixelTextureToBufferUsingComputePass`, for depth/stencil texture and multisampled texture, which later we might want to integrate into `expectTexelViewComparisonIsOkInTexture`

https://github.com/gpuweb/cts/pull/2179/files#diff-57b03e13fafce250e4f531dd2d1ade5f24e8f328f1ac0491502ae4f54d314867

src/webgpu/util/texture/texture_ok.ts
```diff
@@ -22,6 +22,13 @@ export type TexelCompareOptions = {
   maxDiffULPsForFloatFormat?: number;
 };
 
+export type PixelExpectation = PerTexelComponent<number> | Uint8Array;
+
+export type PerPixelComparison<E extends PixelExpectation> = {
+  location: GPUOrigin3D;
```
austinEng, line 28: not sure, but maybe `coord` instead of `location` ? seem to use "coord" in other parts of this PR

lokokung: Done

src/webgpu/gpu_test.ts
```diff
@@ -977,3 +991,193 @@ export class GPUTest extends Fixture<GPUTestSubcaseBatchState> {
...
+      src: GPUImageCopyTexture,
+      {
+        exp,
+        comparisonOptions = {
+          maxIntDiff: 0,
+          maxDiffULPsForNormFormat: 1,
+          maxDiffULPsForFloatFormat: 1,
+        },
+      }: {
+        exp: PerPixelComparison<E>[];
```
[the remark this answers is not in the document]
lokokung: Done

src/webgpu/util/texture.ts
```diff
@@ -59,3 +60,79 @@ export function makeTextureWithContents(
...
+    commandEncoder.copyBufferToTexture(
+      { buffer: stagingBuffer, bytesPerRow },
+      { texture, mipLevel },
+      [mipWidth, mipHeight, mipDepthOrArray]
+    );
+  }
+  device.queue.submit([commandEncoder.finish()]);
+
+  // Cleanup the staging buffers.
+  stagingBuffers.map(value => value.destroy());
```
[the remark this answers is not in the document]
lokokung: Done

src/webgpu/gpu_test.ts
```diff
@@ -977,3 +991,193 @@ export class GPUTest extends Fixture<GPUTestSubcaseBatchState> {
...
+        // Build a sparse map of the coordinates to the expected colors for the texel view.
+        const xMap = expMap.get(loc.x) ?? new Map<number, Map<number, E>>();
+        const yMap = xMap.get(loc.y) ?? new Map<number, E>();
+        assert(
+          !yMap.has(loc.z),
+          `duplicate pixel expectation at coordinate (${loc.x},${loc.y},${loc.z})`
+        );
+        yMap.set(loc.z, e.exp);
+        xMap.set(loc.y, yMap);
+        expMap.set(loc.x, xMap);
```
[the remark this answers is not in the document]
lokokung: Done

src/webgpu/gpu_test.ts
```diff
@@ -977,3 +991,193 @@ export class GPUTest extends Fixture<GPUTestSubcaseBatchState> {
...
+      }
+      const size: GPUExtent3D = [
+        upperCorner[0] - lowerCorner[0] + 1,
+        upperCorner[1] - lowerCorner[1] + 1,
+        upperCorner[2] - lowerCorner[2] + 1,
+      ];
+      let expTexelView: TexelView;
+      if (Symbol.iterator in exp[0].exp) {
+        expTexelView = TexelView.fromTexelsAsBytes(
+          src.texture.format as EncodableTextureFormat,
```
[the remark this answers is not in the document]
lokokung: Done

src/webgpu/gpu_test.ts
```diff
@@ -977,3 +878,193 @@ export class GPUTest extends Fixture<GPUTestSubcaseBatchState> {
...
+      const upperCorner = [0, 0, 0];
+      const expMap = new Map<number, Map<number, Map<number, E>>>();
+      const locs: Required<GPUOrigin3DDict>[] = [];
+      for (const e of exp) {
+        const loc = reifyOrigin3D(e.location);
+        locs.push(loc);
+
+        // Compute the minimum sub-rect that enconpasses all the pixel comparisons. The
+        // `lowerCorner` will become the origin, and the `upperCorner` will be used to compute the
+        // size. At the same time we create a texel view that
```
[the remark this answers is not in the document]
lokokung: Done

src/webgpu/gpu_test.ts
```diff
@@ -977,3 +878,193 @@ export class GPUTest extends Fixture<GPUTestSubcaseBatchState> {
...
+  createTextureFromTexelView(
+    texelView: TexelView,
+    desc: Omit<GPUTextureDescriptor, 'format'>
+  ): GPUTexture;
+
+  /**
+   * Creates a mipmapped texture where each mipmap level's (`i`) content is
+   * from `texelViews[i]` and tracks it for destruction for the test case.
+   */
+  createTextureFromTexelViews(
```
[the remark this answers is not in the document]
lokokung: Done

lokokung:
Ping for re-review as the conflicts might start piling up!

shrekshao (approved):
LGTM with the latest commit resolving comments from Austin and me.
Conflicts seem to be trivial:
pipeline_output_targets.spec.ts: import conflicts
sample_mask.spec.ts, anisotropy.spec.ts: test fn remove async from your earlier refactor

src/webgpu/api/operation/sampling/anisotropy.spec.ts
```diff
@@ -270,7 +272,7 @@ g.test('anisotropic_filter_mipmap_color')
       maxAnisotropy: 1,
       _results: [
         { coord: { x: xMiddle, y: 2 }, expected: colors[2] },
-        { coord: { x: xMiddle, y: 6 }, expected: [colors[0], colors[1]] },
+        { coord: { x: xMiddle, y: 6 }, expected: colors[0] },
```
kainino0x, line 275: Why this change? I think this makes the test stricter. Hopefully that doesn't make it fail on any devices but it's really hard to predict with the anisotropy stuff since its behavior is mostly unspecified.

src/webgpu/util/texture/layout.ts
```diff
@@ -107,7 +107,7 @@ export function getTextureSubCopyLayout(
   );
   assert(
     copySize_.width % blockWidth === 0 && copySize_.height % blockHeight === 0,
-    'copySize must be a multiple of the block size'
+    `copySize (${copySize_.width},${copySize_.height}) must be a multiple of the block size (${blockWidth},${blockHeight})`
```
kainino0x, line 110: nit: you can tag `() =>` on the front here to avoid actually computing the message string unless it's needed. I assume this is probably useful for performance sometimes though I've never tested it.

src/webgpu/util/texture/texel_view.ts
```diff
@@ -76,7 +76,7 @@ export class TexelView {
           coords.x < origin.x + size.width &&
           coords.y < origin.y + size.height &&
           coords.z < origin.z + size.depthOrArrayLayers,
-        'coordinate out of bounds'
+        `coordinate (${coords.x},${coords.y},${coords.z}) out of bounds`
```
kainino0x, line 79: same

src/webgpu/gpu_test.ts
```diff
@@ -517,13 +529,23 @@ export class GPUTest extends Fixture<GPUTestSubcaseBatchState> {
...
     format = resolvePerAspectFormat(format, layout?.aspect);
     const { byteLength, minBytesPerRow, bytesPerRow, rowsPerImage, mipSize } = getTextureCopyLayout(
       format,
       dimension,
       size,
       layout
     );
+    const copySize = [
+      mipSize[0],
+      dimension !== '1d' ? mipSize[1] : 1,
```
kainino0x, line 546: nit: can we just assert (or assume) that if dimension is 1d, then mipSize[1] is 1? I would think that's true...
Similar on the next line.

src/webgpu/gpu_test.ts
```diff
@@ -1065,3 +967,174 @@ export class GPUTest extends Fixture<GPUTestSubcaseBatchState> {
...
+      comparisonOptions = {
+        maxIntDiff: 0,
+        maxDiffULPsForNormFormat: 1,
+        maxDiffULPsForFloatFormat: 1,
+      }
+    ): void {
+      assert(exp.length > 0, 'must specify at least one pixel comparison');
+      assert(
+        (kEncodableTextureFormats as GPUTextureFormat[]).includes(src.texture.format),
+        `${src.texture.format} is not an encodable format`
```
kainino0x, line 1062: same in asserts in this function

kainino0x (approved):
Finally reviewed, no major comments!

Have you dry run this against chromium? it'll be a slight pain but worthwhile to make sure there aren't regressions, I think.

src/webgpu/gpu_test.ts
lokokung: For 1d we can, but for 2d we cannot because of 2d texture arrays. The underlying helper just blindly uses the value in the third dimension for 2d textures, and that causes the copy size to be wrong when we are dealing with a slice of the 2d texture array. I added a comment and a TODO to fix the underlying helper in another change.

src/webgpu/util/texture/texel_view.ts
lokokung: Done

src/webgpu/util/texture/layout.ts
lokokung: Done.

src/webgpu/api/operation/sampling/anisotropy.spec.ts
lokokung: Reverted it for now. My thought was that given it's at anisotropy=1, that the results should be relatively deterministic, but we can revisit this later since I need to revisit this separately anyways.

src/webgpu/gpu_test.ts
lokokung: Done.

lokokung:
Ran the entire suite on https://dawn-review.googlesource.com/c/dawn/+/117591?checksPatchset=5&tab=checks. The failures are all tests that were recently changed and not yet rolled-in/suppressed yet.

greggman:
I want to suggest that this should not be a mixin but instead should be a set a standalone functions that either

(a) get passed `GPUTest` (good) or 
(b) get passed a device and something lower level than GPUTest

I'm not sure I actually understand all the functions in the cts. When to use `assert` vs `t.expect`. I assumed was `assert` is like C++ assert. Just fail here because you can't continue and that `t.expect` was "like assert but includes info about the actual test hence it's a member function of `GPUTest`.

In any case, I'm writing tests that have their own device and I can't use any of these functions that required a GPUTest since they all use GPUTest.device instead of my device.

Further, I have functions that take a `GPUTest` soley to be able to call `t.expectXXX` but now that it's gotten more complicated it's not clear how I make a function that takes whatever is generated by `makeTestGroup(TextureTestMixin(GPUTest))`

Is there some concrete benefit to `expectSinglePixelComparisonsAreOkInTexture` being on an interface and mixing it in with GPUTest rather that it just being a standalone function? It feels like heading toward something like [inheritance hell](https://www.google.com/search?q=inheritance+hell&oq=inheritance+hell) ... I guess as a "mixin" it's not the same but the contortions needed to use it feel similar?

kainino0x:
There's a further step with the mixin refactor that I was hoping to do (but forgot about a long time ago) that should resolve this. I wanted to separate the "has a GPUDevice" mixin from the "provides its own GPUDevice" mixin and have things like TextureTestMixin only require the former.

There's a few goals with the mixin refactor (there may have been others, it's been too long since I started scheming it):
- Discoverability of helpers by similarity
- Avoid having to import tons of helpers separately
- Reduce some verbosity (and temptation to add fixture-level helpers wrapping them)

Though admittedly some of them could be achieved by just exporting one object full of static functions instead of a bunch of separate functions.

> I'm not sure I actually understand all the functions in the cts. When to use `assert` vs `t.expect`. I assumed was `assert` is like C++ assert. Just fail here because you can't continue 

This is theoretically the difference (except it's not like a C++ assert in that it never gets compiled out). In practice the boundary isn't clear because most `expect`s could be `assert`s without causing much problem. So in some cases we end up picking one arbitrarily.

> and that `t.expect` was "like assert but includes info about the actual test hence it's a member function of `GPUTest`.

In this respect it's no different because `assert` just throws an exception, which still gets caught in the specific test/case/subcase that threw it.

> Further, I have functions that take a `GPUTest` soley to be able to call `t.expectXXX` but now that it's gotten more complicated it's not clear how I make a function that takes whatever is generated by `makeTestGroup(TextureTestMixin(GPUTest))`

Hm, someone would need to experiment with the code to see whether there's a big blocker here. There's a way to name that type, some form of `GPUTest & TextureTestMixinType` or maybe `typeof TextureTestMixin(GPUTest)`, and that's a little obtuse but should at least work. The bigger concern is whether something like `TextureTestMixin(BufferTestMixin(GPUTest))` would match that type. That's something to check.

PR #2188, opened 2023-01-21
Adjust fuzzy value on canvas_colorspace_rgba16float ref test

lokokung:
Seems like according to https://web-platform-tests.org/writing-tests/reftests.html#fuzzy-matching specifying range of `0-1` should make it work since the results of the test are only off by 1. I think that specifying `1` only is expecting an error of at most 1/255 = 0?


austinEng (approved):

lokokung:
Upon more debugging and searching I found that the diff algo is this: https://source.chromium.org/chromium/chromium/src/+/main:tools/imagediff/image_diff.cc;l=389;bpv=1;bpt=0 and turns out that the number of different pixels need to be set (and even if they are different within the margin, it still counts).

greggman:
Haha, That's really dumb, because the pixel count is dependent on the devicePixelRatio so a dpr 3 screenshot will have 9x the differences of a dpr 1 screenshot

greggman:
Ah, others have noticed that too

https://github.com/web-platform-tests/wpt/issues/31457

PR #2200, opened 2023-01-25
Implements and updates maxColorAttachmentBytesPerSample tests

lokokung:
Changes:
- Implements new targeted maxColorAttachmentBytesPerSample tests for render pass, render pipelines, and render bundles.
- Rolls webgpu/types forwards for updated type names.
- Updates older tests to respect the new limit.

Issue: #2198 #2199
<hr>

**Requirements for PR author:**

- [x] All missing test coverage is tracked with "TODO" or `.unimplemented()`.
- [x] New helpers are `/** documented */` and new helper files are found in `helper_index.txt`.
- [x] Test behaves as expected in a WebGPU implementation. (If not passing, explain above.)

**Requirements for [reviewer sign-off](https://github.com/gpuweb/cts/blob/main/docs/reviews.md):**

- [ ] Tests are properly located in the test tree.
- [ ] [Test descriptions](https://github.com/gpuweb/cts/blob/main/docs/intro/plans.md) allow a reader to "read only the test plans and evaluate coverage completeness", and accurately reflect the test code.
- [ ] Tests provide complete coverage (including validation control cases). **Missing coverage MUST be covered by TODOs.**
- [ ] Helpers and types promote readability and maintainability.

PR #2207, opened 2023-01-26
Adds typescript-eslint/require-await and fixes existing warnings.

lokokung:
Issue: None

<hr>

### DIFF
--- a/src/unittests/loaders_and_trees.spec.ts
+++ b/src/unittests/loaders_and_trees.spec.ts
@@ -99,13 +99,17 @@ const specsData: { [k: string]: SpecFile } = {
 };
 
 class FakeTestFileLoader extends TestFileLoader {
-  async listing(suite: string): Promise<TestSuiteListing> {
-    return listingData[suite];
+  listing(suite: string): Promise<TestSuiteListing> {
+    return new Promise(resolve => {
+      resolve(listingData[suite]);
+    });
   }
 
-  async import(path: string): Promise<SpecFile> {
+  import(path: string): Promise<SpecFile> {
     assert(path in specsData, '[test] mock file ' + path + ' does not exist');
-    return specsData[path];
+    return new Promise(resolve => {
+      resolve(specsData[path]);
+    });
   }
 }
 
@@ -134,7 +138,7 @@ class LoadingTest extends UnitTest {
 
 export const g = makeTestGroup(LoadingTest);
 
-g.test('suite').fn(async t => {
+g.test('suite').fn(t => {
   t.shouldReject('Error', t.load('suite1'));
   t.shouldReject('Error', t.load('suite1:'));
 });
--- a/src/unittests/loaders_and_trees.spec.ts
+++ b/src/unittests/loaders_and_trees.spec.ts
@@ -100,16 +100,12 @@ const specsData: { [k: string]: SpecFile } = {
 
 class FakeTestFileLoader extends TestFileLoader {
   listing(suite: string): Promise<TestSuiteListing> {
-    return new Promise(resolve => {
-      resolve(listingData[suite]);
-    });
+    return Promise.resolve(listingData[suite]);
   }
 
   import(path: string): Promise<SpecFile> {
     assert(path in specsData, '[test] mock file ' + path + ' does not exist');
-    return new Promise(resolve => {
-      resolve(specsData[path]);
-    });
+    return Promise.resolve(specsData[path]);
   }
 }
 
--- a/src/unittests/test_group.spec.ts
+++ b/src/unittests/test_group.spec.ts
@@ -338,7 +338,7 @@ g.test('shouldReject').fn(async t0 => {
   g.test('a').fn(t => {
     t.shouldReject(
       'Error',
-      (() => {
+      (async () => {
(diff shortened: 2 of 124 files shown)


lokokung:
Ran the suite with the CL and seems to work fine: https://dawn-review.googlesource.com/c/dawn/+/117591?checksPatchset=3&tab=checks

Note that all the failures are from expected failing tests that landed in https://github.com/gpuweb/cts/pull/2200 (and has not been rolled yet). Most of them are related to `maxColorAttachmentBytesPerSample`, while one error is because of the webgpu/types roll that removed the `bgra8unorm-storage` feature, but because I forgot to update the test list for the run in the CL, seems like the runner just timed out on trying to run it.

shrekshao (approved):
LGTM. (I think the intention comes from me saying async is not always needed for the test function `.fn`?)

src/unittests/test_group.spec.ts
```diff
@@ -334,10 +335,10 @@ g.test('shouldReject').fn(async t0 => {
 
   const g = makeTestGroupForUnitTesting(UnitTest);
 
-  g.test('a').fn(async t => {
+  g.test('a').fn(t => {
     t.shouldReject(
       'Error',
-      (async () => {
+      (() => {
```
kainino0x, line 341: This one is changing the meaning of the test. Before it tests shouldReject makes the test 'a' fail, now I'm pretty sure shouldReject is not getting called at all and it's testing that if the test fn throws then 'a' fails

src/unittests/loaders_and_trees.spec.ts
```diff
@@ -99,13 +99,17 @@ const specsData: { [k: string]: SpecFile } = {
 };
 
 class FakeTestFileLoader extends TestFileLoader {
-  async listing(suite: string): Promise<TestSuiteListing> {
-    return listingData[suite];
+  listing(suite: string): Promise<TestSuiteListing> {
+    return new Promise(resolve => {
+      resolve(listingData[suite]);
+    });
```
kainino0x, line 105: nit:
```suggestion
    return Promise.resolve(listingData[suite]);
```

src/unittests/loaders_and_trees.spec.ts
```diff
@@ -99,13 +99,17 @@ const specsData: { [k: string]: SpecFile } = {
...
   }
 
-  async import(path: string): Promise<SpecFile> {
+  import(path: string): Promise<SpecFile> {
     assert(path in specsData, '[test] mock file ' + path + ' does not exist');
-    return specsData[path];
+    return new Promise(resolve => {
+      resolve(specsData[path]);
+    });
```
kainino0x, line 112:
```suggestion
    return Promise.resolve(specsData[path]);
```

src/webgpu/idl/idl_test.ts
```diff
@@ -10,6 +10,7 @@ interface UnknownObject {
  * Base fixture for testing the exposed interface is correct (without actually using WebGPU).
  */
 export class IDLTest extends Fixture {
+  // eslint-disable-next-line @typescript-eslint/require-await
```
kainino0x, line 13: weird this is necessary, without it is the lint breaking compatibility with the overridden method of Fixture?

Unfortunate if the lint isn't precise about this kind of thing. I feel like I've tried to use it before and this kind of issue was probably why I decided not to enable it.

kainino0x (approved):
lgtm with one fix

src/unittests/test_group.spec.ts
lokokung: Ah yea good catch, I think I forgot to revert this one after putting the disable for the lint in this file. Thanks!

src/unittests/loaders_and_trees.spec.ts
lokokung: Done. Oh nice! Didn't know about this syntax. Thanks!

src/unittests/loaders_and_trees.spec.ts
lokokung: Done. Ditto above.

src/webgpu/idl/idl_test.ts
lokokung: Yea, mainly the issue was the return type. I managed to get rid of the suppression by explicitly returning a `Promise.resolve()`. Lmk which one you prefer, the suppression or the explicit return.

lokokung: Actually.. with the `Promise.resolve()` workaround, I may be able to remove all the suppressions entirely also? Basically the problem I was having before was that I didn't know how to force a function to return a `Promise<void>` without the `async` in front of the function def if the function never calls `await` on anything. WDYT?

kainino0x: I thought about suggesting the Promise.resolve. It's a little silly here but no worse than the eslint-disable. For all the others I'd say let's just keep the eslint-disables.

kainino0x (approved):

src/webgpu/idl/idl_test.ts
lokokung: Ok will leave it as in and merge it in then! Thanks Kai!

PR #2217, opened 2023-01-31
Updates error type for create oom buffers.

lokokung:
<hr>


src/webgpu/api/operation/buffers/map_oom.spec.ts
```diff
@@ -38,7 +38,7 @@ g.test('mapAsync')
     const { oom, write, size, unmapBeforeResolve } = t.params;
 
     const buffer = t.expectGPUError(
-      'out-of-memory',
+      'validation',
```
kainino0x, line 41: Why? This seems like an out-of-memory error to me (trying to allocate a buffer of size ~8PiB or 128GiB)

lokokung: According to spec: https://www.w3.org/TR/webgpu/#abstract-opdef-validating-gpubufferdescriptor both of these are greater than maxBufferSize so it should actually be a validation error, not an oom error.

kainino0x: Oh, right, of course. This test was written before we had that limit. This change breaks the intent of the test though :( We should at least update the test description.

I guess we have no way to reliably hit an OOM, but we should add a TODO to have a test that enables the highest available maxBufferSize limit and at least _tries_ to get an OOM.

lokokung: Finally addressed the concerns here! Ultimately removed the old test entirely and moved it into stress. I tried to run the stress test and it occasionally passes? Other times it crashes the GPU process... The test itself should be ok though so let me know what you think.

src/stress/memory/oom.spec.ts
```diff
@@ -3,32 +3,157 @@ Stress tests covering robustness when available VRAM is exhausted.
...
+  mappedAtCreation: boolean
+) {
+  if (!mappable && !mappedAtCreation) {
+    return device.createBuffer({
+      size,
+      usage: GPUBufferUsage.UNIFORM,
+      mappedAtCreation,
+    });
+  }
+  let buffer: GPUBuffer;
```
austinEng, line 27: so here, we could have `mappable = false` but `mappedAtCreation = true` - but then we're using MAP_READ or MAP_WRITE usage. that doesn't make sense

src/stress/memory/oom.spec.ts
```diff
@@ -3,32 +3,157 @@ Stress tests covering robustness when available VRAM is exhausted.
...
+        await Promise.all(buffers.map(value => value.mapAsync(mode)));
+      } else {
+        buffers.forEach(value => {
+          t.expectValidationError(() => {
+            void value.mapAsync(mode);
+          });
+        });
+      }
+      // Finally, destroy all the buffers to free the resources.
+      buffers.forEach(buffer => buffer.destroy());
```
austinEng, line 91: I think the preferred way is to do `trackForCleanup(buffer)` after creating the buffers

src/stress/memory/oom.spec.ts
```diff
@@ -3,32 +3,157 @@ Stress tests covering robustness when available VRAM is exhausted.
...
+        try {
+          t.device.pushErrorScope('out-of-memory');
+          const buffer = createBuffer(t.device, kSize, mappable, mode, mappedAtCreation);
+          if (await t.device.popErrorScope()) {
+            errorBuffer = buffer;
+            break;
+          }
+          buffers.push(buffer);
+        } catch (ex) {
+          await finish();
```
austinEng, line 108: nit: maybe check that the exception was a RangeError

src/stress/memory/oom.spec.ts
```diff
@@ -3,32 +3,157 @@ Stress tests covering robustness when available VRAM is exhausted.
...
+    t.shouldThrow('OperationError', () => {
+      errorBuffer.getMappedRange();
+    });
+
+    // Should't be a validation error even if the buffer failed to be mapped.
+    errorBuffer.unmap();
+    errorBuffer.destroy();
+
+    // Finish the rest of the test w.r.t the mappable buffers.
+    await finish();
```
austinEng, line 155: this is the second time we do finish() in the mappedAtCreation state - is that intended>

lokokung: Ah... good catch, adding a return there to skip all the other stuff.

src/stress/memory/oom.spec.ts
lokokung: Done.

src/stress/memory/oom.spec.ts
lokokung: Done.

src/stress/memory/oom.spec.ts
lokokung: This is intentional. Basically the idea is that the buffer is re-mappable, but not mappable in it's current state. For the case you mention, it is not mappable because it is already mapped.

austinEng: Ah, I see. Maybe we should call the function `createBufferWithMapState` and instead use the [states](https://gpuweb.github.io/gpuweb/#enumdef-gpubuffermapstate) defined in the spec.

Saying "create buffer with mappable = true` reads like you're giving it a mappable usage, not that it's in a mappable state.

lokokung: Done.

austinEng (approved):

PR #2226, opened 2023-02-03
Remove tests of end pass twice in encoder_state.spec.ts

Richard-Yunchao:
End pass twice leads to validation error immediately. The validation error must be reported before finish(). And finish() might be valid. So, the tests are incorrect according to latest WebGPU spec.

See the discussion for end pass twice at this issue: https://github.com/gpuweb/gpuweb/issues/2464

And end pass twice related tests have already been covered in other test file. So this change simply removes corresponding tests in encoder_state.

Issue: #1914

<hr>

### DIFF
--- a/src/webgpu/api/validation/encoding/encoder_state.spec.ts
+++ b/src/webgpu/api/validation/encoding/encoder_state.spec.ts
@@ -52,8 +52,6 @@ g.test('pass_end_invalid_order')
     `
   Test that beginning a {compute,render} pass before ending the previous {compute,render} pass
   causes an error.
-
-  TODO: Update this test according to https://github.com/gpuweb/gpuweb/issues/2464
   `
   )
   .params(u =>
@@ -63,6 +61,8 @@ g.test('pass_end_invalid_order')
       .beginSubcases()
       .combine('firstPassEnd', [true, false])
       .combine('endPasses', [[], [0], [1], [0, 1], [1, 0]])
+      // Don't end the first pass multiple times (that generates a validation error but doesn't invalidate the encoder)
+      .unless(p => p.firstPassEnd && p.endPasses.includes(0))
   )
   .fn(t => {
     const { pass0Type, pass1Type, firstPassEnd, endPasses } = t.params;


Richard-Yunchao:
PTAL. Thanks! Dawn patch at https://dawn-review.googlesource.com/c/dawn/+/116842 depends on this CTS change.

kainino0x:
> Issue: #

#1914

src/webgpu/api/validation/encoding/encoder_state.spec.ts
```diff
@@ -52,8 +52,6 @@ g.test('pass_end_invalid_order')
     `
   Test that beginning a {compute,render} pass before ending the previous {compute,render} pass
   causes an error.
-
-  TODO: Update this test according to https://github.com/gpuweb/gpuweb/issues/2464
   `
   )
   .params(u =>
```
kainino0x, line 57: I think it would be equivalent and simpler to modify the subcases instead:

```
      // Don't end the first pass multiple times (that generates a validation error but doesn't invalidate the encoder)
      .unless(p => p.firstPassEnd && p.endPasses.includes(0))
```

kainino0x:
> And end pass twice related tests have already been covered in other test file.

Just to verify, do the other tests also test that the encoder is kept valid when a pass is ended twice?

Another alternative to the suggestion above would be to keep these cases but change the expectations of when errors come from .end() and .finish(), but it would be unnecessarily complicated if we already have coverage.

src/webgpu/api/validation/encoding/encoder_state.spec.ts
Richard-Yunchao: Good point. Done!

Richard-Yunchao:
> 

Yes. The end twice tests also test that the encoder is valid when a pass is ended twice, see line 162 at https://github.com/gpuweb/cts/pull/1917/files. It is actually in the same file (pass_end_twice in encoder_state.spec.ts). 

Hmm, it is not necessary to make it complicated, because pass end twice is already covered.

Richard-Yunchao:
> > And end pass twice related tests have already been covered in other test file.
> 
> Just to verify, do the other tests also test that the encoder is kept valid when a pass is ended twice?
> 
> Another alternative to the suggestion above would be to keep these cases but change the expectations of when errors come from .end() and .finish(), but it would be unnecessarily complicated if we already have coverage.

Yes. The end twice tests also test that the encoder is valid when a pass is ended twice, see line 162 at https://github.com/gpuweb/cts/pull/1917/files. It is actually in the same file (pass_end_twice in encoder_state.spec.ts).

Hmm, it is not necessary to make it complicated, because pass end twice is already covered.

kainino0x (approved):

PR #2230, opened 2023-02-06
Avoid writable storage buffer binding aliasing

shrekshao:
Issue: https://github.com/gpuweb/cts/issues/2232

Found during Chromium implementation. Reflecting the latest spec changes.

<hr>

### DIFF
--- a/src/webgpu/api/validation/resource_usages/buffer/in_pass_encoder.spec.ts
+++ b/src/webgpu/api/validation/resource_usages/buffer/in_pass_encoder.spec.ts
@@ -231,6 +231,11 @@ referenced by that bind group is "used" in the usage scope. `
         ) {
           return false;
         }
+
+        // Avoid writable storage buffer bindings aliasing.
+        if (t.usage0 === 'storage' && t.usage1 === 'storage') {
+          return false;
+        }
         return true;
       })
       .combine('hasOverlap', [true, false])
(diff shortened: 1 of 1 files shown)


kainino0x (commented):
Do we have another test for this behavior? If not this seems like an easy place to test it, even though it's a different rule from what this was written to test. Will just want to update the description here to note it's tested here along with the resource usage rules, and a note somewhere else (in draw/dispatch validation tests?) noting that it's tested here.

shrekshao:
> Do we have another test for this behavior?
Not yet.

In terms of that, I will defer this change to submit together with the other buffer binding aliasing tests. There is quite a set of combinations need testing.

In the meantime I will just suppress the cts failures for the implementation so everything should be fine.

kainino0x:
It's fine to land this now as long as we have an issue tracked to write buffer binding aliasing tests.

src/webgpu/api/validation/resource_usages/buffer/in_pass_encoder.spec.ts
```diff
@@ -231,6 +231,11 @@ referenced by that bind group is "used" in the usage scope. `
         ) {
           return false;
         }
+
```
kainino0x, line 234: In the test description for both of these tests, add a link to the issue that tracks the new tests (doesn't need to be a TODO, just something so there's a followable reference to the new tests).

kainino0x (approved):

Issue #2195, opened 2023-01-25
val: capability_checks,limits


greggman:
Relevant spec PRs for the tests that these tests should verify coverage for:

- https://github.com/gpuweb/gpuweb/pull/3318
- https://github.com/gpuweb/gpuweb/pull/3366

greggman:
what is needed for this test? Is this just a test that the values are within some range? That they exist and are >= min?

kainino0x:
The title refers to an existing README:
https://github.com/gpuweb/cts/blob/main/src/webgpu/api/validation/capability_checks/limits/README.txt

The idea with `capability_checks` was to test all of the actual validation inside method calls that depends on limits, for example the max texture size check in createTexture. This doesn't necessarily have to be centralized in one place, and it probably makes more sense to write these in the files that test validation for each method.
The reason for centralizing was probably that we didn't have a way to request higher limits on the device yet. We do now, though we still don't have really good ways of parameterizing most tests over limits.

Note I think `capability_checks,features` did mostly get distributed across the test suite, though apparently we still have some here, and a stub leftover.

greggman:
Here's a start. Is this looking good or do you want something else?

https://github.com/greggman/cts/commit/18c1eb4f069b6d88100d2ccbaa64cbb39e7ec216

kainino0x:
@greggman those look great. I think we should add the following cases to the atLimit/overLimit cases you have:

- request the base value: it's easy to add this here, and easier to be sure we have complete coverage than when base limit tests are distributed throughout (makes sure implementations are enforcing requested limits and not allowing the max limits)
- request lower than the base value: implementation is supposed to ignore this and provide the base limit instead
- maybe one that's between the base value and the adapter limit? less necessary

PR #2236, opened 2023-02-08
add maxBufferSize limit test

greggman:
Note: this fails on Chrome. It fails where it asks for one more than the default limit (but not 1 more than the max limit).

Issue: #2195

<hr>

### DIFF
--- a/src/webgpu/api/validation/capability_checks/limits/maxBufferSize.spec.ts
+++ b/src/webgpu/api/validation/capability_checks/limits/maxBufferSize.spec.ts
@@ -1,0 +1,28 @@
+import { kLimitBaseParams, makeLimitTestGroup } from './limit_utils.js';
+
+const limit = 'maxBufferSize';
+export const { g, description } = makeLimitTestGroup(limit);
+
+g.test('createBuffer,at_over')
+  .desc(`Test using at and over ${limit} limit`)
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/capability_checks/limits/maxBufferSize.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+      limitTest,
+      testValueName,
+      async ({ device, testValue, actualLimit, shouldError }) => {
+        await t.testForValidationErrorWithPossibleOutOfMemoryError(
+          () => {
+            const buffer = device.createBuffer({
+              usage: GPUBufferUsage.VERTEX,
+              size: testValue,
+            });
+            buffer.destroy();
```
kainino0x, line 21: Usually we avoid having destroy() calls inside the same error scope with the call we're testing, but in this case it's annoying enough to fix that I don't think it's worth fixing.

kainino0x (approved):

src/webgpu/api/validation/capability_checks/limits/maxBufferSize.spec.ts
kainino0x: Also I don't know why I didn't think about this until now.

PR #2238, opened 2023-02-08
add maxXXXPerShaderStage limit tests

greggman:
Issue: #2195 

random comment: Seems like I should be able to use `GPUShaderStage.FRAGMENT` instead of `GPUConst.ShaderStage.FRAGMENT`. Does something need to be updated?

<hr>

### DIFF
--- a/src/common/util/util.ts
+++ b/src/common/util/util.ts
@@ -219,6 +219,41 @@ export function mapLazy<T, R>(xs: Iterable<T>, f: (x: T) => R): Iterable<R> {
   };
 }
 
+const ReorderOrders = {
+  forward: true,
+  backward: true,
+  shiftByHalf: true,
+};
+export type ReorderOrder = keyof typeof ReorderOrders;
+export const kReorderOrderKeys = keysOf(ReorderOrders);
+
+/**
+ * Creates a new array from the given array with the first half
+ * swapped with the last half.
+ */
+export function shiftByHalf<R>(arr: R[]): R[] {
+  const len = arr.length;
+  const half = (len / 2) | 0;
+  const firstHalf = arr.splice(0, half);
+  return [...arr, ...firstHalf];
+}
+
+/**
+ * Creates a reordered array from the input array based on the Order
+ */
+export function reorder<R>(order: ReorderOrder, arr: R[]): R[] {
+  switch (order) {
+    case 'forward':
+      return arr.slice();
+    case 'backward':
+      return arr.slice().reverse();
+    case 'shiftByHalf': {
+      // should this be pseudo random?
+      return shiftByHalf(arr);
+    }
+  }
+}
+
 const TypedArrayBufferViewInstances = [
   new Uint8Array(),
   new Uint8ClampedArray(),
--- a/src/webgpu/api/validation/capability_checks/limits/limit_utils.ts
+++ b/src/webgpu/api/validation/capability_checks/limits/limit_utils.ts
@@ -2,7 +2,7 @@ import { kUnitCaseParamsBuilder } from '../../../../../common/framework/params_b
 import { makeTestGroup } from '../../../../../common/framework/test_group.js';
 import { keysOf } from '../../../../../common/util/data_tables.js';
 import { getGPU } from '../../../../../common/util/navigator_gpu.js';
-import { assert } from '../../../../../common/util/util.js';
+import { assert, range, reorder, ReorderOrder } from '../../../../../common/util/util.js';
 import { kLimitInfo } from '../../../../capability_info.js';
 import { GPUTestBase } from '../../../../gpu_test.js';
 
@@ -10,25 +10,19 @@ type GPUSupportedLimit = keyof GPUSupportedLimits;
 
 const CreatePipelineTypes = {
   createRenderPipeline: true,
+  createRenderPipelineWithFragmentStage: true,
   createComputePipeline: true,
 };
 export type CreatePipelineType = keyof typeof CreatePipelineTypes;
-
-export const kCreatePipelineTypes = [
-  'createRenderPipeline',
-  'createComputePipeline',
-] as CreatePipelineType[];
+export const kCreatePipelineTypes = keysOf(CreatePipelineTypes);
 
 const CreatePipelineAsyncTypes = {
   createRenderPipelineAsync: true,
+  createRenderPipelineAsyncWithFragmentStage: true,
   createComputePipelineAsync: true,
 };
 export type CreatePipelineAsyncType = keyof typeof CreatePipelineAsyncTypes;
-
-export const kCreatePipelineAsyncTypes = [
-  'createRenderPipelineAsync',
-  'createComputePipelineAsync',
-] as CreatePipelineAsyncType[];
+export const kCreatePipelineAsyncTypes = keysOf(CreatePipelineAsyncTypes);
 
 const RenderEncoderTypes = {
   render: true,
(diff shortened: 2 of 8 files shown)


kainino0x:
> random comment: Seems like I should be able to use `GPUShaderStage.FRAGMENT` instead of `GPUConst.ShaderStage.FRAGMENT`. Does something need to be updated?

We want to be able to generate the test list in node without plugins/polyfills (for presubmit testing, checking for duplicate cases, etc), and browsers without WebGPU enabled (for viewing the test plan). Before adding GPUConst, in Node we polyfilled the global scope, but it felt sketchy to do this since it could interfere with actual test correctness. So I ended up doing this when making the 'test plan viewer' use case work.

src/common/util/util.ts
```diff
@@ -219,6 +219,50 @@ export function mapLazy<T, R>(xs: Iterable<T>, f: (x: T) => R): Iterable<R> {
...
+ * Reorders an array, in place, based on the Order
+ */
+export function reorder<R>(order: Order, arr: R[]): R[] {
+  switch (order) {
+    case 'forward':
+      return arr;
+    case 'backward':
+      return arr.reverse();
+    case 'random': {
+      // should this be pseudo random?
```
kainino0x, line 260: That would be preferable to avoid inherent flakiness, but may mean you need more iterations for better coverage.

In this case I think it would be sufficient to just do something targeted like [1 ... n-1, 0] or [n/2 ... n-1, 0 ... n/2-1]

src/common/util/util.ts
```diff
@@ -219,6 +219,50 @@ export function mapLazy<T, R>(xs: Iterable<T>, f: (x: T) => R): Iterable<R> {
...
+    const other = randInt(i, len);
+    [arr[i], arr[other]] = [arr[other], arr[i]];
+  }
+  return arr;
+}
+
+/**
+ * Reorders an array, in place, based on the Order
+ */
+export function reorder<R>(order: Order, arr: R[]): R[] {
```
kainino0x, line 253: nit: Since this modifies the array in place I think it would be clearer if it didn't return it back.

src/webgpu/api/validation/capability_checks/limits/maxSampledTexturesPerShaderStage.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+                binding: i,
+                visibility,
+                texture: {},
+              }))
+            ),
+          });
+        }, shouldError);
+      }
+    );
+  });
```
kainino0x, line 43: Same as the other PR, will need to test automatic layouts coming from create\*pipeline\*

Should probably be a followup, since it's going to be a bunch more work.

src/webgpu/api/validation/capability_checks/limits/maxSampledTexturesPerShaderStage.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+    `
+  Test using at and over ${limit} limit in createBindGroupLayout
+  
+  Note: We also test order to make sure the implementation isn't just looking
+  at just the last entry.
+  `
+  )
+  .params(
+    kLimitBaseParams
+      .combine('visibility', kShaderStageCombinationsWithStage)
```
kainino0x, line 20: We might want cases like: maxStorageBuffersPerShaderStage vertex bindings and maxStorageBuffersPerShaderStage+1 fragment bindings.

But ISTR we already have tests for that up to the base limit, so maybe that's enough.

Same comment for all of them

src/common/util/util.ts
greggman: switched to something not random

src/common/util/util.ts
greggman: switched to returning a new array. It's more convenient to use

src/webgpu/api/validation/capability_checks/limits/maxSampledTexturesPerShaderStage.spec.ts
greggman: I will add after the bind limits test is merged as it has the utils need to make this easier

src/webgpu/api/validation/capability_checks/limits/maxSampledTexturesPerShaderStage.spec.ts
greggman: done

src/common/util/util.ts
```diff
@@ -219,6 +219,41 @@ export function mapLazy<T, R>(xs: Iterable<T>, f: (x: T) => R): Iterable<R> {
   };
 }
 
+const Orders = {
+  forward: true,
+  backward: true,
+  shiftByHalf: true,
+};
+export type Order = keyof typeof Orders;
```
Kangz, line 227: nit: order seems a bit generic. Also is it really needed since reorder could type `order` as 
"forward" | "backward" | "shiftByHalf"? (if there's a `export ReorderOrderType = [...] as const`?)

src/webgpu/api/validation/capability_checks/limits/limit_utils.ts
```diff
@@ -45,6 +39,194 @@ const EncoderTypes = {
...
+  extraWGSL = ''
+) {
+  return getPerStageWGSLForBindingCombinationImpl(
+    bindingCombination,
+    order,
+    bindGroupTest,
+    storageDefinitionWGSLSnippetFn,
+    (numBindings: number, set: number) => {
+      return bindingCombination === 'compute'
+        ? `d = ${range(numBindings, i => usageWGSLSnippetFn(i, set)).join(' +\n')};`
```
Kangz, line 198: Another possibility that could be simpler is to use a bunch of phony assignments

src/webgpu/api/validation/capability_checks/limits/maxSampledTexturesPerShaderStage.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+          order,
+          bindGroupTest,
+          (i, j) => `var u${j}_${i}: texture_2d<f32>`,
+          (i, j) => `textureLoad(u${j}_${i}, vec2u(0), 0).x`,
+          testValue
+        );
+        const module = device.createShaderModule({ code });
+
+        await t.shouldRejectConditionally(
+          'OperationError',
```
Kangz, line 127: ditto GPUPipelineError here and in other files.

src/webgpu/api/validation/capability_checks/limits/maxSampledTexturesPerShaderStage.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+  kBindingCombinations,
+  getPipelineTypeForBindingCombination,
+  getPipelineAsyncTypeForBindingCombination,
+  getPerStageWGSLForBindingCombination,
+} from './limit_utils.js';
+
+const limit = 'maxSampledTexturesPerShaderStage';
+export const { g, description } = makeLimitTestGroup(limit);
+
+g.test('createBindGroupLayout,at_over')
```
Kangz, line 17: Does the spec also validate the limits at pipeline layout creation? I don't know that this is tested at the moment.

src/webgpu/api/validation/capability_checks/limits/maxStorageBuffersPerShaderStage.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+  `
+  )
+  .params(
+    kLimitBaseParams
+      .combine('visibility', [
+        GPUConst.ShaderStage.FRAGMENT,
+        GPUConst.ShaderStage.COMPUTE,
+        GPUConst.ShaderStage.FRAGMENT | GPUConst.ShaderStage.COMPUTE,
+      ])
+      .combine('type', ['storage', 'read-only-storage'] as GPUBufferBindingType[])
```
Kangz, line 33: We could test readonlystorage for vertex as well. It should be allowed.

Kangz (approved):

src/common/util/util.ts
greggman: I'm not sure what you're suggesting here. Get rid of `reorder`? rename `Order` to `ReorderOrderType`? make reorder just take 3 strings?

src/webgpu/api/validation/capability_checks/limits/maxSampledTexturesPerShaderStage.spec.ts
greggman: When I change it I get

> THREW OperationError, instead of GPUPipelineError:

Is that bug or should the test stay `OperationError`?

src/webgpu/api/validation/capability_checks/limits/limit_utils.ts
greggman: thanks. I didn't know about `_` until recently 😅

fixed.

src/webgpu/api/validation/capability_checks/limits/maxSampledTexturesPerShaderStage.spec.ts
greggman: Yes, good catch! Will add test

src/webgpu/api/validation/capability_checks/limits/maxStorageBuffersPerShaderStage.spec.ts
greggman: added

src/webgpu/api/validation/capability_checks/limits/maxSampledTexturesPerShaderStage.spec.ts
greggman: added

src/common/util/util.ts
greggman: I renamed it `ReorderOrderType`. Happy to change it

src/webgpu/api/validation/capability_checks/limits/maxSampledTexturesPerShaderStage.spec.ts
Kangz: It is a Chromium bug fixed by https://chromium-review.googlesource.com/c/chromium/src/+/4277962

src/common/util/util.ts
Kangz: Thanks, that works, I was mostly worried about the overly generic name confusing people.

PR #2242, opened 2023-02-09
Use new test video files for ExternalTexture

jchen10:
The pixel values in old video files don't accurately match the expectation.

Issue: #

<hr>


src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -27,47 +27,58 @@ const kVideoInfo = /* prettier-ignore */ makeTable(
...
+  'four-colors-vp9-bt709.webm'  : [    'REC709',         'video/webm; codecs=vp9'],
+  'four-colors-vp9-bt2020.webm' : [   'REC2020',         'video/webm; codecs=vp9']
 } as const);
 type VideoName = keyof typeof kVideoInfo;
 type VideoInfo = valueof<typeof kVideoInfo>;
 
+const kBt601Red = new Uint8Array([248, 36, 0, 255]);
+const kBt601Green = new Uint8Array([64, 252, 0, 255]);
+const kBt601Blue = new Uint8Array([26, 35, 255, 255]);
+const kBt601Yellow = new Uint8Array([254, 253, 0, 255]);
```
jchen10, line 43: These values of expectation are calculated from https://chromium-review.googlesource.com/c/chromium/src/+/4232318/3/ui/gfx/color_transform_unittest.cc#142

jchen10:
The "compute" cases seem not working yet. Please hold for a while to review and merge.

jchen10:
> The "compute" cases seem not working yet. Please hold for a while to review and merge.

The old cases are already broken on my Win11 machine, so this PR is likely irrelevant.

jchen10:
> 1. I am seeing a failure on Windows for BT.709 with VideoFrame source. This does not fail when using a HTMLVideoElement source and I haven't been able to understand why.
> 2. The BT.2020 video included in this PR appears to use a PQ transfer function, which hasn't been implemented in Dawn/Tint. The BT.2020 video you are replacing used a standard gamma curve.

I am also seeing the same BT.709 VideoFrame failure. It might be something related to WebCodecs. We will need to spend some time investigating that.

For the BT.2020 case, we can add it later until we can support importing HDR video. It doesn't make much sense if we don't use PQ/HLG for HDR video.

kainino0x (commented):
Procedurally LGTM, but I'll let y'all work out getting the video files and expectations correct. It's OK if we need to increase thresholds but hopefully we can figure out what the correct values should be.

kainino0x:
It is OK to land test cases, like the PQ one, that don't pass in Chromium, as long as we are confident in the correctness of the test (which I know is difficult).

src/webgpu/web_platform/external_texture/video.spec.ts
shaoboyan: Please provide the comments that how these video files generated, where the expected value comes from.

src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -27,46 +27,58 @@ const kVideoInfo = /* prettier-ignore */ makeTable(
...
+    _greenExpectation: kBt601Green,
+    _blueExpectation: kBt601Blue,
+    _yellowExpectation: kBt601Yellow,
   },
   {
-    videoName: 'red-green.bt2020.vp9.webm',
-    _redExpectation: new Uint8Array([0xff, 0x00, 0x00, 0xff]),
-    _greenExpectation: new Uint8Array([0x00, 0xff, 0x00, 0xff]),
+    videoName: 'four-colors-vp9-bt709.webm',
+    _redExpectation: new Uint8Array([255, 0, 1, 255]),
```
bjjones, line 78: I disagree with the expectations here. When converting BT.709 primaries to sRGB, the colors should remain "pure". Because each channel is either '1' or '0', converting according to the gamma curve either way should not have any effect. BT.709 uses the exact same gamut as sRGB, so any "pure" source color will still be "pure" after conversion.

I think this is a case where we need to allow a small tolerance.

bjjones:
I spent some time working out the math for the BT.601 -> sRGB cases. I have attached an [Excel spreadsheet](https://github.com/gpuweb/cts/files/10736802/BT.601.to.sRGB.Worksheet.xlsx) I made to calculate all of these. 

| Color | BT.601 | sRGB |
| ----------- | ----------- | ----------- |
| Red|  255, 0, 0 | 248, 36, 0 |
| Green | 0, 255, 0 |  63, 251, 0 |
| Blue |  0, 0, 255 | 26, 35, 255 |
| Yellow | 255, 255, 0 | 254, 253, 0 |

To conclude, the expectations from the previous tests were correct. The expectations in this current PR are also correct, with the exception of `kBt601Green`, which is off by 1 in the R and G channels.

------------------------------------------
If anyone would like to check my work, here is the full process followed in accordance with [1]:

Linearize source channels using BT.601 inverse transfer function [2]:
X/4.5 for X < 0.081
((X+0.099)/1.099)^(1/.45) for X >= 0.081

Multiply by the BT.601 to XYZ matrix [3]:
0.4163290  0.3931464  0.1547446
0.2216999  0.7032549  0.0750452
0.0136576  0.0913604  0.7201920

Multiply by the inverse XYZ to sRGB matrix [3]:
3.1338561 -1.6168667 -0.4906146
-0.9787684  1.9161415  0.0334540
0.0719453 -0.2289914  1.4052427

Encode each channel according to the sRGB Transfer Function [4]:
12.92C for C <= 0.0031308
1.055C^(1/2.4) - 0.055 for C > 0.0031308

[1] https://skia.org/docs/user/color/
[2] https://en.wikipedia.org/wiki/Rec._709#Transfer_characteristics (Note I'm using BT.709 as a reference, which is okay because BT.709 has the same OETF as BT.601)
[3] http://www.brucelindbloom.com/index.html?Eqn_RGB_XYZ_Matrix.html
[4] https://en.wikipedia.org/wiki/SRGB#From_CIE_XYZ_to_sRGB

src/webgpu/web_platform/external_texture/video.spec.ts
jchen10: I introduced this (1 / 255) diff for "unorm->float" precision loss as revealed in https://chromium-review.googlesource.com/c/chromium/src/+/4232318/4/ui/gfx/color_transform_unittest.cc#227.
However I agree that it's kind of distracting. Actually the new test cases can still pass without this. So let me revert it.

jchen10:
> I spent some time working out the math for the BT.601 -> sRGB cases. I have attached an [Excel spreadsheet](https://github.com/gpuweb/cts/files/10736802/BT.601.to.sRGB.Worksheet.xlsx) I made to calculate all of these.
> 
> Color	BT.601	sRGB
> Red	255, 0, 0	248, 36, 0
> Green	0, 255, 0	63, 251, 0
> Blue	0, 0, 255	26, 35, 255
> Yellow	255, 255, 0	254, 253, 0
> To conclude, the expectations from the previous tests were correct. The expectations in this current PR are also correct, with the exception of `kBt601Green`, which is off by 1 in the R and G channels.
> 
> If anyone would like to check my work, here is the full process followed in accordance with [1]:
> 
> 1. Linearize source channels using BT.601 inverse transfer function [2]:
>    X/4.5 for X < 0.081
>    ((X+0.099)/1.099)^(1/.45) for X >= 0.081
> 2. Multiply by the BT.601 to XYZ matrix [3]:
>    0.4163290  0.3931464  0.1547446
>    0.2216999  0.7032549  0.0750452
>    0.0136576  0.0913604  0.7201920
> 3. Multiply by the inverse XYZ to sRGB matrix [3]:
>    3.1338561 -1.6168667 -0.4906146
>    -0.9787684  1.9161415  0.0334540
>    0.0719453 -0.2289914  1.4052427
> 4. Encode each channel according to the sRGB Transfer Function [4]:
>    12.92C for C <= 0.0031308
>    1.055C^(1/2.4) - 0.055 for C > 0.0031308
> 
> [1] https://skia.org/docs/user/color/ [2] https://en.wikipedia.org/wiki/Rec._709#Transfer_characteristics (Note I'm using BT.709 as a reference, which is okay because BT.709 has the same OETF as BT.601) [3] http://www.brucelindbloom.com/index.html?Eqn_RGB_XYZ_Matrix.html [4] https://en.wikipedia.org/wiki/SRGB#From_CIE_XYZ_to_sRGB

I believe your calculation is right. Maybe you also need to check whether the test video files have expected colors.

jchen10:
Except for the "expired" cases, all "sample" and "compute" cases are clean now with this new test cases. The found bugs in chromium/dawn have been all identified and fixed. We will be able to pass them on various OSes regardless of 1-copy or 0-copy.

PTAL

src/webgpu/web_platform/external_texture/video.spec.ts
bjjones: I think the reason Shaobo wanted to document  where the expectations come from is so people (that aren't us) can arrive at the same value on their own, and I don't think that linking to a CL with the values hardcoded accomplishes this. One idea is just adding the process described in my [comment ](https://github.com/gpuweb/cts/pull/2242#issuecomment-1430382811)into a comment in the file.

bjjones:
I think it's unusual to add the video file for BT.2020 if we aren't adding the corresponding test. Otherwise LGTM.

src/resources/README.md
```diff
@@ -1,2 +1,11 @@
 Always use `getResourcePath()` to get the appropriate path to these resources depending
 on the context (WPT, standalone, worker, etc.)
+
+
+The test video files were generated from https://source.chromium.org/chromium/chromium/src/+/main:media/test/data/four-colors.png with the ffmpeg cmds below:
```
bjjones, line 5: nit: Consider just including the four-colors.png file. I don't think we should hard link chromium if we don't have to.

src/webgpu/web_platform/external_texture/video.spec.ts
jchen10: > I think the reason Shaobo wanted to document where the expectations come from is so people (that aren't us) can arrive at the same value on their own, and I don't think that linking to a CL with the values hardcoded accomplishes this. One idea is just adding the process described in my [comment ](https://github.com/gpuweb/cts/pull/2242#issuecomment-1430382811)into a comment in the file.

Sounds good, done.

jchen10:
> I think it's unusual to add the video file for BT.2020 if we aren't adding the corresponding test. Otherwise LGTM.

BT.2020 removed.

src/resources/README.md
jchen10: Done!

jchen10:
> I think it's unusual to add the video file for BT.2020 if we aren't adding the corresponding test. Otherwise LGTM.

@bjjones If no other concern, I'd like to try to land this PR.

@kainino0x PTAL

bjjones (approved):
No further concerns from me. Do wait for someone else to review before merging though.

austinEng (approved):
LGTM

jchen10:
> LGTM

Thanks.

kainino0x:
> I spent some time working out the math for the BT.601 -> sRGB cases. I have attached an [Excel spreadsheet](https://github.com/gpuweb/cts/files/10736802/BT.601.to.sRGB.Worksheet.xlsx) I made to calculate all of these.

This is awesome. Thanks for providing the spreadsheet, I wanted the more precise float numbers for the R/G/B/Y constants so I opened it up and plugged them in.

| Color | BT.601 | sRGB |
| ----------- | ----------- | ----------- |
| Red | 1, 0, 0 | 0.972945567233341, 0.141794376683341, -0.0209589916711088  |
| Green | 0, 1, 0 | 0.248234279433399, 0.984810378661784, -0.0564701319494314 |
| Blue |  0, 0, 1 | 0.10159735826538, 0.135451122863674, 1.00262982899724 |
| Yellow | 1, 1, 0 | 0.995470750775951, 0.992742114518355, -0.0774291236205402 |

PR #2278, opened 2023-02-15
Add f32 matrix addition tests

zoddicus:
Issue #1626

<hr>

### DIFF
--- a/src/unittests/f32_interval.spec.ts
+++ b/src/unittests/f32_interval.spec.ts
@@ -12,7 +12,7 @@ import {
   acoshAlternativeInterval,
   acoshPrimaryInterval,
   additionInterval,
-  additionMatrixPairInterval,
+  additionMatrixInterval,
   asinInterval,
   asinhInterval,
   atanInterval,
@@ -3570,7 +3570,7 @@ interface MatrixPairCase {
   expected: IntervalBounds[][] | number[][];
 }
 
-g.test('additionMatrixMatrixInterval')
+g.test('additionMatrixInterval')
   .paramsSubcasesOnly<MatrixPairCase>([
     // Only testing that different shapes of matrices are handled correctly, to
     // reduce test duplication. This function uses AdditionIntervalOp for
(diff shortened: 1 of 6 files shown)


zoddicus:
I originally also implemented Matrix-Scalar addition in this patch, but that is not actually defined in the spec, so ripped it out. Please call out any code smells that looks like it might be from there being two types of matrix addition in the original patch.

src/unittests/f32_interval.spec.ts
```diff
@@ -3563,3 +3564,200 @@ g.test('transposeInterval')
...
+  input: [number[][], number[][]];
+  expected: IntervalBounds[][] | number[][];
+}
+
+g.test('additionMatrixInterval')
+  .paramsSubcasesOnly<MatrixPairCase>([
+    // Only testing that different shapes of matrices are handled correctly, to
+    // reduce test duplication. This function uses AdditionIntervalOp for
+    // calculating intervals, so depending on additionInterval's testing to
+    // make sure that things like subnormals are handled as expected.
```
amaiorano, line 3578: Just checking, but this comment about `AdditionIntervalOp` is mostly informative, right, since none of the test cases would actually exercise subnormal handling?

src/webgpu/shader/execution/expression/binary/matrix_arithmetic.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+import { additionMatrixInterval } from '../../../../util/f32_interval.js';
+import { sparseMatrixF32Range } from '../../../../util/math.js';
+import { makeCaseCache } from '../case_cache.js';
+import { allInputSources, generateMatrixPairToMatrixCases, run } from '../expression.js';
+
+import { binary } from './binary.js';
+
+export const g = makeTestGroup(GPUTest);
+
+export const d = makeCaseCache('binary/matrix_arithmetic', {
```
amaiorano, line 17: I know that for now, we're only handling f32 matrices, but eventually we will support and test AFloat and f16 matrices. Do you think it would make sense to name the files and tests with 'f32' now?

amaiorano (approved):
Code looks great. Left a couple of comments, but nothing that should stop this from landing.

src/unittests/f32_interval.spec.ts
zoddicus: That is correct, I have cleaned up the comment a bit to hopefully make it clearer.

src/webgpu/shader/execution/expression/binary/matrix_arithmetic.spec.ts
zoddicus: I have changed the file and cache names to be f32_matrix_arithmetic. I have left the test names as is, since the existing convention is 'f32_arithmetic/addition', and not 'f32_arithmetic/f32_addition', since we are assuming that abstract float and f16 will end up in their own files in the future.

PR #2301, opened 2023-02-22
Add f32 matrix-matrix multiplication tests

zoddicus:
Issue #1626

<hr>


src/unittests/f32_interval.spec.ts
```diff
@@ -3958,14 +3959,512 @@ g.test('subtractionMatrixInterval')
...
+          [1, 2],
+          [3, 4],
+        ],
+        [
+          [10, 20],
+          [30, 40],
+        ],
+      ],
+      expected: [
+        [70, 100],
```
ben-clayton, line 3981: So, without spending too many cycles thinking about column / row major, and just looking at your outputs:

```
 ╭      ╮   ╭        ╮   ╭          ╮
 │ 1, 2 │ x │ 10, 20 │ = │ 70,  100 │
 │ 3, 4 │   │ 30, 40 │   │ 150, 220 │
 ╰      ╯   ╰        ╯   ╰          ╯

 dot([1, 2], [10, 30]) =  70
 dot([1, 2], [20, 40]) = 100
 dot([3, 4], [10, 30]) = 150
 dot([3, 4], [20, 40]) = 220

```

Or in symbols:

```
 ╭      ╮   ╭      ╮   ╭      ╮
 │ a, b │ x │ e, f │ = │ i, j │
 │ c, d │   │ g, h │   │ k, l │
 ╰      ╯   ╰      ╯   ╰      ╯

 dot([a, b], [e, g]) = i
 dot([a, b], [f, h]) = j
 dot([c, d], [e, g]) = k
 dot([c, d], [f, h]) = l
```

So <code>result<sub>[y,x]</sub> = dot(LHS<sub>row y</sub>, RHS<sub>column x</sub>)</code>

src/unittests/f32_interval.spec.ts
```diff
@@ -3958,14 +3959,512 @@ g.test('subtractionMatrixInterval')
...
+          [3, 4],
+        ],
+        [
+          [10, 20],
+          [30, 40],
+          [50, 60],
+        ],
+      ],
+      expected: [
+        [70, 100],
```
ben-clayton, line 3998: Applying the same rules from above, we'd need the vector length of the LHS row to be equal to the vector length of the RHS column, no?

ben-clayton (commented):
I might be missing something here, but I'm not quite wrapping my head around the shape of these matrices.

src/unittests/f32_interval.spec.ts
zoddicus: I think is the annoying column major thing that I keep on hooking myself on.

[[1, 2], [3, 4]] is actually the matrix
```
 ╭      ╮   
 │ 1, 3 │
 │ 2, 4 │
 ╰      ╯
 ```
 b/c the first array `[1, 2]` is the values for column 1, so the first entries in each row.
 
For matrix product in column-major form the restriction is matKxR * matCxK, the 'outer' dimensions have to match, not the 'inner' ones, which is opposite of what I am used to, since I learnt linear algebra in row-major form
 
 When working on this, I actually turn my head sideways for it to read 'correctly' for me

Comment (ben-clayton) [reply src/unittests/f32_interval.spec.ts:3981]: Argh, I see that the transpose of these numbers is probably what your code is doing (which given the data is column major, is correct), however the numbers you are using means either work.

Comment (ben-clayton) [reply src/unittests/f32_interval.spec.ts:3981]: Please can you use different numbers so that `dot([a, b], [e, g])` is not equal to `dot([a, c], [e, f])`?

This will save future Ben from going down the same path.

File: src/unittests/f32_interval.spec.ts (hunk 2 above)
Comment (ben-clayton) [reply src/unittests/f32_interval.spec.ts:3998]: Once transposed, this makes sense.

File: src/unittests/f32_interval.spec.ts (hunk 1 above)
Comment (zoddicus) [reply src/unittests/f32_interval.spec.ts]: Updated the tests to avoid these cases.

File: src/webgpu/shader/execution/expression/binary/f32_matrix_arithmetic.spec.ts [hunk 3]
```diff
@@ -170,6 +171,438 @@ export const d = makeCaseCache('binary/f32_matrix_arithmetic', {
       additionMatrixInterval
     );
   },
+  multiplication_2x2_2x2_const: () => {
```
ben-clayton, line 174: Blimey, do we know how many cases this is in total?

ben-clayton (approved):
Thanks!

src/webgpu/shader/execution/expression/binary/f32_matrix_arithmetic.spec.ts
zoddicus: Just checked, 17004, and that is using sparseMatrixRange.

ben-clayton: Ouch. We'll have to check just how long this adds to the CTS.

PR #2316, opened 2023-02-27
Put the same URL in all "Open" links

greggman:
I find that I use `runnow=1` often and then I use an open link to put a subcase in a new tab. In that new tab I want the same URL options.

Issue: none

<hr>

### DIFF
--- a/src/common/runtime/standalone.ts
+++ b/src/common/runtime/standalone.ts
@@ -24,6 +24,8 @@ window.onbeforeunload = () => {
   return promptBeforeReload ? false : undefined;
 };
 
+const kOpenTestLinkAltText = 'Open';
+
 // The possible options for the tests.
 interface StandaloneOptions {
   runnow: boolean;
@@ -435,7 +437,7 @@ function makeTreeNodeHeaderHTML(
     onChange(true);
   };
 
-  const href = `?${worker ? 'worker&' : ''}${debug ? 'debug&' : ''}q=${n.query.toString()}`;
+  const href = createSearchQuery([n.query.toString()]);
   if (onChange) {
     div.on('toggle', function (this) {
       onChange((this as HTMLDetailsElement).open);
@@ -458,8 +460,8 @@ function makeTreeNodeHeaderHTML(
   $('<a>')
     .addClass('nodelink')
     .attr('href', href)
-    .attr('alt', 'Open')
-    .attr('title', 'Open')
+    .attr('alt', kOpenTestLinkAltText)
+    .attr('title', kOpenTestLinkAltText)
     .appendTo(header);
   if ('testCreationStack' in n && n.testCreationStack) {
     $('<button>')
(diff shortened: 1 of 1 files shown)


src/common/runtime/standalone.ts
```diff
@@ -540,14 +542,16 @@ void (async () => {
...
-  const updateURLWithCurrentOptions = () => {
+  const updateURLsWithCurrentOptions = () => {
     const search = prepareParams(optionsToRecord(options));
     let url = `${window.location.origin}${window.location.pathname}`;
     // Add in q separately to avoid escaping punctuation marks.
     url += `?${search}${search ? '&' : ''}${qs.map(q => 'q=' + q).join('&')}`;
     window.history.replaceState(null, '', url.toString());
+    document
+      .querySelectorAll(`a[alt=${kOpenTestLinkAltText}]`)
+      .forEach(elem => ((elem as HTMLAnchorElement).search = search));
```
kainino0x, line 553: This seems to also lose the `q=` from the URL so now all of the Open buttons go to the same place.

There's a bit of code which explicitly preserves only `worker` and `debug`, I think the more useful thing would be to fix this?:
https://github.com/gpuweb/cts/blob/534e6bb95e5111911229870a26412fde335c6b07/src/common/runtime/standalone.ts#L438

IIRC the reason for this was that preserving "runnow" when you go *up* the tree is not always what you want. But I think it would be fine if we did.

greggman: > This seems to also lose the q= from the URL so now all of the Open buttons go to the same place

color me embarrassed. I can't believe I didn't notice that when I tested 😅

I'm not sure which is better. I pushed a fix which removes that line above, and fixes the code I'd added since that effectively does the same work but for all options.

I feel like all the options being preserved is the more common thing users would want. I know it's what I want which is why I added it 😜

I set "runnow=1" and I'm iterating, then I want to work on a sub test and pop one out via the link and then I have to go manually set runnow. Similarly for all of the other options. 

It seems more confusing and error prone if they don't propagate as I'd pop one out (cmd-click) and expect it to work exactly the same as it was before I popped it out (same GPU, to just run, etc...) but if it silently loses options then it will do something other than what I just saw it do as a sub test.

wdyt?

kainino0x: Preserving all the options is good with me. I was just wondering if it was worth updating the links dynamically or just having it use the options that were set when the page was loaded.

I was thinking it would be slow to update all the buttons (since there can be a ton of them) but it's not - which led me to realize we lazily generate those buttons when they first become expanded. Which means we still need to update the `const href` as well to make it work, otherwise when they get lazily generated they get generated without any of the options.

Hacked that in locally and it does seem to fix it.

kainino0x (approved):

kainino0x:
Thanks for the iterations!

PR #2331, opened 2023-03-01
Removes OOM errors in error scope tests since we can't cause OOM.

lokokung:
- With the maxBufferSize warning turned into a validation error, there isn't a deterministic way to cause an OOM error (without the test becoming a stress test) at the moment, so removing the error scope test paths for now.


austinEng:
why can't we cause OOM?

lokokung:
Ah, sorry the comment I had was lost in the commit message, so added it to the description above.

austinEng:
Oh I see - @kainino0x should probably review

maybe we can reach OOM another way? like by making many many HUGE textures? I think the max texture size (in bytes) is larger than the max buffer size

lokokung:
Added Kai as a reviewer also. I think the problem with doing something like allocating a massive texture is that it gets close to stress testing stuff and idk where we want to draw the line here. Alternatively, (I think I proposed this before also), we could consider some sort of backdoor to the APIInjectError that we already have in native?

Kangz:
What about creating a 8k*8k*256layer RGBA32float texture? That's `2 << (13 + 13 + 8 + 4) = 2 << 38` bytes which is 256GB? (it might be too low on some hardware but maybe it works?)

kainino0x:
That's not a bad idea. Previously we considered allocating many buffers until we run out of memory, but that's very stressful to the system. Attempting to allocate one 256GB resource shouldn't be stressful... hopefully (unless the hardware implements it as many separate allocations). I'd avoid miplevels because those are more likely to be separate allocations.

Also has the bonus of poking a corner case which could reveal other interesting issues.

lokokung:
Hmm... I'm not opposed, but I'm hesitant if we are not sure if it'll work on all platforms and create triaging work to figure out if this test will work or not. Especially since these test are only trying to test error scopes, not other functionalities. The other thought that I had, which is why I asked about whether an OOM error should be surfaced when we hit https://www.w3.org/TR/webgpu/#abstract-opdef-initialize-an-active-buffer-mapping, was to make it so that after we throw the RangeError, we can continue executing as if it was just an OOM error (and returning an error buffer). Then we would have a (more) deterministic way to cause an OOM, but I'm actually not too familiar with JS standards for thrown errors and whether users expect to be able to continue executing or if it's a deal-breaker immediately.

src/webgpu/capability_info.ts
```diff
@@ -77,7 +77,7 @@ export const kErrorScopeFilterInfo: {
 };
 /** List of all GPUErrorFilter values. */
 export const kErrorScopeFilters = keysOf(kErrorScopeFilterInfo);
-export const kGeneratableErrorScopeFilters = kErrorScopeFilters.filter(e => e !== 'internal');
+export const kGeneratableErrorScopeFilters = kErrorScopeFilters.filter(e => e === 'validation');
```
kainino0x, line 80: This filter trivially results in just one element, so instead I think we should add a column in the info table just above for whether it's generatable. I'll just go ahead and do this.

kainino0x (commented):
I think we should try it, as long as it works locally on some machine, because it's a pretty useful test to have. However let's go ahead and land this so if we need to revert we can do a clean revert.

kainino0x (approved):

kainino0x:
> was to make it so that after we throw the RangeError, we can continue executing as if it was just an OOM error (and returning an error buffer)

If an exception is thrown, no value is returned, so there's no GPUBuffer at all (error or not).

PR #2291, opened 2023-02-19
Rename to GPUShaderModule.getCompilationInfo()

takahirox:
To follow the upcoming WebGPU spec change

https://github.com/gpuweb/gpuweb/pull/3805#issuecomment-1433931903

Requires
* https://github.com/gpuweb/gpuweb/pull/3805 needs to be merged
* [@webgpu/types](https://github.com/gpuweb/types) needs to updated with the above change

so this PR is still draft. I will switch to ready for review when the aboves are updated.

Issue: #2290

<hr>

**Requirements for PR author:**

- [x] All missing test coverage is tracked with "TODO" or `.unimplemented()`.
- [x] New helpers are `/** documented */` and new helper files are found in `helper_index.txt`.
- [x] Test behaves as expected in a WebGPU implementation. (If not passing, explain above.)

**Requirements for [reviewer sign-off](https://github.com/gpuweb/cts/blob/main/docs/reviews.md):**

- [x] Tests are properly located in the test tree.
- [x] [Test descriptions](https://github.com/gpuweb/cts/blob/main/docs/intro/plans.md) allow a reader to "read only the test plans and evaluate coverage completeness", and accurately reflect the test code.
- [x] Tests provide complete coverage (including validation control cases). **Missing coverage MUST be covered by TODOs.**
- [x] Helpers and types promote readability and maintainability.

When landing this PR, be sure to make any necessary issue status updates.

PR #2332, opened 2023-03-01
Depend on github.com/gpuweb/types instead of npm @webgpu/types

kainino0x:
This pattern unblocks CTS changes on needing to wait for someone to do a new package release.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,7 +23,7 @@
         "@types/pngjs": "^6.0.1",
         "@types/serve-index": "^1.9.1",
         "@typescript-eslint/parser": "^4.33.0",
-        "@webgpu/types": "0.1.27",
+        "@webgpu/types": "gpuweb/types#48dafa2124c767bcc4f84b0d1126bcc478143fad",
         "ansi-colors": "4.1.1",
         "babel-plugin-add-header-comment": "^1.0.3",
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -1262,9 +1262,10 @@
     },
     "node_modules/@webgpu/types": {
       "version": "0.1.27",
-      "resolved": "https://registry.npmjs.org/@webgpu/types/-/types-0.1.27.tgz",
+      "resolved": "git+ssh://git@github.com/gpuweb/types.git#48dafa2124c767bcc4f84b0d1126bcc478143fad",
```
kainino0x, line 1265: No matter what format I use in package.json, even explicitly specifying `https://`, this always ends up resolving to `git+ssh://`. This seems like it could cause issues if ssh is not set up on a machine, but it seems to pass the build check, so hopefully it's fine. We can revert back to the other format if not.

takahirox:
Thanks, this pattern resolves #2291

takahirox:
By the way, #2291 needs 
https://github.com/gpuweb/types/commit/77cd8d26f9c749dae5f2d105ad5fb2520755183e or 0.1.28, 
not https://github.com/gpuweb/types/commit/48dafa2124c767bcc4f84b0d1126bcc478143fad

PR #2334, opened 2023-03-01
Roll @webgpu/types 48dafa2 -> d1b649b

takahirox:
Needed for #2291

https://github.com/gpuweb/types/commit/d1b649b8baf1c092b07657c8e01bd5a81b3c6e77 is for upgrading to 0.1.28 so we could use `"@webgpu/types": "0.1.28"` but I didn't want to break `"@webgpu/types": "gpuweb/types#xxx..."` pattern.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,7 +23,7 @@
         "@types/pngjs": "^6.0.1",
         "@types/serve-index": "^1.9.1",
         "@typescript-eslint/parser": "^4.33.0",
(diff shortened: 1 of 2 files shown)


kainino0x:
This will need fixes to rename usages of compilationInfo to getCompilationInfo. We should get the Chromium change landed before this if it hasn't already.

kainino0x:
Oh I see you already have the renames in #2291. You can (should) do the dependency update in that PR so the build stays green.

takahirox:
Thanks for the advice. Merged this PR to #2291

PR #2339, opened 2023-03-02
Updates max*BufferBindingSize limit tests to adhere to maxBufferSize

lokokung:
Need buffer creation in these tests to adhere to maxBufferSize in order to enable deprecated features into validation errors.

<hr>


src/webgpu/api/validation/capability_checks/limits/maxStorageBufferBindingSize.spec.ts
```diff
@@ -104,18 +108,19 @@ g.test('createBindGroup,at_over')
...
-          })
-        );
-        const outOfMemoryError = await device.popErrorScope();
 
-        if (!outOfMemoryError) {
-          await t.expectValidationError(
-            () => {
+        await t.expectValidationError(
+          async () => {
+            device.pushErrorScope('out-of-memory');
```
greggman, line 114: I'm a little confused. I get why `kExtraLimits` was added. I don't get why the buffer allocation was moved inside `t.expectValidationError`. if the buffer fails to allocate then there will be no validation error even if one is expected and so the test will fail, vs allocating the buffer before `t.exceptValidationError` and then skipping the test if you couldn't make the buffer.

src/webgpu/api/validation/capability_checks/limits/maxStorageBufferBindingSize.spec.ts
```diff
@@ -146,7 +152,12 @@ g.test('validate,maxBufferSize')
     const { defaultLimit, maximumLimit } = t;
     const requestedLimit = getDeviceLimitToRequest(limitTest, defaultLimit, maximumLimit);
 
-    await t.testDeviceWithSpecificLimits(requestedLimit, 0, ({ device, actualLimit }) => {
-      t.expect(actualLimit <= device.limits.maxBufferSize);
-    });
+    await t.testDeviceWithSpecificLimits(
```
greggman, line 155: This test can be simplified to just this

```
g.test('validate,maxBufferSize')
  .desc(`Test that ${limit} <= maxBufferSize`)
  .fn(t => {
    const { limitTest } = t.params;
    const { adapter, defaultLimit, maximumLimit } = t;
    t.expect(defaultLimit <= t.getDefaultLimit('maxBufferSize'))
    t.expect(maximumLimit <= adapter.limits.maxBufferSize);
  });
```

Sorry, it was on my to-do list to cleanup as this test pre-exists some other changes in `limit_utils.ts`

same in the other file.

lokokung: Done!

src/webgpu/api/validation/capability_checks/limits/maxStorageBufferBindingSize.spec.ts
lokokung: So the issue is when `maxStorageBufferBindingSize` is close or equal to `maxBufferSize`. When that happens, and we try to do the `biggerBufferWithOffset` case, then the allocation will actually fail with a validation error, not an OOM error. The test framework, however, assumes that is a valid case because it doesn't know about the extra restriction from `maxBufferSize`.

The easiest way I was able to workaround this was to move the allocation inside and add the additional condition to the whether the overall function call should produce a validation error or not. LMK what you think.

lokokung: Adding the same handling in `maxUniformStorageBufferSize` test was just to be consistent, since in theory it could also happen, but in reality I don't think that limit will ever come close to `maxBufferSize`.

greggman: Thanks, I get it now

Should we capture the validation error outside of `t.expectValdationError`. The only thing this test is supposed to check is `createBindGroup` generating a validation error. It seems like errors for other reasons could be caught before and the actual test skipped if we can't run it.

Also, a comment about why we need to capture the validation error would be good.

I agree the uniform one won't hit the limit but I also agree it's good to refactor that code too

lokokung: Ah, that's a good point. Ok, I moved it back outside and instead of checking for a validation error, I just added a clause/comment to skip the test.

greggman (approved):

PR #2341, opened 2023-03-02
Fix writable binding aliasing in webgpu:api,validation,resource_usages,texture,in_pass_encoder

shrekshao:
Issue: https://github.com/gpuweb/cts/issues/2232

Related dawn issue: https://bugs.chromium.org/p/dawn/issues/detail?id=1642

<hr>

### DIFF
--- a/src/webgpu/api/validation/resource_usages/texture/in_pass_encoder.spec.ts
+++ b/src/webgpu/api/validation/resource_usages/texture/in_pass_encoder.spec.ts
@@ -234,15 +234,23 @@ g.test('subresources_and_binding_types_combination_for_color')
   .params(u =>
     u
       .combine('compute', [false, true])
-      .combineWithParams([
-        { _usageOK: true, type0: 'sampled-texture', type1: 'sampled-texture' },
-        { _usageOK: false, type0: 'sampled-texture', type1: 'writeonly-storage-texture' },
-        { _usageOK: false, type0: 'sampled-texture', type1: 'render-target' },
-        // Race condition upon multiple writable storage texture is valid.
-        { _usageOK: true, type0: 'writeonly-storage-texture', type1: 'writeonly-storage-texture' },
-        { _usageOK: false, type0: 'writeonly-storage-texture', type1: 'render-target' },
-        { _usageOK: false, type0: 'render-target', type1: 'render-target' },
-      ] as const)
+      .expandWithParams(
+        p =>
+          [
+            { _usageOK: true, type0: 'sampled-texture', type1: 'sampled-texture' },
+            { _usageOK: false, type0: 'sampled-texture', type1: 'writeonly-storage-texture' },
+            { _usageOK: false, type0: 'sampled-texture', type1: 'render-target' },
+            // Race condition upon multiple writable storage texture is valid.
+            // For p.compute === true, fails at pass.dispatch because aliasing exists.
+            {
+              _usageOK: !p.compute,
+              type0: 'writeonly-storage-texture',
+              type1: 'writeonly-storage-texture',
+            },
+            { _usageOK: false, type0: 'writeonly-storage-texture', type1: 'render-target' },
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/resource_usages/texture/in_pass_encoder.spec.ts
```diff
@@ -523,7 +523,13 @@ g.test('subresources_and_binding_types_combination_for_color')
...
     }
 
-    const success = _resourceSuccess || _usageOK;
+    let success: boolean;
+    if (type0 === type1 && type0 === 'writeonly-storage-texture' && compute) {
+      // _resourceSuccess === false causes writable binding aliasing at dispatch
+      success = _resourceSuccess && _usageOK;
+    } else {
+      success = _resourceSuccess || _usageOK;
+    }
```
kainino0x, line 532: I think this means _usageOK is wrong on line 242, can we change it (using `.expand`) so `_usageOK` on that line depends on the value of `compute`?

shrekshao: I think this test checks it is allowed to use the same subresources in multiple writable usages in the same pass. There is a similar dawn test [`src/dawn/tests/unittests/validation/ResourceUsageTrackingTests.cpp`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/src/dawn/tests/unittests/validation/ResourceUsageTrackingTests.cpp) I think is testing the same thing. 

so for 'writeonly-storage-texture' view0(bg0) and view1(bg1) having the same subresource,

this is valid (the render pipeline version only set bind group but don't call pass.draw)
```
pass.setBindGroup(0, bg0);
pass.dispatch(1);
pass.setBindGroup(0, bg1);
pass.dispatch(1);
```

this is invalid
```
pass.setBindGroup(0, bg0);
pass.setBindGroup(1, bg1);
pass.dispatch(1);
```


And in comparison `type0: 'sampled-texture', type1: 'writeonly-storage-texture'`
this is invalid
```
pass.setBindGroup(0, bg0);
pass.dispatch(1);
pass.setBindGroup(0, bg1);
pass.dispatch(1);
```


So maybe I should change the test further make the pipeline use one bindgroup, and setBindGroup for bg0 and bg1 at the same slot in the same pass, but at different dispatch/render?

kainino0x: You can't have usage conflicts between two different dispatches: each dispatch is considered a separate "usage scope" that include only the bind groups used by the current pipeline, and implementations are supposed to inject the necessary barriers/synchronization in between.

In render passes it's different, the whole render pass is one usage scope and it includes every setBindGroup call regardless of whether it's used at all.

Then on top of the usage scope rules (which doesn't care about binding ranges), there's the aliasing rule (which does). The aliasing rule only cares about aliasing within a single draw or dispatch calls, not about usage scopes or races between different calls.

The result of these two things (which I hadn't directly realized before) is that the carveout allowing races:
> NOTE: race condition of multiple writable storage buffer/texture usages in a single usage scope is allowed.

now only applies to render passes.

shrekshao: I see. Thanks for the explanation. IIUC:
* `type0 = type1 = 'writeonly-storage-texture' ` fails `compute` because of failing "aliasing rule" at `dispatch`
* `type0 = type1 = 'writeonly-storage-texture' ` passes `render` because of passing "usage scope rules"

so `_usageOk = is_compatible_usage(type0, type1) && pass_aliasing_rule(compute)`.

reflecting in code I can have `const usageOkRefined = _usageOk && !compute` and have a comment saying compute fails because of "aliasing rule"

kainino0x (approved):
Pretty sure this looks right to me.

PR #2346, opened 2023-03-02
Implement absolute error for `asin` intervals

### LINKED ISSUE #2345: Update `asin` accuracy to match spec
https://github.com/gpuweb/gpuweb/issues/3906 has weakened the requirements for asin to reflect real world implementation behaviours, the CTS needs to be updated to match.

zoddicus:
Fixes #2345

<hr>

### DIFF
--- a/src/unittests/f32_interval.spec.ts
+++ b/src/unittests/f32_interval.spec.ts
@@ -691,6 +691,7 @@ g.test('acosInterval')
       { input: -1/2, expected: [hexToF32(0x4005fa91), hexToF32(0x40061a94)] },  // ~2π/3
       { input: 0, expected: kAny },
       { input: 1/2, expected: [hexToF32(0x3f85fa8f), hexToF32(0x3f861a94)] },  // ~π/3
+      { input: 1, expected: kAny },
       { input: kValue.f32.positive.max, expected: kAny },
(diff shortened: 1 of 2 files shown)


src/unittests/f32_interval.spec.ts
```diff
@@ -764,16 +765,22 @@ g.test('asinInterval')
     // prettier-ignore
     [
       // Some of these are hard coded, since the error intervals are difficult to express in a closed human readable
-      // form due to the inherited nature of the errors.
+      // form due to the irrational nature of the values.
```
dneto0, line 768: nit:  "irrational" actually means something specific for numbers, and not what you mean here. Do you mean something like "non-sensical"?

dneto0 (approved):
Thanks!

src/unittests/f32_interval.spec.ts
zoddicus: This is a bit of legacy cruft from the original rewrite, I meant irrational in reference to the values were derived from Pi, since they were no longer inherited, but absolute errors

I am just going to change this to be 'due to the complexity of their derivation', which will cover all it being the span of the intervals from an absolute error and an inherited error, and around Pi.

Issue #2385, opened 2023-03-07
web_platform: canvas,getCurrentTexture


shrekshao:
Should be always the same JS object in the same frame, and never the same object in different frames.
https://github.com/gpuweb/gpuweb/issues/1388#issuecomment-771260087

[#2905](https://github.com/gpuweb/gpuweb/pull/2905)

shrekshao:
https://github.com/gpuweb/gpuweb/pull/3632

kainino0x:
cc #2326

greggman:
I think these tests already exist.

greggman:
stepping on each others toes

I had already written these tests in response to

https://github.com/gpuweb/cts/issues/2326

I hadn't uploaded them because of the questions in that issue.

shrekshao:
Oops

PR #2386, opened 2023-03-07
Get current texture expiry

shrekshao:
Issue: #2385

<hr>

This version of tests seem to fully pass on current Chrome Canary on Windows.
PTAL.

### DIFF
--- a/src/webgpu/web_platform/canvas/getCurrentTexture.spec.ts
+++ b/src/webgpu/web_platform/canvas/getCurrentTexture.spec.ts
@@ -274,16 +274,18 @@ Test automatic WebGPU canvas texture expiry on all canvas types with the followi
   - after previous frame update the rendering
   - before current frame update the rendering
   - in a microtask off the current frame task
-- getCurrentTexture returns a new texture object and the old texture object becomes invalid as soon as possible after HTML update the rendering.
+- getCurrentTexture returns a new texture object and the old texture object becomes invalid
+  as soon as possible after HTML update the rendering.
   `
   )
   .params(u =>
     u //
       .combine('canvasType', kAllCanvasTypes)
       .combine('prevFrameCallsite', ['requestPostAnimationFrame', 'requestAnimationFrame'] as const)
+      .combine('getCurrentTextureAgain', [true, false] as const)
   )
   .fn(t => {
-    const { canvasType, prevFrameCallsite } = t.params;
+    const { canvasType, prevFrameCallsite, getCurrentTextureAgain } = t.params;
     const ctx = t.initCanvasContext(t.params.canvasType);
     // Create a bindGroupLayout to test invalid texture view usage later.
     const bgl = t.device.createBindGroupLayout({
@@ -311,13 +313,17 @@ Test automatic WebGPU canvas texture expiry on all canvas types with the followi
 
       // Call getCurrentTexture immediately after the frame, the texture object should stay the same.
       queueMicrotask(() => {
-        t.expect(prevTexture === ctx.getCurrentTexture());
+        if (getCurrentTextureAgain) {
+          t.expect(prevTexture === ctx.getCurrentTexture());
+        }
       });
 
       // Call getCurrentTexture immediately after this frame updating the rendering.
       // It should return a new texture object.
       timeout(() => {
-        t.expect(prevTexture !== ctx.getCurrentTexture());
+        if (getCurrentTextureAgain) {
+          t.expect(prevTexture !== ctx.getCurrentTexture());
+        }
 
         // prevTexture expired and is invalid, but createView should still succeed.
         const prevTextureView = prevTexture.createView();
--- a/src/webgpu/web_platform/canvas/getCurrentTexture.spec.ts
+++ b/src/webgpu/web_platform/canvas/getCurrentTexture.spec.ts
@@ -292,7 +292,6 @@ Test automatic WebGPU canvas texture expiry on all canvas types with the followi
   as soon as possible after HTML update the rendering.
 
 TODO: test more canvas types, and ways to update the rendering
-- if on the same thread, expiry happens when the document updates its rendering (window "rPAF") OR transferToImageBitmap is called
 - if on a different thread, expiry happens when the worker updates its rendering (worker "rPAF") OR transferToImageBitmap is called
 - [draw, transferControlToOffscreen, then canvas is displayed] on either {main thread, or transferred to worker}
 - [draw, canvas is displayed, then transferControlToOffscreen] on either {main thread, or transferred to worker}
--- a/src/webgpu/web_platform/canvas/getCurrentTexture.spec.ts
+++ b/src/webgpu/web_platform/canvas/getCurrentTexture.spec.ts
@@ -4,6 +4,7 @@ Tests for GPUCanvasContext.getCurrentTexture.
 
 import { SkipTestCase } from '../../../common/framework/fixture.js';
 import { makeTestGroup } from '../../../common/framework/test_group.js';
+import { timeout } from '../../../common/util/timeout.js';
 import { assert, unreachable } from '../../../common/util/util.js';
 import { GPUTest } from '../../gpu_test.js';
 import { kAllCanvasTypes, createCanvas, CanvasType } from '../../util/create_elements.js';
@@ -300,7 +301,7 @@ Test automatic WebGPU canvas texture expiry on all canvas types with the followi
     // or after transferToImageBitmap (for offscreen canvas).
     function requestPostAnimationFrame(fn: () => void) {
       t.requestNewFrameOnCanvasType(canvasType, ctx, () => {
-        setTimeout(fn);
+        timeout(fn);
       });
     }
 
@@ -315,7 +316,7 @@ Test automatic WebGPU canvas texture expiry on all canvas types with the followi
 
       // Call getCurrentTexture immediately after this frame updating the rendering.
       // It should return a new texture object.
-      setTimeout(() => {
+      timeout(() => {
         t.expect(prevTexture !== ctx.getCurrentTexture());
 
         // prevTexture expired and is invalid, but createView should still succeed.
--- a/src/webgpu/web_platform/canvas/getCurrentTexture.spec.ts
+++ b/src/webgpu/web_platform/canvas/getCurrentTexture.spec.ts
@@ -22,6 +22,24 @@ class GPUContextTest extends GPUTest {
 
     return ctx;
   }
+
+  // Request a new "frame" based on the WebGPU canvas type.
+  requestNewFrameOnCanvasType(canvasType: CanvasType, ctx: GPUCanvasContext, fn: () => void) {
+    switch (canvasType) {
+      case 'onscreen':
+        requestAnimationFrame(fn);
+        break;
+      case 'offscreen': {
+        (ctx.canvas as OffscreenCanvas).transferToImageBitmap();
+        // The beginning of frameCheck runs immediately (in the same task), so this
+        // verifies the state has changed synchronously.
+        void fn();
+        break;
+      }
+      default:
+        unreachable();
+    }
+  }
 }
 
 export const g = makeTestGroup(GPUContextTest);
@@ -179,21 +197,7 @@ g.test('multiple_frames')
         prevTexture = currentTexture;
 
         if (frameCount++ < 5) {
-          // Which method will be used to begin a new "frame"?
-          switch (canvasType) {
-            case 'onscreen':
-              requestAnimationFrame(frameCheck);
-              break;
-            case 'offscreen': {
-              (ctx.canvas as OffscreenCanvas).transferToImageBitmap();
-              // The beginning of frameCheck runs immediately (in the same task), so this
-              // verifies the state has changed synchronously.
-              void frameCheck();
-              break;
-            }
-            default:
-              unreachable();
-          }
+          t.requestNewFrameOnCanvasType(canvasType, ctx, frameCheck);
         } else {
           resolve();
         }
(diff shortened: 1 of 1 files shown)


src/webgpu/web_platform/canvas/getCurrentTexture.spec.ts
```diff
@@ -260,3 +265,80 @@ g.test('resize')
...
+      const prevTexture = ctx.getCurrentTexture();
+
+      // Call getCurrentTexture immediately after the frame, the texture object should stay the same.
+      queueMicrotask(() => {
+        t.expect(prevTexture === ctx.getCurrentTexture());
+      });
+
+      // Call getCurrentTexture immediately after this frame updating the rendering.
+      // It should return a new texture object.
+      timeout(() => {
```
shrekshao, line 319: This is the earliest requirement we want for chrome implementation. The spec allows expiry happens immediately before rendering. https://gpuweb.github.io/gpuweb/#-automatic-expiry-task-source. So we may want to chagne this to a warn, or remove this check and replace with a less strict check.

src/webgpu/web_platform/canvas/getCurrentTexture.spec.ts
```diff
@@ -260,3 +265,80 @@ g.test('resize')
...
+        t.expect(prevTexture !== ctx.getCurrentTexture());
```
austinEng, line 320: I *think* we also should have a version that does *not* call ctx.getCurrentTexture() again.

That is, `prevTexture` should be expired and invalid to use, even if you don't call getCurrentTexture() again.

[queue an automatic expiry task](https://gpuweb.github.io/gpuweb/#abstract-opdef-queue-an-automatic-expiry-task) with device device and the following steps:
   - [commit the canvas texture](https://gpuweb.github.io/gpuweb/#abstract-opdef-commit-the-canvas-texture) for this.
     which says:
      - Update the rendering of context.[canvas](https://gpuweb.github.io/gpuweb/#dom-gpucanvascontext-canvas) with [a copy of the image contents](https://gpuweb.github.io/gpuweb/#abstract-opdef-get-a-copy-of-the-image-contents-of-a-context) of context.
      - If context.[[[currentTexture]]](https://gpuweb.github.io/gpuweb/#dom-gpucanvascontext-currenttexture-slot) is not null:
           - Call context.[[[currentTexture]]](https://gpuweb.github.io/gpuweb/#dom-gpucanvascontext-currenttexture-slot).[destroy()](https://gpuweb.github.io/gpuweb/#dom-gputexture-destroy) (without destroying context.[[[drawingBuffer]]](https://gpuweb.github.io/gpuweb/#dom-gpucanvascontext-drawingbuffer-slot)) to terminate write access to the image.
           - Set context.[[[currentTexture]]](https://gpuweb.github.io/gpuweb/#dom-gpucanvascontext-currenttexture-slot) to null.

src/webgpu/web_platform/canvas/getCurrentTexture.spec.ts
```diff
@@ -260,3 +265,80 @@ g.test('resize')
...
+
```
austinEng, line 321: Also, not totally sure - but @kainino0x should know.

We should have a tests that don't use rAF for:

```
timeout(() => {
  const t = ctx.getCurrentTexture();
  timeout(() => {
    // is using `t` valid???
  });
});
```

```
timeout(() => {
  const t = ctx.getCurrentTexture();
  timeout(() => {
    const t2 = ctx.getCurrentTexture();
     // does t === t2??
  });
});
```

src/webgpu/web_platform/canvas/getCurrentTexture.spec.ts
shrekshao: Added. Still seems to pass with Chrome

src/webgpu/web_platform/canvas/getCurrentTexture.spec.ts
greggman: My understanding of the spec is there are 3 cases

1. next task (eg, setTimeout, rAF)
2. rAF
3. update the rendering

3 is only covered if the "update the rendering" step is run and it is *NOT* run in response to rAF. It is run if something in the page changed. In the test I wrote I added (and removed next rAF) and 1x1 pixel div (display: fixed, left: 0, top:0, width: 1px, height 1px, opacity 50%) to try to cover that case. Of course it can only do this in a page, not in a worker.

kainino0x:
I can finally review this now that I've refreshed myself on how the spec works. It's impossible to keep clear in my head about which iteration was the one we decided on.

src/webgpu/web_platform/canvas/getCurrentTexture.spec.ts
```diff
@@ -260,3 +267,86 @@ g.test('resize')
...
+          visibility: GPUShaderStage.COMPUTE,
+          texture: {},
+        },
+      ],
+    });
+
+    // The fn is called immediately after previous frame updating the rendering.
+    // Polyfill by calling the callback by setTimeout, in the requestAnimationFrame callback (for onscreen canvas)
+    // or after transferToImageBitmap (for offscreen canvas).
+    function requestPostAnimationFrame(fn: () => void) {
```
kainino0x, line 306: nit: We should only use the name requestPostAnimationFrame for exactly that. A better name for this might be "runInNewCanvasFrame"

src/webgpu/web_platform/canvas/getCurrentTexture.spec.ts
```diff
@@ -260,3 +267,86 @@ g.test('resize')
...
+      t.requestNewFrameOnCanvasType(canvasType, ctx, () => {
+        timeout(fn);
```
kainino0x, line 308: The extra timeout only makes sense for rPAF (it's part of the rPAF implementation on top of rAF). After transferToImageBitmap, we're in the new frame *immediately* (without waiting for another task or anything).

src/webgpu/web_platform/canvas/getCurrentTexture.spec.ts
```diff
@@ -260,3 +267,86 @@ g.test('resize')
...
+  - after previous frame update the rendering
+  - before current frame update the rendering
+  - in a microtask off the current frame task
+- getCurrentTexture returns a new texture object and the old texture object becomes invalid
+  as soon as possible after HTML update the rendering.
+  `
+  )
+  .params(u =>
+    u //
+      .combine('canvasType', kAllCanvasTypes)
```
kainino0x, line 285: We should also add cases for transferControlToOffscreen. There's a number of things to test for that case, so probably should just write them down and do another PR. Technically there is already a TODO to improve kAllCanvasTypes but I don't think it will cover all of these.

Cases:

- if on the same thread, expiry happens when the document updates its rendering (window "rPAF") OR transferToImageBitmap is called
- if on a different thread, expiry happens when the worker updates its rendering (worker "rPAF") OR transferToImageBitmap is called
- [draw, transferControlToOffscreen, then canvas is displayed] on either {main thread, or transferred to worker}
- [draw, canvas is displayed, then transferControlToOffscreen] on either {main thread, or transferred to worker}
- reftests for the above 2 (what gets displayed when the canvas is displayed?)

Also possible things to test for canvases generally (canvas element or transferControlToOffscreen):
- canvas is added to DOM after being rendered
- canvas is already in DOM but becomes visible after being rendered

Unfortunately we can't technically test anything (except transferToImageBitmap) when the canvas is not in the DOM because expiry is never guaranteed to happen.

kainino0x (commented):
I confused myself a lot and wrote too much. I hope my comments aren't too confusing.

src/webgpu/web_platform/canvas/getCurrentTexture.spec.ts
```diff
@@ -260,3 +267,86 @@ g.test('resize')
     currentTexture = ctx.getCurrentTexture();
     t.expect(prevTexture === currentTexture);
   });
+
+g.test('expiry')
+  .desc(
+    `
+Test automatic WebGPU canvas texture expiry on all canvas types with the following requirements:
+- getCurrentTexture returns the same texture object within the same frame, throughout:
```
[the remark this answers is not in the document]
shrekshao: I see, so
- Task (getCurrentTexture retuns t1)
  - micro task (getCurrentTexture returns t1)
- Task (getCurrentTexture returns ??. We want chrome to return t2 by then)
- ...
- canvas get presented
- ...
- Task (getCurrentTexture returns t2)

src/webgpu/web_platform/canvas/getCurrentTexture.spec.ts
shrekshao: > when the canvas is not in the DOM
Is it that in cts we can only have these test in reftests?

src/webgpu/web_platform/canvas/getCurrentTexture.spec.ts
```diff
@@ -260,3 +267,86 @@ g.test('resize')
...
+  )
+  .params(u =>
+    u //
+      .combine('canvasType', kAllCanvasTypes)
+      .combine('prevFrameCallsite', ['requestPostAnimationFrame', 'requestAnimationFrame'] as const)
+      .combine('getCurrentTextureAgain', [true, false] as const)
+  )
+  .fn(t => {
+    const { canvasType, prevFrameCallsite, getCurrentTextureAgain } = t.params;
+    const ctx = t.initCanvasContext(t.params.canvasType);
```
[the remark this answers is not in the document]
shrekshao: Does this "visible" requires the canvas to be in DOM (appendChild)?

src/webgpu/web_platform/canvas/getCurrentTexture.spec.ts
greggman: reftests compare screenshots so they're only useful for things that make it to the page.

greggman: There are already tests for at least one of thseo

Also possible things to test for canvases generally (canvas element or transferControlToOffscreen):

> * canvas is added to DOM after being rendered
> * canvas is already in DOM but becomes visible after being rendered

It would be easy for me to add another case or 2.

src/webgpu/web_platform/canvas/getCurrentTexture.spec.ts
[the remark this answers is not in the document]
kainino0x: Correct (except if you do call getCurrentTexture in all 3 places in chrome you'll get a third texture t3 on the last one

src/webgpu/web_platform/canvas/getCurrentTexture.spec.ts
```diff
@@ -14,6 +14,15 @@ const kFormat = 'bgra8unorm';
...
   initCanvasContext(canvasType: CanvasType = 'onscreen'): GPUCanvasContext {
     const canvas = createCanvas(this, canvasType, 2, 2);
+    if (canvasType === 'onscreen') {
+      // To make sure onscreen canvas are visible
+      const onscreencanvas = canvas as HTMLCanvasElement;
+      onscreencanvas.style.position = 'fixed';
+      onscreencanvas.style.top = '0';
+      onscreencanvas.style.left = '0';
+      // Set it to transparent so that if multiple canvas are created, they are still visible.
+      onscreencanvas.style.opacity = '50%';
```
kainino0x, line 24: Canvas also needs to be added to the document.

shrekshao: Is there a preferred place to appendChild?
I do `document.body.appendChild(onscreencanvas);` now. And in developer tools element view seems canvas are added to the end of the body and that seems okay?

kainino0x: I think that's fine.

Also I forgot - we need to remove the canvas after the test is done.

src/webgpu/web_platform/canvas/getCurrentTexture.spec.ts
```diff
@@ -260,3 +277,103 @@ g.test('resize')
...
+      // Call getCurrentTexture immediately after the frame, the texture object should stay the same.
+      queueMicrotask(() => {
+        if (getCurrentTextureAgain) {
+          t.expect(prevTexture === ctx.getCurrentTexture());
+        }
+      });
+
+      // Call getCurrentTexture immediately after this frame updating the rendering.
+      // It could return a new texture object as early as the next task. [1]
+      timeout(() => {
```
kainino0x, line 352: In the `prevFrameCallsite="runInNewCanvasFrame"` case, I think this timeout() happens during the same frame as the first getCurrentTexture() (line 341). runInNewCanvasFrame() instead of timeout() here should work?

kainino0x: wait, I misunderstood something. Let me get back to you about this

src/webgpu/web_platform/canvas/getCurrentTexture.spec.ts
```diff
@@ -179,21 +210,7 @@ g.test('multiple_frames')
         prevTexture = currentTexture;
 
         if (frameCount++ < 5) {
-          // Which method will be used to begin a new "frame"?
-          switch (canvasType) {
-            case 'onscreen':
-              requestAnimationFrame(frameCheck);
```
kainino0x, line 185: On the first iteration I think this will call getCurrentTexture twice in the same frame. The first frameCheck isn't scheduled by rAF, so the rAF called from it may be run at the end of the same frame.

Fix would I think be to initially call frameCheck from requestAnimationFrame on line L202/R219 below.

src/webgpu/web_platform/canvas/getCurrentTexture.spec.ts
```diff
@@ -4,24 +4,55 @@ Tests for GPUCanvasContext.getCurrentTexture.
...
-      format: 'rgba8unorm',
+      format: kFormat,
       usage: GPUTextureUsage.RENDER_ATTACHMENT | GPUTextureUsage.COPY_SRC,
     });
 
     return ctx;
   }
+
+  // Request a new "frame" based on the WebGPU canvas type.
+  requestNewFrameOnCanvasType(canvasType: CanvasType, ctx: GPUCanvasContext, fn: () => void) {
```
kainino0x, line 40: I don't think this abstracts well over the two places we're using it right now [2]. One wants to do a frame loop so rAF makes sense, the other wants to run code at the beginning of the frame which is rPAF. I think I'd move the code back to those 2 places so we can modify them accordingly.

src/webgpu/web_platform/canvas/getCurrentTexture.spec.ts
kainino0x: I think this could be done with `this.trackForCleanup({ close() { document.body.removeChild(onscreencanvas); } });`

src/webgpu/web_platform/canvas/getCurrentTexture.spec.ts
```diff
@@ -179,21 +210,7 @@ g.test('multiple_frames')
...
         if (frameCount++ < 5) {
-          // Which method will be used to begin a new "frame"?
-          switch (canvasType) {
-            case 'onscreen':
-              requestAnimationFrame(frameCheck);
-              break;
-            case 'offscreen': {
-              (ctx.canvas as OffscreenCanvas).transferToImageBitmap();
-              // The beginning of frameCheck runs immediately (in the same task), so this
-              // verifies the state has changed synchronously.
```
kainino0x, line 190: nit: frameCheck is not async anymore, so this comment should be updated

src/webgpu/web_platform/canvas/getCurrentTexture.spec.ts
```diff
@@ -179,21 +210,7 @@ g.test('multiple_frames')
...
-              void frameCheck();
-              break;
-            }
-            default:
-              unreachable();
-          }
+          t.requestNewFrameOnCanvasType(canvasType, ctx, frameCheck);
```
kainino0x, line 213: [2] I'd remove this abstraction and restore the old code, but with a few changes

src/webgpu/web_platform/canvas/getCurrentTexture.spec.ts
```diff
@@ -260,3 +277,103 @@ g.test('resize')
...
+          texture: {},
+        },
+      ],
+    });
+
+    // The fn is called immediately after previous frame updating the rendering.
+    // Polyfill by calling the callback by setTimeout, in the requestAnimationFrame callback (for onscreen canvas)
+    // or after transferToImageBitmap (for offscreen canvas).
+    function runInNewCanvasFrame(fn: () => void) {
+      t.requestNewFrameOnCanvasType(canvasType, ctx, () => {
```
kainino0x, line 329: [2] I'd remove this abstraction and call requestPostAnimationFrame / transferToImageBitmap directly here.

where we implement requestPostAnimationFrame as a standalone thing (instead of having the rAF and timeout in different places in the code)

src/webgpu/web_platform/canvas/getCurrentTexture.spec.ts
kainino0x: OK I understand, this problem is the TODO about making it a warning. Let's not worry about the warning anymore since we will write a chrome-specific test, and just write the spec-compliant test.

So replacing this timeout() with runInNewCanvasFrame() seems right.

shrekshao: Turns out changing this `timeout` to `runInNewCanvasFrame`, makes that when canvas is offscreen, the new frame fn is called before the `queueMicroTask` fn, because offscreen canvas new frame method `transferToImageBitmap` returns immediately.

I can think of putting this `runInNewCanvasFrame` inside the fn of `queueMicroTask` to pass the test. Do you think this still keeps the semantic we want?

kainino0x: That's probably fine. Theoretically we would also want to test transferToImageBitmap without any task boundaries in between but that could be another test, or not at all. We might need another test anyway if we want to test things like transferToImageBitmap interaction with presentation of a transferControlToOffscreen'd canvas.

kainino0x (approved):

src/webgpu/web_platform/canvas/getCurrentTexture.spec.ts
```diff
@@ -260,3 +279,106 @@ g.test('resize')
...
+Test automatic WebGPU canvas texture expiry on all canvas types with the following requirements:
+- getCurrentTexture returns the same texture object until the next task:
+  - after previous frame update the rendering
+  - before current frame update the rendering
+  - in a microtask off the current frame task
+- getCurrentTexture returns a new texture object and the old texture object becomes invalid
+  as soon as possible after HTML update the rendering.
+
+TODO: test more canvas types, and ways to update the rendering
+- if on the same thread, expiry happens when the document updates its rendering (window "rPAF") OR transferToImageBitmap is called
```
kainino0x, line 295: This one is covered by this test
```suggestion
```

Issue #2277, opened 2023-02-15
Test restrictions on limits that can be exposed by adapters


kainino0x:
- [ ] `maxBindingsPerBindGroup`: "The supported limit for an adapter must be ≥ (max bindings per shader stage × max shader stages per pipeline) for the same adapter."
- [x] `minUniformBufferOffsetAlignment`: "Implementations must not support values less than 32 bytes."
- [x] `minStorageBufferOffsetAlignment`: "Implementations must not support values less than 32 bytes."
- [x] `maxBufferSize`: "The supported limit for an adapter must be ≥ maxUniformBufferBindingSize and ≥ maxStorageBufferBindingSize."
- [ ] `maxFragmentCombinedOutputResources`: https://github.com/gpuweb/gpuweb/pull/3829
- [x] [EDIT] all alignment limits: "May only be set to powers of 2 which are ≤ the default. Values which are not powers of 2 are invalid. Higher powers of 2 are clamped to the default."
- [ ] [EDIT] gpuweb/gpuweb#3888

I'm not sure if we have tests for any of these yet, but probably not.

kainino0x:
cc @greggman

greggman:
> * maxBufferSize: "The supported limit for an adapter must be ≥ maxUniformBufferBindingSize and ≥ maxStorageBufferBindingSize."

This one is tested the opposite way, the checks are `maxUniformBufferBindingSize <= maxBufferSize` and `maxStorageBufferBindingSize <= maxBufferSize`

I haven't put up the PR yet.

> * minUniformBufferOffsetAlignment: "Implementations must not support values less than 32 bytes."
> * minStorageBufferOffsetAlignment: "Implementations must not support values less than 32 bytes."

Added to the same PR

> * maxBindingsPerBindGroup: "The supported limit for an adapter must be ≥ (max bindings per shader stage × max shader stages per pipeline) for the same adapter."

I don't understand this one. the limit is "per bind group" which is less than stage per pipeline and less then binding per stage

kainino0x:
Forgot https://github.com/gpuweb/gpuweb/pull/3829, added above

kainino0x:
Also forgot to mention all `alignment`-class limits have extra restrictions, added above

greggman:
>all alignment limits: "May only be set to powers of 2 which are ≤ the default. Values which are not powers of 2 are invalid. Higher powers of 2 are clamped to the default."

[...]

PR #2396, opened 2023-03-09
Validate maxFragmentCombinedOutputResources

greggman:
Issue: #2277 #2195 

<hr>

### DIFF
--- a/src/webgpu/api/validation/capability_checks/limits/maxBindGroups.spec.ts
+++ b/src/webgpu/api/validation/capability_checks/limits/maxBindGroups.spec.ts
@@ -38,7 +38,7 @@ g.test('createPipelineLayout,at_over')
     );
   });
 
-g.test('createPipeline,async,at_over')
+g.test('createPipeline,at_over')
   .desc(
     `Test using createRenderPipeline(Async) and createComputePipeline(Async) at and over ${limit} limit`
   )
--- a/src/webgpu/api/validation/capability_checks/limits/maxBindGroups.spec.ts
+++ b/src/webgpu/api/validation/capability_checks/limits/maxBindGroups.spec.ts
@@ -2,7 +2,6 @@ import { range } from '../../../../../common/util/util.js';
 
 import {
   kCreatePipelineTypes,
-  kCreatePipelineAsyncTypes,
   kEncoderTypes,
   kMaximumLimitBaseParams,
   makeLimitTestGroup,
@@ -39,11 +38,17 @@ g.test('createPipelineLayout,at_over')
     );
   });
 
-g.test('createPipeline,at_over')
-  .desc(`Test using createRenderPipeline and createComputePipeline at and over ${limit} limit`)
-  .params(kMaximumLimitBaseParams.combine('createPipelineType', kCreatePipelineTypes))
+g.test('createPipeline,async,at_over')
+  .desc(
+    `Test using createRenderPipeline(Async) and createComputePipeline(Async) at and over ${limit} limit`
+  )
+  .params(
+    kMaximumLimitBaseParams
+      .combine('createPipelineType', kCreatePipelineTypes)
+      .combine('async', [false, true] as const)
+  )
   .fn(async t => {
-    const { limitTest, testValueName, createPipelineType } = t.params;
+    const { limitTest, testValueName, createPipelineType, async } = t.params;
 
     await t.testDeviceWithRequestedMaximumLimits(
       limitTest,
@@ -54,32 +59,7 @@ g.test('createPipeline,at_over')
         const code = t.getGroupIndexWGSLForPipelineType(createPipelineType, lastIndex);
         const module = device.createShaderModule({ code });
 
-        await t.expectValidationError(() => {
-          t.createPipeline(createPipelineType, module);
-        }, shouldError);
-      }
-    );
-  });
-
-g.test('createPipelineAsync,at_over')
-  .desc(
-    `Test using createRenderPipelineAsync and createComputePipelineAsync at and over ${limit} limit`
-  )
-  .params(kMaximumLimitBaseParams.combine('createPipelineAsyncType', kCreatePipelineAsyncTypes))
-  .fn(async t => {
-    const { limitTest, testValueName, createPipelineAsyncType } = t.params;
-
-    await t.testDeviceWithRequestedMaximumLimits(
-      limitTest,
-      testValueName,
-      async ({ device, testValue, shouldError }) => {
-        const lastIndex = testValue - 1;
-
-        const code = t.getGroupIndexWGSLForPipelineType(createPipelineAsyncType, lastIndex);
-        const module = device.createShaderModule({ code });
-
-        const promise = t.createPipelineAsync(createPipelineAsyncType, module);
-        await t.shouldRejectConditionally('GPUPipelineError', promise, shouldError);
+        await t.testCreatePipeline(createPipelineType, async, module, shouldError);
       }
     );
   });
--- a/src/webgpu/api/validation/capability_checks/limits/maxColorAttachmentBytesPerSample.spec.ts
+++ b/src/webgpu/api/validation/capability_checks/limits/maxColorAttachmentBytesPerSample.spec.ts
@@ -155,7 +155,7 @@ const kExtraLimits: LimitsRequest = {
 const limit = 'maxColorAttachmentBytesPerSample';
 export const { g, description } = makeLimitTestGroup(limit);
 
-g.test('createRenderPipeline,async,at_over')
+g.test('createRenderPipeline,at_over')
   .desc(`Test using at and over ${limit} limit in createRenderPipeline(Async)`)
   .params(
     kMaximumLimitBaseParams
--- a/src/webgpu/api/validation/capability_checks/limits/maxColorAttachmentBytesPerSample.spec.ts
+++ b/src/webgpu/api/validation/capability_checks/limits/maxColorAttachmentBytesPerSample.spec.ts
@@ -4,7 +4,7 @@ import { align } from '../../../../util/math.js';
 
 import {
   kMaximumLimitBaseParams,
-  kLimitModes,
+  LimitsRequest,
   LimitTestsImpl,
   makeLimitTestGroup,
 } from './limit_utils.js';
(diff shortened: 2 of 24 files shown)


src/webgpu/api/validation/capability_checks/limits/maxColorAttachmentBytesPerSample.spec.ts
```diff
@@ -184,26 +195,19 @@ g.test('createRenderPipeline,at_over')
           code
         );
       },
-      { maxColorAttachments: maxColorAttachmentsLimitMode }
+      kExtraLimits
     );
   });
 
 g.test('createRenderPipelineAsync,at_over')
```
lokokung, line 202: This test is almost identical to the one above aside from the final expectation check. We might be able to coalesce them into one and have an addition param `.combine('async', [true, false])`.

src/webgpu/api/validation/capability_checks/limits/maxStorageTexturesPerShaderStage.spec.ts
```diff
@@ -130,6 +135,13 @@ g.test('createPipeline,at_over')
       limitTest,
```
lokokung, line 135: Same comment as above with these tests w.r.t using a param for async and not async. But feel free to do it in another change if it gets too messy in this one or if there are other similar cases in other limit tests.

src/webgpu/api/validation/capability_checks/limits/maxFragmentCombinedOutputResources.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+
+const kTextureStorageTypes = [
+  { type: 'texture_storage_1d', coords: '0' },
+  { type: 'texture_storage_2d', coords: 'vec2u(0)' },
+  { type: 'texture_storage_2d_array', coords: 'vec2u(0), 0' },
+  { type: 'texture_storage_3d', coords: 'vec3u(0)' },
+];
+
+function getPipelineDescriptor(device: GPUDevice, testValue: number) {
+  const numStorageTextures = (testValue / 2) | 0;
```
lokokung, line 22: So for the limits here, would be nice to fill it with both storage buffers and textures, leaving the rest for the color attachments. I wrote similar tests in `RenderPipelineValidationTests.cpp` here https://dawn-review.googlesource.com/c/dawn/+/122461 for reference. I think the `kExtraLimits` also probably wants to reference `maxStorageBuffersPerShaderStage` and `maxStorageTexturesPerShaderStage` as those limits are also relevant when trying to fill out the resource limit.

lokokung (changes requested):

src/webgpu/api/validation/capability_checks/limits/maxColorAttachmentBytesPerSample.spec.ts
```diff
@@ -4,7 +4,7 @@ import { align } from '../../../../util/math.js';
 
 import {
   kMaximumLimitBaseParams,
-  kLimitModes,
+  LimitsRequest,
```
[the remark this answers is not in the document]
greggman: I feel like it's more accurate the way it is.

1. First it takes the bytesPerSample so far and adjusts it to the the alignment requirements for the next type.
2. Then it checks, given the aligned bytesPerSample, is there enough room for the next target.

The suggested way to me that logic is lost. It's instead "newBytesPerSize = magic math" if newBytesPerSize > testValue exit.

It might be 6 of 1, half dozen of another but it's not clear to me changing it makes it more understandable and not less.

src/webgpu/api/validation/capability_checks/limits/maxFragmentCombinedOutputResources.spec.ts
greggman: done

src/webgpu/api/validation/capability_checks/limits/maxStorageTexturesPerShaderStage.spec.ts
greggman: it was pretty messy but it was a good to cleanup

src/webgpu/api/validation/capability_checks/limits/maxColorAttachmentBytesPerSample.spec.ts
greggman: done

src/webgpu/api/validation/capability_checks/limits/maxColorAttachmentBytesPerSample.spec.ts
```diff
@@ -141,25 +147,23 @@ function createTextures(t: LimitTestsImpl, targets: GPUColorTargetState[]) {
...
       .combine('sampleCount', kTextureSampleCounts)
       .combine('interleaveFormat', kInterleaveFormats)
   )
   .fn(async t => {
-    const {
-      limitTest,
-      testValueName,
-      maxColorAttachmentsLimitMode,
-      sampleCount,
-      interleaveFormat,
```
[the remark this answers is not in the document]
lokokung: Hmm, fair enough that the other tests are not testing the max limit, just the default. I guess the thing I had against having it this way was that it makes it really hard to understand exactly that we are testing the alignment. In the linked tests, I made sure to use the same set of formats but ordered differently to clearly show that the alignment is under test, but just interleaving a format doesn't necessarily have the same guarantees.

Either way, fine to leave as is for now. We can always update them as necessary.

src/webgpu/api/validation/capability_checks/limits/maxColorAttachmentBytesPerSample.spec.ts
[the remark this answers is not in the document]
lokokung: Hmm fair enough. I guess I came from the spec knowing the exact algorithm that should be applied here (https://www.w3.org/TR/webgpu/#abstract-opdef-calculating-color-attachment-bytes-per-sample) so the magic math part wasn't so much magic, but just a given operation/calculation. I think either leaving it as is, or having the magic math part with a comment linking to the spec is fine.

src/webgpu/api/validation/capability_checks/limits/maxBindGroups.spec.ts
```diff
@@ -39,47 +38,28 @@ g.test('createPipelineLayout,at_over')
...
-
-        await t.expectValidationError(() => {
-          t.createPipeline(createPipelineType, module);
-        }, shouldError);
-      }
-    );
-  });
-
-g.test('createPipelineAsync,at_over')
+g.test('createPipeline,async,at_over')
```
lokokung, line 41: nit. Maybe we don't even need the word `async` here (and in the other tests). I think that the parameter makes it pretty obvious?

src/webgpu/api/validation/capability_checks/limits/maxColorAttachmentBytesPerSample.spec.ts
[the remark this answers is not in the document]
greggman: I like that reasoning. I'll check it to match the spec algo

src/webgpu/api/validation/capability_checks/limits/maxFragmentCombinedOutputResources.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+          code,
+          numStorageBuffers,
+          numStorageTextures,
+        } = getPipelineDescriptor(device, testValue);
+        const targets = pipelineDescriptor.fragment?.targets as GPUColorTargetState[];
+        const bytesPerSample = computeBytesPerSample(targets);
+        if (
+          targets.length > device.limits.maxColorAttachments ||
+          bytesPerSample > device.limits.maxColorAttachmentBytesPerSample ||
+          numStorageBuffers > device.limits.maxStorageBuffersPerShaderStage ||
```
lokokung, line 114: Ah... because I know that in the Dawn backend, `maxStorage[Buffer|Texture]PerShaderStage` are both clamped to 8, but `maxFragmentCombinedOutputResources` is computed as sum of `maxStorage[Buffer|Texture]PerShaderStage` and `maxColorAttachment` pre-clamping, I think that this condition might happen more often than ideal... WDYT about clamping the number of storageBuffers and storageTextures to their max limit also in case of skipping? Then worse case scenario we are still testing the effective (allowed) maximum of `maxFragmentCombinedOutputResources`?

src/webgpu/api/validation/capability_checks/limits/maxBindGroups.spec.ts
greggman: Does it? If I saw a test called `createPipeline` would assume it does not test async.

lokokung: Ah, I see your point since I realized that the `async` param will be hidden in the subcases. What if we promoted the `async` param to come before `kMaximumLimitBaseParams` that way it can be in the explicit case name? Then in the UI viewer and in tests names we would see something like: `createPipeline,at_over:async=true;*` and `createPipeline,at_over:async=false;*`.

src/webgpu/api/validation/capability_checks/limits/maxFragmentCombinedOutputResources.spec.ts
greggman: done

src/webgpu/api/validation/capability_checks/limits/maxBindGroups.spec.ts
greggman: I could do that ... my point was it seems reasonable to search just the names. For example searching for `g.test\('[^']*createPipeline[^']*'`

Screenshot 2023-03-10 at 15 13 50

or `g.test\('[^']*create[\w]*Pipeline[^']*'`

Screenshot 2023-03-10 at 15 17 17

to find tests that test `createXXXPipelineAsync`. If I didn't see "async" I wouldn't know where to look.

Surprised that this isn't useful though as it didn't find any other tests :P

lokokung: Hmm yea, I guess I'm not too set on either way, but searching through the codebase I also see other cases:

https://github.com/gpuweb/cts/blob/main/src/webgpu/api/validation/compute_pipeline.spec.ts
https://github.com/gpuweb/cts/blob/main/src/webgpu/api/validation/render_pipeline/fragment_state.spec.ts

where we have a param named `isAsync` instead of `async`. And the test name themselves don't both with saying async in them. Maybe it is good enough to include a mention of it in the description?

greggman: done

lokokung (approved):

PR #2421, opened 2023-03-15
Fix render_pass,store_op_and_read_only test

greggman:
Fixes webgpu:api,validation,render_pass,storeOp:store_op_and_read_only:readonly=false

This PR only prevents the error. I didn't check out the spec changed, but given the current spec, it seems like in another PR this test should be re-written to test loadOp as well?

Issue: none

<hr>


lokokung (dismissed):
LGTM with a TODO comment to note that the test needs to be re-worked a bit.

greggman:
Ok, please review.

Note: the depthClearValue test fails on the first test

> webgpu:api,validation,render_pass,loadOp_and_storeOp:depthClearValue:depthClearValue="_undef_";depthLoadOp="clear" 

It passes unexpectedly.

src/webgpu/api/validation/render_pass/loadOp_and_storeOp.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+      hasStencil && !stencilReadOnly ? hasBothStencilOps : hasNeitherStencilOps;
+
+    const shouldError = !goodAspectCombo || !goodDepthCombo || !goodStencilCombo;
+
+    t.expectValidationError(() => {
+      encoder.finish();
+    }, shouldError);
+  });
+
+g.test('depthClearValue')
```
lokokung, line 108: So couple thoughts here, initially I saw this and thought that the test name doesn't really make sense with the name of the suite so I was looking around to see whether there was another more suitable suite for it. I then came across https://github.com/gpuweb/cts/blob/main/src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts#L885 which is a very similar test, and also https://github.com/gpuweb/cts/blob/main/src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts#L811. So I was thinking that (1) we should consolidate those tests so that we don't have duplicates, and that (2) maybe we can have a `depth_stencil_attachment` suite under `render_pass` and keep all these tests in there?

Sorry I didn't see those tests earlier! D:

greggman: agreed. done

src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts
```diff
@@ -808,78 +808,93 @@ g.test('depth_stencil_attachment,sample_counts_mismatch')
...
-      setStencilLoadStoreOp,
+      stencilLoadOp,
+      stencilStoreOp,
     } = t.params;
 
-    let isValid = true;
-    const info = kTextureFormatInfo[format];
-    if (info.depth && info.stencil) {
-      isValid &&= depthReadOnly === stencilReadOnly;
+    if (format === 'depth32float-stencil8' && !t.device.features.has('depth32float-stencil8')) {
```
lokokung, line 844: I think the general pattern we use for this kind of thing is:
```
t.selectDeviceOrSkipTestCase(info.feature);
```

src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts
```diff
@@ -808,78 +808,93 @@ g.test('depth_stencil_attachment,sample_counts_mismatch')
...
+g.test('depth_stencil_attachment,loadOp_storeOp_match_depthReadOnly_stencilReadOnly')
   .desc(
     `
   Test GPURenderPassDepthStencilAttachment Usage:
     - depthReadOnly and stencilReadOnly must match if the format is a combined depth-stencil format.
-    - depthLoadOp and depthStoreOp must be provided iff the format has a depth aspect and
-      depthReadOnly is not true.
-    - stencilLoadOp and stencilStoreOp must be provided iff the format has a stencil aspect and
-      stencilReadOnly is not true.
+    - depthLoadOp and depthStoreOp must be provided if the format has a depth aspect and
```
lokokung, line 816: I agree with the iff change, but I think it might just be clearer to have break these points up a bit (even if it is a little more verbose) like the spec, (applies for stencil part also) i.e:

- if the format has a depth aspect:
    - if depthReadOnly is true:
        -  depthLoadOp and depthStoreOp must be provided
    - else:
        - depthLoadOp and depthStoreOp must not be provided

src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts
```diff
@@ -893,6 +908,7 @@ g.test('depth_stencil_attachment,depth_clear_value')
     u
       .combine('depthLoadOp', ['load', 'clear', undefined] as const)
       .combineWithParams([
+        { depthClearValue: undefined },
```
lokokung, line 911: I think this could just be simplified to:
```
.combine('depthClearValue', [undefined, -1.0, 0.0, 0.5, 1.0, 1.5])
```

lokokung (approved):
Overall LGTM with some nits! Thanks Gregg!

src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts
greggman: Thanks. I was wondering why it wasn't that way in the first place :P   Done

src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts
greggman: done

src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts
greggman: done

lokokung: I think we should be able to remove this if clause with the change.

greggman: sorry I missed that. done

Issue #2416, opened 2023-03-15
Add AbstractFloat support to floating point testing framework


zoddicus:
This covers all of the refactoring of the existing code to support parameterization on floating point type, and related work to get that working. This does not cover actually enabling specific AbstractFloat tests, those are covered by other issues.

PR #2456, opened 2023-03-22
Use (number | IntervalBounds) for specifying unittests

### LINKED ISSUE #2254: `f32_inteval.spec.ts` tests can use `IntervalBounds | number` pattern to simplify test cases
This was originally noted in the review for the `transpose` tests, https://github.com/gpuweb/cts/pull/2249. This pattern for allowing points to be specified via numbers could be used to clean up other tests.

existing tests use `[1]` for a point F32Interval instead of `1`, because the type of IntervalBounds is `[number]  | [number, number]`. Utility functions like `toF32Interval`, `toF32Vector`, and `toF32Matrix` have been added, since this pattern was established,  for converting a variety of types down to F32Interval

Existing tests could be updated to allow passing in arrays of numbers when all of the values are points. This could be taken even further by having the various toF32Foo function support mixed number| IntervalBounds.

zoddicus:
This allows tests that only need a point value to be listed as `1` instead of `[1]`.

This helps readability, especially for tests where there are arrays of values, i.e. (number | IntervalBounds)[], since [[1], [2], ...], can be written as [1, 2, ...]

This require changing the signatures of toVector and toMatrix (and related functions) to support mixed arrays, i.e. (number | IntervalBounds)[] instead of number[] | IntervalBuilder[], which honestly is just more ergonomic, and the code already supported it.

Some tests that were calling the F32Interval constructor directly have been flipped to using `toInterval`, since they are not testing the constructor and that is the preferred pattern.

A bunch of documentation lines in the test file have been re-formatted to be 80 cols, which I think is easier to read, since this PR is all about readability/ergonomics anyway. I was not super consistent about line length when writing the tests originally.

Issue #2416
Fixes #2254

<hr>

### DIFF
--- a/src/unittests/f32_interval.spec.ts
+++ b/src/unittests/f32_interval.spec.ts
@@ -119,11 +119,13 @@ function minusOneULP(x: number): number {
  * @param expected the bounds to be adjusted
  * @param error error function to adjust the bounds via
  */
-function applyError(expected: IntervalBounds, error: (n: number) => number): IntervalBounds {
+function applyError(
+  expected: number | IntervalBounds,
+  error: (n: number) => number
+): IntervalBounds {
   if (expected !== kAny) {
-    const begin = expected[0];
-    const end = expected.length === 2 ? expected[1] : begin;
-    expected = [begin - error(begin), end + error(end)];
+    const interval = toF32Interval(expected);
+    expected = [interval.begin - error(interval.begin), interval.end + error(interval.end)];
   }
 
   return expected;
@@ -171,7 +173,7 @@ g.test('constructor')
   });
 
 interface ContainsNumberCase {
-  bounds: IntervalBounds;
+  bounds: number | IntervalBounds;
   value: number;
   expected: boolean;
 }
@@ -196,13 +198,13 @@ g.test('contains_number')
     { bounds: [-5, 10], value: -10, expected: false },
 
     // Point
-    { bounds: [0], value: 0, expected: true },
-    { bounds: [0], value: 10, expected: false },
-    { bounds: [0], value: -1000, expected: false },
-    { bounds: [10], value: 10, expected: true },
-    { bounds: [10], value: 0, expected: false },
-    { bounds: [10], value: -10, expected: false },
-    { bounds: [10], value: 11, expected: false },
+    { bounds: 0, value: 0, expected: true },
+    { bounds: 0, value: 10, expected: false },
+    { bounds: 0, value: -1000, expected: false },
+    { bounds: 10, value: 10, expected: true },
+    { bounds: 10, value: 0, expected: false },
+    { bounds: 10, value: -10, expected: false },
+    { bounds: 10, value: 11, expected: false },
 
     // Upper infinity
     { bounds: [0, kValue.f32.infinity.positive], value: kValue.f32.positive.min, expected: true },
@@ -281,7 +283,7 @@ g.test('contains_number')
     ]
   )
   .fn(t => {
-    const i = new F32Interval(...t.params.bounds);
+    const i = toF32Interval(t.params.bounds);
     const value = t.params.value;
     const expected = t.params.expected;
 
@@ -290,8 +292,8 @@ g.test('contains_number')
   });
 
 interface ContainsIntervalCase {
-  lhs: IntervalBounds;
-  rhs: IntervalBounds;
+  lhs: number | IntervalBounds;
+  rhs: number | IntervalBounds;
   expected: boolean;
 }
 
@@ -300,7 +302,7 @@ g.test('contains_interval')
     // prettier-ignore
     [
       // Common usage
-      { lhs: [-10, 10], rhs: [0], expected: true},
+      { lhs: [-10, 10], rhs: 0, expected: true},
       { lhs: [-10, 10], rhs: [-1, 0], expected: true},
       { lhs: [-10, 10], rhs: [0, 2], expected: true},
       { lhs: [-10, 10], rhs: [-1, 2], expected: true},
@@ -310,7 +312,7 @@ g.test('contains_interval')
       { lhs: [-10, 10], rhs: [-100, 10], expected: false},
 
       // Upper infinity
-      { lhs: [0, kValue.f32.infinity.positive], rhs: [0], expected: true},
+      { lhs: [0, kValue.f32.infinity.positive], rhs: 0, expected: true},
       { lhs: [0, kValue.f32.infinity.positive], rhs: [-1, 0], expected: false},
       { lhs: [0, kValue.f32.infinity.positive], rhs: [0, 1], expected: true},
       { lhs: [0, kValue.f32.infinity.positive], rhs: [0, kValue.f32.positive.max], expected: true},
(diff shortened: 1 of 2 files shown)


src/webgpu/util/f32_interval.ts
```diff
@@ -169,15 +169,15 @@ export type F32Vector =
   | [F32Interval, F32Interval, F32Interval, F32Interval];
 
 /** Coerce F32Interval[] to F32Vector if possible */
-function isF32Vector(v: number[] | IntervalBounds[] | F32Interval[] | F32Vector): v is F32Vector {
```
ben-clayton, line 172: I'm not sure if this worked as intended before.

`F32Vector` is a `2`, `3` or `4` element array of `F32Interval`.

`F32Interval[] | F32Vector` is essentially the same as just saying `F32Interval[]`, as the former is a superset of the latter.

So `isF32Vector` would return `true` when passing a `2`, `3` or `4` sized-array of `F32Interval`, even if the intention was not to treat it as a vector - is that intentional?

If not, I think you need to make `F32Vector` a true class to treat it as a distinct type. This might not be a bad idea as the number of nested `[` `]` in these tests are becoming rather hard to mentally map to their semantic meaning. A `vec(1, 2)` style helper would make it obvious that you are dealing with vectors, and would help with the RTTI problems.

If so, then I'd suggest just dropping the the `| F32Vector` part of the parameter - it's not doing anything, and is adding more confusion to the helper.

The changes to this method don't appear to address this issue.

src/webgpu/util/f32_interval.ts
```diff
@@ -169,15 +169,15 @@ export type F32Vector =
...
+function isF32Vector(v: (number | IntervalBounds | F32Interval)[] | F32Vector): v is F32Vector {
   if (v[0] instanceof F32Interval) {
     return v.length === 2 || v.length === 3 || v.length === 4;
   }
   return false;
 }
 
 /** @returns an F32Vector representation of an array fo F32Intervals if possible */
-export function toF32Vector(v: number[] | IntervalBounds[] | F32Interval[] | F32Vector): F32Vector {
+export function toF32Vector(v: (number | IntervalBounds | F32Interval)[] | F32Vector): F32Vector {
```
ben-clayton, line 180: So you're now saying that `v` can be an array of any *mixed* element type of `number`, `IntervalBounds`, or `F32Interval`.

So, you could call `toF32Vector` with `[1, [2], [2,3], F32Interval.any()]`.

Please can we have some unit tests to check that this can correctly cope with the mix of element types?

src/webgpu/util/f32_interval.ts
```diff
@@ -169,15 +169,15 @@ export type F32Vector =
   | [F32Interval, F32Interval, F32Interval, F32Interval];
 
 /** Coerce F32Interval[] to F32Vector if possible */
```
ben-clayton, line 171: Comment doesn't seem to match method.

src/webgpu/util/f32_interval.ts
```diff
@@ -169,15 +169,15 @@ export type F32Vector =
...
-function isF32Vector(v: number[] | IntervalBounds[] | F32Interval[] | F32Vector): v is F32Vector {
+function isF32Vector(v: (number | IntervalBounds | F32Interval)[] | F32Vector): v is F32Vector {
   if (v[0] instanceof F32Interval) {
```
ben-clayton, line 173: This now looks wrong.
`isF32Vector([F32Interval, number, number])` would return true while `isF32Vector([number, F32Interval, number])` would not.

Tests please.

ben-clayton (changes requested):

src/webgpu/util/f32_interval.ts
zoddicus: True being returned for F32Interval[] with size 2, 3, 4 was intentional. This specifically get used when using .maps. The return type of .map on a F32Vector is a an array, not a F32Vector, even through for all intents and purposes it will behave like a F32Vector. It get used like this `toF32Vector(vec.map(foo))` when running an operation component-wise over a vector.

I have been somewhat resistant to introducing a F32Vector class, since that would lose a lot of the 'free' array functionality, unless we reimplemented it. I have considered it previously, and will take another look at it while implementing the first pass of unittests.

zoddicus: I agree that the `| F32Vector` should be droppable

src/webgpu/util/f32_interval.ts
zoddicus: Tests and fixed.

src/webgpu/util/f32_interval.ts
zoddicus: Updated

src/webgpu/util/f32_interval.ts
zoddicus: That is correct, a vector of things that is valid to toF32Interval should be convertible in toF32Vector. I have added tests for this.

zoddicus:
PTAL

src/unittests/f32_interval.spec.ts
```diff
@@ -300,134 +304,990 @@ g.test('contains_interval')
...
-      // Double Intervals
-      { intervals: [[0, 1], [2, 5]], expected: [0, 5]},
-      { intervals: [[2, 5], [0, 1]], expected: [0, 5]},
-      { intervals: [[0, 2], [1, 5]], expected: [0, 5]},
-      { intervals: [[0, 5], [1, 2]], expected: [0, 5]},
-      { intervals: [[kValue.f32.infinity.negative, 0], [0, kValue.f32.infinity.positive]], expected: kAny},
+    // F32Interval
+    {
+      input: [
+        [toF32Interval(1), toF32Interval(2)],
```
ben-clayton, line 1128: Would be nice if some of these provided something other than a single `number` to `toF32Interval`

src/unittests/f32_interval.spec.ts
```diff
@@ -300,134 +304,990 @@ g.test('contains_interval')
...
+      expected: [
+        [1, 2],
+        [2, 3],
+        [3, 4],
+        [4, 5],
+      ],
+    },
+
+    // F32Interval
+    { input: [toF32Interval([1]), toF32Interval([2])], expected: [1, 2] },
```
ben-clayton, line 568: Would be nice if some of these provided something other than a single `number` to `toF32Interval`

src/unittests/f32_interval.spec.ts
```diff
@@ -300,134 +304,990 @@ g.test('contains_interval')
...
+        [1, 2],
+        [2, 3],
+        [3, 4],
+        [4, 5],
+      ],
+      expected: false,
+    },
+
+    // F32Interval, valid dimensions
+    { input: [toF32Interval([1]), toF32Interval([2])], expected: true },
```
ben-clayton, line 479: Would be nice if some of these provided something other than a single `number` to `toF32Interval`

src/unittests/f32_interval.spec.ts
```diff
@@ -300,134 +304,990 @@ g.test('contains_interval')
...
+        [[9], [10], [11], [12]],
+        [[13], [14], [15], [16]],
+      ],
+      expected: false,
+    },
+
+    // F32Interval, valid dimensions
+    {
+      input: [
+        [toF32Interval(1), toF32Interval(2)],
```
ben-clayton, line 753: Would be nice if some of these provided something other than a single `number` to `toF32Interval`

ben-clayton (approved):

PR #2457, opened 2023-03-22
Fixes remaining tests that hit deprecated paths.

lokokung:
- Make sure to use 'unfilterable-float' when multisampling.
- Make sure not to pass depth/stencil Load/Store ops in readonly mode.
- Also makes some of the tests under 'subresources,depth_stencil_attachment_and_bind_group' into subcases. This speeds up the tests considerably. Before each test when not hitting any caching took about 1s, but now we batch 8 cases together (running them in parallel) resulting still in about 1s runtime... So almost an 8x speed up! Also greatly increases load-speed of the page since we don't need to create ~300 entries, and just create ~40.

<hr>


greggman (approved):
lgtm

I took me a while to figure out why `unfiltered-float` is need but finally got it. 😅

PR #2461, opened 2023-03-24
Fix unexpected failure in api,validation,queue,copyToTexture,CopyExte…

shaoboyan:
…rnalImageToTexture:source_image,crossOrigin:*

In crossOrigin test case, we call getCanvasWithContent() API and assuming context is CanvasRenderingContext2D.

But for OffscreenCanvas, it returns OffscreenCanvasRenderingContext2D, which is different from CanvasRenderingContext2D.

This PR fix the unexpected failure by adding OffscreenCanvasRenderingContext2D into assert

Issue: #

<hr>

### DIFF
--- a/src/webgpu/api/validation/queue/copyToTexture/CopyExternalImageToTexture.spec.ts
+++ b/src/webgpu/api/validation/queue/copyToTexture/CopyExternalImageToTexture.spec.ts
@@ -157,7 +157,9 @@ class CopyExternalImageToTextureTest extends ValidationTest {
   ): HTMLCanvasElement | OffscreenCanvas {
     const canvas = createCanvas(this, canvasType, 1, 1);
     const ctx = canvas.getContext('2d');
-    assert(ctx instanceof CanvasRenderingContext2D);
+    assert(
+      ctx instanceof CanvasRenderingContext2D || ctx instanceof OffscreenCanvasRenderingContext2D
+    );
     ctx.drawImage(content, 0, 0);
 
     return canvas;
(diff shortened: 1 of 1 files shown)


Jiawei-Shao (approved):
Do you think if we should do as below:

```javascript
    switch (canvasType) {
      case 'onscreen':
        assert(ctx instanceof CanvasRenderingContext2D);
        break;
      case 'offscreen':
        assert(ctx instanceof OffscreenCanvasRenderingContext2D);
        break;
    }
```

src/webgpu/api/validation/queue/copyToTexture/CopyExternalImageToTexture.spec.ts
```diff
@@ -157,7 +157,14 @@ class CopyExternalImageToTextureTest extends ValidationTest {
...
     const ctx = canvas.getContext('2d');
-    assert(ctx instanceof CanvasRenderingContext2D);
+    switch (canvasType) {
+      case 'onscreen':
+        assert(ctx instanceof CanvasRenderingContext2D);
+        break;
+      case 'offscreen':
+        assert(ctx instanceof OffscreenCanvasRenderingContext2D);
+        break;
+    }
```
kainino0x, line 167: nit, but don't have to fix this: I think this assert is just here to make typescript happy, it would be fine to just `assert(ctx !== null);` or simply do `const ctx = canvas.getContext('2d')!;` since if it's null it'll crash on the next line anyway.

PR #2464, opened 2023-03-24
Move "device_destroyed_before_importing_video_frame" case to "api,validation,states,device_lost,destroy"

shaoboyan:
According to WebGPU spec, calling GPUDevice.destroy() results in "Destroys the device, preventing further operations on it. Outstanding asynchronous operations will fail."
(https://www.w3.org/TR/webgpu/#dom-gpudevice-destroy).

"device_destroyed_before_importing_video_frame" test case says it "Tests that imported GPUExternalTexture from destroyed device and submit
 it should result in an error".

It seems this test case violate the spec defination about destroyed GPUDevice.

This PR removed this test case.

Issue: #

<hr>


src/webgpu/api/validation/gpu_external_texture_expiration.spec.ts
```diff
@@ -322,42 +322,3 @@ g.test('import_from_different_video_frame')
...
-    let bindGroup: GPUBindGroup;
-    let externalTexture: GPUExternalTexture;
-    let source: HTMLVideoElement | VideoFrame;
-    await startPlayingAndWaitForVideo(videoElement, async () => {
-      source =
-        sourceType === 'VideoFrame'
-          ? await getVideoFrameFromVideoElement(t, videoElement)
-          : videoElement;
-
-      t.device.destroy();
```
kainino0x, line 350: If we keep this test, perhaps add something like:
`if (t.params.awaitAfterDestroy) await t.device.lost;`
here.

src/webgpu/api/validation/gpu_external_texture_expiration.spec.ts
```diff
@@ -322,42 +322,3 @@ g.test('import_from_different_video_frame')
       t.submitCommandBuffer(bindGroup, true);
     });
   });
-
-g.test('device_destroyed_before_importing_video_frame')
-  .desc(
-    `
-    Tests that imported GPUExternalTexture from destroyed device and submit
-    it should result in an error.
```
kainino0x, line 330: Agree this description looks wrong, but the test itself doesn't seem to actually be checking for an error. I think it's OK to test this behavior. Even though we can't assert anything, we can still detect if the browser crashes or something.

shaoboyan: Ok, I think I'd like to move this test case to a proper place here (https://github.com/gpuweb/cts/blob/904c74af6fee0530ce33434974aa0df6316fdef1/src/webgpu/api/validation/state/device_lost/destroy.spec.ts#L44)

kainino0x: That makes sense.

src/webgpu/api/validation/gpu_external_texture_expiration.spec.ts
shaoboyan: Move test case to api,validation,state,device_lost,destroy

shaoboyan:
@kainino0x would you mind to take another look at this? I think it is ready to be merged but requires your double check, thanks!

src/webgpu/api/validation/state/device_lost/destroy.spec.ts
```diff
@@ -500,6 +505,43 @@ Tests creating query sets on destroyed device.
...
+      source =
+        sourceType === 'VideoFrame'
+          ? await getVideoFrameFromVideoElement(t, videoElement)
+          : videoElement;
+
+      await t.executeAfterDestroy(() => {
+        t.device.importExternalTexture({
+          /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
+          source: source as any,
+        });
```
kainino0x, line 540: nit: It would probably be good to keep something that uses the texture, like a createBindGroup call

kainino0x (approved):
Apologies! GitHub doesn't send emails when a new commit is pushed.

src/webgpu/api/validation/state/device_lost/destroy.spec.ts
```diff
@@ -500,6 +505,43 @@ Tests creating query sets on destroyed device.
...
+      .combine('sourceType', ['VideoElement', 'VideoFrame'] as const)
+      .beginSubcases()
+      .combine('awaitLost', [true, false])
+  )
+  .fn(async t => {
+    const { awaitLost, sourceType } = t.params;
+
+    const videoElement = getVideoElement(t, 'four-colors-vp9-bt601.webm');
+    if (!('requestVideoFrameCallback' in videoElement)) {
+      t.skip('HTMLVideoElement.requestVideoFrameCallback is not supported');
```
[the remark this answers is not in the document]
shaoboyan: > The latest expiration semantics might not need it.

I think we could remove the requirement of rVFC here but not for gpu_external_texture_expiration.spec.ts. If we import with a not ready video element, we might get error (https://www.w3.org/TR/webgpu/#dom-gpudevice-importexternaltexture). Am I misunderstand something?

src/webgpu/api/validation/state/device_lost/destroy.spec.ts
shaoboyan: Done.

PR #2468, opened 2023-03-29
Refactor how floating point intervals are calculated

zoddicus:
Introduces a FPContext class that things like limits, constants, utilities, and interval generators exist in. This reorganizes the existing collection of free functions that all loosely depend on F32Interval into a tighter structure where they pull information about how to build intervals and what the limits for floating point numbers are from their context.

There is currently one concrete implementation of FPContext class for f32, there will be more in the future to support abstrate float and f16.

There is currently code in the generic FPContext class that is f32 specific, like magic numbers for precision. As things like abstract float and f16 are implemented I expect there to be refactorings to move these things around, either adding more constants, or moving specific interval generators down to specific floating point implementations.

All of the implementation code has been moved to floating_point.ts

f32_interval.ts contains pass-through shims that allow existing code and tests to use the old API for generating intervals. Future patches will work through migrating the rest of the CTS off of the old API. Currently testing for the interval generators runs through the shim, so that the codepath being used in CTS is under test, and it reduces the diff for this patch.

Most of the existing code outside of the interval generators doesn't need to change at this point, expect for specific locations like serialization/deserialization, where FPInterval (formally F32Interval) is used explicitly. This is due to FPInterval being an actual class vs type/interface, so it it is non-trivial to alias it in the shim. So for the few places where the refactoring needed to leak out it has, instead of going through heroeics for conceptual purity.

Issue #2416

<hr>

### DIFF
--- a/src/webgpu/util/floating_point.ts
+++ b/src/webgpu/util/floating_point.ts
@@ -1,0 +1,3090 @@
+import { assert, unreachable } from '../../common/util/util.js';
+import { Float16Array } from '../../external/petamoriken/float16/float16.js';
+
+import { kValue } from './constants.js';
+import { f32, reinterpretF32AsU32, reinterpretU32AsF32, Scalar } from './conversion.js';
+import {
+  calculatePermutations,
+  cartesianProduct,
+  correctlyRoundedF16,
+  correctlyRoundedF32,
+  flatten2DArray,
+  FlushMode,
+  flushSubnormalNumberF32,
+  isFiniteF16,
+  isFiniteF32,
+  isSubnormalNumberF16,
+  isSubnormalNumberF32,
+  map2DArray,
+  oneULPF32,
+  unflatten2DArray,
+} from './math.js';
+
+/** Indicate the kind of WGSL floating point numbers being operated on */
+export type FPKind = 'f32';
+
+// Containers
+
+/**
+ * Representation of bounds for an interval as an array with either one or two
+ * elements. Single element indicates that the interval is a single point. For
+ * two elements, the first is the lower bound of the interval and the second is
+ * the upper bound.
+ */
+export type IntervalBounds = [number] | [number, number];
+
+/** Represents a closed interval of floating point numbers */
+export class FPInterval {
+  public readonly kind: FPKind;
+  public readonly begin: number;
+  public readonly end: number;
+  // context is JS private, so that objectEquals does not see it when comparing
+  // FPIntervals, since it may not have been initialized yet, but that should
+  // impact equality.
+  #context?: FPContext;
+
+  /**
+   * Constructor
+   *
+   * `FPContext.toInterval` is the preferred way to create FPIntervals
+   *
+   * @param kind the floating point number type this is an interval for
+   * @param bounds beginning and end of the interval
+   */
+  public constructor(kind: FPKind, ...bounds: IntervalBounds) {
+    this.kind = kind;
+
+    const [begin, end] = bounds.length === 2 ? bounds : [bounds[0], bounds[0]];
+    assert(!Number.isNaN(begin) && !Number.isNaN(end), `bounds need to be non-NaN`);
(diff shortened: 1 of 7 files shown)


src/webgpu/util/floating_point.ts
```diff
@@ -0,0 +1,3090 @@
...
+  /** @returns the floating point context for this interval */
+  public context(): FPContext {
+    // context is not fetched and stored at creation time, because there is a
+    // potential dependency cycle between kFPContext and interval on startup.
+    if (this.#context === undefined) {
+      assert(this.kind !== undefined);
+      this.#context = kFPContext[this.kind];
+    }
+    return this.#context;
+  }
```
ben-clayton, line 74: Why do you need to cache this at all? Could you not just use `kFPContext[this.kind]`, or use a getter?

src/webgpu/util/floating_point.ts
```diff
@@ -0,0 +1,3090 @@
...
+    };
+    4: {
+      2: FPMatrix;
+      3: FPMatrix;
+      4: FPMatrix;
+    };
+  };
+}
+
+abstract class FPContext {
```
ben-clayton, line 564: `Context` at least to me is something that carries some form of state. `FPInfo`, `FPTraits` ? Naming things is hard :).

AFAICT, this class is entirely stateless. It carries the `FPKind`, which is immutable. I probably would have gone with an interface here, just listing all the `public` members. I don't know JS well enough to know if there's any real pros or cons to using a class.

src/webgpu/util/floating_point.ts
```diff
@@ -0,0 +1,3090 @@
...
+  // Overrides - Utilities
+  public readonly correctlyRounded = correctlyRoundedF32;
+  public readonly isFinite = isFiniteF32;
+  public readonly isSubnormal = isSubnormalNumberF32;
+  public readonly flushSubnormal = flushSubnormalNumberF32;
+  public readonly oneULP = oneULPF32;
+  public readonly scalarBuilder = f32;
+}
+
+export const kFPContext = {
```
ben-clayton, line 3088: This is going to get used a lot. I'd be inclined to just call it `FP` to keep the noise down on usage. Or even just expose `f32` without wrapping it in a object.

ben-clayton (approved):
A big step in the right direction. I have a few design questions, but can certainly be done as followup.

src/webgpu/util/floating_point.ts
zoddicus: I agree context implies a stateful thing that is pass around . I think I will rename this to be FPTraits, since it is telling the system about the traits that distinguish each floating point type. (My first choice would have been FPType, but whoa boy that seems like it is going to clash a lot with things mentally)

wrt class v interface.
There is a lot of generic code for calculating intervals that lives in the abstract class here, some which has been factored into multiple functions to encapsulate and hide the details, so uses protected/private a lot. I don't think interfaces really support having private/protected members.
As is, since there is only one FP kind, this class looks like a lot of deck chair organization, but I am hoping when it comes to implementing abstract and f16 the actual implementation is going to be as terse as the f32 section, just defining the constants and specific utilities (other then maybe having to add in a few new constants for things that are magic number currently)

I think there is possibility of refactoring FPTraits to be an interface and having all of the generic framework code live off in a separate FPFramework or something object, but I think that would be a later refinement of what I have here if we feel we need it. I think the focus here is to organize the pile of free functions that I had into a structure that allows expressing new floating point types with minimum code duplication.

src/webgpu/util/floating_point.ts
zoddicus: I have renamed it FP. I prefer having it in const object like this, for the future where there is more than one kind defined, so code can use the kind as a key `FP[kind]`

src/webgpu/util/floating_point.ts
ben-clayton: SGTM

src/webgpu/util/floating_point.ts
ben-clayton: Ack

PR #2471, opened 2023-03-31
Fix maxStorageTexturesPerShaderStage validation tests

haoxli:
- The total number of fragment output resources must be not greater than maxFragmentCombinedOutputResources, but when testing maximum value, the number of storage textures + color targets exceeds this limit. The combination of fragment output resources has been covered in other tests, we just need to test single storage textures, not need color targets here.
- Remove redundant commas in getPerStageWGSLForBindingCombinationStorageTextures in limit_utils.ts
- Fix testCreatePipeline with async in limit_utils.ts

Issue: #

<hr>

### DIFF
--- a/src/webgpu/api/validation/capability_checks/limits/limit_utils.ts
+++ b/src/webgpu/api/validation/capability_checks/limits/limit_utils.ts
@@ -210,7 +210,7 @@ export function getPerStageWGSLForBindingCombinationStorageTextures(
     bindGroupTest,
     storageDefinitionWGSLSnippetFn,
     (numBindings: number, set: number) =>
-      `${range(numBindings, i => usageWGSLSnippetFn(i, set)).join('\n          ')};`,
+      `${range(numBindings, i => usageWGSLSnippetFn(i, set)).join('\n          ')}`,
     numBindings,
     extraWGSL
   );
@@ -783,6 +783,13 @@ export class LimitTestsImpl extends GPUTestBase {
     msg = ''
   ) {
     if (async) {
+      await this.shouldRejectConditionally(
+        'GPUPipelineError',
+        this.createPipelineAsync(createPipelineType, module),
+        shouldError,
+        msg
+      );
+    } else {
       await this.expectValidationError(
         () => {
           this.createPipeline(createPipelineType, module);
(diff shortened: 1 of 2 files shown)


greggman:
Sorry about these errors. I was unaware that you could make a fragment pipeline with no colorAttachments. Given that, it seems best to just fix the pipeline creation in limits_test.ts to not have colorAttachments and use a depthStencilAttachment for all of these tests, rather than special case this one test.

So I put up this

https://github.com/gpuweb/cts/pull/2472

lokokung (approved):

greggman (changes requested):
see other comment. I put up a different pr

haoxli:
Using depth-stencil attachment instead is a specific requirement for the `createPipeline, at_over` tests in maxStorageTexturesPerShaderStage.spec.ts due to the extra limit `maxFragmentCombinedOutputResources`, we‘d better just decouple this case from the extra limit.

I also thought about doing the changes in limit_utils.ts, but I didn't know how to explain why we need to enable the opitional depth -stencil attachment and empty color attachmens in the fragment stage for common cases, most cases have nothing to do with depth-stencil attachment. 

After the  #2471, I will also add `create`Compute/Render`PipelineDescriptor` helpers and simply the `createPipeline` function using `createComputePipeline`  and `createRenderPipeline` in the limit_utils.ts, currently the  descriptor statements are redundant in `createPipeline` and `createPipelineAsync`, but it's an optimization, not related to the test issues, to make your review easier, I did not put the changes in the fix PR.

lokokung:
I think I agree with Gregg that we should be fine to just use the depth stencil in the base. I'm not sure I understood the logic of it being specific to `maxStorageTexturesPerShaderStage` tests that Hao mentioned.

haoxli:
> I'm not sure I understood the logic of it being specific to maxStorageTexturesPerShaderStage tests that Hao mentioned.

For most of cases under limits, they have nothing to do with using color targets or depth-stencil targets, they work with both and no error even if we use the wrong targets.
But the maxStorageTexturesPerShaderStage tests are different, they are related to the limitation with color targets, targets number will affect the test results. So it's a checkpoint in these tests. We need to explain why we didn't test the point because other tests covered it and highlight we must need to use depth-stencil attachment instead here, which may help other readers to unstandstand these cases easliy. That is what the specific means. 

In my opinion using the color target by default could help to detect the problems, we should keep it, and if I saw this code for the first time, I would wonder why an empty color target is used.  That's why I insist on putting the changes in maxStorageTexturesPerShaderStage tests.

If these are not problems, I have no more comments in https://github.com/gpuweb/cts/pull/2472.

lokokung:
Ah, sorry I got confused with the name. I assume you mean the relationship of the limit w.r.t maxFragmentCombinedOutputResources? I think that it would make more sense to test that in the maxFragmentCombinedOutputResources test (which IIUC we already do), and only isolate to test the number of storage textures in the maxStorageTexturesPerShaderStage test.

haoxli:
Sure, I update the [comment](https://github.com/gpuweb/cts/pull/2471#issuecomment-1498414159).

PR #2472, opened 2023-03-31
Fix issues in LimitTestsImpl

greggman:
* swapped incorrect async path in testCreatePipeline

* added shared pipeline descriptor creation so the code is more D.R.Y.

* changed the RenderPipelineWithFragmentShader to have no colorAttachments and just a depthStencilAttachment

* removed semicolon in getPerStageWGSLForBindingCombinationStorageTextures

Thanks to @haoxli for finding these issues

see: https://github.com/gpuweb/cts/pull/2471

Issue: none

<hr>

### DIFF
--- a/src/webgpu/api/validation/capability_checks/limits/limit_utils.ts
+++ b/src/webgpu/api/validation/capability_checks/limits/limit_utils.ts
@@ -210,7 +210,7 @@ export function getPerStageWGSLForBindingCombinationStorageTextures(
     bindGroupTest,
     storageDefinitionWGSLSnippetFn,
     (numBindings: number, set: number) =>
-      `${range(numBindings, i => usageWGSLSnippetFn(i, set)).join('\n          ')};`,
+      `${range(numBindings, i => usageWGSLSnippetFn(i, set)).join('\n          ')}`,
     numBindings,
     extraWGSL
   );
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/capability_checks/limits/limit_utils.ts
```diff
@@ -700,41 +700,62 @@ export class LimitTestsImpl extends GPUTestBase {
...
+    return {
+      layout: 'auto',
+      vertex: {
+        module,
+        entryPoint: 'mainVS',
+      },
+      fragment: {
+        module,
+        entryPoint: 'mainFS',
+        targets: [],
```
haoxli, line 725: Using depth-stencil attachment instead is a specific requirement for the `createPipeline, at_over` tests in maxStorageTexturesPerShaderStage.spec.ts due to the extra limit `maxFragmentCombinedOutputResources`, we‘d better just decouple this case from the extra limit.

I also thought about doing the changes in limit_utils.ts, but I didn't know how to explain why we need to enable the opitional depth -stencil attachment and empty color attachmens in the fragment stage for common cases, most cases have nothing to do with depth-stencil attachment. 

After the  #2471, I will also add `create`Compute/Render`PipelineDescriptor` helpers and simply the `createPipeline` function using `createComputePipeline`  and `createRenderPipeline` in the limit_utils.ts, currently the  descriptor statements are redundant in `createPipeline` and `createPipelineAsync`, but it's an optimization, not related to the test issues, to make your review easier, I did not put the changes in the fix PR.

greggman: > Using depth-stencil attachment instead is a specific requirement for the createPipeline, at_over tests in maxStorageTexturesPerShaderStage.spec.ts due to the extra limit

The requirement not to use a depthStencilAttachment. The requirement is to NOT use a colorAttachment. None of the other tests require a colorAttachment either, They just require a pipeline that succeeds. So, removing the colorAtthacments and using a depthStencilAttachment works for all tests.

haoxli (approved):

PR #2505, opened 2023-04-26
Clarify that f64 castable constants are about WebIDL casting

### LINKED ISSUE #2501: Clarify the meaning of the castable constants in constants.ts
Specifically f64_not_castable and last_f64_castable.

This may just require a comment/name change, but how they are calculated should be examined to confirm it is correct.

zoddicus:
Adds comments clarifying what the values are, and changes the name to be more explicit.

The f16 versions were removed because a) there are not currently used, and b) looking at 
the WebIDL spec, I don't think conversion to half float is actually specified, just to single 
float.

Fixes #2501

<hr>

### DIFF
--- a/src/webgpu/util/constants.ts
+++ b/src/webgpu/util/constants.ts
@@ -403,10 +403,15 @@ export const kValue = {
         sixth: reinterpretU32AsF32(kBit.f32.positive.pi.sixth),
       },
       e: reinterpretU32AsF32(kBit.f32.positive.e),
-      first_f64_not_castable: reinterpretU32AsF32(kBit.f32.positive.max) / 2 + 2 ** 127, // mid point of 2**128 and largest f32
-      last_f64_castable: reinterpretU64AsF64(
+      // The positive number/f64 with the smallest magnitude which when cast to
+      // f32 will produce infinity, per the rounding rules of WebIDL.
+      first_webidl_not_castable: reinterpretU32AsF32(kBit.f32.positive.max) / 2 + 2 ** 127,
+      // The positive number/f64 with the largest magnitude which when cast to
+      // f32 will not produce infinity, instead rounding to FLT_MAX, per
+      // the rounding rules of WebIDL
+      last_webidl_castable: reinterpretU64AsF64(
         BigInt(reinterpretU32AsF32(kBit.f32.positive.max) / 2 + 2 ** 127) - BigInt(1)
-      ), // first_f64_not_castable minus one fraction bit of the 64 bit float representation
+      ),
     },
     negative: {
       max: reinterpretU32AsF32(kBit.f32.negative.max),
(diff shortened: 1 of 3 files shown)


src/webgpu/util/constants.ts
```diff
@@ -466,17 +476,17 @@ export const kValue = {
       min: reinterpretU16AsF16(kBit.f16.positive.min),
       max: reinterpretU16AsF16(kBit.f16.positive.max),
       zero: reinterpretU16AsF16(kBit.f16.positive.zero),
-      first_f64_not_castable: reinterpretU16AsF16(kBit.f16.positive.max) / 2 + 2 ** 16, // mid point of 2**16 and largest f16
-      last_f64_castable: reinterpretU64AsF64(
+      first_webidl_not_castable: reinterpretU16AsF16(kBit.f16.positive.max) / 2 + 2 ** 16, // mid point of 2**16 and largest f16
```
shrekshao, line 479: As Kai pointed out there seems to be a typo here. I revisited and I think it should be `reinterpretU16AsF16(kBit.f16.positive.max) / 2 + 2 ** 15, // mid point of 2**16 and largest f16`

`2**128` and `2**16` should be the infinity of f32 and f16

f32: exponent bit = 11111111<sub>2</sub> (is infinity in IEEE754) -> 2<sup>255-127</sup> -> `2**128`
f16: exponent bit = 11111<sub>2</sub> (is infinity in IEEE754) -> 2<sup>31-15</sup> -> `2**16`

(largest_f16 + infinity_16) / 2 = reinterpretU16AsF16(kBit.f16.positive.max) / 2 + 2 ** 15

zoddicus: Hmmm... apparently I didn't commit a change on my branch before pushing. I will do that now.

I have actually removed the f16 forms of this, since from my quick reading of the WebIDL spec, https://webidl.spec.whatwg.org/#es-float, half float is not defined/the conversion is not specified at all.
Since nothing in the repo depends on it, it makes sense to me to remove it entirely.

shrekshao: What I'm not sure is:

`2**128` is treated as a special value explicitly mentioned in [WebIDL](https://webidl.spec.whatwg.org/#es-float), when converting number/f64 to f32

But number/f64 to f16 conversion according to WebGPU/WGSL [spec](https://gpuweb.github.io/gpuweb/#conversions-to-wgsl) is first casting to wgslF32 then do f16(wgslF32). So taking the symmetry special value of f32 to fit f16 (`2**128` to `2**16`) doesn't make sense here. Maybe this is also the reason you didn't put the same comment here like you did for the f32 ones?

Kai: Did we have discussion about this when I add this code? I vaguely remember we did, and we found the case is over-complicated and we cannot fully test what we want. I think we may want to remove the `f16` ones and its usages to avoid confusion?

shrekshao: > I have actually removed the f16 forms of this

Yeah that was what I want to type at the same time. Removing LGTM.

You also need to remove its usages in `compute_pipeline.spec.ts` and `overrides.spec.ts`
(search with`/f16\..*first_f64_not_castable/`)

kainino0x (changes requested):
The spec uses the WebIDL casting rules but the casting in question does *not* happen at the WebIDL layer so I think this is more confusing. It is, currently, used only in conversion of pipeline-overridable constants, see spec references to
https://gpuweb.github.io/gpuweb/#abstract-opdef-to-wgsl-type
.  But when WGSL has f64 then it might have the same rules (if they can be consistent across hardware).

Sorry, I forgot to mention about f16. That algorithm does handle f16 (for f16 overrides), but it does so not with the WebIDL rules, but a combination of WebIDL + WGSL rules, for better or worse. Maybe we want to change that before f16 is done.

Referring to pipeline-overridable constants, instead of WebIDL, here, would be fine.

zoddicus:
I have taken another crack at this, I think this aligns with what @kainino0x was saying.

I noticed when working on this, that for calculating the last value that is castable, the code got the first non-castable value as a float, subtracted 1, then converted the result from a bitfield to a float, which seemed wrong to me, since we were applying a conversion from bits to floats to a float value...
I think there was a missing convert the non-castable float to a bitfield before the subtraction, so I added that in.

Also, I am not sure if this is correct in general, because the CTS tests that are affected already had a bunch of failing cases. This change doesn't appear to change how many are failing or why. But yeah...

src/webgpu/util/constants.ts
```diff
@@ -403,10 +413,17 @@ export const kValue = {
         sixth: reinterpretU32AsF32(kBit.f32.positive.pi.sixth),
       },
       e: reinterpretU32AsF32(kBit.f32.positive.e),
-      first_f64_not_castable: reinterpretU32AsF32(kBit.f32.positive.max) / 2 + 2 ** 127, // mid point of 2**128 and largest f32
-      last_f64_castable: reinterpretU64AsF64(
-        BigInt(reinterpretU32AsF32(kBit.f32.positive.max) / 2 + 2 ** 127) - BigInt(1)
```
zoddicus, line 408: The just raw converting this to a BigInt I think was wrong, since this is getting passed through a bitfield -> float conversion. I am pretty sure there needed to be a float -> bitfield conversion, to 1 bit could be removed, before converting to a float... I did the same thing for the other instances of this

zoddicus:
@kainino0x could you take another look at this?

kainino0x (approved):
lgtm, apologies for the delay!

PR #2526, opened 2023-04-27
[wgsl] Add workgroup_size validation tests.

### LINKED ISSUE #1691: wgsl: validate workgroup_size
From [9.2.1 Function Attributes for Entry Points](https://www.w3.org/TR/WGSL/#entry-point-attributes):

```rust
@compute @workgroup_size(8,4,1)
fn sorter() { }
```

```rust
@compute @workgroup_size(8u)
fn reverser() { }
```

```rust
// Using an pipeline-overridable constant.
@id(42) override block_width = 12u;
@compute @workgroup_size(block_width)
fn shuffler() { }
```

dj2:
This CL adds validation tests for the `workgroup_size` attribute.

Fixes #1691

<hr>

### DIFF
--- a/src/webgpu/shader/validation/shader_io/workgroup_size.spec.ts
+++ b/src/webgpu/shader/validation/shader_io/workgroup_size.spec.ts
@@ -73,6 +73,13 @@ const kWorkgroupSizeTests = {
     pass: true,
   },
 
+  const_expr: {
+    src: `const a = 4;
+    const b = 5;
+    @workgroup_size(a, b, a + b)`,
+    pass: true,
+  },
+
   override: {
     src: `@id(42) override block_width = 12u;
 @workgroup_size(block_width)`,
@@ -94,6 +101,13 @@ const kWorkgroupSizeTests = {
   },
   trailing_comma_z: {
     src: `@workgroup_size(8, 8, 8,)`,
+    pass: true,
+  },
+
+  override_expr: {
+    src: `override a = 5;
+    override b = 6;
+    @workgroup_size(a, b, a + b)`,
     pass: true,
   },
 
--- a/src/webgpu/shader/validation/shader_io/workgroup_size.spec.ts
+++ b/src/webgpu/shader/validation/shader_io/workgroup_size.spec.ts
@@ -96,6 +96,51 @@ const kWorkgroupSizeTests = {
     src: `@workgroup_size(8, 8, 8,)`,
     pass: true,
   },
+
+  // Mixed abstract is ok
+  mixed_abstract_signed: {
+    src: `@workgroup_size(8, 8i)`,
+    pass: true,
+  },
+  mixed_abstract_unsigned: {
+    src: `@workgroup_size(8u, 8)`,
+    pass: true,
+  },
+  // Mixed signed and unsigned is not
+  mixed_signed_unsigned: {
+    src: `@workgroup_size(8i, 8i, 8u)`,
+    pass: false,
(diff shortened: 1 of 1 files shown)


zoddicus:
Another corner, since floats are allowed, what happens if a fractional value, like 2.5 is supplied?

dj2:
Added test for mixing `u` and `i`. Added test for the max value listed in the spec `(256, 256, 64)`. Added `0` and negative value tests.

Did not test really large values as that's a pipeline-creation error and an adapter value. The float is not a valid value, which is tested already that floats are rejected.

This is testing the `0` and negative values here, which means they're shader-creation errors. Per spec, they're listed as pipeline-creation errors. Is it valid to error at shader-creation for them or does it have to be pipeline-creation (@dneto0)

zoddicus (commented):
lgtm, but I will let @dneto0 do the final approval

Sorry for misreading the float tests.

zoddicus (commented):
Trying the clear the changes requested bit

zoddicus (approved):
Apparently I have to approve to clear the changes requested bit.

src/webgpu/shader/validation/shader_io/workgroup_size.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+  },
+  negative_z: {
+    src: `@workgroup_size(8, 8, -8)`,
+    pass: false,
+  },
+
+  max_values: {
+    src: `@workgroup_size(256, 256, 64)`,
+    pass: true,
+  },
```
alan-baker, line 143: This is still reasonably likely to fail on many implementations due to the number of invocations limit. So maybe we should have a test that validates beyond these numbers?

alan-baker (commented):
Suggest adding some expression variants for both const-expression and override-expression.

src/webgpu/shader/validation/shader_io/workgroup_size.spec.ts
dj2: These numbers are the values listed in the WebGPU spec, so I believe they are the most that needs to be supported. These are all validation tests so they won't be executed on actual devices, but the values should compile I believe.

alan-baker: maxComputeInvocationsPerWorkgroup is 256 for reference. So (256,256,64) is too large. Not a big deal since it's only a validation test, but likewise we could ensure larger values are allowed (someone could also ask for an adaptor with higher limits).

dj2: I wasn't sure if requesting more would be portable, because the adapter can reject the higher limit, no? So, we can only validate that the minimum passes, otherwise it's out of our hands?

Or, since that's validation outside WGSL, we can just treat this as an i32 and pass in whatever and leave it for execution tests to catch big values?

alan-baker: I expect https://github.com/gpuweb/gpuweb/pull/4073 to be approved. So there is no upper bound to hit for a shader-creation error. So unless you're checking for pipeline errors the test could (but don't have to) cover larger values. Not really asking for a change here just more as an FYI. The main point of the review was the request for expression (not just declaration or literal) based values.

dj2:
Added `const_expr` and `override_expr` tests. PTAL.

alan-baker (approved):

PR #2591, opened 2023-05-04
Import spec files async to improve load times.

kdashg:
Issue: -

<hr>


kainino0x (approved):
Good idea, LGTM!

src/common/internal/tree.ts
```diff
@@ -253,6 +253,9 @@ export async function loadTreeForQuery(
   let foundCase = false;
   // L0 is suite:*
   const subtreeL0 = makeTreeForSuite(suite, isCollapsible);
+
+  const imports_start = performance.now();
```
kainino0x, line 257: `performance.now()` is currently disallowed because it doesn't exist in some versions of Node (though this might be fixed now). You can either import `now` from `util.ts` or drop this logging code, both are fine with me.

kdashg: It looks like node from 2017 has it. Do we support very far back?

kainino0x: I don't think we do anymore, but our lints are still set up to catch this. Let me just clean it up and rebase your PR.

kainino0x: Actually performance.now wasn't added as a *global* until Node 16.0.0 (2021-04-20, in https://github.com/nodejs/node/commit/f0bf373176). The thing we're using (`require('perf_hooks').performance`) is what was added in Node 8.5.0 in 2017.

kainino0x: There's another issue here with console.log, so I will figure it out and get this landed for you.

Issue #2495, opened 2023-04-22
Per-aspect capability data for multi-aspect texture formats


kainino0x:
// MAINTENANCE_TODO: Refactor this to separate per-aspect data for multi-aspect formats. In particular:
// - bytesPerBlock only makes sense on a per-aspect basis. But this table can't express that.
//   So we put depth24unorm-stencil8 and depth32float-stencil8 to be unsized formats for now.

kainino0x:
The only thing left right now is bytesPerBlock, which is a huge pain to refactor, but was also always the biggest problem, so it still needs to get fixed. But I'm not going to do it right now I think.

PR #2598, opened 2023-05-06
Refactor texture format info table (part 1)

kainino0x:
In part 1:
- Replaces the giant 2d table with just an object literal, which is organized better by with color/depth/stencil/colorRender sub-objects.
- Updates SOME tests which can be more easily updated to use the new fields. A lot of updates just insert a `!!` cast to boolean, because `.color`/`.depth`/`.stencil` changed from `boolean` to `undefined | object`.
- Keeps a bunch of the old (redundant) fields around, because it's nontrivial to update all tests to remove them. Right now, more than half of the size of the table is the deprecated fields. The values of both the new fields and the deprecated fields in the new object have been verified programmatically. The deprecated fields are verified in an intermediate commit of this PR, just for record-keeping. The new fields don't map directly, but doing the best mapping (not included in this PR anymore) possible shows these (expected) differences from old to new:
```diff
 ...
   "depth32float-stencil8": {
     "blockHeight": 1,
     "blockWidth": 1,
+    // bytesPerBlock was intentionally omitted before to avoid problems with dual-aspect format.
+    // It shows up here because this is derived from (info.color ?? info.depth ?? info.stencil).bytes.
+    "bytesPerBlock": 4,
     "color": false,
     "copyDst": false,
-    // Bug (possibly intentional) in the old info.
-    "copySrc": false,
+    "copySrc": true,
     "depth": true,
     "feature": "depth32float-stencil8",
     "multisample": true,
     "renderable": true,
     "resolve": false,
     "sampleType": "depth",
     "stencil": true,
     "storage": false
   },
 ...
   "rg11b10ufloat": {
     "blockHeight": 1,
     "blockWidth": 1,
     "bytesPerBlock": 4,
     "color": true,
     "copyDst": true,
     "copySrc": true,
     "depth": false,
     "multisample": false,
-    // Format is not renderable in core, so these apply only to optional feature tests.
-    // alignment/byteCost are now part of a substructure, so, reasonably, can't be
-    // specified here anymore. The constants will move into that test when it's updated.
-    // Eventually, this will be fixed with better support for optional features in this table.
-    "renderTargetComponentAlignment": 4,
-    "renderTargetPixelByteCost": 8,
     "renderable": false,
     "resolve": false,
     "sampleType": "float",
     "stencil": false,
     "storage": false
   },
 ...
```
Texture format info also influences a lot of test parameterizations. Verified that the total list of test cases hasn't changed.

In upcoming parts:
- Move this into a separate file because it's so large.
- Fix all of the tests using old data, incrementally.
- Remove all of the deprecated fields.
- Eventually, support optional features.

Passing test run: https://dawn-review.googlesource.com/c/dawn/+/132081/5?tab=checks
Issue: #2495

<hr>


kainino0x:
@lokokung this is finally ready, PTAL! Diff looks huge but most of it is the table itself, and everything else should be fairly trivial. More interesting changes will come in future PRs.
Comments welcome on the structure of the new table, though I may fix comments on that as followups rather than in this PR.

src/webgpu/capability_info.ts
```diff
@@ -87,135 +87,1083 @@ export const kGeneratableErrorScopeFilters = kErrorScopeFilters.filter(
...
+
+  // IMPORTANT:
+  // Add new top-level keys both here and in TextureFormatInfo_TypeCheck.
+} as const;
+/**
+ * Takes `table` and applies `defaults` to every row, i.e. for each row,
+ * `{ ... kUniversalDefaults, ...defaults, ...row }`.
+ * This only operates at the first level; it doesn't support defaults in nested objects.
+ */
+function tableWithDefaults<Defaults extends {}, Table extends { readonly [K: string]: {} }>({
```
lokokung, line 123: This is only for texture formats, so maybe `textureTableWithDefaults`? Or maybe it's possible to generalize this more and pass in like a templated `UniveralDefaults` also? Not sure if that's possible/worth the effort though... So maybe just a rename is good enough?

For naming also, `kUniversalDefaults`, maybe `kUniversalFormatDefaults`?

src/webgpu/capability_info.ts
```diff
@@ -295,56 +1243,124 @@ export const kCanvasColorSpacesInfo: {
...
+interface TextureFormatStencilAspectInfo extends TextureFormatAspectInfo {
+  bytes: 1;
+  type: 'uint';
+}
+
+/**
+ * Per-GPUTextureFormat info.
+ * This is not actually the type of values in kTextureFormatInfo; that type is fully const
+ * so that it can be narrowed very precisely at usage sites by the compiler.
+ * This type exists only a type check on the inferred type of kTextureFormatInfo.
```
lokokung, line 1278: nit. This type exists only as* a

src/webgpu/capability_info.ts
```diff
@@ -87,135 +87,1083 @@ export const kGeneratableErrorScopeFilters = kErrorScopeFilters.filter(
...
+    r8unorm: {
+      color: { type: 'float', copySrc: true, copyDst: true, storage: false, bytes: 1 },
+      colorRender: { blend: true, resolve: true, byteCost: 1, alignment: 1 },
+      renderable: true,
+      /*prettier-ignore*/ get renderTargetComponentAlignment() { return this.colorRender.alignment; },
+      /*prettier-ignore*/ get renderTargetPixelByteCost() { return this.colorRender.byteCost; },
+      multisample: true,
+      /*prettier-ignore*/ get sampleType() { return this.color.type; },
+      copySrc: true,
+      copyDst: true,
```
lokokung, line 159: I think for the regular formats `copySrc` and `copyDst` seem to always be `true` so maybe we can default them?

src/webgpu/capability_info.ts
```diff
@@ -87,135 +87,1083 @@ export const kGeneratableErrorScopeFilters = kErrorScopeFilters.filter(
...
+    // plain, 8 bits per component
+
+    r8unorm: {
+      color: { type: 'float', copySrc: true, copyDst: true, storage: false, bytes: 1 },
+      colorRender: { blend: true, resolve: true, byteCost: 1, alignment: 1 },
+      renderable: true,
+      /*prettier-ignore*/ get renderTargetComponentAlignment() { return this.colorRender.alignment; },
+      /*prettier-ignore*/ get renderTargetPixelByteCost() { return this.colorRender.byteCost; },
+      multisample: true,
+      /*prettier-ignore*/ get sampleType() { return this.color.type; },
```
lokokung, line 157: Not sure if it's possible but the getters are repeated a lot (and clutter the table quite a bit...), idk if it's possible to make them conditionally defined depending on what sub-types are defined? i.e. if `color` is defined, then define `get sampleType() { return this.color.type; };`, otherwise don't define it? Not sure what is possible in TS though.

If it were possible, it would also make the deprecating part a lot easier since we would only have to delete the default parts (not the defs in every entry).

Feel free to leave these for now if it's not immediate obvious how to accomplish that though since I assume they are removed entirely in a follow up or something anyways.

src/webgpu/capability_info.ts
```diff
@@ -87,135 +87,1083 @@ export const kGeneratableErrorScopeFilters = kErrorScopeFilters.filter(
...
+      bytesPerBlock: undefined,
+      sampleType: 'depth',
+      renderable: true,
+    },
+    'depth24plus-stencil8': {
+      depth: { type: 'depth', copySrc: false, copyDst: false, storage: false, bytes: undefined },
+      stencil: { type: 'uint', copySrc: true, copyDst: true, storage: false, bytes: 1 },
+      copySrc: false,
+      copyDst: false,
+      bytesPerBlock: undefined,
```
lokokung, line 637: Unnecessary

src/webgpu/capability_info.ts
```diff
@@ -87,135 +87,1083 @@ export const kGeneratableErrorScopeFilters = kErrorScopeFilters.filter(
...
-  'bc6h-rgb-float':        [            ,              ,          ,        ,        ,          ,          ,          ,          ,              'float',              16,            4,             4,                            ,                                 ,   'texture-compression-bc'],
-  'bc7-rgba-unorm':        [            ,              ,          ,        ,        ,          ,          ,          ,          ,              'float',              16,            4,             4,                            ,                                 ,   'texture-compression-bc',   'bc7-rgba-unorm'],
-  'bc7-rgba-unorm-srgb':   [            ,              ,          ,        ,        ,          ,          ,          ,          ,              'float',              16,            4,             4,                            ,                                 ,   'texture-compression-bc',   'bc7-rgba-unorm'],
+const kBCTextureFormatInfo = tableWithDefaults({
+  defaults: {
```
lokokung, line 654: `sampleType`, `copySrc`, and `copyDst` look default-able for easier deprecation. (Both here and in the other compressed formats too.)

src/webgpu/capability_info.ts
```diff
@@ -295,56 +1243,124 @@ export const kCanvasColorSpacesInfo: {
...
+  bytes: 1;
+  type: 'uint';
+}
+
+/**
+ * Per-GPUTextureFormat info.
+ * This is not actually the type of values in kTextureFormatInfo; that type is fully const
+ * so that it can be narrowed very precisely at usage sites by the compiler.
+ * This type exists only a type check on the inferred type of kTextureFormatInfo.
+ * Documentation is also written here, but not actually visible to the IDE.
```
lokokung, line 1279: Ah... this is unfortunate... Is this fixable in follow up CLs once we aren't in between states?

lokokung (approved):
Thanks for the work here Kai!! Overall looks pretty good! Just some minor suggestions to hopefully make the follow up CLs easier to review.

src/webgpu/capability_info.ts
kainino0x: Good point, originally I was moving all this stuff into another file and then I realized it was too much for one PR. I'll update the names.

src/webgpu/capability_info.ts
kainino0x: All of these getters are kind of hacked in. They're all for the deprecated fields so I didn't put much effort into them... just wrote a small block of code that I could copy paste into every format. They could be the same for every format but I didn't know how to / didn't want to figure out how to make copy getters from the default object into the final object - the tableWithDefaults thing actually calls all of the getters when the table is being constructed so they get baked into the object as actual values.

I'm just going to leave them until they're removed.

src/webgpu/capability_info.ts
kainino0x: Unlikely. I didn't realize this before this PR, but we actually already had this problem.

We need the type of the entire object to be const because we rely on having the exact type of every format for a bunch of type narrowing that happens elsewhere - like if you call `kTextureFormat[format]` where `format` is of type `"rgba8unorm" | "bgra8unorm"` then a bunch of the fields that are normally optional become guaranteed.

But then because every single piece of data in the entire structure has its own place in the type, every single one of those would need the documentation attached to it individually. Which would require defining the entire type explicitly (redundantly) and duplicating the documentation all over.

I think typedoc would need a pretty substantial piece of magic to let us declare the documentation in a way that will distribute it throughout the inferred type for us, https://github.com/microsoft/tsdoc/issues/241 probably.

kainino0x: for the record, this turned out to be solvable: #3351

Issue #2593, opened 2023-05-04
It is too easy to introduce non-serializable Comparators


zoddicus:
It is too easy to create Comparators in the testing framework that don't have defined serialization for using the case cache and then try to pass them into the cache.

Depending on your setup, you might not even realize you have introduced a bug depending on how you are running the CTS.

Also it isn't obvious is review that this is happening.

This has been introduced in the bitcast tests `src/webgpu/shader/execution/expression/call/builtin/bitcast.spec.ts` and is preventing rolling new versions of the CTS into the Dawn/Chromium project (https://bugs.chromium.org/p/dawn/issues/detail?id=1795), since the cache is throwing an error.

zoddicus:
Two steps for fixing this:
1. Remove the usage of the case cache from `bitcast.spec.ts`, since these tests are correct and will run without the cache, just a bit slower.
2. Refactor the existing Comparator code to include serialization logic on the Comparator itself, so that when you create a new instance of it, you have to either supply serialization functionality, or explicitly leave it undefined.

PR #2601, opened 2023-05-08
Make need for serialization of Comparators more obvious

zoddicus:
As-is the CTS framework just requires you to define a function to create a Comparator and pass it as a Expectation in a Case. This causes issues if the writer then attempts to implement using the CaseCache, since that requires the Comparator to have defined serialization and deserialization. There is no type level enforcement of this, so errors are not detected until runtime with the cache enabled.

This PR reworks the Comparator type to have an impl function and kind tag as required properties, along with optional data for serialization. This forces the user to actively bypass the type system to define a Comparator without serialization. Additional documentation has been added around the relevant definitions to point the user in the correct direction.

This PR doesn't change the entire type system to strictly require serialization/deserialization. This is because some of the internals of the testing framework use Comparators as part of building and running tests.

One could add in serialization/deserialization code to these Comparators, but this would allow silent errors when internal components leak out into the cache unintentionally. The other option is to split the current Case/Expectation system to have Serializable and non-Serializable implementations. This creates a bunch of duplicated code and significant amounts of type conversion for a bit of intellectual cleanliness.

This implementation chooses to make the framework user's error a type level error, so detectable by tooling, since there is likely to be many more changes to code defining tests and using the framework vs the framework itself. Issues in the framework will still cause runtime issues, and thus integration failures, but hopefully localize them to code that is changing less frequently.

The bitcast tests are given an explicit opt-out from the serialization. A follow up PR will rewrite them to properly define their Comparators and remove this legacy opt-out.

Issue #2593

<hr>

### DIFF
--- a/src/webgpu/shader/execution/expression/call/builtin/bitcast.spec.ts
+++ b/src/webgpu/shader/execution/expression/call/builtin/bitcast.spec.ts
@@ -75,15 +75,15 @@ const f32RangeWithInfAndNaN: number[] = [
 ];
 
 const anyF32: Comparator = {
-  impl: v => ({ matched: true, got: `${v}`, expected: 'any f32' }),
+  compare: v => ({ matched: true, got: `${v}`, expected: 'any f32' }),
   kind: 'bitcast',
 };
 const anyI32: Comparator = {
-  impl: v => ({ matched: true, got: `${v}`, expected: 'any i32' }),
+  compare: v => ({ matched: true, got: `${v}`, expected: 'any i32' }),
   kind: 'bitcast',
 };
 const anyU32: Comparator = {
-  impl: v => ({ matched: true, got: `${v}`, expected: 'any u32' }),
+  compare: v => ({ matched: true, got: `${v}`, expected: 'any u32' }),
   kind: 'bitcast',
 };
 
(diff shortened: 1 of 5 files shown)


src/webgpu/shader/execution/expression/call/builtin/bitcast.spec.ts
```diff
@@ -74,9 +74,18 @@ const f32RangeWithInfAndNaN: number[] = [
   ].map(u => reinterpretU32AsF32(u)),
 ];
 
-const anyF32: Comparator = v => ({ matched: true, got: `${v}`, expected: 'any f32' });
-const anyI32: Comparator = v => ({ matched: true, got: `${v}`, expected: 'any i32' });
-const anyU32: Comparator = v => ({ matched: true, got: `${v}`, expected: 'any u32' });
+const anyF32: Comparator = {
+  impl: v => ({ matched: true, got: `${v}`, expected: 'any f32' }),
```
ben-clayton, line 78: nit: `Comparator` is just TypeScript metadata, and there's no way to know that an object exposing an `impl` field has anything to do with a `Comparator`.
I'd suggest renaming `impl` to `compare` so the field make sense in isolation.

ben-clayton (approved):

src/webgpu/shader/execution/expression/call/builtin/bitcast.spec.ts
zoddicus: Done

Issue #2610, opened 2023-05-12
Handle uncaught internal errors in device_pool


kainino0x:
It currently catches `out-of-memory` and `validation` errors but not `internal` errors because they didn't exist when the code was written.

PR #2611, opened 2023-05-12
Handle uncaught internal errors in device_pool

austinEng:
Also reorder the error scopes such that we catch OOM in the inner most scope. We also check for OOM first. This is because it does not allow device reuse. Errors that result in no reuse should be checked first. Otherwise, if there is both a validation error and an OOM error, and we check for validation first, we may end up throwing the error and reusing the device because we didn't check for OOM as well.

Issue: #2610

<hr>


kainino0x (approved):
Thanks for the fix, nice catch about the check ordering! I don't think the pushErrorScope order is supposed to matter but the if() order definitely does.

PR #2619, opened 2023-05-17
Add execution tests for AbstractFloat -> f32 assignment

### LINKED ISSUE #2617: wgsl: execution tests for assigning `AbstractFloat` to `f32`
For values precisely defined in f32, the result should be exact. Otherwise it should be correctly rounded.

zoddicus:
This includes related infra changes; AbstractFloat range generators, passing AbstractFloats into shaders, bigints being
usable directly in test cases/names (JSONification).

Fixes #2617

<hr>


src/common/internal/query/json_param_value.ts
```diff
@@ -42,12 +45,21 @@ function stringifyFilter(k: string, v: unknown): unknown {
...
+      `${v} matches bigint magic pattern for stringification, so cannot be used`
+    );
   }
 
   if (Object.is(v, -0)) {
     return jsNegativeZeroMagicValue;
   }
 
+  if (typeof v === 'bigint') {
+    return `_${v}n_`;
```
ben-clayton, line 60: Do we need the underscores? Is `(\d+)n` not sufficient?

src/webgpu/shader/execution/expression/expression.ts
```diff
@@ -418,6 +422,27 @@ struct Output {
...
+  expressionBuilder: ExpressionBuilder
+): string {
+  // f64/AbstractFloat values cannot be stored in an array
+  if (parameterTypes.some(ty => scalarTypeOf(ty).kind === 'f64')) {
+    return '';
+  }
+  return `
+const values = array(
+  ${cases
+    .map(c => toStorage(resultType, expressionBuilder(map(c.input, v => v.wgsl()))))
```
ben-clayton, line 441: I can't recall why we go `toStorage` for this `const` array. It seems like it would be generally simpler to fill this with the non-storage values, which supports both `bool`s and abstract-numerics.

src/webgpu/shader/execution/expression/expression.ts
```diff
@@ -467,7 +492,19 @@ export function basicExpressionBuilder(expressionBuilder: ExpressionBuilder): Sh
...
+        // intermediate store, which will concretize the value earlier
+        body = cases
+          .map(
+            (c, i) =>
+              `  outputs[${i}].value = ${toStorage(
+                resultType,
+                expressionBuilder(map(c.input, v => v.wgsl()))
+              )};`
+          )
+          .join('\n  ');
```
ben-clayton, line 506: Remove the `toStorage` from the `values` array, and I don't think you need this any more.

src/webgpu/shader/execution/expression/expression.ts
```diff
@@ -467,7 +492,19 @@ export function basicExpressionBuilder(expressionBuilder: ExpressionBuilder): Sh
...
+      } else if (globalTestConfig.unrollConstEvalLoops) {
         body = cases.map((_, i) => `  outputs[${i}].value = values[${i}];`).join('\n  ');
```
ben-clayton, line 508: ... but you'll need to add a `toStorage` here.

src/webgpu/shader/execution/expression/unary/abstract_float_assignment.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+      return { input: f64(f), expected: FP.f32.correctlyRoundedInterval(f) };
+    });
+  },
+});
+
+g.test('f32')
+  .specURL('https://www.w3.org/TR/WGSL/#floating-point-conversion')
+  .desc(
+    `
+narrowing down to f32
```
ben-clayton, line 29: So if the RHS does not first explicitly convert the abstract-float to a `f32`, then this is testing [`concretization`](https://www.w3.org/TR/WGSL/#concretization).

src/webgpu/util/conversion.ts
```diff
@@ -779,6 +779,8 @@ export class Scalar {
     };
     if (isFinite(this.value as number)) {
       switch (this.type.kind) {
+        case 'f64':
```
ben-clayton, line 782: I'm wondering whether we should have a different type for `abstract-float`. While they are currently represented as `f64`, they are special - no suffix, unique rules around concretization, untypable, etc.
The day WGSL gets a proper `f64` type, or `abstract-float`'s precision gets bumped to something > 64 bits, it's going to be hard to distinguish `f64` from `abstract-float`.

zoddicus: Done

src/common/internal/query/json_param_value.ts
kainino0x: I'm fine with no underscores, since we'll detect if a collision ever happens.

kainino0x: Also good catch on the `^$`.

PR #2630, opened 2023-05-24
Convert integer arithmetic tests to use sparse inputs

### LINKED ISSUE #2623: wgsl: Reduce number of test cases for integer arithmetic operations
These tests are regularly listed as being the longest running for integration into [Chrome](https://bugs.chromium.org/p/dawn/issues/detail?id=1783) , this is likely due to using the full integer test range in a binary operation, so generating N^2 test cases.

This is the integer analogue to https://github.com/gpuweb/cts/issues/2613

zoddicus:
This reduces the number of test cases substantially, since these will be generating N^2 cases per operation.

Fixes #2623

<hr>

### DIFF
--- a/src/webgpu/util/math.ts
+++ b/src/webgpu/util/math.ts
@@ -919,11 +919,15 @@ export function filteredF64Range(
 /** Short list of i32 values of interest to test against */
 const kInterestingI32Values: number[] = [
   kValue.i32.negative.max,
-  kValue.i32.negative.max / 2,
+  Math.trunc(kValue.i32.negative.max / 2),
+  -256,
+  -10,
   -1,
   0,
   1,
-  kValue.i32.positive.max / 2,
+  10,
+  256,
+  Math.trunc(kValue.i32.positive.max / 2),
   kValue.i32.positive.max,
 ];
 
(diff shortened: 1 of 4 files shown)


src/webgpu/util/math.ts
```diff
@@ -919,11 +919,15 @@ export function filteredF64Range(
...
+  Math.trunc(kValue.i32.negative.max / 2),
+  -256,
+  -10,
   -1,
   0,
   1,
-  kValue.i32.positive.max / 2,
+  10,
+  256,
+  Math.trunc(kValue.i32.positive.max / 2),
```
amaiorano, line 930: Why do we need to truncate these?

amaiorano (approved):
One minor question, but LGTM!

src/webgpu/util/math.ts
zoddicus: We chatted about this OOB, I am including the summary of it here, for future reference.

These values are odd integers, so when divided return a fractional result, which is stored in the `number` since it is a f64 under the hood. This leads to other calculations ending up slightly off and the test failing.
The changes that I made to the test generation code should catch this and quantize things correctly, but it feels weird to have a known non-integer in a list of nominally integer value.s

Issue #2628, opened 2023-05-24
wpt lint errors


sagudev:
When running `wpt lint` on `out-wpt`:
```
ERROR:lint:webgpu/webgpu/shader/execution/memory_model/coherence.spec.js:467: Whitespace at EOL (TRAILING WHITESPACE)
ERROR:lint:webgpu/webgpu/shader/execution/expression/call/builtin/atomics/harness.js:49: Whitespace at EOL (TRAILING WHITESPACE)
ERROR:lint:webgpu/webgpu/shader/execution/expression/call/builtin/atomics/harness.js:117: Whitespace at EOL (TRAILING WHITESPACE)
ERROR:lint:webgpu/webgpu/web_platform/reftests/ref/canvas_complex-ref.html:24: Whitespace at EOL (TRAILING WHITESPACE)
ERROR:lint:webgpu/webgpu/api/validation/capability_checks/limits/maxSampledTexturesPerShaderStage.spec.js:35: Whitespace at EOL (TRAILING WHITESPACE)
ERROR:lint:webgpu/webgpu/api/validation/capability_checks/limits/maxSampledTexturesPerShaderStage.spec.js:63: Whitespace at EOL (TRAILING WHITESPACE)
ERROR:lint:webgpu/webgpu/api/validation/capability_checks/limits/maxSampledTexturesPerShaderStage.spec.js:97: Whitespace at EOL (TRAILING WHITESPACE)
ERROR:lint:webgpu/webgpu/api/validation/capability_checks/limits/maxSamplersPerShaderStage.spec.js:35: Whitespace at EOL (TRAILING WHITESPACE)
ERROR:lint:webgpu/webgpu/api/validation/capability_checks/limits/maxSamplersPerShaderStage.spec.js:63: Whitespace at EOL (TRAILING WHITESPACE)
ERROR:lint:webgpu/webgpu/api/validation/capability_checks/limits/maxSamplersPerShaderStage.spec.js:97: Whitespace at EOL (TRAILING WHITESPACE)
ERROR:lint:webgpu/webgpu/api/operation/compute_pipeline/overrides.spec.js:87: Whitespace at EOL (TRAILING WHITESPACE)
ERROR:lint:webgpu/webgpu/api/operation/compute_pipeline/overrides.spec.js:91: Whitespace at EOL (TRAILING WHITESPACE)
[...]

PR #2668, opened 2023-05-25
Include testharness.js in exposed interface tests

kainino0x:
These tests should be run as testharness.js tests, even though they don't actually use any of testharness.js. Include testharness.js anyway to make `wpt lint` not misidentify these as "visual" tests.

Initially I was going to rewrite these as vanilla testharness.js files, but I'm pretty sure I verified these tests worked in the past, and that testharness.js will catch plain exceptions thrown at the toplevel (of an imported module, in this case).

Issue #2628

### DIFF
--- a/src/webgpu/idl/exposed.http.html
+++ b/src/webgpu/idl/exposed.http.html
@@ -5,6 +5,8 @@
     <title>WebGPU exposed items (non-HTTPS)</title>
     <meta name=assert content="WebGPU should not be exposed on a non-[SecureContext]">
     <link rel=help href='https://gpuweb.github.io/gpuweb/'>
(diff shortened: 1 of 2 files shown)


kainino0x:
@sagudev PTAL

kainino0x:
@sagudev I've verified this makes `wpt lint` happy but I'm not set up to run the actual tests. Are you able to easily run them and verify that top-level exceptions will indeed cause the tests to fail?

sagudev:
@kainino0x I only have servo set up, but unfortunately it still cannot run test properly.

kainino0x:
OK, I think this is right but if not we can fix it afterward. I'm just going to land it to make forward progress.

PR #2744, opened 2023-05-25
Remove console.log from a file exported to wpt

kainino0x:
WPT has a lint that disallows console.log. This print is somewhat useful for debugging but seems generally unnecessary. I would keep it, but I'm not sure how other than doing shenanigans in the WPT build process (can't suppress the lint because the suppression file lives outside the webgpu directory).

Issue: #2628

### DIFF
--- a/src/common/util/navigator_gpu.ts
+++ b/src/common/util/navigator_gpu.ts
@@ -1,6 +1,8 @@
 /// <reference types="@webgpu/types" />
 
-import { assert } from './util.js';
+import { TestCaseRecorder } from '../framework/fixture.js';
+
+import { ErrorWithExtra, assert } from './util.js';
 
 /**
  * Finds and returns the `navigator.gpu` object (or equivalent, for non-browser implementations).
@@ -45,7 +47,7 @@ export function setDefaultRequestAdapterOptions(options: GPURequestAdapterOption
  * Finds and returns the `navigator.gpu` object (or equivalent, for non-browser implementations).
  * Throws an exception if not found.
  */
-export function getGPU(): GPU {
+export function getGPU(recorder: TestCaseRecorder): GPU {
   if (impl) {
     return impl;
   }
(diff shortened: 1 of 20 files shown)


greggman:
The reason it's there is I've wasted several days of my life debugging the wrong adapter on a 2 GPU device 😅

So, I'd be really nice if the adapter info was displayed somewhere when running the CTS. On the HTML page, in each test, somewhere where you can see you got the adapter you thought you were testing. If there's no info, then I'll likely run into that issue again where I think I'm testing one adapter but actually testing a different adapter.

Can you think of a good way to surface this info ... and does it belong in the logs so you can see it on the bots?

kainino0x:
OK, makes sense. I thought about surfacing this in a way that it would log to WPT if using WPT, and console if not, similar to what we have for setTimeout. Will figure out how to do that.

src/common/util/navigator_gpu.ts
```diff
@@ -58,14 +64,16 @@ export function getGPU(): GPU {
...
-          console.log(info);
-        }
-      });
+      const promise = oldFn.call(this, { ...defaultRequestAdapterOptions, ...options });
+      if (recorder) {
+        void promise.then(async adapter => {
+          if (adapter) {
+            const info = await adapter.requestAdapterInfo();
+            const infoString = `Adapter: ${info.vendor} / ${info.architecture} / ${info.device}`;
+            recorder.debug(new ErrorWithExtra(infoString, () => ({ adapterInfo: info })));
```
kainino0x, line 73: Instead of falling back to `console.log`ing I've just wired this up to the TestCaseRecorder, since it's the only way to get logs out in WPT anyway.

WDYT of making it a `debug` level log? That means it'll only show up if you specify `?debug=1` in standalone (and not at all in WPT, and at least not by default in Chromium's harness)

kainino0x:
@greggman PTAL

src/common/util/navigator_gpu.ts
greggman: sgtm

greggman (approved):
lgtm

PR #2748, opened 2023-06-01
Run `generate-listings` before `build-wpt`

### LINKED ISSUE #2747: wpt: listing.js is not available in the first run
When running `npm run wpt` in clean checkout, `out-wpt/webgpu` does not contain `listing.js`. Running same command for the second time without cleaning, gives `listing.js`.

sagudev:
Fixes #2747


kainino0x (approved):
Ahh, I probably messed this up thinking generate-listings should run afterward to overwrite the listing.js generate by build-wpt. But instead it needs to run first so build-wpt can copy it from out/.

This build system is getting out of hand :(

PR #2761, opened 2023-06-09
Uniformity validation tests for function variable values

alan-baker:
* Add validation tests to cover Function-scope Variable Value Analysis
* Pointers are deferred to a subsequent set of tests (i.e. not let decls of pointers)

Issue: #

<hr>

### DIFF
--- a/src/webgpu/shader/validation/uniformity/uniformity.spec.ts
+++ b/src/webgpu/shader/validation/uniformity/uniformity.spec.ts
@@ -557,6 +557,16 @@ const kFuncVarCases = {
     cond: `x > 0`,
     uniform: `never`,
   },
+  if_nonescaping_nonuniform: {
+    typename: `u32`,
+    typedecl: ``,
+    assignment: `if uniform_cond {
+      x = nonuniform_value[0];
+      return;
+    }`,
+    cond: `x > 0`,
+    uniform: `init`,
+  },
   loop_body_depends_on_continuing_uniform: {
     typename: `u32`,
     typedecl: ``,
--- a/src/webgpu/shader/validation/uniformity/uniformity.spec.ts
+++ b/src/webgpu/shader/validation/uniformity/uniformity.spec.ts
@@ -1,6 +1,7 @@
 export const description = `Validation tests for uniformity analysis`;
 
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { keysOf } from '../../../../common/util/data_tables.js';
 import { unreachable } from '../../../../common/util/util.js';
 import { ShaderValidationTest } from '../shader_validation_test.js';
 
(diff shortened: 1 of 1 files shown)


src/webgpu/shader/validation/uniformity/uniformity.spec.ts
```diff
@@ -354,3 +355,900 @@ fn main(@builtin(${t.params.builtin}) p : ${t.params.type}) {
...
+      return ``;
+    }
+    case `uniform`: {
+      return `= uniform_value[3]`;
+    }
+    case `nonuniform`: {
+      return `= nonuniform_value[3]`;
+    }
+    default: {
+      unreachable(`Unhandled init`);
```
zoddicus, line 401: nit: Can you include the unexpected init value in the unreachable string?

src/webgpu/shader/validation/uniformity/uniformity.spec.ts
```diff
@@ -354,3 +355,900 @@ fn main(@builtin(${t.params.builtin}) p : ${t.params.type}) {
...
+          return true;
+        }
+        case `nonuniform`: {
+          return false;
+        }
+        default: {
+          break;
+        }
+      }
+      unreachable(`Unhandled init`);
```
zoddicus, line 380: nit: Can you include the unexpected init value in the unreachable string?

src/webgpu/shader/validation/uniformity/uniformity.spec.ts
```diff
@@ -354,3 +355,900 @@ fn main(@builtin(${t.params.builtin}) p : ${t.params.type}) {
...
+      break;
+    }
+    default: {
+      unreachable(`Unhandled uniform`);
```
zoddicus, line 384: nit: Can you include the unexpected uniform value in the unreachable string?

src/webgpu/shader/validation/uniformity/uniformity.spec.ts
```diff
@@ -354,3 +355,900 @@ fn main(@builtin(${t.params.builtin}) p : ${t.params.type}) {
...
+    }
+  }
+}
+
+function generateFunctionVarInit(init: string): string {
```
zoddicus, line 389: nit:
Could this be a const?
Something like this:
```
const var_init = {
  none: '';
  uniform: '= uniform_value[3]';
  nonuniform: '= nonuniform_value[3]';
};
```

src/webgpu/shader/validation/uniformity/uniformity.spec.ts
```diff
@@ -354,3 +355,900 @@ fn main(@builtin(${t.params.builtin}) p : ${t.params.type}) {
 
     t.expectCompileResult(t.params.uniform, code);
   });
+
+function expectedUniformity(uniform: string, init: string): boolean {
```
zoddicus, line 359: nit:
Could this be a const?
Something like this:
```
const expected_uniformity = {
  always: true;
  never: false;
  init: {
     none: true;
     uniform: true;
     nonuniform: false;
  };
};
```

zoddicus: You can then use `expected_uniformity['always']` or `expected_uniformity[foo]`, where foo is a string

src/webgpu/shader/validation/uniformity/uniformity.spec.ts
zoddicus: You can then use `var_init['none']` or `var_init[oo]`, where foo is a string

src/webgpu/shader/validation/uniformity/uniformity.spec.ts
```diff
@@ -354,3 +355,900 @@ fn main(@builtin(${t.params.builtin}) p : ${t.params.type}) {
...
+
+g.test('function_variables')
+  .desc(`Test uniformity of function variables`)
+  .params(u =>
+    u
+      .combine('case', keysOf(kFuncVarCases))
+      .combine('init', ['none', 'uniform', 'nonuniform'] as const)
+  )
+  .fn(t => {
+    const code = `
```
zoddicus, line 1219: nit: Since you are pulling multiple fields from a single entry in kFuncCases below, possibly pull the element out one and save it to a local variable `const func_case = kFuncVarCases[t.params.case];`

src/webgpu/shader/validation/uniformity/uniformity.spec.ts
zoddicus: Though I guess this would require a bit of switching logic in the test runner, i.e. `uniform === 'init' ? expected_uniformity['init'][init]: expected_uniformity[uniform];`

zoddicus (commented):
overall looks reasonable from a CTS/TS perspective, some style cleanup things, but nothing major.

I will let @jrprice do final approval, since he is more familiar with the specifics of uniformity.

src/webgpu/shader/validation/uniformity/uniformity.spec.ts
zoddicus: This isn't needed if changed to a const as I suggest above

zoddicus (commented):
look good to me from CTS/TS perspective.

@jrprice can you review the uniformity specific aspects?

src/webgpu/shader/validation/uniformity/uniformity.spec.ts
```diff
@@ -354,3 +355,867 @@ fn main(@builtin(${t.params.builtin}) p : ${t.params.type}) {
...
+    typename: `u32`,
+    typedecl: ``,
+    assignment: `if uniform_cond {
+      return;
+    } else {
+      x = nonuniform_value[0];
+    }`,
+    cond: `x > 0`,
+    uniform: `never`,
+  },
```
jrprice, line 559: Suggest maybe adding a case that does something like this (if not already covered somewhere that I missed):
```
if uniform_cond {
  x = nonuniform_value[0];
  return;
}
```
This tests that the non-uniform assignment doesn't make it out of the if-else because of the return statement, so `x` would still be uniform afterwards.

jrprice (approved):
LGTM, with one suggestion for an additional test.

src/webgpu/shader/validation/uniformity/uniformity.spec.ts
alan-baker: Added in the latest commit.

jrprice: Thanks!

jrprice (approved):

PR #2765, opened 2023-06-14
Update Expiration cts for GPUExternalTexture imported from VideoFrame

shaoboyan:
According to WebGPU spec, GPUExternalTexture from VideoFrame expires when, and only when, the source VideoFrame is closed (https://www.w3.org/TR/webgpu/#external-texture-creation).

Update related cts to follow the spec

Issue: #

<hr>

### DIFF
--- a/src/webgpu/api/validation/gpu_external_texture_expiration.spec.ts
+++ b/src/webgpu/api/validation/gpu_external_texture_expiration.spec.ts
@@ -155,7 +155,8 @@ g.test('import_and_use_in_different_task')
   .desc(
     `
     Tests that in the different task scope, previous imported GPUExternalTexture
-    should be expired.
+    should be expired if it is imported from HTMLVideoElment. GPUExternalTexture
+    imported from WebCodec VideoFrame is not expired.
     `
   )
   .params(u =>
@@ -187,8 +188,10 @@ g.test('import_and_use_in_different_task')
     });
 
     await waitForNextTask(() => {
-      // Enter in another task scope, previous GPUExternalTexture should be expired.
-      t.submitCommandBuffer(bindGroup, false);
+      // Enter in another task scope. For GPUExternalTexture imported from WebCodec,
+      // it shouldn't be expired because VideoFrame is not 'closed'.
+      // For GPUExternalTexutre imported from HTMLVideoElement, it should be expired.
+      t.submitCommandBuffer(bindGroup, sourceType === 'VideoFrame' ? true : false);
     });
   });
 
@@ -196,26 +199,18 @@ g.test('use_import_to_refresh')
   .desc(
     `
     Tests that in the different task scope, imported GPUExternalTexture
-    again on the same source frame should return the same GPUExternalTexture
+    again on the same HTMLVideoElement should return the same GPUExternalTexture
     object and refresh it.
     `
   )
-  .params(u =>
-    u //
-      .combine('sourceType', ['VideoElement', 'VideoFrame'] as const)
-  )
   .fn(async t => {
-    const sourceType = t.params.sourceType;
     const videoElement = t.getDefaultVideoElementAndCheck();
 
     let bindGroup: GPUBindGroup;
     let externalTexture: GPUExternalTexture;
     let source: HTMLVideoElement | VideoFrame;
-    await startPlayingAndWaitForVideo(videoElement, async () => {
-      source =
-        sourceType === 'VideoFrame'
-          ? await getVideoFrameFromVideoElement(t, videoElement)
-          : videoElement;
+    await startPlayingAndWaitForVideo(videoElement, () => {
+      source = videoElement;
       externalTexture = t.device.importExternalTexture({
         /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
         source: source as any,
--- a/src/webgpu/api/validation/gpu_external_texture_expiration.spec.ts
+++ b/src/webgpu/api/validation/gpu_external_texture_expiration.spec.ts
@@ -232,12 +232,12 @@ g.test('use_import_to_refresh')
     });
 
     await waitForNextTask(() => {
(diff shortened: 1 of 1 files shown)


shaoboyan:
The CTS changes should fail current chrome canary but should pass with https://chromium-review.googlesource.com/c/chromium/src/+/4613007 this change

src/webgpu/api/validation/gpu_external_texture_expiration.spec.ts
```diff
@@ -187,35 +188,29 @@ g.test('import_and_use_in_different_task')
...
     });
   });
 
 g.test('use_import_to_refresh')
   .desc(
     `
     Tests that in the different task scope, imported GPUExternalTexture
-    again on the same source frame should return the same GPUExternalTexture
+    again on the same HTMLVideoElement should return the same GPUExternalTexture
     object and refresh it.
```
kainino0x, line 203: Why isn't this test valid for VideoFrame? Even though auto-expiry doesn't happen when importing from VideoFrame, I think the spec still allows importExternalTexture to return the same GPUExternalTexture again.

That said, in no case does the spec actually guarantee that the same GPUExternalTexture will be returned twice. The browser is always allowed to return a new one. So I think this test is stricter than the spec (though perhaps this is OK until someone complains about it).

shaoboyan: >That said, in no case does the spec actually guarantee that the same GPUExternalTexture will be returned twice. 

That's true. I think I should move the stricter case to chromium content test and release the condition of this cts. That says, the test should test GPUExternalTexture returned by import is active in any task scope.

src/webgpu/api/validation/gpu_external_texture_expiration.spec.ts
```diff
@@ -230,15 +232,21 @@ g.test('use_import_to_refresh')
     });
 
     await waitForNextTask(() => {
-      // Video frame is not updated, import should return the same GPUExternalTexture object.
-      const shouldBeTheSameExternalTexture = t.device.importExternalTexture({
+      const maydBeTheSameExternalTexture = t.device.importExternalTexture({
```
kainino0x, line 235:
```suggestion
      const mayBeTheSameExternalTexture = t.device.importExternalTexture({
```

src/webgpu/api/validation/gpu_external_texture_expiration.spec.ts
```diff
@@ -230,15 +232,21 @@ g.test('use_import_to_refresh')
...
         /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
         source: source as any,
       });
-      assert(externalTexture === shouldBeTheSameExternalTexture);
 
-      // ImportExternalTexture should refresh expired GPUExternalTexture.
-      t.submitCommandBuffer(bindGroup, true);
+      if (externalTexture === maydBeTheSameExternalTexture) {
```
kainino0x, line 240:
```suggestion
      if (externalTexture === mayBeTheSameExternalTexture) {
```

src/webgpu/api/validation/gpu_external_texture_expiration.spec.ts
```diff
@@ -187,35 +196,28 @@ g.test('import_and_use_in_different_task')
...
     Tests that in the different task scope, imported GPUExternalTexture
-    again on the same source frame should return the same GPUExternalTexture
-    object and refresh it.
+    again on the same HTMLVideoElement should return active GPUExternalTexture.
     `
   )
-  .params(u =>
-    u //
-      .combine('sourceType', ['VideoElement', 'VideoFrame'] as const)
-  )
```
kainino0x, line 206: I think it makes sense to bring this back now?

src/webgpu/api/validation/gpu_external_texture_expiration.spec.ts
shaoboyan: Oops... Thanks for pointing it out...

src/webgpu/api/validation/gpu_external_texture_expiration.spec.ts
```diff
@@ -187,35 +196,28 @@ g.test('import_and_use_in_different_task')
     });
 
     await waitForNextTask(() => {
-      // Enter in another task scope, previous GPUExternalTexture should be expired.
-      t.submitCommandBuffer(bindGroup, false);
+      // Enter in another task scope. For GPUExternalTexture imported from WebCodec,
+      // it shouldn't be expired because VideoFrame is not 'closed'.
+      // For GPUExternalTexutre imported from HTMLVideoElement, it should be expired.
+      t.submitCommandBuffer(bindGroup, sourceType === 'VideoFrame' ? true : false);
```
shaoboyan, line 202: [1]

src/webgpu/api/validation/gpu_external_texture_expiration.spec.ts
shaoboyan: A bit hesitate on this.  
First, Refresh seems specific to `GPUExternalTexture` from  `HTMLVideoElement`. Second, the test case  "import_and_use_in_different_task" ([1], https://github.com/gpuweb/cts/pull/2765/files#diff-8e3cd7410d080fa936b8cbe9c85289ca6de0548752bcf72432096d0ed8f5d9b0R162) has already covered the case that `GPUExternalTexture` from `VideoFrame` is valid in other task scope if source frame is not closed

kainino0x: I was thinking we need a test that could trigger caching for VideoFrame, where it could return the same GPUExternalTexture twice for the same VideoFrame. But it looks like that's covered by `import_multiple_times_in_same_task_scope`. The two tests are slightly different but between `import_multiple_times_in_same_task_scope` and `import_and_use_in_different_task` I think the coverage is complete. Thanks for explaining!

kainino0x (approved):

PR #2767, opened 2023-06-14
Adding 0x80000000/0x8000 cases to `ceil`, `fract`, `floor` & `round`

### LINKED ISSUE #2766: Test that result of `fract()` is between 0 to 1
It was reported in https://github.com/gpuweb/gpuweb/issues/4144 that for large numbers `fract()` appears to be returning values outside of the [specified range of `[0 .. 1]`](https://www.w3.org/TR/WGSL/#fract-builtin).

[I've speculated](https://github.com/gpuweb/gpuweb/issues/4144#issuecomment-1566784622) that some drivers are implementing `fract` as `x - f32(i32(x))`, which will fail when `x` is larger than `0x7fffffff`.

We should be testing for the result of `fract` being between 0 and 1 here:
https://github.com/gpuweb/cts/blob/e957d8fd868b0793b5e9f14fd523f436ac3f761c/src/webgpu/util/floating_point.ts#L3264-L3282

zoddicus:
This should detect if an implementation is doing something like fract(x) = x - floor(i32(x)), or floor(x) = i32(x), etc, which
will fail to conform to the spec.

Fixes #2766

<hr>

### DIFF
--- a/src/unittests/floating_point.spec.ts
+++ b/src/unittests/floating_point.spec.ts
@@ -2690,10 +2690,12 @@ const kFloorIntervalCases = {
   f32: [
     { input: 2 ** 30, expected: 2 ** 30 },
     { input: -(2 ** 30), expected: -(2 ** 30) },
+    { input: 0x80000000, expected: 0x80000000 }, // https://github.com/gpuweb/cts/issues/2766
   ],
   f16: [
-    { input: 2 ** 15, expected: 2 ** 15 },
-    { input: -(2 ** 15), expected: -(2 ** 15) },
+    { input: 2 ** 14, expected: 2 ** 14 },
+    { input: -(2 ** 14), expected: -(2 ** 14) },
+    { input: 0x8000, expected: 0x8000 }, // https://github.com/gpuweb/cts/issues/2766
   ],
 } as const;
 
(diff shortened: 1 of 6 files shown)


src/unittests/floating_point.spec.ts
```diff
@@ -2690,10 +2690,12 @@ const kFloorIntervalCases = {
   f32: [
     { input: 2 ** 30, expected: 2 ** 30 },
     { input: -(2 ** 30), expected: -(2 ** 30) },
+    { input: 0x80000000, expected: 0x80000000 }, // https://github.com/gpuweb/cts/issues/2766
   ],
   f16: [
-    { input: 2 ** 15, expected: 2 ** 15 },
```
zoddicus, line 2695: Changed these to avoid duplicated case issues with 0x8000

jzm-intel (approved):
LGTM

jzm-intel:
Shall we also add these cases to `ceil` and `round`?

zoddicus:
Added cases to `ceil` and `round` since they may experience similar issues, depending on the implementation.

jzm-intel:
Thanks, still LGTM!

PR #2770, opened 2023-06-16
Maintain test timing metadata in-repo and auto-chunk WPT

### LINKED ISSUE #2749: Chunk tests for WPT using actual test time measurements
WPT wants tests to stay under about 5s. By estimating:
- the overhead of switching test cases / subcases
- the average time-per-case of each test

we can generate a list of WPT tests (either as "variants" or as individual HTML files) where each one takes a pretty consistent amount of time, yet retains the hierarchical structure.

The basic machinery to do timing and chunking exists already, so what we need is basically just a way to capture test timings in bulk and turn them into chunkings.

Open questions:
- Make this part of the roll process so each browser has to generate their own data?
- Just have a static table of this info in the repo?
- How to keep it up to date? Maybe OK if it doesn't stay perfectly up to date?
- Are runtimes across browsers and platforms consistent enough that we can semi-automatically maintain this? Probably set up tooling such that chunking always gets finer over time and never gets coarser without manual intervention? Unlikely to get so fine  through this process that it's problematic.

kainino0x:
Commits may be reviewed separately.

This adds the following:
- `src/webgpu/listing_metadata.json` which contains estimated times-per-subcase for (almost) every test in the CTS
- Tools for maintaining that file
- A new output at `out-wpt/cts-chunked2sec.https.html` which is chunked such that each variant takes under 2 seconds. (The old cts.https.html remains untouched - verified.)
- A few related things that I needed to do those (see first few commits).

**Important notes:**
- I'll be unavailable after today (Sep 8). Revert this if anything breaks (and isn't trivial to fix). (Note to importers - if it does get reverted, it might be possible to pin to the revision before it was reverted if you need this functionality now.)
- Presubmit requires this metadata file to be kept up-to-date. This will be a little annoying to test authors, but hopefully not too bad. (I'll make sure it still passes on `main` after this lands.)
- The metadata file is thus a complete listing of all of the tests in the repository. It can be used for other purposes too.
- This will generate two redundant `.https.html` files in `out-wpt`, depending on which one the importer wants to use. They shouldn't both be run. Hopefully this is apparent to importers.

Fixes #2749

<hr>

- [ ] ~Tests are properly located in the test tree.~
- [ ] ~[Test descriptions](https://github.com/gpuweb/cts/blob/main/docs/intro/plans.md) allow a reader to "read only the test plans and evaluate coverage completeness", and accurately reflect the test code.~
- [ ] ~Tests provide complete coverage (including validation control cases). **Missing coverage MUST be covered by TODOs.**~


kainino0x:
FYI @litherum @ErichDonGubler. Please see the "important notes" above.

src/common/internal/tree.ts
```diff
@@ -185,16 +205,45 @@ export class TestTree {
...
         subtree.subtreeCounts.nodesWithTODO += counts.nodesWithTODO;
+        subtree.subtreeCounts.totalTimeMS += counts.totalTimeMS;
+        subtree.subcaseCount += counts.subcaseCount;
+      } else {
+        subtree.subcaseCount = child.subcaseCount;
       }
     }
-    return subtree.subtreeCounts;
+
+    if (chunking && subtree.query instanceof TestQueryMultiCase) {
```
austinEng, line 226: sanity check: chunking will mean we do/don't split at the test level. so in the end you could a mix of any:
 - TestQuerySingleCase meaning we broke up all the cases like `s:f:t:foo,abc`, `s:f:t:foo,def`
 - TestQueryMultiCase where it is all cases for a test like `s:f:t:*`
 - TestQueryMultiCase where some of the cases are split - like `s:f:t:foo,*` and `s:f:t:bar,*`
 - mix of single/multi like `s:f:t:foo,abc`, `s:f:t:foo,def`, `s:f:t:bar,*`, `s:f:t:bar,*`
 
but it'll never merge up to `s:f:t1,*` or `s:f:*` ?

kainino0x:
FYI also @ben-clayton as upcoming webgpu gardener for chrome

src/common/tools/crawl.ts
```diff
@@ -88,6 +106,55 @@ export async function crawl(suiteDir: string, validate: boolean): Promise<TestSu
...
+    }
+
+    const missingEntries = [];
+    for (const metadataKey of validateTimingsEntries.testsFoundInFiles) {
+      if (!(metadataKey in validateTimingsEntries.metadata)) {
+        missingEntries.push(metadataKey);
+      }
+    }
+    if (missingEntries.length) {
+      console.log(
```
austinEng, line 147: console.error?

src/common/tools/crawl.ts
```diff
@@ -88,6 +106,55 @@ export async function crawl(suiteDir: string, validate: boolean): Promise<TestSu
...
+      }
+      if (metadataValue.subcaseMS <= 0) {
+        zeroEntries.push(metadataKey);
+      }
+      if (!validateTimingsEntries.testsFoundInFiles.has(metadataKey)) {
+        staleEntries.push(metadataKey);
+      }
+    }
+    if (zeroEntries.length) {
+      console.log('WARNING: subcaseMS≤0 found in listing_meta.json (allowed, but try to avoid):');
```
austinEng, line 127: console.warn?

src/common/tools/merge_listing_times.ts
```diff
@@ -0,0 +1,177 @@
...
+          caseTimes.set(parsed.q, parsed);
+        }
+      } else {
+        caseTimes.set(parsed.q, parsed);
+      }
+    }
+  }
+
+  // Accumulate total times per test. Map of suite -> query -> {totalTimeMS, caseCount}.
+  const testTimes = new Map<string, Map<string, { totalTimeMS: number; subcaseCount: number }>>();
```
austinEng, line 113: doesn't the query have the suite in it? why do we need suite -> query ?

you can still check below `if (!suites.includes(castQ.suite)) continue;`

src/common/tools/merge_listing_times.ts
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  for (const suite of suites) {
+    const currentMetadata: TestMetadataListing = JSON.parse(
+      fs.readFileSync(`./src/${suite}/listing_meta.json`, 'utf8')
+    );
+
+    const metadata = { ...currentMetadata };
+    for (const [testQString, { totalTimeMS, subcaseCount }] of testTimes.get(suite)!) {
+      const aveTime = totalTimeMS / Math.max(1, subcaseCount);
```
austinEng, line 148: uuunit: I think `avg` is the typical abbreviation for average

austinEng (approved):
mostly minor comments

src/common/internal/logging/test_case_recorder.ts
```diff
@@ -19,7 +19,8 @@ const kMinSeverityForStack = LogSeverity.Warn;
 
 /** Holds onto a LiveTestCaseResult owned by the Logger, and writes the results into it. */
 export class TestCaseRecorder {
-  private result: LiveTestCaseResult;
+  readonly result: LiveTestCaseResult;
+  public unskippedSubcaseCount: number = 0;
```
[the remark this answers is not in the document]
kainino0x: notSkippedSubcaseCount is kinda hard to parse so I went with nonskippedSubcaseCount

src/common/internal/tree.ts
kainino0x: correct.

    // - TestQueryMultiTest and higher nodes are never set to collapsible anyway, so we ignore them.
    // - TestQuerySingleCase nodes can't be collapsed, so we ignore them.

src/common/tools/merge_listing_times.ts
kainino0x: It was just to optimize that but that's a silly optimization because we basically never use more than one suite. done

kainino0x: Hmm actually it's not quite so simple. The loop after that, which updates the metadata, has to sort the results into which metadata file they go in, which requires knowing the suite from just a query string. It's slightly annoying to parse the suite out so I'm just going to leave this as is...

greggman:
Should this not error if there's no web socket?

For example if I go here

https://gpuweb.github.io/cts/standalone/?runnow=1&q=webgpu:api,operation,adapter,requestAdapter:requestAdapter_no_parameters:

I get

> websocket_logger.ts:27 WebSocket connection to 'ws://localhost:59497/optional_cts_websocket_logger' failed: 
(anonymous) @ websocket_logger.ts:27
logToWebsocket @ websocket_logger.ts:21
run @ test_group.ts:705

PR #2773, opened 2023-06-20
Test async pipeline creation before/after device destroy

### LINKED ISSUE #2265: val: state,device_lost,destroy:{createRenderPipelineAsync, createComputePipelineAsync}
Expand these validation tests to test performing async pipeline compilation after the device has been lost.

austinEng:
Fixes #2265

<hr>

### DIFF
--- a/src/webgpu/api/validation/state/device_lost/destroy.spec.ts
+++ b/src/webgpu/api/validation/state/device_lost/destroy.spec.ts
@@ -482,18 +482,11 @@ Tests creating compute pipeline asynchronously on destroyed device.
       t.expect(lostInfo.reason === 'destroyed');
     }
 
-    // After device destroy, expect a GPUPipelineError with "validation" reason.
-    t.shouldReject(
-      'GPUPipelineError',
+    // After device destroy, creation should still resolve successfully.
+    t.shouldResolve(
       (async () => {
-        try {
-          await fn();
-        } catch (ex) {
-          if (ex instanceof GPUPipelineError) {
-            t.expect(ex.reason === 'validation');
-          }
-          throw ex;
-        }
+        const pipeline = await fn();
+        t.expect(pipeline instanceof GPUComputePipeline);
       })()
     );
   });
@@ -532,18 +525,11 @@ Tests creating render pipeline asynchronously on destroyed device.
       t.expect(lostInfo.reason === 'destroyed');
     }
 
-    // After device destroy, expect a GPUPipelineError with "validation" reason.
-    t.shouldReject(
-      'GPUPipelineError',
+    // After device destroy, creation should still resolve successfully.
+    t.shouldResolve(
       (async () => {
-        try {
-          await fn();
-        } catch (ex) {
-          if (ex instanceof GPUPipelineError) {
-            t.expect(ex.reason === 'validation');
-          }
-          throw ex;
-        }
+        const pipeline = await fn();
+        t.expect(pipeline instanceof GPURenderPipeline);
       })()
     );
   });
--- a/src/webgpu/api/validation/state/device_lost/destroy.spec.ts
+++ b/src/webgpu/api/validation/state/device_lost/destroy.spec.ts
@@ -5,6 +5,7 @@ Tests for device lost induced via destroy.
 `;
 
 import { makeTestGroup } from '../../../../../common/framework/test_group.js';
+import { assert } from '../../../../../common/util/util.js';
 import {
   allBindingEntries,
   bindingTypeInfo,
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/state/device_lost/destroy.spec.ts
```diff
@@ -454,6 +454,156 @@ Tests creating render pipeline on destroyed device.
...
+
+    // Destroy the device, and expect it to be lost.
+    t.expectDeviceLost('destroyed');
+    t.device.destroy();
+    if (awaitLost) {
+      const lostInfo = await t.device.lost;
+      t.expect(lostInfo.reason === 'destroyed');
+    }
+
+    // After device destroy, expect a GPUPipelineError with "validation" reason.
```
lokokung, line 485: @kainino0x Want to confirm with you whether this is the intended behavior? W.r.t the other APIs, I think that this should actually just return a pipeline? If so, maybe the `during` variants can be combined.

kainino0x: The intent is that it returns a pipeline, not rejects. Although reading the spec closely I think the spec is actually wrong and says it'll reject. Reopened https://github.com/gpuweb/gpuweb/issues/1629#issuecomment-1599988786

austinEng: I've updated these tests to now expect success for all the destroyed cases

src/webgpu/api/validation/state/device_lost/destroy.spec.ts
```diff
@@ -454,6 +454,142 @@ Tests creating render pipeline on destroyed device.
     }, awaitLost);
   });
 
+g.test('createComputePipelineAsync')
```
lokokung, line 457: WDYT about merging the variants with a `awaitLost` param?

austinEng: Merged.

I also ended up turning `awaitLost` into a param not a subcase param. I realized that as a subcase param, it was reusing the device between subcases! which meant that the lost device was already lost before .destroy()'ed in some subcases

austinEng: That is - overall, anything that destroys the device shouldn't use subcases, I think.

austinEng:
@kainino0x is validation still supposed to run on lost devices?

I should probably add test cases that do:
1. destroy the device
2. async pipeline creation with *invalid* WGSL source
3. expect validation error? or "success" ?

src/webgpu/api/validation/state/device_lost/destroy.spec.ts
lokokung: Oh great catch!

lokokung (approved):
Thanks Austin!

kainino0x:
> @kainino0x is validation still supposed to run on lost devices?

No, it should never be observable whether validation is running or not on a lost device. https://gpuweb.github.io/gpuweb/#errors-and-debugging

kainino0x:
and https://github.com/gpuweb/gpuweb/issues/1629#issuecomment-1495166485

austinEng:
> > @kainino0x is validation still supposed to run on lost devices?
> 
> No, it should never be observable whether validation is running or not on a lost device. https://gpuweb.github.io/gpuweb/#errors-and-debugging

Ok, @kainino0x PTAL again

I'm not sure how we should handle invalid creation while losing the device though.
I believe this coudl be a race since the validation could happen before or after the device was destroyed.
If it happened before, you see the error. If it happened after, you don't.
Right now, I have the test written assuming that validation occurs on the same timeline as device.destroy() (not on some thread), so if you do something invalid before device.destroy(), you get the error.

kainino0x:
I think it's a race, yeah. But maybe we can rely on the ordering of the `lost` promise and the createPipelineAsync promise matching the result you see. explained more here: https://github.com/gpuweb/gpuweb/issues/1629#issuecomment-1603468868
and possible test code:

```
let isLost = false;
const lostP = device.lost.then(() => { isLost = true; });

let createSucceeded;
const createP = createPipelineAsync().then(() => {
  assert(isLost);
}, () => {
  assert(!isLost);
});

device.destroy();
await Promise.all([lostP, createP]);
```

wdyt?
Unfortunately probably only one of these paths will actually be hit by chrome.

kainino0x:
oh and i haven't actually reviewed yet so apologies if this is close to what you have. Will review momentarily

austinEng:
@kainino0x updated to handle the validation ordering with the device lost promise

kainino0x (approved):

kainino0x:
assuming you've gotten the test to pass with the chrome patch, lgtm

PR #2781, opened 2023-06-22
Compat: Skip bgra8uorm-srgb in test

greggman:
bgra8unorm-srgb is not supported in compat

I thought about putting `skipIfTextureFormatNotSupported` in `GPUTestSubcaseBatchState` instead of `GPUTestBase` as then it could be called in `beforeAllSubcases` but it's not useful there because subcase type info is not available in `beforeAllSubcases`

<hr>

### DIFF
--- a/src/webgpu/api/operation/command_buffer/copyTextureToTexture.spec.ts
+++ b/src/webgpu/api/operation/command_buffer/copyTextureToTexture.spec.ts
@@ -75,6 +75,8 @@ class F extends GPUTest {
     srcCopyLevel: number,
     dstCopyLevel: number
   ): void {
+    this.skipIfTextureFormatNotSupported(srcFormat, dstFormat);
+
     const mipLevelCount = dimension === '1d' ? 1 : 4;
 
     // Create srcTexture and dstTexture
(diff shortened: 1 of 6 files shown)


kainino0x (approved):
> I thought about putting `skipIfTextureFormatNotSupported` in `GPUTestSubcaseBatchState` instead of `GPUTestBase` as then it could be called in `beforeAllSubcases` but it's not useful there because subcase type info is not available in `beforeAllSubcases`

It's possible to skip from beforeAllSubcases in the cases where the format _is_ available in beforeAllSubcases - which is true for most tests, either because we need to call `selectDeviceOrSkipTestCase` or simply because we wanted to be able to easily suppress tests by format (back when we had a few core formats which were unimplemented). It's not a big deal, I think the amount of wasted runtime on skipped subcases will be quite small. But I do think the logic for some of these tests would be a bit more straightforward if we call `skipIfTextureFormatNotSupported` before `selectDeviceOrSkipTestCase`, or maybe even combine the two into a single call.

kainino0x:
OK to land if it unblocks you, we can talk about whether to do those changes as followup

greggman:
I updated it, ptal

kainino0x (commented):
lgtm!

PR #2795, opened 2023-07-01
Fix image_copy tests

greggman:
I'm not sure what it was doing before. I refactored `updateLinearTextureDataSubBox`.

1. It was calling `iterateBlockRows` and passing in `sourceOrigin` at which point `texel` is in sourceOrigin coordinates.

   But `texel` was passed to `getTexelOffsetInBytes` for the destination and the code just substracted `destinationOrigin` from `texel` which makes no sense

   I changed it so texel is just a relative coord (0 to size) and `getTexelOffsetInBytes` adds the origin passed in.

2. I found it extremely hard to follow what was happening at `updateLinearTextureDataSubBox` call sites so I refactored so it takes a `src` and `dest` params so that hopefully it's much easier to read at the call site.

This seems to have uncovered at least 2 implementation bugs

1. Mac M1, format `rgb9e5ufloat` does not always round trip

   Filed a [chrome bug](https://bugs.chromium.org/p/dawn/issues/detail?id=1895) after spending the afternoon wondering if I was crazy and writing lots of diagnostics and a standalone repo

2. On Windows, dimension='1d' fails for webgpu:api,operation,command_buffer,image_copy:offsets_and_sizes:initMethod="WriteTexture";checkMethod="PartialCopyT2B

   Given these passes on M1 Mac I'm assming 🤞 that it's an bug in WebGPU, not in the test but I have no had time to dig into it.

Also ran the these test on Intel Mac, all of them passed.

Issue: #

<hr>


greggman:
turns out th rgb9e5 issue is alternate representations of the same values. I'll update the test and use the functions Ben is addi g here https://github.com/gpuweb/cts/pull/2796

src/webgpu/api/operation/command_buffer/image_copy.spec.ts
```diff
@@ -390,7 +513,14 @@ class ImageCopyTest extends GPUTest {
       bufferData,
       GPUBufferUsage.COPY_SRC | GPUBufferUsage.COPY_DST
     );
-
+    // At this point both buffer and bufferData have the same random data in
+    // them. We'll use copyTextureToBuffer to up buffer with data from the
```
austinEng, line 517:
```suggestion
    // them. We'll use copyTextureToBuffer to update buffer with data from the
```

austinEng (approved):
sorry for the delay; took a while to parse all the padding

Issue #2671, opened 2023-05-25
wgsl: f32 validation tests for `atanh`


zoddicus:
Need coverage that when abs(x) >= 1, it is rejected for const-eval (shader creation error) and override (pipeline creation error).

Issue #2665, opened 2023-05-24
wgsl: f32 validation tests for `asin`


zoddicus:
Need coverage that when abs(x) > 1 or x == 0, it is rejected for const-eval (shader creation error) and override (pipeline creation error).

Issue #2663, opened 2023-05-24
wgsl: f16 validation tests for `acosh`


zoddicus:
Need coverage that when x < 1, it is rejected for const-eval (shader creation error) and override (pipeline creation error).

Issue #2660, opened 2023-05-24
wgsl: f16 validation tests for `acos`


zoddicus:
Need coverage that when abs(x) > 1 or x == 0, it is rejected for const-eval (shader creation error) and override (pipeline creation error).

PR #2799, opened 2023-07-04
shader/validation: Validate trig builtins

ben-clayton:
Fixes: #2658
Fixes: #2659
Fixes: #2660
Fixes: #2661
Fixes: #2662
Fixes: #2663
Fixes: #2664
Fixes: #2665
Fixes: #2666
Fixes: #2670
Fixes: #2671
Fixes: #2672

<hr>


src/webgpu/shader/validation/expression/call/builtin/trig.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+
+function runTest(
+  t: ShaderValidationTest,
+  builtin: string,
+  expectedResult: boolean,
+  value: number,
+  type: string,
+  stage: string
+) {
+  const enables = requiresShaderF16(type) ? 'enable f16' : '';
```
jrprice, line 41: Enable directives need a terminating semicolon.

src/webgpu/shader/validation/expression/call/builtin/trig.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+
+/// An array of values ranging from -1 to 2
+const kMinusOneToTwo = [-1.0, -0.9, -0.1, 0.0, 0.1, 0.5, 0.9, 1.0, 1.1, 1.5, 1.9, 2.0] as const;
+
+/// The evaluation stages to test
+const kEvaluationStages = ['constant', 'override'] as const;
+
+/// All floating-point scalar and vector types to test
+const kFloatingPointTypes = [
+  '',
```
jrprice, line 18: IIUC this is the AbstractFloat case? Presumably this means we aren't testing vectors of AbstractFloats here, but maybe that's OK?

jrprice (commented):
Should we be testing that `AbstractInt` implicitly converts to `AbstractFloat` for these builtins here, or will that be covered elsewhere?

Likewise, should we be testing that stuff like `acos(i32(42))` is rejected?

src/webgpu/shader/validation/expression/call/builtin/trig.spec.ts
ben-clayton: Added abstract vector types to this

src/webgpu/shader/validation/expression/call/builtin/trig.spec.ts
ben-clayton: Done

ben-clayton:
> Should we be testing that `AbstractInt` implicitly converts to `AbstractFloat` for these builtins here, or will that be covered elsewhere?

I think this is already covered. JS will omit the `.0` suffix if the number is an integer, which we have cases for.

> Likewise, should we be testing that stuff like `acos(i32(42))` is rejected?

A fair question. But how far do you go with these negative tests? Should we check for 2 arguments? etc.

I can followup with negative tests for bad argument types, if we want to go down this path.

ben-clayton:
Added tests to ensure integer arguments are rejected, and split up the `trig.spec.ts` file into one per builtin.

jrprice (approved):

PR #2812, opened 2023-07-05
Fix texel view for signed values

greggman:
The code was pulling out unsigned values and then complaining when value was > max for a signed value. Example: rg8int with a unsigned value of 131 would fail because it pulled out the value unsigned instead of signed.

<hr>

### DIFF
--- a/src/webgpu/util/texture/texel_data.ts
+++ b/src/webgpu/util/texture/texel_data.ts
@@ -453,6 +453,10 @@ function makeIntegerInfo(
   opt: { signed: boolean }
 ): TexelRepresentationInfo {
   assert(bitLength <= 32);
+  const numericRange = opt.signed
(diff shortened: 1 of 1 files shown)


austinEng (approved):

greggman:
I went through several iterations (you can see the pushes)

1. tried to change `unpackComponentsBits` to take a sign. It seemed to work but seemed wrong as AFAICT `unpackComponentBits` is supposed to return uninterpreted bits.

2. tried to change `decode` to use the sign. It worked for my use case but other code was calling `decode` with signed values.

3. final, in `makeIntegerInfo` `decode` is used for `bitsToNumber`. So made `bitsToNumber` its own function which converts from unsigned to signed. This seems to work and doesn't break the other path

Issue #1690, opened 2022-07-20
wgsl: validate function restrictions


dj2:
* [x] `vertex` shader with `position` output value
* [x] `vertex` shader without `position` output fails
* [x] function calling an `entry_point` is invalid
* [x] function with `constructible return type
* [x] function with non-`constructible` return type is invalid
* [x] function param with each of `constructible types`, `pointer`, `texture`, `sampler` types
* [x] function param with other types is invalid
* [x] function call with value of wrong type fails
* [x] function call to `ptr` with differing address space, or store type, or access mode fails
* [x] function with `ptr` in `function`, `private`~~, `workgroup`~~ address space 
* [x] function with `ptr` in other address spaces fails
* [ ] ~~function with `ptr` param with an `atomic` fails~~
* [ ] ~~function with `ptr` param with nested `atomic` fails~~
* [ ] ~~call to builtin with `ptr` in correct address space~~
* [ ] ~~call to builtin with `ptr` of wrong address space fails~~
* [x] call user function with address-of works
* [x] call user function with `ptr` param of another function works
* [x] call user function with `address-of` of non-identifier fails
* [x] Recursive calls fail `a -> b -> a`

PR #2817, opened 2023-07-06
Function restriction validation tests

alan-baker:
Contributes to #1690

* Tests:
  * vertex must return position
  * entry points cannot be the target of a call
  * function return types
  * function parameter types
  * function call argument type matching
  * no recursion

Issue: #1690
<hr>

### DIFF
--- a/src/webgpu/shader/validation/functions/restrictions.spec.ts
+++ b/src/webgpu/shader/validation/functions/restrictions.spec.ts
@@ -1,0 +1,591 @@
+export const description = `Validation tests for function restrictions`;
+
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { keysOf } from '../../../../common/util/data_tables.js';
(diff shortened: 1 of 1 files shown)


src/webgpu/shader/validation/functions/restrictions.spec.ts
```diff
@@ -0,0 +1,591 @@
...
+${t.params.stage}
+fn bar() -> ${ret_attr} vec4f {
+  return vec4f();
+}
+
+fn foo() {
+  let tmp = bar();
+}
+`;
+    const without_entry_point = `
```
dj2, line 66: Would this be better making the with/without entry point a param that is passed in, then the same shader ends up being used for both the pass/fail cases?

alan-baker: sure, I'll make that change.

dj2 (approved):

Issue #1283, opened 2022-04-26
wgsl: execution test for atomicCompareExchangeWeak builtin


dj2:
```
fn atomicCompareExchangeWeak(atomic_ptr: ptr<SC, atomic<T>, A>, cmp: T, v: T) -> __atomic_compare_exchange_result<T>

struct __atomic_compare_exchange_result<T> {
  old_value : T; // old value stored in the atomic
  exchanged : bool; // true if the exchange was done
}
```

Performs the following steps atomically:

 1. Load the original value pointed to by atomic_ptr.
 1. Compare the original value to the value v using an equality operation.
 1. Store the value v only if the result of the equality comparison was true.

Returns a two member structure, where the first member, old_value, is the original value of the atomic object and the second member, exchanged, is whether or not the comparison succeeded.

Issue #1282, opened 2022-04-26
wgsl: execution test for atomicExchange builtin


dj2:
```
fn atomicExchange(atomic_ptr: ptr<SC, atomic<T>, A>, v: T) -> T
```

Atomically stores the value v in the atomic object pointed to atomic_ptr and returns the original value stored in the atomic object.

See `atomicExchange` in https://www.w3.org/TR/WGSL/#atomic-rmw

Issue #1274, opened 2022-04-26
wgsl: execution test for atomicStore builtin


dj2:
See https://www.w3.org/TR/WGSL/#atomic-store

Issue #1273, opened 2022-04-26
wgsl: execution test for atomicLoad builtin


dj2:
See https://www.w3.org/TR/WGSL/#atomic-load

PR #2819, opened 2023-07-06
Tests for atomic builtins: atomicLoad, atomicStore, atomicExchange, a…

amaiorano:
…tomicCompareExchangeWeak

Issue: #1273
Issue: #1274
Issue: #1282
Issue: #1283

<hr>

### DIFF
--- a/src/webgpu/shader/execution/expression/call/builtin/atomics/atomicCompareExchangeWeak.spec.ts
+++ b/src/webgpu/shader/execution/expression/call/builtin/atomics/atomicCompareExchangeWeak.spec.ts
@@ -14,11 +14,21 @@ component of the result vector equals cmp.
 `;
 
 import { makeTestGroup } from '../../../../../../../common/framework/test_group.js';
+import { keysOf } from '../../../../../../../common/util/data_tables.js';
+import { assert } from '../../../../../../../common/util/util.js';
 import { GPUTest } from '../../../../../../gpu_test.js';
+
+import {
+  dispatchSizes,
+  workgroupSizes,
+  typedArrayCtor,
+  kMapId,
+  onlyWorkgroupSizes,
+} from './harness.js';
 
 export const g = makeTestGroup(GPUTest);
 
-g.test('exchange')
+g.test('compare_exchange_weak_storage_basic')
   .specURL('https://www.w3.org/TR/WGSL/#atomic-rmw')
   .desc(
     `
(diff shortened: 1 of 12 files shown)


src/webgpu/shader/execution/expression/call/builtin/atomics/atomicCompareExchangeWeak.spec.ts
```diff
@@ -34,6 +37,264 @@ struct __atomic_compare_exchange_result<T> {
...
+    pass.end();
+    t.queue.submit([encoder.finish()]);
+
+    // Output buffer should be the same as the initial input buffer as it contains
+    // values returned from atomicCompareExchangeWeak
+    const outputExpected = new (typedArrayCtor(t.params.scalarKind))(bufferNumElements);
+    outputExpected.forEach((_, i) => (outputExpected[i] = i));
+    t.expectGPUBufferValuesEqual(outputBuffer, outputExpected);
+
+    // Exchanged buffer should be 'true' for every third value
```
alan-baker, line 144: atomicCompareExchangeWeak can fail spuriously. 
>The equality comparison may spuriously fail on some implementations. That is, the second component of the result vector may be false even if the first component of the result vector equals cmp.

So you can't really require exchanged to be true. You could require unexchanged values to definitely be false. You should probably also check the output based on the value of exchanged at that index.

src/webgpu/shader/execution/expression/call/builtin/atomics/atomicCompareExchangeWeak.spec.ts
```diff
@@ -34,6 +37,264 @@ struct __atomic_compare_exchange_result<T> {
...
+      fn main(
+          @builtin(local_invocation_index) local_invocation_index: u32,
+          @builtin(workgroup_id) workgroup_id : vec3<u32>
+          ) {
+        let id = ${scalarKind}(local_invocation_index);
+        let global_id = ${scalarKind}(workgroup_id.x * ${wgNumElements} + local_invocation_index);
+
+        // Initialize wg[id] with this invocations global id
+        atomicStore(&wg[id], global_id);
+        workgroupBarrier();
```
alan-baker, line 211: Strictly speaking, I think this barrier is unnecessary. Each invocation operates on a separate piece of data. Program ordering should ensure there is no reordering.

src/webgpu/shader/execution/expression/call/builtin/atomics/atomicCompareExchangeWeak.spec.ts
```diff
@@ -34,6 +37,264 @@ struct __atomic_compare_exchange_result<T> {
 `
   )
   .params(u =>
-    u.combine('SC', ['storage', 'uniform'] as const).combine('T', ['i32', 'u32'] as const)
+    u
+      .combine('workgroupSize', workgroupSizes)
+      .combine('dispatchSize', dispatchSizes)
+      .combine('mapId', keysOf(kMapId))
+      .combine('scalarKind', ['u32', 'i32'])
```
ben-clayton, line 44: nit: `scalarType`

src/webgpu/shader/execution/expression/call/builtin/atomics/atomicCompareExchangeWeak.spec.ts
amaiorano: Does this spurious failure have anything to do with multiple writes to the same address? I know the spec doesn't say so, but am wondering if we can reasonably assume no spurious failures if we only have a single write to a given location.

src/webgpu/shader/execution/expression/call/builtin/atomics/atomicCompareExchangeWeak.spec.ts
amaiorano: Right, this is a vestige from other code I wrote where only invocation with `local_invocation_index == 0` would initialize all values. Will remove.

src/webgpu/shader/execution/expression/call/builtin/atomics/atomicCompareExchangeWeak.spec.ts
ben-clayton: Not unless the spec is changed to state this.

ben-clayton (commented):
Only just started looking. Will review properly when I'm back at my machine.

src/webgpu/shader/execution/expression/call/builtin/atomics/atomicCompareExchangeWeak.spec.ts
alan-baker: I don't have a good answer for this. Probably you won't see failures, but it wouldn't be incorrect if you did. Really you should loop on the compare exchange, but then you have problem from a lack of forward progress guarantees.

amaiorano: Fair enough, I'll fix this up to only expect the exchanged value if the function returned true that it did.

amaiorano: Done.

src/webgpu/shader/execution/expression/call/builtin/atomics/atomicCompareExchangeWeak.spec.ts
amaiorano: Removed.

src/webgpu/shader/execution/expression/call/builtin/atomics/atomicCompareExchangeWeak.spec.ts
amaiorano: I went ahead and renamed. I think it's alright for this change.

src/webgpu/shader/execution/expression/call/builtin/atomics/atomicCompareExchangeWeak.spec.ts
```diff
@@ -34,6 +37,273 @@ struct __atomic_compare_exchange_result<T> {
...
+          @builtin(global_invocation_id) global_invocation_id : vec3<u32>,
+          ) {
+        let id = ${scalarType}(global_invocation_id[0]);
+
+        // Exchange every third value
+        var comp = id + 1;
+        if (id % 3 == 0) {
+          comp = id;
+        }
+        let r = atomicCompareExchangeWeak(&input[id], comp, map_id(id * 2));
```
ben-clayton, line 75: While this is a good basic test for checking the comparison behaviour - the one-invocation to one-memory-location aspect is not really exercising the atomic RMW properties of this builtin.

For example, you could replace this line with:

```ts

        let r = atomicLoad(&wg[id]);
        let eq = r == comp;
        if (eq) {
          atomicStore(&wg[id], map_id(global_id * 2));
        }
```

And with minor tweaks to the usage of `r`, the test will still pass.

I recommend adding another test for multiple-invocations to one-memory-location.

src/webgpu/shader/execution/expression/call/builtin/atomics/atomicExchange.spec.ts
```diff
@@ -18,6 +21,201 @@ fn atomicExchange(atomic_ptr: ptr<AS, atomic<T>, read_write>, v: T) -> T
...
+      @group(0) @binding(1)
+      var<storage, read_write> output : array<${scalarType}>;
+
+      @compute @workgroup_size(${t.params.workgroupSize})
+      fn main(
+          @builtin(global_invocation_id) global_invocation_id : vec3<u32>,
+          ) {
+        let id = ${scalarType}(global_invocation_id[0]);
+
+        output[id] = atomicExchange(&input[id], map_id(id * 2));
```
ben-clayton, line 51: Similar to `atomicCompareExchangeWeak`, the 1:1 invocation to memory location isn't testing the RMW property of `atomicExchange`. You could substitute this line with:

```ts
output[id] = atomicLoad(&input[id]);
// input[id] could change between the line above and below
atomicStore(&input[id], map_id(id * 2));
```

And still pass the test.

I suggest adding another test where:
* all the invocations `atomicExchange` to the same single memory address `A`, using the hash of the invocation IDs
* the old values returned by `atomicExchange` are atomically appended to another storage buffer `B`.
* you can then check the results by ensuring the final value held by `A` combined with the unordered values in `B` match the set of invocation ID hashes.

src/webgpu/shader/execution/expression/call/builtin/atomics/atomicStore.spec.ts
```diff
@@ -18,6 +28,73 @@ fn atomicStore(atomic_ptr: ptr<AS, atomic<T>, read_write>, v: T)
...
   )
-  .unimplemented();
+  .fn(t => {
+    const numInvocations = t.params.workgroupSize * t.params.dispatchSize;
+    const bufferNumElements = numInvocations;
+    const mapId = kMapId[t.params.mapId];
+    const extra = mapId.wgsl(numInvocations, t.params.scalarType); // Defines map_id()
+
+    const initValue = 0;
+    const op = `atomicStore(&output[id], map_id(id))`;
```
ben-clayton, line 44: Similar comment here. Nothing really atomic about this, when there's a 1:1 invocation to memory location.
Suggest multiple invocations writing to the same location with different values, then checking that the written value was one of the values. That would at least demonstrate there was no tearing in the written word.

dneto0:
@reeselevine pushed the state of the art with the memory model tests in https://github.com/gpuweb/cts/tree/main/src/webgpu/shader/execution/memory_model 

It doesn't exercise all the atomic operations. It does check atomicLoad, atomicStore, atomicExchange, atomicAdd.

It's based on the research in https://reeselevine.github.io/assets/pdf/mc_mutants.pdf

So.... we have to pick how far to push things.  Are we trying to have a good chance of observing memory consistency failures?  That's what the research was about.   

Or is this more about "these functions work in a basic way".  That's a gap left to be filled.

amaiorano:
> Or is this more about "these functions work in a basic way". That's a gap left to be filled.

That's right. We need basic execution tests for each builtin. Complex tests that use builtins are not sufficient on their own.

src/webgpu/shader/execution/expression/call/builtin/atomics/atomicExchange.spec.ts
amaiorano: Done. Added a "_basic" suffix to the original tests I wrote, and added "_advanced" ones for these you recommended.

src/webgpu/shader/execution/expression/call/builtin/atomics/atomicStore.spec.ts
amaiorano: Did as you suggested. Added two more atomicStore "_advanced" tests that read/write to the same atomic value.

src/webgpu/shader/execution/expression/call/builtin/atomics/atomicCompareExchangeWeak.spec.ts
```diff
@@ -34,6 +177,337 @@ struct __atomic_compare_exchange_result<T> {
...
+
+    // Number of times each workgroup attempts to write the same value to the same memory address
+    const numWrites = 4;
+
+    const bufferNumElements = numInvocations * numWrites;
+    const defaultValue = 99999999;
+
+    const wgsl = `
+      @group(0) @binding(0)
+      var<storage, read_write> input : array<atomic<${scalarType}>>;
```
ben-clayton, line 345: No need for the array if there's only ever one element.

src/webgpu/shader/execution/expression/call/builtin/atomics/atomicCompareExchangeWeak.spec.ts
```diff
@@ -34,6 +177,337 @@ struct __atomic_compare_exchange_result<T> {
...
+
+        // Each invocation attempts to write 'compare + 1' if 'compare' is at input[0], once per loop.
+        // Only one invocation per loop iteration should succeed. Note the workgroupBarrier() used
+        // to synchronize each invocation in the loop.
+
+        var compare = ${scalarType}(0);
+        for (var i = 0u; i < ${numWrites}u; i++) {
+          // Only one invocation will succeed at writing its value,
+          // or none if spurious failures occur.
+          let r = atomicCompareExchangeWeak(&input[0], compare, compare + 1);
```
ben-clayton, line 367: We discussed bouncing between two comparison values, so that a single spurious failure doesn't break the chain, and instead the passes after a spurious failure can resynchronise. This seems to be incrementing?

src/webgpu/shader/execution/expression/call/builtin/atomics/atomicCompareExchangeWeak.spec.ts
```diff
@@ -34,6 +177,337 @@ struct __atomic_compare_exchange_result<T> {
...
+
+          let slot = i * ${numInvocations}u + u32(id);
```
ben-clayton, line 369: Can use an array-of-arrays to simplify this.

src/webgpu/shader/execution/expression/call/builtin/atomics/atomicCompareExchangeWeak.spec.ts
```diff
@@ -34,6 +177,337 @@ struct __atomic_compare_exchange_result<T> {
...
+      };
+
+      // Only one of the invocations should have succeeded to exchange - or none if a spurious failure occured
+      if (exchanged.every(v => v === 0)) {
+        // Spurious failure, all values in output be the default value
+        if (!output.every(v => v === defaultValue)) {
+          t.fail(
+            `Spurious failure detected, expected only default value of ${defaultValue} in output buffer.${dumpValues()}`
+          );
+          return;
```
ben-clayton, line 480: One spurious failure and we bail. Can we do better?

src/webgpu/shader/execution/expression/call/builtin/atomics/atomicCompareExchangeWeak.spec.ts
```diff
@@ -34,6 +177,337 @@ struct __atomic_compare_exchange_result<T> {
...
+          t.fail(`More than one invocation exchanged its value.${dumpValues()}`);
+          return;
+        }
+
+        // Get its index
+        const idx = exchanged.findIndex(v => v === 1);
+        assert(idx !== -1);
+
+        // Output should contain the old value after exchange
+        if (output[idx] !== nextExpectedValue) {
```
ben-clayton, line 494: All invocations should have the same old value, no?

src/webgpu/shader/execution/expression/call/builtin/atomics/atomicCompareExchangeWeak.spec.ts
```diff
@@ -34,6 +177,337 @@ struct __atomic_compare_exchange_result<T> {
...
+  )
+  .fn(async t => {
+    const numInvocations = t.params.workgroupSize * t.params.dispatchSize;
+    const scalarType = t.params.scalarType;
+
+    // Number of times each workgroup attempts to write the same value to the same memory address
+    const numWrites = 4;
+
+    const bufferNumElements = numInvocations * numWrites;
+    const defaultValue = 99999999;
```
ben-clayton, line 341: Recommend declaring these after the WGSL, so a reader doesn't look for their usages in the string.

src/webgpu/shader/execution/expression/call/builtin/atomics/atomicCompareExchangeWeak.spec.ts
```diff
@@ -34,6 +177,337 @@ struct __atomic_compare_exchange_result<T> {
...
+
+    let nextExpectedValue = 0;
+    for (let w = 0; w < numWrites; ++w) {
+      const offset = w * numInvocations;
+      const exchanged = exchangedBufferResult.subarray(offset, offset + numInvocations);
+      const output = outputBufferResult.subarray(offset, offset + numInvocations);
+
+      const dumpValues = () => {
+        return `\n\tFor write: ${w}
+        \texchanged: ${exchanged}
```
ben-clayton, line 469: nit: I'd recommend avoiding tabs. Especially as you're indenting these two lines with whitespace.

src/webgpu/shader/execution/expression/call/builtin/atomics/atomicCompareExchangeWeak.spec.ts
amaiorano: Done.

src/webgpu/shader/execution/expression/call/builtin/atomics/atomicCompareExchangeWeak.spec.ts
amaiorano: Done.

src/webgpu/shader/execution/expression/call/builtin/atomics/atomicCompareExchangeWeak.spec.ts
amaiorano: Yeah, I just prefer to keep it like this as it matches the `subarray` analog in JS/TS.

src/webgpu/shader/execution/expression/call/builtin/atomics/atomicCompareExchangeWeak.spec.ts
amaiorano: To clarify, this code isn't failing the test if a spurious failure is detected. This is failing if the output buffer was written to. If no value was exchanged at all, then the output buffer should contain the default initial value.

src/webgpu/shader/execution/expression/call/builtin/atomics/atomicCompareExchangeWeak.spec.ts
amaiorano: Done.

src/webgpu/shader/execution/expression/call/builtin/atomics/atomicCompareExchangeWeak.spec.ts
amaiorano: Done.

src/webgpu/shader/execution/expression/call/builtin/atomics/atomicCompareExchangeWeak.spec.ts
amaiorano: It didn't, as this version was only updating the output buffer for the invocation that succeeded at exchanging. But I've since updated the code so that we always write the returned value to output, and test that it contains what we expect (that is: the old value for the invocation that exchanged, else the old or new value for all other invocations).

amaiorano:
@ben-clayton I added the workgroup test. PTAL, thanks!

ben-clayton (approved):
Approving. Made a bunch of suggestions for renaming things, but logic looks good.
Thanks for the changes!

src/webgpu/shader/execution/expression/call/builtin/atomics/atomicCompareExchangeWeak.spec.ts
```diff
@@ -34,6 +331,412 @@ struct __atomic_compare_exchange_result<T> {
...
+
+    // Number of times each workgroup attempts to exchange the same value to the same memory address
+    const numWrites = 4;
+
+    const bufferNumElements = numInvocations * numWrites;
+    const pingPongValues = [24, 68];
+
+    const wgsl = `
+      @group(0) @binding(0)
+      var<storage, read_write> input : atomic<${scalarType}>;
```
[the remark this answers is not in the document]
amaiorano: Done.

src/webgpu/shader/execution/expression/call/builtin/atomics/atomicCompareExchangeWeak.spec.ts
```diff
@@ -34,6 +331,412 @@ struct __atomic_compare_exchange_result<T> {
...
+
+      const dumpValues = () => {
+        return `
+        For write: ${w}
+        exchanged: ${exchanged}
+        output:    ${output}`;
+      };
+
+      // Only one of the invocations should have succeeded to exchange - or none if spurious failures occured
+      const allSpuriouslyFailed = exchanged.every(v => v === 0);
```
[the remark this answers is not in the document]
amaiorano: Done.

src/webgpu/shader/execution/expression/call/builtin/atomics/atomicExchange.spec.ts
```diff
@@ -18,6 +22,449 @@ fn atomicExchange(atomic_ptr: ptr<AS, atomic<T>, read_write>, v: T) -> T
...
+    const numInvocations = t.params.workgroupSize * t.params.dispatchSize;
+    const bufferNumElements = numInvocations;
+    const scalarType = t.params.scalarType;
+    const mapId = kMapId[t.params.mapId];
+    const extra = mapId.wgsl(numInvocations, t.params.scalarType); // Defines map_id()
+
+    const wgsl =
+      `
+      @group(0) @binding(0)
+      var<storage, read_write> input : array<atomic<${scalarType}>>;
```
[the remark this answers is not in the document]
amaiorano: Done.

src/webgpu/shader/execution/expression/call/builtin/atomics/atomicCompareExchangeWeak.spec.ts
```diff
@@ -34,6 +331,412 @@ struct __atomic_compare_exchange_result<T> {
...
+
+    const bufferNumElements = numInvocations * numWrites;
+    const pingPongValues = [24, 68];
+
+    const wgsl = `
+      @group(0) @binding(0)
+      var<storage, read_write> input : atomic<${scalarType}>;
+
+      @group(0) @binding(1)
+      var<storage, read_write> output : array<${scalarType}>;
```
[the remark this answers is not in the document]
amaiorano: Done.

src/webgpu/shader/execution/expression/call/builtin/atomics/atomicExchange.spec.ts
```diff
@@ -18,6 +22,449 @@ fn atomicExchange(atomic_ptr: ptr<AS, atomic<T>, read_write>, v: T) -> T
...
```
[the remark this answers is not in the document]
amaiorano: Done.

Issue #2716, opened 2023-05-25
wgsl: f32 validation tests for `length`


zoddicus:
Need coverage that when result or intermediaries go OOB, it is rejected for const-eval (shader creation error) and override (pipeline creation error).

PR #2821, opened 2023-07-06
shader/validation: Validate length()

ben-clayton:
I'll followup with a non-splatted version.

Fixed: #2715
Fixed: #2716
Fixed: #2717

<hr>

### DIFF
--- a/src/webgpu/shader/validation/expression/call/builtin/length.spec.ts
+++ b/src/webgpu/shader/validation/expression/call/builtin/length.spec.ts
@@ -1,0 +1,178 @@
+const builtin = 'length';
+export const description = `
+Validation tests for the ${builtin}() builtin.
+`;
+
+import { makeTestGroup } from '../../../../../../common/framework/test_group.js';
+import {
+  ScalarType,
+  TypeF16,
+  TypeF32,
+  elementType,
+  kAllFloatScalars,
+  kAllFloatVector2,
+  kAllFloatVector3,
+  kAllFloatVector4,
+  kAllIntegerScalarsAndVectors,
+} from '../../../../../util/conversion.js';
+import { isRepresentable } from '../../../../../util/floating_point.js';
+import { ShaderValidationTest } from '../../../shader_validation_test.js';
+
+import {
+  fullRangeForType,
+  kConstantAndOverrideStages,
+  stageSupportsType,
+  validateConstOrOverrideBuiltinEval,
+} from './const_override_validation.js';
+
+export const g = makeTestGroup(ShaderValidationTest);
+
+/** @returns true if the sum of the squares of each element in @p vec is representable by @p type */
+function isSquareSumRepresentable(vec: number[], type: ScalarType) {
+  const squareSum = vec.reduce((prev, curr) => prev + curr * curr, 0);
+  return isRepresentable(squareSum, type);
+}
+
+g.test('scalar')
+  .desc(
+    `
+Validates that constant evaluation and override evaluation of ${builtin}() with
+the input scalar value always compiles without error
+`
+  )
+  .params(u =>
+    u
+      .combine('stage', kConstantAndOverrideStages)
+      .combine('type', kAllFloatScalars)
+      .filter(u => stageSupportsType(u.stage, u.type))
+      .expand('value', u => fullRangeForType(u.type))
+  )
+  .beforeAllSubcases(t => {
+    if (elementType(t.params.type) === TypeF16) {
+      t.selectDeviceOrSkipTestCase('shader-f16');
+    }
+  })
+  .fn(t => {
+    // We only validate with numbers known to be representable by the type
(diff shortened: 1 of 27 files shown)


src/webgpu/shader/validation/expression/call/builtin/length.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+import { makeTestGroup } from '../../../../../../common/framework/test_group.js';
+import {
+  TypeF16,
+  TypeF32,
+  VectorType,
+  elementType,
+  kAllFloatScalarsAndVectors,
+  kAllIntegerScalarsAndVectors,
+} from '../../../../../util/conversion.js';
+import { fpTraitsFor, isRepresentable } from '../../../../../util/floating_point.js';
```
dj2, line 15: isRepresentable is never used

src/webgpu/shader/validation/expression/call/builtin/length.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+
+import {
+  kConstantAndOverrideStages,
+  stageSupportsType,
+  validateConstOrOverrideBuiltinEval,
+} from './const_override_validation.js';
+
+export const g = makeTestGroup(ShaderValidationTest);
+
+g.test('splat')
```
dj2, line 27: Are non-splat cases coming? Like using `-val, val, -val` where it can't splat?

ben-clayton: Yeah, will follow up with non-splat cases. Requires more harness work.
I'll also add cases that are guaranteed to fail, so we have some negative tests.

dj2 (approved):

src/webgpu/shader/validation/expression/call/builtin/length.spec.ts
```diff
@@ -0,0 +1,178 @@
...
+`
+  )
+  .params(u =>
+    u
+      .combine('stage', kConstantAndOverrideStages)
+      .combine('type', kAllFloatVector2)
+      .filter(u => stageSupportsType(u.stage, u.type))
+      .expand('x', u => fullRangeForType(u.type, 5))
+      .expand('y', u => fullRangeForType(u.type, 5))
+      .filter(u => isSquareSumRepresentable([u.x, u.y], elementType(u.type)))
```
dneto0, line 80: Instead of filtering, check the expected overflow

expectedResult = isSquareSumRepresentable(. ...);

dneto0 (changes requested):

dneto0 (changes requested):
Something to consider.
I'm not sure, but it looks like the same value is replicated across all elements.

So an implementation could be broken in the sense that it's only sensitive to the overflow in the first vector element, instead of all of them.

It would be better to sweep the value across all elements and let the other components be 0.

src/webgpu/shader/validation/expression/call/builtin/length.spec.ts
ben-clayton: That's depending on an implementation detail. Internally the built-in could be evaluated with a higher precision than the data type.

ben-clayton:
> Something to consider. I'm not sure, but it looks like the same value is replicated across all elements.
> 
> So an implementation could be broken in the sense that it's only sensitive to the overflow in the first vector element, instead of all of them.
> 
> It would be better to sweep the value across all elements and let the other components be 0.

No. The implementation sweeps each of the components individually.

src/webgpu/shader/validation/expression/call/builtin/length.spec.ts
```diff
@@ -0,0 +1,178 @@
...
+      t.selectDeviceOrSkipTestCase('shader-f16');
+    }
+  })
+  .fn(t => {
+    const expectedResult = true;
+    validateConstOrOverrideBuiltinEval(
+      t,
+      builtin,
+      expectedResult,
+      [t.params.type.create([t.params.x, t.params.y, t.params.z])],
```
ben-clayton, line 125: Observe that we're constructing the argument from 3 parameterised values

src/webgpu/shader/validation/expression/call/builtin/length.spec.ts
ben-clayton: Kept the filter for the reason stated above, but also included negative tests where `isRepresentable(intermediate) == isRepresentable(result)`

src/webgpu/shader/validation/expression/call/builtin/length.spec.ts
ben-clayton: Done

src/webgpu/shader/validation/expression/call/builtin/length.spec.ts
ben-clayton: Done

Issue #2832, opened 2023-07-07
spelling of pointer types 


dneto0:
This is the pointer part of the original #1538.

The spec has been reorganized so now the relevant section is:
https://gpuweb.github.io/gpuweb/wgsl/#ref-ptr-types

> Pointer types may appear.

> A pointer type is spelled with parameterization by:

>*  [address space](https://gpuweb.github.io/gpuweb/wgsl/#address-spaces),

>*  [store type](https://gpuweb.github.io/gpuweb/wgsl/#store-type), and

> * sometimes by [access mode](https://gpuweb.github.io/gpuweb/wgsl/#access-mode), as specified in [§ 13.3 Address Spaces](https://gpuweb.github.io/gpuweb/wgsl/#address-space).

> If a pointer type appears in the program source, it [must](https://gpuweb.github.io/gpuweb/wgsl/#shader-creation-error) also be valid to [declare a variable](https://gpuweb.github.io/gpuweb/wgsl/#variable-declaration), somewhere in the program, with the pointer type’s [address space](https://gpuweb.github.io/gpuweb/wgsl/#address-spaces), [store type](https://gpuweb.github.io/gpuweb/wgsl/#store-type), and [access mode](https://gpuweb.github.io/gpuweb/wgsl/#access-mode).

For  pointer types:

* [ ] Storage address space no address mode (defaults to `read`)
* [ ] Storage address space with `read` mode accepted
* [ ] Storage address space with `read_write` mode accepted
* [ ] Storage address space with `write` mode rejected
* [ ] Other address spaces {`function`, `private`, `workgroup`, `uniform`, `storage`}.  Access mode must not be written.

dneto0:
Pointer types also appear as formal parameters to user declared functions.
Other tasks touching on pointers to user declared functions
- #1926
- #1589 
- #1427

Issue #1589, opened 2022-06-29
wgsl: for user-defined function, param of pointer-to-workgroup must not have store type of atomic type or containing an atomic type


dneto0:
Spec change is https://github.com/gpuweb/gpuweb/pull/3110

dneto0:
Fixed by #2840

Issue #1427, opened 2022-05-17
forbid spelling pointer types that are not instantiable


dneto0:
From spec change https://github.com/gpuweb/gpuweb/pull/2904 

E.g. the following shaders must produce shader-creation error
```rust
type p = ptr<workgroup,array<i32>>;
```

```rust
fn foo( p: ptr<workgroup,array<i32>> ) { }
```

dneto0:
Fixed by #2840

PR #2840, opened 2023-07-10
Issue 2832 validate pointer types spelled in WGSL source

dneto0:
Issue: #2832 #1427 #1589

<hr>

### DIFF
--- a/src/webgpu/shader/validation/decl/ptr_spelling.spec.ts
+++ b/src/webgpu/shader/validation/decl/ptr_spelling.spec.ts
@@ -1,0 +1,79 @@
+export const description = `
+Validate spelling of pointer types.
+
+Pointer types may appear.
+
+They are parameterized by:
+- address space, always
+- store type
+- and access mode, as specified by the table in Address Spaces.
+   Concretely, only 'storage' address space allows it, and allows 'read', and 'read_write'.
+
+A pointer type can be spelled only if it corresponds to a variable that could be
+declared in the program.  So we need to test combinations against possible variable
+declarations.
+`;
+
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { keysOf } from '../../../../common/util/data_tables.js';
+import {
+  AddressSpace,
+  kAccessModeInfo,
+  kAddressSpaceInfo,
+} from '../../types.js';
+import { ShaderValidationTest } from '../shader_validation_test.js';
+
+import {
+  pointerType,
+  infoExpander,
+  explicitSpaceExpander,
+  accessModeExpander,
+  getVarDeclShader,
+  ShaderStage,
+} from './util.js';
+
+// Address spaces that can hold an i32 variable.
+const kNonHandleAddressSpaces = keysOf(kAddressSpaceInfo).filter(
+  as => as !== 'handle'
+) as AddressSpace[];
+
+export const g = makeTestGroup(ShaderValidationTest);
+
+g.test('param_ptr_type')
+  .desc('Validate formal parameter of pointer type on user-declared functions')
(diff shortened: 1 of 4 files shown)


src/webgpu/shader/validation/decl/util.ts
```diff
@@ -40,3 +48,128 @@ fn ${arg.name}() {
...
+ * @returns a WGSL var declaration with given parameters for variable 'x' and
+ * store type i32.
+ */
+export function declareVarX(addressSpace: AddressSpace | '', accessMode: AccessMode | ''): string {
+  const parts: string[] = [];
+  if (addressSpace && kAddressSpaceInfo[addressSpace].binding) parts.push('@group(0) @binding(0) ');
+  parts.push('var');
+
+  const template_parts: string[] = [];
+  if (addressSpace) template_parts.push(addressSpace as string);
```
ben-clayton, line 62: I think you can drop the `as string` here

src/webgpu/shader/validation/decl/ptr_spelling.spec.ts
```diff
@@ -0,0 +1,157 @@
...
+
+g.test('let_ptr_explicit_type_matches_var')
+  .desc(
+    'Let-declared pointer with explicit type initialized from var with same address space and access mode'
+  )
+  .specURL('https://w3.org/TR#ref-ptr-types')
+  .params(u =>
+    u // Generate non-handle variables in all valid permutations of address space and access mode.
+      .combine('addressSpace', kNonHandleAddressSpaces)
+      .expand('info', infoExpander)
```
ben-clayton, line 52: Maybe I'm misunderstanding things, but it looks like you're using the test parameters as a scratch location for putting intermediate state used by other `expand()` and `combine()` transforms.

Parameters should be just that - things that appear in the case query. There's a limit to the length of the query string, I think our CTS runner will fail at 250-ish characters. You should also consider that these are things that people type, so we want to avoid redundant information.

As far as I understand, `info` is immutable data fetched by looking up `kAddressSpaceInfo` with `addressSpace`. Given that this deterministically derived from `addressSpace`, I imagine you can push this lookup into `explicitSpaceExpander` and `accessModeExpander`, and drop `info` from the parameters.

I do believe that parameters with a leading underscore are excluded from the query, but I have no idea how efficient it is to do what you're doing.

src/webgpu/shader/validation/decl/ptr_spelling.spec.ts
```diff
@@ -0,0 +1,157 @@
...
+      .expand('explicitSpace', explicitSpaceExpander)
+      .combine('explicitMode', [false, true])
```
ben-clayton, line 54: nit: I had to look up what `explicitMode` meant. I think `explicitAccess` might be clearer.

ben-clayton (approved):
Approving, but I think you might need to fix `.expand('info', infoExpander)` - this might break CTS rolls.

src/webgpu/shader/validation/decl/ptr_spelling.spec.ts
dneto0: Fixed now.  Thanks.  Yep, it was dodgy.

src/webgpu/shader/validation/decl/ptr_spelling.spec.ts
dneto0: Fixed.

src/webgpu/shader/validation/decl/util.ts
dneto0: Fixed.

ben-clayton (approved):
Thank you for all the changes!

PR #2843, opened 2023-07-10
Fix unpackRGB9E5UFloat usage and add tests

greggman:
Fix TexelView for rgb9e5uflroat

The issue was `unpackRGB9e5ufloat` was being called in `unpackBits`
but `unpackRGB9e5ufloat` doesn't return bits, it returns "number".
The code in `TexelView`

https://github.com/gpuweb/cts/blob/7f06106f08266ecd6178f66c14eb54f6dd94dbc8/src/webgpu/util/texture/texel_view.ts#L103

calls `bitsToNumber` with the results of `unpackBits` and `bitsToNumber` expects bits

Added a few tests. And made NaN match NaN. The tests often
put in pseudo random binary data which might be NaN and expect
NaN to come back out but the tests fail if comparing NaN to NaN
returns false.

<hr>

### DIFF
--- a/src/webgpu/util/conversion.ts
+++ b/src/webgpu/util/conversion.ts
@@ -131,6 +131,8 @@ type FloatFormat = { signed: 0 | 1; exponentBits: number; mantissaBits: number; 
 export const kFloat32Format = { signed: 1, exponentBits: 8, mantissaBits: 23, bias: 127 } as const;
 /** FloatFormat defining IEEE754 16-bit float. */
 export const kFloat16Format = { signed: 1, exponentBits: 5, mantissaBits: 10, bias: 15 } as const;
+/** FloatFormat for 9 bit mantissa, 5 bit exponent unsigned float */
+export const kUFloat9e5Format = { signed: 0, exponentBits: 5, mantissaBits: 9, bias: 15 } as const;
 
 /**
  * Once-allocated ArrayBuffer/views to avoid overhead of allocation when converting between numeric formats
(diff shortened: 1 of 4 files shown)


src/webgpu/util/texture/texel_data.ts
```diff
@@ -775,24 +775,32 @@ export const kTexelRepresentationInfo: {
             components.B ?? unreachable()
           ),
         ]).buffer,
-      // For the purpose of unpacking, expand into three "ufloat14" values.
       unpackBits: (data: Uint8Array) => {
-        return unpackRGB9E5UFloat((data[3] << 24) | (data[2] << 16) | (data[1] << 8) | data[0]);
```
kainino0x, line 780: I think this was the only usage of this function, so maybe it should be deleted.

While checking that, I accidentally found that this is actually brand new in #2796. You should sync with @ben-clayton about it.

src/webgpu/util/conversion.ts
```diff
@@ -179,6 +181,22 @@ export function floatBitsToNumber(bits: number, fmt: FloatFormat): number {
...
 }
 
+/**
+ * Convert ufloat9e5 bits to a JS number
+ *
+ * The difference between `floatBitsToNumber` and `ufloatBitsToNumber`
+ * is
+ *
+ * floatBitsToNumber  = 2^(exponent - bias) * (1 + mantissa / 2 ^ numMantissaBits)
+ * ufloatBitsToNumber = 2^(exponent - bias - numMantissaBits) * mantissa
```
kainino0x, line 191: This more clearly shows the difference between the two
```suggestion
 * The difference between `floatBitsToNumber` and `ufloatBitsToNumber`
 * is that the latter doesn't use an implicit leading bit:
 *
 * floatBitsToNumber  = 2^(exponent - bias) * (1 + mantissa / 2 ^ numMantissaBits)
 * ufloatBitsToNumber = 2^(exponent - bias) * (mantissa / 2 ^ numMantissaBits)
 *                    = 2^(exponent - bias - numMantissaBits) * mantissa
```

But more importantly I believe this is different from the other ufloat format, rg11b10ufloat. So the name of this function should be more specific. Since this format is quite special it probably would make sense to just specialize it to remove the `fmt` (which doesn't contain enough information to fully decode the number anyway) and rename it to be ufloat9e5-specific.

kainino0x (commented):
LGTM % comments but I'd also like Ben's input

src/webgpu/util/conversion.ts
greggman: I renamed it and added your comment edit. I'm not sure what you meant by "remove the `fmt`"

kainino0x:
BTW please avoid rebase-force-pushes, as AFAICT GitHub can't show me an incremental diff when that happens. Instead use a merge commit, or neither if no update is needed.

src/webgpu/util/texture/texel_data.ts
ben-clayton: I was using it for texture execution tests, which I'm still working on. Please don't remove it.

ben-clayton (approved):
LGTM

PR #2853, opened 2023-07-14
Compat: Test bindgroup usage validation

greggman:
Bindgroups in compat must not use different
views of the same texture.

<hr>

### DIFF
--- a/src/compat/api/validation/encoding/programmable/pipeline_bind_group_compat.spec.ts
+++ b/src/compat/api/validation/encoding/programmable/pipeline_bind_group_compat.spec.ts
@@ -307,8 +307,8 @@ Tests that you can not use 2 different views of the same texture in a render pas
 - Test you can bind the same view in different bindGroups
 - Test binding incompatible bindGroups is ok as long as they are fixed before draw/dispatch
 
-  The last 2 tests are to check validation happens at the correct time (draw/dispatch) and not
-  at finish or setBindGroup.
+  The last test is to check validation happens at the correct time (draw/dispatch) and not
+  at setBindGroup.
     `
   )
   .params(u =>
@@ -370,8 +370,8 @@ Tests that you can not use 2 different views of the same texture in a compute pa
 - Test can bind the same view in different bindGroups
 - Test that binding incompatible bindGroups is ok as long as they are fixed before draw/dispatch
 
-  The last 2 tests are to check validation happens at the correct time (draw/dispatch) and not
-  at finish or setBindGroup.
+  The last test is to check validation happens at the correct time (draw/dispatch) and not
+  at setBindGroup.
     `
   )
   .params(u =>
--- a/src/compat/api/validation/encoding/programmable/pipeline_bind_group_compat.spec.ts
+++ b/src/compat/api/validation/encoding/programmable/pipeline_bind_group_compat.spec.ts
@@ -21,50 +21,62 @@ type BindConfig = kBindConfigs[number];
  * Gets the WGSL needed for testing a render pipeline using texture_2d or texture_storage_2d
  * and either 2 bindgroups or 1
  */
-function getRenderPipelineWGSL(textureType: TextureType, bindConfig: BindConfig) {
+function getRenderShaderModule(
+  device: GPUDevice,
+  textureType: TextureType,
+  bindConfig: BindConfig
+) {
   const textureTypeWGSL = getTextureTypeWGSL(textureType);
   const secondGroup = bindConfig === 'one bindgroup' ? 0 : 1;
   const secondBinding = secondGroup === 0 ? 1 : 0;
-  return `
-  @vertex
-  fn vs(@builtin(vertex_index) VertexIndex : u32) -> @builtin(position) vec4f {
-    var pos = array(
-      vec4f(-1,  3, 0, 1),
-      vec4f( 3, -1, 0, 1),
-      vec4f(-1, -1, 0, 1));
-    return pos[VertexIndex];
-  }
+  return device.createShaderModule({
+    code: `
+      @vertex
+      fn vs(@builtin(vertex_index) VertexIndex : u32) -> @builtin(position) vec4f {
+        var pos = array(
+          vec4f(-1,  3, 0, 1),
+          vec4f( 3, -1, 0, 1),
+          vec4f(-1, -1, 0, 1));
+        return pos[VertexIndex];
+      }
 
-  @group(0) @binding(0) var tex0 : ${textureTypeWGSL};
-  @group(${secondGroup}) @binding(${secondBinding}) var tex1 : ${textureTypeWGSL};
+      @group(0) @binding(0) var tex0 : ${textureTypeWGSL};
+      @group(${secondGroup}) @binding(${secondBinding}) var tex1 : ${textureTypeWGSL};
 
-  @fragment
-  fn fs(@builtin(position) pos: vec4f) -> @location(0) vec4f {
-    _ = tex0;
-    _ = tex1;
-    return vec4f(0);
-  }
-  `;
+      @fragment
+      fn fs(@builtin(position) pos: vec4f) -> @location(0) vec4f {
+        _ = tex0;
+        _ = tex1;
+        return vec4f(0);
+      }
+  `,
+  });
 }
 
 /**
  * Gets the WGSL needed for testing a compute pipeline using texture_2d or texture_storage_2d
  * and either 2 bindgroups or 1
  */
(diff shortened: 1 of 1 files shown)


src/compat/api/validation/encoding/programmable/pipeline_bind_group_compat.spec.ts
```diff
@@ -0,0 +1,373 @@
...
+  fn cs() {
+    _ = tex0;
+    _ = tex1;
+  }
+  `;
+}
+
+type GPUEncoderType = GPURenderPassEncoder | GPUComputePassEncoder | GPURenderBundleEncoder;
+
+const kBindCases: {
```
lokokung, line 72: Ah... some of these feel like they should be their own test case instead of a parameter. Especially `'binding incompatible bindGroups without using'` because that case is being parameterized over draw/dispatch cases even though it skips them resulting in a bunch of duplicated tests.

Might be a bit of duplicated code but you should be able to make a helper to help. Could also get rid of [1] in that case since it was a bit confusing to me at first to have that.

src/compat/api/validation/encoding/programmable/pipeline_bind_group_compat.spec.ts
```diff
@@ -0,0 +1,373 @@
...
+}
+
+type kBindConfigs = ['one bindgroup', 'two bindgroups'];
+type BindConfig = kBindConfigs[number];
+
+/**
+ * Gets the WGSL needed for testing a render pipeline using texture_2d or texture_storage_2d
+ * and either 2 bindgroups or 1
+ */
+function getRenderPipelineWGSL(textureType: TextureType, bindConfig: BindConfig) {
```
lokokung, line 24: Both here and `getComputePipelineWGSL` could consider taking the device also and directly returning the module instead of the WGSL.

src/compat/api/validation/encoding/programmable/pipeline_bind_group_compat.spec.ts
```diff
@@ -0,0 +1,373 @@
...
+import { kRenderEncodeTypes } from '../../../../../webgpu/util/command_buffer_maker.js';
+
+const kTextureTypes = ['regular', 'storage'];
+type TextureType = typeof kTextureTypes[number];
+
+function getTextureTypeWGSL(textureType: TextureType) {
+  return textureType === 'storage' ? 'texture_storage_2d<rgba8unorm, write>' : 'texture_2d<f32>';
+}
+
+type kBindConfigs = ['one bindgroup', 'two bindgroups'];
```
lokokung, line 17: [1]

src/compat/api/validation/encoding/programmable/pipeline_bind_group_compat.spec.ts
```diff
@@ -0,0 +1,373 @@
...
+    `
+  )
+  .params(u =>
+    u
+      .combine('encoderType', kRenderEncodeTypes)
+      .combine('bindCase', kBindCaseNames)
+      .combine('useCase', kDrawCaseNames)
+      .combine('textureType', kTextureTypes)
+      .filter(
+        // storage textures can't have 2 bind groups point to the same
```
lokokung, line 255: Should these actually be a test that expect failure instead of being skipped then? (Especially the compat-specific behavior.)

src/compat/api/validation/encoding/programmable/pipeline_bind_group_compat.spec.ts
greggman: done

src/compat/api/validation/encoding/programmable/pipeline_bind_group_compat.spec.ts
greggman: There is no compat specific behavior for this case. You can't bind a storage texture to two or more bind points. You'll get an "aliasing disallowed" validation error. That error is already tested. The test that checks that you can't alias storage texture bindings will also be run in compat mode.

src/compat/api/validation/encoding/programmable/pipeline_bind_group_compat.spec.ts
greggman: I split the tests. I'm not sure what you mean by "get rid of [1]". It's still needed for the other tests.

lokokung: Ah, sorry that was probably a bit unclear. I was imagining that we would split all of the other ones also. I am just worried that this abstraction might be hiding too much into "parameters" when in reality the tests are really testing fairly different things. Here's some pseudo-code of what I am imagining:

```
g.test('incompatibleViews,sameBindgroup')
  .params(u =>
    u
      .combine('encoderType', kProgrammableEncoderTypes)
      .combine('textureType', kTextureTypes)
      .expand('useCase', p => {
        if (p.encoderType === 'compute pass') {
          return kDispatchCaseNames;
        }
        return kDrawCaseNames;
      }
  .fn( t => {
    // Create encode based on encoderType
    // Create resources based on encoderType
    // Do the equivalent of the fn in the bindCase for incompatibleViews,sameBindgroup
    // Call use case based on encoderType
  });

g.test('incompatibleViews,differentBindgroup')

...
```

Looking at it again though, like you said, maybe [1] can stay as it helps with determining the shader, but I would just hard-code which one is needed per test. Overall my thought process is just that these bindCases are not so much a parameter whereas `encoderType` could be one instead if we want to minimize code duplication. I could be convinced otherwise if you feel strongly though!

src/compat/api/validation/encoding/programmable/pipeline_bind_group_compat.spec.ts
lokokung: Ack.

src/compat/api/validation/encoding/programmable/pipeline_bind_group_compat.spec.ts
greggman: I'm not against changing but I'm not really following what the concrete argument is why separating the tests is better how they are now. Is it your personal preference or is there some guideline I missing as to how tests are supposed to be structured that the current code is not following?

To me, there's just 1 issue being tested. At draw time, are 2 incompatible views drawn/dispatched. The rest is the just the combination of ways you can get there

1. encoder type (renderPass, render bundle, compute pass)
2. one bindgroup, two bindgroups

The other 2 cases are just there to double check compat validation didn't go overboard.

* 2 compatible views

  this should already be tested in the webgpu: suite but I put this here in case it's not

* 2 incompatible views but fixed before draw
 
  this is just to check, like it says in the comments, that validation happens at `draw` and not the moment you bind an incompatible view.

lokokung: Hmm, fair enough! There aren't concrete guidelines (especially about parameterization, IIRC, it just says use your own discretion), so my comment is definitely just my personal opinion. I think my concern just stems from the fact that I haven't really seen other tests parameterize in a similar way, especially not with such a verbose logical parameter. (Most of the parameters I've seen tend to be like a single variable or type, whereas these are almost like a concept?).

I do agree that these are all essentially testing the same thing though, so I feel like that is a good enough argument to keep it as is. We can always update if necessary in the future!

src/compat/api/validation/encoding/programmable/pipeline_bind_group_compat.spec.ts
```diff
@@ -0,0 +1,423 @@
...
+
+  return { texture, pipeline };
+}
+
+export const g = makeTestGroup(ValidationTest);
+
+g.test('twoDifferentTextureViews,render_pass,used')
+  .desc(
+    `
+Tests that you can not use 2 different views of the same texture in a render pass in compat mode..
```
lokokung, line 303: nit. extra `.`

src/compat/api/validation/encoding/programmable/pipeline_bind_group_compat.spec.ts
```diff
@@ -0,0 +1,423 @@
...
+    const { encoder, validateFinish } = t.createEncoder(encoderType);
+    encoder.setPipeline(pipeline);
+    createAndBindTwoBindGroupsWithDifferentViewsOfSameTexture(t.device, pipeline, encoder, texture);
+    validateFinish(true);
+  });
+
+g.test('twoDifferentTextureViews,compute_pass,used')
+  .desc(
+    `
+Tests that you can not use 2 different views of the same texture in a compute pass in compat mode..
```
lokokung, line 366: nit. extra `.`

lokokung (approved):

Issue #2860, opened 2023-07-18
WGSL test parameters are incompatible with CTS harness


greggman:
A bunch of new tests are incompatible with the test harness.

Example:

Go here: [https://gpuweb.github.io/cts/standalone/?q=webgpu:shader,validation,expression,call,builtin,abs:values:stage=%22constant%22;*](https://gpuweb.github.io/cts/standalone/?q=webgpu:shader,validation,expression,call,builtin,abs:values:stage=%22constant%22;*)

Click the `[out]` icon of a test (the icon to the right of the ▶ icon)

For example the first test leads here: [https://gpuweb.github.io/cts/standalone/?q=webgpu:shader,validation,expression,call,builtin,abs:values:stage=%22constant%22;type={%22kind%22:%22abstract-float%22,%22_size%22:8};value=-1.7976931348623157e%2B308](https://gpuweb.github.io/cts/standalone/?q=webgpu:shader,validation,expression,call,builtin,abs:values:stage=%22constant%22;type={%22kind%22:%22abstract-float%22,%22_size%22:8};value=-1.7976931348623157e%2B308)

Which gets an error

> Error: Query `webgpu:shader,validation,expression,call,builtin,abs:values:stage="constant";type={"kind":"abstract-float","_size":8};value=-1.7976931348623157e%2B308` does not match any cases

The issue is the test is parameterized with classes. For example that test effectively does this

```
  params(u =>
    ...
    u.combine('type', [ ScalarType, ...])
    ...
```

But `ScalarType` can't be serialized so when the harness tries to match the query `type={"kind":"abstract-float","_size":8}` it fails because it expects that they're the same type of object but one is `Object` and the other `ScalarType`. Further, if I remove the check for them being the same type of object it expects all of the properties to match but the query only has `kind` and `_size` where as `ScalarType` has also has `read` which is a function

dneto0:
Oof.  Nice catch.
It would be nice if something would have caught this during paramaterization itself.

zoddicus:
Ugh, so this is the whole not being able to JSON-ify thing again (I think). I don't think adding a bunch of logic to json_param_value.ts for scalar and vector types is the appropriate thing to do.

Instead of arrays of objects, the types should be stored in maps with string keys, and the keys can be used for the 
 test params. I believe there is already functionality to automate combine by key instead of whole entry.

zoddicus:
oh, it is just use `keysOf(foo)`, and then index into the map. That should be fine.

[...]

PR #2862, opened 2023-07-19
Reject invalid test params

greggman:
This PR rejects test params who's values are functions or non-plain JavaScript objects.

This affects
  src/unittests/floating_point.spec.ts (toVector, toMatrix)
  src/webgpu/api/validation/buffer/mapping.spec.ts (mapAsync,usage)
  src/webgpu/api/operation/render_pipeline/culling_tests.spec.ts (culling)
  the test mentioned in https://github.com/gpuweb/cts/issues/2860

This validates both case params and subcase params. I thought about not validating subcase params because they are not part of the public interface (they don't appear in URLs nor expectations) but, when debugger it's common to comment out `beginSubcases`. This makes it easy to run and debug individual subcases as cases. But, if subcases are allowed to break these rules then the ability to quickly comment out `beginSubcases` would not be possible.

One issue, is the error message is terse. Ideally it would print out the entire path to the test. As it is, all it prints is the test name (the part inside `g.test('namegoeshere')`)

NOTE: this PR can not be committed until the tests above are refactored

<hr>


zoddicus:
Just to confirm my understanding, these tests will need to be updated to let this PR pass the checks
- `src/unittests/floating_point.spec.ts`
- `src/webgpu/api/validation/buffer/mapping.spec.ts`
- `src/webgpu/api/operation/render_pipeline/culling_tests.spec.ts`

greggman:
correct.

I was about to fix the other 2

kainino0x:
> But, if subcases are allowed to break these rules then the ability to quickly comment out `beginSubcases` would not be possible. We could decide that's okay though and not validate subcases. Thoughts?

I think it's best to be strict here so we can always convert subcases to cases. It's intended to be easy to do that.

> It also wasn't clear if I should allow typedarrays. If I don't then some other tests fail.

I guess if queries work with them, they're technically fine, but in principle I think it's very odd to allow typedarrays as part of the public parameters since they're not representable in JSON. It's fine to allow them (and functions) in private params (those with names starting with `_`).

greggman:
ok, I disallowed typedarrays. There was only one test to fix

Also

* `src/webgpu/api/validation/buffer/mapping.spec.ts` (mapAsync,usage)
* `src/webgpu/api/operation/render_pipeline/culling_tests.spec.ts` (culling)

did not need to be fixed. I just needed to check for null in this PR

greggman:
this is ready for review. Had to rebase to get in the test fixes

zoddicus (approved):

PR #2871, opened 2023-07-24
Massively speed up zero_init workgroup memory tests on Android

toji:
These tests are currently so slow that the Chrome bots time out when running them. It's about 30-40s on a Pixel 6 for each batch of 15.

Profiling with Chrome's about:tracing on one batch to see where the time was going, I got back a trace with fifteen (the test batch size) ~2s long pipeline creation blocks. Worth noting that the Tint translation portion of each block was only about 0.3ms.
test_profile_01

Each test creates a custom pipeline to check the memory zero init state, so at first glance it felt like there wasn't much that could be done here. But by zooming in closer you could see another pipeline creation immediately following the massive one. This one only took ~2ms each time.
test_profile_02

Turns out the smaller of the two pipeline builds is the custom shader for each test. The larger one is from a pipeline that preps for the test by stomping on the entirely of workgroup memory with a known pattern so we can ensure the zeroing behavior is legit. This shader has a loop in it the size of which is determined by the workgroup memory size, and it's not too hard to imagine a Vulkan driver trying to unroll it internally.

Fortunately, this pipeline is the same for every single run, so we can cache it! Doing so cuts the test time down by a couple orders of magnitude! (~2s for the first test batch, then ~300ms for every other batch.)

There's a few more things that could be cached here (like the buffer used to copy the fill value from) but those are already performing well and I'd prefer to add as few "globally" cached values as possible.

Issue: https://bugs.chromium.org/p/dawn/issues/detail?id=1901

<hr>

### DIFF
--- a/src/webgpu/shader/execution/zero_init.spec.ts
+++ b/src/webgpu/shader/execution/zero_init.spec.ts
@@ -423,8 +423,21 @@ g.test('compute,zero_init')
       }
       `;
 
+      const fillLayout = t.device.createBindGroupLayout({
+        entries: [{
+          binding: 0,
+          visibility: GPUShaderStage.COMPUTE,
+          buffer: { type: 'read-only-storage' },
+        }, {
+          binding: 1,
+          visibility: GPUShaderStage.COMPUTE,
+          buffer: { type: 'storage' },
+        }]
+      });
+
       const fillPipeline = t.device.createComputePipeline({
-        layout: 'auto',
+        layout: t.device.createPipelineLayout({ bindGroupLayouts: [fillLayout] }),
+        label: 'Workgroup Fill Pipeline',
         compute: {
           module: t.device.createShaderModule({
             code: wgsl,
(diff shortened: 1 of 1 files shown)


austinEng (approved):
LGTM but we should look into why this isn't hitting a cache in Chrome

lokokung (commented):
Haven't done a full review yet, but my first impression of this is that maybe we could reuse the `SubcaseBatchState` instead of using a global map. Looking at the test, it seems like the big pipeline is only needed in the `addressSpace == 'workgroup'` cases, so maybe that's possible? Nonetheless I am not even remotely certain this is possible...

toji:
Not really familiar with the `SubcaseBatchState` but I was wondering if we had a mechanism for this already. I'll look into that.

toji:
Updated this PR in light of discovering https://bugs.chromium.org/p/dawn/issues/detail?id=1933

Now it simply uses an explicit layout rather than `'auto'`, which should be a less disruptive change overall. I think there's still some benefit to be had from doing some explicit caching on the CTS side as way of helping other implementations and nudging people towards best practices, but it doesn't have to be handled here.

austinEng (approved):
still LGTM! thanks for adding labels too :)

PR #2874, opened 2023-07-25
Compat: Make compat tests skip in non-compat mode

greggman:
This should make it easier to run webgpu: and compat: tests together as the
compat tests will not have to be manually excluded (for example via expectations.txt)
when run in non-compat

<hr>

### DIFF
--- a/src/compat/api/validation/encoding/cmds/copyTextureToBuffer.spec.ts
+++ b/src/compat/api/validation/encoding/cmds/copyTextureToBuffer.spec.ts
@@ -3,14 +3,14 @@ Tests limitations of copyTextureToBuffer in compat mode.
 `;
 
 import { makeTestGroup } from '../../../../../common/internal/test_group.js';
-import { ValidationTest } from '../../../../../webgpu/api/validation/validation_test.js';
 import {
   kCompressedTextureFormats,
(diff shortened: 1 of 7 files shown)


lokokung (approved):
LGTM! Separate question just for my knowledge, IIUC, these tests still "run" (start and then get skipped) when in non-compat mode right? Wondering it it would be possible to make them not even appear somehow? Probably doesn't matter much atm since I assume the number of tests is significantly smaller than the non-compat mode suite.

greggman:
Yes, the tests run and get skipped

The issue is, at least in Chromium's infra, there is a list of tests generated

https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/third_party/gn/webgpu-cts/test_list.txt;l=1?q=test_list.txt&sq=&ss=chromium

The testing infra walks that list and expects each one to run.

It seems like there are 2 ways to solve this

1. Have 2 lists, one for `compat:*`, one for `webgpu:*`
2. Have 1 list and have all tests run in both

It was suggested it's easier if there is just one list. If we split there would also need to be separate expectations.txt files etc....

🤷‍♂️

lokokung:
Ahh.. right makes sense. Thanks!

PR #2886, opened 2023-08-09
Add HTMLImageElement and ImageData to copyExternalImageToTexture

beaufortfrancois:
This PR adds tests for HTMLImageElement and ImageData support to the GPUQueue copyExternalImageToTexture() method.

Chromium CL: https://chromium-review.googlesource.com/c/chromium/src/+/4729404

### DIFF
--- a/src/webgpu/web_platform/copyToTexture/ImageData.spec.ts
+++ b/src/webgpu/web_platform/copyToTexture/ImageData.spec.ts
@@ -3,66 +3,10 @@ copyExternalImageToTexture from ImageData source.
 `;
 
 import { makeTestGroup } from '../../../common/framework/test_group.js';
-import {
-  kTextureFormatInfo,
-  kValidTextureFormatsForCopyE2T,
-  EncodableTextureFormat,
-} from '../../format_info.js';
+import { kTextureFormatInfo, kValidTextureFormatsForCopyE2T } from '../../format_info.js';
 import { CopyToTextureUtils, kCopySubrectInfo } from '../../util/copy_to_texture.js';
-import { PerTexelComponent } from '../../util/texture/texel_data.js';
-import { TexelView } from '../../util/texture/texel_view.js';
 
-type TestColor = PerTexelComponent<number>;
-// None of the dst texture format is 'uint' or 'sint', so we can always use float value.
-const kColors = {
-  Red: { R: 1.0, G: 0.0, B: 0.0, A: 1.0 },
-  Green: { R: 0.0, G: 1.0, B: 0.0, A: 1.0 },
-  Blue: { R: 0.0, G: 0.0, B: 1.0, A: 1.0 },
-  Black: { R: 0.0, G: 0.0, B: 0.0, A: 1.0 },
-  White: { R: 1.0, G: 1.0, B: 1.0, A: 1.0 },
-  SemitransparentWhite: { R: 1.0, G: 1.0, B: 1.0, A: 0.6 },
-} as const;
-const kTestColorsOpaque = [
-  kColors.Red,
-  kColors.Green,
-  kColors.Blue,
-  kColors.Black,
-  kColors.White,
-] as const;
-const kTestColorsAll = [...kTestColorsOpaque, kColors.SemitransparentWhite] as const;
-
-function makeTestColorsTexelView({
-  testColors,
-  format,
-  width,
-  height,
-  premultiplied,
-  flipY,
-}: {
-  testColors: readonly TestColor[];
-  format: EncodableTextureFormat;
-  width: number;
-  height: number;
-  premultiplied: boolean;
-  flipY: boolean;
-}) {
-  return TexelView.fromTexelsAsColors(format, coords => {
-    const y = flipY ? height - coords.y - 1 : coords.y;
-    const pixelPos = y * width + coords.x;
-    const currentPixel = testColors[pixelPos % testColors.length];
-
-    if (premultiplied && currentPixel.A !== 1.0) {
-      return {
-        R: currentPixel.R! * currentPixel.A!,
-        G: currentPixel.G! * currentPixel.A!,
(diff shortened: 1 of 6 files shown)


src/webgpu/web_platform/copyToTexture/image.spec.ts
```diff
@@ -0,0 +1,366 @@
...
+        (imageCanvas as HTMLCanvasElement).toBlob(blob => resolve(blob));
+      } else {
+        (imageCanvas as OffscreenCanvas).convertToBlob().then(resolve, reject);
+      }
+    });
+    const blob = (await blobFromCanvas) as Blob;
+    const url = URL.createObjectURL(blob);
+    const image = new Image(imageCanvas.width, imageCanvas.height);
+    image.src = url;
+    await raceWithRejectOnTimeout(image.decode(), 5000, 'load image timeout');
```
kainino0x, line 175: image.decode() isn't appropriate here because that asks the image to be ready to add to the DOM, but that's not what we're going to do. Instead we should do roughly:

```ts
await raceWithRejectOnTimeout(new Promise(resolve => {
    image.onload = () => { resolve(); };
}), 5000, 'load image timeout');
```

even though that's annoyingly clunky.

src/webgpu/web_platform/copyToTexture/image.spec.ts
```diff
@@ -0,0 +1,366 @@
...
```
kainino0x, line 322: I think all of the logic above this point could be nicely wrapped into a helper function and deduplicated between the two tests (even though it's about to become a lot smaller).
The helper function can just be local to this file.

beaufortfrancois: Agree!

src/webgpu/web_platform/copyToTexture/image.spec.ts
```diff
@@ -0,0 +1,366 @@
...
+    // support '2d' contexts on OffscreenCanvas.
+    // In this situation, the case will be skipped.
+    let imageCanvas: HTMLCanvasElement | OffscreenCanvas;
+    if (typeof document !== 'undefined') {
+      imageCanvas = document.createElement('canvas');
+      imageCanvas.width = width;
+      imageCanvas.height = height;
+    } else if (typeof OffscreenCanvas === 'undefined') {
+      t.skip('OffscreenCanvas is not supported');
+      return;
```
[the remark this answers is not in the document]
beaufortfrancois: I've removed unnecessary `t.skip` where possible. Note that I had to keep two as otherwise TS complains `imageCanvasContext` can be null.

src/webgpu/web_platform/copyToTexture/image.spec.ts
```diff
@@ -0,0 +1,366 @@
...
+    const {
+      width,
+      height,
+      orientation,
+      dstColorFormat,
+      dstPremultiplied,
+      srcDoFlipYDuringCopy,
+    } = t.params;
+
+    // CTS sometimes runs on worker threads, where document is not available.
```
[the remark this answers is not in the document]
beaufortfrancois: It's much better indeed. Thx!

src/webgpu/web_platform/copyToTexture/ImageData.spec.ts
```diff
@@ -0,0 +1,286 @@
...
+  in CPU back resource.
+
+  It generates pixels in ImageData one by one based on a color list:
+  [Red, Green, Blue, Black, White, SemitransparentWhite].
+
+  Then call copyExternalImageToTexture() to do a full copy to the 0 mipLevel
+  of dst texture, and read the contents out to compare with the ImageData contents.
+
+  Do premultiply alpha during copy if 'premultipliedAlpha' in 'GPUImageCopyTextureTagged'
+  is set to 'true' and do unpremultiply alpha if it is set to 'false'.
```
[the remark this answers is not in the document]
beaufortfrancois: Got it!

src/webgpu/web_platform/copyToTexture/image.spec.ts
```diff
@@ -0,0 +1,366 @@
...
+        R: currentPixel.R! * currentPixel.A!,
+        G: currentPixel.G! * currentPixel.A!,
+        B: currentPixel.B! * currentPixel.A!,
+        A: currentPixel.A,
+      };
+    } else {
+      return currentPixel;
+    }
+  });
+}
```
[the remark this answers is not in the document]
beaufortfrancois: You're right! I moved everything I could in a `util.ts` file.

beaufortfrancois (commented):
Thanks for reviewing @kainino0x
Your feedback is much appreciated!

kainino0x (approved):
looks great, thanks!

src/webgpu/web_platform/copyToTexture/image.spec.ts
kainino0x: I thought onload was supposed to make it 'ready'. I was certain there was a way to do this without `.decode()`. I'll see if I can figure it out.

The spec does already have this check, see content timeline step 7 here:
https://gpuweb.github.io/gpuweb/#dom-gpuqueue-copyexternalimagetotexture
> Let usability be ? check the usability of the image argument(source).

kainino0x: > When switching to `onload`, tests don't pass anymore as the image needs to be fully decoded to be extracted for canvas.

Ahh, I suspect this is because onload was set after .src. It has to be set before, so that it'll already be registered when loading starts.

Anyway, I thought about this some more and decided it's not worth switching decode() to onload. Even though decode() is supposed to be for adding to the DOM I think browsers in practice don't mind it for WebGPU/WebGL uploads. If it works as it looks like it should (even though the spec doesn't guarantee it), it'll still save blocking time inside copyExternalImageToTexture where it would otherwise have to synchronously decode.

PR #2889, opened 2023-08-10
Only trigger WPT Lint test on pull_request

kainino0x:
Previously it was run every time a branch other than wpt-lint was pushed (including on branches in contributors' forks). This doesn't really achieve anything. (We could run it on main to make sure main stays green, but all that really matters is that PRs can be landed.)

Issue: none

<hr>

**Review requirements not applicable**


mehmetoguzderin (approved):
LGTM, great!

PR #2890, opened 2023-08-10
Optimize loadTreeForQuery by filtering eagerly in ParamsBuilder

kainino0x:
Previously, if you load a query like
`webgpu:api,validation,encoding,cmds,render,draw:vertex_buffer_OOB:type="draw";VBSize="exile";IBSize="exile";VStride0=false;IStride0=true;AStride="zero";offset=1`
or
`webgpu:api,validation,encoding,cmds,render,draw:vertex_buffer_OOB:type="draw";VBSize="exile";IBSize="exile";*`
then loadTreeForQuery would iterate *all* of the cases for that test
`webgpu:api,validation,encoding,cmds,render,draw:*`
and filter out the wrong ones at the very end.

This changes the ParamsBuilder iteration to be filter-aware, so it doesn't even start generating any of the subspaces of the combinatorial space that it can already see have a mismatch.

For the case above, this improves the runtime of loadTreeForQuery from 205ms to ~2ms (100x faster!). This saves tons of time if tests are run in a way where loadTreeForQuery is called for many sub-test or single-case queries (as is done in Chromium).

Issue: https://crbug.com/1470849

<hr>

- [ ] N/A ~All missing test coverage is tracked with "TODO" or `.unimplemented()`.~

- [ ] N/A ~Tests are properly located in the test tree.~
- [ ] N/A ~[Test descriptions](https://github.com/gpuweb/cts/blob/main/docs/intro/plans.md) allow a reader to "read only the test plans and evaluate coverage completeness", and accurately reflect the test code.~
- [ ] N/A ~Tests provide complete coverage (including validation control cases). **Missing coverage MUST be covered by TODOs.**~

### DIFF
--- a/src/common/framework/params_builder.ts
+++ b/src/common/framework/params_builder.ts
@@ -1,6 +1,9 @@
-import { Merged, assertMergedWithoutOverlap, mergeParams } from '../internal/params_utils.js';
+import { Merged, mergeParams, mergeParamsChecked } from '../internal/params_utils.js';
+import { comparePublicParamsPaths, Ordering } from '../internal/query/compare.js';
 import { stringifyPublicParams } from '../internal/query/stringify_params.js';
 import { assert, mapLazy } from '../util/util.js';
+
+import { TestParams } from './fixture.js';
 
 // ================================================================
 // "Public" ParamsBuilder API / Documentation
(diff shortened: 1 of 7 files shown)


kainino0x:
Had a false start with this until I realized: we can't special-case when we know the full case param, because that still won't tell us the _private_ params (or anything that doesn't serialize perfectly, but I don't think this happens).
So instead, I did this! It's invasive but it works great. And it works for multi-case queries, not just single-case queries.

src/common/framework/params_builder.ts
```diff
@@ -292,54 +370,6 @@ export class SubcaseParamsBuilder<CaseP extends {}, SubcaseP extends {}>
   }
 }
 
-/** Creates a generator function for expandWithParams() methods above. */
```
kainino0x, line 295: All this logic got moved up inline, because it was starting to get really messy (and this caused a bug that I had a really hard time debugging because the types were screwy because function(x) can be implicitly cast to function(x,y)).
Being able to specialize this above also saves a bunch of unnecessary `mergeParams({}, x)`.

toji (approved):
Bit tricky to read through, given all the layers of generators, but LGTM as far as I can tell! Really fantastic how much difference this can make to the test initialization time, I'm excited to see the impact it has on the Chrome bots!

Looks like you'll have to update at least one more unittest in `unittests:loaders_and_trees:case:` to make the checks happy.

kainino0x:
Confused, I thought I had them passing. I guess I forgot about this one.

PR #2891, opened 2023-08-11
Cover ImageBitmap option "colorSpaceConversion: none"

shaoboyan:
ImageBitmap creation could set option colorSpaceConversion:none. 

Cts should cover this option to check whether it has side effect for copyEI2T.

Issue: #

<hr>

### DIFF
--- a/src/webgpu/web_platform/copyToTexture/ImageBitmap.spec.ts
+++ b/src/webgpu/web_platform/copyToTexture/ImageBitmap.spec.ts
@@ -106,6 +106,7 @@ g.test('from_ImageData')
     u
       .combine('alpha', ['none', 'premultiply'] as const)
       .combine('orientation', ['none', 'flipY'] as const)
+      .combine('colorSpaceConversion', ['none', 'default'] as const)
       .combine('srcDoFlipYDuringCopy', [true, false])
       .combine('dstColorFormat', kValidTextureFormatsForCopyE2T)
       .combine('dstPremultiplied', [true, false])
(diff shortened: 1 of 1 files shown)


kainino0x (commented):
Change itself looks fine, but I don't think we have any test cases right now that actually have a color space on the source image? which I believe means the colorSpaceConversion option is expected to have no effect:
- 'default' means convert from the image color space (defaults to srgb if untagged) to the destination space (srgb)
- 'none' means reinterpret the image as untagged (==srgb) and convert to the destination space (srgb)

We could (theoretically) do this by constructing the ImageBitmap from an ImageData or canvas that has a colorSpace set on it. (And skipping the test, I suppose, if that's not supported.)

shaoboyan:
> Change itself looks fine, but I don't think we have any test cases right now that actually have a color space on the source image? which I believe means the colorSpaceConversion option is expected to have no effect:
> 
> * 'default' means convert from the image color space (defaults to srgb if untagged) to the destination space (srgb)
> * 'none' means reinterpret the image as untagged (==srgb) and convert to the destination space (srgb)

Yes, that's exactly the side effect happens (affect premultiply in chrome implementation). See details here https://bugs.chromium.org/p/chromium/issues/detail?id=1457649

shaoboyan:
> We could (theoretically) do this by constructing the ImageBitmap from an ImageData or canvas that has a colorSpace set on it. (And skipping the test, I suppose, if that's not supported.)
I'll try this on ImageData with color space set.

kainino0x (approved):
> Yes, that's exactly the bug root cause(affect premultiply in chrome implementation).

Ah, I see, so this change catches the old broken behavior. LGTM then (you can follow up with color space tests if you're looking at them, but it's not critical and we already have a big TODO about that)

Issue #1609, opened 2022-07-04
wgsl: execution test, bitcast: f16 cases remain


dj2:
* [x] Identity `bitcast` for i32, u32
* [x] Identity `bitcast` for f32
* [x] Identity `bitcast` for f16
* [x] `bitcast<u32>(i32)`
* [x] `bitcast<u32>(f32)`
* [x] `bitcast<i32>(u32)`
* [x] `bitcast<i32>(f32)`
* [x] `bitcast<f32>(u32)`
* [x] `bitcast<f32>(i32)`
* [x] `bitcast<vecN<u32>>(vecN<i32>)`
* [x] `bitcast<vecN<u32>>(vecN<f32>)`
* [x] `bitcast<vecN<i32>>(vecN<u32>)`
* [x] `bitcast<vecN<i32>>(vecN<f32>)`
* [x] `bitcast<vecN<f32>>(vecN<u32>)`
* [x] `bitcast<vecN<f32>>(vecN<i32>)`
* [x] `bitcast<u32>(vec2<f16>)`
* [x] `bitcast<i32>(vec2<f16>)`
* [x] `bitcast<f32>(vec2<f16>)`
* [x] `bitcast<vec2<f16>>(u32)`
* [x] `bitcast<vec2<f16>>(i32)`
* [x] `bitcast<vec2<f16>>(f32)`
* [x] `bitcast<vec2<u32>>(vec4<f16>)`
* [x] `bitcast<vec2<i32>>(vec4<f16>)`
* [x] `bitcast<vec2<f32>>(vec4<f16>)`
* [x] `bitcast<vec4<f16>>(vec2<u32>)`
* [x] `bitcast<vec4<f16>>(vec2<i32>)`
* [x] `bitcast<vec4<f16>>(vec2<f32>)`

Validation tests inplemented in `shader\validation\expression\call\builtin\bitcast.spec.ts`:

* [x] `bitcast` to/from `mat` types fail
* [x] `bitcast` to/from `vec3<f16>` fail
* [x] `bitcast` to/from `f16` fails
* [x] `bitcast` to non-numeric types fails
* [x] const `bitcast` from integer types to inf and nan values fail

dneto0:
#2754 tests failure of:
- to/from mat
- to/from non-numeric types
- const bitcast resulting in to inf nan floats

dneto0:
The remaining tests require f16 support.

jzm-intel:
[...]

Issue #1248, opened 2022-04-25
wgsl: f16 built-in execution tests


dj2:
List all f16 related built-in here to track the overall built-in execution tests implementation.

* [x] abs
* [x] acos (#2551)
* [x] acosh (#2559)
* [x] arrayLength
* [x] asin (#2553)
* [x] asinh (#2561)
* [x] atan
* [x] atan2
* [x] atanh (#2563)
* [x] bitcast (#1609)
* [x] ceil
* [x] clamp
* [x] cos
* [x] cosh (#2538)
* [x] cross (#2540)
* [x] degrees (#2565)
* [x] determinant (#2522)
* [x] distance (#2557)
* [x] dot (#2513)
* [x] exp (#2543)
* [x] exp2 (#2544)
* [x] faceForward (#2547)
* [x] floor
* [x] fma (#2549)
* [x] fract
* [x] frexp (#2587)
* [x] inversesqrt
* [x] ldexp
* [x] length (#2555)
* [x] log
* [x] log2
* [x] max
* [x] min
* [x] mix (#2567)
* [x] modf (#2568)
* [x] normalize (#2571)
* [x] pow (#2573)
* [x] radians (#2577)
* [x] reflect (#2579)
* [x] refract (#2580)
* [x] round (#2510)
* [x] saturate
* [x] select
* [x] sign (#2583)
* [x] sin
* [x] sinh (#2535)
* [x] smoothstep (#2533)
* [x] sqrt (#2532)
* [x] step (#2529)
* [x] tan (#2507)
* [x] tanh (#2527)
* [x] transpose (#2520)
* [x] trunc (#2524)

jzm-intel:
Upload this issue to make it a tracker for the whole f16 built-in execution test implementation.

zoddicus:
All subtasks completed. Thanks @jzm-intel for all of the hard work :1st_place_medal:

PR #2897, opened 2023-08-16
wgsl: f16 built-in execution test for bitcast

jzm-intel:
This PR add execution tests for `bitcast` built-in from and to f16 types.

Issue: #1248, #1609

<hr>

### DIFF
--- a/src/webgpu/shader/execution/expression/call/builtin/bitcast.spec.ts
+++ b/src/webgpu/shader/execution/expression/call/builtin/bitcast.spec.ts
@@ -119,23 +119,30 @@ const f16ZerosInterval: FPInterval = new FPInterval('f16', -0.0, 0.0);
 
 /**
  * @returns an u32 whose lower and higher 16bits are the two elements of the
- * given array of two u16 respectively.
+ * given array of two u16 respectively, in little-endian.
  */
 function u16x2ToU32(u16x2: number[]): number {
   assert(u16x2.length === 2);
-  const array = new Uint16Array(2);
-  [array[0], array[1]] = u16x2;
-  return new Uint32Array(array.buffer)[0];
+  // Create a DataView with 4 bytes buffer.
+  const buffer = new ArrayBuffer(4);
+  const view = new DataView(buffer);
+  // Enforce little-endian.
+  view.setUint16(0, u16x2[0], true);
+  view.setUint16(2, u16x2[1], true);
+  return view.getUint32(0, true);
 }
 
 /**
- * @returns an array of two u16, respectively the lower and higher 16bits of given u32.
+ * @returns an array of two u16, respectively the lower and higher 16bits of
+ * given u32 in little-endian.
  */
 function u32ToU16x2(u32: number): number[] {
-  const array = new Uint32Array(1);
-  array[0] = u32;
-  const dst_array = new Uint16Array(array.buffer);
-  return [dst_array[0], dst_array[1]];
+  // Create a DataView with 4 bytes buffer.
+  const buffer = new ArrayBuffer(4);
+  const view = new DataView(buffer);
+  // Enforce little-endian.
+  view.setUint32(0, u32, true);
+  return [view.getUint16(0, true), view.getUint16(2, true)];
 }
 
 /**
--- a/src/webgpu/shader/execution/expression/call/builtin/bitcast.spec.ts
+++ b/src/webgpu/shader/execution/expression/call/builtin/bitcast.spec.ts
@@ -42,10 +42,9 @@ import {
   TypeU32,
   TypeF16,
   TypeVec,
-  vec2,
-  vec4,
   Vector,
   Scalar,
+  toVector,
 } from '../../../../../util/conversion.js';
 import { FPInterval, FP } from '../../../../../util/floating_point.js';
 import {
@@ -144,7 +143,7 @@ function u32ToU16x2(u32: number): number[] {
  */
 function u16x2ToVec2F16(u16x2: number[]): Vector {
   assert(u16x2.length === 2);
-  return vec2(f16(reinterpretU16AsF16(u16x2[0])), f16(reinterpretU16AsF16(u16x2[1])));
+  return toVector(u16x2.map(reinterpretU16AsF16), f16);
 }
 
 /**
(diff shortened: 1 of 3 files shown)


zoddicus (commented):
Overall this looks good to me, but I would like @dneto0 to take a look at the changes in bitcast.spec.ts, since he is more familiar with that part of the spec.

dneto0:
Hi, sorry for the delay in reviewing.  I intend to review this carefully, but that happen on Monday.

src/webgpu/shader/execution/expression/call/builtin/bitcast.spec.ts
```diff
@@ -71,44 +86,190 @@ const f32InfAndNaNInI32 = f32InfAndNaNInU32.map(u => reinterpretU32AsI32(u));
...
+// F16 values, finite, Inf/NaN, and zeros. Represented in float and u16.
+const f16FiniteInF16: number[] = [...fullF16Range(), kValue.f16.negative.zero];
+const f16FiniteInU16: number[] = f16FiniteInF16.map(u => reinterpretF16AsU16(u));
+
+const f16InfAndNaNInU16: number[] = [
+  // Cover NaNs evenly in integer space.
+  // The positive NaN with the lowest integer representation is the integer
+  // for infinity, plus one.
+  // The positive NaN with the highest integer representation is u16 0x7fff i.e. 32767.
+  ...linearRange(kBit.f16.infinity.positive + 1, 32767, numNaNs).map(v => Math.ceil(v)),
```
dneto0, line 105: Right.   The call to Math.ceil(.)   forces the value to be integral. That makes sense.

src/webgpu/shader/execution/expression/call/builtin/bitcast.spec.ts
```diff
@@ -71,44 +86,190 @@ const f32InfAndNaNInI32 = f32InfAndNaNInU32.map(u => reinterpretU32AsI32(u));
...
+const f16ZerosInU16 = [kBit.f16.negative.zero, 0];
+
+// f16 interval that match +/-0.0.
+const f16ZerosInterval: FPInterval = new FPInterval('f16', -0.0, 0.0);
+
+/**
+ * @returns an u32 whose lower and higher 16bits are the two elements of the
+ * given array of two u16 respectively.
+ */
+function u16x2ToU32(u16x2: number[]): number {
```
dneto0, line 124: Both Uint16Array and Uint32Array assume the platform byte order.  (e.g. see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint32Array )

I suspect this code assumes little-endian (which is the most common).
Please use a [DataView](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView) for this function and the next one. 

Hey @zoddicus do you know of other endianness sensitivities in the CTS codebase?

src/webgpu/shader/execution/expression/call/builtin/bitcast.spec.ts
```diff
@@ -71,44 +86,190 @@ const f32InfAndNaNInI32 = f32InfAndNaNInU32.map(u => reinterpretU32AsI32(u));
...
+    .map(u16 => isFiniteF16(reinterpretU16AsF16(u16)))
+    .reduce((a, b) => a && b, true);
+}
+
+/**
+ * @returns an array of N elements with the i-th element being an array of len elements
+ * [a_i, a_((i+1)%N), ..., a_((i+len-1)%N)], for the input array of N element [a_1, ... a_N]
+ * and the given len. For example, slidingSlice([1, 2, 3], 2) result in
+ * [[1, 2], [2, 3], [3, 1]].
+ * This helper function is used for generating vector cases from scalar values array.
```
dneto0, line 172: Nice!

src/webgpu/shader/execution/expression/call/builtin/bitcast.spec.ts
```diff
@@ -71,44 +86,190 @@ const f32InfAndNaNInI32 = f32InfAndNaNInU32.map(u => reinterpretU32AsI32(u));
...
+function u16x4ToVec4F16(u16x4: number[]): Vector {
+  assert(u16x4.length === 4);
+  return toVector(u16x4.map(reinterpretU16AsF16), f16);
+}
+
+/**
+ * @returns true if and only if a given u32 can bitcast to a vec2<f16> with all elements
+ * being finite f16 values.
+ */
+function canU32BitcastToFiniteVec2F16(u32: number): boolean {
```
dneto0, line 161: Uses of this function are really nice!

src/webgpu/shader/execution/expression/call/builtin/bitcast.spec.ts
```diff
@@ -158,13 +319,320 @@ function bitcastI32ToF32Comparator(i: number): Comparator {
...
+  if (!isFiniteF16(bitcastedF16Value)) {
+    return [f16UnboundedInterval];
+  }
+  // If the casted f16 value is +/-0.0, the result can be one of both. Note that in JS -0.0 === 0.0.
+  if (bitcastedF16Value === 0.0) {
+    return [f16ZerosInterval];
+  }
+  const exactInterval = FP.f16.toInterval(bitcastedF16Value);
+  // If the casted f16 value is subnormal, it also may be flushed to +/-0.0.
+  return [exactInterval, ...(isSubnormalNumberF16(bitcastedF16Value) ? [f16ZerosInterval] : [])];
```
dneto0, line 345: Nice catch in handling the case that a subnormal value can map to a zero of the opposite sign.

dneto0 (approved):
This is really good work!
Sorry for the delay.

I have the tiniest of nits:  I think there is an assumption that the host is little-endian.  I suggest using DataView to remove that assumption.

I'm approving now because there is no little harm in landing this.  Anybody who runs these tests on a big-endian machine will quickly find the problem and fix it.
It's up to you whether to land this now and fix the endian issue later, or to address it now.

thanks!

src/webgpu/shader/execution/expression/call/builtin/bitcast.spec.ts
jzm-intel: I see, use the `DataView` to explicitly enforce the little-endian.

All `reinterpret*As*` in `util/conversion.ts` use TypedArray to do the conversion, but most of them use array with one single element so that would be fine. `reinterpretF64AsU32s` and `reinterpretU32sAsF64` assume the little endian, but they are only used for serialization and deserialization for testcases cache, so I thought they would be fine too.

jzm-intel:
Thanks for reviewing! I changed to use the `DataView` for casting between U32 and U16s.

src/webgpu/shader/execution/expression/call/builtin/bitcast.spec.ts
zoddicus: I agree with @jzm-intel's analysis.
I would add that all of the AF work is going to assume little endianess for the time being.
Until there is someone with a big endian platform needing bidirectional support, I think it is ok to assume little endianess.

Issue #2028, opened 2022-11-24
Support reading structs as outputs from builtin tests


zoddicus:
Currently the testing framework only support reading out Scalars and Vectors, but some builtins, i.e. `modf`, return structured data. The current workaround, https://github.com/gpuweb/cts/pull/2014, is to have a test for each member of the output structure that only pulls that member out for testing. This means for a structure with N members there needs to be N almost identical tests being run.

The better solution is for the framework to support something like a  TypeStruct similar to TypeVector that handle packing and unpack out of the buffers, so all the members can be tested in a single invocation. This will probably get tricky to get right.

zoddicus:
Sending to @ben-clayton, since he mentioned he might take a pass at implementing this. Feel free to send back to me.

zoddicus:
For supporting AbstractFloats, I am running into really wanting to be able to return something like a TypeStruct (of 2 u32s). Everything I can think to make this work without just eating the frog on this issue is just hacking workarounds in to the code base.

Given that there are multiple existing tests that could be improved by adding a TypeStruct, and AbstractFloat tests are basically predicated on them, I think it is time to invest the effort. Additionally, I have already added in TypeMatrix, so I am pretty familiar with how to add new types to the framework.

dj2:
This was handled in other ways, the full general solution wasn't necessary.

PR #2905, opened 2023-08-17
Add support for structs to framework and use for modf

zoddicus:
This adds support for structs both in the input specifying code and expectation code.

modf executions tests ares re-written to use this new support to demonstrate it working and flesh out the details.

Future PRs will convert other tests to use structs instead of multiple passes or other hacks.

Issue #2028

<hr>

### DIFF
--- a/src/webgpu/shader/execution/expression/call/builtin/modf.spec.ts
+++ b/src/webgpu/shader/execution/expression/call/builtin/modf.spec.ts
@@ -18,8 +18,8 @@ Returns the result_struct for the given type.
 
 import { makeTestGroup } from '../../../../../../common/framework/test_group.js';
 import { GPUTest } from '../../../../../gpu_test.js';
-import { toVector, TypeF32, TypeVec } from '../../../../../util/conversion.js';
-import { FP, FPKind } from '../../../../../util/floating_point.js';
+import { toVector, TypeF32, TypeStruct, TypeVec } from '../../../../../util/conversion.js';
+import { FP, FPInterval, FPKind, FPStruct } from '../../../../../util/floating_point.js';
 import { fullF32Range, vectorF32Range } from '../../../../../util/math.js';
 import { makeCaseCache } from '../../case_cache.js';
 import {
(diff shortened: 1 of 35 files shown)


src/webgpu/util/conversion.ts
```diff
@@ -755,8 +771,69 @@ export function TypeMat(cols: number, rows: number, elementType: ScalarType): Ma
...
+   * given byte offset
+   */
+  public read(buf: Uint8Array, offset: number): Struct {
+    const elements: Value[] = this.elementTypes.map(e => {
+      const result = e.read(buf, offset);
+      offset += e.size;
+      // vec3s have a padding element
+      if (e instanceof VectorType) {
+        if (e.width === 3) {
+          offset += e.elementType.size;
```
jzm-intel, line 795: I am afraid that a vec3 in structure doesn't take up the padding space, see [the example in WGSL spec](https://gpuweb.github.io/gpuweb/wgsl/#example-fc0bb4df).
For example, in the following structure
``` rust
struct A {                            //            align(16) size(16)
    a: vec3<f32>,                     // offset(0)  align(16) size(12)
    b: f32,                           // offset(12) align(4)  size(4)
}
```
there is not padding space between elements `a` and `b`. Actually, the spec says ["Each subsequent member is placed at the lowest offset that satisfies the member type alignment, and which avoids overlap with the previous member"](https://gpuweb.github.io/gpuweb/wgsl/#example-fc0bb4df:~:text=Each%20subsequent%20member%20is%20placed%20at%20the%20lowest%20offset%20that%20satisfies%20the%20member%20type%20alignment%2C%20and%20which%20avoids%20overlap%20with%20the%20previous%20member.), and the "overlap" here only consider the actual size of previous member.

So, if I get it right, instead of adding offset after a vec3, we should align the offset before reading each element in line 790. I thought the `modf` vec3 testcases passed as expected is because the result structure is
```rust
struct ModfOutput {
    a: vec3<f32>,                                       // offset(0)  align(16)  size(12)
    // -- implicit vec3<f32> alignment padding for b -- // offset(12)            size(4)
    b: vec3<f32>,                                       // offset(16) align(16)  size(12)
}
```
and the alignment padding size is coincidently equal to the "padding element" size.

src/webgpu/util/conversion.ts
```diff
@@ -755,8 +771,69 @@ export function TypeMat(cols: number, rows: number, elementType: ScalarType): Ma
...
+    return this.name;
+  }
+
+  public get size(): number {
+    return this.elementTypes.reduce((sum, cur) => {
+      sum += cur.size;
+      // vec3s have a padding element
+      if (cur instanceof VectorType) {
+        if (cur.width === 3) {
+          sum += cur.elementType.size;
```
jzm-intel, line 813: The same here, we should compute the size of structure by the last member's position + it's size and round up to the maximum alignment of structure's members. [Spec](https://gpuweb.github.io/gpuweb/wgsl/#example-fc0bb4df:~:text=roundUp(AlignOf(S)%2C%20justPastLastMember)%0A%0Awhere%20justPastLastMember%20%3D%20OffsetOfMember(S%2CN)%20%2B%20SizeOfMember(S%2CN))

So, the size of
``` rust
struct A {                            //            align(16) size(16)
    a: vec3<f32>,                     // offset(0)  align(16) size(12)
    b: f32,                           // offset(12) align(4)  size(4)
}
```
would be 16 rather than 20.

src/webgpu/util/floating_point.ts
```diff
@@ -265,6 +265,16 @@ export type FPMatrix =
       [FPInterval, FPInterval, FPInterval, FPInterval]
     ];
 
+export type FPValue = FPInterval | FPVector | FPMatrix | FPStruct;
+
+/** Container representing a struct of FP values */
+export class FPStruct {
```
ben-clayton, line 271: The choice of `FPStruct` was a little surprising.
In WGSL, a struct doesn't need to have members of the same type, and `frexp` returns a structure of `f32` and `i32`, which presumably this can't be used for.

Its not clear what the rules are for this. Do each of the members need to be the same FP type, or can you mix a f32 with a f16?

src/webgpu/shader/execution/expression/call/builtin/modf.spec.ts
```diff
@@ -32,74 +32,51 @@ import {
...
-    };
-    return fullF32Range().map(makeCase);
-  },
-  f32_vec2_fract: () => {
-    return vectorF32Range(2).map(makeVectorCaseFract.bind(null, 'f32'));
-  },
-  f32_vec2_whole: () => {
-    return vectorF32Range(2).map(makeVectorCaseWhole.bind(null, 'f32'));
+  f32_vec2: () => {
+    return vectorF32Range(2).map(makeVectorCase.bind(null, 'f32'));
```
ben-clayton, line 69: nit: I think `n => makeVectorCase('f32', n)` is easier to read than `makeVectorCase.bind(null, 'f32')`. It's also shorter.

src/webgpu/shader/execution/expression/expression.ts
```diff
@@ -123,6 +133,12 @@ function valueStride(ty: Type): number {
     );
   }
 
+  if (ty instanceof StructType) {
+    return ty.elementTypes.reduce((sum, cur) => {
+      return sum + valueStride(cur);
```
ben-clayton, line 138: This isn't taking alignment into consideration.

```
struct S {
  a : i32,
  b : vec4<f32>,
}
```

Would be misaligned by 12 bytes.

src/webgpu/shader/execution/expression/expression.ts
```diff
@@ -455,7 +471,7 @@ function wgslValuesArray(
   expressionBuilder: ExpressionBuilder
 ): string {
   // AbstractFloat values cannot be stored in an array
```
ben-clayton, line 473: They can.

```
const an_array_of_abstract_float = array(1.0, 2.0, 3.0);
```

But like all abstract numeric types, they'll be concreted before being assigned to a `var`, `let` or parameter.

src/webgpu/shader/execution/expression/expression.ts
```diff
@@ -485,23 +501,55 @@ function wgslInputVar(inputSource: InputSource, count: number) {
...
+
+  const structTypes: StructType[] = [];
+  if (resultType instanceof StructType) {
+    structTypes.push(resultType);
+  }
+
+  structTypes.push(...(parameterTypes.filter(t => t instanceof StructType) as StructType[]));
+
+  // Passing through set to make sure added decls are unique
+  header += [...new Set(structTypes)].map(s => {
```
ben-clayton, line 516: This looks like it'll deduplicate structs if the JS objects are the same. Does this still work when serialising and deserializing a cached case, or do we potentially end up with multiple struct objects with the same name and members? I think it would be safer to use a map of struct name to struct.

ben-clayton, line 516: Is there any determinism in the result of an array -> set -> array? If not, I'd sort based on the struct name after going via the set.

src/webgpu/shader/execution/expression/expression.ts
```diff
@@ -670,7 +718,7 @@ ${wgslInputVar(inputSource, cases.length)}
 @compute @workgroup_size(1)
 fn main() {
   for (var i = 0; i < ${cases.length}; i++) {
-    outputs[i].value = ${expr};
+    outputs[i].value = ${assignmentBuilder(expr)};
```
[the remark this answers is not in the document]
ben-clayton: Good spot.
You'd have to make sure that `${expr}` does all the abstract-numeric logic first, because as soon as it hits the `let`, it'll concretize.

zoddicus:
I forgot about different storage types requiring different layout requirements... I am going to take a step back and try to re-work this using multiple Outputs, so that the CTS doesn't need to have yet another chunk of complex alignment logic. This might possibly get rid of the need for the FPStruct, which is problematic also, due to needing to support mixed types.

zoddicus:
Abandoning this, since the path I was going down that included this PR, implementing struct support to eventually use for AF tests won't work, since structs cannot be the type for vectors, which is the problem I was running into with my previous AF implementation.

I need to take a step back and think about how to properly handle AF results... which is probably going to be super low road and require two passes, one for upper bits and one for lower bits...

PR #2928, opened 2023-09-07
wgsl: Add non-matrix AbstractFloat subtraction execution tests

zoddicus:
Issue #1626

<hr>


src/webgpu/shader/execution/expression/binary/af_subtraction.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+    return FP.abstract.generateScalarVectorToVectorCases(
+      sparseF64Range(),
+      sparseVectorF64Range(4),
+      'finite',
+      subtractionScalarVectorInterval
+    );
+  },
+});
+
+g.test('scalar')
```
jzm-intel, line 84: Actually this test covers not only scalar-scalar cases but also vector-vector cases by vectorization. Not sure if we would need to change this name?

src/unittests/floating_point.spec.ts
```diff
@@ -4554,12 +4554,28 @@ const kSubtractionInterval64BitsNormalCases = {
...
+    { input: [0.1, 0], expected: [0.1] },
+    { input: [0, -0.1], expected: [0.1] },
+    { input: [-0.1, 0], expected: [-0.1] },
+    { input: [0, 0.1], expected: [-0.1] },
+
+    { input: [0.1, 0.1], expected: [0] },
+    { input: [-0.1, -0.1], expected: [0] },
+    // f64 0x3FB999999999999A - 0xBFB999999999999A = 0x3FC999999999999A
+    { input: [0.1, -0.1], expected: [reinterpretU64AsF64(0x3fc999999999999an)] },  // ~0.2
+    // f64 0xBFB999999999999A- 0x3FB999999999999A = 0xBFC999999999999A
```
jzm-intel, line 4570:
```suggestion
    // f64 0xBFB999999999999A - 0x3FB999999999999A = 0xBFC999999999999A
```

jzm-intel (approved):
Looks good to me, thanks

src/webgpu/shader/execution/expression/binary/af_subtraction.spec.ts
zoddicus: Hmmm... naming things is hard :smile:
I agree that the name scalar doesn't convey that scalar-scalar and vector-vector tests are occuring. The name comes from these being named things like 'f16' and 'f32' in other test files.

The sticking point for a new name I am hitting is that fact that we want something that indicates that homogenous operations are occuring, but it not be so generic as to suggest that it might also cover matrix-matrix operations.

I am tempted to separate these up into two test fixtures 'scalar' and 'vector', and only use `.combine('vectorize', [2, 3, 4] as const)` on 'vector'

I think this requires a bit of thinking to decide on the correct change, so I am going to leave the naming as-is for these PRs, and file a new issue (https://github.com/gpuweb/cts/issues/2935), since this will affect the f32 and f16 tests too

PR #2947, opened 2023-09-13
Update websocket-logger instructions to use `npm ci`

### LINKED ISSUE #2945: Should `websocket-logger` instructions use `npm ci` instead of `npm i`?
During the review of my [PR](https://github.com/gpuweb/cts/pull/2942) adding further documentation about updating timing metadata, the question came up about whether the websocket-logger instructions should be using `npm install` as they currently state, versus `npm ci`.

My broad understanding of the difference between the two is that `npm i` is meant to be used during active development, since it attempts to update dependencies to the newest compatible version, where `npm ci` is meant to be used in situations like CI or automated builds where a specific version of the dependencies is desired.

@greggman pointed out that using `npm i` is a potential security risk, because if a semantically compatible newer version of a dependency with a flaw in it is deployed to the ecosystem, the instructions will have to you installing it.

Considering all of this, I am pretty sure that we want to be using `npm ci`. The only case that I can think of preferring `npm i` is if there is a fast moving dependency that we own in there, so it needs to be rev'd regularly.

I am going to take a look into this.

zoddicus:
Changing to using `npm ci` instead of `npm install` to avoid potentially pulling in unvetted dependencies.

Fixes #2945

<hr>

### DIFF
--- a/docs/adding_timing_metadata.md
+++ b/docs/adding_timing_metadata.md
@@ -66,8 +66,8 @@ throughout the following steps.
 
 At `tools/websocket-logger/`
 ```
-npm install
-npm run
+npm ci
+npm start
 ```
 
 The output from this command will indicate where the results are being logged,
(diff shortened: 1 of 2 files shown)


zoddicus:
There was already a package-lock.json file, so didn't need to add one. (I must of have accidently removed my local copy when previously stating there wasn't one)

austinEng (approved):

PR #2953, opened 2023-09-13
Compat: Limit max attributes

greggman:
In compat @builtin(vertex_index) and @builtin(instance_index) each take an attribute so account for that in this test.

It's possible we should refactor this test to not
use vertex_index, instance_instance. For example we could make each pair of data generate the correct pixel position.

For now it seemed best to get it to pass.
The test in webgpu/compat/api/validation/encoding/render_pipeline/vertex_state.spec.ts does a simple test that you can use maxVertexAttributes but it does not test all the combinations like this test. It only tests that creating a pipeline passes if you use maxVertexAttributes and fails if you use maxVertexAttributes + the builtins above

<hr>

### DIFF
--- a/src/webgpu/api/operation/vertex_state/correctness.spec.ts
+++ b/src/webgpu/api/operation/vertex_state/correctness.spec.ts
@@ -941,13 +941,15 @@ g.test('max_buffers_and_attribs')
   .params(u => u.combine('format', kVertexFormats))
   .fn(t => {
     const { format } = t.params;
-    const attributesPerBuffer = Math.ceil(kMaxVertexAttributes / kMaxVertexBuffers);
+    // In compat mode, @builtin(vertex_index) and @builtin(instance_index) each take an attribute
+    const maxVertexAttributes = t.isCompatibility ? kMaxVertexAttributes - 2 : kMaxVertexAttributes;
+    const attributesPerBuffer = Math.ceil(maxVertexAttributes / kMaxVertexBuffers);
     let attributesEmitted = 0;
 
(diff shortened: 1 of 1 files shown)


src/webgpu/api/operation/vertex_state/correctness.spec.ts
```diff
@@ -941,13 +941,15 @@ g.test('max_buffers_and_attribs')
   .params(u => u.combine('format', kVertexFormats))
   .fn(t => {
     const { format } = t.params;
-    const attributesPerBuffer = Math.ceil(kMaxVertexAttributes / kMaxVertexBuffers);
+    // In compat mode, @builtin(vertex_index) and @builtin(instance_index) each take an attribute
+    const maxVertexAttributes = t.isCompatibility ? kMaxVertexAttributes - 2 : kMaxVertexAttributes;
```
lokokung, line 945: Is it possible to query the device for this info? If not this is fine for now, but a TODO to make like a separate limit tables or something for compat is probably a good idea!

lokokung (approved):

src/webgpu/api/operation/vertex_state/correctness.spec.ts
greggman: `device.limits.maxVertexAttribute` will be 16 on both compat and non-compat. The compat limit is that the 2 builtins each count as 1 attribute. So this test, knowing that it is using those 2 builtins, needs to subtract 2. If it wasn't using the builtins then it would not need to subtract them.

PR #2958, opened 2023-09-14
Add a copy button

greggman:
Clicking it puts the query in the clipboard


toji (approved):
LGTM! Thanks, tis will be really helpful!

PR #2982, opened 2023-09-21
Show parseQuery error

greggman:
Some errors in parsing the query string were only showing in the JavaScript console so that a bad query the test page would just show nothing. This is an attempt to try to surface those errors to the page.


toji (approved):
LGTM

PR #3004, opened 2023-09-25
Compat: make vertex state correctness tests handle different limits

greggman:
<hr>

### DIFF
--- a/src/webgpu/api/operation/vertex_state/correctness.spec.ts
+++ b/src/webgpu/api/operation/vertex_state/correctness.spec.ts
@@ -4,11 +4,14 @@ float tolerance.
 `;
 
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
-import { assert, memcpy, unreachable } from '../../../../common/util/util.js';
 import {
-  kMaxVertexAttributes,
-  kMaxVertexBufferArrayStride,
-  kMaxVertexBuffers,
+  assert,
+  filterUniqueValueTestVariants,
+  makeValueTestVariant,
+  memcpy,
+  unreachable,
+} from '../../../../common/util/util.js';
+import {
   kPerStageBindingLimits,
   kVertexFormatInfo,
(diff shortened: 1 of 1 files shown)


src/webgpu/api/operation/vertex_state/correctness.spec.ts
```diff
@@ -600,11 +606,21 @@ g.test('vertex_format_to_shader_format_conversion')
       .combine('format', kVertexFormats)
       .combine('shaderComponentCount', [1, 2, 3, 4])
       .beginSubcases()
-      .combine('slot', [0, 1, kMaxVertexBuffers - 1])
-      .combine('shaderLocation', [0, 1, kMaxVertexAttributes - 1])
+      .combine('slotVariant', [
+        { mult: 0, add: 0 },
+        { mult: 0, add: 1 },
+        { mult: 1, add: -1 },
+      ])
```
lokokung, line 613: Hmm we have this sort of thing in quite a few places... I assume we need to do this for a couple reasons, one being so that the test names are consistent? (does it actually matter for subcases?). Do you know if the subcase naming is important? If not, maybe we could add some sort of utility that converts `[{ mult: number, add: number }]` and a limit name into a a list of actual values and then use that in these sort of scenarios? I think it would be helpful to see the actual value when looking at the test results instead of the operations because the value isn't always easily deducible in the test name.

(Not critical, just if you have an idea for it. And definitely would be better in a follow up.)

lokokung (approved):

src/webgpu/api/operation/vertex_state/correctness.spec.ts
greggman: I don't know if subcase names matter but, as the testing framework is currently designed, you don't know at subcase creation time, whether or not you're in compat mode so you can't know the actual limit :(   Maybe that could/should be refactored.

PR #3011, opened 2023-09-26
Compat limit render pass desc

greggman:
<hr>

### DIFF
--- a/src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts
+++ b/src/webgpu/api/validation/render_pass/render_pass_descriptor.spec.ts
@@ -6,9 +6,10 @@ TODO: review for completeness
 
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
 import { range } from '../../../../common/util/util.js';
-import { kMaxColorAttachments, kQueryTypes } from '../../../capability_info.js';
+import { kMaxColorAttachmentsToTest, kQueryTypes } from '../../../capability_info.js';
 import { GPUConst } from '../../../constants.js';
 import {
+  computeBytesPerSampleFromFormats,
   kDepthStencilFormats,
   kRenderableColorTextureFormats,
   kTextureFormatInfo,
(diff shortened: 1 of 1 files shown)


greggman:
fyi: g.test('color_attachments,limits,maxColorAttachmentBytesPerSample,unaligned') test only works with 8 attachments (it's skipped with less currently based on the check) 

The test is testing that some attachments, with alignment requirements, exceed `maxColorAttachmentBytesPerSample` and get a validation error, where as if rearranged, they they're within the limit.

AFAICT there is no combination of 4 attachments (the limit in compat) that can trigger the "exceed limits because of alignment" check.

Also, maybe we should file a bug for this, but this test is effectively tested twice. This same issue is tested in `maxColorAttachmentBytesPerSample.spec.ts`. That one tests more combinations. It's not clear if we should delete this test. There are several other issues that are tested twice like this. For now, I think it's fine we have 2 tests since there is so much to do but maybe we should file a bug to remove duplicate tests.

austinEng (approved):

austinEng:
> fyi: g.test('color_attachments,limits,maxColorAttachmentBytesPerSample,unaligned') test only works with 8 attachments (it's skipped with less currently based on the check)
> 
> The test is testing that some attachments, with alignment requirements, exceed `maxColorAttachmentBytesPerSample` and get a validation error, where as if rearranged, they they're within the limit.
> 
> AFAICT there is no combination of 4 attachments (the limit in compat) that can trigger the "exceed limits because of alignment" check.
> 
> Also, maybe we should file a bug for this, but this test is effectively tested twice. This same issue is tested in `maxColorAttachmentBytesPerSample.spec.ts`. That one tests more combinations. It's not clear if we should delete this test. There are several other issues that are tested twice like this. For now, I think it's fine we have 2 tests since there is so much to do but maybe we should file a bug to remove duplicate tests.

worth filing a bug about the duplication. Kai should know best when he is back where the tests should go.

PR #3046, opened 2023-10-04
wgsl: Correctly invoke multiplicationInterval instead impl directly

### LINKED ISSUE #3044: wgsl: scalar-matrix multiplication not correctly rounding operations
Chromium/Dawn is reporting errors that look like this:
```
AssertionError:   - EXPECTATION FAILED: (mat2x2<f32>(-3.4028234663852886e+38 (0xff7fffff), 0.0 (0x80000000 subnormal), 0.0 (0x80000000 subnormal), 0.0 (0x00000000 subnormal)), -1.1754942106924411e-38 (0x807fffff subnormal))
        returned: mat2x2<f32>(-1.1754942106924411e-38 (0x807fffff subnormal), -1.1754942106924411e-38 (0x807fffff subnormal), -1.1754942106924411e-38 (0x807fffff subnormal), -1.1754942106924411e-38 (0x807fffff subnormal))
        expected: [[[{ 'f32', [0.0 (0x80000000 subnormal),3.999999523162842 (0x407ffffe)] }],[{ 'f32', [0.0 (0x80000000 subnormal)] }]],[[{ 'f32', [0.0 (0x80000000 subnormal)] }],[{ 'f32', [0.0 (0x80000000 subnormal)] }]]]
    
    (mat2x2<f32>(-3.4028234663852886e+38 (0xff7fffff), 0.0 (0x80000000 subnormal), 0.0 (0x80000000 subnormal), 0.0 (0x00000000 subnormal)), -1.401298464324817e-45 (0x80000001 subnormal))
        returned: mat2x2<f32>(-1.401298464324817e-45 (0x80000001 subnormal), -1.401298464324817e-45 (0x80000001 subnormal), -1.401298464324817e-45 (0x80000001 subnormal), -1.401298464324817e-45 (0x80000001 subnormal))
        expected: [[[{ 'f32', [0.0 (0x80000000 subnormal),4.7683712978141557e-7 (0x34ffffff)] }],[{ 'f32', [0.0 (0x80000000 subnormal)] }]],[[{ 'f32', [0.0 (0x80000000 subnormal)] }],[{ 'f32', [0.0 (0x80000000 subnormal)] }]]]
    
    (mat2x2<f32>(-3.4028234663852886e+38 (0xff7fffff), 0.0 (0x80000000 subnormal), 0.0 (0x80000000 subnormal), 0.0 (0x00000000 subnormal)), 1.401298464324817e-45 (0x00000001 subnormal))
        returned: mat2x2<f32>(1.401298464324817e-45 (0x00000001 subnormal), 1.401298464324817e-45 (0x00000001 subnormal), 1.401298464324817e-45 (0x00000001 subnormal), 1.401298464324817e-45 (0x00000001 subnormal))
        expected: [[[{ 'f32', [-4.7683712978141557e-7 (0xb4ffffff),0.0 (0x80000000 subnormal)] }],[{ 'f32', [0.0 (0x80000000 subnormal)] }]],[[{ 'f32', [0.0 (0x80000000 subnormal)] }],[{ 'f32', [0.0 (0x00000000 subnormal)] }]]]
```

The critical part here is that we have a mat2x2 * a subnormal scalar. Since that scalar can be FTZ, one of the acceptable answers should be the zero matrix, which in turn should accept elements that are subnormal, since they can be FTZ.

The rejected matrix is all subnormal values.

Looking at the code, it looks like multiplicationMatrixScalarInterval is reaching directly into the multiplicationInterval impl instead of calling multiplicationInterval, which is like the source of this bug, since correctlyRounded will be applied by the scaffolding code called by multiplicationInterval to invoke the impl

zoddicus:
This corrects issues with missing possible FTZ

Fixes #3044

<hr>

### DIFF
--- a/src/unittests/floating_point.spec.ts
+++ b/src/unittests/floating_point.spec.ts
@@ -7574,135 +7574,171 @@ interface MatrixScalarToMatrixCase {
   expected: (number | IntervalBounds)[][];
 }
 
+const kMultiplicationMatrixScalarIntervalCases = {
+  f32: [
+    // From https://github.com/gpuweb/cts/issues/3044
+    {
+      matrix: [
+        [kValue.f32.negative.min, 0],
+        [0, 0],
+      ],
+      scalar: kValue.f32.negative.subnormal.min,
+      expected: [
+        [[0, reinterpretU32AsF32(0x407ffffe)], 0], // [[0, 3.9999995...], 0],
+        [0, 0],
+      ],
+    },
+  ] as MatrixScalarToMatrixCase[],
+  f16: [
+    // From https://github.com/gpuweb/cts/issues/3044
+    {
+      matrix: [
+        [kValue.f16.negative.min, 0],
+        [0, 0],
+      ],
+      scalar: kValue.f16.negative.subnormal.min,
+      expected: [
+        [[0, reinterpretU16AsF16(0x43fe)], 0], // [[0, 3.99609375], 0]
+        [0, 0],
+      ],
+    },
+  ] as MatrixScalarToMatrixCase[],
+} as const;
+
 g.test('multiplicationMatrixScalarInterval')
   .params(u =>
     u
       .combine('trait', ['f32', 'f16'] as const)
       .beginSubcases()
-      .combineWithParams<MatrixScalarToMatrixCase>([
-        // Only testing that different shapes of matrices are handled correctly
-        // here, to reduce test duplication.
-        // multiplicationMatrixScalarInterval uses MultiplicationIntervalOp for calculating intervals,
-        // so the testing for multiplcationInterval covers the actual interval
+      .expandWithParams<MatrixScalarToMatrixCase>(p => {
+        // Primarily testing that different shapes of matrices are handled correctly
+        // here, to reduce test duplication. Additional testing for edge case
+        // discovered in https://github.com/gpuweb/cts/issues/3044.
+        //
+        // multiplicationMatrixScalarInterval uses for calculating intervals,
+        // so the testing for multiplicationInterval covers the actual interval
         // calculations.
-        {
-          matrix: [
-            [1, 2],
-            [3, 4],
-          ],
-          scalar: 10,
-          expected: [
-            [10, 20],
-            [30, 40],
-          ],
-        },
-        {
-          matrix: [
-            [1, 2],
-            [3, 4],
-            [5, 6],
-          ],
-          scalar: 10,
-          expected: [
-            [10, 20],
-            [30, 40],
-            [50, 60],
-          ],
-        },
-        {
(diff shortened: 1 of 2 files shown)


amaiorano (approved):

src/unittests/floating_point.spec.ts
```diff
@@ -7574,135 +7574,169 @@ interface MatrixScalarToMatrixCase {
...
     u
       .combine('trait', ['f32', 'f16'] as const)
       .beginSubcases()
-      .combineWithParams<MatrixScalarToMatrixCase>([
+      .expandWithParams<MatrixScalarToMatrixCase>(p => {
         // Only testing that different shapes of matrices are handled correctly
         // here, to reduce test duplication.
         // multiplicationMatrixScalarInterval uses MultiplicationIntervalOp for calculating intervals,
-        // so the testing for multiplcationInterval covers the actual interval
+        // so the testing for multiplicationInterval covers the actual interval
```
jzm-intel, line 7617: So it seems that the assumption in this comment doesn't stand (being an implementation bug), because the bug is considered caused by misusing `multiplicationInterval` without calling it but actually call its implementation directly.

Maybe we can add to this comment with some note like `Some subnormal cases are also tested to ensure that inputs are correctly flushed when calling multiplicationInterval.`?

I am not sure what to do to help avoid similar problem. I looks through the `floating_point.ts`, and the only other interval implementation that calls other implementation (`.impl`) is `ldexpIntervalImpl`, and it explicitly handles the flushing (hopefully it does it right)

jzm-intel (approved):
Looks good to me, thanks

src/unittests/floating_point.spec.ts
zoddicus: I have added to the comment language calling out that additional testing is occuring, and pointing to the original bug as the reason why.

I am not sure of a good way to avoid these issues, since the fundamental issue from my perspective is that the various run*Op functions do manipulation of the inputs and then need to invoke the *Op.impl functions, but we want things that are not the run*Op functions to be prevented from invoking the impls.

Currently they are all members of the same class, so there isn't something simple like changing access qualifiers that can be done. We could decompose the FP classes into different functional units so that the API, Impl, and Run portions are in different classes, so that we could use OO relationships to limit accidental misuse. I am not sure off hand if all of that refactoring and additional complexity is worth it.

I am not sure in TS if there is a better way to control the invokability of the impls. Like being able to inspect the calling function to make sure it is a run*Op (I know older JS allowed you to reflect the caller in a function, but I am pretty sure that is deprecated.)

zoddicus: I think ldexp is fine, since it is explicitly invoking the roundAndFlush code that run*Op would.

PR #3054, opened 2023-10-05
Make Pass > Skip

greggman:
As it is, the test recorder only keeps the "highest" status. It was Pass = 0, Skip = 1, that means if 1000 subcases passes but just one was skipped the entire case was marked as skip.

Making Skip less than Pass means they'll as long as one subcase runs the test will be marked as Pass unless there is higher a higher status subcase

Added tests, also tested locally via telemetry and it seems to work.

This is a fix for the revert: https://github.com/gpuweb/cts/pull/3043

<hr>

### DIFF
--- a/src/common/internal/logging/test_case_recorder.ts
+++ b/src/common/internal/logging/test_case_recorder.ts
@@ -3,27 +3,43 @@ import { globalTestConfig } from '../../framework/test_config.js';
 import { now, assert } from '../../util/util.js';
 
 import { LogMessageWithStack } from './log_message.js';
-import { Expectation, LiveTestCaseResult } from './result.js';
+import { Expectation, LiveTestCaseResult, Status } from './result.js';
 
 enum LogSeverity {
-  Pass = 0,
+  NotRun = 0,
   Skip = 1,
-  Warn = 2,
-  ExpectFailed = 3,
-  ValidationFailed = 4,
-  ThrewException = 5,
+  Pass = 2,
+  Warn = 3,
+  ExpectFailed = 4,
+  ValidationFailed = 5,
+  ThrewException = 6,
 }
 
 const kMaxLogStacks = 2;
 const kMinSeverityForStack = LogSeverity.Warn;
+
+function logSeverityToString(status: LogSeverity): Status {
+  switch (status) {
+    case LogSeverity.NotRun:
+      return 'notrun';
+    case LogSeverity.Pass:
+      return 'pass';
+    case LogSeverity.Skip:
+      return 'skip';
+    case LogSeverity.Warn:
+      return 'warn';
+    default:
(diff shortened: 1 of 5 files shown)


greggman:
Note: an example of a bug this fixes (you'll need --enable-webgpu-developer-feature in Chrome)

This test gets the error: EXCEPTION: internal error: case is "skip" but has nonskipped subcases

https://gpuweb.github.io/cts/standalone/?runnow=1&compatibility=1&q=webgpu:api,validation,encoding,cmds,copyTextureToTexture:texture_format_compatibility:srcFormatFeature=%22_undef_%22;dstFormatFeature=%22_undef_%22

Where as with this change the error goes away, as well as the skipped issue mentioned above.

src/common/internal/logging/test_case_recorder.ts
```diff
@@ -56,20 +72,13 @@ export class TestCaseRecorder {
...
     // Convert numeric enum back to string (but expose 'exception' as 'fail')
-    this.result.status =
-      this.finalCaseStatus === LogSeverity.Pass
-        ? 'pass'
-        : this.finalCaseStatus === LogSeverity.Skip
-        ? 'skip'
-        : this.finalCaseStatus === LogSeverity.Warn
-        ? 'warn'
-        : 'fail'; // Everything else is an error
+    this.result.status = logSeverityToString(this.finalCaseStatus);
```
austinEng, line 75: Could the final result status ever be "notrun" ?

It might be weird to ever see that reported. We could consider asserting it is never notrun 

In Chrome, telemetry will silently treat this as Pass. Chrome checks for "skip" and "fail". Other stuff is considered pass (can be changed of course).

src/unittests/logger.spec.ts
```diff
@@ -59,13 +71,27 @@ g.test('skip').fn(t => {
...
   rec.start();
   rec.skipped(new SkipTestCase());
-  rec.debug(new Error('hello'));
   rec.finish();
 
   t.expect(res.status === 'skip');
   t.expect(res.timems >= 0);
 });
 
+// Tests if there's some skips and at last one pass it's pass.
```
austinEng, line 80:
```suggestion
// Tests if there's some skips and at least one pass it's pass.
```

austinEng (approved):

src/common/internal/logging/test_case_recorder.ts
greggman: It is 'notrun' if you don't run anything and this is tested in `logger.spec.ts`

We should probably fix telemetry. Assuming pass doesn't seem like a good default.

src/unittests/logger.spec.ts
greggman: fixed

Issue #3057, opened 2023-10-06
framework: add ability filter device based on WGSLLanguageFeatures


dneto0:
This is necessary to be able to conditionally test for WGSL langauge extensions. See https://gpuweb.github.io/gpuweb/#gpuwgsllanguagefeatures 

This gates #3056

Issue #3056, opened 2023-10-06
wgsl: test unrestricted_pointer_parameters feature


dneto0:
https://github.com/gpuweb/gpuweb/pull/4312 

Example code:

```
// enable chromium_experimental_full_ptr_parameters;  // needed for Chrome.

@group(0) @binding(0) var<storage,read_write> a: f32;
@group(0) @binding(0) var<storage> b: f32;

const count: u32 = 12;
var<workgroup> w: array<atomic<u32>,count>;

fn mutf(p: ptr<storage,f32>) -> f32 {
  return *p; 
}

fn getw(p: ptr<workgroup,atomic<u32>>) -> u32 {
  return atomicLoad(p);
}

@compute @workgroup_size(64)
fn main() {
  // _ = mutf(&a); // invalide due to read_write access mode
  _ = mutf(&b);
  let u: u32 = getw(&w[8]);
}
```

dneto0:
Is this done @ben-clayton ?

ben-clayton:
Yup

PR #3058, opened 2023-10-06
framework: allow skipping tests based on wgslLanguageFeatures

dneto0:
Issue: #3057

This is in preparation for #3056 

<hr>


dneto0 (commented):
I haven't tested the case where this actually skips the test.

src/webgpu/util/device_pool.ts
```diff
@@ -296,9 +307,13 @@ class DeviceHolder implements DeviceProvider {
   // If the device is lost, DeviceHolder.lost gets set.
   static async create(
     recorder: TestCaseRecorder,
-    descriptor: CanonicalDeviceDescriptor | undefined
+    descriptor: CanonicalDeviceDescriptor | undefined,
+    langFeature: WGSLLanguageFeature | undefined
   ): Promise<DeviceHolder> {
     const gpu = getGPU(recorder);
+    if (langFeature && gpu.wgslLanguageFeatures.has(langFeature)) {
```
dneto0, line 314: I haven't checked that this triggers.

kainino0x (commented):
It's surprising to me to wire wgslLanguageFeatures through DeviceSelectionDescriptor when it's never needed for device selection. I'm not sure how this is intended to be used - could we just have a `requireLanguageFeatureOrSkipTestCase` (that just calls `this.skip()` alongside `selectDeviceOrSkipTestCase` in `gpu_test.ts`?

src/webgpu/gpu_test.ts
```diff
@@ -245,6 +247,23 @@ export class GPUTestSubcaseBatchState extends SubcaseBatchState {
       }
     }
   }
+
+
+  /** Skips this test case if the `langFeature` is *not* supported. */
+  requireLanguageFeatureOrSkipTestCase(langFeature: WGSLLanguageFeature) {
+    const lf = getGPU(this.recorder).wgslLanguageFeatures;
+    if (!(langFeature in lf)) {
```
dneto0, line 255: Using      lf.has(langFeature) failed with an UNIMPLEMENTED exception.
This way of doing it properly skips or does not skip.

src/webgpu/wgsl_language_feature.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * @fileoverview Declares a type for each string tested against
+ * the WGSLLanguageFeatures property of the GPU object.
+ */
```
kainino0x, line 4: Unfortunately I don't think TSDoc supports file-level documentation. This is getting attached to `kKnownWGSLLanguageFeatures` (visible in vscode or `docs/tsdoc`).

Looking into this it seems like TSDoc is kind of dying and we should probably use JSDoc? I'll attempt this sometime.

src/webgpu/wgsl_language_feature.ts
```diff
@@ -0,0 +1,12 @@
+/**
```
kainino0x, line 1: Usually we just add new things like this to `capability_info.ts`. This will stay small enough that I'd say just put it there.

src/webgpu/gpu_test.ts
kainino0x: wgslLanguageFeatures appears to be incorrectly implemented in dawn.node.
`langFeature in lf` should always return false because that's just checking for a key with a name equal to `langFeature` in the JS object `lf`. But that's just the interface to the set, so it only has keys like `has`, etc.
`lf.has(langFeature)` is the correct way.

Also we may want to skip if `wgslLanguageFeatures` doesn't exist (to make life easier for incomplete WebGPU implementations). In this case `lf` would be `undefined`.

src/resources/cache/hashes.json
```diff
@@ -1,106 +1,106 @@
 {
```
kainino0x, line 1: It makes sense that these hash files change but the binary files should not have. See the bold thing here: https://github.com/gpuweb/cts/blob/main/docs/case_cache.md#file-cache

If this is all it took to trigger this - and I should have realized it would be - then I think we're going to need to fix that issue sooner rather than later. @ben-clayton

src/webgpu/gpu_test.ts
```diff
@@ -245,6 +247,22 @@ export class GPUTestSubcaseBatchState extends SubcaseBatchState {
       }
     }
   }
+
+  /** Skips this test case if the `langFeature` is *not* supported. */
+  requireLanguageFeatureOrSkipTestCase(langFeature: WGSLLanguageFeature) {
```
kainino0x, line 252: On second thought (re: my comment from October), this should probably be on GPUTest and not GPUTestSubcaseBatchState.

The reason for this is that, unlike selectDeviceOrSkipTestCase, this doesn't have to be done before getting a GPUDevice - wgslLanguageFeatures is static.

kainino0x: GPUTestBase, actually, not GPUTest. It can go next to skipIfTextureFormatNotSupported.

src/resources/cache/hashes.json
ben-clayton: https://github.com/gpuweb/cts/pull/3187

src/webgpu/gpu_test.ts
dneto0: Thanks. fixed now. Tested with Chrome.
At least the "unsupported" test does trigger and detect the compilation failures.

src/webgpu/wgsl_language_feature.ts
```diff
@@ -0,0 +1,12 @@
...
+ * @fileoverview Declares a type for each string tested against
+ * the WGSLLanguageFeatures property of the GPU object.
+ */
+
+export const kKnownWGSLLanguageFeatures = [
+  'readonly_and_readwrite_storage_textures',
+  'packed_4x8_integer_dot_product',
+  'unrestricted_pointer_parameters',
+];
```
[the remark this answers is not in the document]
dneto0: committed

kainino0x (commented):
LGTM, but please undo all of the .bin changes - none of those files actual (decompressed) content should have changed due to this patch.

dneto0:
> LGTM, but please undo all of the .bin changes - none of those files actual (decompressed) content should have changed due to this patch.

I've squashed the primary content, removed the .bin changes, and updated the hashes.
(The CI failed before I updated the caches)

dneto0:
> > LGTM, but please undo all of the .bin changes - none of those files actual (decompressed) content should have changed due to this patch.
> 
> I've squashed the primary content, removed the .bin changes, and updated the hashes. (The CI failed before I updated the caches)

I'm installing node v20 to see if that makes a difference

kainino0x (approved):

PR #3080, opened 2023-10-23
wgsl: Add AbstractFloat `round` execution tests

### LINKED ISSUE #2509: wgsl: AbstractFloat execution test for `round` builtin
Add an AbstractFloat execution test for the `round` builtin.
```
@const @must_use fn round(e: T) -> T
S is AbstractFloat
T is S or vecN<S>
```
Result is the integer k nearest to e, as a floating point value.
When e lies halfway between integers k and k + 1, the result is k when k is even, and k + 1 when k is odd.
[Component-wise](https://www.w3.org/TR/WGSL/#component-wise) when T is a vector.

Accuracy is `Correctly rounded`

See https://www.w3.org/TR/WGSL/#round-builtin

zoddicus:
Fixes #2509

<hr>

### DIFF
--- a/src/webgpu/shader/execution/expression/call/builtin/round.spec.ts
+++ b/src/webgpu/shader/execution/expression/call/builtin/round.spec.ts
@@ -12,13 +12,18 @@ Component-wise when T is a vector.
 
 import { makeTestGroup } from '../../../../../../common/framework/test_group.js';
 import { GPUTest } from '../../../../../gpu_test.js';
-import { TypeF32, TypeF16 } from '../../../../../util/conversion.js';
+import { TypeF32, TypeF16, TypeAbstractFloat } from '../../../../../util/conversion.js';
 import { FP } from '../../../../../util/floating_point.js';
-import { fullF32Range, fullF16Range } from '../../../../../util/math.js';
+import {
+  fullF32Range,
+  fullF16Range,
+  reinterpretU64AsF64,
+  fullF64Range,
+} from '../../../../../util/math.js';
 import { makeCaseCache } from '../../case_cache.js';
-import { allInputSources, run } from '../../expression.js';
+import { allInputSources, onlyConstInputSource, run } from '../../expression.js';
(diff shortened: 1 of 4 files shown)


src/webgpu/shader/execution/expression/call/builtin/round.spec.ts
```diff
@@ -43,15 +48,30 @@ export const d = makeCaseCache('round', {
       FP.f16.roundInterval
     );
   },
+  abstract: () => {
+    return FP.abstract.generateScalarToIntervalCases(
+      [
+        reinterpretU64AsF64(0x8000_0000_0000_0000n), // https://github.com/gpuweb/cts/issues/2766
```
dneto0, line 54: This is a reinterpret cast, so this generates  -0.0 in f64.

I don't think that's what you mean.

dneto0 (changes requested):
I think the "large int value" case is not exercised the way it was intended.

src/webgpu/shader/execution/expression/call/builtin/round.spec.ts
dneto0: The point of 2766 was to test integers that are larger in magnitude than can fit inside the mantissa bits.  So for f64 that would be something in excess of 2** 53 ish.

zoddicus: Correct, thanks for catching this.

zoddicus: arg... I updated the unittests, but not the CTS test itself :face_exhaling:

zoddicus: Updated

zoddicus:
@dneto0 PTAL, I think this should be fixed now

jzm-intel (approved):
Looks good to me, thanks

PR #3082, opened 2023-10-23
wgsl: Cleanup cruft related to quantizeToF16

zoddicus:
This is only defined for f32, so doesn't really need to be defined in the other classes that will never implement it.

<hr>

### DIFF
--- a/src/unittests/floating_point.spec.ts
+++ b/src/unittests/floating_point.spec.ts
@@ -3423,7 +3423,7 @@ g.test('negationInterval')
     );
   });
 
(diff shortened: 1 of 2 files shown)


dneto0 (approved):
LGTM.

I saw there was an assertion for code that would have invoked a test for quantizeToF16 for  a non-f32 based argument. But that seems to not be exercised.
I concluded this was fine because the assert itself was never triggered (otherwise the TS copmijle would fail).  It would be some *other* test  or MR that would check for a validation failure. So, after all, this is fine.

PR #3083, opened 2023-10-23
Fix a bad slice operation in image_copy stencil tests

toji:
Found that in the `checkStencilTextureContent()` method of the `image_copy` tests an `ArrayBuffer.slice()` is being used that was passing arguments as if they were `(offset, length)`, when in fact they must be `(start, end)`. This was causing any tests with a buffer offset other than 0 to pass even where there was an error because the arrays that were being compared were zero length.

After correcting this issue I also found that a separate buffer offset needed to be passed to the `expectGPUBufferValuesEqual()` method to account for the offset of the row being tested. This was preventing anything other than the first row from being compared properly, though that didn't cause an error due to the aforementioned `slice()` issue.

Once these two fixes are in place the test can now properly check all subcases. Confirmed passing on Chrome.

<hr>


kainino0x (approved):

toji:
Oh, and for what it's worth I did a quick scan of other instances of `slice()` in the CTS codebase. None of the other uses appear to have this same argument mixup. (It's mostly called without arguments to clone the array)

PR #3089, opened 2023-10-24
wgsl: Filter `atan2` tests based on if const-eval or not

### LINKED ISSUE #3088: wgsl: atan2 execution tests should filter test cases if doing const-eval
Currently the generators are passing unfiltered cases in, which becomes a problem for atan2's unique corner cases, i.e. y = subnormal returning unbounded, since for const-eval this will cause an execution error.

This is probably a hold over from all of the revisions that atan2 and unbounded handling has gone through. There might be other examples of this.

This was found by Chromium, https://bugs.chromium.org/p/tint/issues/detail?id=2070

zoddicus:
Rewrites how test cases are generated for atan2, so that if running in const-eval unbounded results will not be generated, since those will cause compilation errors.

Fixes #3088

<hr>

### DIFF
--- a/src/webgpu/shader/execution/expression/call/builtin/atan2.spec.ts
+++ b/src/webgpu/shader/execution/expression/call/builtin/atan2.spec.ts
@@ -9,7 +9,6 @@ Returns the arc tangent of e1 over e2. Component-wise when T is a vector.
 
 import { makeTestGroup } from '../../../../../../common/framework/test_group.js';
 import { GPUTest } from '../../../../../gpu_test.js';
-import { kValue } from '../../../../../util/constants.js';
 import { TypeF32, TypeF16 } from '../../../../../util/conversion.js';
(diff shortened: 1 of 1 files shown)


jzm-intel (approved):
Looks good to me, thanks

jzm-intel:
Just wonder, but why I don't get error on constant cases with current implementation?

zoddicus:
> Just wonder, but why I don't get error on constant cases with current implementation?

I think it is only some specific devices that have an issue with y == 0, or near 0 (and other boundary conditions). It is related to if they do FTZ and how they implement atan2. It is only recently with adding Android bots to Chrome that we regularly saw runs that were failing because of this.

That being said, this implies that some bots are being too tolerant of bad inputs, since these should have been failing in general. validation tests should be catching when const-eval is being too tolerant, but that coverage is yet to be implemented.

PR #3098, opened 2023-10-26
Compat: Test textureBindingViewDimension validation

greggman:
Tests that if textureBindingViewDimension is incompatible with a texture's dimension in createTexture a validation error is generated

Tests that if a texture view's dimension does not match the texture's textureBindingViewDimension when passed to createBindGroup a validation error is generated.

<hr>


greggman:
These pass in Chrome with the proper validating compiled in.

src/webgpu/compat/api/validation/createBindGroup.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+import { CompatibilityTest } from '../../compatibility_test.js';
+
+export const g = makeTestGroup(CompatibilityTest);
+
+function isTextureBindingViewDimensionCompatibleWithDimension(
+  dimension?: GPUTextureDimension,
+  textureBindingViewDimension?: GPUTextureViewDimension
+) {
+  dimension = dimension || '2d';
+  textureBindingViewDimension = textureBindingViewDimension || '2d';
```
kainino0x, line 20:
```suggestion
  dimension: GPUTextureDimension = '2d',
  textureBindingViewDimension: GPUTextureViewDimension = '2d'
) {
```

src/webgpu/compat/api/validation/createBindGroup.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+      )
+  )
+  .fn(t => {
+    const { dimension, textureBindingViewDimension, viewDimension, depthOrArrayLayers } = t.params;
+
+    const texture = t.device.createTexture({
+      size: [1, 1, depthOrArrayLayers],
+      format: 'rgba8unorm',
+      usage: GPUTextureUsage.TEXTURE_BINDING,
+      ...(dimension && { dimension }),
```
kainino0x, line 139: weird, but I'm going to have to file this one in my book of JS tricks.

We usually don't bother making sure unused keys won't be set to `undefined` - webidl `dictionary` guarantees that `{}` and `{x: undefined}` are the same. IMO probably not necessary here, but OK.

src/webgpu/compat/api/validation/createBindGroup.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+    case '2d':
+    case undefined:
+      return depthOrArrayLayers > 1 ? '2d-array' : '2d';
+      break;
+    case '3d':
+      return '3d';
+  }
+}
+
+g.test('dimension_matches_textureBindingViewDimension')
```
kainino0x, line 104: iiuc
```suggestion
g.test('viewDimension_matches_textureBindingViewDimension')
```

src/webgpu/compat/api/validation/createBindGroup.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+        },
+      ],
+    });
+
+    const shouldError = effectiveTextureBindingViewDimension !== effectiveViewDimension;
+
+    t.expectValidationError(() => {
+      t.device.createBindGroup({
+        layout,
+        entries: [{ binding: 0, resource: texture.createView({ dimension: viewDimension }) }],
```
kainino0x, line 173: It's important especially for this test to check whether the error comes from createView or createBindGroup. I'm assuming createBindGroup, so the createView should be moved out of the expectValidationError.

src/webgpu/compat/api/validation/createBindGroup.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+      texture.depthOrArrayLayers,
+      viewDimension
+    );
+
+    const layout = t.device.createBindGroupLayout({
+      entries: [
+        {
+          binding: 0,
+          visibility: GPUShaderStage.COMPUTE,
+          texture: {
```
kainino0x, line 161: Would be good to verify this isn't enforced for storage textures

src/webgpu/util/texture/base.ts
```diff
@@ -157,22 +157,52 @@ export function viewDimensionsForTextureDimension(textureDimension: GPUTextureDi
...
     case '3d':
       return '3d';
     default:
       unreachable();
   }
 }
 
+/** Returns the effective view dimension for a given texture */
+export function effectiveViewDimensionForTexture(
+  texture: Readonly<GPUTexture>,
```
kainino0x, line 186: nit: GPUTexture should be inherently readonly, I think
```suggestion
  texture: GPUTexture,
```

kainino0x (approved):
Sorry for the delay.

src/webgpu/compat/api/validation/createBindGroup.spec.ts
greggman: The storage textures should (already?) be tested by core. The compat specific tests only test that things not allowed are validated out in compat. For things that still work they should be tested already (or in any case belong in the storage texture tests... right?)

PR #3113, opened 2023-10-30
binary_stream: Simplify away a no-op TypedArray

kainino0x:
I just noticed this while looking around for TypedArray allocations that could be cleaned up. The code here didn't match the intent. This change is a no-op, but maybe there is a bug to be fixed instead.

I only verified the unittests pass, so far, since I figured I should send this to you first.

Issue: None

<hr>

### DIFF
--- a/src/webgpu/util/binary_stream.ts
+++ b/src/webgpu/util/binary_stream.ts
@@ -18,9 +18,9 @@ export default class BinaryStream {
     this.view = new DataView(buffer);
   }
 
-  /** buffer() returns the stream's buffer sliced to the 8-byte rounded read or write offset */
+  /** buffer() returns the stream's whole buffer */
   buffer(): ArrayBufferLike {
-    return new Uint8Array(this.view.buffer, align(this.offset, 8)).buffer;
+    return this.view.buffer;
   }
 
   /** writeBool() writes a boolean as 255 or 0 to the buffer at the next byte offset */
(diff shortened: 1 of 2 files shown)


src/webgpu/util/binary_stream.ts
```diff
@@ -18,9 +18,9 @@ export default class BinaryStream {
     this.view = new DataView(buffer);
   }
 
-  /** buffer() returns the stream's buffer sliced to the 8-byte rounded read or write offset */
+  /** buffer() returns the stream's whole buffer */
   buffer(): ArrayBufferLike {
-    return new Uint8Array(this.view.buffer, align(this.offset, 8)).buffer;
```
kainino0x, line 23: `new Uint8Array` wraps a view around `this.view.buffer`. `Uint8Array.buffer` returns the underlying `ArrayBuffer`. Therefore this was the same as simply `this.view.buffer`.

zoddicus (approved):

src/webgpu/util/binary_stream.ts
ben-clayton: This was supposed to be `new Uint8Array(this.view.buffer, 0, align(this.offset, 8)).buffer` as in: truncate the array to the written size. I actually noticed this bug yesterday, as all the cached files are 33.6MB (!). Fortunately tar.gz does a good job removing all those 0's.

Issue #3049, opened 2023-10-05
memory_model tests use 6 storage buffers but compat has a default limit of 4


greggman:
Thoughts?

I haven't dug through this code to see what it's doing.

* Would it be not too much work to refactor them use 4 storage buffers instead of 6?  Like if 3 are inputs then just merge them. Would it still be testing the same thing?
* Should we just skip them in compat?
* Should we make them request 6 storage buffers so they run in compat IF the device supports 6
* Should we raise the compat limit? Not sure we can, limits appear to be 4 for vertex, 4 for fragment, 8 for compute. Maybe we could split the limits in WebGPU?

https://github.com/gpuweb/cts/blob/17c90bca804385a5a959196e10c5229571434d16/src/webgpu/shader/execution/memory_model/memory_model_setup.ts#L256

greggman:
@SenorBlanco

dneto0:
I own these tests now, and am currently iterating on them. (They were contributed by UCSC @reeselevine, in a really nice piece of work).

dneto0:
These tests only run compute shaders. So maybe there's not issue?

greggman:
We don't have limits separate for compute vs fragment vs vertex in WebGPU so the limit is currently the min of each of the GL limits which is 4. 

In other words `maxStorageBuffersPerShaderStage` defaults to 4 on compat

reeselevine:
I don't think there's any technical reason some of the buffers couldn't be merged, at the expense of some readability.

greggman:
I'm not sure the best way to approach this. As it is, several arrays are dynamic and different sizes so I can't just interleave their data into a struct and make one dynamic array. I guess I could make the array as large as the largest number of elements needed. In other words, if it was doing 

```
    a: array(TypeA, 10)
    b: array(TypeB, 20)
    c: array(TypeC, 15)
```

Then I could change it to

```
    struct TypeABC {
      a: TypeA,
      b: TypeB,
      c: TypeC,
    }

    abc: array(TypeABC, 20)
```

And code that was using `a: array(TypeA)` will just not use the extra 10 elements

Or, going back to the first example

```
    a: array(TypeA, 10)
    b: array(TypeB, 20)
    c: array(TypeC, 15)
```

I could change to

```
   struct ABC {
     a: array(typeA, 10),
     b: array(typeB, 20),
     c: array(typeC, 15),
   }
   abc: ABC;  // no array here
```

[...]

PR #3117, opened 2023-10-31
Compat: Refactor memory module tests to use 4 storage buffers

greggman:
So ... maybe this test was written against older WGSL but it's not clear to me why the code was

```
struct IndexMemory {
  value: array<u32>,
};

@group(0) @binding(2) var<storage, read_write> foo: IndexMemory;
```

Instead of just

```
@group(0) @binding(2) var<storage, read_write> foo: array<u32>;
```

But, because it was already that way, the smallest change seemed to be to leave it that way. I can go change them all to get rid of the struct and just use arrays directly. Access would switch from `combo.thing.value[ndx]` to `combo.thing[ndx]`.

Similarly, it's unclear why `barrier` is defined as an array and not just a single value. I could change that as well assuming there isn't some reason it's an array.

Also, compat has a maxWorkgroupSizeX of 128.  I switched the code to use 128 in compat and didn't change any other code. It passes. I'd have expected to have to change something else to say "only check 128 results" or something to that effect. That makes me concerned the test is not actually testing anything. I didn't dig though.

Issue: #3049

<hr>


greggman:
ping

src/webgpu/shader/execution/memory_model/memory_model_setup.ts
```diff
@@ -191,47 +202,59 @@ export class MemoryModelTester {
...
+      const kNumBarriers = 1u;  // MAINTENANCE_TODO: make barrier not an array
+      const kMaxWorkgroups = ${params.maxWorkgroups}u;
+      const kScratchMemorySize = ${params.scratchMemorySize}u;
+      const kWorkgroupXSize = ${Math.min(256, t.device.limits.maxComputeWorkgroupSizeX)}u;
+    `;
+    testShader = shaderPreamble(accessValueType, constants) + testShader;
+    resultShader = shaderPreamble(accessValueType, constants) + resultShader;
 
     // set up buffers
     const testingThreads = this.params.workgroupSize * this.params.testingWorkgroups;
```
dneto0, line 215: Please use the same workgroup size throughout.  This now differs from the tweaked value set at line 209 for `kWorkgroupSize`.

src/webgpu/shader/execution/memory_model/memory_model_setup.ts
```diff
@@ -249,52 +272,60 @@ export class MemoryModelTester {
       size: shuffledWorkgroupsSize,
     };
 
+    // Combine 3 arrays into 1 buffer as we need to keep the number of storage buffers to 4 for compat.
```
dneto0, line 275: The three logical arrays are jammed up right beside each other.

This induces false sharing between the combo.barrier.value[0] location and the  scratchpad beside it.
These tests were developed as part of research on how to heuristically evoke weak behaviours to more effectively test memory models.  https://dl.acm.org/doi/10.1145/3575693.3575750 
So I'd like to avoid adding false sharing as that would likely perturb the results in a meaningful way.

Please quantize the sizes so they are unlikely to have overlapping cache lines.
E.g.    
```js
const falseSharingAvoidanceQuantum = 4096;
const barrierSize = Math.max(bytesPerWord, falseSharingAvoidanceQuantum);
const scratchpadSize = Math.max(this.params.scratchMemorySize * bytesPerWord, falseSharingAvoidanceQuantum);
const scratchMemoryLocationsSize = Math.max(this.params.maxWorkgroups * bytesPerWord, falseSharingAvoidanceQuantum);
```

dneto0 (changes requested):
Thanks for taking this on.
Thank you for your patience as well.

I'm requesting two changes:
- use the same tweaked workgroup throughout
- avoid false sharing between the sub-buffers.

dneto0:
> So ... maybe this test was written against older WGSL but it's not clear to me why the code was

It was written quite some time ago.  Originated here: https://gpuharbor.ucsc.edu/webgpu-mem-testing/ 

I'm not particularly bothered, but thank you for avoiding style-based rewrites in this PR.

dneto0:
> So ... maybe this test was written against older WGSL but it's not clear to me why the code was
> 
> ```
> struct IndexMemory {
>   value: array<u32>,
> };
> 
> @group(0) @binding(2) var<storage, read_write> foo: IndexMemory;
> ```
> 
> Instead of just
 
> Also, compat has a maxWorkgroupSizeX of 128. I switched the code to use 128 in compat and didn't change any other code. It passes. I'd have expected to have to change something else to say "only check 128 results" or something to that effect. That makes me concerned the test is not actually testing anything. I didn't dig though.

For the long answer, read the "MC Mutants" paper: https://dl.acm.org/doi/10.1145/3575693.3575750 

The short answer is @reeselevine did experiments to heuristically search for ways to evoke weak memory behaviours via various techniques (read the paper).  The 256 is basicaly arbitrary.  Changing it to 128 is still *valid* bit pushes it into an unexamined regime. My intuition is the test is probably as effective as before, or close to it.   IMHO it's good to keep it 256 for baseline WebGPU and limit it to 128 for the compat case.  It's fine, no sweat.

Does it do anything?  Yes. 
For example go to https://gpuharbor.ucsc.edu/webgpu-mem-testing/tests/message-passing/ and hit the "Start Test" button at the bottom.   You are likely to see weak behaviours exhibited (blue and red bars). That's success. It's doing something.

src/webgpu/shader/execution/memory_model/memory_model_setup.ts
greggman: done?

src/webgpu/shader/execution/memory_model/memory_model_setup.ts
greggman: Is `max` what you wanted here or should it be `align` as in

const barrierSize = align(bytesPerWord, falseSharingAvoidanceQuantum);
const scratchpadSize = align(this.params.scratchMemorySize * bytesPerWord, falseSharingAvoidanceQuantum);
const scratchMemoryLocationsSize = align(this.params.maxWorkgroups * bytesPerWord, falseSharingAvoidanceQuantum);

I put `max` in for now.

dneto0: You're right align would be better.  I didn't look hard enough at `util/math.ts`.   :-)

src/webgpu/shader/execution/memory_model/memory_model_setup.ts
```diff
@@ -191,19 +202,28 @@ export class MemoryModelTester {
     this.test = t;
     this.params = params;
 
-    testShader = shaderPreamble(accessValueType) + testShader;
-    resultShader = shaderPreamble(accessValueType) + resultShader;
+    const workgroupXSize = Math.min(params.workgroupSize, t.device.limits.maxComputeWorkgroupSizeX);
```
dneto0, line 205: Interestingly, the OpenCL CTS also recently had to be updated to accommodate devices that only handle up to 128-size workgroups.

See https://github.com/KhronosGroup/OpenCL-CTS/pull/1846

dneto0 (approved):
LGTM.
Your choice whether to use `align` instead of `Math.max`.  align would be better.

thanks

PR #3120, opened 2023-10-31
wgsl: Convert `quantizeToI32/U32` to use `Math.trunc`

zoddicus:
Another small bump (~5%) to be gained through using a builtin instead of trampolining through a TypedArray.

<hr>

### DIFF
--- a/src/webgpu/util/math.ts
+++ b/src/webgpu/util/math.ts
@@ -2023,22 +2023,32 @@ export function quantizeToF16(num: number): number {
   return hfround(num);
 }
 
-/** Statically allocate working data, so it doesn't need per-call creation */
-const quantizeToI32Data = new Int32Array(new ArrayBuffer(4));
-
-/** @returns the closest 32-bit signed integer value to the input */
+/**
+ * @returns the closest 32-bit signed integer value to the input, rounding
+ * towards 0, if not already an integer
+ */
 export function quantizeToI32(num: number): number {
-  quantizeToI32Data[0] = num;
-  return quantizeToI32Data[0];
+  if (num >= kValue.i32.positive.max) {
+    return kValue.i32.positive.max;
+  }
+  if (num <= kValue.i32.negative.min) {
+    return kValue.i32.negative.min;
+  }
+  return Math.trunc(num);
 }
 
-/** Statically allocate working data, so it doesn't need per-call creation */
-const quantizeToU32Data = new Uint32Array(new ArrayBuffer(4));
-
-/** @returns the closest 32-bit signed integer value to the input */
+/**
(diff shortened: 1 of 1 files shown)


src/webgpu/util/math.ts
```diff
@@ -2030,22 +2030,26 @@ export function quantizeToF16(num: number): number {
...
 export function quantizeToI32(num: number): number {
-  quantizeToI32Data[0] = num;
-  return quantizeToI32Data[0];
+  if (num >= kValue.i32.positive.max) {
+    return kValue.i32.positive.max;
+  }
+  if (num <= kValue.i32.negative.min) {
+    return kValue.i32.negative.min;
+  }
+  return Math.round(num);
```
ben-clayton, line 2041: Despite the function's comment, `quantizeToI32` wasn't rounding to closest, but was rounding towards zero:

```js
const F = 0.7;

const quantizeToI32Data = new Int32Array(new ArrayBuffer(4));
quantizeToI32Data[0] = F;
var message = `${quantizeToI32Data[0]}, ${Math.round(F)}`;

console.log(message)
```

This is a change of behaviour. Have you checked that all the places that call this expect the new behaviour?

zoddicus: Changed these to use Math.trunc to replicate the behaviour. I have run all of the tests under webgpu:shader,execution,expression,* and the unittests, so I think this should be fine.

jzm-intel:
The [spec](https://gpuweb.github.io/gpuweb/wgsl/#floating-point-conversion:~:text=In%20other%20words%2C%20floating%20point%20to%20integer%20conversion%20rounds%20toward%20zero%2C%20then%20saturates%20in%20the%20target%20type.) said `In other words, floating point to integer conversion rounds toward zero, then saturates in the target type.`

On the other hand, if all tests are passed no matter using `Math.round` or `Math.trunc`, does it imply that some tests on this fp-int conversion is missing?

ben-clayton (approved):
Approving this change, but I agree with:

> On the other hand, if all tests are passed no matter using `Math.round` or `Math.trunc`, does it imply that some tests on this fp-int conversion is missing?

It sounds like we could do with some more unit tests, and probably some more CTS cases.

zoddicus:
Looking at the call sites of these functions, since it has been awhile since I worked with this code. They are only being used as filtering/cleanup of initial input values, i.e. I want you to run CTS on this set of numbers, and these functions are used to fix up the values to be the correct type of integer. They are not being used to adjust things internally in the framework to make sure that results comply with the WGSL rounding rules.

This is why the specific rounding rule doesn't affect if the test is passing, changing it just changes the set of inputs being tested on, not the expected result for a specific value.

PR #3122, opened 2023-10-31
Reduce draw validation test parameterization

kainino0x:
Draw validation tests were taking a huge chunk (maybe 25%) of the time of the whole api,validation suite. This should reduce their time by roughly 80-90%.

It may be useful to look at each commit separately to have a good sense of how much they're reducing the test coverage by. My estimate is that we still probably have all of the coverage we need.

cc @jzm-intel in case you have opinions about these tests.

Issue: None

<hr>

### DIFF
--- a/src/webgpu/api/validation/encoding/cmds/render/draw.spec.ts
+++ b/src/webgpu/api/validation/encoding/cmds/render/draw.spec.ts
@@ -426,7 +426,6 @@ success/error as expected. Such set of buffer parameters should include cases li
               { firstVertex: 0, vertexCount: 1 },
               { firstVertex: 0, vertexCount: 10000 },
               { firstVertex: 10000, vertexCount: 0 },
-              { firstVertex: 10000, vertexCount: 1 },
               { firstVertex: 10000, vertexCount: 10000 },
             ]
       )
@@ -437,7 +436,6 @@ success/error as expected. Such set of buffer parameters should include cases li
               { firstInstance: 0, instanceCount: 1 },
               { firstInstance: 0, instanceCount: 10000 },
               { firstInstance: 10000, instanceCount: 0 },
-              { firstInstance: 10000, instanceCount: 1 },
               { firstInstance: 10000, instanceCount: 10000 },
             ]
       )
--- a/src/webgpu/api/validation/encoding/cmds/render/draw.spec.ts
+++ b/src/webgpu/api/validation/encoding/cmds/render/draw.spec.ts
@@ -419,12 +419,28 @@ success/error as expected. Such set of buffer parameters should include cases li
       .beginSubcases()
       .combine('setBufferOffset', [0, 200]) // must be a multiple of 4
       .combine('attributeFormat', ['snorm8x2', 'float32', 'float16x4'] as GPUVertexFormat[])
-      .combine('vertexCount', [0, 1, 10000])
-      .combine('firstVertex', [0, 10000])
-      .filter(p => p.VStride0 === (p.firstVertex + p.vertexCount === 0))
-      .combine('instanceCount', [0, 1, 10000])
-      .combine('firstInstance', [0, 10000])
-      .filter(p => p.IStride0 === (p.firstInstance + p.instanceCount === 0))
+      .expandWithParams(p =>
+        p.VStride0
+          ? [{ firstVertex: 0, vertexCount: 0 }]
+          : [
+              { firstVertex: 0, vertexCount: 1 },
+              { firstVertex: 0, vertexCount: 10000 },
+              { firstVertex: 10000, vertexCount: 0 },
+              { firstVertex: 10000, vertexCount: 1 },
+              { firstVertex: 10000, vertexCount: 10000 },
+            ]
+      )
+      .expandWithParams(p =>
+        p.IStride0
+          ? [{ firstInstance: 0, instanceCount: 0 }]
+          : [
+              { firstInstance: 0, instanceCount: 1 },
+              { firstInstance: 0, instanceCount: 10000 },
+              { firstInstance: 10000, instanceCount: 0 },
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/encoding/cmds/render/draw.spec.ts
```diff
@@ -404,27 +404,46 @@ success/error as expected. Such set of buffer parameters should include cases li
...
       .combine('VStride0', [false, true] as const)
       // should the instance stride count be zero
       .combine('IStride0', [false, true] as const)
-      // the state of array stride
-      .combine('AStride', ['zero', 'exact', 'oversize'] as const)
       // the factor for offset of attributes in vertex layout
       .combine('offset', [0, 1, 2, 7]) // the offset of attribute will be factor * MIN(4, sizeof(vertexFormat))
-      .beginSubcases()
-      .combine('setBufferOffset', [0, 200]) // must be a multiple of 4
+      .combine('setBufferOffset', [200]) // must be a multiple of 4
```
lokokung, line 425: nit. If we are just using 1 value, maybe we can just make it a constant in the test?

src/webgpu/api/validation/encoding/cmds/render/draw.spec.ts
```diff
@@ -459,7 +478,7 @@ success/error as expected. Such set of buffer parameters should include cases li
     }
 
     const calcSetBufferSize = (
-      boundBufferSizeState: 'zero' | 'exile' | 'enough',
+      boundBufferSizeState: 'zero' | 'oneTooSmall' | 'exact',
```
lokokung, line 481: nit. maybe just `'tooSmall'` is good enough? Not sure whether the one part is relevant.

src/webgpu/api/validation/encoding/cmds/render/draw.spec.ts
```diff
@@ -592,17 +611,34 @@ buffer slot and index buffer will cause no validation error, with completely/par
...
       .combine('drawType', ['draw', 'drawIndexed', 'drawIndirect', 'drawIndexedIndirect'] as const)
       .beginSubcases()
-      .combine('vertexBoundOffestFactor', [0, 0.5, 1, 1.5, 2])
-      .combine('instanceBoundOffestFactor', [0, 0.5, 1, 1.5, 2])
-      .combine('indexBoundOffestFactor', [0, 0.5, 1, 1.5, 2])
       .combine('arrayStrideState', ['zero', 'exact', 'oversize'] as const)
+      .combineWithParams([
+        { vertexOffsetFactor: 0.0, instanceOffsetFactor: 0.0, indexOffsetFactor: 0.0 },
+        { vertexOffsetFactor: 1.0, instanceOffsetFactor: 1.0, indexOffsetFactor: 1.0 },
+        //
```
lokokung, line 618: nit. Since you already included comment sections to separate the cases, maybe a brief one-liner explaining what the cases are testing?

lokokung (approved):

src/webgpu/api/validation/encoding/cmds/render/draw.spec.ts
kainino0x: It's a bit silly but I left it here since the test is already parameterized around it. Having it here makes it clear we could add more values in the future if needed.

src/webgpu/api/validation/encoding/cmds/render/draw.spec.ts
kainino0x: I think it's kinda helpful to a test reader to know that we're testing the boundary condition. Also both 'zero' and 'oneTooSmall' are too small.

src/webgpu/api/validation/encoding/cmds/render/draw.spec.ts
kainino0x: Honestly I don't have much of an explanation here. I just picked some reasonable looking subset of the original combinatorial explosion that would maybe have the right amounts of overlaps. But I don't fully understand how they're used so I can't really write intelligently here.

jzm-intel: Initially these factors are designed to emulate all overlay possibility of three buffer binding within a single GPUBuffer. Consider each binding size is `1`, and the offset factor indicate the begin point of each binding form `0` to `2` in `0.5` steps.
So some of the cases are: 
1. No overlay: [0, 1, 2]
2. The first two binding no overlay, third binding overlay: [0, 1, (1, 1.5)]
3. The first two partially overlay, and the third one may or may not overlay with them: [0, 0.5, (0.5, 1, 1.5)]
4. The first two complete overlay, and the third one may or may not overlay with them: [0, 0, (0, 0.5, 1)]

And three different kind of binding (vertex, instance, index) map to these three binding in several ways.
Yes there are too many combination and many of them is duplicated...

And oops I realize that the expected binding size of three kind is not necessary the same, so combining the factor doesn't ensure all overlay cases are tested.
Maybe I should review this tests later and replace these factor things with explicit overlay expectation...

kainino0x: Thanks for the explanation, that's about what I thought but I wasn't sure.
Let me see if we can do something like:

* No overlay: [0, 2, 4]
* The first two binding no overlay, third binding overlay: [0, 2, (2, 2.5)]
* The first two partially overlay, and the third one may or may not overlay with them: [0, 0.5, (0.5, 1, 4)]
* The first two complete overlay, and the third one may or may not overlay with them: [0, 0, (0, 0.5, 4)]

kainino0x: Actually - this test is only a tiny fraction of the slowness here, so I'm just going to revert this, add a TODO for what you suggested, and land this PR.

PR #3123, opened 2023-11-01
Optimize some checkElements* implementations

kainino0x:
`webgpu:api,operation,command_buffer,copyTextureToTexture:*` accounts for a huge chunk of the api,operation test time (maybe 25%). This makes it almost 40% faster on my machine (13.5s -> 8.5s).

Most of the speedup is just from splitting the failure case code out into a different function. This must impact inlining or something in V8; unfortunately I suspect the improvement wouldn't translate directly to other JS engines.

Those tests could probably also use some parameterization reduction, but I'll do that separately.

Issue: None

<hr>

### DIFF
--- a/src/webgpu/util/check_contents.ts
+++ b/src/webgpu/util/check_contents.ts
@@ -135,14 +135,10 @@ export function checkElementsPassPredicate(
   predicate: CheckElementsPredicate,
   { predicatePrinter }: { predicatePrinter?: CheckElementsSupplementalTableRows }
 ): ErrorWithExtra | undefined {
-  const size = actual.length;
-  const ctor = actual.constructor as TypedArrayBufferViewConstructor;
-  const printAsFloat = ctor === Float16Array || ctor === Float32Array || ctor === Float64Array;
-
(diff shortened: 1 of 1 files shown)


src/webgpu/util/check_contents.ts
```diff
@@ -44,7 +44,28 @@ export function checkElementsEqual(
 ): ErrorWithExtra | undefined {
   assert(actual.constructor === expected.constructor, 'TypedArray type mismatch');
   assert(actual.length === expected.length, 'size mismatch');
-  return checkElementsEqualGenerated(actual, i => expected[i]);
```
lokokung, line 47: Hmmm... I would be interested to see if the change here where instead of creating an additional generator, we are now just comparing the arrays is the real culprit of the speed up? (Sort of the only thing that seems to really have changed functionally?)

lokokung (approved):

src/webgpu/util/check_contents.ts
kainino0x: I measured these changes separately - see the commit messages. The largest speedup was from just splitting the check function in half.

PR #3124, opened 2023-11-01
Prevent Double Runs and make run clearer.

greggman:
Prevents clicking a run button while tests are running. I seem to do this once in a while and get errors about DevicePool acquire device in use 😅

Also made the progress bar have a background color so it's slightly clearer that tests are running.


toji (approved):
LGTM! I've definitely hit this problem several times.

PR #3131, opened 2023-11-02
Remove preview workflow, do WPT lint in PR workflow

kainino0x:
The artifact upload for the PR preview workflow was really slow. We could re-enable this if we find it's needed though.

The WPT lint workflow was doing a redundant build of most of the CTS. We can just run WPT's lint after a regular build (it seems to take only 11+2 seconds on past runs, including the one on this PR).

Issue: None

<hr>

**Requirements for PR author:** N/A

**Requirements for [reviewer sign-off](https://github.com/gpuweb/cts/blob/main/docs/reviews.md):** N/A


mehmetoguzderin (approved):
SGTM, and the solution in main repo works better so we can orient that way in case to be revived. Thank you!

kainino0x:
Yes, I was thinking that too. Sounds good.

BTW, I forgot to mention in the PR - the preview also hasn't been working for at least a few days. Guessing that switching to the new solution would work better.

PR #3132, opened 2023-11-02
Test errors for out-of-range limit requests

kainino0x:
See https://crbug.com/1499030

Issue: None

<hr>


src/webgpu/util/math.ts
```diff
@@ -2056,6 +2056,7 @@ export function isPowerOfTwo(n: number): boolean {
   if (!Number.isInteger(n)) {
     return false;
   }
+  assert((n | 0) === n, 'isPowerOfTwo only supports 32-bit numbers');
```
kainino0x, line 2059: I verified that the tests that use this helper still pass.
Could switch to `Math.log2(n) % 1 === 0` but I'm not 100% if that's a sound implementation either so I just did this.
`webgpu:api,validation,capability_checks,limits,*`
`webgpu:api,operation,adapter,requestDevice:*`

src/webgpu/api/operation/adapter/requestDevice.spec.ts
```diff
@@ -311,6 +311,61 @@ g.test('limit,better_than_supported')
...
+    const requiredLimits = {
+      [limit]: value,
+    };
+
+    const errorName =
+      value < 0 || value > Number.MAX_SAFE_INTEGER
+        ? 'TypeError'
+        : limitInfo.class === 'maximum' && value > adapter.limits[limit]
+        ? 'OperationError'
+        : limitInfo.class === 'alignment' && (value > 2 ** 31 || !isPowerOfTwo(value))
```
kainino0x, line 362: The spec does not actually currently say that `2**32` or `2**53` is an error for alignment limits. I'm proposing it to the spec:
https://github.com/gpuweb/gpuweb/pull/4358

greggman (commented):
I am confused. The limits tests already test trying to go over the max limit and they all pass. 

https://github.com/gpuweb/cts/blob/41f89e77b67e6b66cb017be4e00235a0a9429ca7/src/webgpu/api/validation/capability_checks/limits/limit_utils.ts#L376

What is the bug this is trying to surface that wasn't already tested?

Can you just add a case there instead of making a new test?

kainino0x:
> What is the bug this is trying to surface that wasn't already tested?

We mishandle values &ge; 2^32 for 32-bit limit types, by `static_cast<uint32_t>(uint64_t)`:
https://chromium-review.googlesource.com/c/chromium/src/+/5002999

> Can you just add a case there instead of making a new test?

I may have missed the appropriate test to modify, but I only saw tests that used `mul`/`add` and I couldn't easily get specific corner cases in those tests.
I could certainly reduce the number of subcases in this test though.

greggman (commented):
ok, lgtm 🤷‍♂️

If you wanted to give it a try you could add a new limit here

https://github.com/gpuweb/cts/blob/41f89e77b67e6b66cb017be4e00235a0a9429ca7/src/webgpu/api/validation/capability_checks/limits/limit_utils.ts#L207

atLimit, overLimit, overStorageSize

but I this fine too I guess. It tests more values.

greggman (approved):

kainino0x:
Ack, I think I'll keep this since it can easily test a lot of values. It also does fit in with the tests next to it (even if we have some redundancy between these and the limit_utils based tests).

PR #3141, opened 2023-11-08
Fixes error scope tests for Linux/Intel

lokokung:
Fixes the error scope tests on Linux/Intel once the associated [fix](https://dawn-review.googlesource.com/c/dawn/+/159662) lands.

<hr>


src/webgpu/api/validation/error_scope.spec.ts
```diff
@@ -146,7 +155,7 @@ Tests that popping an empty error scope stack should reject.
   )
   .fn(t => {
     const promise = t.device.popErrorScope();
-    t.shouldReject('OperationError', promise);
```
kainino0x, line 149: this is fine since we already test popErrorScope has stacks elsewhere, but why is this needed?

src/webgpu/api/validation/error_scope.spec.ts
```diff
@@ -26,7 +26,16 @@ class ErrorScopeTests extends Fixture {
...
     const adapter = await gpu.requestAdapter();
     assert(adapter !== null);
-    const device = await adapter.requestDevice();
+
+    // We need to max out the adapter limits related to texture dimensions to more reliably cause an
+    // OOM error when asked for it, so set that on the device now.
+    const device = this.trackForCleanup(
+      await adapter.requestDevice({
+        requiredLimits: {
+          ['maxTextureDimension2D']: adapter.limits.maxTextureDimension2D,
```
kainino0x, line 35: nit
```suggestion
          maxTextureDimension2D: adapter.limits.maxTextureDimension2D,
```

kainino0x (approved):

src/webgpu/api/validation/error_scope.spec.ts
lokokung: I'm actually not entirely sure? Maybe I had an outdated version of chromium that didn't plumb the stack through, but when I was running the tests, without this they were all failing with the missing stack error... I can't easily rebase/update my chromium atm so I just added these suppressions for the time being. If we are confident it should work without them I can remove them also.

lokokung: Actually, since I'm OOO I will just land this for now. If it turns out we don't need the allow `allowMissingStacks` we can fix them in a follow up!

kainino0x: You didn't need to check back on this while OOO! I was assuming it was just an old chromium build, and I was going to try removing these changes, but they really don't matter, so no need to follow up.

PR #3149, opened 2023-11-09
Revert "Skip compressed image_copy tests in Compatibility mode."

greggman:
Reverts gpuweb/cts#3140

this is an invalid change isn't it? Suppressions shouldn't happen in the CTS


greggman:
committed via #3150

Issue #3147, opened 2023-11-09
Put Exceptions/Validation errors at the top in standalone


greggman:
If you run this test

https://gpuweb.github.io/cts/standalone/?compatibility=1&runnow=1&q=webgpu:api,operation,resource_init,texture_zero:uninitialized_texture_is_zero:dimension=%222d%22;readMethod=%22Sample%22;format=%22bgra8unorm%22

on a browser that supports compat it gets several screens worth of errors about results not matching but the source of the results not matching are at the very end. In other words, the most important info is hidden at the bottom. I was unaware of this and assumed the bad results shown in the first 30 screens of errors had a different cause then the actual cause bury below.

https://github.com/gpuweb/cts/assets/234804/68b1b1ef-4a96-4193-89e4-1048c047144d

Would it be okay for me to add any of the following?

1. Sort the errors, put the exceptions at the top. 

   It appears the errors say "EXPECTATION FAILED", "INFO", "EXCEPTION" so maybe it's possible to sort and put "EXCEPTION" at the top

   Maybe could add a "sort by" button but default it to whatever put's "EXCEPTION" at the top (vs say "chronological"

2. Change the colors or add some other indicator for tests that failed by "EXCEPTION" vs "EXPECTATION FAILED"

3. Anything else that might surface incorrect use of the API first?

...

@kainino0x

kainino0x:
The best sorting does seem to be a bit test- and failure-dependent, but some kind of fix here would be a good idea. It's a common issue because uncaught validation errors are the last thing that get checked but usually the most important.

Additional ideas:
2a: Use text to display what the worst severity of failure was (more than just the pass/skip/fail/warn status)
2b: Similar but list *all* of the severities of failures that happened in the test
4: Insert a log at the top when there were uncaught errors from the test
5: Elide the entire array data section, like we do for stacks with `(elided: below max severity)`. This can probably be implemented by moving it into the stack property instead of the message property.

IMO: definitely should do 5, probably 2, maybe 4.

PR #3156, opened 2023-11-10
Standalone: call out exceptions at the top of results

greggman:
Issue: #3147

This was the simplest thing I could think of.


kainino0x (commented):
This looks OK but I think I'd prefer to do something for all runtimes, propose #3163

kainino0x (approved):

PR #3162, opened 2023-11-13
Add case cache to repo

ben-clayton:
Add case cache to repo

* Add the cache files to the CTS repo under `src/resources/cache`.
* Compress & decompress the cache with gzip using the Compression Streams API. All the cache files sum to just over 4 meg.
* Update `gen_cache --validate` to check that all the cache files are present and up to date. Also error if there are old cache files that are no longer needed.
* Update `pr.yml` to use run `gen_cache --validate` before building with `all` (which re-generates the cache)

<hr>


ben-clayton:
Looks like I need to update the version of node that the presubmits use.

ben-clayton:
This is fun. The cache seems to differ based on the version of node used. Investigating.

Gruntfile.js
```diff
@@ -244,6 +252,10 @@ module.exports = function (grunt) {
     'parallel:write-out-wpt-cts-html-all',
     'run:autoformat-out-wpt',
   ]);
+  grunt.registerTask('build-node', 'Build out/ (no checks; run after generate-common)', [
```
kainino0x, line 255:
```suggestion
  grunt.registerTask('build-node', 'Build out-node/ (no checks; run after generate-common)', [
```

Gruntfile.js
```diff
@@ -30,9 +30,13 @@ module.exports = function (grunt) {
...
         args: ['tools/validate', ...kAllSuites.map(s => 'src/' + s)],
       },
+      'generate-cache': {
+        cmd: 'node',
+        args: ['tools/gen_cache', 'src/webgpu'],
+      },
       'validate-cache': {
         cmd: 'node',
-        args: ['tools/gen_cache', 'out', 'src/webgpu', '--validate'],
+        args: ['tools/gen_cache', 'src/webgpu', '--validate'],
```
kainino0x, line 39: If we can make the compression result stable, I would say, we don't need this step, we can instead just regenerate the cache and then check that git is clean at the end of the GitHub Actions check.

src/common/runtime/server.ts
```diff
@@ -71,13 +71,13 @@ if (!sys.existsSync('src/common/runtime/cmdline.ts')) {
   console.log('Must be run from repository root');
   usage(1);
 }
+setBaseResourcePath('out-node/resources');
```
kainino0x, line 74: Can this read the cache from `src` instead of `out-node`?

package.json
```diff
@@ -53,6 +54,7 @@
     "babel-plugin-add-header-comment": "^1.0.3",
     "babel-plugin-const-enum": "^1.2.0",
     "chokidar": "^3.5.3",
+    "compression": "^1.7.4",
```
kainino0x, line 57: It seems like this is unused as you've used CompressionStream?
However if this provides stable compression results then it makes sense to use it instead.

Gruntfile.js
```diff
@@ -30,9 +30,13 @@ module.exports = function (grunt) {
         cmd: 'node',
         args: ['tools/validate', ...kAllSuites.map(s => 'src/' + s)],
       },
+      'generate-cache': {
```
kainino0x, line 33:
```suggestion
      'generate-cache': {
        // Note this generates files into the src/ directory (not the gen/ directory).
```

kainino0x (commented):
> This is fun. The cache seems to differ based on the version of node used. Investigating.
> 
> Update: Looks like gzip compresses differently between node versions. I'll change the validation to test post-decompression tomorrow. Even still, there's one 4-byte diff between node 18 and 20 where a float number (looks non-finite, possibly `f16`) is encoded differently.

I think it would be much preferable if we can make the compression results stable. If they're not stable we'll end up with different developers' machines fighting over the compressed files.

Another idea - not sure if this would work - would be to store in `src/` in a non-compressed format, compress into `gen/`, and copy into `out*/`.
Or even store in a text format (to make git happier) and serialize+compress into `gen/`.
Never mind if this defeats the whole point of checking the cache into the repository. (Though come to think of it - I don't actually understand why we can't generate the cache at compile time)

Gruntfile.js
ben-clayton: Is there any existing code in place to check the git status is clean after building?
Note: This also checks for stale output, and gives a clear message on how to fix it.

src/common/runtime/server.ts
ben-clayton: It could, yes, but it seems a lot cleaner to have all the build under one directory for serving.

package.json
ben-clayton: I think this was needed for TS to know the types, but maybe there's a better package for that. I'll double check tomorrow.

ben-clayton:
> > This is fun. The cache seems to differ based on the version of node used. Investigating.
> > Update: Looks like gzip compresses differently between node versions. I'll change the validation to test post-decompression tomorrow. Even still, there's one 4-byte diff between node 18 and 20 where a float number (looks non-finite, possibly `f16`) is encoded differently.
> 
> I think it would be much preferable if we can make the compression results stable. If they're not stable we'll end up with different developers' machines fighting over the compressed files.

Understood. I'm not entirely sure it can be though. What if we pinned the generation to a node version, erroring if you try to generate with the wrong version of node? It would only affect the developers that re-generate the cases, which so far has been ~3 people. 

> Another idea - not sure if this would work - would be to store in `src/` in a non-compressed format, compress into `gen/`, and copy into `out*/`. Or even store in a text format (to make git happier) and serialize+compress into `gen/`. Never mind if this defeats the whole point of checking the cache into the repository. (Though come to think of it - I don't actually understand why we can't generate the cache at compile time)

We did generate the cache at compile time. That was the first version. It was deemed too expensive to do each time we run the CTS.

src/common/runtime/server.ts
kainino0x: Ah, if the server runtime is dependent on out-node already then SGTM.

package.json
kainino0x: `compression` appears to be a compression library unrelated to CompressionStream.
The CompressionStream types should be built into TypeScript (I'm seeing at least the `CompressionStream` symbol is defined).

Gruntfile.js
kainino0x: I thought we had something either here or in the spec repo but I'm not sure.
I think all we would need to do is add a step that runs `git diff-files --quiet` which will return an error code > 0 if there are changes to the working tree.
https://stackoverflow.com/questions/2657935/checking-for-a-dirty-index-or-untracked-files-with-git

Regardless this isn't that much code, and having a clear error message is worth it, so never mind.

kainino0x:
> We did generate the cache at compile time. That was the first version. It was deemed too expensive to do each time we run the CTS.

OK. I'd like to understand this a bit more. We already build all of Chromium in order to run the CTS in Dawn - does the cache take so long to build that we don't want to add it onto that build stage?

ben-clayton:
> > We did generate the cache at compile time. That was the first version. It was deemed too expensive to do each time we run the CTS.
> 
> OK. I'd like to understand this a bit more. We already build all of Chromium in order to run the CTS in Dawn - does the cache take so long to build that we don't want to add it onto that build stage?

It did. It previously took around 6 minutes, then spiked to 50, which caused major upset and forced me to drop everything to get the cache build removed from the CTS runs. Since we've all been optimizing the CTS, this has been brought down to less than 2 minutes on a moderately powerful single core machine.

Clearly we've brought this down to something _almost negligible_ in the grand scheme of things, however:
* This is only going to climb again as we add more tests.
* This was being run for each CTS bots configuration, and the data is purely static. It's a bunch of immutable constant expression constants that can be completely calculated once, ahead of time.
* CTS runs will repeat the generation N times to build this, so would each and every presubmit and CI run.
* (I could be wrong) but I believe the bots that were doing the cache calculation were the GPU bots, which are limited in their numbers. Infra were not happy that we were spending cycles doing non-gpu work with them.

Obviously it would be nicer if we could just build this when we build the CTS, but we've yet to find a place where this can be done without past infra impact and future scalability concerns.

Gruntfile.js
ben-clayton: Done

package.json
ben-clayton: You're right. Removed.

Gruntfile.js
ben-clayton: Done

ben-clayton:
I've added a followup change that adds a `cache/hashes.json` file, which holds the CRC32 hashes of all the source files used to build the cache binary file. This massively speeds up the validation, and means we only have to rebuild the cases that have been modified.
It also fixes the validation issue between node versions, as the output binary is no longer regenerated as part of validation.

ben-clayton:
PTAL

src/resources/cache/hashes.json
```diff
@@ -0,0 +1,106 @@
...
+  "webgpu/shader/execution/binary/f32_division.bin": "593b194",
+  "webgpu/shader/execution/binary/f32_matrix_addition.bin": "5937c46",
+  "webgpu/shader/execution/binary/f32_matrix_matrix_multiplication.bin": "5931426",
+  "webgpu/shader/execution/binary/f32_matrix_scalar_multiplication.bin": "593ff67",
+  "webgpu/shader/execution/binary/f32_matrix_subtraction.bin": "59328ea",
+  "webgpu/shader/execution/binary/f32_matrix_vector_multiplication.bin": "5939a29",
+  "webgpu/shader/execution/binary/f32_multiplication.bin": "593aa04",
+  "webgpu/shader/execution/binary/f32_remainder.bin": "593e328",
+  "webgpu/shader/execution/binary/f32_subtraction.bin": "593c524",
+  "webgpu/shader/execution/binary/i32_arithmetic.bin": "5914697",
```
kainino0x, line 32: These hashes are very poorly distributed. Is that a problem?

src/unittests/parse_imports.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+  t.test(`\n\n`, []);
+});
+
+g.test('simple').fn(t => {
+  t.test(`import 'x/y/z.js';`, ['a/b/x/y/z.js']);
+  t.test(`import * as blah from 'x/y/z.js';`, ['a/b/x/y/z.js']);
+  t.test(`import { blah } from 'x/y/z.js';`, ['a/b/x/y/z.js']);
+});
+
+g.test('multiline').fn(t => {
```
kainino0x, line 38: Probably should also test multiple imports in the same file

src/common/tools/gen_cache.ts
```diff
@@ -14,21 +16,23 @@ DataCache will load this instead of building the expensive data at CTS runtime.
 Options:
   --help          Print this message and exit.
   --list          Print the list of output files without writing them.
-  --nth i/n       Only process every file where (file_index % n == i)
-  --validate      Check that cache should build (Tests for collisions).
+  --force         Rebuild cache even if they're up to date
```
kainino0x, line 19: Note somewhere that different node versions can produce different outputs (so this may change files unnecessarily)

kainino0x (approved):
Sorry for the delay, LGTM

kainino0x:
> It did. It previously took around 6 minutes, then spiked to 50, which caused major upset and forced me to drop everything to get the cache build removed from the CTS runs. Since we've all been optimizing the CTS, this has been brought down to less than 2 minutes on a moderately powerful single core machine.

This is a major improvement that sounds like it came from one particular change - if I knew what that change was I could judge better how likely it is to be a problem again in the future.

> Clearly we've brought this down to something _almost negligible_ in the grand scheme of things, however:
> 
> * This is only going to climb again as we add more tests.

It's not clear to me how much this will climb. We could 3x the number of tests and if it takes 6min to build then I think we're still in an OK realm.

> * This was being run for each CTS bots configuration, and the data is purely static. It's a bunch of immutable constant expression constants that can be completely calculated once, ahead of time.

This doesn't concern me that much if it's a few minutes of compiler bot time. It should be just once per platform, not once per swarming config.

I worry slightly about the maintainability of the solution you've built, though I'm pretty sure it will be fine. But if it does turn out to cause difficulties then we can reevaluate.

> * CTS runs will repeat the generation N times to build this, so would each and every presubmit and CI run.

True on GitHub. On Dawn it's probably cached?

> * (I could be wrong) but I believe the bots that were doing the cache calculation were the GPU bots, which are limited in their numbers. Infra were not happy that we were spending cycles doing non-gpu work with them.

I hope not - theoretically the C++ compile should definitely not be happening on GPU bots, so would be OK if it's part of that build.

> Obviously it would be nicer if we could just build this when we build the CTS, but we've yet to find a place where this can be done without past infra impact and future scalability concerns.

kainino0x:
> I worry slightly about the maintainability of the solution you've built

I'll add I was pleasantly surprised by how little code it took to do this nice hashing solution, so if it turns out to be robust then that's great.

src/resources/cache/hashes.json
ben-clayton: It was my shoddy `hashCombine` which was shifting right then xor-ing. This meant the MSBs of the hash were always taken from the last import. Fixed by CRC'ing the string of the two hashes.

src/unittests/parse_imports.spec.ts
ben-clayton: Done

src/common/tools/gen_cache.ts
ben-clayton: Done

kainino0x (approved):

kainino0x:
I just remembered one more thought I had while reviewing (not blocking this, but I'll let you decide when to land it).

What if we checked in the uncompressed cache, and did only the compression in the build step? This would be fast and free us from variations in gzip results.

Depending on how these files change over time, Git might also do a better job storing incremental changes to the binary files if they're uncompressed. https://stackoverflow.com/a/20690323

PR #3167, opened 2023-11-14
Updates for writeTimestamp removal

kainino0x:
- `webgpu:api,validation,capability_checks,features,query_types:timestamp:*` *always* throws TypeError
  - `webgpu:api,validation,capability_checks,features,query_types:timestamp:featureContainsTimestampQuery=true` fails if writeTimestamp is enabled in Chrome `--enable-dawn-features=allow_unsafe_apis --enable-webgpu-developer-features`
- Other tests skip writeTimestamp **but only when it throws TypeError**. These all still pass if writeTimestamp is enabled in Chrome.
  - `webgpu:api,validation,encoding,encoder_open_state:non_pass_commands:command="writeTimestamp"`
  - `webgpu:api,validation,encoding,queries,general:writeTimestamp,*`
  - `webgpu:api,validation,queue,destroyed,query_set:writeTimestamp:`
  - `webgpu:api,validation,state,device_lost,destroy:command,writeTimestamp:*`

Issue: spec issue https://github.com/gpuweb/gpuweb/pull/4370

<hr>

**Requirements for PR author:**

- [x] All missing test coverage is tracked with "TODO" or `.unimplemented()`.
- [x] New helpers are `/** documented */` and new helper files are found in `helper_index.txt`.
- [x] Test behaves as expected in a WebGPU implementation. (If not passing, explain above.)

**Requirements for [reviewer sign-off](https://github.com/gpuweb/cts/blob/main/docs/reviews.md):**

- [ ] Tests are properly located in the test tree.
- [ ] [Test descriptions](https://github.com/gpuweb/cts/blob/main/docs/intro/plans.md) allow a reader to "read only the test plans and evaluate coverage completeness", and accurately reflect the test code.
- [ ] Tests provide complete coverage (including validation control cases). **Missing coverage MUST be covered by TODOs.**
- [ ] Helpers and types promote readability and maintainability.

When landing this PR, be sure to make any necessary issue status updates.

PR #3177, opened 2023-11-17
Ignore compilation error in writeTimestamp tests

haoxli:
As writeTimestamp API has been removed from SPEC and webgpu types, the tests including it will cause build error when rolling latest types.

Add pass's timestampWrites tests for destoryed query set.

Issue: #

<hr>


kainino0x:
I updated these recently in #3167 and intentionally kept the tests for writeTimestamp. We're definitely going to try to add it back in some form, and I think if we remove the tests then we will end up forgetting about them and having to rewrite them someday.

I think it would be OK to keep writeTimestamp in the types. But if we don't do that, we can use unsafe casts, like:

```ts
interface WithWriteTimestamp : GPUCommandEncoder {
    writeTimestamp(querySet: GPUQuerySet, queryIndex: GPUSize32): undefined;
}

(encoder.encoder as WithWriteTimestamp).writeTimestamp(/* ... */);
```

haoxli:
We can keep types to align with the spec. 
Could we simply add "@ts-ignore" before writeTimestamp API to ignore the compile error?

kainino0x:
Not sure `@ts-ignore` works with our lint setup, but yes it would be fine to turn off the type checking here. You can use `as any` + `eslint-disable-next-line @typescript-eslint/no-explicit-any`.

kainino0x (approved):
I replaced the two-line eslint+tsc ignore with just `as any` and a lint ignore. I prefer this because it is a more controlled escape hatch compared with `@ts-ignore`.

I also found a different way to include writeTimestamp in the list of encoder commands without `as any`.

LGTM, please merge if this looks good to you.

haoxli:
LGTM, thanks for the update.

PR #3178, opened 2023-11-17
Add documentation for the case cache

ben-clayton:
Add `npm run node` helper to sit alongside `npm run standalone` and `npm run wpt`.

Update build.md for `out-node`.


docs/build.md
```diff
@@ -1,16 +1,41 @@
...
+- A copy of any `.html` test cases from `src/`
+
+### 3. `out-node` directory
+
+**Built with**: `npm run node`
+
+**Used by**: NodeJS test runners:
+
+- [`src/common/runtime/cmdline.ts`](../src/common/runtime/cmdline.ts) - A command line interface test runner
+- [`src/common/runtime/server.ts`](../src/common/runtime/server.ts) - An HTTP server for executing CTS tests with a REST interface
```
kainino0x, line 36: As noted in #3160 nothing in this repository actually uses the out-node build. It's only used if you use it explicitly.
Though I just noticed there is no `tools/` version of the `server` runtime. We should have one - #3179.

```suggestion
**Used for**: Running NodeJS tools, if you want to specifically avoid the live-compilation overhead of the `tools/` versions, or are running on a deployment which no longer has access to `ts-node` (which is a build-time dependency). For example:

- `node out-node/common/runtime/cmdline.js` ([source](../src/common/runtime/cmdline.ts)) - A command line interface test runner
- `node out-node/common/runtime/server.js` ([source](../src/common/runtime/server.ts)) - An HTTP server for executing CTS tests with a REST interface
```

package.json
```diff
@@ -7,6 +7,7 @@
     "all": "grunt all",
     "standalone": "grunt standalone",
     "wpt": "grunt wpt",
+    "node": "grunt node",
```
kainino0x, line 10: This is confusing (already, before your change) because `npm run node` will compile `out-node` while `tools/run_node` will run tests under node.

The out-node build is quite niche and I think we should just leave it out of here. (If removed here, change to `npx grunt node` at [1].)

docs/build.md
```diff
@@ -1,16 +1,41 @@
...
+
+Contains:
+
+- An adapter for running WebGPU CTS tests under WPT
+- A copy of the needed files from `out/`
+- A copy of any `.html` test cases from `src/`
+
+### 3. `out-node` directory
+
+**Built with**: `npm run node`
```
kainino0x, line 31: [1]

docs/case_cache.md
```diff
@@ -0,0 +1,71 @@
...
+which can be run with `npx grunt run:generate-cache`.
+This tool is automatically run by the various Grunt build commands.
+
+As generating the cache is expensive (hence why we build it ahead of time!) the
+cache generation tool will only re-build the cache files it believes may be out
+of date. To determine which files it needs to rebuild, the tool calculates a
+hash of all the transitive source TypeScript files that are used to build the
+output, and compares this hash to the hash stored in
+[`src/resources/cache/hashes.json`](`../src/resources/cache/hashes.json`). Only
+those cache files with differing hashes are rebuilt.
```
kainino0x, line 25: We should have a note that different versions of Node will compress differently, so people know to expect if they regenerate a file it might change unnecessarily. It makes sense to mention here, because the hash thing tries to avoid unnecessary rebuilds, but it will sometimes rebuild when not necessary.

```suggestion
those cache files with differing hashes are rebuilt.

Since source changes will sometimes change the hash without changing the generated cache,
sometimes the cache will be regenerated unnecessarily. **This is OK, but try to avoid committing
no-op regenerations - this will happen if your version of Node produces different gzip outputs
than the original committer's Node did for the same input.**
```

kainino0x (approved):

package.json
ben-clayton: I sort of agree, but also feel like not having something that matches the other two confusing. With the same argument, I would expect `node standalone` to run a server, not just build it.

docs/build.md
ben-clayton: If we can get everything to work from just `out`, then that would be great. I've struggled to wrap my head around the N different flavors of build.

package.json
kainino0x: ack, no strong feeling.

ben-clayton:
Thank you for the thorough review and great suggestions, @kainino0x !

PR #3183, opened 2023-11-21
wgsl: Add `sparseMatrixRange` to FPTraits API

### LINKED ISSUE #3174: wgsl: Expose various test range generators in FPTraits
There are a variety of functions for generating test ranges that are effective parameterized by fp type (vectorF16Range vs vectorF32Range vs vectorF64). Because these are not part of the FPTraits there is a bunch of cases where case generation code is needing different functions instead of doing something like FP[trait].vectorRange.

dot for example was actually doing the wrong thing, and using F32 for both f32 and f16, so there are probably some bugs from this to be quashed.

This bug covers adding the APIs to FPTraits, and refactoring the existing code to use them.

zoddicus:
Fixes #3174

<hr>


zoddicus:
Another stacked change, all the new code is in https://github.com/gpuweb/cts/pull/3183/commits/744e31bd39b2d1f32468166431f3b3f92eb9ce44

src/webgpu/shader/execution/expression/call/builtin/transpose.spec.ts
```diff
@@ -10,71 +10,37 @@ import { makeTestGroup } from '../../../../../../common/framework/test_group.js'
...
-          sparseMatrixF64Range(cols, rows),
-          'finite',
-          FP.abstract.transposeInterval
-        );
-      },
-    }))
-  )
-  .reduce((a, b) => ({ ...a, ...b }), {});
-
 export const d = makeCaseCache('transpose', {
```
jzm-intel, line 42: Just
``` suggestion
export const d = makeCaseCache('transpose', cases);
```

jzm-intel (approved):
Looks good to me, thanks

src/webgpu/shader/execution/expression/call/builtin/atanh.spec.ts
```diff
@@ -12,46 +12,39 @@ Note: The result is not mathematically meaningful when abs(e) >= 1.
...
-  ...fullF32Range(),
-];
-const f16_inputs = [
-  ...biasedRange(kValue.f16.negative.less_than_one, -0.9, 20), // discontinuity at x = -1
-  -1,
-  ...biasedRange(kValue.f16.positive.less_than_one, 0.9, 20), // discontinuity at x = 1
-  1,
-  ...fullF16Range(),
-];
+const cases = (['f32', 'f16'] as const)
```
ben-clayton, line 25: For this and the other changes, a comment describing what `cases` holds (non-exhaustive fields) would be good as this made me go cross-eyed trying to figure out what it was actually doing

ben-clayton (approved):
LGTM, however I'm not sure if the cache should be re-generated or not based on the code changes. It might be worth checking whether you get the same diffs if we all agree on a NodeJS version to use. See https://github.com/gpuweb/cts/pull/3187

zoddicus:
> LGTM, however I'm not sure if the cache should be re-generated or not based on the code changes. It might be worth checking whether you get the same diffs if we all agree on a NodeJS version to use. See #3187

I am currently using 20.9.0, which I believe is the newest LTS release and the same one that you pin in that PR.

This PR shouldn't be changing the actual expectations significantly (There are a couple of tests that were using the wrong generator that will be changing, but most should be a no-op). It does change the code in floating_point.ts and the various .spec.ts files, so I think they should have a different hash.

ben-clayton:
SGTM then, thanks!

src/webgpu/shader/execution/expression/call/builtin/atanh.spec.ts
zoddicus: Do you think something like
```
// Cases: [f32|f16]_[non_]const
```

is sufficient?

This is what jzm has included for the usage of this pattern that they wrote, but I didn't replicate consistantly.

src/webgpu/shader/execution/expression/call/builtin/transpose.spec.ts
zoddicus: Done

Issue #3078, opened 2023-10-20
support read-only, read-write storage textures


dneto0:
https://github.com/gpuweb/gpuweb/pull/4326 has landed

It's a core feature, except that 'bgra8unorm' read-write and read-only support is guarded by GPUFeature [bgra8unorm-storage](https://gpuweb.github.io/gpuweb/#bgra8unorm-storage)

greggman:
related? https://github.com/gpuweb/cts/issues/2271

dneto0:
> related? #2271

Yes.  :-)  Also #1272

PR #3186, opened 2023-11-22
Add tests on createBindGroupLayout() with new storage texture accesses

Jiawei-Shao:
This patch adds tests on the use of new storage texture accesses `read-only` and `read-write` in the creation of bind group
layouts.

This patch also upgrades the version of @webgpu/types to the latest one to support these two storage texture accesses.

Note that right now this test can only pass when Chromium is launched with `--enable-dawn-features=allow_unsafe_apis`.

Issue: #3078

<hr>


src/webgpu/api/validation/createBindGroupLayout.spec.ts
```diff
@@ -163,8 +163,10 @@ g.test('visibility,VERTEX_shader_stage_storage_texture_access')
   .fn(t => {
     const { shaderStage, access } = t.params;
 
+    const appliedSuccess = access ?? 'write-only';
```
Kangz, line 166: nit: appliedAccess?

src/webgpu/api/validation/createBindGroupLayout.spec.ts
```diff
@@ -436,29 +438,35 @@ g.test('storage_texture,layout_dimension')
...
     `
-  Test that a validation error is generated if the format doesn't support the storage usage.
+  Test that a validation error is generated if the format doesn't support the storage usage. A
+  validation error is also generated if the format doesn't support the 'read-write' storage access
+  when the storage access is 'read-write'.
   `
   )
-  .params(u => u.combine('format', kAllTextureFormats))
+  .params(u =>
+    u.combine('format', kAllTextureFormats).combine('access', kStorageTextureAccessValues)
```
Kangz, line 447:
```suggestion
    u \\
    .combine('format', kAllTextureFormats) \\
       .combine('access', kStorageTextureAccessValues)
```

Kangz (commented):
@ben-clayton I'm surprised that Jiawei had to regenerate the cache here, is that expected?

ben-clayton:
> @ben-clayton I'm surprised that Jiawei had to regenerate the cache here, is that expected?

So the hashes will change if there's files modified that _could_ impact the generation of the cache files. The fact the files have actually changed is likely due to differences in NodeJS versions. [This is discussed here](https://github.com/gpuweb/cts/blob/main/docs/case_cache.md#file-cache):

> Since source changes will sometimes change the hash without changing the generated cache, sometimes the cache will be regenerated unnecessarily. **This is OK, but try to avoid committing no-op regenerations - this will happen if your version of Node produces different gzip outputs than the original committer's Node did for the same input.**

It would be good to try keeping the CTS developers on the same version of the NodeJS, and avoiding no-op regeneration like this.

ben-clayton:
Created https://github.com/gpuweb/cts/pull/3187 to try to prevent continual cache regeneration.

src/webgpu/api/validation/createBindGroupLayout.spec.ts
```diff
@@ -436,29 +438,37 @@ g.test('storage_texture,layout_dimension')
...
-            },
-          ],
-        });
-      },
-      !info.color?.storage
-    );
+    let success = info.color?.storage;
+    if (access === 'read-write' && !info.color?.readWriteStorage) {
+      success = false;
+    }
```
jzm-intel, line 461: How about
``` suggestion
    const success =
      info.color?.storage && !(access === 'read-write' && !info.color?.readWriteStorage);
```

jzm-intel (approved):
Looks good to me, thanks

Kangz (approved):
LGTM, I just don't understand while all these new files where created.

PR #3187, opened 2023-11-22
Pin the JavaScript runtime version for the data cache

ben-clayton:
Prevents accidental no-op regeneration of the CTS cache files.

See https://github.com/gpuweb/cts/blob/main/docs/case_cache.md#file-cache


ben-clayton:
@zoddicus , @Jiawei-Shao - fyi

kainino0x (commented):
IIUC this is going to force all CTS developers to figure out how to install this specific version of Node, and if we ever want to change it then everyone will have to do that.

If we can find a pinnable dependency that implements gzip without using the Node version, or switch to not compressing these files at all, I think it would be much preferable.

kainino0x:
_Maybe_ we could actually pin node by installing it as a dependency:
https://www.npmjs.com/package/node
But I'm not actually sure how well this works.

kainino0x:
Also do we know if pinned node versions have stable gzip output across different OSes/bitness?

ben-clayton:
> IIUC this is going to force all CTS developers to figure out how to install this specific version of Node, and if we ever want to change it then everyone will have to do that.

Yes, although I don't know how frequently that would happen. 

> If we can find a pinnable dependency that implements gzip without using the Node version

There might be. I'd be slightly concerned with the latency this might have on builds, if compression is done in JS.

> or switch to not compressing these files at all, I think it would be much preferable.

Unfortunately, compression is not the only diff between node versions. There's slightly different output for what I think is emitted for `f16` cases between 16 and later versions. 

Dropping compression may also negatively affect the "download" speed for some CTS runners. Our python web server appears to be surprisingly slow to serve large files. 

> Also do we know if pinned node versions have stable gzip output across different OSes/bitness?

Good question. Can certainly test between Mac Intel and Linux. Would likely need help to test beyond that.

kainino0x:
> > or switch to not compressing these files at all, I think it would be much preferable.
> 
> Unfortunately, compression is not the only diff between node versions. There's slightly different output for what I think is emitted for `f16` cases between 16 and later versions.

Sorry, I misspoke. What I was thinking of was compressing in the build step instead of in the checked-in files.

kainino0x:
OK, let's just do this for now. I don't want this as a permanent solution but hopefully it won't disrupt anyone too badly.

kainino0x:
Actually... on second thought, I think there's still a very good chance that this won't fix it because zlib has different output on different machines, and will just end up creating hassle for people. Not going to land this yet.

ben-clayton:
I'm surprised and depressed to see that Node does indeed produce different compressed output for Windows, macOS and Linux.

I've made another PR based on your suggestion to just remove the compression and let Git deal with it: https://github.com/gpuweb/cts/pull/3197

Closing this PR.

PR #3197, opened 2023-11-29
Remove compression on the case cache, split files to reduce transitive deps

ben-clayton:
Compression is not stable between Node versions, nor between operating systems.

This inflates the cache size 10x (from 4.3mb to 49mb), however these files should not be frequently changing, and Git should be performing compressions with packfiles and during network operations.

Logic for building the cache has been moved out of the `foo.spec.ts` files and moved into a new sibling `foo.cache.ts` files. The reason for this is to reduce the number of transitive imports, and so reduce the frequency that the cache needs to be rebuilt.

A couple of monolithic typescript files have been broken up to further reduce the number of transitive imports.

The cache generator now scans for a `MUST_NOT_BE_IMPORTED_BY_DATA_CACHE` token that can be used to figure out why the cache wants to regenerate for a given file. This has been added to `device_pool.ts`, as this really shouldn't be depended on for the cache.

Updated documentation for all these changes.


kainino0x:
The on-disk size (and as a proxy for the network size) in Git:
- Local `.git` before fetching this PR: 226M
- Local `.git` after fetching this PR: 231M

This is very good.

kainino0x:
> The cache generator now scans for a `MUST_NOT_BE_IMPORTED_BY_DATA_CACHE` token that can be used to figure out why the cache wants to regenerate for a given file. This has been added to `device_pool.ts`, as this really shouldn't be depended on for the cache.

FWIW if we later find that this check isn't strong enough, we could do the opposite and add ALLOW_IMPORT_BY_DATA_CACHE. I checked this would be feasible:
```
Final number of files hashed is 140, of which 36 are non-cache
```
Another way could be by include lint rules, but I don't think it would work well.

kainino0x:
> This inflates the cache size 10x (from 4.3mb to 49mb)

Are we still concerned about web server transfer time for these files (on localhost on test machines)?
If so we can still compress them - compress instead of copy in the `copy-assets` steps.

ben-clayton:
> FWIW if we later find that this check isn't strong enough, we could do the opposite and add ALLOW_IMPORT_BY_DATA_CACHE. I checked this would be feasible:

I considered this. Cherry picking the few unavoidable GPU / test harness files seemed like it would suffice. Happy to see how it goes and flip it though. 

> Are we still concerned about web server transfer time for these files (on localhost on test machines)?

These are less than 1/10 the file size we had before moving to binary files. We're also no longer incurring the cost of decompression, so it might balance out. I guess we'll see.

Issue #3201, opened 2023-12-04
wgsl: Change domain restriction pattern for numeric framework


zoddicus:
Currently in the numeric framework uses a pattern where the impl is wrapped in a domain limiting function, that forwards to the implementation as needed.

This causes complexity about how and when the `FPTrait`/`this` gets bound.

Changing this to be a optional domain generator function on the operation interface, like `extrema` to make it easier to reason about creating timelines.

Issue #3139, opened 2023-11-06
wgsl: Resolve how OOB intervals behave in the numeric framework


zoddicus:
While adding test cases for this comment, https://github.com/gpuweb/cts/pull/3087/files/de68a79d02318f3f451017917a94e310863176ef#r1381686423

I discovered that the testing framework doesn't do what I expect, eed.g. calculate all bounds as f64 precise values and then at the compare step determine if the interval is finite, if not just return true...

The current implementation of .contains only tests for being the infinite interval if the input is NaN. And does not have special logic for handling non-finite intervals in general.

I am pretty sure the rest of the framework is currently depending on .contains() to have something like if (!this.isFinite()) return true at the beginning of it.

I don't think this is actually the correct behaviour anyway, because for a lower then f64 precision calculation, if a calculation in the intermediate steps goes OOB, but then get pulled back inbounds in the intervening steps, then it should still behave as the unbounded interval, b/c an implementation would be totally valid to use lower precision storage throughout the calculation.

The simple solution would be if an interval is created with OOB values for it precision type, return the unbounded interval instead. This though becomes an issue for validation later, because there is a subtle differentiation between near OOB values and very OOB, so we need to preserve the calculated bounds in 'infinite' precision, and if the value ever went OOB for the 'finite' precision.

So the proper solution I think is maintain the existing interval calculations, but add a bit to the interval type that indicates if it ever went OOB during an intermediate calculation. Thi I am pretty sure would be done by adding a check for 'isFinite' in the spanIntervals function. .contains would be changed to just do the simple check if in the interval, the compareInterval() function would be responsible for checking this new OOB bit before calling contains.

zoddicus:
I was so young and naive when I thought that this would be simple :upside_down_face:

PR #3208, opened 2023-12-05
wgsl: Change *Unbounded[Interval|Vector|...] to *Infinite[Interval|Vector|...]

zoddicus:
The current 'unbounded' constants are for [-∞, ∞] intervals. Changing
the name to avoid confusion once the 'bounded' state bit is added to
the interval class.

Issue  #3201

<hr>


zoddicus:
This is dependent on the `bounds` -> `endpoints` change, so new code is in the second diff, bd95de97a2a25296922b3e0d6331fd6c9f727cc6

src/webgpu/util/floating_point.ts
```diff
@@ -4259,7 +4259,7 @@ export abstract class FPTraits {
    * [0, 0] and [1, 1] indicate that the correct answer in point they encapsulate.
    * [0, 1] should not be treated as a span, i.e. 0.1 is acceptable, but instead
    * indicate either 0.0 or 1.0 are acceptable answers.
-   * [-∞, +∞] is treated as unbounded interval, since an unbounded or
+   * [-∞, +∞] is treated as unbounded interval, since an out of bounds or
    * infinite value was passed in.
```
jzm-intel, line 4263: How about
``` suggestion
   * [-∞, +∞], i.e. infinite interval, will be returned when an out-of-bound or
   * infinite value was passed in.
```

jzm-intel (approved):
The code change generally looks good to me, thanks!

But I am not sure about the usage of name `infinite`. If I get it right, we use "unbounded interval" to indicate that we accept any result of the given FP type, or as the WGSL spec said "indeterminate value". But technically yes, it is a interval of [-inf, inf].

I hope others can have comments on this.

jzm-intel:
> The code change generally looks good to me, thanks!
> 
> But I am not sure about the usage of name `infinite`. If I get it right, we use "unbounded interval" to indicate that we accept any result of the given FP type, or as the WGSL spec said "indeterminate value". But technically yes, it is a interval of [-inf, inf].
> 
> I hope others can have comments on this.

@ben-clayton @dneto0

zoddicus:
This is the issue that I have filed, https://github.com/gpuweb/cts/issues/3139

In brief, for mix specifically, I have written a test case that goes OOB in one of the intermediate steps, but comes back in bounds by the end. This leads to a situation where (because of the intermediate OOB) the CTS should accept any input, but because we have an in bounds end interval, it is checking against that.

As implemented the framework is supposed to have converted that OOB value to the infinite interval at the intermediate step (which I probably could find the bug and fix) but that would eliminate the endpoint values. For validation there is distinction between going OOB near the edge of the range versus being way out there, so retaining what the actual calculated endpoints are is going to be needed for the validation testing.

zoddicus:
The bit should resolve this issue by allowing .contains to be rewritten so that if bit == true return true at the beginning.

ben-clayton:
> In brief, for mix specifically, I have written a test case that goes OOB in one of the intermediate steps, but comes back in bounds by the end. This leads to a situation where (because of the intermediate OOB) the CTS should accept any input, but because we have an in bounds end interval, it is checking against that.

This is roughly what I understood, but I'm vague about:
(a) How it goes OOB and then back into bounds.
(b) Why the modeling of this out-then-in-bounds doesn't accurately replicate the desired outcome.

Would it be possible to describe this in steps, with a concrete example, highlighting where the mismatch occurs?

zoddicus:
Took me a moment to come up with a solid example, because my new workstation apparently does the math in a higher precision, so wasn't causing the CTS to trip anymore :upside_down_face: 

For mix one of the allowed formulas is `e1 * (1 - e3) + e2 * e3`
If we use these inputs
```
e1 = 1                                                                                                                                           
e2 = 2                                                                                                                                           
e3 = (max/2) + 1
```

We have the following operations:
1)
```
e2 * e3 = 2 * (max/2) + 2                                                                                                                        
        = max + 2
```

2)
```
1 - e3 = 1 - max/2 + 1                                                                                                                           
       = -max / 2 
```

3)
```
e1 * (1 - e3) = 1 * ( -max/2)                                                                                                                    
              = -max/2
```

4)
```
e1 * (1 - e3) + e2 * e3 = -max/2 + (max + 2)                                                                                                     
                        = max/2 + 2
```

So the final result 4) is in bounds, but 1) went OOB, and if we are operating in f16, number in TS can accurately represent all of this.

The framework for f16 currently returns [32752.0, 32768.0], which is in bounds. The current implementation isn't behaving as intended, because at 1) this should have be splatted to [-∞, ∞], which should have propegated through. This in theory could be fixed, but will need to be revisited for validation testing.

Specifically https://www.w3.org/TR/WGSL/#floating-point-overflow states that if a value is OOB, but near the edge of the range, than it is fair for a platform to round into range, so doesn't nescarily cause a validation error. So how far things go OOB needs to be tracked, so this unbounded bit will probably need to be expanded to be a tri-state, always in bounds, always alteast near OOB, or went far OOB.

PR #3229, opened 2023-12-19
Test builtin(position) and interpolation

greggman:
Interpolation can affect position so these 2 seem like they need to be tested together.

Note: Running the tests on M1 Mac vs AMD Linux, sample and centroid tests get different values and it's not yet clear to me what the correct values are. Rather then continuing I'd like to at least get the working tests in as they are needed to surface known issues in WebGPU implementations.

<hr>


src/webgpu/shader/execution/shader_io/fragment_builtins.spec.ts
```diff
@@ -0,0 +1,777 @@
...
+    },
+  });
+
+  const texture0 = textures[0];
+  const uniformBuffer = t.device.createBuffer({
+    size: 8,
+    usage: GPUBufferUsage.UNIFORM | GPUBufferUsage.COPY_DST,
+  });
+  t.trackForCleanup(uniformBuffer);
+  t.device.queue.writeBuffer(uniformBuffer, 0, new Float32Array([texture0.width, texture0.height]));
```
lokokung, line 567: nit. Maybe could get rid of `texture0` and just use `width` and `height` instead of `texture0.width` aand `texture0.height`?

src/webgpu/shader/execution/shader_io/fragment_builtins.spec.ts
```diff
@@ -0,0 +1,777 @@
...
+    {
+      predicatePrinter: [
+        {
+          leftHeader: 'expected ==',
+          getValueForCell: i => expected[i],
+        },
+      ],
+    }
+  );
+}
```
lokokung, line 631: [1]

src/webgpu/shader/execution/shader_io/fragment_builtins.spec.ts
```diff
@@ -0,0 +1,777 @@
...
+      [4 / 16, 2 / 16],
+      [2 / 16, 12 / 16],
+
+      [0 / 16, 8 / 16],
+      [15 / 16, 4 / 16],
+      [14 / 16, 15 / 16],
+      [1 / 16, 0 / 16],
+    ],
+  ],
+]);
```
lokokung, line 389: I know that Shrek had a very similar table here https://github.com/gpuweb/cts/blob/main/src/webgpu/api/operation/render_pipeline/sample_mask.spec.ts#L143

The link that Shrek included in that instance is super helpful if you wanted to include it here also. (https://learn.microsoft.com/en-us/windows/win32/api/d3d11/ne-d3d11-d3d11_standard_multisample_quality_levels)

Also wondering if we should just move this to a more centralized place and provide a getter like: `getMultisampleTable(count: number)`. Would also be easier to read if the 16 was factored out and just divided in the end somehow.

src/webgpu/shader/execution/shader_io/fragment_builtins.spec.ts
```diff
@@ -0,0 +1,777 @@
...
+/**
+ * Renders float32 fragment shader inputs values to 4 rgba8unorm textures that
+ * can be multisampled textures. It stores each of the channels, r, g, b, a of
+ * the shader input to a separate texture, doing the math required to store the
+ * float32 value into an rgba8unorm texel.
+ *
+ * Note: We could try to store the output to an vec4f storage buffer.
+ * Unfortunately, using a storage buffer has the issue that we need to compute
+ * an index with the very thing we're trying to test. Similarly for a storage
+ * texture. Also, using a storage buffer seems to affect certain backends like
```
lokokung, line 439: nit. the "Similarly for storage texture" is a bit confusing? Maybe it was leftover comment?

lokokung (commented):
Thanks Gregg!! Sorry for the delay. I am going to CC Shrek this PR also and see if he could take a look as well since he also worked with multisample tests before.

src/webgpu/shader/execution/shader_io/fragment_builtins.spec.ts
```diff
@@ -0,0 +1,777 @@
...
+
+        @compute @workgroup_size(1) fn cs(@builtin(global_invocation_id) id: vec3u) {
+          let numSamples = ${numSamples};
+          let dimensions = textureDimensions(texture);
+          let sampleIndex = id.x % numSamples;
+          let tx = id.x / numSamples;
+          let offset = (id.y * dimensions.x + tx) * numSamples + sampleIndex;
+          let v = vec4u(textureLoad(texture, vec2u(tx, id.y), ${sampleIndex}) * 255.0);
+
+          buffer[offset] = bitcast<f32>(dot(v, vec4u(0x1000000, 0x10000, 0x100, 0x1)));
```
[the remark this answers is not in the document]
greggman: done

src/webgpu/shader/execution/shader_io/fragment_builtins.spec.ts
```diff
@@ -0,0 +1,777 @@
...
+  return pipeline;
+}
+
+/**
+ * Copies a texture (even if multisampled) to a buffer
+ * @param t a gpu test
+ * @param texture texture to copy
+ * @returns buffer with copy of texture, mip level 0, array layer 0.
+ */
+function copyTextureToBufferIncludingMultisampledTextures(t: GPUTest, texture: GPUTexture) {
```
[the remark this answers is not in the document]
greggman: That's a good idea.

For the copy, checking the code over at gpu_test `copySinglePixelTextureToBufferUsingComputePass` is (a) only copying a single pixel and (b) isn't recovering f32 values from an rgba8unorm texture. Do we want to try to combine all of those options into a single function or just keep them separate.

I agree that using texelviews to do the comparison would be really good to do. I think I'll mark it as a MAINTANENCE_TODO and try to change it in the next PR.

src/webgpu/shader/execution/shader_io/fragment_builtins.spec.ts
```diff
@@ -0,0 +1,777 @@
...
+function subtractVectors(v1: readonly number[], v2: readonly number[]) {
+  return v1.map((v, i) => v - v2[i]);
+}
+
+/**
+ * Computes the dot product of 2 vectors
+ */
+function dotProduct(v1: readonly number[], v2: readonly number[]) {
+  return v1.reduce((a, v, i) => a + v * v2[i], 0);
+}
```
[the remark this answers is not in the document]
greggman: moved

src/webgpu/shader/execution/shader_io/fragment_builtins.spec.ts
greggman: updated

shrekshao:
Sorry I'm ooo. I just defer to your decision

src/webgpu/shader/execution/shader_io/fragment_builtins.spec.ts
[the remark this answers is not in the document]
lokokung: I think that we should probably keep the value recovering part separate. (Might be able to do it via something internal to the `TexelViews` where we could specify how to interpret the binary data.)

But copying the data in/out would be nice to unify.

src/webgpu/shader/execution/shader_io/fragment_builtins.spec.ts
greggman: I got rid of the `/ 16` but I'm not convinced it's more readable. As it is, fractions seem clear to me (how far into the pixel the sample is) where-as integers feel like something I have to go read up on on what they are.  By a maybe poor analogy `1 / 64` (read as "one 64th") is much easier to understand than "0.015625" (what even is that?)

I guess I feel leaving them as x / 16 with a comment that these are fractional positions in a pixel might be more readable but for now I updated it.

src/webgpu/shader/execution/shader_io/fragment_builtins.spec.ts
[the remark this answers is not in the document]
greggman: So for now I changed it. Instead of getting the values from the textures one texture at a time and then combining them after, it reads all 4 textures at once into a single buffer. This results in a buffer that's ready to use as `rgba32float` which makes it easy to call the TexelView code. So the code now uses TexelView

We can decide whether to switch to reading them as rgba8unorm and keeping them as rgba8unorm in a separate PR, but, at a glance, the TexelView code assumes a single buffer and so, we'd either have to merge/interleave the 4 rgba8uorm buffers into one buffer or else refactor the TexelView code so it doesn't assume one buffer. As it is, it calls "theOneBuffer.subarray" to get data for each texel but that won't work here unles the buffers are interleaved into a single buffer.

There's also the issue that the TexelView code has no concept of samples. So, if it prints an error in texel 11, 5 and your sampleCount is 4 then that really means texel 2, 5, sample index 3,

It might be simple to pass in a sampleCount and have the TexelView code do `texelX = x / sampleCount | 0; sampleIndex = texelX % sampleCount` but that should probably happen in another PR

lokokung (approved):

src/webgpu/shader/execution/shader_io/fragment_builtins.spec.ts
```diff
@@ -525,46 +556,64 @@ async function renderFragmentShaderInputsTo4TexturesAndReadbackValues(
...
+    new Uint8Array(expected.buffer),
+    areaDesc
+  );
+
+  const failedPixelsMessage = findFailedPixels(
+    format,
+    { x: 0, y: 0, z: 0 },
+    { width, height, depthOrArrayLayers: 1 },
+    { actTexelView, expTexelView },
+    { maxFractionalDiff }
```
lokokung, line 605: Hmm, I think that there is a preference to use ULPs if possible. Given the fractional diff you are using I think that it's likely within 1 ULP.

lokokung (approved):

src/webgpu/shader/execution/shader_io/fragment_builtins.spec.ts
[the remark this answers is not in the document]
lokokung: Gotcha! This is great already! I think that copying the data out into a single buffer is fine, we probably just need to introduce the concept of samples to the TexelView. If we default it to 1, and then allow users to specify, we can just assume that the textures will be one after another in the buffer. (As long as we define and use the same copy functions we should be fine.) And yes, absolutely those follow ups should be in a separate PR! Happy holidays!!

Issue #1045, opened 2022-03-07
web_platform: reference tests imported to WPT need to be fuzzy


Juanmihd:
This is work related to https://bugs.chromium.org/p/chromium/issues/detail?id=1083478

Some of the reference tests that use canvas 2d as reference need some fuzziness to pass.

kainino0x:
Thanks for the fixes! Is this fully resolved now, or might there be more to do to get things fully passing?

Juanmihd:
I'm waiting to validate this with Chrome and the test infrastructure to see that everything passes and then I will close this.

kainino0x:
I think this is fixed now.

PR #3230, opened 2023-12-21
reftests: Allow large totalPixels for fuzzy matching

kainino0x:
For these tests the fuzzy matching is only to allow for slight differences like rounding errors. We don't care about how many pixels are different.
https://web-platform-tests.org/writing-tests/reftests.html#fuzzy-matching

Also adds fuzzy matching to two more canvas_colorspace tests because it's not important to have extremely precise results for any test that does color conversions or blending (at least right now).

I haven't tested this exact change but in the linked Chromium CL below we saw the following:
```
01:01:43.047 7604 Found pixels_different: 500, max_channel_diff: 1
Allowed pixels_different; 0-400, max_channel_diff: 0-2
```
```
01:01:44.062 7604 Found pixels_different: 700, max_channel_diff: 2
Allowed pixels_different; 0-400, max_channel_diff: 0-2
```
So this should definitely work.

Some other tests that we're not currently passing for other reasons render quite large areas, so we set totalPixels very large. For example the canvas_colorspace tests render twelve `(128 * devicePixelRatio)`-squared canvases.

Issue: #1045, https://chromium-review.googlesource.com/c/chromium/src/+/5134470

<hr>

### DIFF
--- a/src/webgpu/web_platform/reftests/canvas_colorspace_bgra8unorm.https.html
+++ b/src/webgpu/web_platform/reftests/canvas_colorspace_bgra8unorm.https.html
@@ -14,6 +14,7 @@
   <link rel="help" href="https://gpuweb.github.io/gpuweb/" />
   <meta name="assert" content="WebGPU bgra8norm canvas with colorSpace set should be rendered correctly" />
   <link rel="match" href="./ref/canvas_colorspace-ref.html" />
+  <meta name=fuzzy content="maxDifference=0-2;totalPixels=0-99999">
   <script type="module">
     import { runColorSpaceTest } from './canvas_colorspace.html.js';
     runColorSpaceTest('bgra8unorm');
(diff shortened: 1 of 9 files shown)


src/webgpu/web_platform/reftests/canvas_colorspace_bgra8unorm.https.html
```diff
@@ -14,6 +14,7 @@
   <link rel="help" href="https://gpuweb.github.io/gpuweb/" />
   <meta name="assert" content="WebGPU bgra8norm canvas with colorSpace set should be rendered correctly" />
   <link rel="match" href="./ref/canvas_colorspace-ref.html" />
+  <meta name=fuzzy content="maxDifference=0-2;totalPixels=0-99999">
```
kainino0x, line 17: Actually these tests render a very large area (twelve 128x128 canvases, possibly times the devicePixelRatio^2) so I'm going to increase these.

greggman (approved):
lgtm

I find it strange that totalPixels is any difference even if they are under maxDifference but 🤷‍♂️

kainino0x:
Yeah, it's unintuitive, but that's how it's defined. I think they wanted to define it such that you could use "fuzzy" matching to almost-precisely match a particular expected error. But I don't know if it actually is ever used like that.
https://web-platform-tests.org/writing-tests/reftests.html#fuzzy-matching

PR #3235, opened 2024-01-03
Compat: make atomicCompareExchangeWeak tests handle limits

greggman:
<hr>

### DIFF
--- a/src/webgpu/shader/execution/expression/call/builtin/atomics/atomicCompareExchangeWeak.spec.ts
+++ b/src/webgpu/shader/execution/expression/call/builtin/atomics/atomicCompareExchangeWeak.spec.ts
@@ -339,6 +339,11 @@ struct __atomic_compare_exchange_result<T> {
     const numInvocations = t.params.workgroupSize;
     const scalarType = t.params.scalarType;
 
+    t.skipIf(
+      numInvocations > t.device.limits.maxComputeWorkgroupSizeX,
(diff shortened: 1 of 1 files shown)


src/webgpu/shader/execution/expression/call/builtin/atomics/atomicCompareExchangeWeak.spec.ts
```diff
@@ -339,6 +339,11 @@ struct __atomic_compare_exchange_result<T> {
     const numInvocations = t.params.workgroupSize;
     const scalarType = t.params.scalarType;
 
+    t.skipIf(
```
shrekshao, line 342: nit: maybe filter out the params instead of skipping in test body?

shrekshao (approved):
LGTM

src/webgpu/shader/execution/expression/call/builtin/atomics/atomicCompareExchangeWeak.spec.ts
greggman: Thanks for the suggestion. I don't think I  can filter these out before without a device since I'm comparing to `device.limits.maxComputeWorkgroupSizeX`.  I could hard code those limits somewhere but it seems better to look them up in the one place that would be correct in the future.

shrekshao: Np for leaving as is then

PR #3236, opened 2024-01-03
Fix scaling issue in canvas image rendering test

greggman:
IIUC the issue is scaling can happen edge cases, especially given some APIs render with flipped Y so in the pixelated/crisp-edges case they might choose a different edge than the canvas API.

Making the display size exactly double the rendered size seems like it might solve this issue. The test would still show a difference between image-rendering: initial and image-rendeirng: pixelated / crisp-edges but should probably not show any difference otherwise

<hr>


kainino0x (approved):
(For any other readers more info can be found on https://crbug.com/1515185)

This looks like it'll at least make the test reliable for devices with devicePixelRatio=1. For higher non-integer devicePixelRatio values I suspect we may still see the same issues though maybe this will still be more reliable on average.

PR #3249, opened 2024-01-08
Remove grunt-ts

greggman:
IIUC, grunt-ts is only used to run typescript to do type checking which can be done more simply by just spawning typescript.

This removes 146 packages and most of the npm warnings


kainino0x (approved):
Nice find! I have noticed this dependency seemed silly, but did not realize it was so hefty. -_-

PR #3255, opened 2024-01-09
Allow Path as query

greggman:
Allows using paths as queries


kainino0x (approved):
LGTM. What happened with bitcast.bin though?

greggman:
> LGTM. What happened with bitcast.bin though?

¯\_(ツ)_/¯

Issue #3302, opened 2024-01-23
Make AbstractInts Go.


dj2:
Need framework support for AbstractInts in the shader CTS framework.

 * [x] Use AbstractInts in a shader
 * [ ] Retrieve an AbstractInt value back from a shader

PR #3306, opened 2024-01-23
Add support for AbstractInts as inputs

zoddicus:
This PR adds in abstract ints to the Scalar type system, and wires up being able to write them as inputs as literals in shaders. Since they are abstract values they will never be supported as buffer inputs. Returning the values via buffers will be handled in a seperate PR.

I have implemented the tests for AbstractInt -> i32 & u32 conversion to confirm the implementation operates as expected.

There is a substantial number of changes related to the fact that the existing code assumes all numerics are `number`s, but AbstractInts need to be represented as BigInts.

I have also cleaned up a couple of places related to the AbstractFloat implementation where there wasn't full testing coverage, etc.

Issue #3302

<hr>


src/webgpu/util/check_contents.ts
```diff
@@ -263,10 +263,10 @@ ${generatePrettyTable(opts, [
 // Helper helpers
 
 /** Convert an integral `number` into a hex string, padded to the specified `byteLength`. */
-function intToPaddedHex(number: number, { byteLength }: { byteLength: number }) {
-  assert(Number.isInteger(number), 'number must be integer');
-  let s = Math.abs(number).toString(16);
+function intToPaddedHex(n: number | bigint, { byteLength }: { byteLength: number }) {
+  assert(Number.isInteger(n), 'number must be integer');
+  let s = (n < 0n ? -n : n).toString(16);
```
dj2, line 268: nit: maybe rename `n` such that it doesn't also match the suffix for the BigInt value of `0n`.

src/webgpu/util/conversion.ts
```diff
@@ -762,37 +766,44 @@ function valueFromBytes(workingDataOut: TypedArrayBufferView, buf: Uint8Array, o
   return workingDataOut[0];
 }
 
+export const TypeAbstractInt = new ScalarType(
+  'abstract-int',
+  8,
+  (buf: Uint8Array, offset: number) =>
+    abstractInt(valueFromBytes(workingDataF64, buf, offset) as bigint)
```
dj2, line 773: Why `workingDataF64`? It just happens to be the right byte length?

zoddicus: That is a copy-pasta mistake.

zoddicus:
This I think is ready for re-review

src/webgpu/shader/execution/expression/call/builtin/atomics/atomicAdd.spec.ts
```diff
@@ -35,7 +35,7 @@ fn atomicAdd(atomic_ptr: ptr<AS, atomic<T>, read_write>, v: T) -> T
     u
       .combine('workgroupSize', workgroupSizes)
       .combine('dispatchSize', dispatchSizes)
-      .combine('scalarType', ['u32', 'i32'])
+      .combine('scalarType', ['u32', 'i32'] as ('u32' | 'i32')[])
```
ben-clayton, line 38:
```suggestion
      .combine('scalarType', ['u32', 'i32'] as const
```

src/webgpu/util/conversion.ts
```diff
@@ -991,12 +1034,30 @@ export interface ScalarBuilder {
...
+      );
+    }
+  } else {
+    if (typeof value === 'number') {
+      workingDataArray[0] = value;
+    } else {
+      unreachable(
+        `Mismatch between value type and working data array type, non-BigInt64Array requires 'number'`
+      );
+      workingDataArray[0] = value;
```
ben-clayton, line 1057: remove?

src/webgpu/util/conversion.ts
```diff
@@ -991,12 +1034,30 @@ export interface ScalarBuilder {
 function scalarFromValue(
   type: ScalarType,
   workingDataArray: TypedArrayBufferView,
-  value: number
+  value: number | bigint
```
ben-clayton, line 1037: Can't you use the `ArrayElementType<A>` trick above to statically enforce the correct value type?

src/webgpu/util/pretty_diff_tables.ts
```diff
@@ -1,5 +1,27 @@
...
 
+/**
+ * @returns a function that converts numerics to strings, depending on if they
+ * should be treated as integers or not.
+ */
+export function numericToStringBuilder(is_integer: boolean): (n: number | bigint) => string {
+  if (is_integer) {
+    return (val: number | bigint): string => {
+      if (typeof val === 'number') {
+        return val.toFixed();
```
ben-clayton, line 11: I'm curious: Why do we use `toFixed()` if we want to print an integer?

ben-clayton (commented):
Looking much nicer, thank you. A couple more suggestions to remove the casting noise.

src/webgpu/shader/execution/expression/call/builtin/atomics/atomicAdd.spec.ts
zoddicus: Done

src/webgpu/util/conversion.ts
zoddicus: Nice :+1:

src/webgpu/util/conversion.ts
zoddicus: Done

src/webgpu/util/pretty_diff_tables.ts
zoddicus: I didn't write this code originally, but quickly looking at .toFixed(), it has a arg `fractionDigits` which defaults to 0, which means don't print any digits after the decimal point (and the decimal point). So this would be doing the same thing as trunc'ing the value.

I am not sure why this is needed, since the value is supposed to be an integer, and an integer would be displayed the same way using .toString().
I assume there is some way that an non-int might get in here? :man_shrugging:

zoddicus:
This is ready for another review

src/webgpu/shader/execution/expression/expression.ts
```diff
@@ -165,16 +170,21 @@ function storageType(ty: Type): Type {
...
   }
   if (ty instanceof VectorType) {
+    assert(
+      ty.elementType.kind !== 'abstract-int',
+      `'abstract-int' values cannot appear in input storage`
+    );
     assert(
       ty.elementType.kind !== 'abstract-float',
-      `AbstractFloat values cannot appear in input storage`
+      `'abstract'float' values cannot appear in input storage`
```
ben-clayton, line 187:
```suggestion
      `'abstract-float' values cannot appear in input storage`
```

src/webgpu/util/pretty_diff_tables.ts
ben-clayton: Okidoki.

ben-clayton (approved):
Thank you for all the changes!

src/webgpu/shader/execution/expression/expression.ts
zoddicus: Done

PR #3311, opened 2024-01-24
Update getCurrentTexture test to match spec.

greggman:
It's not entirely clear from the current WebGPU spec if "modified" = setting to the same dimensions. It seems like that shouldn't matter and is inconsistent with other canvas contents.

Note: This fails in Chrome

<hr>

### DIFF
--- a/src/webgpu/web_platform/canvas/getCurrentTexture.spec.ts
+++ b/src/webgpu/web_platform/canvas/getCurrentTexture.spec.ts
@@ -40,6 +40,27 @@ class GPUContextTest extends GPUTest {
 
     return ctx;
   }
+
+  expectTextureDestroyed(texture: GPUTexture, expectDestroyed = true) {
+    this.expectValidationError(() => {
+      // Try using the texture in a render pass. Because it's a canvas texture
+      // it should have RENDER_ATTACHMENT usage.
+      assert((texture.usage & GPUTextureUsage.RENDER_ATTACHMENT) !== 0);
+      const encoder = this.device.createCommandEncoder();
+      const pass = encoder.beginRenderPass({
+        colorAttachments: [
+          {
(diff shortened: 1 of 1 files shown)


src/webgpu/web_platform/canvas/getCurrentTexture.spec.ts
```diff
@@ -263,21 +286,35 @@ g.test('resize')
...
     // Ensure that texture contents are transparent black.
     t.expectSingleColor(currentTexture, currentTexture.format, {
       size: [currentTexture.width, currentTexture.height, 1],
       exp: { R: 0, G: 0, B: 0, A: 0 },
     });
 
-    // Simply setting the canvas width and height values to their current values should not trigger
-    // a change in the texture.
-    ctx.canvas.width = 4;
-    ctx.canvas.height = 4;
```
toji, line 277: This feels like a valid test scenario that we don't want to lose? Either this counts as being modified and we should verify that it destroys the current texture in the process or it doesn't and we should ensure that the current texture is retained.

toji (changes requested):
LGTM, though I'd request that we restore a removed test.

src/webgpu/web_platform/canvas/getCurrentTexture.spec.ts
greggman: This was moved to like 311

```
    // Setting the canvas width and height values to their current values should
    // still trigger a change in the texture.
    const { width, height } = ctx.canvas;
    ctx.canvas.width = width;
    ctx.canvas.height = height;
```

toji: Apologies for missing that!

toji (approved):

Issue #3301, opened 2024-01-23
Running webgpu:web_platform,external_texture,video:* multiple times hangs browser


shaoboyan:
Try on Windows and MacOS several times through chrome canary by:
- access Running webgpu:web_platform,external_texture,video:* , and click several runs
- Canary hangs.

Suspect this is due to we hold so many js objects.  Change logic in `getVideoElement()` https://github.com/gpuweb/cts/blob/main/src/webgpu/web_platform/util.ts#L516 by creating video element once and return the same element always improve the behaviour (And if we remove all video frame related cases, this change fix the hang). But failed to handle video frame cases.

Not sure how to do a clean fix but maybe a start to change `getVideoElement()`

shaoboyan:
cc @kainino0x for ideas

kainino0x:
On my Mac (M1), what I'm seeing is that if I run the tests repeatedly then they get slower and slower. The CPU usage is very high when the tests are being slow.

If I give the page a few seconds to rest (in the background?), then it becomes fast again, so there is almost certainly a garbage collection problem here. (I thought maybe it was due to videos completing, but we have `loop: true`, so they shouldn't.)
The profiler isn't much help but if I click "pause" in the debugger then it always puts me on callbackAndResolve here:
https://github.com/gpuweb/cts/blob/c7e1a4f5c5455c9f1eb4f57b74a28ea73d959d84/src/webgpu/web_platform/util.ts#L407-L409
so what's slow is requestVideoFrameCallback.

We might be leaking the HTMLVideoElement objects which keep calling requestVideoFrameCallback until the GC realizes that that callback is dead?
I think we should try explicitly cleaning up the HTMLVideoElement objects, at least by clearing the `src` attribute. Though apparently more steps help on some browsers:
https://stackoverflow.com/a/28060352

I also saw 2 or so instances of `A VideoFrame was garbage collected without being closed. Applications should call close() on frames when done with them to prevent stalls.` I thought this should have been fixed already but we may be creating a lot more VideoFrames than we want because of the above issue. If we keep seeing it after cleaning up HTMLVideoElement objects then we should investigate it too.

kainino0x:
Fixing whatever is going on here will probably help a lot with test reliability in these tests. Nice find! Great that it's easy to reproduce.

mwyrzykowski:
[...]

PR #3312, opened 2024-01-24
Unload HTMLVideoElement actively after cases finished

shaoboyan:
This PR unload HTMLVideoElement by setting src to null string.

Issue:#3301

<hr>


src/webgpu/web_platform/util.ts
```diff
@@ -386,6 +386,7 @@ export function startPlayingAndWaitForVideo(
           try {
             await callback();
             resolve();
+            video.src = '';
```
beaufortfrancois, line 389: May I suggest adding `video.srcObject = null;` so that when we're using MediaStream as well, we're also handling it properly>

shaoboyan:
I could observe significant memory reduce on windows platform through task manager. But there still a slow memory increase (e.g. after run 60 times of cts sets, the memory grows from 12.0g to 12.6g).

src/webgpu/web_platform/util.ts
shaoboyan: Thanks @beaufortfrancois ! I've uploaded a new commit to address this. But for curious, we use MediaStream through `video.captureStream()`. Is this API set `srcObject` or clean the `srcObject` helps to release the MediaStream?

beaufortfrancois:
> I could observe significant memory reduce on windows platform through task manager. But there still a slow memory increase (e.g. after run 60 times of cts sets, the memory grows from 12.0g to 12.6g). Another trying is to call `mediaStream.removeTrack()` explicitly, but find no further improvements.

Did you look at Memory panel in DevTools?
Can you share steps to reproduce and I'd be happy to give it a try as well?

shaoboyan:
> > I could observe significant memory reduce on windows platform through task manager. But there still a slow memory increase (e.g. after run 60 times of cts sets, the memory grows from 12.0g to 12.6g). Another trying is to call `mediaStream.removeTrack()` explicitly, but find no further improvements.
> 
> Did you look at Memory panel in DevTools? Can you share steps to reproduce and I'd be happy to give it a try as well?

The most strange thing is that from DevTools/ performance trace, we cannot find any js heap explode clue. To reproduce, pls see issue https://github.com/gpuweb/cts/issues/3301

shaoboyan:
Without this PR, trace screenshot (run 3 times): 
trace-origin

Without this PR, the memory usage in task manager
Start:
main-stream-start
Run four times:
main-stream-run-four-times

Apply this PR (set video.src= ''), the trace screenshot (run 20 times):
Video-src-clear

shaoboyan:
> Looking at `chrome://media-internals/` when CTS is running, I can tell media players are quite slow to be marked as ended.

One guess here is this might be chrome implementation issue. Chrome implementation holds a cache for html_video_element here (https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/webgpu/gpu_external_texture.h;drc=fa4f751671b4641fe7a7c6adc1deea578b4ad4be;l=76). I don't think we have mechanism to remove cache actively when video finished playing.

beaufortfrancois (approved):

PR #3319, opened 2024-01-25
Re-review initMethod = render in image_copy tests

shrekshao:
This change simply removes a TODO planning comment.

In `webgpu:api,operation,command_buffer,image_copy`, the coveraged added by the planning for `initMethod="Draw";checkMethod="FullCopyT2B"` seems to be partly coverred by `webgpu:api,operation,memory_sync,texture,same_subresource:wr:boundary="command-buffer";write={"op":"attachment-store","in":"command-encoder"};read={"op":"t2b-copy","in":"command-encoder"}`

If we want to add this coverage in `image_copy`, it involves supporting writing arbitrary buffer data to the texture as a render attachment, via uint/int/float pipeline output in shader for all color renderable formats, and depth-stencil formats, which seems overkilled.

The missing coverage in `same_subresource` is: Use non-solid-color texture contents. Which seems easier to add as it is using a 4x4 rgba8unorm texture for all the subcases.

<hr>


kainino0x:
> The missing coverage in `same_subresource` is: Use non-solid-color texture contents. Which seems easier to add as it is using a 4x4 rgba8unorm texture for all the subcases.

Do you think this should be a TODO in same_subresource? Personally I think it's OK as is - doesn't seem too likely to catch new bugs.

kainino0x (approved):
LGTM. Inspecting this more closely since I didn't earlier, it does seem out of place to add any other initMethod. Both initMethod and checkMethod are part of the test - initMethod tests writeTexture and copyBufferToTexture, and checkMethod tests copyTextureToBuffer.

Maybe it would be good to have a little documentation about this, I guess in this file-level description? Not too important though.

shrekshao:
> Do you think this should be a TODO in same_subresource?

It is the TODO in `same_subresource`. Think we will keep leave it there as lower priority.

> Maybe it would be good to have a little documentation about this

I add a sentence: "Note: more coverage of memory synchronization for different read and write texture methods are in same_subresource.spec.ts."

PR #3320, opened 2024-01-25
Move the query manipulation to parseQuery

greggman:
I think this is ok. Checked dawn.node and standalone. I'm not sure why I didn't put it here first.


src/common/internal/query/parseQuery.ts
```diff
@@ -17,12 +17,49 @@ import {
...
+  return queryOrPath.endsWith('.spec.ts') &&
+    (queryOrPath.includes('/') || queryOrPath.includes('\\'))
+    ? convertPathToQuery(queryOrPath)
+    : queryOrPath;
+}
+
+/**
+ * Convert long suite names (the part before the first colon) to the
+ * shortest last word
+ *    foo.bar.moo:test/subtest/foo -> moo:test/subtest/foo
```
kainino0x, line 47: nit:
```suggestion
 *    foo.bar.moo:test,subtest,foo -> moo:test,subtest,foo
```

kainino0x (approved):

PR #3330, opened 2024-01-25
Compat: Fix readonly depth-stencil sampling

greggman:
It's not entirely clear to me what we should do here. For now, the failing tests are skipped in compat mode

The current compat spec says you can't have 2 different TEXTURE_BINDING views with different settings for the same texture. So you can't view a depth-stencil texture as both an `aspect: 'stencil-only'` in one view and `aspect: 'depth-only'` in another view in the same draw call. 

You can view them as one or the other in separate draw calls. Should we add a path that uses 2 draw calls? If so, can that be a separate PR?

<hr>

### DIFF
--- a/src/webgpu/api/operation/memory_sync/texture/readonly_depth_stencil.spec.ts
+++ b/src/webgpu/api/operation/memory_sync/texture/readonly_depth_stencil.spec.ts
@@ -31,7 +31,16 @@ testing while the other one is used for sampling.
       })
   )
   .beforeAllSubcases(t => {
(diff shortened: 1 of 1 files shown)


shrekshao (approved):
> You can view them as one or the other in separate draw calls. Should we add a path that uses 2 draw calls? If so, can that be a separate PR?

Sounds reasonable.

LGTM

Kangz:
Whoops I should have thought of compat, thanks for looking into that!

PR #3333, opened 2024-01-26
webgpu,api,operation,texture_view,write:format:*

shrekshao:
Mostly implements `webgpu,api,operation,texture_view,write:format:*`. Skipping `rgb10a2uint` and `sampleCount === 4` with maintanence TODO on helpers.

<hr>

**Requirements for PR author:**

- [x] All missing test coverage is tracked with "TODO" or `.unimplemented()`.
- [x] New helpers are `/** documented */` and new helper files are found in `helper_index.txt`.
- [x] Test behaves as expected in a WebGPU implementation. (If not passing, explain above.)

**Requirements for [reviewer sign-off](https://github.com/gpuweb/cts/blob/main/docs/reviews.md):**

- [x] Tests are properly located in the test tree.
- [x] [Test descriptions](https://github.com/gpuweb/cts/blob/main/docs/intro/plans.md) allow a reader to "read only the test plans and evaluate coverage completeness", and accurately reflect the test code.
- [x] Tests provide complete coverage (including validation control cases). **Missing coverage MUST be covered by TODOs.**
- [x] Helpers and types promote readability and maintainability.

When landing this PR, be sure to make any necessary issue status updates.

PR #3340, opened 2024-01-26
Extend compute pass multisampled copy texture helper to copy multiple pixels

shrekshao:
MAINTAINENCE_TODO: Extend the `copySinglePixelTextureToBufferUsingComputePass` to `copy2DTextureToBufferUsingComputePass` for copying multiple texels for multisampled texture.

This is to support sampleCount=4 for texture/write test in https://github.com/gpuweb/cts/pull/3333
<hr>


src/webgpu/gpu_test.ts
```diff
@@ -887,14 +913,24 @@ export class GPUTestBase extends Fixture<GPUTestSubcaseBatchState> {
...
     });
 
     const encoder = this.device.createCommandEncoder();
     const pass = encoder.beginComputePass();
     pass.setPipeline(computePipeline);
     pass.setBindGroup(0, uniformBindGroup);
-    pass.dispatchWorkgroups(1);
+    pass.dispatchWorkgroups(
+      (width + kWorkgroupSizeX - 1) / kWorkgroupSizeX,
+      (height + kWorkgroupSizeY - 1) / kWorkgroupSizeY,
```
kainino0x, line 931: Use Math.floor here to explicitly get an integer (instead of relying on WebIDL conversion from double to integer)

src/webgpu/gpu_test.ts
```diff
@@ -846,13 +854,26 @@ export class GPUTestBase extends Fixture<GPUTestSubcaseBatchState> {
       ${textureSrcCode}
       @group(0) @binding(1) var<storage, read_write> dst : Buffer;
 
-      @compute @workgroup_size(1) fn main() {
-        var coord = vec2<i32>(0, 0);
-        for (var sampleIndex = 0; sampleIndex < ${sampleCount};
+      struct Params {
+        origin: vec3u,
```
kainino0x, line 858: nit: If only supporting 2d textures, I think use vec2u instead of vec3u+pad. We can add vec3u later if we extend this to 3d. Right now it makes it a bit harder to read.

src/webgpu/gpu_test.ts
```diff
@@ -846,13 +854,26 @@ export class GPUTestBase extends Fixture<GPUTestSubcaseBatchState> {
...
+      };
+      @group(0) @binding(2) var<uniform> params : Params;
+
+      @compute @workgroup_size(${kWorkgroupSizeX}, ${kWorkgroupSizeY}, 1) fn main(@builtin(global_invocation_id) id : vec3u) {
+        let boundary = params.origin + params.extent;
+        let coord = params.origin + id;
+        if (any(coord >= boundary)) {
+          return;
+        }
+        let offset = (coord.x + coord.y * params.extent.x) * ${componentCount} * ${sampleCount};
```
kainino0x, line 871: If I'm reading this correctly then the origin of the texture (0,0) would map to the beginning of the buffer [0]. But I think copyTextureToBuffer would map the origin of the _copy_ to the beginning of the buffer which is more practical so it doesn't require a ton of unused padding at the start of the buffer.

kainino0x (approved):

src/webgpu/gpu_test.ts
shrekshao: You are right. Changed from using `coord.xy` to `id.xy`

src/webgpu/gpu_test.ts
```diff
@@ -869,11 +888,16 @@ export class GPUTestBase extends Fixture<GPUTestSubcaseBatchState> {
     });
 
     const storageBuffer = this.device.createBuffer({
-      size: sampleCount * type.size * componentCount,
+      size: sampleCount * type.size * componentCount * width * height,
```
shrekshao, line 891: Fixed the storage buffer size.

PR #3358, opened 2024-02-01
Remove usage of the Theora codec

padenot:
The WebGPU CTS references exactly one Theora file, and after years of service, this codec is being retired from the Web.

Some context, for the two implementations that have historically supported Theora, Chromium (except on Android) and Gecko (all platforms):
 
- https://chromestatus.com/feature/5158654475239424 (cc @dalecurtis)
- https://bugzilla.mozilla.org/show_bug.cgi?id=1860492 (I speak for Mozilla here, I'm handling this)

If need be, I can provide a media file that has exactly the same characteristics and switch it here with a different codec. However, my understanding is that what we care about here are primaries / transfer function / color space / color range and then pixel format / plane arrangement, and this seems to be covered by other codecs, so we should not lose any coverage here, as far as WebGPU is concerned.


dalecurtis:
lgtm, I agree the webm bt601 version covers this case.

kainino0x (approved):
LGTM! We copied these cases originally from WebGL. You'll want to remove the Theora WebGL tests too, see:
https://github.com/KhronosGroup/WebGL/tree/main/sdk/tests/resources
(and whatever tests reference them)

padenot:
> LGTM! We copied these cases originally from WebGL. You'll want to remove the Theora WebGL tests too, see:
> https://github.com/KhronosGroup/WebGL/tree/main/sdk/tests/resources
> (and whatever tests reference them)

Yes, this is planned as well, thanks!

PR #3367, opened 2024-02-06
Test maxBindGroupsPlusVertexBuffers limits

greggman:
Testing this is fairly confusing because there's so many things in play. I'm also not sure related tests don't need revisiting.

Testing `maxBindGroupsPlusVertexBuffers` implies to me, testing `maxBindGroups` and `maxVertexBuffers` at their max values. `maxBindGroups` in turn, implies setting `maxSampledTexturesPerShaderStage`, `maxSamplersPerShaderStage`, `maxStorageBuffersPerShaderStage`, `maxStorageTexturesPerShaderStage`, and `maxUniformBuffersPerShaderStage`. That's because, given the combination of parameters, there's no guarantee you can hit `maxBindGroups` if those don't sum to more than `maxBindGroups`.

And, similarly, testing `maxVertexBuffers` implies testing `maxVertexAttributes`, or so it feels. Like, in order to actually make a pipeline that will use `maxVertexBuffers`, the pipeline needs `maxVertexAttributes >= maxVertexBuffers`. 

Note that this doesn't currently test manually created layouts via `createPipelineLayout` + vertex buffers, it only test (1) `createPipeline` with `layout: 'auto'` and (2) `drawXXX`

I can add that but it would be nice to get some feedback on this first

<hr>

### DIFF
--- a/src/webgpu/api/validation/capability_checks/limits/limit_utils.ts
+++ b/src/webgpu/api/validation/capability_checks/limits/limit_utils.ts
@@ -1,7 +1,13 @@
 import { kUnitCaseParamsBuilder } from '../../../../../common/framework/params_builder.js';
 import { makeTestGroup } from '../../../../../common/framework/test_group.js';
 import { getGPU } from '../../../../../common/util/navigator_gpu.js';
-import { assert, range, reorder, ReorderOrder } from '../../../../../common/util/util.js';
+import {
+  assert,
+  range,
+  reorder,
+  ReorderOrder,
+  unreachable,
+} from '../../../../../common/util/util.js';
 import { getDefaultLimitsForAdapter } from '../../../../capability_info.js';
 import { GPUTestBase } from '../../../../gpu_test.js';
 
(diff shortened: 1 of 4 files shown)


greggman:
sorry, I forgot to add the file 😅

kainino0x:
> And, similarly, testing `maxVertexBuffers` implies testing `maxVertexAttributes`, or so it feels. Like, in order to actually make a pipeline that will use `maxVertexBuffers`, the pipeline needs `maxVertexAttributes >= maxVertexBuffers`.

Does GPUVertexBufferLayout.attributes actually have to be non-empty? I can't find a rule saying that, though there is one bit of spec text that assumes it:

> Let lastStride be max(attribute.offset + sizeof(attribute.format)) for each attribute in buffers[slot].attributes.

mwyrzykowski:
> Does GPUVertexBufferLayout.attributes actually have to be non-empty? I can't find a rule saying that, though there is one bit of spec text that assumes it:
> 
> > Let lastStride be max(attribute.offset + sizeof(attribute.format)) for each attribute in buffers[slot].attributes.

I don't think it needs to be non-empty and I think there are CTS tests which test that. My reading is that lastStride would be zero if all the attributes are empty.

kainino0x:
That makes logical sense. I think the text is ambiguous as written though.

kainino0x:
Sorry, couldn't find time to review this today.

PR #3374, opened 2024-02-07
Test default layout bindgroups

greggman:
It's probably a good idea to decide what we want to happen here and fix it ASAP because as it is, Chrome will allow bind groups made with different `layout: 'auto'` pipelines to work with each other but the spec seems to say that's disallowed. It's easy to imagine lots of WebGPU pages are already breaking this rule.

I also suspect the spec would like to say that all empty bind group layouts are compatible but I don't see that in the spec currently so it's tested that they are not compatible in this PR.

<hr>

### DIFF
--- a/src/webgpu/api/validation/encoding/programmable/pipeline_bind_group_compat.spec.ts
+++ b/src/webgpu/api/validation/encoding/programmable/pipeline_bind_group_compat.spec.ts
@@ -358,9 +358,9 @@ class F extends ValidationTest {
         const nonEmptyBindGroupLayout = pipeline.getBindGroupLayout(kNonEmptyBindGroupNdx);
         return { emptyBindGroupLayout, nonEmptyBindGroupLayout, pipeline };
       } else {
-        // Testing non-empty
-        //   emptyBindGroupLayout comes from the pipeline we'll render/compute with
-        //   nonEmptyBindGroupLayout comes from a possibly incompatible place.
+        // Testing non-empty:
+        // - emptyBindGroupLayout comes from the pipeline we'll render/compute with.
+        // - nonEmptyBindGroupLayout comes from a possibly incompatible place.
         const pipeline = pipelineType === 'auto0' ? pipelineAuto0 : pipelineExplicit;
         const nonEmptyBindGroupLayout =
           bindingType === 'explicit'
--- a/src/webgpu/api/validation/encoding/programmable/pipeline_bind_group_compat.spec.ts
+++ b/src/webgpu/api/validation/encoding/programmable/pipeline_bind_group_compat.spec.ts
@@ -37,6 +37,9 @@ type PipelineType = (typeof kPipelineTypes)[number];
 const kBindingTypes = ['auto0', 'auto1', 'explicit'] as const;
 type BindingType = (typeof kBindingTypes)[number];
 
+const kEmptyBindGroupNdx = 0;
+const kNonEmptyBindGroupNdx = 1;
+
 // Test resource type compatibility in pipeline and bind group
 // [1]: Need to add externalTexture
 const kResourceTypes: ValidBindableResource[] = [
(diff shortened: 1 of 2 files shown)


src/webgpu/api/validation/encoding/programmable/pipeline_bind_group_compat.spec.ts
```diff
@@ -277,6 +285,117 @@ class F extends ValidationTest {
...
+    );
+
+    const buffer = device.createBuffer({
+      size: 16,
+      usage: GPUBufferUsage.UNIFORM,
+    });
+    this.trackForCleanup(buffer);
+
+    const kEmptyBindGroupNdx = 0;
+    const kNonEmptyBindGroupNdx = 1;
```
kainino0x, line 339: nit: duplicated
```suggestion
```

src/webgpu/api/validation/encoding/programmable/pipeline_bind_group_compat.spec.ts
```diff
@@ -277,6 +285,117 @@ class F extends ValidationTest {
...
+      pipeline,
+    }: {
+      emptyBindGroupLayout: GPUBindGroupLayout;
+      nonEmptyBindGroupLayout: GPUBindGroupLayout;
+      pipeline: T;
+    } = (() => {
+      if (empty) {
+        // Testing empty so
+        //   nonEmptyBindGroupLayout comes from the pipeline we'll render/compute with
+        //   emptyBindGroupLayout comes from a possibly incompatible place.
```
kainino0x, line 353: nit: to be able to read where the sentence breaks are
```suggestion
        // Testing empty:
        // - nonEmptyBindGroupLayout comes from the pipeline we'll render/compute with.
        // - emptyBindGroupLayout comes from a possibly incompatible place.
```

src/webgpu/api/validation/encoding/programmable/pipeline_bind_group_compat.spec.ts
```diff
@@ -277,6 +285,117 @@ class F extends ValidationTest {
...
+            ? explicitEmptyBindGroupLayout
+            : bindingType === 'auto0'
+            ? pipelineAuto0.getBindGroupLayout(kEmptyBindGroupNdx)
+            : pipelineAuto1.getBindGroupLayout(kEmptyBindGroupNdx);
+        const nonEmptyBindGroupLayout = pipeline.getBindGroupLayout(kNonEmptyBindGroupNdx);
+        return { emptyBindGroupLayout, nonEmptyBindGroupLayout, pipeline };
+      } else {
+        // Testing non-empty
+        //   emptyBindGroupLayout comes from the pipeline we'll render/compute with
+        //   nonEmptyBindGroupLayout comes from a possibly incompatible place.
```
kainino0x, line 366:
```suggestion
        // Testing non-empty:
        // - emptyBindGroupLayout comes from the pipeline we'll render/compute with.
        // - nonEmptyBindGroupLayout comes from a possibly incompatible place.
```

src/webgpu/api/validation/encoding/programmable/pipeline_bind_group_compat.spec.ts
```diff
@@ -277,6 +285,117 @@ class F extends ValidationTest {
...
+
+    const {
+      emptyBindGroupLayout,
+      nonEmptyBindGroupLayout,
+      pipeline,
+    }: {
+      emptyBindGroupLayout: GPUBindGroupLayout;
+      nonEmptyBindGroupLayout: GPUBindGroupLayout;
+      pipeline: T;
+    } = (() => {
```
kainino0x, line 349: nit: just use `let` and assign inside the if/else?

src/webgpu/api/validation/encoding/programmable/pipeline_bind_group_compat.spec.ts
```diff
@@ -793,3 +912,165 @@ g.test('empty_bind_group_layouts_requires_empty_bind_groups,render_pass')
...
+    `
+  Test that bind groups created with default bind group layouts never match other layouts, including empty bind groups.
+
+  * Test that a pipeline with an explicit layout can not be used with a bindGroup from an auto layout
+  * Test that a pipeline with an auto layout can not be used with a bindGroup from an explicit layout
+  * Test that an auto layout from one pipeline can not be used with an auto layout from a different pipeline.
+
+  TODO:
+  * Test empty bindgroup layouts on the same default layout pipeline are not compatible. In other words if
+    you only define group(2) then group(0)'s empty layout and group(1)'s empty layout should be incompatible.
```
kainino0x, line 927: The spec rules as written only require it to come from the same pipeline, not the same slot:
https://gpuweb.github.io/gpuweb/#bind-group-compatibility
This is a little odd, but probably fine as I don't think it's much of a footgun?

src/webgpu/api/validation/encoding/programmable/pipeline_bind_group_compat.spec.ts
```diff
@@ -793,3 +912,165 @@ g.test('empty_bind_group_layouts_requires_empty_bind_groups,render_pass')
...
+  .params(u =>
+    u
+      .combine('pipelineAndBinding', [
+        { pipelineType: 'auto0', bindingType: 'auto0' }, // successful case
+        { pipelineType: 'explicit', bindingType: 'explicit' }, // successful case
+        { pipelineType: 'explicit', bindingType: 'auto0' },
+        { pipelineType: 'auto0', bindingType: 'explicit' },
+        { pipelineType: 'auto0', bindingType: 'auto1' },
+      ] as const)
+      .combine('empty', [false, true])
```
kainino0x, line 939: nit: factor this part out into a variable (using kUnitCaseParamsBuilder) to show it's the same across the render and compute test

src/webgpu/api/validation/encoding/programmable/pipeline_bind_group_compat.spec.ts
```diff
@@ -793,3 +912,165 @@ g.test('empty_bind_group_layouts_requires_empty_bind_groups,render_pass')
...
+  * Test that an auto layout from one pipeline can not be used with an auto layout from a different pipeline.
+
+  TODO:
+  * Test empty bindgroup layouts on the same default layout pipeline are not compatible. In other words if
+    you only define group(2) then group(0)'s empty layout and group(1)'s empty layout should be incompatible.
+  `
+  )
+  .params(u =>
+    u
+      .combine('pipelineAndBinding', [
```
kainino0x, line 932: nit: Flatten this by changing this to

```
      .combineWithParams([
```

kainino0x (approved):

src/webgpu/api/validation/encoding/programmable/pipeline_bind_group_compat.spec.ts
greggman: In that case we need to test that 2 bind groups that appear the same, are the same? In other words

```
      @group(0) @binding(0) var s1: sampler;
      @group(0) @binding(1) var t1: texture_2d<f32>;

      @group(1) @binding(0) var s2: sampler;
      @group(1) @binding(1) var t2: texture_2d<f32>;
```

If we use `layout: 'auto'` on this then a bindGroup made with `pipeline.getBindGroupLayout(0)` should work bound to bind group 1.

I don't think there is a test for that and it would be easy for browsers to disagree on this without a test?

It does make me wonder why they aren't just compatible in general. The rules are spelled out for what gets generated.

kainino0x: good point, we would definitely need to test that. Unless we want to make a breaking change in the spec and make it stricter.

The exclusivePipeline thing was added as a way to prevent footguns because auto layouts are sensitive to static uses in WGSL which are fairly subtle and easy to change (like by commenting out a function call, it removes from the BGL some binding that was used by that function's call tree). But I do wonder if it's really needed. Maybe with some good error messages there would be no footgun, like, "bindgroup created with auto BGL from pipeline 'X' but is used with pipeline 'Y' which has an incompatible auto BGL. Note that auto BGLs do not include declared bindings which are not reachable from the entrypoint of the pipeline".
Jasper and Brandon both had more opinions about this. We definitely picked the current behavior knowing it was a more "safe for 1.0" decision that we didn't necessarily have to keep forever.

Issue #1644, opened 2022-07-07
wgsl: validate assignment


dj2:
# Simple assignment
* [ ] `vec.{x,y,z,w,r,g,b,a} = 1`
* [ ] `vec.xy = vec2(1)` fails (no multi-swizzle on lhs)
* [ ] `vec[0] = 1` 
* [ ] `vec[-1] = 1` is not a validation error
* [ ] `vec[99] = 1` is not a validation error
* [ ] `ary[1] = 1` 
* [ ] `ary[-1] = 1` is not a validation error
* [ ] `ary[99] = 1` is not a validation error
* [ ] `vec[min(max(1, 2), 1)] = 1`
* [ ] `struct.a = 1`
* [ ] `struct.a.b.c.d.e.r = 1`
* [ ] `*ptr = 1`, write to a de-ref'd pointer
* [ ] `structA = structB` succeeds
* [ ] `structA = structB` validation error when structB has {runtime array, atomic, sampler?} (not constructable things)
* [ ] `func().r = 1` where func returns a vec ... ?
* [ ] Assign where RHS type != LHS type is validation error
* [ ] Write to LHS where access mode is `write` fails
* [ ] Write to LHS in `uniform` address space fails
* [ ] Write to LHS in `storage, read` address space fails
* [ ] Write to {texture, sampler, pointer} fails
* [ ] RHS of {texture, sampler, pointer} fails validation (maybe the same as the above?)

# Phony assignment
* [ ] RHS of {texture, sampler, pointer} succeed
* [ ] RHS of constructable passes
* [ ] RHS of non-constructable fails (struct with runtime array)
* [ ] RHS of function call

# Compound assignment
* [ ] Compound assignment with space between the `<char>` and `=` fails parse
* [ ] Same set of tests as simple assignment
* [ ] Validate that `e1` is only evaluated once (function call returning vec that increments global?)
* [ ] `_ += 1` fails (for all compound types)

- https://www.w3.org/TR/WGSL/#assignment

dj2:
I believe these are covered. We can fill individual issues for anything which is missing.

PR #3379, opened 2024-02-08
Fix assignment of AInt to u32 in the expression generator

amaiorano:
Apply the same fix that was made for AFloat to AInt when emitting the value assignment WGSL. Otherwise, if
globalTestConfig.unrollConstEvalLoops is true, as it is on Windows, then we emit invalid WGSL because we index an array of AInts with a runtime index, forcing concretization, which results in an assignment of i32 to u32. The fix is to do as was done with AFloat: we emit direct assignments of the literal values into the output array. Also made sure we don't emit the array of values in this case, as it's unused.

This fixes `'webgpu:shader,execution,expression,unary,ai_assignment:*'` tests that were failing on Windows.

Issue: #1644 

<hr>

### DIFF
--- a/src/webgpu/shader/execution/expression/expression.ts
+++ b/src/webgpu/shader/execution/expression/expression.ts
@@ -13,6 +13,7 @@ import {
   Value,
   Vector,
   VectorType,
+  isAbstractType,
   scalarTypeOf,
 } from '../../../util/conversion.js';
 
@@ -601,7 +602,7 @@ function basicExpressionShaderBody(
     // Constant eval
     //////////////////////////////////////////////////////////////////////////
     let body = '';
-    if (parameterTypes.some(ty => scalarTypeOf(ty).kind === 'abstract-float')) {
+    if (parameterTypes.some(isAbstractType)) {
       // Directly assign the expression to the output, to avoid an
       // intermediate store, which will concretize the value early
       body = cases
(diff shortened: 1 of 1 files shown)


src/webgpu/shader/execution/expression/expression.ts
```diff
@@ -48,6 +48,13 @@ export type Config = {
   vectorize?: number;
 };
 
+// Returns true if types contains an abstract float or abtract int
+function isAbstract(types: Type[]) {
```
ben-clayton, line 52: Already exists here:
https://github.com/gpuweb/cts/blob/3f04968bcc3bacf13aabf3c631a9459ffaddf262/src/webgpu/util/conversion.ts#L1613

src/webgpu/shader/execution/expression/expression.ts
```diff
@@ -601,7 +608,7 @@ function basicExpressionShaderBody(
     // Constant eval
     //////////////////////////////////////////////////////////////////////////
     let body = '';
-    if (parameterTypes.some(ty => scalarTypeOf(ty).kind === 'abstract-float')) {
+    if (isAbstract(parameterTypes)) {
```
ben-clayton, line 611:
```suggestion
    if (parameterTypes.some(isAbstractType)) {
```

src/webgpu/shader/execution/expression/expression.ts
```diff
@@ -627,10 +634,16 @@ function basicExpressionShaderBody(
   }`;
     }
 
+    // If params are abstract, we will assign them directly to the storage array, so skip the values array.
+    let valuesArray = '';
+    if (!isAbstract(parameterTypes)) {
```
ben-clayton, line 639:
```suggestion
    if (parameterTypes.some(isAbstractType)) {
```

ben-clayton (approved):
Thanks!

amaiorano:
> Thanks!

I'm still not used to GitHub. I had replied "Thanks!" to your suggetions, but those got lost when I resolved them...

PR #3391, opened 2024-02-12
Switch memory model tests to use PRNG

alan-baker:
* Switch to PRNG from Math.random

Issue: #

<hr>

### DIFF
--- a/src/webgpu/shader/execution/memory_model/memory_model_setup.ts
+++ b/src/webgpu/shader/execution/memory_model/memory_model_setup.ts
@@ -522,12 +522,12 @@ export class MemoryModelTester {
     );
   }
 
-  /** Returns a random integer between 0 and the max. */
+  /** Returns a random integer in the range [0, max). */
   protected getRandomInt(max: number): number {
     return this.prng.randomU32() % max;
   }
 
-  /** Returns a random number in between the min and max values. */
+  /** Returns a random number in the range [min, max). */
   protected getRandomInRange(min: number, max: number): number {
     if (min === max) {
       return min;
--- a/src/webgpu/shader/execution/memory_model/memory_model_setup.ts
+++ b/src/webgpu/shader/execution/memory_model/memory_model_setup.ts
@@ -1,6 +1,7 @@
 import { GPUTest } from '../../../gpu_test.js';
 import { checkElementsPassPredicate } from '../../../util/check_contents.js';
 import { align } from '../../../util/math.js';
+import { PRNG } from '../../../util/prng.js';
 
 /* All buffer sizes are counted in units of 4-byte words. */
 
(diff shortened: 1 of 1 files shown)


alan-baker:
@reeselevine PTAL, I wasn't 100% about whether `getRandomInt()` was meant to be inclusive or exclusive end point.

We've added a seedable PRNG to CTS since the memory model tests were originally added. This should make the tests more consistent across platforms, etc.

src/webgpu/shader/execution/memory_model/memory_model_setup.ts
```diff
@@ -521,7 +524,7 @@ export class MemoryModelTester {
 
   /** Returns a random integer between 0 and the max. */
   protected getRandomInt(max: number): number {
-    return Math.floor(Math.random() * max);
+    return this.prng.randomU32() % max;
```
zoddicus, line 527: To replicate the existing behaviour, you should use .random(), though I agree the comment is ambigious, since it is unclear if [0, max) or [0, max] (or even (0, max)) was intended.

Looking at the usages, i.e. shuffleArray, it looks like the code is depending on on the [0, max) behaviour or atleast comensating for it.

I would suggest keeping the existing behaviour and updating the comment to be clear on the return range, baring feedback from the original author.

zoddicus: Math.random() returns on [0, 1), which is what PRNG.random() does too

alan-baker: I've pinged Reese on this PR and don't mind waiting until I hear from him. I switched to `randomU32()` because that seemed more inline with the function's intent.

zoddicus: :+1: to waiting for a response

looking at the code in prng.ts, what you are doing will actually save an operation, since .random() does a scaling to get to [0, 1), so :+1:

alan-baker: Another thought: should each instantiation of the `MemoryModelTester` get a different seed?

reeselevine: Thanks Alan, that looks good to me. Sorry for the ambiguous comment, [0, max) is the correct interval.

As for having each instantiation getting a separate seed, I don't think it should matter. The random numbers are used to toggle parameters based on percentages and to shuffle threads, and there should be enough calls that a bad initial sequence won't cause the tests to be ineffective. When we've run our experiments we've used the same seed to make results more reproducible/line up runs on different devices, which could be useful here if failures are ever observed.

zoddicus (approved):

src/webgpu/shader/execution/memory_model/memory_model_setup.ts
alan-baker: Thanks Reese.

PR #3393, opened 2024-02-13
Separate tests for target format supporting usage and being a color format

kainino0x:
This splits two logically separate tests that were part of the same test before. This will make it easier to fix usage of the texture info table: Previously, it would access certain members of the info table assuming they were color, but would actually get the depth component of depth and depth-stencil formats. This wasn't really a problem because those formats were already expected to fail. But cleaner to instead just scope that test to only color formats.

Part 4 of Issue: #2495

<hr>

### DIFF
--- a/src/webgpu/api/validation/render_pipeline/fragment_state.spec.ts
+++ b/src/webgpu/api/validation/render_pipeline/fragment_state.spec.ts
@@ -3,7 +3,7 @@ This test dedicatedly tests validation of GPUFragmentState of createRenderPipeli
 `;
 
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
-import { range } from '../../../../common/util/util.js';
+import { assert, range } from '../../../../common/util/util.js';
 import {
   kBlendFactors,
   kBlendOperations,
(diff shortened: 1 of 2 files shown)


src/webgpu/api/validation/render_pipeline/fragment_state.spec.ts
```diff
@@ -49,9 +52,57 @@ g.test('color_target_exists')
     t.doCreateRenderPipelineTest(isAsync, false, badDescriptor);
   });
 
+g.test('targets_format_is_color_format')
+  .desc(
+    `Tests that color target state format must be a color format, regardless of how the
+    fragment shader writes to it.`
+  )
+  .params(u => {
+    assert(kColorTextureFormats.length + kDepthStencilFormats.length === kAllTextureFormats.length);
```
lokokung, line 61: nit. Does it make sense to put this here or as like an assert in the format table file?

lokokung (approved):

src/webgpu/api/validation/render_pipeline/fragment_state.spec.ts
kainino0x: I think it makes sense for it to be here because generally most tests don't care. This one just needs to make sure that it's testing every non-color format.

But actually I can just do this in a more constructive way: start with all formats, then filter out the color formats.

PR #3404, opened 2024-02-14
Test fragment_input @builtin(sample_mask)

greggman:
<hr>


src/webgpu/shader/execution/shader_io/fragment_builtins.spec.ts
```diff
@@ -1210,3 +1236,92 @@ g.test('inputs,front_facing')
...
+
+    Draws various triangles that should trigger different sample_mask values.
+    Checks that sample_mask matches what's expected.
+  `
+  )
+  .params(u =>
+    u //
+      .combine('nearFar', [[0, 1] as const, [0.25, 0.75] as const] as const)
+      .combine('sampleCount', [1, 4] as const)
+      .combine('interpolation', [
```
dneto0, line 1253: In this test the sample_mask value does not depend on the interpolation. So shouldn't we only set on kind of interpolation, to reduce test redundancy?

greggman: It does depend on interpolation because `sample` turns on per-sample rendering. I can remove all but 2

greggman: done

dneto0: ah, right. Thanks

src/webgpu/shader/execution/shader_io/fragment_builtins.spec.ts
```diff
@@ -1218,7 +1244,86 @@ g.test('inputs,front_facing')
...
+  .fn(async t => {
+    const {
+      x,
+      y,
+      nearFar,
+      sampleCount,
+      interpolation: { type, sampling },
+    } = t.params;
+    // prettier-ignore
+    const clipSpacePoints = [
```
dneto0, line 1283: If I understand this correctly, the triangles generated by this test always:
 - have an edge going through the origin
 - cover various portions of the upper right, upper left, and lower right quadrants
 - but never cover the bottom left quadrant.

If so, I'd suggest adding a second variant where the bottom left quadrant is also tested, basically by rotating the triangle around the origin by 180degrees.

dneto0 (changes requested):
Very nice test, even I understand what's going on, I think.

dneto0 (approved):
thanks!  that's great.

src/webgpu/shader/execution/shader_io/fragment_builtins.spec.ts
shrekshao: Would like to see more comments revealing the relation between the coord [x,y] and sample points they cover, like you did previously. Now it seems less intuitive.

Optional: So maybe a better way to specify the triangle vertices is (untested code):
```
.combine('corner', [[-1,-1], [1,-1], [-1,1], [1,1]])
.combine('x', [0.75, 1.25, 2])
.combine('y', [0.75, 1.25, 2])
//...
const xs = -Math.sign(corner[0]);
const ys = -Math.sign(corner[1]);
const clipSpacePoints = [
  [...corner, 0, 1],
  [corner[0] + xs * x, corner[1], 0, 1],
  [corner[0], corner[1] + ys * y, 0, 1],
];
```
No culling so I think the ccw, cw ordering won't matter?
Given a corner vertex, the x y length combination => n sample points coverred relations:
- 0.75, 0.75 => 0
- 0.75, 1.25 => 0
- 0.75, 2 => 1
- 1.25, 1.25 => 1
- 1.25, 2 => 2
- 2, 2 => 2

The comment can be something like:

```
Different x, y combinations draw different triangles covering different sets of sample points i.e. producing different sample_mask input

e.g. corner_v = [-1, 1], x=1.25, y=2, it draws the following triangle and is expected to emit sample_mask = 0b0101

    +-------------+
    |....s1|/     |
    |......|      |
    |...../|   s2 |
    +------C------+
    |s3./  |      |
    |../   |      |
    |./    |s4    |
    +-------------+
```

ref: https://learn.microsoft.com/en-us/windows/win32/api/d3d11/ne-d3d11-d3d11_standard_multisample_quality_levels

shrekshao (approved):
LGTM with optional comment

src/webgpu/shader/execution/shader_io/fragment_builtins.spec.ts
greggman: done

Issue #3405, opened 2024-02-14
add f16 to kScalarTypeInfo in src/webgpu/shader/types.ts


dneto0:
This ultimately adds to the list of values yielded by supportedScalarTypes in that same file, which affects:

webgpu:shader,execution,robust_access:*

webgpu:shader,execution,zero_init

PR #3407, opened 2024-02-14
add f16 to webgpu/shader/types.ts utilities

dneto0:
Udpate affected tests;
- webgpu:shader,execution,zero_init: skip f16 cases
- webgpu:shader,execution,robust_access: generalize to cover most f16 cases

Bug: #3405

Issue: #

<hr>

### DIFF
--- a/src/webgpu/shader/types.ts
+++ b/src/webgpu/shader/types.ts
@@ -247,7 +247,7 @@ export function* generateTypes({
     assert(scalarInfo.supportsAtomics, 'type does not support atomics');
     assert(
       containerType === 'scalar' || containerType === 'array',
-      'atomics are only supported for scalars and arrays'
+      "can only generate atomic inner types with containerType 'scalar' or 'array'"
     );
   }
   const scalarType = isAtomic ? `atomic<${baseType}>` : baseType;
--- a/src/webgpu/shader/types.ts
+++ b/src/webgpu/shader/types.ts
@@ -2,19 +2,20 @@ import { keysOf } from '../../common/util/data_tables.js';
 import { assert } from '../../common/util/util.js';
 import { align } from '../util/math.js';
 
-const kArrayLength = 3;
+const kDefaultArrayLength = 3;
 
 export type Requirement = 'never' | 'may' | 'must'; // never is the same as "must not"
 export type ContainerType = 'scalar' | 'vector' | 'matrix' | 'atomic' | 'array';
-export type ScalarType = 'i32' | 'u32' | 'f32' | 'bool';
+export type ScalarType = 'i32' | 'u32' | 'f16' | 'f32' | 'bool';
 
-export const HostSharableTypes = ['i32', 'u32', 'f32'] as const;
+export const HostSharableTypes = ['i32', 'u32', 'f16', 'f32'] as const;
 
 /** Info for each plain scalar type. */
 export const kScalarTypeInfo =
   /* prettier-ignore */ {
   'i32':    { layout: { alignment:  4, size:  4 }, supportsAtomics:  true, arrayLength: 1, innerLength: 0 },
   'u32':    { layout: { alignment:  4, size:  4 }, supportsAtomics:  true, arrayLength: 1, innerLength: 0 },
+  'f16':    { layout: { alignment:  2, size:  2 }, supportsAtomics: false, arrayLength: 1, innerLength: 0, feature: 'shader-f16' },
   'f32':    { layout: { alignment:  4, size:  4 }, supportsAtomics: false, arrayLength: 1, innerLength: 0 },
   'bool':   { layout:                   undefined, supportsAtomics: false, arrayLength: 1, innerLength: 0 },
 } as const;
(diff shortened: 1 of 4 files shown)


src/webgpu/shader/types.ts
```diff
@@ -212,33 +261,43 @@ export function* generateTypes({
 
   // Array types
   if (containerType === 'array') {
+    // Buffer affective binding size must be a multiple of 4. Adjust array length as needed.
```
kainino0x, line 264:
```suggestion
    // Buffer's effective binding size must be a multiple of 4. Adjust array length as needed.
```

src/webgpu/shader/types.ts
```diff
@@ -212,33 +261,43 @@ export function* generateTypes({
...
+      arrayLength = align(arrayLength, 4);
+    }
+
     const arrayTypeInfo = {
       elementBaseType: `${baseType}`,
-      arrayLength: kArrayLength,
+      arrayLength,
       layout: scalarInfo.layout
         ? {
             alignment: scalarInfo.layout.alignment,
```
kainino0x, line 279: aside: I'm not exactly sure what the type layouts are supposed to express, but should this be 16 in uniform storage class?

(I did not change this in the code above.)

src/webgpu/shader/types.ts
```diff
@@ -212,33 +261,43 @@ export function* generateTypes({
...
   // Array types
   if (containerType === 'array') {
+    // Buffer affective binding size must be a multiple of 4. Adjust array length as needed.
+    let arrayLength = kDefaultArrayLength;
+    if (
+      addressSpace === 'storage' &&
+      scalarInfo.layout !== undefined &&
+      scalarInfo.layout.alignment % 4 !== 0
+    ) {
+      arrayLength = align(arrayLength, 4);
```
kainino0x, line 271: I think this can make arrayLength bigger than it should be. For example for f16, scalarInfo.layout.alignment is 2.
If arrayLength is 6, the byte size will be 12. That's already a multiple of 4.
But this increases arrayLength to 8, making the byte size 16.
(I am not sure if this is a big problem, or just a nit.)

```suggestion
      scalarInfo.layout.alignment < 4
    ) {
      arrayLength = align(arrayLength, 4 / scalarInfo.layout.alignment);
```

The need for the `addressSpace === 'storage'` condition is also surprising until you dig into what happens later with arrayStride(). Since this setup has grown more complex I think it would be best to rework the whole block for readability:
```ts
  if (containerType === 'array') {
    let arrayLength = kArrayLength;
    let arrayLayout = undefined;
    if (scalarInfo.layout) {
      // Uniform storage class must have array elements aligned to 16.
      const elementAlignment = addressSpace === 'uniform' ? 16 : scalarInfo.layout.alignment;

      // Buffer's effective binding size must be a multiple of 4.
      if (elementAlignment < 4) {
        arrayLength = align(arrayLength, 4 / elementAlignment);
      }

      const arrayByteSize = kDefaultArrayLength * align(scalarInfo.layout.size, elementAlignment);
      arrayLayout = { alignment: scalarInfo.layout.alignment, size: arrayByteSize };
    }

    const arrayTypeInfo = { elementBaseType: `${baseType}`, arrayLength, layout: arrayLayout };
```

kainino0x (approved):
LGTM but 1-2 important small things

src/webgpu/shader/types.ts
dneto0: I totally reworked this.  Should have much better documentation about what the outputs should be.

dneto0 (commented):
Thanks for the feedback.
The array generation is mostly rewritten.

dneto0:
Retested with the things that depend on this: robust_access and zero_init execution tests.

kainino0x:
nit: avoid rebases that delete previously-reviewed commits, to make incremental diffs possible (use merge commits instead):
https://github.com/gpuweb/cts/blob/main/docs/intro/developing.md#pull-requests

kainino0x:
Today I discovered `git range-diff main 704bde62c08304c41818e43cb204376010ee928c 49b74692f63856b112043f2a842319edd9f98206` which can tell me that the first commit in the series didn't change during the rebase. So that's cool.

```
1:  704bde62c08 = 1:  2d3f878638a add f16 to webgpu/shader/types.ts utilities
-:  ----------- > 2:  d7b2b0cd9d4 Move TODO to test description
-:  ----------- > 3:  49b74692f63 Rewrite the array generation
```

src/webgpu/shader/types.ts
```diff
@@ -204,10 +216,34 @@ export function* generateTypes({
...
+      // - the base type if the generated type is an array
+      elementBaseType: string;
+      // Layout details if host-shareable, and undefined otherwise.
+      layout: undefined | AlignmentAndSize;
+      supportsAtomics: boolean;
+      // The number of elementBaseType items in the container.
+      arrayLength: number;
+      // 0 for scalar and vector.
+      // For a matrix type, this is the number of rows in the matrix.
+      innerLength?: number;
```
kainino0x, line 235: nit: write these as `/** doc comments */`. Editors should be able to pick those docs up on hover on usages of these members, I think.

```suggestion
    /** WGSL name for the generated type */
    type: string;
    _kTypeInfo: {
      /**
       * WGSL name for:
       * - the generated type if it is scalar or atomic
       * - the column vector type if the generated type is a matrix
       * - the base type if the generated type is an array
       */
      elementBaseType: string;
      /** Layout details if host-shareable, and undefined otherwise. */
      layout: undefined | AlignmentAndSize;
      supportsAtomics: boolean;
      /** The number of elementBaseType items in the container. */
      arrayLength: number;
      /**
       * 0 for scalar and vector.
       * For a matrix type, this is the number of rows in the matrix.
       */
      innerLength?: number;
```

src/webgpu/shader/types.ts
```diff
@@ -204,10 +216,34 @@ export function* generateTypes({
...
+    };
+  },
+  void
+> {
   const scalarInfo = kScalarTypeInfo[baseType];
   if (isAtomic) {
     assert(scalarInfo.supportsAtomics, 'type does not support atomics');
+    assert(
+      containerType === 'scalar' || containerType === 'array',
+      'atomics are only supported for scalars and arrays'
```
kainino0x, line 245: nit: I was confused about this for a minute because it sounded like `atomic<array<f32>>`
```suggestion
      "can only generate atomic inner types with containerType 'scalar' or 'array'"
```

src/webgpu/shader/types.ts
```diff
@@ -261,51 +299,64 @@ export function* generateTypes({
...
+      // Compute the layout of the array type.
+      // Adjust the array element count or element type as needed.
+      if (addressSpace === 'uniform') {
+        // Use a vec4 of the scalar type, to achieve a 16 byte alignment without internal padding.
+        // This works for 4-byte scalar types, and does not work for f16.
+        // It is the caller's responsibility to filter out the f16 case.
+        assert(!isAtomic, 'the uniform case is making vec4 of scalar, which cannot handle atomics');
+        arrayElemType = `vec4<${baseType}>`;
+        supportsAtomics = false;
+        const arrayElemLayout = vectorLayout('vec4', baseType) as AlignmentAndSize;
```
kainino0x, line 316: nit: `!` is safer than `as`
```suggestion
        const arrayElemLayout = vectorLayout('vec4', baseType)!;
```

src/webgpu/shader/types.ts
```diff
@@ -261,51 +299,64 @@ export function* generateTypes({
...
+    let arrayElementCount: number = kDefaultArrayLength;
+    let supportsAtomics = scalarInfo.supportsAtomics;
+    let layout: undefined | AlignmentAndSize = undefined;
+    if (scalarInfo.layout) {
+      // Compute the layout of the array type.
+      // Adjust the array element count or element type as needed.
+      if (addressSpace === 'uniform') {
+        // Use a vec4 of the scalar type, to achieve a 16 byte alignment without internal padding.
+        // This works for 4-byte scalar types, and does not work for f16.
+        // It is the caller's responsibility to filter out the f16 case.
```
kainino0x, line 312: Is it possible in WGSL to use f16 in an array in an uniform buffer with
```wgsl
array<struct { @align(16) x: f16 }>
```
? I assume we can't easily do that here because IIRC there are no inline structs, but would it be worth adding for the extra test coverage? (not in this PR)

src/webgpu/shader/types.ts
```diff
@@ -261,51 +299,64 @@ export function* generateTypes({
...
+        assert(!isAtomic, 'the uniform case is making vec4 of scalar, which cannot handle atomics');
+        arrayElemType = `vec4<${baseType}>`;
+        supportsAtomics = false;
+        const arrayElemLayout = vectorLayout('vec4', baseType) as AlignmentAndSize;
+        // assert(arrayElemLayout.alignment % 16 === 0); // Callers responsibility to avoid
```
kainino0x, line 317: Seems like this assert should be uncommented?

src/webgpu/shader/types.ts
```diff
@@ -261,51 +299,64 @@ export function* generateTypes({
...
-                  arrayLength *
-                  arrayStride({
-                    ...scalarInfo.layout,
-                    alignment: 16,
-                  })
-                : arrayLength * arrayStride(scalarInfo.layout),
-          }
-        : undefined,
+      arrayLength: arrayElementCount,
+      layout: scalarInfo.layout ? layout : undefined,
```
kainino0x, line 340: nit: I think this could equivalently just be:
```suggestion
      layout,
```

src/webgpu/shader/types.ts
```diff
@@ -261,51 +299,64 @@ export function* generateTypes({
...
     if (addressSpace === 'uniform') {
       yield {
-        type: `array<vec4<${scalarType}>,${arrayLength}>`,
+        type: `array<${arrayElemType},${arrayElementCount}>`,
         _kTypeInfo: arrayTypeInfo,
       };
     } else {
-      yield { type: `array<${scalarType},${arrayLength}>`, _kTypeInfo: arrayTypeInfo };
+      yield { type: `array<${arrayElemType},${arrayElementCount}>`, _kTypeInfo: arrayTypeInfo };
     }
```
kainino0x, line 352: These two cases are now identical.
```suggestion
    yield { type: `array<${arrayElemType},${arrayElementCount}>`, _kTypeInfo: arrayTypeInfo };
```

kainino0x (approved):
I didn't go back and try to re-understand the old logic. But pretty sure this looks good now.

src/webgpu/shader/types.ts
dneto0: Yes.  It would require the two users of this function (in robust-access and zero-init tests) to adjust because now there is padding in the data.

I wanted to leave that to a followup.

src/webgpu/shader/types.ts
dneto0: The callers generate everything, then filter out the bad case.   So this assert does actually trigger.
It requires additional work to avoid this.  I'd rather rework the f16 flow to a followup.

PR #3415, opened 2024-02-16
Generate Web Worker script for every test file

kainino0x:
This will be needed to launch Service Workers, which cannot be generated at runtime and cannot use dynamic import().

WIP. @beaufortfrancois 

Issue: #3345

<hr>


src/common/tools/gen_listings_and_webworkers.ts
```diff
@@ -0,0 +1,91 @@
...
+      fs.writeFileSync(
+        outFile,
+        `\
+// AUTO-GENERATED - DO NOT EDIT. See ${myself}.
+
+// oldG is a TestGroup<Fixture> object (defined in common/internal/test_group.ts).
+import { g as oldG } from '${relPathToSuiteRoot}/${entry.file.join('/')}.spec.js';
+
+// FIXME: Expose a proxied test interface. I think this can completely replace test_worker-worker.js
+// (using this instead of that), but if not then hopefully it can at least share code with it.
```
kainino0x, line 85: It may or may not replace test_worker-worker.js. It would be nice to avoid the duplicated functionality. But that one is a lot more flexible - it just creates one worker and runs tests on it, instead of having to manage one worker for every test file. But since we need to do that for Service Workers anyway, it's probably simplest to use the same setup and logic for all types of workers.

src/common/tools/gen_listings_and_webworkers.ts
```diff
@@ -0,0 +1,91 @@
...
+    const outFile = path.normalize(path.join(outDir, `${suite}/listing.js`));
+    fs.mkdirSync(path.join(outDir, suite), { recursive: true });
+    fs.writeFileSync(
+      outFile,
+      `\
+// AUTO-GENERATED - DO NOT EDIT. See ${myself}.
+
+export const listing = ${JSON.stringify(listing, undefined, 2)};
+`
+    );
```
kainino0x, line 59: All the code above this point is the same as the old gen_listings.ts. Git didn't detect it as a rename because I added too much code.

kainino0x:
This landed in #3419

PR #3419, opened 2024-02-19
Add service worker support

beaufortfrancois:
Spec PR: gpuweb/gpuweb#4465

<hr>


src/common/tools/gen_listings_and_webworkers.ts
```diff
@@ -0,0 +1,140 @@
...
+      // MAINTENANCE_TODO: Change this to whatever the option ends up being
+      ...(compatibility && { compatibilityMode: true }),
+    });
+  }
+
+  // const testcases = Array.from(await loader.loadCases(parseQuery(query)));
+  // assert(testcases.length === 1, 'worker query resulted in != 1 cases');
+
+  // const testcase = testcases[0];
+  const testcase = { query }; // FIXME! I failed to figure out how to get a testcase from oldG and query ;(
```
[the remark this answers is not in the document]
kainino0x: We've never done this before so it'll require work. I'm not quite sure.

I think you can cast oldG to `TestGroup<Fixture>` to get at its internals.
If your query is a string, use `parseQuery` to parse it to get the [`testParts`](url) and [`params`](https://github.com/gpuweb/cts/blob/62389bc770bb9d86670881dadc24a8ea8e259d3e/src/common/internal/query/query.ts#L105) from it.
Use the testParts to search for the test in the TestGroup.
Then you can probably just call the test's function with the params?
loadCases wraps this up normally, you can look at how it does it.

(Aside: I realized it doesn't need to be called `oldG`. There will be no new `g`. So you can just change the code I wrote to `import { g } ...` I think.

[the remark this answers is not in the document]
beaufortfrancois: I've tried this code without success.
I believe I'm missing instantiating the test but it's not clear where it's done.

```js
  try {
    const testQuerySingleCase = parseQuery(query);
    console.log({testQuerySingleCase});
    let testcase = null;
    for (const t of oldG.iterate()) {
      for (const c of t.iterate(testQuerySingleCase.params)) {
        if (comparePublicParamsPaths(c.id.params, testQuerySingleCase.params) === Ordering.Equal) {
          testcase = c;
        }
      }
    }
    console.log({testcase});
    testcase.fn();
    console.log('yeah');
  } catch(error) {
    console.log(error);
  }
```

image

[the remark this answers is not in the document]
beaufortfrancois: Argh... I thought I was cracking it with the following code but I reached a WebSocket error `WebSocket connection to 'ws://localhost:59497/optional_cts_websocket_logger' failed: Error in connection establishment: net::ERR_CONNECTION_REFUSED`

```js
  const testQuerySingleCase = parseQuery(query);
  let testcase = null;
  for (const t of oldG.iterate()) {
    for (const c of t.iterate(testQuerySingleCase.params)) {
      if (comparePublicParamsPaths(c.id.params, testQuerySingleCase.params) === Ordering.Equal) {
        testcase = c;
      }
    }
  }
  const [rec, result] = log.record(query);
  await testcase.run(rec, expectations);
  this.postMessage({ query, result });
```

image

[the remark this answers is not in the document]
beaufortfrancois: It turns out the WebSocket error is not the blocker issue.
The fail now tests (properly?) with the following surprising error as `expectations` is an Array of size 0.

```js
  const [rec, result] = log.record(query);
  console.log({expectations}); // <----
  await testcase.run(rec, expectations);
  this.postMessage({ query, result });
}
```

```
Error: expectations is not iterable
TypeError: expectations is not iterable
    at getExpectedStatus (http://localhost:8080/out/common/internal/test_group.js:561:25)
    at RunCaseSpecific.run (http://localhost:8080/out/common/internal/test_group.js:717:13)
    at async WindowClient.reportTestResults (http://localhost:8080/out/webgpu/webworker/api/operation/adapter/requestAdapter.worker.js:47:3)
```

image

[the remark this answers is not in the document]
beaufortfrancois: Using the following code allows me to run a small subset of tests (http://localhost:8080/standalone/?worker=service&q=webgpu:api,operation,adapter,requestAdapter:requestAdapter:*), which is a small victory (yeah!):

```js
  const testQuerySingleCase = parseQuery(query);
  let testcase = null;
  for (const t of g.iterate()) {
    for (const c of t.iterate(testQuerySingleCase.params)) {
      if (comparePublicParamsPaths(c.id.params, testQuerySingleCase.params) === Ordering.Equal) {
        testcase = c;
      }
    }
  }
  const [rec, result] = log.record(query);
  await testcase.run(rec, testQuerySingleCase, expectations);
  this.postMessage({ query, result });
  ```

The problem now comes from Service Worker I believe.
When running many tests at once (http://localhost:8080/standalone/?worker=service&q=webgpu:api,operation,*), the service worker file I get is not the one I expect. More specifically, the following example shows I got requestAdapter.worker.js instead of requestAdapterInfo.worker.js.
It's not clear to me what is happening but the creation of many service worker files may not help ;(

image


```js

    const registration = await navigator.serviceWorker.register(serviceWorkerPath, {
      type: 'module',
      scope: '/',
    });
    await navigator.serviceWorker.ready;
    if (registration.active.scriptURL !== `${window.location.origin}${serviceWorkerPath}`) {
      console.error(`Failed to retrieve proper file from service worker registration.
          Got ${registration.active.scriptURL} instead of ${window.location.origin}${serviceWorkerPath}`);
      return;
    }
 ```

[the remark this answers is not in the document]
Comment (beaufortfrancois) [reply src/common/tools/gen_listings_and_webworkers.ts]: 👋  I finally got it working!

**Comment by kainino0x** [comment]
Nice!! Sorry, I ended up with a lot of code reviewing to do today. I will review soon.

File: src/common/runtime/helper/test_worker.ts [hunk 2]
```diff
@@ -94,3 +94,53 @@ export class TestSharedWorker {
...
+      const result: TransferredTestCaseResult = ev.data.result;
+      if (result.logs) {
+        for (const l of result.logs) {
+          Object.setPrototypeOf(l, LogMessageWithStack.prototype);
+        }
+      }
+      this.resolvers.get(query)!(result as LiveTestCaseResult);
+
+      // MAINTENANCE_TODO(kainino0x): update the Logger with this result (or don't have a logger and
+      // update the entire results JSON somehow at some point).
```
kainino0x, line 132: I think we have three copies of this code, could we refactor it into a function (or a method on a base class)?

src/common/tools/dev_server.ts
```diff
@@ -150,14 +150,17 @@ app.get('/out/**/*.js', async (req, res, next) => {
   const tsUrl = jsUrl.replace(/\.js$/, '.ts');
   if (compileCache.has(tsUrl)) {
     res.setHeader('Content-Type', 'application/javascript');
+    res.setHeader('Service-Worker-Allowed', '/');
```
kainino0x, line 153: Is this needed? Seems like it's only needed for intercepting network requests which we don't need to do.

src/common/tools/dev_server.ts
```diff
@@ -150,14 +150,17 @@ app.get('/out/**/*.js', async (req, res, next) => {
...
     res.send(compileCache.get(tsUrl));
     return;
   }
 
-  let absPath = path.join(srcDir, tsUrl);
+  // I'm not sure if this is the way I should handle it...
```
kainino0x, line 158: This would require `out` to be built in order to use the dev server. Dev server is supposed to work without any build step. So instead when we get a `.worker.js` request we want to generate the response on the fly similar to what is done for `listing.js` (except we don't need to cache it because the generation should be very fast).

src/common/tools/dev_server.ts
```diff
@@ -166,6 +169,7 @@ app.get('/out/**/*.js', async (req, res, next) => {
       compileCache.set(tsUrl, result.code);
 
       res.setHeader('Content-Type', 'application/javascript');
+      res.setHeader('Service-Worker-Allowed', '/');
```
kainino0x, line 172: same

src/common/tools/gen_listings_and_webworkers.ts
```diff
@@ -78,12 +78,62 @@ export const listing = ${JSON.stringify(listing, undefined, 2)};
...
+self.onconnect = (event) => {
+  const port = event.ports[0];
 
-// FIXME: Expose a proxied test interface. I think this can completely replace test_worker-worker.js
-// (using this instead of that), but if not then hopefully it can at least share code with it.
-console.log(oldG.iterate());
+  port.onmessage = (ev) => {
+    void reportTestResults.call(port, ev);
+  };
+};
```
kainino0x, line 136: If possible, move all of this into a new file (in src/common/runtime/helper probably) so this string can be as minimal as possible, something like this:

```ts
import { g } from '${relPathToSuiteRoot}/${entry.file.join('/')}.spec.js';
import { wrapTestGroupForWorker } from '${relPathToSuiteRoot}/../common/runtime/helper/wrap_for_worker.js';

wrapTestGroupForWorker(g);
```

src/common/tools/gen_listings_and_webworkers.ts
```diff
@@ -78,12 +78,62 @@ export const listing = ${JSON.stringify(listing, undefined, 2)};
...
+    setDefaultRequestAdapterOptions({
+      ...(powerPreference && { powerPreference }),
+      // MAINTENANCE_TODO: Change this to whatever the option ends up being
+      ...(compatibility && { compatibilityMode: true }),
+    });
+  }
+
+  const testQuerySingleCase = parseQuery(query);
+  let testcase = null;
+  for (const t of g.iterate()) {
```
kainino0x, line 113: Between these two lines, `continue` the loop if the test name doesn't match, so we don't have to iterate all the cases of every test in the file

src/common/tools/gen_listings_and_webworkers.ts
```diff
@@ -78,12 +78,62 @@ export const listing = ${JSON.stringify(listing, undefined, 2)};
...
+};
+
+self.onconnect = (event) => {
+  const port = event.ports[0];
 
-// FIXME: Expose a proxied test interface. I think this can completely replace test_worker-worker.js
-// (using this instead of that), but if not then hopefully it can at least share code with it.
-console.log(oldG.iterate());
+  port.onmessage = (ev) => {
+    void reportTestResults.call(port, ev);
```
kainino0x, line 134: nit: can we just pass `port` as a regular argument instead of using `.call()`?

src/common/tools/gen_listings_and_webworkers.ts
```diff
@@ -78,12 +78,62 @@ export const listing = ${JSON.stringify(listing, undefined, 2)};
...
+  Logger.globalDebugMode = debug;
+  const log = new Logger();
+
+  if (powerPreference || compatibility) {
+    setDefaultRequestAdapterOptions({
+      ...(powerPreference && { powerPreference }),
+      // MAINTENANCE_TODO: Change this to whatever the option ends up being
+      ...(compatibility && { compatibilityMode: true }),
+    });
+  }
```
kainino0x, line 109: This looks copy-pasted, I think? Could it be factored out into a function to avoid duplicated code?

src/webgpu/web_platform/worker/worker.spec.ts
```diff
@@ -51,3 +51,21 @@ g.test('shared_worker')
...
+  .fn(async t => {
+    if (isNode()) {
+      t.skip('node does not support 100% compatible workers');
+      return;
+    }
+    // Note: we load worker_launcher dynamically because ts-node support
+    // is using commonjs which doesn't support import.meta. Further,
+    // we need to put the url in a string add pass the string to import
+    // otherwise typescript tries to parse the file which again, fails.
+    // worker_launcher.js is excluded in node.tsconfig.json.
```
kainino0x, line 66: We don't really need three copies of this comment, perhaps move it just before the definition of `isNode`.

Also "string add pass" should be "string and pass"

src/webgpu/web_platform/worker/worker.spec.ts
```diff
@@ -51,3 +51,21 @@ g.test('shared_worker')
     const result = await launchSharedWorker();
     assert(result.error === undefined, `should be no error from worker but was: ${result.error}`);
   });
+
+g.test('service_worker')
+  .desc(`test WebGPU is available in service workers and check for basic functionality`)
+  .fn(async t => {
+    if (isNode()) {
+      t.skip('node does not support 100% compatible workers');
+      return;
```
kainino0x, line 60: Return is redundant because skip throws an exception. Also I think you can use:
`t.skipIf(isNode(), '...');`
instead.

kainino0x (commented):
This looks great so far, not too complex!

src/common/runtime/helper/test_worker.ts
beaufortfrancois: Done.

src/common/runtime/helper/test_worker.ts
```diff
@@ -94,3 +94,53 @@ export class TestSharedWorker {
...
+    registration.active?.postMessage({
+      query,
+      expectations,
+      ctsOptions: this.ctsOptions,
+    });
+    const serviceWorkerResult = await new Promise<LiveTestCaseResult>(resolve => {
+      this.resolvers.set(query, resolve);
+      void registration.unregister();
+    });
+    rec.injectResult(serviceWorkerResult);
```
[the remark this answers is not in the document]
beaufortfrancois: I was not able to create a proper helper method for `run()` as we use either a Worker, MessagePort, or ServiceWorkerRegistration and we do different cleaning. Let me know if that's okay with you.

src/common/tools/gen_listings_and_webworkers.ts
beaufortfrancois: I cleaned up this code so that we don't use `call()` anymore here.

src/common/tools/gen_listings_and_webworkers.ts
beaufortfrancois: Done. Note that I had to export `comparePath()`

src/common/tools/gen_listings_and_webworkers.ts
beaufortfrancois: The `reportTestResults()` function used by dedicated workers and shared workers in `src/common/runtime/helper/test_worker-worker.ts` is similar in the first part, but not in how handle discovering the test case and running it.
I didn't find a way to avoid duplication code.

src/common/tools/dev_server.ts
beaufortfrancois: I wish I'd understand that bit ;)
Can you expand on that?

beaufortfrancois (commented):
Thank you @kainino0x for the review!

src/common/tools/gen_listings_and_webworkers.ts
beaufortfrancois: Done

src/common/tools/dev_server.ts
kainino0x: Yes, sorry. For the on-disk builds, we generate a bunch of files and write them to disk. However, the dev server does all building in-memory. When the webpage requests a .js file, we builds the .ts file. When it requests a listing.js file, we generate one by crawling the src directory for tests. Similarly here when it requests a .worker.js file, we need to generate one in the same way 

The generation code of course should be shared between the dev server and the gen_listings_and_webworkers script.

src/common/tools/gen_listings_and_webworkers.ts
kainino0x: I mean just the selection of code I put this comment on - from `const ... = ctsOptions` to `setDefaultRequestAdapterOptions`.
This part has to be updated every tim ewe add new global test/adapter options so it would be nice to have it in one place.

src/common/tools/dev_server.ts
kainino0x: I see. We still don't need a large scope though, which is specified in `navigator.serviceWorker.register` as `scope: '/'`. Can we do the other option:

> Adjust the scope

and just use the default scope which is minimal and seems like it should work fine.

kainino0x: Also in particular we need this to work on gpuweb.github.io and we can't set headers there.

src/common/internal/query/compare.ts
```diff
@@ -58,7 +58,7 @@ function compareOneLevel(ordering: Ordering, aIsBig: boolean, bIsBig: boolean):
   return Ordering.Unordered;
 }
 
-function comparePaths(a: readonly string[], b: readonly string[]): Ordering {
+export function comparePaths(a: readonly string[], b: readonly string[]): Ordering {
```
kainino0x, line 61: Since this is exported now, it should have documentation.
```suggestion
/** Compare two file paths, or file-local test paths, returning an Ordering between the two. */
export function comparePaths(a: readonly string[], b: readonly string[]): Ordering {
```

beaufortfrancois:
> nit: Avoid force-pushing: https://github.com/gpuweb/cts/blob/main/docs/intro/developing.md#pull-requests

I will be more vigilant in the future. Sorry for that.

src/common/internal/query/compare.ts
beaufortfrancois: Done

src/common/tools/dev_server.ts
beaufortfrancois: You're right. I've removed `scope` and it works fine now.

src/common/runtime/helper/test_worker.ts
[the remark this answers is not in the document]
kainino0x: It can't be the whole `run` method but everything from `postMessage` to `injectResult` is almost the same.
Here's a proposal which isn't actually simpler but I think it would be a bit more maintainable:
```ts
// also use this definition in wrapTestGroupForWorker and test_worker-worker.ts
interface WorkerTestRunRequest {
  query: string,
  expectations: TestQueryWithExpectation[] = [],
  ctsOptions: CTSOptions,
}
```
```ts
async makeRequestAndRecordResult(target: /* ? */, request: WorkerTestRunRequest) {
  target.postMessage({
    query,
    expectations,
    ctsOptions: this.ctsOptions,
  });
  const serviceWorkerResult = await new Promise<LiveTestCaseResult>(resolve => {
    this.resolvers.set(query, resolve);
  });
  rec.injectResult(serviceWorkerResult);
}
```
and for service worker:
```ts
await this.makeRequestAndRecordResult(/* ... */);
void registration.unregister();
```

kainino0x:
All good! Left one more comment.

src/common/tools/dev_server.ts
beaufortfrancois: IIUC (which is not likely) gen_listings_and_webworkers does not generate code. It writes to a file that is then handled by dev_server and processed. How would the generation code work there?

src/common/runtime/helper/test_worker.ts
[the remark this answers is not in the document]
beaufortfrancois: Note: `rec: TestCaseRecorder` is missing in `makeRequestAndRecordResult()`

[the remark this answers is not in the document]
kainino0x: ah yes, that would need to be passed through too. Ends up being very verbose :(

src/common/tools/gen_listings_and_webworkers.ts
```diff
@@ -0,0 +1,90 @@
...
+
+function usage(rc: number): void {
+  console.error(`Usage: tools/gen_listings [options] [OUT_DIR] [SUITE_DIRS...]
+
+For each suite in SUITE_DIRS, generate listings into OUT_DIR/{suite}/listing.js,
+and generate Web Worker proxies in OUT_DIR/{suite}/webworker/**/*.worker.js for
+every .spec.js file. (Note {suite}/webworker/ is reserved for this purpose.)
+
+Example:
+  tools/gen_listings gen/ src/unittests/ src/webgpu/
```
kainino0x, line 15: oops, I missed something here
```suggestion
  tools/gen_listings_and_webworkers gen/ src/unittests/ src/webgpu/
```

src/common/tools/dev_server.ts
kainino0x: `gen_listings_and_webworkers` is only used in the offline build, it shouldn't affect the dev server at all. It generates `gen/*/listing.js` and `gen/*/webworker/**/*.worker.js`. Those are then copied into the out directories by Grunt.

In the dev server instead we have a special request handler when a `listing.js` file is requested:
`app.get('/out/:suite([a-zA-Z0-9_-]+)/listing.js', async (req, res, next) => {`
and we generate the listing and return it, instead of using the on-disk `listing.js`.

src/common/runtime/helper/test_worker.ts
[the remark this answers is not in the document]
kainino0x: (I think it's probably still better though, for maintainability.)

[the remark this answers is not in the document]
beaufortfrancois: Done.

src/common/tools/gen_listings_and_webworkers.ts
beaufortfrancois: Done

src/common/tools/dev_server.ts
beaufortfrancois: Thank you for bearing with me @kainino0x. I think I now understand what you were telling me (for a long time). Let me know if latest commit looks good to you.

src/common/runtime/helper/utils_worker.ts
```diff
@@ -4,6 +4,9 @@ import { setDefaultRequestAdapterOptions } from '../../util/navigator_gpu.js';
 
 import { CTSOptions } from './options.js';
 
+/*
```
kainino0x, line 7: Make this a doc-comment:
```suggestion
/**
```

src/common/tools/dev_server.ts
```diff
@@ -147,19 +147,30 @@ app.get('/out/:suite([a-zA-Z0-9_-]+)/listing.js', async (req, res, next) => {
...
+  if (jsUrl.endsWith('.worker.js')) {
+    const specUrl = req.url.replace(/\/webworker/, '').replace(/\.worker.js$/, '.spec.js');
+    const result = `\
+import { g } from '${specUrl}';
+import { wrapTestGroupForWorker } from '/out/common/runtime/helper/wrap_for_worker.js';
+
+wrapTestGroupForWorker(g);
+`;
+    res.setHeader('Content-Type', 'application/javascript');
+    res.send(result);
```
kainino0x, line 160:
```suggestion
    res.send(result);
    return;
```

src/common/tools/dev_server.ts
kainino0x: No problem, looks good!

src/common/runtime/helper/test_worker.ts
```diff
@@ -2,9 +2,16 @@ import { LogMessageWithStack } from '../../internal/logging/log_message.js';
 import { TransferredTestCaseResult, LiveTestCaseResult } from '../../internal/logging/result.js';
 import { TestCaseRecorder } from '../../internal/logging/test_case_recorder.js';
 import { TestQueryWithExpectation } from '../../internal/query/query.js';
+import { assert } from '../../util/util.js';
 
 import { CTSOptions, kDefaultCTSOptions } from './options.js';
 
+interface WorkerTestRunRequest {
+  rec: TestCaseRecorder;
```
kainino0x, line 10: I'd like to use WorkerTestRunRequest as the actual object that we post to the worker so the worker can cast to it (in wrapTestGroupForWorker and test_worker-worker.ts)
though I see that because we have `this.ctsOptions` that might not work. Let me look into it a little.

src/common/tools/dev_server.ts
```diff
@@ -147,19 +147,30 @@ app.get('/out/:suite([a-zA-Z0-9_-]+)/listing.js', async (req, res, next) => {
 // Serve all other .js files by fetching the source .ts file and compiling it.
 app.get('/out/**/*.js', async (req, res, next) => {
   const jsUrl = path.relative('/out', req.url);
+  // Worker files are generated on the fly.
+  if (jsUrl.endsWith('.worker.js')) {
+    const specUrl = req.url.replace(/\/webworker/, '').replace(/\.worker.js$/, '.spec.js');
```
kainino0x, line 152: Note to self to see if Express allows us to do this with a route.

kainino0x (approved):
LGTM, I need to try this locally and I might address a few nits

src/common/runtime/helper/test_worker.ts
kainino0x: Ended up using it only in the postMessage/onmessage interface and not in makeRequestAndRecordResult. It's still nice though.

src/common/tools/dev_server.ts
```diff
@@ -147,6 +147,20 @@ app.get('/out/:suite([a-zA-Z0-9_-]+)/listing.js', async (req, res, next) => {
...
+    const result = `\
+import { g } from '${specUrl}';
+import { wrapTestGroupForWorker } from '/out/common/runtime/helper/wrap_for_worker.js';
+
+wrapTestGroupForWorker(g);
+`;
```
kainino0x, line 158: Ideally this can be deduplicated with gen_listings_and_webworkers

src/common/runtime/helper/test_worker.ts
```diff
@@ -2,95 +2,125 @@ import { LogMessageWithStack } from '../../internal/logging/log_message.js';
...
+    const registration = await navigator.serviceWorker.register(serviceWorkerPath, {
+      type: 'module',
     });
-    rec.injectResult(workerResult);
+    await registration.update();
+    navigator.serviceWorker.onmessage = ev => this.onmessage(ev);
+    assert(!!registration.active);
+    const request: WorkerTestRunRequest = { rec, query, expectations };
+    await this.makeRequestAndRecordResult(registration.active, request);
+    void registration.unregister();
```
kainino0x, line 124: Running this locally on /standalone/ it's very slow. I think the problem is that we're creating and destroying a ServiceWorker for every test case. We really only want to create one ServiceWorker for each .worker.js, at most (= one per spec file). Which means we need to unregister later than this anyway.

This can be a followup though.

src/common/runtime/helper/test_worker.ts
```diff
@@ -2,95 +2,125 @@ import { LogMessageWithStack } from '../../internal/logging/log_message.js';
...
-      this.resolvers.set(query, resolve);
+    const [suite, name] = query.split(':', 2);
+    const fileName = name.split(',').join('/');
+    const serviceWorkerPath = `/out/${suite}/webworker/${fileName}.worker.js`;
+
+    const registration = await navigator.serviceWorker.register(serviceWorkerPath, {
+      type: 'module',
     });
-    rec.injectResult(workerResult);
+    await registration.update();
```
kainino0x, line 119: @beaufortfrancois Is this supposed to be the right way to wait for a registration to be active? It appears to be redundant, and it's really slow (90-100ms), but I can't find any reliable way to wait until the service worker is active. `navigator.serviceWorker.ready` is global, and doesn't work anyway (probably because of the scope). 

The best idea I have so far is
```js
      while (!registration.active) {
        await new Promise(resolve => setTimeout(resolve, 0));
      }
```
which is much faster (5-6ms).

kainino0x:
@beaufortfrancois I pushed some commits polishing up some bits. Please review, then this should be ready to squash-and-merge!

src/common/runtime/helper/test_worker.ts
beaufortfrancois: `await registration.update();` is actually required to make sure the service worker is registered with the new code.

Steps to reproduce issue:
1. Go to http://localhost:8080/standalone/?worker=service&q=webgpu:api,operation,*
2. Run "webgpu:api,operation,adapter,requestAdapter:*" subtree
3. Run "webgpu:api,operation,adapter,requestAdapterInfo:*" subtree

Without `registration.update();`, the second service worker is the one from the first subtree.

beaufortfrancois:
@kainino0x Everything but https://github.com/gpuweb/cts/pull/3419/commits/ffff9fdd345989d7fd1c00ab9fcf241cc2db7966#r1517395245 looks good to me. Thank you for polishing this PR!

src/common/runtime/helper/test_worker.ts
beaufortfrancois: LGTM ;)
Please merge

src/common/runtime/helper/wrap_for_worker.ts
```diff
@@ -10,25 +11,33 @@ import { setupWorkerEnvironment, WorkerTestRunRequest } from './utils_worker.js'
...
+          }
         }
       }
+      assert(!!testcase, 'testcase not found');
+      const [rec, result] = log.record(query);
+      await testcase.run(rec, testQuery, expectations);
+
+      ev.source?.postMessage({ query, result });
+    } catch (thrown) {
+      const ex = thrown instanceof Error ? thrown : new Error(`${thrown}`);
```
beaufortfrancois, line 36: Out of curiosity, when is `thrown` not an Error?

kainino0x: It probably won't ever be, but it's theoretically possible, so TypeScript makes us handle it. Since this is error handling code I prefer to handle it than to just type-cast from `unknown` to `Error`. Otherwise a new exception could be thrown and lost the same way `thrown` was lost before.

beaufortfrancois:
@kainino0x I think we're good to merge now. I'll let you do it.

PR #3425, opened 2024-02-21
Add compat mode testing to the PR checklist

greggman:
I'm not sure it's valid to put this here as for example, Safari will not have compat mode. But, at least at the moment, we have the issue that new tests can fail in compat and should either be refactored if they can be or else skipped.

It's also possible to enforce compat validation via an extension so I could go write one for Safari if the need is there. 

Most just hoping that adding this will remind me and others to test compat if possible.


kainino0x (approved):
LGTM. This list isn't really enforced anyway and just serves as a list of reminders.

Issue #1631, opened 2022-07-06
wgsl: Bitwise tests


dj2:
# Complement
* [x] `~` for i32, u32 and vec of each
* [x] `~` for AbstractInt, and vectors of AbstractInt

# Or, And, XOR
(`|`, `&` and `^` already have tests in shader/execution/expression/binary/bitwise.spec.ts)
For each of those, validate mixed types fails to parse

# Left Shift expressions
* [x] `1 << 2`
* [x] Modulo for i32/u32 shift e.g. `1i << 33` shifts `== 2`
* [ ] most significant bits must be same for abstract (1<<63 is error, 2 <<62, etc)
* [x] `1 << -1` is validation error
* [x] `1i << -1` is validation error
* [x] `1u << -1` is validation error
* [x] Component wise shift with `vec`
* [x] `vec` sizes don't match for shift is error

# Right shift expressions
* [x] `0x80000000{i|u} >> 32` == `0x8000000` (shift is modulo)
* [x] `0x80000000{i|u} >> 31` == `1`
* [x] `0x80000000 >> 32` == `0` (abstract)
* [x] `1{i32, u32} >> -1` is validation error
* [x] `1{Abstract} >> -1` is validation error
* [x] `vec` sizes don't match is validation error
* [x] `vec` shifts component wise.

PR #3440, opened 2024-02-27
wgsl: Implement AbstractInt complement (~) execution tests

zoddicus:
Issue #1631

<hr>


src/webgpu/shader/execution/expression/unary/ai_complement.cache.ts
```diff
@@ -0,0 +1,11 @@
+import { abstractInt } from '../../../../util/conversion.js';
+import { fullI64Range } from '../../../../util/math.js';
+import { makeCaseCache } from '../case_cache.js';
+
+export const d = makeCaseCache('unary/ai_complement', {
```
ben-clayton, line 5: Having a case cache for integer tests like this seems like it would be a pessimism. I would strongly suspect that the overhead for deserializing these cases would far outweigh the cost of runtime generation.

zoddicus: The existing integer complement tests use this pattern. Should I convert them away from using the cache too?

ben-clayton: Given that there's no heavy interval logic being precomputed, that seems like a good idea.

zoddicus: Done

ben-clayton (approved):

PR #3446, opened 2024-02-28
wgsl: Implement AbstractFloat matrix multiplcation tests

zoddicus:
Issue #1626

<hr>


zoddicus:
PTAL, this depends on the implementation of dot, so all of the new code is in the later diff.

src/webgpu/listing_meta.json
```diff
@@ -918,6 +918,11 @@
   "webgpu:shader,execution,expression,binary,af_multiplication:scalar_vector:*": { "subcaseMS": 2025.534 },
   "webgpu:shader,execution,expression,binary,af_multiplication:vector:*": { "subcaseMS": 710.667 },
   "webgpu:shader,execution,expression,binary,af_multiplication:vector_scalar:*": { "subcaseMS": 2085.300 },
+  "webgpu:shader,execution,expression,binary,af_matrix_matrix_multiplication:matrix_matrix:*": { "subcaseMS": 0 },
```
ben-clayton, line 921: Those binary files are huge. I'd be curious to get a sense of how long these take to run. My spidey-sense is telling me that we have too many permutations here.

zoddicus: Took a bit of time to at the case generation code. There isn't any obvious accidently using dense ranges vs sparse ones that I can see relative to the other floating point types.

Looking at the size of .bins, the AbstractFloat ones are actually slightly smaller than the equivalent f32 ones, propbably because though the size will be inflated due to 32 vs 64 bits, there are fewer total cases due to OOB being trimmed.

Time to generate the case cache doesn't appear to be significantly affected by whether or not this PR is present.

With a pre-built case cache, running locally on lava pipe,
`-j 1 'webgpu:shader,execution,expression,binary,af_matrix_scalar_multiplication,*'` takes 5m30.410694302s

and

`-j 1 'webgpu:shader,execution,expression,binary,f32_matrix_scalar_multiplication,*'` takes 14.364973476s

So yeah, AF is much much slower than f32, but I think that is because of the all of the extraction logic that needs to be run in the shader vs cache size.

ben-clayton: Thank you for doing investigations here.
5½ minutes is a huge amount of time for a single test case. I strongly suspect that's going to cause timeouts for our test runner and likely others. I think we're going to have to take a look at reducing the case permutations.

src/webgpu/shader/execution/expression/binary/af_matrix_matrix_multiplication.cache.ts
```diff
@@ -1,23 +1,23 @@
+import { crc32 } from '../../../../../common/util/crc32.js';
 import { FP } from '../../../../util/floating_point.js';
 import { sparseMatrixF64Range } from '../../../../util/math.js';
-import { StochasticFilter } from '../../../../util/stochastic_filter.js';
 import { makeCaseCache } from '../case_cache.js';
 
-const sf = new StochasticFilter(0);
+const filter_percentage = 50;
```
ben-clayton, line 6: Please document.

src/webgpu/shader/execution/expression/binary/af_matrix_matrix_multiplication.cache.ts
```diff
@@ -0,0 +1,27 @@
...
+      ([2, 3, 4] as const).map(rows => ({
+        [`mat${k}x${rows}_mat${cols}x${k}`]: () => {
+          return FP.abstract
+            .generateMatrixPairToMatrixCases(
+              sparseMatrixF64Range(k, rows),
+              sparseMatrixF64Range(cols, k),
+              'finite',
+              FP.abstract.multiplicationMatrixMatrixInterval
+            )
+            .filter(c => filter_percentage <= crc32(c.input.toString()) % 100);
```
ben-clayton, line 20: It might be preferable to specify an absolute target value instead of a percentage. The target value could then be used to calculate the relative threshold (`0xffffffff*n/count <= crc32(c.input.toString())` or something)

This could also be put in a reusable helper function.

zoddicus: Done

zoddicus:
PTAL, this should be ready for review

src/webgpu/shader/execution/expression/case.ts
```diff
@@ -25,6 +27,41 @@ export type Case = {
...
+ * @param cases list of Cases to be selected from.
+ */
+export function selectNCases(n: number, cases: CaseList): CaseList {
+  assert(n > 0 && Math.round(n) === n, `n ${n} is expected to be a positive integer`);
+  const count = cases.length;
+  if (n >= count) {
+    return cases;
+  }
+
+  return cases.filter(c => n * (0xffff_ffff / count) > crc32(c.input.toString()));
```
ben-clayton, line 62: Idea: Should we xor-in another `crc32` of a string parameter (i.e. cache key name)? I'm thinking that we'll get the same N-cases selected for different tests with the same parameters. XOR-ing in something unique to the test would prevent this.

ben-clayton (approved):
Approving - although if you like the xor idea, it might be good to do that as part of this PR, to avoid bloating the git history.

src/webgpu/shader/execution/expression/case.ts
zoddicus: That is a good point

Issue #3363, opened 2024-02-03
Make all of src/webgpu/api/operation use max limits


greggman:
IIUC, the current CTS doesn't test max limits operationally. It only tests the default limits and it tests validation for max limits. 

Would it be a good idea to make a MaxLimitTest that requires all max limits and have all relevant tests under src/webgpu/api/operation use that so more tests the maximum limits?

WebGL always tested maximum limits. I don't know what bugs, if any, to expect, but, it's at the max limits where the edge cases are. 

I don't think we have the capacity to write a whole bunch of new tests nor do we want to run so many tests twice, once at default and once at max. So, maybe something like just making a bunch of tests in the operation subtree use max limits would more thoroughly test?

greggman:
Thoughts? @kainino0x @kdashg @Kangz @mwyrzykowski

mwyrzykowski:
Seems reasonable to me

kainino0x:
yes, I think we should. Having it be enabled on a per test basis is probably good, rather than doing this globally for operation tests, because some existing tests could break (or lose coverage) otherwise.

PR #3466, opened 2024-03-07
Add AdapterLimitsGPUTest that sets all limits to the adapter's

greggman:
Note: I'm not sure how I refactored device_pool.ts is good but, the old code created one initial device with no descriptor the first time in acquire. It then created another of the type actually wanted by the user. That no longer works becauase the adapter is passed in and so if the first requestDevice succeeds, then the adapter has been used and can't be used for the second one.

Also, assuming this change is approved,  I can (and should?) probably refactor LimitsTest to use this vs how it's doing it now. That would end up using cached devices.

Issue: #3363

<hr>


kainino0x (approved):
LGTM. This makes more sense than the thing that tried to get a device twice.

greggman:
sorry about that. I'll fix assp

PR #3472, opened 2024-03-08
Type texture

shrekshao:
First try adding to WGSL validation tests: texel_formats, storage_texture_types

Failing cases in Chrome:
~~- storage_texture_types: format without readWriteStorage usage didn't get rejected with 'read-write' access.~~
- bgra8unorm

For other texture types (e.g. sampled textures, multisampled textures etc.) I think they need api side binding creation to test the format validation. Do we want to include that as part of these shader validation tests?

<hr>


src/webgpu/shader/validation/types/textures.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+    `Test that for texture_storage_xx<format, access>
+- format must be an enumerant for one of the texel formats for storage textures
+- access must be an enumerant for one of the access modes
+`
+  )
+  .params(u =>
+    u
+      .combine('access', ['read', 'write', 'read_write'] as const)
+      .combine('format', kAllTextureFormats)
+      // bgra8unorm-storage feature is tested at webgpu,api,validation,texture,bgra8unorm_storage
```
ben-clayton, line 81: Overlap is fine. Suggest removing.

shrekshao: Remove filter (add bgra8unorm).
Seems to fail for all access (when bgra8unorm-storage is not requried)

ben-clayton: > Seems to fail for all access (when bgra8unorm-storage is not requried)

Sorry, please can you elaborate a little? You believe you've found an implementation bug?

shrekshao:
Yeah seems to be an implementation bug. But I'm not 100% sure given there's
createShaderModule test in the bgra8unorm_stoarge test that passes, with
and without bgra8unorm_storage feature

src/webgpu/shader/validation/types/textures.spec.ts
ben-clayton: Please can you use `run-cts --dump-shaders` and file a tint bug with the WGSL that you think is broken? I'll investigate.

shrekshao: > So I think the problem is that the test uses the API's expectations for whether this is valid, instead of WGSL's.

Do you refer to `bgra8unorm_storage.spec.ts` or "the test added in this CL" for "the tests"?

shrekshao: FYI the dumpped wgsl for `webgpu:shader,validation,types,textures:storage_texture_types:access="read";format="bgra8unorm"`
```
// Dumped WGSL:
@group(0) @binding(0) var tex: texture_storage_2d<bgra8unorm, read>;
```

shrekshao: For this test, it uses info hardcoded in `format_info.ts` for storage and readWriteStorage usage.
for bgra8unorm, both storage and readWriteStorage are false. If WGSL should always accept the format as storage, that information from `format_info.ts` seems to include API restriction.

Given that l think for bgra8unorm we should skip it here, and let `bgra8unorm_storage.spec.ts` to test for this special case. WDYT?

jrprice: I meant the tests added here, but it seems that `bgra8unorm_storage.spec.ts` also expects shader modules containing `bgra8unorm_storage` storage textures to fail when `bgra8unorm-storage` is not present. That doesn't seem quite right to me, since I don't see anywhere in the API spec that imposes format validation on the shader module itself.

shrekshao: Update the test with reinterpreted spec. Now the only case failing is bgra8unorm. PTAL

src/webgpu/shader/validation/types/textures.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+  )
+  .fn(t => {
+    const { format, access } = t.params;
+    const info = kTextureFormatInfo[format];
+    let isFormatValid =
+      info.color?.storage || info.depth?.storage || info.stencil?.storage || false;
+    if (t.isCompatibility) {
+      // Adjust if storage is supported under compatibility mode for formats in kCompatModeUnsupportedStorageTextureFormats.
+      isFormatValid = isTextureFormatUsableAsStorageFormat(format, t.isCompatibility);
+    }
```
jrprice, line 92: I'm not 100% sure, but I don't think compat's restrictions here should cause shader module creation to fail. Much like the `bgra8unorm` thing, I imagine that should be validated at texture creation and BGL creation.

src/webgpu/shader/validation/types/textures.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+`
+  )
+  .params(u =>
+    u.combine('access', [...kAccessModes, 'storage'] as const).combine('format', kAllTextureFormats)
+  )
+  .fn(t => {
+    const { format, access } = t.params;
+    const info = kTextureFormatInfo[format];
+    let isFormatValid =
+      info.color?.storage || info.depth?.storage || info.stencil?.storage || false;
```
jrprice, line 88: Nit: Remove `|| false`?

src/webgpu/shader/validation/types/textures.spec.ts
shrekshao: I think you are right. Removed.

src/webgpu/shader/validation/types/textures.spec.ts
shrekshao: Without the `|| false`, type of `isFormatValid` is `boolean | undefined`. Keep as is (Change `let` to `const`)

jrprice: Ah I see, I missed the `?`s.

jrprice (approved):
LGTM

PR #3478, opened 2024-03-12
Test `textureSample` WGSL validation

greggman:
Looking for feedback / guidance since this is my first WGSL test - I'm not sure every case that should be tested. 

`textureSample` takes a bunch of formats. I tested that the wrong type can't be passed to each parameter except for the texture (not sure how you'd pass an incorrect texture type since the function is determined from the texture type. I suppose I could test passing an external texture ? Similarly I didn't test not passing a sampler since there is only one type of sampler.

I wasn't exactly sure how to tell if types are compatible so I hacked in `isCompatible`. I wanted the test to at least have some success cases because if it only expects failure than any typo will cause it appear to be correctly failing all tests.

I also wasn't sure what to check for const-expression. The spec says the offset parameter must be a `const-experssion` so I made 3 variables `c` (a const) `u` an uniform, `l` a `let` expression based off something non-uniform. It works but I'm not sure if there is some more official way or if there are more types I should be testing. I was surprised a `let` expressions like `let a = 1` is not considered const even though AFAICT it is actually const. I'm guessing that's in the spec but it seems unintuitive.


src/webgpu/shader/validation/expression/call/builtin/textureSample.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+    @fragment fn fs() -> @location(0) vec4f {
+      let v = textureSample(t, s, ${coordWGSL}${arrayWGSL}${offsetWGSL});
+      return vec4f(0);
+    }
+    `;
+
+    const expectSuccess = isConvertible(coordType, coords);
+
+    t.expectCompileResult(expectSuccess, code);
+  });
```
ben-clayton, line 137:
```suggestion
g.test('coords_argument')
  .desc(
    `
Validates that only incorrect coords arguments are rejected by ${builtin}
`
  )
  .params(u =>
    u
      .combine('textureType', keysOf(kValidTextureSampleParameterTypes))
      .combine('coordType', keysOf(kValuesTypes))
      .beginSubcases()
      .combine('value', [-1, 0, 1] as const)
      .expand('offset', ({ textureType }) => {
        const offset = kValidTextureSampleParameterTypes[textureType].offset;
        return offset ? [false, true] : [false];
      })
  )
  .fn(t => {
    const { value, offset } = t.params;
    const coordArgType = kValuesTypes[t.params.coordType];
    const coordParamType = kValidTextureSampleParameterTypes[t.params.textureType].coords;
    const offsetParamType = kValidTextureSampleParameterTypes[t.params.textureType].offset;
    const arrayIndex = kValidTextureSampleParameterTypes[t.params.textureType].arrayIndex;

    const coordWGSL = coordArgType.create(value).wgsl();
    const arrayWGSL = arrayIndex ? ', 0' : '';
    const offsetWGSL = offset ? `, ${offsetParamType}(0)` : '';

    const code = `
@group(0) @binding(0) var s: sampler;
@group(0) @binding(1) var t: ${t.params.textureType};
@fragment fn fs() -> @location(0) vec4f {
  let v = textureSample(t, s, ${coordWGSL}${arrayWGSL}${offsetWGSL});
  return vec4f(0);
}
`;

    const expectSuccess = isConvertible(coordArgType, coordParamType);

    t.expectCompileResult(expectSuccess, code);
  });
```

src/webgpu/shader/validation/expression/call/builtin/textureSample.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+        case 'i32':
+        case 'i8':
+          return true;
+        default:
+          return false;
+      }
+    default:
+      return false;
+  }
+}
```
ben-clayton, line 95:
```suggestion
/** @returns true if an argument of type 'src' can be used for a parameter of type 'dst' */
function isConvertible(src: Type, dst: Type) {
  if (src === dst) {
    return true;
  }

  const widthOf = (ty: Type) => {
    return ty instanceof VectorType ? ty.width : 1;
  };

  if (widthOf(src) !== widthOf(dst)) {
    return false;
  }

  const elSrc = scalarTypeOf(src);
  const elDst = scalarTypeOf(dst);

  switch (elSrc.kind) {
    case 'abstract-float':
      switch (elDst.kind) {
        case 'abstract-float':
        case 'f16':
        case 'f32':
        case 'f64':
          return true;
        default:
          return false;
      }
    case 'abstract-int':
      switch (elDst.kind) {
        case 'abstract-int':
        case 'abstract-float':
        case 'f16':
        case 'f32':
        case 'f64':
        case 'u16':
        case 'u32':
        case 'u8':
        case 'i16':
        case 'i32':
        case 'i8':
          return true;
        default:
          return false;
      }
    default:
      return false;
  }
}
```

Might be useful to promote this up to someplace reusable

ben-clayton (commented):
Made some suggestions for how you can use the Type / Value utilities to remove the string comparisons.

ben-clayton:
> I was surprised a `let` expressions like `let a = 1` is not considered const even though AFAICT it is actually const. I'm guessing that's in the spec but it seems unintuitive.

You want `const a = 1` for that to be a const expression.  A `let` and `var` usage is a runtime expression.

src/webgpu/shader/validation/expression/call/builtin/textureSample.spec.ts
ben-clayton: Note that `'f64'`, `'u16'`, `'u8'`, `'i16'`, `'i32'`, `'i8'` do not exist in WGSL. I'm guessing they were added to the framework for future compatibility, but you don't really need to handle them here.

src/webgpu/util/conversion.ts
```diff
@@ -882,10 +882,10 @@ function valueFromBytes<A extends TypedArrayBufferView>(
 const abstractIntType = new ScalarType('abstract-int', 8, (buf: Uint8Array, offset: number) =>
   abstractInt(valueFromBytes(workingDataI64, buf, offset))
 );
-const i32Type = new ScalarType('i32', 4, (buf: Uint8Array, offset: number) =>
+export const i32Type = new ScalarType('i32', 4, (buf: Uint8Array, offset: number) =>
```
ben-clayton, line 885: Doesn't need exporting. Use `Type.i32`. Same for `u32`.

src/webgpu/shader/validation/expression/call/builtin/textureSample.spec.ts
```diff
@@ -0,0 +1,219 @@
...
+import { ShaderValidationTest } from '../../../shader_validation_test.js';
+
+type TextureSampleParams = {
+  coords: string;
+  array_index?: boolean;
+  offset?: string;
+};
+
+const kValidTextureSampleParameterTypes: { [n: string]: TextureSampleParams } = {
+  'texture_1d<f32>': { coords: 'f32' },
```
ben-clayton, line 30: This is the second time that Github has thrown a large comment on the floor.
Earlier today I made a suggestion block here that you replaced these string with `f32` -> `Type.f32`, `vec3<f32>` -> `Type.vec3f`, etc. You can then use `elementTypeOf()` to obtain the element type, instead of more string manipulation with `innerType`.

Maybe my other code examples I provided would make more sense had Github not discarded that comment.

greggman: no worries. thanks for guiding me through these type types.

updated.

src/webgpu/shader/validation/expression/call/builtin/textureSample.spec.ts
```diff
@@ -0,0 +1,223 @@
...
+      .filter(
+        ({ textureType }) => kValidTextureSampleParameterTypes[textureType].offset !== undefined
+      )
+  )
+  .fn(t => {
+    const { textureType, varType } = t.params;
+    const {
+      coords,
+      array_index,
+      offset = Type.vec2f,
```
ben-clayton, line 197: Little bit odd to my not-so-familiar-with-JS-eye to have an assignment in here, then another assignment outside. Might be better to call this `offsetType` too, to match the other types.

ben-clayton (approved):
Great job. One nit, but feel free to ignore.

src/webgpu/shader/validation/expression/call/builtin/textureSample.spec.ts
greggman: I'm conflicted here and would love your input. I picked those names originally because they match the spec

https://gpuweb.github.io/gpuweb/wgsl/#texturesample

But I'm happy to change them. They do seem ... confusing.  Thoughts?

The assignment here offset can be undefined since the set of textureSampler overloads, some include offset and some don't. So, even though all the textureSampler overloads with no offset are filtered out of the parameters, TypeScript complains later that offset might be undefined. I can probably just add a `!` at the right place instead of this `= Type.vec2f`

ben-clayton: Ah - I didn't even realize what this was doing. `offset = Type.vec2f` as a default (which is an invalid type), which gets replaced with `Type.vec2i` if its present on the `kValidTextureSampleParameterTypes` entry - which it always is? What a strange language.

I was wondering if you can get TypeScript to infer that `offset` is never undefined after filtering, but after 20 mins of playing around, I've conceded that my TS-foo is weak. Maybe @kainino0x might know the magic required to narrow the types here.

I'm happy to land this, but a comment here would be good to explain what this is doing.

greggman: I removed it and added a `!` suppression later.

PR #3481, opened 2024-03-12
WGSL: Add binary + / - / * validation tests but for matrices

Kangz:
Issue: #

<hr>


src/webgpu/shader/validation/expression/binary/add_sub_mul.spec.ts
```diff
@@ -0,0 +1,170 @@
...
+  kAllScalarsAndVectors,
+  scalarTypeOf,
+  Type,
+  VectorType,
+} from '../../../../util/conversion.js';
+import { ShaderValidationTest } from '../../shader_validation_test.js';
+
+export const g = makeTestGroup(ShaderValidationTest);
+
+// A list of operators and a flag for whether they support boolean values or not.
```
jrprice, line 17: Nit: comment mentions a flag that isn't there.

src/webgpu/shader/validation/expression/binary/add_sub_mul.spec.ts
```diff
@@ -0,0 +1,170 @@
...
+    const op = kOperators[t.params.op];
+    const lhs = kScalarAndVectorTypes[t.params.lhs];
+    const rhs = kScalarAndVectorTypes[t.params.rhs];
+    const lhsElement = scalarTypeOf(lhs);
+    const rhsElement = scalarTypeOf(rhs);
+    const hasF16 = lhsElement === Type.f16 || rhsElement === Type.f16;
+    const code = `
+${hasF16 ? 'enable f16;' : ''}
+const lhs = ${lhs.create(lhsElement.kind === 'abstract-int' ? 0n : 0).wgsl()};
+const rhs = ${rhs.create(rhsElement.kind === 'abstract-int' ? 0n : 0).wgsl()};
```
jrprice, line 65: Nit: `lhs.create(0)` should work now

src/webgpu/shader/validation/expression/binary/add_sub_mul.spec.ts
```diff
@@ -0,0 +1,170 @@
...
+      elementsCompatible = false;
+
+      // AbstractInt is allowed with everything but booleans which are already checked above.
+    } else if (elementTypes.includes(Type.abstractInt)) {
+      elementsCompatible = true;
+
+      // AbstractFloat is allowed with AbstractInt (checked above) or float types
+    } else if (elementTypes.includes(Type.abstractFloat)) {
+      elementsCompatible = elementTypes.every(e =>
+        ['abstract-float', 'f64', 'f32', 'f16'].includes(e.kind)
```
jrprice, line 83: Nit: could do `elementTypes.every(e => kConvertableToFloatScalar.includes(e))`

jrprice (approved):
We should also add some coverage for f32/f16 expressions whose results go out of range of the type.

Doesn't have to be in this CL, but just want to make sure we don't forget (e.g. add an `.unimplemented()` if you're not planning on doing it soon, or leave a comment on the spreadsheet about the remaining work).

Kangz:
> We should also add some coverage for f32/f16 expressions whose results go out of range of the type.

I'm slightly surprised that out-of-range concrete type expressions would be an error ... ah ok it's because WGSL has builtin consteval of everything? How do we want to test this though? It feels like it would require a lot of range-interval computations, more akin to what we do in execution tests.

jrprice:
> How do we want to test this though? It feels like it would require a lot of range-interval computations, more akin to what we do in execution tests.

Yes, we have helpers we can reuse (e.g. `isRepresentable()`). I'd take a look at the kind of thing we've done for some of the builtin validation tests, e.g. [`atan2`](https://github.com/gpuweb/cts/blob/main/src/webgpu/shader/validation/expression/call/builtin/atan2.spec.ts#L31). The `values` test generates ranges of values using helpers and then uses `isRepresentable()` to determine whether the result should be accepted or not.

We can't quite reuse `validateConstOrOverrideBuiltinEval()` for this, but we could make a similar helper for binary expressions.

As I say, don't have to do this in this PR if it turns into a bunch of extra work.

Kangz:
Thanks for the explanation. I added an unimplemented but will relinquish further tasks as I'll be OOO for a bit.

PR #3498, opened 2024-03-13
Adds validation tests for count* builtin functions.

lokokung:
Piggy-backing off of Brandon's PR at [](https://github.com/gpuweb/cts/pull/3497), also implementing something a bit simpler to test the waters. As Brandon points out in his PR for min/max, some tests (these included) are VERY similar, to a point where pretty much the only difference is the first line of the test files, so I'm also wondering whether this is fine or whether there's precedence for what to do.

More specific to this suite of tests though, the WGSL spec for these count* builtins does not explicitly state that AbstractInts are a part of the parameterization, but during my testing, I noticed that passing an AbstractInt does not fail compilation because of how overload resolution handles it. As a result, I decided to not have a specific test checking that AbstractInt works or fails. Feel free to let me know if that's not what we want.

<hr>


src/webgpu/shader/validation/expression/call/builtin/countLeadingZeros.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+  });
+
+const kGoodArgs = '(1u)';
+const kBadArgs = {
+  // Bad number of args
+  '0args': '',
+  '2args': '(1u,2u)',
+  // Bad value for arg 0
+  '0bool': '(false)',
+  '0array': '(array(1u))',
```
dneto0, line 77: Please add a floating point case, and a vector case (vector float or vector bool)

src/webgpu/shader/validation/expression/call/builtin/countLeadingZeros.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+    validateConstOrOverrideBuiltinEval(
+      t,
+      builtin,
+      /* expectedResult */ false,
+      [type.create(0)],
+      'constant'
+    );
+  });
+
+const kGoodArgs = '(1u)';
```
dneto0, line 70: @ben-clayton gave me feedback about how to improve these.
See discussion here on @toji 's PR
https://github.com/gpuweb/cts/pull/3497#pullrequestreview-1937798113 

For latest advice on how to organize and name cases, see

https://github.com/gpuweb/cts/blob/278383fad33d67229a7a592ce39762ae9f0addf5/src/webgpu/shader/validation/expression/call/builtin/radians.spec.ts#L77

dneto0 (changes requested):
My advice: you don't have to test the good cases such as AbstractInt because that should already be covered by the execution cases.

regarding apparent duplication  /  DRY, we discussed and decided the duplication is useful because it makes it trivial to see and verify coverage for each given builtin.

src/webgpu/shader/validation/expression/call/builtin/countLeadingZeros.spec.ts
lokokung: Done.

src/webgpu/shader/validation/expression/call/builtin/countLeadingZeros.spec.ts
lokokung: Talked offline about this and how this is covered via the `float_argument` case.

lokokung:
Updated the tests according to comments in https://github.com/gpuweb/cts/pull/3565. PTAL! (Also adding Dan as a reviewer since a lot of those comments were from him.)

dj2 (approved):

dneto0 (approved):
Thanks!  Sorry for the delay

PR #3500, opened 2024-03-14
WGSL Validation tests for textureSampleCompare

greggman:
I notice the offset option was missing in one case in the `textureSample` tests so that's included here. I can put it in another PR if it's important.

<hr>


src/webgpu/shader/validation/expression/call/builtin/textureSampleCompare.spec.ts
```diff
@@ -0,0 +1,250 @@
...
+    `
+Validates that only incorrect array_index arguments are rejected by ${builtin}
+`
+  )
+  .params(u =>
+    u
+      .combine('textureType', kTextureTypes)
+      // filter out types with no array_index
+      .filter(
+        ({ textureType }) =>
```
ben-clayton, line 99: nit: `t => kValidTextureSampleParameterTypes[t.textureType].hasArrayIndexArg ?? false` is a bit more common, and IMO a bit easier to read.

src/webgpu/shader/validation/expression/call/builtin/textureSampleCompare.spec.ts
```diff
@@ -0,0 +1,250 @@
...
+          !!kValidTextureSampleCompareParameterTypes[textureType].hasArrayIndexArg
+      )
+      .combine('arrayIndexType', keysOf(kValuesTypes))
+      .beginSubcases()
+      .combine('value', [-9, -8, 0, 7, 8])
+  )
+  .fn(t => {
+    const { textureType, arrayIndexType, value } = t.params;
+    const arrayIndexArgType = kValuesTypes[arrayIndexType];
+    const args = [arrayIndexArgType.create(value)];
```
ben-clayton, line 109: It took me a while to figure out why this was not causing failures for `u32`. `value` includes `-9`, `-8`, and I was expecting this to result in `-9u`, `-8u` which would result in: `error: no matching overload for operator - (u32)`.

However, `ScalarType.create()` calls `u32(value)`, which is implemented as:
https://github.com/gpuweb/cts/blob/f1aa9428a23f53c98aaa9588af36e319b37ed94e/src/webgpu/util/conversion.ts#L1631-L1635

So, we end up with the two's complement. This is unexpected in this context.

We're not trying to test for unsigned integers handling negative numbers, so I'd suggest changing the `value` range to just include positive numbers.

Same thing applies for the other tests.

src/webgpu/shader/validation/expression/call/builtin/textureSampleCompare.spec.ts
```diff
@@ -0,0 +1,250 @@
...
+    const offsetWGSL = `${offsetArgType}(${varType})`;
+    const castWGSL = offsetArgType!.elementType.toString();
+
+    const code = `
+@group(0) @binding(0) var s: sampler_comparison;
+@group(0) @binding(1) var t: ${textureType};
+@group(0) @binding(2) var<uniform> u: ${offsetArgType};
+@fragment fn fs(@builtin(position) p: vec4f) -> @location(0) vec4f {
+  const c = 1;
+  let l = ${offsetArgType}(${castWGSL}(p.x));
```
ben-clayton, line 243: The same test could be achieved with `var l : ${offsetArgType};`, which would reduce the complexity of the shader.

ben-clayton (approved):
Looks good. Just some nits.

src/webgpu/shader/validation/expression/call/builtin/textureSampleCompare.spec.ts
greggman: It needed to be `var l : ${offsetArgType.create(0).wgsl()};,` AFAICT. Just checking I'm not missing something.

src/webgpu/shader/validation/expression/call/builtin/textureSampleCompare.spec.ts
greggman: I see what you're saying about `-8u` getting converted via `Uint32Array` but I need negative numbers to test `offset` which is required to generate an error for values < -8 and it will end up doing the same thing, passing `-8` to `offsetArgType.create()` with those values when `offsetArgType` is `u32`.

I switched to filtering out negative values with unsigned types

PR #3509, opened 2024-03-14
wgsl: Add normalize builtin validation tests

toji:
Adds validation tests for `normalize()`.

Took a while to get the right logic for testing the validity of the numeric values which didn't result in false positives, but at this point all the test cases pass. More than happy to take feedback on how that's being computed, though!

Also, I have no idea what's going on with the changes to hashes.json. I'm assuming it's because I reorganized the scalar/vector types in `conversion.js` in order to get lists of *only* the vector types?

<hr>


toji:
Ugh. Noticed that this is failing on devices with f16 support, I'll work on that.

toji:
Okay, think I worked out all the kinks, just in time for the weekend.

src/webgpu/shader/validation/expression/call/builtin/normalize.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+  kConstantAndOverrideStages,
+  stageSupportsType,
+  validateConstOrOverrideBuiltinEval,
+} from './const_override_validation.js';
+
+export const g = makeTestGroup(ShaderValidationTest);
+
+const kValidArgumentTypes = objectsToRecord(kConvertableToFloatVectors);
+
+// These should probably go in a utilities file somewhere?
```
zoddicus, line 27: Looks like you are just doing fp quanitzation, so I think you should be able to use quantizeToF32 and quantizeToF16 from src/webgpu/util/math.ts

zoddicus (changes requested):

src/webgpu/shader/validation/expression/call/builtin/normalize.spec.ts
toji: You're right! Thanks for pointing out those functions!

zoddicus (approved):

PR #3525, opened 2024-03-18
wgsl: Implement AbstractFloat `step` execution tests

### LINKED ISSUE #2530: wgsl: AbstractFloat execution test for `step` builtin
Add an AbstractFloat execution test for the `step` builtin.
```
@const @must_use fn step(edge: T,
                         x: T) -> T

S is AbstractFloat
T is S or vecN<S>
```
Returns 1.0 if edge ≤ x, and 0.0 otherwise. [Component-wise](https://www.w3.org/TR/WGSL/#component-wise) when T is a vector.

Accuracy is `Correctly rounded`

See https://www.w3.org/TR/WGSL/#step-builtin

zoddicus:
Also switching to using sparse ranges since there are N^2 cases being run.

Fixes #2530

<hr>


src/unittests/floating_point.spec.ts
```diff
@@ -4903,12 +4903,15 @@ g.test('stepInterval')
           { input: [1, -1], expected: 0 },
 
           // 64-bit normals
-          { input: [0.1, 0.1], expected: [0, 1] },
+          // number is f64 internally, so the value for 0.1/-0.1 will be precise
```
dneto0, line 4906: This is ok... but slightly misleading perhaps.  By "precise" did you mean "perfectly accurate"? Because  0.1 is not exactly representable in binary floating point. (Because 1/5 is not a terminating binary fraction).

use something like 0.375  ( = 3/8) for a perfectly representable fraction.

dneto0 (approved):

src/unittests/floating_point.spec.ts
zoddicus: I agree I wasn't clear here, I was getting at the fact that number is internally a f64, the same as abstract float, so no matter what normal literal we put in there, we will never be able to create the the step could go either way behaviour that lower precision floating points have. I will clarify the comment.

PR #3543, opened 2024-03-21
wgsl: Validation for quantizeToF16 builting function

toji:
Patterned after the `normalize()` tests.
<hr>


src/webgpu/shader/validation/expression/call/builtin/quantizeToF16.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+  Type.vec(2, Type.abstractFloat),
+  Type.vec(3, Type.abstractFloat),
+  Type.vec(4, Type.abstractFloat),
+  ...kConcreteF32ScalarsAndVectors,
+]);
+
+g.test('values')
+  .desc(
+    `
+Validates that constant evaluation and override evaluation of ${builtin}() never errors
```
dneto0, line 32: nit:  It *does* error.  Comment is out of date.

src/webgpu/shader/validation/expression/call/builtin/quantizeToF16.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+      .filter(u => stageSupportsType(u.stage, kValidArgumentTypes[u.type]))
+      .beginSubcases()
+      .expand('value', u => fullRangeForType(kValidArgumentTypes[u.type]))
+  )
+  .fn(t => {
+    let expectedResult = true;
+
+    // Should be invalid if the quantized value exceeds the maximum representable
+    // 16-bit float value.
+    const f16Value = quantizeToF16(Number(t.params.value));
```
dneto0, line 48: what about NaNs?

Does hfround yield NaNs?

src/webgpu/shader/validation/expression/call/builtin/quantizeToF16.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+  good: '(vec3f())',
+  bad_no_parens: '',
+  // Bad number of args
+  bad_0args: '()',
+  bad_2arg: '(1.0, 2.0)',
+  // Bad value for arg 0
+  bad_0bool: '(false)',
+  bad_0array: '(array(1.1,2.2))',
+  bad_0struct: '(modf(2.2))',
+  bad_0f16: '(1h)',
```
dneto0, line 74: This is interesting. It can fail for two reasons:
shader-f16 not supported, so the literal doesn't parse.
Or the overload is not available.

So this is correct, for a subtle reason.

dneto0 (changes requested):
Just a question about NaNs.

src/webgpu/shader/validation/expression/call/builtin/quantizeToF16.spec.ts
toji: Looks like hfround will return a NaN if a NaN is passed in. (Because passing NaN to `Number.isFinite()` returns `false`)

I thought WGSL officially didn't have NaNs, though? So if we're not explicitly passing NaNs in here doesn't that still cover the usable value range?

src/webgpu/shader/validation/expression/call/builtin/quantizeToF16.spec.ts
toji: Good point! Updated this to be less "accidentally correct". :)

src/webgpu/shader/validation/expression/call/builtin/quantizeToF16.spec.ts
toji: Fixed

src/webgpu/shader/validation/expression/call/builtin/quantizeToF16.spec.ts
zoddicus: I agree with the assessment that hfround should only return NaN if NaN is passed in. NaN should not be part of the values being generated, thus hfround should never generate a NaN here.

zoddicus (approved):

src/webgpu/shader/validation/expression/call/builtin/quantizeToF16.spec.ts
toji: David, any further concerns here?

dneto0: Ah right.  If NaN is in the source then the shader will fail the compile anywya.

dneto0 (approved):
LGTM. Thanks for the extra consideration.

PR #3536, opened 2024-03-20
Disable logToWebSocket by default

kainino0x:
Fixes #3504

<hr>

**Requirements for PR author:**

- [N/A] All missing test coverage is tracked with "TODO" or `.unimplemented()`.
- [x] New helpers are `/** documented */` and new helper files are found in `helper_index.txt`.
- [x] Test behaves as expected in a WebGPU implementation. (If not passing, explain above.)
- [N/A] Test have be tested with compatibility mode validation enabled and behave as expected. (If not passing, explain above.)

**Requirements for [reviewer sign-off](https://github.com/gpuweb/cts/blob/main/docs/reviews.md):**

- [N/A] Tests are properly located in the test tree.
- [N/A] [Test descriptions](https://github.com/gpuweb/cts/blob/main/docs/intro/plans.md) allow a reader to "read only the test plans and evaluate coverage completeness", and accurately reflect the test code.
- [ ] Tests provide complete coverage (including validation control cases). **Missing coverage MUST be covered by TODOs.**
- [ ] Helpers and types promote readability and maintainability.

When landing this PR, be sure to make any necessary issue status updates.

PR #3548, opened 2024-03-22
Don't validate or warn about listing_meta.json contents

### LINKED ISSUE #3522: Clarification on `src/webgpu/listing_meta.json`
With the increase of people working on the WGSL CTS, I have seen a noticeable uptick in people tripping over needing to add entries to  `src/webgpu/listing_meta.json` and I am haven't been adequately able explain its neccessity/usage, since I wasn't involved in the design/roll out.

So I have some questions/comments, that hopefully can be answered, and maybe documentation can be updated based on the responses.

My understanding is that this file is a meant to a be a set of estimates of the time taken to run tests, to assist in load balancing of execution shards. (But to my knowledge no one is currently using it).
- Is this actively being used at the moment? i.e. is there any implementors of WebGPU that use this information as part of their their CI, etc.
- If not, is it being actively worked on as a near term project?
- If it is not, should not having these entries be a blocking error for commits or even a warning for zeroes?

There is some concerns about methodology, because there doesn't appear to be a common baseline implementation/machine being used.
Specifically people have called out that they are generating values orders of magnitude different from what is in the file, which may be partially due to issues in dawn.node. But I am not aware of any mitigation of the situation of person a) generates some values on a slower CPU based run  and person b) generates others on a fast modern GPU. (This is could also be compounded by node vs browser based timing or using different WebGPU implementations.)

The concern here is that it is unclear that using these values for load balancing decisions would be any better then something like just round-robining down the list of tests to run, since though the cost values are in the same units (mS), there is no way to know if they are comareable.
- Is there a doc or something that discusses this?

The final concern I have heard is about the location of this timing data.
- As mentioned above, it appears to be pretty tightly coupled to the machine and environment that the tests are executed on, so is in the the general CTS tree the proper location? 

- Having the common load balancing logic live in CTS repo seems sensible, but should the specific values be provided by the different implementations of WebGPU as part of their CI, since it is unlikley one set of values would apply to all implementations?

(This could be somewhat mitigated if the values were all calculated by a common baseline machine, so aleast they could be considered be relative costs, but that would not address if different implementations have significantly different runtimes between, i.e. test A is faster than test B on one implementation but reversed on another)

kainino0x:
These can be updated as needed rather than forcing test developers to update them continuously.

Generates a new `gen/webgpu_variant_list.json` and `gen/webgpu_variant_list_chunked2sec.json` for any automated tooling needs. These are not used for anything internally, they're just there for external tools.

**Note: `listing_meta.json` is no longer an exhaustive list of tests. Use `gen/webgpu_variant_list*.json` if you need this.**

Also:
- Deduplicate some docs between merge_listing_times.ts and adding_timing_metadata.md
- Update docs for this change
- Update docs for the change in #3536

Fixes #3522

<hr>

- [N/A] Test behaves as expected in a WebGPU implementation. (If not passing, explain above.)
- [N/A] Test have be tested with compatibility mode validation enabled and behave as expected. (If not passing, explain above.)

- [N/A] Tests are properly located in the test tree.
- [N/A] [Test descriptions](https://github.com/gpuweb/cts/blob/main/docs/intro/plans.md) allow a reader to "read only the test plans and evaluate coverage completeness", and accurately reflect the test code.
- [N/A] Tests provide complete coverage (including validation control cases). **Missing coverage MUST be covered by TODOs.**


kainino0x:
PTAL

src/common/tools/validate.ts
```diff
@@ -14,23 +14,40 @@ For each suite in SUITE_DIRS, validate some properties about the file:
 - That each case query is not too long
 
 Example:
-  tools/validate src/unittests/ src/webgpu/
+  tools/validate src/unittests src/webgpu
 
 Options:
-  --help          Print this message and exit.
+  --help           Print this message and exit.
+  --validate-meta  Validate listing_meta.json files.
```
dneto0, line 21: The arg checked for on line 37 is `--print-metadata-warnings`

dneto0 (changes requested):

src/common/tools/validate.ts
kainino0x: Oops, I renamed it and forgot to fully rename it.

zoddicus (approved):

PR #3549, opened 2024-03-23
wgsl: Validation for pow builtin function

toji:
Posting now, but there's still some open questions about the validation as discussed in http://github.com/gpuweb/gpuweb/issues/4527 The tests pass with Dawn as-is. Happy to hold off if we want to wait till the spec questions are resolved.

<hr>


dj2:
I'm not sure if we should land this until the spec is clarified. @dneto0 opinions?

src/webgpu/shader/validation/expression/call/builtin/pow.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+    if (scalarTypeOf(kValidArgumentTypes[t.params.type]) === Type.f16) {
+      t.selectDeviceOrSkipTestCase('shader-f16');
+    }
+  })
+  .fn(t => {
+    let expectedResult = true;
+
+    const a = Number(t.params.a);
+    const b = Number(t.params.b);
+    if (a < 0 || (a === 0 && b <= 0)) {
```
dneto0, line 68: LGTM to land.

This matches GLSL exactly.

Also, MSL says for both pow and powr:
"Implemented as exp2(y * log2(x)).
Undefined for x = 0 and y = 0"
Because it uses log2(x) , then x < 0 is disallowed.

HLSL doesn't say.

IEEE 754-2018 defines pown, pow, and powr.  
*  pown: exponent is an integer
* pow  exponent y is real, but must agree with pown when y is integral.  Domain is [-inf,+inf]x[-inf,+inf]
* powr is implemented as exp( y * log(x)) so domain is [ 0, +inf ] x [-inf, +inf].
Closest we have in shader languages is powr.

Taking the intersection of GLSL, MSL, and IEEE powr, we end up using GLSL's rule.

So LGTM to land.

dneto0 (approved):
Thanks!

PR #3550, opened 2024-03-23
WGSL validation test for textureGather/compare

greggman:
<hr>


src/webgpu/shader/validation/expression/call/builtin/textureGather.spec.ts
```diff
@@ -0,0 +1,333 @@
...
+    `
+Validates that only non-const components arguments are rejected by ${builtin}
+`
+  )
+  .params(u =>
+    u
+      .combine('textureType', keysOf(kValidTextureGatherParameterTypes))
+      // filter out types with no component argument
+      .filter(t => !!kValidTextureGatherParameterTypes[t.textureType].hasComponentArg)
+      .combine('varType', ['c', 'u', 'l'])
```
alan-baker, line 126: I don't see a declaration for `u` so it will fail validation, but for the wrong reason.

alan-baker: I'd also consider making this an expression (e.g. `c + 1`).

src/webgpu/shader/validation/expression/call/builtin/textureGather.spec.ts
```diff
@@ -0,0 +1,333 @@
...
+  .specURL('https://gpuweb.github.io/gpuweb/wgsl/#texturegather')
+  .desc(
+    `
+Validates that only non-const offset arguments are rejected by ${builtin}
+`
+  )
+  .params(u =>
+    u
+      .combine('textureType', kTextureTypes)
+      .combine('varType', ['c', 'u', 'l'])
```
alan-baker, line 306: Similar here, there seems to be a missing declaration for `u`.

src/webgpu/shader/validation/expression/call/builtin/textureGatherCompare.spec.ts
```diff
@@ -0,0 +1,264 @@
...
+  .specURL('https://gpuweb.github.io/gpuweb/wgsl/#texturegathercompare')
+  .desc(
+    `
+Validates that only non-const offset arguments are rejected by ${builtin}
+`
+  )
+  .params(u =>
+    u
+      .combine('textureType', kTextureTypes)
+      .combine('varType', ['c', 'u', 'l'])
```
alan-baker, line 238: Same here.

alan-baker (changes requested):
I think there are some missing declarations in some tests.

I don't see any negative tests for the texture type parameter. Is that covered elsewhere or todo?

src/webgpu/shader/validation/expression/call/builtin/textureGather.spec.ts
greggman: this one has `u` at line 323

src/webgpu/shader/validation/expression/call/builtin/textureGatherCompare.spec.ts
greggman: this one is defined at 254

https://github.com/greggman/cts/blob/ac09118d5a1ff0a6240aad334ed26c462937a908/src/webgpu/shader/validation/expression/call/builtin/textureGatherCompare.spec.ts#L254

src/webgpu/shader/validation/expression/call/builtin/textureGather.spec.ts
greggman: `u` was missing. Added. Not sure what you mean by `c + 1`.

greggman:
> I don't see any negative tests for the texture type parameter. Is that covered elsewhere or todo?

Do you have an example of a negative test for texture type?  The function's override is selected by texture type.

src/webgpu/shader/validation/expression/call/builtin/textureGather.spec.ts
alan-baker: Making it an expression instead of just a reference to a declaration.

alan-baker (commented):
For texture types I mean a non-sampled texture (e.g. `texture_storage_2d<r32uint>` or `texture_multisampled_2d`). Those should all fail without an appropriate overload.

greggman:
> For texture types I mean a non-sampled texture (e.g. `texture_storage_2d<r32uint>` or `texture_multisampled_2d`). Those should all fail without an appropriate overload.

My point is, since the overload comes from the texture type, what parameters to do I pass to it to try to make it incorrectly compile?

```
@group(0) @binding(0) var t: texture_storage_2d<r32uint>;

...
  _ = textureGather(t, --what-parameters-go-here-?);
```

Since there is no overload for `texture_storage_2d<r32uint>` there is no way to deduce which parameters to pass to try to get it to incorrectly compile. All the other tests pass all valid parameters except for the one they're testing. i can't do that for texture type because I have no idea what parameters that bad type requires

src/webgpu/shader/validation/expression/call/builtin/textureGather.spec.ts
greggman: I'm not following. Why does need to be an expression? The thing being tested is does `c` (a const) work and `u` a uniform and `l` an non-constant, fail. It's testing that. What does `c + 1` change about that?

alan-baker:
I'd suggest just picking a few where you just use the parameters from an equivalent sampled texture. So `texture_storage_2d<r32uint>` -> `texture_2d<u32>` for example. Doesn't need to be exhausted but good to have some coverage in my opinion.

src/webgpu/shader/validation/expression/call/builtin/textureGather.spec.ts
alan-baker: To me, it tests that an expression is acceptable and not just a name.

greggman: constant expressions are already tested by the other test though right?

I think I'm confused. We're testing in other tests which types are acceptable. Expressions evaluate to a type so testing expressions is redundant. Instead, in some other part of the CTS, that each expression evaluates to the correct type would be tested. Then, here, all that's tested is (1) that only the correct types are allowed (that's the previous test to this one) and (2) the extra restrictions (in this case that this parameter must be const).

Otherwise, I'm not sure why would wouldn't have to test every possible expression `c + 1`, `c - 1` (because maybe minus converts to a different type, `someFunc(c)`, `1 - c`, ` + c`, etc, etc.

alan-baker: If they are covered elsewhere that's ok then.

alan-baker (approved):

PR #3565, opened 2024-03-27
Add validation tests for *Bit builtin functions

lokokung:
Note that `insertBits` will come in a follow up because it has multiple arguments unlike these simpler ones.

<hr>


lokokung:
Note that these are also practically identical to the ones in https://github.com/gpuweb/cts/pull/3498 since these builtins all take the same argument types.

src/webgpu/shader/validation/expression/call/builtin/firstLeadingBit.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+Validates that float arguments are rejected by ${builtin}()
+`
+  )
+  .params(u => u.combine('type', keysOf(kFloatTypes)))
+  .fn(t => {
+    const type = kFloatTypes[t.params.type];
+    validateConstOrOverrideBuiltinEval(
+      t,
+      builtin,
+      /* expectedResult */ false,
```
dj2, line 64: Can you add a case into here that will pass, and then conditionalize this? We like to make sure the tests are failing due to the input not other issues.

src/webgpu/shader/validation/expression/call/builtin/firstLeadingBit.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+  good: '(1u)',
+  bad_no_parens: '',
+  // Bad number of args
+  bad_too_few: '()',
+  bad_too_many: '(1u,2u)',
+  // Bad value for arg 0 (Note that float type arguments are handled in 'float_argument' above)
+  bad_0bool: '(false)',
+  bad_0array: '(array(1u))',
+  bad_0struct: '(modf(2.2))',
+};
```
dj2, line 80: Few more to add here, textures, samplers, matrices, vector, atomic

src/webgpu/shader/validation/expression/call/builtin/firstLeadingBit.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+    t.expectCompileResult(
+      t.params.arg === 'good',
+      `const c = ${builtin}${kArgCases[t.params.arg]};`
+    );
+  });
+
+g.test('must_use')
+  .desc(`Result of ${builtin} must be used`)
+  .fn(t => {
+    t.expectCompileResult(false, `fn f() { ${builtin}${kArgCases['good']}; }`);
```
dj2, line 95: Can you conditionalize this and make a version which assigns to a phony to verify it passes?

dj2 (commented):
Comments on the first test apply to the other two test files as well.

src/webgpu/shader/validation/expression/call/builtin/firstLeadingBit.spec.ts
dneto0: We haven't done those types in other similar tests.  So I'd be ok with filing a bug to followup to do this more broadly.

Also, textures and samplers can't be constructed in the middle of an expression, so we'd have to add a preamble to declare those variables.

src/webgpu/shader/validation/expression/call/builtin/firstLeadingBit.spec.ts
dneto0: Agreed, like we've done in others like `sin` and `cos`

src/webgpu/shader/validation/expression/call/builtin/firstLeadingBit.spec.ts
dneto0: Atomic is another case you can't construct in an expresison. you'd need a variable declaration somewhere.

dneto0 (changes requested):
Please update the must_use tests to be like the other ones such as `cos`:

https://github.com/gpuweb/cts/blob/916580420fa306db53e7934c172b1a94f42aba16/src/webgpu/shader/validation/expression/call/builtin/cos.spec.ts#L108 

I'm ok adding a TODO for the other bad type cases that require a declared variable elsewhere: textures, samplers, atomics.

dj2:
You can see an example of some of those tests in https://github.com/gpuweb/cts/blob/c06f419947827ece5335e4593ca466a31166c7d9/src/webgpu/shader/validation/expression/call/builtin/acos.spec.ts#L188

src/webgpu/shader/validation/expression/call/builtin/firstLeadingBit.spec.ts
dj2: These are the tests I've been adding as I worked down the list of builtins. You can see some examples in https://github.com/gpuweb/cts/blob/c06f419947827ece5335e4593ca466a31166c7d9/src/webgpu/shader/validation/expression/call/builtin/acos.spec.ts#L188

lokokung: Done.

src/webgpu/shader/validation/expression/call/builtin/firstLeadingBit.spec.ts
lokokung: Done.

src/webgpu/shader/validation/expression/call/builtin/firstLeadingBit.spec.ts
lokokung: Done.

dneto0 (approved):
Thanks

PR #3567, opened 2024-03-28
wgsl: Add short-circuiting validation tests

jrprice:
Test that the short-circuiting logical expressions are only accepted for scalar boolean types.

Also test that they guard invalid expressions on the right-hand-side when the left-hand-side is a const-expression.

<hr>


src/webgpu/shader/validation/expression/binary/short_circuiting_and_or.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+
+    t.expectCompileResult(t.params.control, code);
+  });
+
+// A list of expressions that are invalid unless guarded by a short-circuiting const-expression.
+const kInvalidRhsExpressions: Record<string, string> = {
+  overflow: 'i32(1<<31) < 0',
+  binary: '(1.0 / 0) == 0',
+  builtin: 'sqrt(-1) == 0',
+  array_size: 'array<bool, 3 - 4>()[0]',
```
jrprice, line 154: This is the contentious case that we've been discussing internally, and likely needs clarification in the spec as to whether this is valid or not on the RHS of a short-circuited expression.

FYI @alan-baker too.

jrprice: I've opened https://github.com/gpuweb/gpuweb/issues/4551 to get clarification on this.

dneto0: We should land this test before the spec update is reviewed (if needed)

src/webgpu/shader/validation/expression/binary/short_circuiting_and_or.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+  .params(u =>
+    u
+      .combine('op', ['&&', '||'])
+      .combine('rhs', keysOf(kInvalidRhsExpressions))
+      .combine('skip_rhs', [true, false])
+      .beginSubcases()
+  )
+  .fn(t => {
+    const lhs =
+      t.params.op === '&&'
```
dneto0, line 172: It took a while to read this but yes, this is correct.

An alternative:
```js
const lookup =  { '&&': { true: false, false: true}, '||': { true: true, false: false}};
const lhs = lookup[t.params.op][t.params.skip_rhs];
```

dneto0 (approved):
Thanks!

src/webgpu/shader/validation/expression/binary/short_circuiting_and_or.spec.ts
jrprice: Did this a slightly different way because TypeScript was being picky.

src/webgpu/shader/validation/expression/binary/short_circuiting_and_or.spec.ts
jrprice: I pulled out the array-element-count tests to their own suite(?) and made them expected-to-fail (with a control case), in anticipation of the spec clarification falling in that direction. No need to rush to land this though.

jrprice (commented):
Also added `override` tests.

dneto0 (approved):
Thanks!

alan-baker:
https://github.com/gpuweb/gpuweb/pull/4801 resolves the spec issue.

PR #3569, opened 2024-03-29
wgsl: Validation for reflect builtin function

toji:
This has the most complex algorithm validation thus far (though it's still not too bad.) So a couple of questions I have regarding that:

 - Is it worthwhile to check every separate calculation? Feels like some of these calculations could be grouped.
 - Similarly, is it worth it to quantize at each step?
 - Finally, we're testing 5 values from `fullRangeForType` because we've got two different values we're checking, but this means we mostly hit the extremes, which won't be where most real-world usage happens. Do we care about including more "normal" test cases for algorithms like this? (Especially relevant for pow, where most exponents will be smallish.)

<hr>


ben-clayton (approved):
LGTM

PR #3572, opened 2024-03-29
Better-preserve log messages sent from workers

kainino0x:
We were relying on postMessage to turn LogMessageWithStack objects into Error objects, but this was kind of magic, and didn't work well - in particular, the name was lost.

Flattening this into a plain object before postMessage makes it much less magic. Now logs sent from workers print correctly, showing for example "DEBUG:" instead of "Error:".

Issue: none

<hr>


src/common/internal/logging/log_message.ts
```diff
@@ -1,26 +1,43 @@
...
 
-  constructor(name: string, ex: Error | ErrorWithExtra) {
-    super(ex.message);
+  /** Wrap an Error (which was created to capture the stack at that point) into a
+   * LogMessageWithStack (which has extra stuff for good log messages).
+   */
+  static wrapError(name: string, ex: Error | ErrorWithExtra) {
+    return new LogMessageWithStack({
+      name,
+      message: ex.message,
```
greggman, line 17: Do you want to send `ex.constructor.name`?

greggman (approved):
lgtm, just random thoughts, no changes requested.

src/common/internal/logging/log_message.ts
kainino0x: `name` is sort of meant to override it. Chrome includes the original name in the `stack`, but other browsers don't so we should preserve it somehow for cases where it matters (I think just `rec.threw()`).

PR #3573, opened 2024-03-29
wgsl: Validation for mix builtin function

toji:
Mostly working at the moment, but some of the f16 values tests are failing due to out-of-range computations. For example:

```
EXCEPTION: Unexpected validation error occurred: Error while parsing WGSL: :2:11 error: '-65504.0 + -0.0039043426513671875' cannot be represented as 'f16'
const v = mix(vec2(-65504.0h, -65504.0h), vec2(63.96875h, 63.96875h), vec2(-0.00006103515625h, -0.00006103515625h));
```

As you can see, the code IS trying to check for that, so there must be something here about how the tests are quantizing to f16 that doesn't quite match how Tint is checking things.

<hr>


zoddicus:
So I think you have gotten hoisted on the same issue I hit when I initially implemented the mix execution tests, https://github.com/gpuweb/gpuweb/issues/3260. Basically the spec gives an example of how mix can be implemented, but does not mandate it, though kinda does in the accuracy interval... some implementations (I think Vulkan based ones) use a different formula, which will give you different intermediates.

The execution tests calculate both common forms for mix and use the combined interval, so you will probably have to do something similar, https://github.com/gpuweb/cts/blob/1779ce8568baba98730ea2322a1328d1c997cbe9/src/webgpu/util/floating_point.ts#L3771.

I am going to ping on the original spec issue, to see if we can get this resolved in the spec.

toji:
Actually, I suspect that this is due to F16 quantization behavior that #3636 should help address. But I won't be able to confirm that till I get home to test on a device with F16 support.

toji:
Making use of the utilities introduced in #3643 has fixed the issues mentioned in the opening comment, so this is now ready for review.

zoddicus (approved):

PR #3577, opened 2024-03-29
wgsl: Validation of fma builtin function

toji:
Currently exhibiting the same f16 validation failures as #3573 

<hr>


ben-clayton:
Adding Ryan as reviewer for this, as this overlaps with the execution tests.

src/webgpu/shader/validation/expression/call/builtin/fma.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+    }
+  })
+  .fn(t => {
+    let expectedResult = true;
+
+    const scalarType = scalarTypeOf(kValidArgumentTypes[t.params.type]);
+    const quantizeFn = quantizeFunctionForScalarType(scalarType);
+
+    // Fused Multiply Add equation: a * b + c
+    // Should be invalid if the calculations result in intermediate values that
```
dneto0, line 67: Actually, WGSL allows a faked fma, which is _non_-fused multiply-add.  And in that case this rule is correct.

But a proper IEEE fma, which is allowed by WGSL, allows cases where the intermediate may overflow.  The "fusing" part says compute the intermediates at infinite precision, then round.    so you can have `A*B+C` result in an in-range value.
Example, A= FLOAT_MAX, B=1.5, C = - FLOAT_MAX.  And the in-range result should approximate `0.5*FLOAT_MAX`.

src/webgpu/shader/validation/expression/call/builtin/fma.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+    // exceed the maximum representable float value for the given type.
+    const a = Number(t.params.a);
+    const b = Number(t.params.b);
+    const c = Number(t.params.c);
+    const ab = quantizeFn(a * b);
+    const abc = quantizeFn(ab + c);
+
+    if (!Number.isFinite(ab) || !Number.isFinite(abc)) {
+      expectedResult = false;
```
dneto0, line 76: I think skip the test in this case, per the description above.
Because either result is ok.

src/webgpu/shader/validation/expression/call/builtin/fma.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+    }
```
dneto0, line 77: but we should still test this:

```
if (!Number.isFinite(quantizeFn( a * b + c)) {expectedResult = false; }
```

dneto0 (changes requested):

src/webgpu/shader/validation/expression/call/builtin/fma.spec.ts
toji: Good to know, thanks!

src/webgpu/shader/validation/expression/call/builtin/fma.spec.ts
toji: Done!

toji:
Thanks for the review, @dneto0. Updated according to your recommendation and added a comment to explain why.

src/webgpu/shader/validation/expression/call/builtin/fma.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+  })
+  .fn(t => {
+    let expectedResult = true;
+
+    const scalarType = scalarTypeOf(kValidArgumentTypes[t.params.type]);
+    const quantizeFn = quantizeFunctionForScalarType(scalarType);
+
+    // Fused Multiply Add equation: a * b + c
+    // Should be invalid if the calculation results in values that
```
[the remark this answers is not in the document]
toji: Got it. Updated this to only fail if the end result is OOB, but if the end result is valid and the intermediate a*b goes OOB it'll now skip the test.

zoddicus (commented):
This looks good to me, will let @dneto0 give final approval.

toji:
Review ping?

zoddicus (approved):

ben-clayton (approved):

toji:
Thank you! Still need @dneto0 to clear his changes requested bit.

src/webgpu/shader/validation/expression/call/builtin/fma.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+} from './const_override_validation.js';
+
+export const g = makeTestGroup(ShaderValidationTest);
+
+const kValidArgumentTypes = objectsToRecord(kConvertableToFloatVectors);
+
+g.test('values')
+  .desc(
+    `
+Validates that constant evaluation and override evaluation of ${builtin}() never errors
```
dneto0, line 26: This is inaccurate?  There are error cases.

dneto0 (approved):
LGTM except the comment on the test is inaccurate.  It is supposed to fail sometimes.

src/webgpu/shader/validation/expression/call/builtin/fma.spec.ts
toji: Fixed, thanks for calling that out!

PR #3578, opened 2024-04-01
WGSL execution TextureSample tests

greggman:
This was originally @ben-clayton's PR. It's from March of 2023 so merged with the latest, got it to run, found f16 and f32 formats failed on the spiral (not texel center) tests on all devices. Tried quantizing the texture coordinates within texels to powers of 2. Found that on M1 and AMD (linux) quantizing to 256 units per texel made the tests pass. NVidia requires quantizing to 16 units per pixel. It still seems like a valid test with this quantization.

There's a lot more to do in this file but I thought it was good to get something checked in.

Note: Once approved, the cache needs rebuilding so don't merge this. I didn't want to add the cache files yet because it's super annoying to have to keep fixing them when merging. After approved I'll generate the cache files, add them to the PR, and merge.

<hr>


greggman:
fyi: The code picks pseudo random values for the texels in the texture. It does this by using effectively doing `lerp(finiteMin, finiteMax, Math.random())` (except Math.random() is some pseudo random function)

In anycase, finiteMin can be an extremely negative value and finiteMax an extremely positive value and I guessed that bilinear interpreting between values with a very large range might have been why the test was failing. So, I added code like this as a test

```js
function getLimit(v: number) {
  switch (v) {
    case Number.POSITIVE_INFINITY:
      return -1;
    case Number.NEGATIVE_INFINITY:
      return 2;
    default:
      return v;
  }
}

export function createRandomTexelView(info: {
  format: GPUTextureFormat;
  size: GPUExtent3D;
}): TexelView {
  const rep = kTexelRepresentationInfo[info.format as EncodableTextureFormat];
  const generator = (coords: Required<GPUOrigin3DDict>): Readonly<PerTexelComponent<number>> => {
    const texel: PerTexelComponent<number> = {};
    for (const component of rep.componentOrder) {
      const rnd = hashU32(coords.x, coords.y, coords.z, component.charCodeAt(0));
      const normalized = clamp(rnd / 0xffffffff, { min: 0, max: 1 });
      // texel[component] = lerp(rep.numericRange!.finiteMin, rep.numericRange!.finiteMax, normalized);
      texel[component] = lerp(
        getLimit(rep.numericRange!.min),
        getLimit(rep.numericRange!.max),
        normalized
      );
    }
    return quantize(texel, rep);
  };
  return TexelView.fromTexelsAsColors(info.format as EncodableTextureFormat, generator);
}
```

This would effectively make it pick values only between -1.0 and 2.0 in floating point textures. It didn't help. What did help was quantizing the texture coordinates.

ben-clayton:
Update: Dry run unfortunately has quite a few failures for these new tests.

greggman:
So I bump the quantizing all the way down to 2 units per texel. That's the lowest it can go and still actually test sampling. That means it's at least testing between 2 texels (could probably make it 4 or 8 but haven't tried those). 

I also had to remove the subcases as it was timing out on android and sometimes other machines. I'm not sure if setting `listing_meta.json` for these tests to some high number would change that. What I've noticed is, on my M1, if not tests fail it runs fast, all cases (as subcases), it's 2-3 seconds in total for all the tests. On the other hand if they start failing it takes much longer. This is in standalone which doesn't have the overhead of the bots.  In any case, just to see it work I made them all cases.

Testing here: https://dawn-review.googlesource.com/c/dawn/+/181924?tab=checks

It passes everywhere except android and linux

On android it's failing on `NEAREST` filter with texel center coordinates. That's a bug in that driver or implementation since it's passing everywhere else. 

Example:
https://ci.chromium.org/ui/labs/p/chromium/inv/build-8751721781070413873/test/ninja%3A%2F%2Fchrome%2Ftest%3Atelemetry_gpu_integration_test_android_chrome%2Fgpu_tests.webgpu_cts_integration_test.WebGpuCtsIntegrationTest.webgpu%3Ashader%2Cexecution%2Cexpression%2Ccall%2Cbuiltin%2CtextureSample%3Asampled_2d_coords%3Aformat%3D%22rgba8unorm%22%3Bsample_points%3D%22texel-centre%22%3BaddressModeU%3D%22mirror-repeat%22%3BaddressModeV%3D%22repeat%22%3BminFilter%3D%22linear%22%3Boffset%3Dtrue/variant/33151f17ffa75e34?resultIndex=0

On Linux all the failing tests are crashing. That also seems to suggest a bug in the implementation, not in the test since they're passing everywhere else including my on Linux in both Vulkan and OpenGL (compat)

https://ci.chromium.org/ui/p/chromium/builders/try/linux-dawn-rel/b8751721782341272257/test-results?q=ExactID%3Aninja%3A%2F%2Fchrome%2Ftest%3Atelemetry_gpu_integration_test%2Fgpu_tests.webgpu_cts_integration_test.WebGpuCtsIntegrationTest.webgpu%3Ashader%2Cexecution%2Cexpression%2Ccall%2Cbuiltin%2CtextureSample%3Asampled_2d_coords%3Aformat%3D%22bgra8unorm%22%3Bsample_points%3D%22texel-centre%22%3BaddressModeU%3D%22mirror-repeat%22%3BaddressModeV%3D%22clamp-to-edge%22%3BminFilter%3D%22nearest%22%3Boffset%3Dtrue+VHash%3A087f2e276273f1a0&clean=&sortby=&groupby=

That seems like the tests are good to check in and bugs should be filed for Android and Linux

greggman:
Actually, with the spiral coord pattern and `nearest` filtering and a quantization of 2 units per texel, some coords may be right on the edge between 2 texel. Maybe I should make it so if it's `spiral` coords and it's NEARST then never choose a value at the edge between 2 texels.

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
```diff
@@ -439,9 +441,20 @@ export function generateSamplePoints(
...
-  const kSubdivisionsPerTexel = 2;
+  const kSubdivisionsPerTexel = 4;
   const q = [args.textureWidth * kSubdivisionsPerTexel, args.textureHeight * kSubdivisionsPerTexel];
-  return out.map(c => c.map((v, i) => Math.floor(v * q[i]) / q[i]) as vec2);
+  return out.map(
+    c =>
+      c.map((v, i) => {
+        // Quantize to kSubdivisionsPerPixel
+        const v1 = Math.floor(v * q[i]);
+        // If it's nearest and we're on the edge of a texel then move us off the edge
```
ben-clayton, line 451: I actually implemented something similar to deal with 'bumping' edge texels - apparently I reverted it, or have a local branch somewhere I never pushed.

Please correct me if I'm wrong, but I think what you have here will bump the texels between `[0, ε)`, but you have to consider both sides of the edge (`[-ε, ε]`). IIRC my approach was something like:

```ts
// Bump all texels that sit on a texel boundary, so that they're deterministically sampling.
out = out.map(c => c.map((v, i) => {
  const dims = [args.textureWidth, args.textureHeight];
  const t = v * dims[i]; // normalized -> integer texels
  const onEdge = Math.abs(fract(t + 0.5) - 0.5) < epsilon;
  if (onEdge) {
    t += bumpFactor;
  }
  return t / dims[i];
})
```

greggman: The code as is, works

```
nearest = true;
args = { textureWidth: 8, textureHeight: 8};
kSubdivisionsPerTexel = 4;
q = [args.textureWidth * kSubdivisionsPerTexel, args.textureHeight * kSubdivisionsPerTexel];
bumpFromTexelBoundry = (v, i) => {
      // Quantize to kSubdivisionsPerPixel
      const v1 = Math.floor(v * q[0]);
      // If it's nearest and we're on the edge of a texel then move us off the edge
      // since the edge could choose one texel or another in nearest mode
      const v2 = (nearest && v1 % kSubdivisionsPerTexel === 0) ? v1 + 1 : v1;
      // Convert back to texture coords
      return v2 / q[0];
}
toTexel = v => v * args.textureWidth;
toCoord = v => v / args.textureWidth;
[-2, -1.5, -1, -0.5, 0, 0.5, 1, 1.5, 2].map(toCoord).map(bumpFromTexelBoundry).map(toTexel)
```

> `(9) [-1.75, -1.5, -0.75, -0.5, 0.25, 0.5, 1.25, 1.5, 2.25]`

Unfortunately it didn't help get the tests to pass on android

src/webgpu/shader/execution/expression/call/builtin/textureSample.spec.ts
```diff
@@ -70,11 +81,74 @@ Parameters:
...
+      method: t.params.sample_points,
+      textureWidth: 8,
+      textureHeight: 8,
+    }).map((c, i) => {
+      const hash = hashU32(i) & 0xff;
+      return {
+        builtin: 'textureSample',
+        coordType: 'f',
+        coords: c,
+        offset: t.params.offset ? [(hash & 15) - 8, ((hash >> 8) & 15) - 8] : undefined,
```
kainino0x, line 125: `hash >> 8` would always be 0, I think.
```suggestion
        offset: t.params.offset ? [(hash & 15) - 8, ((hash >> 4) & 15) - 8] : undefined,
```

kainino0x (commented):
A few notes, I haven't reviewed the big file yet

kainino0x:
> I'm not sure if setting `listing_meta.json` for these tests to some high number would change that.

It would not, Chrome doesn't use it.

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
kainino0x: I believe it works because it's quantized first. Basically you're just mapping
`[0, 1, 2, 3, 4, 5, 6, 7, ...]` to
`[1, 1, 2, 3, 5, 5, 6, 7, ...]`

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
```diff
@@ -0,0 +1,747 @@
...
+  const prototype = calls[0];
+  if (prototype.builtin.startsWith('textureSample')) {
+    // textureSample*() builtins take a sampler as the second argument
+    args.push('S');
+  }
+
+  for (const name of kTextureCallArgNames) {
+    const value = prototype[name];
+    if (value !== undefined) {
+      if (name === 'offset') {
```
kainino0x, line 513:
```suggestion
      if (name === 'offset') {
        // offset must be a const-expression.
```

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
```diff
@@ -0,0 +1,747 @@
...
+  ddy?: T;
+  offset?: T;
+}
+
+export interface TextureCall<T extends Dimensionality> extends TextureCallArgs<T> {
+  builtin: 'textureSample' | 'textureLoad';
+  coordType: 'f';
+}
+
+function toArray(coords: Dimensionality): number[] {
```
kainino0x, line 61: nit: possibly just require callers to use `[number]` instead of `number` for scalars, simplifying this?

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
```diff
@@ -0,0 +1,747 @@
...
+        errs.push(layoutTwoColumns(expectedSamplePoints, gotSamplePoints).join('\n'));
+        errs.push('', '');
+      }
+    }
+  }
+
+  return errs.length > 0 ? new Error(errs.join('\n')) : undefined;
+}
+
+export async function identifySamplePoints(
```
kainino0x, line 261: Doc comments on functions (at least exported ones) would be very helpful.

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
```diff
@@ -0,0 +1,747 @@
+/* eslint-disable prettier/prettier */
```
kainino0x, line 1: There doesn't seem to be anything in this file that needs prettier disabled.

If there is, please use `/* prettier-ignore */` where needed instead of globally disabling prettier.
`// prettier-ignore` or `/* prettier-ignore */` applies to the entire AST subtree that immediately follows it.

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
kainino0x, line 261: For this function in particular, also please add a comment somewhere in the middle so it's easier to tell where the "interesting" part finishes and the string construction starts. Or consider splitting the function in two.

Also maybe put sample output in a comment so the reader has a picture of what the code is attempting to do.

```ts
// component was not as expected:
//       call: textureSample(texture: T, sampler: S, coords: vec2(1.125, 0.21875))
//  component: R
//        got: 0.4960784316062927
//   expected: 0.496078431372549
//   abs diff: 0.0000
//   rel diff: 0.00%
//   ulp diff: 1
//   sample points:
// 
// expected:                            | got:
//     0   1   2   3   4   5   6   7    |     0   1   2   3   4   5   6   7 
//   ┌───┬───┬───┬───┬───┬───┬───┬───┐  |   ┌───┬───┬───┬───┬───┬───┬───┬───┐
// 0 │   │   │   │   │   │   │   │   │  | 0 │   │   │   │   │   │   │   │   │
//   ├───┼───┼───┼───┼───┼───┼───┼───┤  |   ├───┼───┼───┼───┼───┼───┼───┼───┤
// 1 │   │   │   │   │   │   │   │ a │  | 1 │   │   │   │   │   │   │   │ a │
//   ├───┼───┼───┼───┼───┼───┼───┼───┤  |   ├───┼───┼───┼───┼───┼───┼───┼───┤
// 2 │   │   │   │   │   │   │   │ b │  | 2 │   │   │   │   │   │   │   │ b │
//   ├───┼───┼───┼───┼───┼───┼───┼───┤  |   ├───┼───┼───┼───┼───┼───┼───┼───┤
// 3 │   │   │   │   │   │   │   │   │  | 3 │   │   │   │   │   │   │   │   │
//   ├───┼───┼───┼───┼───┼───┼───┼───┤  |   ├───┼───┼───┼───┼───┼───┼───┼───┤
// 4 │   │   │   │   │   │   │   │   │  | 4 │   │   │   │   │   │   │   │   │
//   ├───┼───┼───┼───┼───┼───┼───┼───┤  |   ├───┼───┼───┼───┼───┼───┼───┼───┤
// 5 │   │   │   │   │   │   │   │   │  | 5 │   │   │   │   │   │   │   │   │
//   ├───┼───┼───┼───┼───┼───┼───┼───┤  |   ├───┼───┼───┼───┼───┼───┼───┼───┤
// 6 │   │   │   │   │   │   │   │   │  | 6 │   │   │   │   │   │   │   │   │
//   ├───┼───┼───┼───┼───┼───┼───┼───┤  |   ├───┼───┼───┼───┼───┼───┼───┼───┤
// 7 │   │   │   │   │   │   │   │   │  | 7 │   │   │   │   │   │   │   │   │
//   └───┴───┴───┴───┴───┴───┴───┴───┘  |   └───┴───┴───┴───┴───┴───┴───┴───┘
// a: at: [7, 1], weights: [R: 0.75000] | a: at: [7, 1], weights: [R: 0.75000]
// b: at: [7, 2], weights: [R: 0.25000] | b: at: [7, 2], weights: [R: 0.25000]
```

kainino0x (commented):
No more big comments on the big file. I didn't get too deep into the math, though.

src/webgpu/util/texture/texel_data.ts
```diff
@@ -809,7 +817,12 @@ export const kTexelRepresentationInfo: {
         G: floatBitsToNormalULPFromZero(components.G!, kFloat11Format),
         B: floatBitsToNormalULPFromZero(components.B!, kFloat10Format),
       }),
-      numericRange: { min: 0, max: Number.POSITIVE_INFINITY },
+      numericRange: {
+        min: 0,
+        max: Number.POSITIVE_INFINITY,
+        finiteMin: 0,
+        finiteMax: float32ToFloatBits(0b11110_111111, 0, 5, 6, 15),
```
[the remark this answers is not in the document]
greggman: To be honest, I've been curious if we should be choosing random numbers between finiteMin and finiteMax, especially for f16, f32, and other floating point values. If the range between the chosen numbers is large I'd expect the differences in interpolation between GPUs to get exaggerated.

On the other hand I tried using just `Math.max(min, -1)` and `Math.min(max, 1)` and it didn't seems to help so I guess my intuition is wrong.

src/webgpu/util/texture/texel_data.ts
```diff
@@ -854,7 +867,12 @@ export const kTexelRepresentationInfo: {
         G: floatBitsToNormalULPFromZero(components.G!, kUFloat9e5Format),
         B: floatBitsToNormalULPFromZero(components.B!, kUFloat9e5Format),
       }),
-      numericRange: { min: 0, max: Number.POSITIVE_INFINITY },
+      numericRange: {
+        min: 0,
+        max: Number.POSITIVE_INFINITY,
+        finiteMin: 0,
+        finiteMax: float32ToFloatBits(0b11110_111111111, 0, 5, 9, 15),
```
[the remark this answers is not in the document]
greggman: done

src/webgpu/util/math.ts
```diff
@@ -413,6 +413,31 @@ export function oneULPF32(target: number, mode: FlushMode = 'flush'): number {
   }
 }
 
+/** Converts two 32-bit hex values to a 64-bit float value */
+export function hexToF64(h32: number, l32: number): number {
+  const u32Arr = new Uint32Array(2);
```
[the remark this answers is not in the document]
greggman: just deleted it

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
greggman: done. though for this one in particular I just didn't export it.

greggman: done

kainino0x:
(I'm pretty sure this is not quite ready to land? But please ping me if/when you need approval from me)

src/webgpu/shader/execution/expression/call/builtin/textureSample.spec.ts
greggman: done

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
```diff
@@ -0,0 +1,747 @@
...
+      // ┌───┬───┬───┬───┐
+      // │ a │   │   │   │  norm: a = 1/8, b = 7/8
+      // ├───┼───┼───┼───┤   abs: a = 0,   b = 3
+      // │   │   │   │   │
+      // ├───┼───┼───┼───┤
+      // │   │   │   │   │
+      // ├───┼───┼───┼───┤
+      // │   │   │   │ b │
+      // └───┴───┴───┴───┘
+      let at = coords.map((v, i) => (v - 0.5 / textureSize[i]) * textureSize[i]);
```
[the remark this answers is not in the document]
greggman: done

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
greggman: I don't think it's that simple to change it. It used for all args, many of which are more natural to think of as `number` instead of `[number]`. I spent a couple of hours trying to get it work and failed. Maybe we can do this later if it's important?

greggman:
@kainino0x this is ready 

There are runs [here](https://dawn-review.googlesource.com/c/dawn/+/182983). AFAIK the only failures are real. Linux-Nvidia crashes. Android gets truly bad results (nearest sampling fails)

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
kainino0x: It's not important, just thought it would remove some of the special casing.

src/webgpu/util/math.ts
```diff
@@ -413,6 +413,31 @@ export function oneULPF32(target: number, mode: FlushMode = 'flush'): number {
...
+  u32Arr[0] = l32;
+  u32Arr[1] = h32;
+  const f64Arr = new Float64Array(u32Arr.buffer);
+  return f64Arr[0];
+}
+
+/** @returns an integer value between 0..0xffffffff using a simple non-cryptographic hash function */
```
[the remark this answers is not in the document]
kainino0x: Comment still should be added

kainino0x (approved):

PR #3579, opened 2024-04-01
wgsl: Validation for dot builtin function

toji:
Because of the `dot` function's ability to accept abstract-int values I needed to add some additional scaffolding to this one to make it more tractable to write the kind of step-by-step equation validation we've been doing so far. Not 100% sure if I'm happy with it, but whatever we do land on here I'll probably want to push to a common util and update the previous builtin validations that I've done to use.

<hr>


src/webgpu/shader/validation/expression/call/builtin/dot.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+  stageSupportsType,
+  validateConstOrOverrideBuiltinEval,
+} from './const_override_validation.js';
+
+export const g = makeTestGroup(ShaderValidationTest);
+
+const kValidArgumentTypes = objectsToRecord(kConvertableToFloatVectors);
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function quantizeFunctionForScalarType(type: ScalarType): QuantizeFunc<any> {
```
ben-clayton, line 30: Do we need to use `any` here? Could we not use `QuantizeFunc<number | bigint>` or something?

src/webgpu/shader/validation/expression/call/builtin/dot.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+    case Type.abstractInt:
+      return BigInt;
+    default:
+      return Number;
+  }
+}
+
+export interface CheckFunc {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  (num: any): boolean;
```
ben-clayton, line 45: ditto, here any everywhere else `any` is used

src/webgpu/shader/validation/expression/call/builtin/dot.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+  )
+  .beforeAllSubcases(t => {
+    if (scalarTypeOf(kValidArgumentTypes[t.params.type]) === Type.f16) {
+      t.selectDeviceOrSkipTestCase('shader-f16');
+    }
+  })
+  .fn(t => {
+    const scalarType = scalarTypeOf(kValidArgumentTypes[t.params.type]);
+
+    // Dot product equation: (a[0]*b[0]) + (a[1]*b[1]) + ... (a[N]*b[N])
```
ben-clayton, line 112: Because we use a single scalar replicated for all elements of the vector (splat), order of the addition doesn't need to be considered. Cool.

src/webgpu/shader/validation/expression/call/builtin/dot.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+    case Type.abstractInt:
+      return (num: bigint) => !kValue.i64.isOOB(num);
+    default:
+      return (num: number) => Number.isFinite(num);
+  }
+}
+
+interface StepArgs {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  [name: string]: any;
```
ben-clayton, line 59: The `checkEquation` stuff is cool, but I am wondering if it will have use beyond `dot` tests. If not, it might be easier to understand if you just did the explicit checks without the abstractions. It would also remove the need for `any` here.

src/webgpu/shader/validation/expression/call/builtin/dot.spec.ts
toji: Yeah, I struggled with this for a while. The problem is that if you say it's a `number | bigint` then every math operation you do throws up TypeScript errors saying that it can't perform math operations between those types. I do pass every incoming argument through the quantize function (which also serves as a cast when needed) to ensure that we're always using the same data types between args, but the TypeScript linter doesn't have a way of recognizing that.

`any` isn't desirable, but at least it keeps you from needing to do lots of redundant checks and casts for every step, and I figure it would be best if that type of type fudging was kept to a utility function.

(I should at least add comments about all this.)

src/webgpu/shader/validation/expression/call/builtin/dot.spec.ts
toji: It could be used for most of the builtin tests I've done so far (normalize, cross, fma, etc) And if I removed the abstractions I'd still have to account for the differences between BigInt and number, which would probably lead to nearly as much supporting code. 

That said, I'm not against approaching it from a different direction as long as we can find a pattern that can keep the equations being tested reasonably clear and separate from the supporting cast/quantize/check code.

src/webgpu/shader/validation/expression/call/builtin/dot.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+  (args: StepArgs): StepArgs;
+}
+
+function checkEquation(type: ScalarType, args: StepArgs, steps: EquationStep[]) {
+  const quantizeFn = quantizeFunctionForScalarType(type);
+  const checkFn = checkFunctionForScalarType(type);
+
+  const quantizedArgs: StepArgs = {};
+  for (const key in args) {
+    quantizedArgs[key] = quantizeFn(args[key]);
```
ben-clayton, line 72: This is converting `vecSize` to the scalar type, which is very subtle.

I had a play to see if I could restructure things in a way that avoided `any`, and kept things a bit more type safe. What do you think about this:

```ts

/** Quantizes and validates if the value is representable by the scalar type. */
function quantizeAndCheck<N extends number | bigint>(
  value: N,
  type: ScalarType
): {
  // 'value' quantized to the scalar type.
  quantized: N;
  // true iff the value is representable by the scalar type without error.
  isValid: boolean;
} {
  switch (type) {
    case Type.f32: {
      assert(typeof value === 'number');
      const quantized = quantizeToF32(value) as N;
      return { quantized, isValid: Number.isFinite(quantized) };
    }
    case Type.f16: {
      assert(typeof value === 'number');
      const quantized = quantizeToF16(value) as N;
      return { quantized, isValid: Number.isFinite(quantized) };
    }
    case Type.abstractInt:
      assert(typeof value === 'bigint');
      return { quantized: value, isValid: !kValue.i64.isOOB(value) };
    default:
      assert(typeof value === 'number');
      return { quantized: value, isValid: Number.isFinite(value) };
  }
}

/** Multiplies the two values, coercing to number if the types are not the same */
function mul(a: number | bigint, b: number | bigint): number | bigint {
  if (typeof a === typeof b) {
    return (a as number) * (b as number); // also works for bigint
  }
  // Type mismatch. Coerce to number.
  return Number(a) * Number(b);
}

class Equation<VALUES extends { [key: string]: number | bigint }> {
  /** The dictionary of values */
  readonly values: VALUES;
  /** The scalar type used for equation calculations */
  readonly type: ScalarType;
  /** True if all computed values are valid for type */
  readonly isValid: boolean;

  constructor(type: ScalarType, data: VALUES, isValid?: boolean) {
    this.type = type;
    this.values = data;
    this.isValid = isValid ?? true;
  }

  /**
   * @returns a new Equation extended from this Equation, with an additional field with the name
   * 'name' and value calculated from calling 'fn'. The returned Equation.isValid will be true iff
   * 'this.isValid' is true and the returned value by 'fn' is valid for 'this.type'.
   */
  then<T extends number | bigint, N extends string>(
    name: N,
    fn: (data: VALUES) => T
  ): Equation<VALUES & { [name in N]: T }> {
    const result = fn(this.values);
    const { quantized, isValid } = quantizeAndCheck(result, this.type);
    const out = { ...this.values, [name]: quantized };
    return new Equation(this.type, out, isValid && this.isValid);
  }
}

g.test('values')
  .desc(
    `
Validates that constant evaluation and override evaluation of ${builtin}() never errors
`
  )
  .params(u =>
    u
      .combine('stage', kConstantAndOverrideStages)
      .combine('type', keysOf(kValidArgumentTypes))
      .filter(u => stageSupportsType(u.stage, kValidArgumentTypes[u.type]))
      .beginSubcases()
      .expand('a', u => fullRangeForType(kValidArgumentTypes[u.type], 5))
      .expand('b', u => fullRangeForType(kValidArgumentTypes[u.type], 5))
  )
  .beforeAllSubcases(t => {
    if (scalarTypeOf(kValidArgumentTypes[t.params.type]) === Type.f16) {
      t.selectDeviceOrSkipTestCase('shader-f16');
    }
  })
  .fn(t => {
    const scalarType = scalarTypeOf(kValidArgumentTypes[t.params.type]);

    // Dot product equation: (a[0]*b[0]) + (a[1]*b[1]) + ... (a[N]*b[N])
    // Should be invalid if the dot product calculations result in intermediate
    // values that exceed the maximum representable float value for the given type.

    const expectedResult = new Equation(scalarType, {
      a: t.params.a,
      b: t.params.b,
      vecSize: BigInt(kValidArgumentTypes[t.params.type].width),
    })
      .then('ab', v => mul(v.a, v.b))
      .then('dp', v => mul(v.ab, v.vecSize));

    const type = kValidArgumentTypes[t.params.type];

    // Validates dot(vecN(a), vecN(b));
    validateConstOrOverrideBuiltinEval(
      t,
      builtin,
      expectedResult.isValid,
      [type.create(t.params.a), type.create(t.params.b)],
      t.params.stage
    );
  });
```

toji:
Okay, refactored this to get rid of the more specialized `CheckedEquation` utility and use the `ConstantOrOverrideValueChecker` utility that I've been using for other tests. In order to handle the BigInt/Number cases I've simply split it into an conditional and repeated the equation, which is annoying but avoids a the `any` shenanigans and in this case it's not bad because the equation is short.

PTAL!

src/webgpu/shader/validation/expression/call/builtin/dot.spec.ts
toji: I appreciate this! Seems viable, but after walking away for a bit and revisiting this I think something more straightforward is probably more prudent here, so now I'm using a simpler utility I put in place for other tests and just doing a conditional, repeating the (short) equation for both bigint and number.

toji:
Review ping?

ben-clayton (approved):
LGTM. Thanks for all the changes here!

PR #3585, opened 2024-04-03
wgsl: Validation for distance builtin function

toji:
Pretty straightforward. The fact that Tint uses `abs(a-b)` internally for validation of scalars tripped me up a bit at first. It's a sensible optimization, but makes me question how much we should be depending on every implementation to make the same optimization choices.

<hr>


src/webgpu/shader/validation/expression/call/builtin/distance.spec.ts
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    // Only calculates the full length if the type is a vector. Otherwise abs(a-b) is used.
+    if (kValidArgumentTypes[t.params.type].width > 1) {
+      const ab2 = quantizeFn(ab * ab);
+      const sqrLen = quantizeFn(ab2 * kValidArgumentTypes[t.params.type].width);
+      // Square root does not need to be calculated because it can never fail if
+      // the previous results are finite.
+
+      if (!Number.isFinite(ab2) || !Number.isFinite(sqrLen)) {
```
ben-clayton, line 83: What about DBZ when the length is zero?

toji: Given that this isn't `normalize`, the input vector doesn't need to be divided by the length, the result is the length itself. And square root is fine because `sqrt(0) == 0`. Unless I'm missing something?

ben-clayton: Sorry, getting my builtins mixed up. Ignore me.

ben-clayton (approved):

PR #3604, opened 2024-04-05
Fail tests in worker=* mode, when worker can't be created

### LINKED ISSUE #3591: Skip or fail tests if the worker= mode is not supported
If the environment doesn't support a worker type (for example SharedWorker isn't supported on Chrome on Android) then tests should all skip or fail if we attempt to run them under that worker type. Currently they hang instead - most likely because of an exception being thrown early before we set up exception handling, but I haven't checked yet.

Offhand I don't know how to do this (can't do it _inside_ the test, because that doesn't run until after the shared worker is created).

I'm also not sure if the tests should fail or skip. Could skip because the test relies on a feature that isn't available. Could fail because `worker=shared` specified a config that isn't available. I'm leaning "skip".

cc @beaufortfrancois

kainino0x:
Rather than just throwing an error and letting the runtime break down when this happens (which in Chromium's testing causes timeouts).

I decided to fail instead of skip for now because probably nothing should actually rely on the skip - and failing loudly is less error-prone.

Issue: fixes #3591

<hr>

- [N/A] Test have be tested with compatibility mode validation enabled and behave as expected. (If not passing, explain above.)

- [N/A] Tests are properly located in the test tree.
- [N/A] [Test descriptions](https://github.com/gpuweb/cts/blob/main/docs/intro/plans.md) allow a reader to "read only the test plans and evaluate coverage completeness", and accurately reflect the test code.
- [N/A] Tests provide complete coverage (including validation control cases). **Missing coverage MUST be covered by TODOs.**


greggman (approved):
lgtm

Issue #3350, opened 2024-01-30
Test CopyEI2T with full transparent source


shaoboyan:
A chromium bug https://bugs.chromium.org/p/chromium/issues/detail?id=1514809 here reports a `ImageElement` uploading issue. I think we should cover this in cts but not sure how to achieve this. 

@kainino0x  and @greggman  for more thoughts.

kainino0x:
Tests for HTMLImageElement added in #3621.
We still should add tests from other sources (see https://github.com/gpuweb/cts/pull/3621#discussion_r1558354041).

PR #3621, opened 2024-04-09
Add cts to cover copyEI2T copying from fully transparent image

shaoboyan:
Issue: #

<hr>

### DIFF
--- a/src/webgpu/web_platform/copyToTexture/image.spec.ts
+++ b/src/webgpu/web_platform/copyToTexture/image.spec.ts
@@ -146,6 +146,79 @@ g.test('from_image')
     );
   });
 
+g.test('from_fully_transparent_image')
+  .desc(
+    `
+  Test HTMLImageElement with alpha 0 can be copied to WebGPU texture correctly.
+
+  Use a prebaked 2x2 fully transparent image as source.
+
+  Then call copyExternalImageToTexture() to do a copy to the 0 mipLevel of dst texture,
+  and read the contents out to compare with the HTMLImageElement contents.
+  When dest alpha mode is:
+  - premultiplied, the content should be (0, 0, 0, 0)
+  - not premultiplied, the content should be the same as prebaked
+    pixel values (255, 102, 153, 0).
+
+  The tests covers:
+  - Source HTMLImageElement is fully transparent with valid dest alphaMode.
+  And the expected results are all passed.
+  `
+  )
+  .params(u => u.combine('dstPremultiplied', [true, false]))
+  .beforeAllSubcases(t => {
+    if (typeof HTMLImageElement === 'undefined') t.skip('HTMLImageElement not available');
+  })
+  .fn(async t => {
+    const { dstPremultiplied } = t.params;
+
+    const kColorFormat = 'rgba8unorm';
+    const kImageWidth = 2;
+    const kImageHeight = 2;
+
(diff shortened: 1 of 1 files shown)


shaoboyan:
This cts tend to cover issues like https://issues.chromium.org/issues/41487362

src/webgpu/web_platform/copyToTexture/image.spec.ts
```diff
@@ -146,6 +146,79 @@ g.test('from_image')
...
+    const kColorFormat = 'rgba8unorm';
+    const kImageWidth = 2;
+    const kImageHeight = 2;
+
+    const imageCanvas = document.createElement('canvas');
+    imageCanvas.width = kImageWidth;
+    imageCanvas.height = kImageHeight;
+
+    // Prebaked fully transparent image with content (255, 102, 153, 0)
+    const image = new Image(kImageWidth, kImageHeight);
```
kainino0x, line 184: We should probably test this for other inputs type other than `Image` (`HTMLImageElement`) as well. Not sure if the Chromium bug was specific to HTMLImageElement, but it's useful coverage anyway.

Could you file a bug for it?

(Maybe we can find a way to factor these tests so we don't have to duplicate it too much. For example this test could easily have a parameter for whether we go through HTMLImageElement or ImageBitmap, or could even be generalized to many input types, since the input data is trivial.)

kainino0x, line 184: Just wondering, do we have tests for partially-transparent images as well? I don't see any here.

If not, perhaps file a bug for this as well.

kainino0x (approved):

src/webgpu/web_platform/copyToTexture/image.spec.ts
shaoboyan: >Not sure if the Chromium bug was specific to HTMLImageElement, but it's useful coverage anyway.

I think the chromium bug is specific to HTMLImageElement. The root cause for that bug is the decoder decodes image with `premultiplied-alpha` and `CopyEI2T` loses the info. So we need a redecoded path for this. I think other inputs (maybe `HTMLVideoElement` ?) doesn't have such problem?

shaoboyan: No, I don't think we have partially-transparent image test coverage(And only ImageBitmap has partially-transparent coverage). File issue 3668

kainino0x: Still, it would probably be relatively easy for a bug to creep in the future, or in another browser's implementation, so more coverage would be useful.

kainino0x: (Please feel free to land this PR though.)

shaoboyan: Thanks and I've got time to back to this PR recently.  Let me file an issue for other inputs in case we don't forget it.

kainino0x: Thanks! Can you link to the issue?

kainino0x: Actually #3350 is still open so I'll just write this down there (if you haven't already filed another issue)

PR #3612, opened 2024-04-08
Validate pack2x16float

Jiawei-Shao:
Fixed: #3610 

<hr>

**Requirements for PR author:**

- [*] All missing test coverage is tracked with "TODO" or `.unimplemented()`.
- [*] New helpers are `/** documented */` and new helper files are found in `helper_index.txt`.
- [*] Test behaves as expected in a WebGPU implementation. (If not passing, explain above.)
- [*] Test have be tested with compatibility mode validation enabled and behave as expected. (If not passing, explain above.)

**Requirements for [reviewer sign-off](https://github.com/gpuweb/cts/blob/main/docs/reviews.md):**

- [ ] Tests are properly located in the test tree.
- [ ] [Test descriptions](https://github.com/gpuweb/cts/blob/main/docs/intro/plans.md) allow a reader to "read only the test plans and evaluate coverage completeness", and accurately reflect the test code.
- [ ] Tests provide complete coverage (including validation control cases). **Missing coverage MUST be covered by TODOs.**
- [ ] Helpers and types promote readability and maintainability.

When landing this PR, be sure to make any necessary issue status updates.

PR #3626, opened 2024-04-09
wgsl: Validation for refract builtin function

toji:
Longest equation that I've validated thus far, but it works.

<hr>


src/webgpu/shader/validation/expression/call/builtin/refract.spec.ts
```diff
@@ -0,0 +1,185 @@
...
+      return quantizeToF16;
+    default:
+      return (v: number) => v;
+  }
+}
+
+g.test('values')
+  .desc(
+    `
+Validates that constant evaluation and override evaluation of ${builtin}() never errors
```
dneto0, line 42: This comment is out of date.

src/webgpu/shader/validation/expression/call/builtin/refract.spec.ts
```diff
@@ -0,0 +1,185 @@
...
+
+    const type = kValidArgumentTypes[t.params.type];
+    const scalarType = scalarTypeOf(kValidArgumentTypes[t.params.type]);
+    const quantizeFn = quantizeFunctionForScalarType(scalarType);
+
+    // Refract equation:
+    //   let k = 1.0 - c * c * (1.0 - dot(b, a) * dot(b, a))
+    //   if (k < 0.0) { return vecN(0.0); }
+    //   return c * a - (c * dot(b, a) + sqrt(k)) * b
+    // Should be invalid if the calculations result in intermediate values that
```
dneto0, line 71: This *might* fall into the "may fail" case gray area.
I recommend adding a comment that we might have to choose sample values more selectively in the future.

src/webgpu/shader/validation/expression/call/builtin/refract.spec.ts
```diff
@@ -0,0 +1,185 @@
...
+  bad_2array: '(vec3(0), vec3(1), array(1.1,2.2))',
+  bad_2struct: '(vec3(0), vec3(1), modf(2.2))',
+  bad_2int: '(vec3(0), vec3(1), 2i)',
+  bad_2uint: '(vec3(0), vec3(1), 2u)',
+  bad_2veci: '(vec3(0), vec3(1), vec3i(2))',
+  bad_2vecu: '(vec3(0), vec3(1), vec3u(2))',
+  bad_2vecf: '(vec3(0), vec3(1), vec3f(2))',
+  bad_2vech: '(vec3(0), vec3(1), vec3h(2))',
+};
+
```
dneto0, line 168: Now we're adding 'return' variants that try to assign the result to an explicitly-typed declaration like `let c: vec2f = ...`.

e.g. see https://github.com/gpuweb/cts/pull/3612/files#diff-fe62cb0fc68ba76c7b93bf333a9a9542cdb242f14292bab49db741f6b4f72409R46

dneto0 (changes requested):

zoddicus (changes requested):
This should be migrated to use your new utility code.

src/webgpu/shader/validation/expression/call/builtin/refract.spec.ts
toji: Done

src/webgpu/shader/validation/expression/call/builtin/refract.spec.ts
toji: Done

src/webgpu/shader/validation/expression/call/builtin/refract.spec.ts
toji: The utility that I just converted this to use should detect at least the ambiguous overflow cases. If there's other ambiguous pitfalls to look out for I'd like to know about it!

toji:
> This should be migrated to use your new utility code.

Done

src/webgpu/shader/validation/expression/call/builtin/refract.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+    const b_dot_a_2 = vCheck.checkedResult(b_dot_a * b_dot_a);
+    const one_minus_b_dot_a_2 = vCheck.checkedResult(1.0 - b_dot_a_2);
+    const c2 = vCheck.checkedResult(c * c);
+    const c2_one_minus_b_dot_a_2 = vCheck.checkedResult(c2 * one_minus_b_dot_a_2);
+    const k = vCheck.checkedResult(1.0 - c2_one_minus_b_dot_a_2);
+
+    if (k >= 0) {
+      const ca = vCheck.checkedResult(c * a);
+      const cbda = vCheck.checkedResult(c * b_dot_a);
+      const sqrt_k = vCheck.checkedResult(Math.sqrt(k));
```
zoddicus, line 70: Since there is a sqrt being performed here, you will probably want to put in a isNearZero check similar to distance, since you could potentially hit the same 1/inverseqqrt(x) problems

src/webgpu/shader/validation/expression/call/builtin/refract.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+    //   let k = 1.0 - c * c * (1.0 - dot(b, a) * dot(b, a))
+    //   if (k < 0.0) { return vecN(0.0); }
+    //   return c * a - (c * dot(b, a) + sqrt(k)) * b
+    // Should be invalid if the calculations result in intermediate values that
+    // exceed the maximum representable float value for the given type.
+    const a = Number(t.params.a);
+    const b = Number(t.params.b);
+    const c = Number(t.params.c);
+
+    const b_dot_a = vCheck.checkedResult(b * a * type.width);
```
zoddicus, line 60: Addition/subtraction operations are a potential source of may-fail behaviour, since OOB intermediates can be pull back in bounds by them.

Quickly looking at the examples here, the one_minus_b_dot_a_2, and c2_one_minus_b_dot_a_2, should be fine, since they are operating on non-negative values, due to the squaring, and the pull back in bounds can only occur when the operands have different signs (or are 0)... I guess you need to make sure one of the operands isn't 0 even for them.

For cdba_sqrt_k, I think there is a potential difference in signs, and thus potential may-fail. Specifically sqrt_k must be positive, but cdba can be negative, since I don't think the spec restricts the sign of c. I would take a closer look at that one to determine if it needs to be checked for may-fail.

zoddicus (changes requested):

src/webgpu/shader/validation/expression/call/builtin/refract.spec.ts
toji: `checkedResult()` checks every value passed to it to see if it's in that ambiguous overflow zone, and skips the test internally if that's the case.

```js
  checkedResult(value: number): number {
    if (this.isAmbiguousOverflow(value)) {
      this.t.skip(`Checked value, ${value}, was within the ambiguous overflow rounding range.`);
    }

    const quantizedValue = this.quantize(value);
    if (!Number.isFinite(quantizedValue)) {
      this.#allChecksPassed = false;
    }
    return quantizedValue;
  }
```

Does that satisfy your concern here, or is there something more that I'm missing?

src/webgpu/shader/validation/expression/call/builtin/refract.spec.ts
toji: Done.

toji:
Review ping?

src/webgpu/shader/validation/expression/call/builtin/refract.spec.ts
zoddicus: Yeah it does

zoddicus (approved):

ben-clayton (approved):
LGTM

toji:
Thank you! Still need @dneto0 to clear his changes requested bit.

src/webgpu/shader/validation/expression/call/builtin/refract.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+  });
+
+const kArgCases = {
+  good: '(vec3(0), vec3(1), 2.0)',
+  bad_no_parens: '',
+  // Bad number of args
+  bad_0args: '()',
+  bad_1arg: '(vec3(0))',
+  bad_2arg: '(vec3(0), vec3(1))',
+  bad_3arg: '(vec3(0), vec3(1), 2.0, vec3(3))',
```
dneto0, line 99: nit: this is bad_4arg.

src/webgpu/shader/validation/expression/call/builtin/refract.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+      t.params.arg === 'good',
+      `const c = ${builtin}${kArgCases[t.params.arg]};`
+    );
+  });
+
+const kReturnType = 'vec3f';
+
+g.test('return')
+  .desc(`Test ${builtin} return value type`)
+  .params(u => u.combine('type', ['vec3f', 'vec3u', 'vec3i', 'vec2f', 'u32', 'i32', 'f32', 'bool']))
```
dneto0, line 150: This is slightly weak sauce because only one one of the 9 possible return types is checked.
But I'm ok with that: those other ones are surely(?) checked in the execution tests, at least implicitly.

dneto0 (approved):
Thanks!  That vCheck mechanism makes for very readable tests!

src/webgpu/shader/validation/expression/call/builtin/refract.spec.ts
toji: Done.

src/webgpu/shader/validation/expression/call/builtin/refract.spec.ts
toji: I see what you mean. Refactored to do a better job of it, I'd appreciate you taking another look!

zoddicus (approved):

Issue #3623, opened 2024-04-09
`webgpu:shader,validation,expression,call,builtin,distance:values:type="f16";*` test failures


ben-clayton:
https://github.com/gpuweb/cts/pull/3585 introduced `distance()` validation tests.

These tests fail for `f16` with the error:

```
  - EXCEPTION: Error: Unexpected validation error occurred: Tint program failure: :4:18 error: '-65504.0 - 0.062469482421875' cannot be represented as 'f16'
    var<private> v = distance(f16(o0), f16(o1));
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^
    
    :4:18 note: when calculating distance
    var<private> v = distance(f16(o0), f16(o1));
```

The finite representable range for `f16` is `[-65504.0 .. +65504.0]`, so this looks like a valid failure (Tint is correct, CTS is wrong).

The `distance()` CTS attempts to detect the non-representable intermediate with:

https://github.com/gpuweb/cts/blob/afef918253ae6ba28bee20dfbf5b4e980103949f/src/webgpu/shader/validation/expression/call/builtin/distance.spec.ts#L65-L74

For `Type.f16`, `quantizeFn()` calls `quantizeToF16()`, which in turn calls `hfround()`, which is implemented as:
https://github.com/gpuweb/cts/blob/afef918253ae6ba28bee20dfbf5b4e980103949f/src/external/petamoriken/float16/float16.js#L1219-L1226
https://github.com/gpuweb/cts/blob/afef918253ae6ba28bee20dfbf5b4e980103949f/src/external/petamoriken/float16/float16.js#L344-L349

This does not appear to convert values outside of the finite range to infinity.

Interestingly, the `petamoriken/float16` library on GitHub appears substantially different, and does treat these as infinities:

https://github.com/petamoriken/float16/blob/60551c22d2eba68c829e04da8956a917786e7a92/src/_util/converter.mjs#L32-L63

@zoddicus - Should we try to update our version of `petamoriken/float16` ?

toji:
Well, for one I just realized that my Windows box at work doesn't have F16 support, which surprises me (and is why I missed this initially.)

[...]

PR #3636, opened 2024-04-10
Update petamoriken/float16 to v3.8.6

toji:
Should allow #3623 to be addressed in a follow-up PR.

Specifically merging to be able to replace uses of `roundToFloat16Bits` with `roundToFloat16` in our `hround` function in order to ensure it behaves more like Tint's internal behavior.


zoddicus (approved):
Thanks for doing this :+1:

PR #3643, opened 2024-04-11
wgsl: Add utility to help check for ambiguous overflows

### LINKED ISSUE #3623: `webgpu:shader,validation,expression,call,builtin,distance:values:type="f16";*` test failures

toji:
Fixes #3623

Primary thing to pay attention to is the new `ConstantOrOverrideValueChecker` utility class (rolls off the tongue, huh?) This is performing the same quantization step as before but now has the opportunity to force a subcase to be skipped if the checked value is found to be in a range where the overflow rounding behavior is ambiguous.

If we think this is a good direction then I'll go back and update a bunch of the tests I've already completed to use the new utility class for consistency/completeness.

<hr>

### DIFF
--- a/src/webgpu/shader/validation/expression/call/builtin/const_override_validation.ts
+++ b/src/webgpu/shader/validation/expression/call/builtin/const_override_validation.ts
@@ -1,6 +1,7 @@
 import { assert, unreachable } from '../../../../../../common/util/util.js';
 import { kValue } from '../../../../../util/constants.js';
 import {
+  ScalarType,
   Type,
   Value,
   elementTypeOf,
@@ -14,6 +15,9 @@ import {
   scalarF64Range,
   linearRange,
   linearRangeBigInt,
+  quantizeToF32,
+  quantizeToF16,
+  QuantizeFunc,
 } from '../../../../../util/math.js';
 import { ShaderValidationTest } from '../../../shader_validation_test.js';
 
@@ -311,4 +315,75 @@ export function unique<T>(...arrays: Array<readonly T[]>): T[] {
     }
   }
   return [...set];
+}
+
+interface FloatLimits {
+  positive: {
+    max: number;
+  },
+  negative: {
+    min: number;
+  },
+  emax: number;
+}
+
+/**
+ * Provides an easy way to validate steps in an equation that will trigger a validation error with
+ * constant or override values due to overflow/underflow. Typical call pattern is:
+ * 
+ * const vCheck = new ConstantOrOverrideValueChecker(t, Type.f32);
+ * const c = vCheck.checkedResult(a + b);
+ * const d = vCheck.checkedResult(c * c);
+ * const expectedResult = vCheck.allChecksPassed();
+ */
+export class ConstantOrOverrideValueChecker {
+  #allChecksPassed = true;
+  #floatLimits?: FloatLimits;
+  #quantizeFn: QuantizeFunc<number>;
+
+  constructor(private t: ShaderValidationTest, type: ScalarType) {
+    switch (type) {
+      case Type.f32:
+        this.#quantizeFn = quantizeToF32;
+        this.#floatLimits = kValue.f32;
+        break;
+      case Type.f16:
+        this.#quantizeFn = quantizeToF16;
+        this.#floatLimits = kValue.f16;
+        break;
+      default:
+        this.#quantizeFn = (v: number) => v;
+        break;
+    }
+  }
+
+  quantize(value: number): number {
+    return this.#quantizeFn(value);
+  }
+
+  // Some overflow floating point values may fall into an abiguously rounded scenario, where they
+  // can either round up to Infinity or down to the maximum representable value. In these cases the
+  // test should be skipped, because it's valid for implementations to differ.
+  isAmbiguousOverflow(value: number): boolean {
+    if (!this.#floatLimits ||
+        (value <= this.#floatLimits.positive.max &&
+          value >= this.#floatLimits.negative.min)) {
+          return false;
+    }
+    return Math.abs(value) < Math.pow(2, this.#floatLimits.emax+1);
+  }
+
+  checkedResult(value: number): number {
+    if (this.isAmbiguousOverflow(value)) {
+      this.t.skip(`Checked value, ${value}, was within the ambiguous overflow rounding range.`);
+    }
+    
+    const quantizedValue = this.quantize(value);
+    if (!Number.isFinite(quantizedValue)) {
+      this.#allChecksPassed = false;
+    }
+    return value;
+  }
+
+  allChecksPassed(): boolean { return this.#allChecksPassed; }
 }
--- a/src/webgpu/shader/validation/expression/call/builtin/const_override_validation.ts
+++ b/src/webgpu/shader/validation/expression/call/builtin/const_override_validation.ts
@@ -320,17 +320,19 @@ export function unique<T>(...arrays: Array<readonly T[]>): T[] {
 interface FloatLimits {
   positive: {
     max: number;
-  },
-  negative: {
     min: number;
-  },
+  };
+  negative: {
+    max: number;
+    min: number;
+  };
   emax: number;
 }
 
 /**
  * Provides an easy way to validate steps in an equation that will trigger a validation error with
  * constant or override values due to overflow/underflow. Typical call pattern is:
- * 
+ *
  * const vCheck = new ConstantOrOverrideValueChecker(t, Type.f32);
  * const c = vCheck.checkedResult(a + b);
  * const d = vCheck.checkedResult(c * c);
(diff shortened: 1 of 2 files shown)


src/webgpu/shader/validation/expression/call/builtin/const_override_validation.ts
```diff
@@ -312,3 +316,74 @@ export function unique<T>(...arrays: Array<readonly T[]>): T[] {
...
+ * const c = vCheck.checkedResult(a + b);
+ * const d = vCheck.checkedResult(c * c);
+ * const expectedResult = vCheck.allChecksPassed();
+ */
+export class ConstantOrOverrideValueChecker {
+  #allChecksPassed = true;
+  #floatLimits?: FloatLimits;
+  #quantizeFn: QuantizeFunc<number>;
+
+  constructor(private t: ShaderValidationTest, type: ScalarType) {
```
dneto0, line 344: I don't know JS well enough but I presume this makes class fields `t` and `type`. Otherwise you couldn't access `t` to skip the test later.

dneto0 (commented):
Looks pretty nice.

I'd like the isAmbiguousOverflow to be explicit about what it's supposed to do  with Infininties and naNs. (Well, Nans are the more interesting one, because the comparison at the end should take care of the infinity.)

And this would be improved with unit tests of isAmbiguousOverflow.

src/webgpu/shader/validation/expression/call/builtin/const_override_validation.ts
```diff
@@ -312,3 +316,74 @@ export function unique<T>(...arrays: Array<readonly T[]>): T[] {
...
+  // Some overflow floating point values may fall into an abiguously rounded scenario, where they
+  // can either round up to Infinity or down to the maximum representable value. In these cases the
+  // test should be skipped, because it's valid for implementations to differ.
+  isAmbiguousOverflow(value: number): boolean {
+    if (!this.#floatLimits ||
+        (value <= this.#floatLimits.positive.max &&
+          value >= this.#floatLimits.negative.min)) {
+          return false;
+    }
+    return Math.abs(value) < Math.pow(2, this.#floatLimits.emax+1);
```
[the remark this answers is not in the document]
zoddicus: I agree in about f64, since number is f64 we are already hitting issues with AF being an f64, if we add a concrete f64 (and presumably move AF to f128) there is going to need to be some major reworking in the CTS to support that.

src/webgpu/shader/validation/expression/call/builtin/distance.spec.ts
```diff
@@ -57,32 +45,22 @@ Validates that constant evaluation and override evaluation of ${builtin}() never
...
-    }
+    const ab = vCheck.checkedResult(a - b);
 
     // Only calculates the full length if the type is a vector. Otherwise abs(a-b) is used.
     if (kValidArgumentTypes[t.params.type].width > 1) {
-      const ab2 = quantizeFn(ab * ab);
-      const sqrLen = quantizeFn(ab2 * kValidArgumentTypes[t.params.type].width);
+      const ab2 = vCheck.checkedResult(ab * ab);
+      const sqrLen = vCheck.checkedResult(ab2 * kValidArgumentTypes[t.params.type].width);
       // Square root does not need to be calculated because it can never fail if
```
zoddicus, line 62: I don't think this comment is correct...
sqrt's accuracy is defined as 1/inverseSqrt(x)... which becomes a pain, because the inteval for inverseSqrt is 2 ULP, and division is 2.5 ULP plus not being define very close to 0 (https://www.w3.org/TR/WGSL/#floating-point-accuracy) So there should probably be a check that the value going into sqrt isn't near 0... and if it is, just skip, because it is likely to be a may fail case.

src/webgpu/shader/validation/expression/call/builtin/const_override_validation.ts
```diff
@@ -312,3 +316,74 @@ export function unique<T>(...arrays: Array<readonly T[]>): T[] {
...
+  }
+
+  quantize(value: number): number {
+    return this.#quantizeFn(value);
+  }
+
+  // Some overflow floating point values may fall into an abiguously rounded scenario, where they
+  // can either round up to Infinity or down to the maximum representable value. In these cases the
+  // test should be skipped, because it's valid for implementations to differ.
+  isAmbiguousOverflow(value: number): boolean {
```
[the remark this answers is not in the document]
zoddicus: From reading the code I think that for infinities it will always return false, because abs(+/-Infinit) is Infinity, which will never be less that 2^emax+1.

I think NaN just flows through and returns a false at the end, but I am not 100% sure about that.

PRobably worth being explict in the comment that values greater than 2^emax+1 will round to infinity and thus are not ambigious, i.e. will always fail validation

src/webgpu/shader/validation/expression/call/builtin/const_override_validation.ts
zoddicus: Yeah, it makes them private fields. I had to look this up along with the # syntax, since I don't normally use them

zoddicus (changes requested):
Overall I like it, other than the prexisting issue with sqrt handling that I think needs to be addressed, sorry for not catching that in an earlier review.

src/webgpu/shader/validation/expression/call/builtin/distance.spec.ts
toji: Thanks for pointing that out! Updated to add a near-zero check, implemented by testing to see if the value is between the min positive and max negative values for the type.

src/webgpu/shader/validation/expression/call/builtin/const_override_validation.ts
toji: Yep, though it's a typescript feature, not a JavaScript one.

Happy to remove any syntax that feels out of place or confusing, I'm just used to using these in my own code.

src/webgpu/shader/validation/expression/call/builtin/const_override_validation.ts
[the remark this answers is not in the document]
toji: Added a bunch of comments and made the finite check explicit.

src/webgpu/shader/validation/expression/call/builtin/const_override_validation.ts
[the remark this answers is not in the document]
toji: To clarify: Sounds like we don't need to do anything regarding f64 support now, but may need to in the future? Or are there changes that you'd like to see to how this is being handled today?

src/webgpu/shader/validation/expression/call/builtin/const_override_validation.ts
zoddicus: I am fine with leaving it as is, I don't find it particularly cryptic, was just a syntax I wasn't familar with.

src/webgpu/shader/validation/expression/call/builtin/const_override_validation.ts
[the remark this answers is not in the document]
zoddicus: No changes needed at this point (This is just mostly us waxing philospically about future evolutions of WGSL)

src/webgpu/shader/validation/expression/call/builtin/const_override_validation.ts
toji: TypeScript allows for private attributes to also be declared as:

`private someValue: number;`

If that reads better.

zoddicus (approved):

src/webgpu/shader/validation/expression/call/builtin/const_override_validation.ts
zoddicus: I am fine with either way

PR #3648, opened 2024-04-12
shader/validation: Add compound assignment validation

ben-clayton:
Extend the binary operator expression validation to also validate compound assignment.

Also calculate the binary operator's resulting type, and use that to validate the result type is as expected.

Issue: #1644

<hr>


src/webgpu/shader/validation/expression/binary/result_type.ts
```diff
@@ -0,0 +1,67 @@
...
+    // vector <op> vector
+    if (lhs.width !== rhs.width) {
+      return null;
+    }
+    const elementType = resultType({
+      lhs: lhs.elementType,
+      rhs: rhs.elementType,
+      canConvertScalarToVector,
+    }) as ScalarType | null;
+    return elementType !== null ? Type.vec(lhs.width, elementType) : null;
```
Kangz, line 33: I wonder if here and other places we can use the ?? operator (nullish coalescing). So it would be `return elementType ?? Type.vec(lhs.width, elementType)`

Kangz (approved):

src/webgpu/shader/validation/expression/binary/result_type.ts
ben-clayton: That's not quite the same thing - we're returning a vector of `elementType`, if `elementType` is not `null`, not `elementType` itself.

ben-clayton:
Thanks!

Issue #3251, opened 2024-01-09
reftests: Canvas display after device lost/destroyed


kainino0x:
I believe that according to the WebGPU spec, canvas [display](https://gpuweb.github.io/gpuweb/#abstract-opdef-get-a-copy-of-the-image-contents-of-a-context) should interact with device loss (and destroy()) as follows:

- After the device is lost but before calling getCurrentTexture, it should keep returning the last displayed frame
- After the device is lost and after calling getCurrentTexture, it should return:
    - a solid black image, if configure()d as opaque
    - a blank transparent image, if configure()d as transparent

Rough test cases:
- Render something, destroy the device, then wait a few frames before the screenshot is taken
- Render something, destroy the device, wait a few frames, call getCurrentTexture

kainino0x:
@perryuwang
The general instructions for CTS contributions are here:
https://github.com/gpuweb/cts/blob/main/docs/intro/README.md

The reftests live in this directory:
https://github.com/gpuweb/cts/tree/main/src/webgpu/web_platform/reftests

A good example to look at is the simplest one. Here is the test:
https://github.com/gpuweb/cts/blob/main/src/webgpu/web_platform/reftests/canvas_clear.https.html
Note its script is `canvas_clear.html.js` which is actually `canvas_clear.html.ts` in the repository.

Also note it points to it's `match` reference rendering `./ref/canvas_clear-ref.html`. A reftest compares the rendering of the page with the rendering of the `match` reference. Usually our reftests compare WebGPU rendering to 2D canvas rendering, which make it easy to render a solid color like that:
https://github.com/gpuweb/cts/blob/main/src/webgpu/web_platform/reftests/ref/canvas_clear-ref.html

For this test, I think a solid color canvas (one opaque and one transparent), similar to the canvas_clear test, is sufficient, except you will probably want to set the background color of the element or the page in order to test the transparency rendering.

Usually I recommend just visually comparing the rendering of the test and the reference. In order to run the pages locally, follow the setup instructions here:
https://github.com/gpuweb/cts/blob/main/docs/intro/developing.md
and see here for how to open the page on localhost:
https://github.com/gpuweb/cts/blob/main/docs/intro/developing.md#web-platform-tests-wpt---ref-tests
There are also instructions for actually running the reftest (with automatic comparison) in that documentation section, but since this test just needs to show a solid color it probably won't be necessary.

perryuwang:
[...]

PR #3669, opened 2024-04-16
Add reftests: canvas display after device lost

perryuwang:
Issue: #3251

<hr>


kainino0x (approved):
LGTM; please review the commit I pushed on the end!

src/webgpu/web_platform/reftests/canvas_display_after_device_lost.html.ts
```diff
@@ -0,0 +1,73 @@
...
+    }
+  }
+
+  function drawAll() {
+    draw('cvs00', 'opaque', { reconfigureAfterLost: false, drawAfterLost: false });
+    draw('cvs01', 'opaque', { reconfigureAfterLost: false, drawAfterLost: true });
+    draw('cvs02', 'premultiplied', { reconfigureAfterLost: false, drawAfterLost: false });
+    draw('cvs03', 'premultiplied', { reconfigureAfterLost: false, drawAfterLost: true });
+
+    draw('cvs10', 'opaque', { reconfigureAfterLost: true, drawAfterLost: false });
```
perryuwang, line 58: I think `cvs10` should display transparent color.

kainino0x: Hm, why is that? I think the spec says, reconfiguring should `Replace the drawing buffer` (which sets the drawing buffer to transparent black), then when the canvas is presented, `get a copy of the image contents of a context` clears the alpha to 0.

perryuwang: My understanding is that although reconfiguring has set the drawing buffer to black, but drawAfterLoss is false(getCurrentTexture() is not called), canvas is not redrawn and it is no content to presented, so it should be transparent.
For cvs11, because drawAfterLost is true, it should be displayed in black.

kainino0x: According to the spec, configure() also replaces the drawing buffer with a blank image (as do context creation, resize, unconfigure(), getCurrentTexture(), and transferToImageBitmap()).

Which makes me realize there should probably be some tests for unconfigure as well. I was going to add some but didn't think of why it would be useful.

kainino0x:
Let me remove my change from the end and open a new PR for the extra tests I want to add.

PR #3671, opened 2024-04-16
shader/validation: Move `break if` expression type validation out of `parse`

ben-clayton:
Use the same pattern by the other statement validation tests.

<hr>


ben-clayton:
Also adds testing for abstract types.

src/webgpu/shader/validation/statement/break_if.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+  'i32',
+  'u32',
+  'f32',
+  'f16',
+  'vec2f',
+  'vec3h',
+  'vec4u',
+  'vec3b',
+  'mat2x3f',
+  'mat4x2h',
```
dj2, line 20: This removed a bunch of the extra testing for matrices and vectors, which I think is fine, but it also dropped Struct which I think we possibly want to keep. We may also want to add array, texture, atomic and sampler here as well?

ben-clayton: User-declared and non-constructable types require a bit of work with this pattern. I agree, but I'd like to do this for all the statement tests in this directory, as they all follow the same pattern.

dj2: K, can you please file a bug with the details so we don't lose track?

dj2 (approved):

src/webgpu/shader/validation/statement/break_if.spec.ts
ben-clayton: Done: https://github.com/gpuweb/cts/issues/3685

PR #3675, opened 2024-04-17
Fix texel data test

Jiawei-Shao:
Issue: #

<hr>


Jiawei-Shao:
Hi @Kangz  and @jzm-intel ,

This PR is far from being mature enough but I'd like to discuss how to test the data conversion about texel data, and it seems the current way in the CTS is incorrect.

PTAL, thanks!

jzm-intel:
Currently the [WebGPU spec](https://gpuweb.github.io/gpuweb/#texture-formats:~:text=If%20the%20format%20has%20the%20%2Dsrgb%20suffix%2C%20then%20sRGB%20conversions%20from%20gamma%20to%20linear%20and%20vice%20versa%20are%20applied%20during%20the%20reading%20and%20writing%20of%20color%20values%20in%20the%20shader.) said
> If the format has the -srgb suffix, then sRGB conversions from gamma to linear and vice versa are applied during the reading and writing of color values in the shader.

Since the sRGB conversions are heavily non-linear, we should expect the result floats end up within some certain error boundary, like the case we seen for WGSL tests.

Jiawei suggested that current PR follows the accuracy requirement of [D3D spec](https://microsoft.github.io/DirectX-Specs/d3d/archive/D3D11_3_FunctionalSpec.htm#3.2.3.7%20SRGB%20-%3E%20FLOAT), which might be reasonable. It seems that Vulkan spec only refers to the conversions formula, didn't mention the accuracy; and the Metal spec (7.7.7 Conversion Rules for sRGBA and sBGRA Textures, P285) said the same accuracy requirement
> The precision of the above conversion must ensure that the delta between the resulting infinitely precise floating point value when converting result back to an unnormalized sRGB value but without rounding to an 8-bit unsigned integer value (call it r) and the original sRGB 8-bit unsigned integer color value (call it rorig) is <= 0.5


However, this seems like a gap in the WebGPU spec, not specifying the accuracy of such conversions. Maybe working group shall discuss this?

And another concern I have is, currently the [WebGPU spec](https://gpuweb.github.io/gpuweb/#color-spaces:~:text=WebGPU%20does%20not%20provide%20color%20management.%20All%20values%20within%20WebGPU%20(such%20as%20texture%20elements)%20are%20raw%20numeric%20values%2C%20not%20color%2Dmanaged%20color%20values.) also said that
> WebGPU does not provide color management. ***All values within WebGPU (such as texture elements) are raw numeric values, not color-managed color values.***
> 
> WebGPU does interface with color-managed outputs (via [GPUCanvasConfiguration](https://gpuweb.github.io/gpuweb/#dictdef-gpucanvasconfiguration)) and inputs (via [copyExternalImageToTexture()](https://gpuweb.github.io/gpuweb/#dom-gpuqueue-copyexternalimagetotexture) and [importExternalTexture()](https://gpuweb.github.io/gpuweb/#dom-gpudevice-importexternaltexture)). Thus, color conversion must be performed between the WebGPU numeric values and the external color values. Each such interface point locally defines an encoding (color space, transfer function, and alpha premultiplication) in which the WebGPU numeric values are to be interpreted.

Does this description conflict?

src/webgpu/util/conversion.ts
```diff
@@ -18,24 +18,40 @@ import {
...
- * Input must be between -1 and 1 if signed, or 0 and 1 if unsigned.
+ * Encodes a JS `number` into "normalized" (unorm/snorm) integer scale with `bits` bits for ULP
+ * comparison. Input must be between -1 and 1 if signed, or 0 and 1 if unsigned.
+ * e.g. float 0.5 -> "unorm8" 127.5
  *
  * MAINTENANCE_TODO: See if performance of texel_data improves if this function is pre-specialized
  * for a particular `bits`/`signed`.
  */
-export function floatAsNormalizedInteger(float: number, bits: number, signed: boolean): number {
+export function floatAsNormalizedIntegerScale(
```
Kangz, line 28: Scale -> Unquantized maybe? And explain that this doesn't cut off the fractional part of what would otherwise be an integer so that additional computations can be performed?

src/webgpu/util/texture/texel_data.spec.ts
```diff
@@ -106,16 +107,67 @@ function doTest(
...
+    // The procedure for measuring this tolerance is to convert the result back into SRGB space
+    // using the ideal FLOAT -> SRGB conversion specified below but WITHOUT the rounding to integer,
+    // and taking the floating point difference versus the original SRGB value to yield the error.
+    // Exact conversion is required: 0.0f and 1.0f (the ends) must be exactly achievable.
+    // https://microsoft.github.io/DirectX-Specs/d3d/archive/D3D11_3_FunctionalSpec.htm#3.2.3.7%20SRGB%20-%3E%20FLOAT
+    const readBackValue = await t.readGPUBufferRangeTyped(outputBuffer, {
+      type: Float32Array,
+      typedLength: 4,
+    });
+    t.expect(idealReadbackData[3] === readBackValue.data[3]);
```
Kangz, line 129: nit: comment that alpha doesn't gamma curved?

Kangz (approved):
LGTM

kainino0x:
> This doesn't conflict with the color space handling of `GPUCanvasContext` and friends because it is the responsibility of the rest of the Web platform to define these color spaces. `rgba8unorm-srgb` and `bgra8unorm-srgb` are part of the "number computations" of WebGPU and should work correctly.

Right. For example:

- `rgba8unorm` textures are not color-managed as sRGB-linear; they very often actually contain sRGB-nonlinear values, display-native (nonlinear) values, or non-color values.
- `rgba8unorm-srgb` textures are not color-managed as sRGB; they apply sRGB's gamma conversion when read or written in certain ways, but do not have to be sRGB - for instance Display P3 also uses sRGB's gamma conversion, and it's a decent approximation of display-native gamma curves too.

(This is a useful explanation. I will file a bug to put it in the spec.)

src/webgpu/util/texture/texel_data.spec.ts
```diff
@@ -106,16 +107,67 @@ function doTest(
...
+    })
   );
+
+  if (format === 'rgba8unorm-srgb' || format === 'bgra8unorm-srgb') {
+    // In D3D the SRGB -> Float conversion is permitted a tolerance of 0.5f ULP on the SRGB side.
+    // The procedure for measuring this tolerance is to convert the result back into SRGB space
+    // using the ideal FLOAT -> SRGB conversion specified below but WITHOUT the rounding to integer,
+    // and taking the floating point difference versus the original SRGB value to yield the error.
+    // Exact conversion is required: 0.0f and 1.0f (the ends) must be exactly achievable.
+    // https://microsoft.github.io/DirectX-Specs/d3d/archive/D3D11_3_FunctionalSpec.htm#3.2.3.7%20SRGB%20-%3E%20FLOAT
```
kainino0x, line 124: NIce fix! We've had problems here from the very beginning - a lot of tests that use the various texel data checking functions allow variances &le; 1 ULP, where really I wanted to allow variances &lt; 1 ULP (or for cases like this one, &le; 0.50001 ULP for "correctly rounded" results with a little tolerance for JS-side error) - but it seemed like a pretty difficult thing to fix so I decided it was OK to swallow the imprecision. Didn't realize this test had a real problem with it.

src/webgpu/util/texture/texel_data.spec.ts
```diff
@@ -106,16 +107,67 @@ function doTest(
...
+      type: Float32Array,
+      typedLength: 4,
+    });
+    t.expect(idealReadbackData[3] === readBackValue.data[3]);
+
+    // Do the ideal FLOAT -> SRGB conversion on the readback float value WITHOUT the rounding to
+    // integer.
+    const readBackValueToSRGB = new Float32Array(3);
+    for (let i = 0; i < 3; ++i) {
+      const gamaCompressed = gammaCompress(readBackValue.data[i]);
```
kainino0x, line 135: s/gama/gamma

src/webgpu/util/texture/texel_data.spec.ts
```diff
@@ -106,16 +107,67 @@ function doTest(
...
+    // https://microsoft.github.io/DirectX-Specs/d3d/archive/D3D11_3_FunctionalSpec.htm#3.2.3.7%20SRGB%20-%3E%20FLOAT
+    const readBackValue = await t.readGPUBufferRangeTyped(outputBuffer, {
+      type: Float32Array,
+      typedLength: 4,
+    });
+    t.expect(idealReadbackData[3] === readBackValue.data[3]);
+
+    // Do the ideal FLOAT -> SRGB conversion on the readback float value WITHOUT the rounding to
+    // integer.
+    const readBackValueToSRGB = new Float32Array(3);
```
kainino0x, line 133: Not sure we want the intermediate round from f64 to f32...? The 0/255 cases may depend on it currently, but I think those could be simplified to simply require 0.0/1.0 specifically. Then you can use a plain array of 3 `number`s here instead.

src/webgpu/util/texture/texel_data.spec.ts
```diff
@@ -106,16 +107,67 @@ function doTest(
...
+      const gamaCompressed = gammaCompress(readBackValue.data[i]);
+      let outputIndex = i;
+      if (format === 'bgra8unorm-srgb' && (i === 0 || i === 2)) {
+        outputIndex = 2 - i;
+      }
+      readBackValueToSRGB[outputIndex] = floatAsNormalizedIntegerScale(gamaCompressed, 8, false);
+    }
+
+    // Take the floating point difference versus the original SRGB value to yield the error.
+    const toleranceULP = 0.5;
```
kainino0x, line 144: nit: kToleranceULP

kainino0x, line 144: Should this be _slightly_ larger than 0.5 to allow for error on the JS side? (Though, if it passes, maybe it doesn't matter)

kainino0x (commented):
LGTM with a few notes, thank you for looking at this bug, we put it off too long!

kainino0x (approved):

jzm-intel:
Thanks @Kangz and @kainino0x for the explanations!

src/webgpu/util/texture/texel_data.spec.ts
Jiawei-Shao: I just keep using `0.5` as all the tests can pass right now. What do you think?

Issue #3657, opened 2024-04-12
errors in webgpu:shader,execution,expression,constructor,non_zero:*


kainino0x:
Example:
`webgpu:shader,execution,expression,constructor,non_zero:abstract_array_elements:abstract_type="abstract-int";concrete_type="f32";length=1`

`EXCEPTION: Error: Unexpected validation error occurred: Error while parsing WGSL: :22:22 error: cannot assign 'array<i32, 1>' to 'array<f32, 1>'`

Found in CTS->Dawn roll. It sounds like a test bug, I don't think this is supposed to be valid.

All of the following are failing with similar errors locally (M1 Mac, Chrome `125.0.6415.0 (Official Build) canary (arm64) `):

```
webgpu:shader,execution,expression,constructor,non_zero:abstract_vector_splat:*
webgpu:shader,execution,expression,constructor,non_zero:abstract_vector_elements:*
webgpu:shader,execution,expression,constructor,non_zero:abstract_vector_mix:*
webgpu:shader,execution,expression,constructor,non_zero:abstract_matrix_elements:*
webgpu:shader,execution,expression,constructor,non_zero:abstract_matrix_column_vectors:*
webgpu:shader,execution,expression,constructor,non_zero:abstract_array_elements:*
```

From #3631

kainino0x:
Tangentially(?) this test is unexpectedly passing on most configurations on Chrome's CQ: https://bugs.chromium.org/p/dawn/issues/detail?id=2514

kainino0x:
IIUC fixed in #3659? @ben-clayton

ben-clayton:
I believe so, but we might want to wait until the next roll confirms.

kainino0x:
The roll did not remove the expectations:
https://dawn-review.googlesource.com/c/dawn/+/184140/4/webgpu-cts/expectations.txt#1483

And locally they still fail with `unroll_const_eval_loops` disabled (and pass with it enabled).

ben-clayton:
Argh. Okay. I'll take another look tomorrow. Sorry about that.

kainino0x:
Thanks!

ben-clayton:
Fixed.

PR #3677, opened 2024-04-17
shader/execution: Fix abstract non_zero constructor tests

ben-clayton:
Depending on the platform, we either directly assign the expression evaluation to the output buffer, or go via a const array. The latter caused compilation failures as it would force a concretization of the abstract values in the array, which would become incompatible with the output type.

To force the harness to avoid the array, add a new optional `ConstEvaluationMode` flag to force the evaluation mode used.

Also replace the growing list of `ShaderBuilder` parameters with a structure, so that fewer places need to be updated for this sort of change in the future.

Issue: #3657

<hr>

### DIFF
--- a/src/webgpu/shader/execution/expression/expression.ts
+++ b/src/webgpu/shader/execution/expression/expression.ts
@@ -51,17 +51,21 @@ export type ConstEvaluationMode = 'direct' | 'unrolled' | 'loop';
 
 /** Configuration for running a expression test */
 export type Config = {
-  // Where the input values are read from
+  /** Where the input values are read from */
   inputSource: InputSource;
-  // If defined, scalar test cases will be packed into vectors of the given
-  // width, which must be 2, 3 or 4.
-  // Requires that all parameters of the expression overload are of a scalar
-  // type, and the return type of the expression overload is also a scalar type.
-  // If the number of test cases is not a multiple of the vector width, then the
-  // last scalar value is repeated to fill the last vector value.
+  /**
+   * If defined, scalar test cases will be packed into vectors of the given
+   * width, which must be 2, 3 or 4.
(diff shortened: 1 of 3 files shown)


src/webgpu/shader/execution/expression/expression.ts
```diff
@@ -48,6 +60,9 @@ export type Config = {
   // If the number of test cases is not a multiple of the vector width, then the
   // last scalar value is repeated to fill the last vector value.
   vectorize?: number;
+  // The evaluation mode used when 'inputSource' is 'const'. If undefined, then an appropriate mode
+  // will be picked based on the input types.
```
kainino0x, line 64: nit: These comments should be doc-comments.

src/webgpu/shader/execution/expression/expression.ts
```diff
@@ -664,41 +667,53 @@ function basicExpressionShaderBody(
     uniqueID: () => `cts_symbol_${nextUniqueIDSuffix++}`,
   };
   if (inputSource === 'const') {
+    let constEvaluationMode = params.constEvaluationMode;
+    if (constEvaluationMode === undefined) {
+      if (parameterTypes.some(ty => isAbstractType(scalarTypeOf(ty)))) {
+        // Directly assign the expression to the output, to avoid an
+        // intermediate store, which will concretize the value early
```
kainino0x, line 674: Are there valid use cases for non-'direct' modes with abstract types? If no, add an assert to make sure no one accidentally triggers this early concretization. If yes, maybe consider names like "unrolled-with-early-concretization"..?

kainino0x (approved):

src/webgpu/shader/execution/expression/expression.ts
ben-clayton: Done

PR #3699, opened 2024-04-22
Test that submit validates command buffers uniqueness

toji:
Issue first raised in https://github.com/gpuweb/gpuweb/issues/4367 with proposed validation behavior added to the spec in https://github.com/gpuweb/gpuweb/pull/4593. This PR should probably not land until that spec change has landed.

Ensures that command buffers are validated for uniqueness when submitting. Prevents an edge case which may allow command buffers to be submitted twice.

This behavior is not yet implemented in Chrome. Other browsers have not been tested but given that the requirement didn't exist in the spec previously it would be unsurprising if they lacked it as well.

<hr>

### DIFF
--- a/src/webgpu/api/validation/queue/submit.spec.ts
+++ b/src/webgpu/api/validation/queue/submit.spec.ts
@@ -45,3 +45,18 @@ g.test('command_buffer,device_mismatch')
       t.device.queue.submit([cb0, cb1]);
     }, mismatched);
   });
+
+g.test('command_buffer,duplicate_buffers')
+  .desc(
+    `
+    Tests submit cannot be called with the same command buffer listed multiple times:
+    `
+  )
+  .fn(t => {
+    const encoder0 = t.device.createCommandEncoder();
+    const cb = encoder0.finish();
+
+    t.expectValidationError(() => {
+      t.device.queue.submit([cb, cb]);
+    }, true);
+  });


lokokung:
Agree that we should land this after the spec changes, but LGTM otherwise! Thanks for raising this Brandon!

lokokung (approved):

src/webgpu/api/validation/queue/submit.spec.ts
```diff
@@ -45,3 +45,18 @@ g.test('command_buffer,device_mismatch')
...
+    `
+    Tests submit cannot be called with the same command buffer listed multiple times:
+    `
+  )
+  .fn(t => {
+    const encoder0 = t.device.createCommandEncoder();
+    const cb = encoder0.finish();
+
+    t.expectValidationError(() => {
+      t.device.queue.submit([cb, cb]);
```
kainino0x, line 60: Should test whether `cb` is invalid after this or not. (Your spec PR would say that it is still valid.)

toji: Hm... looks like we're still figuring that one out? According to https://bugs.chromium.org/p/dawn/issues/detail?id=1863 anyway. (Dawn invalidates them whether validation passes or not: https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/src/dawn/native/Queue.cpp;l=285;drc=f0f5f3ceebb00da9363ccc7a1e2c0f17b6b383ba)

kainino0x: OK, we can test it afterward. I dunno why I didn't file a spec issue about it?

kainino0x (approved):

src/webgpu/api/validation/queue/submit.spec.ts
kainino0x: https://github.com/gpuweb/gpuweb/issues/4599

PR #3708, opened 2024-04-24
wgsl: Add validation tests for vector access of abstract types

zoddicus:
<hr>


src/webgpu/shader/validation/expression/access/vector.spec.ts
```diff
@@ -174,23 +174,189 @@ const kCases = {
...
+  ryb: { wgsl: 'const r = V.ryb;', ok: false },
+  xgza: { wgsl: 'const r = V.xgza;', ok: false },
+
+  // error: too many swizzle elements
+  xxxxx: { wgsl: 'const r = V.xxxxx;', ok: false },
+  rrrrr: { wgsl: 'const r = V.rrrrr;', ok: false },
+  yxwxy: { wgsl: 'const r = V.yxwxy;', ok: false },
+  rgbar: { wgsl: 'const r = V.rgbar;', ok: false },
+
+  // error: inValid index Value
```
ben-clayton, line 333: übernit:
```suggestion
  // error: invalid index Value
```

src/webgpu/shader/validation/expression/access/vector.spec.ts
```diff
@@ -174,23 +174,189 @@ const kCases = {
...
+  literal_5: { wgsl: 'const r = V[5];', ok: false },
+  literal_minus_1: { wgsl: 'const r = V[-1];', ok: false },
+
+  // error: inValid index type
```
ben-clayton, line 337:
```suggestion
  // error: invalid index type
```

src/webgpu/shader/validation/expression/access/vector.spec.ts
```diff
@@ -174,23 +174,189 @@ const kCases = {
...
+  const_expr_2_via_struct: {
+    wgsl: 'const r = V[S(2).i];',
+    ok: (width: number) => width > 2,
+  },
+  const_expr_3_via_struct: {
+    wgsl: 'const r = V[S(3).i];',
+    ok: (width: number) => width > 3,
+  },
+
+  // single element conVenience name accesses
```
ben-clayton, line 280:
```suggestion
  // single element convenience name accesses
```

src/webgpu/shader/validation/expression/access/vector.spec.ts
```diff
@@ -174,23 +174,189 @@ const kCases = {
   array_idx: { wgsl: 'let r : T = v[array<i32, 2>()];', ok: false },
 };
 
-g.test('vector')
-  .desc('Tests validation of vector indexed and swizzles')
+const kAbstractCases = {
+  // indexing with literal
+  literal_0: { wgsl: 'const r = V[0];', ok: true },
```
ben-clayton, line 179: Maybe as a followup, but it would be good to check the expression type (same thing for the concrete tests), by using an explicit type for `r` (`const r : T = V[0];`).

Unlike the concrete tests, the abstract tests would need to consider that `abstract-int` can convert to `abstract-float`, but the reverse is not true.

ben-clayton (approved):

src/webgpu/shader/validation/expression/access/vector.spec.ts
zoddicus: Done

src/webgpu/shader/validation/expression/access/vector.spec.ts
zoddicus: Done

src/webgpu/shader/validation/expression/access/vector.spec.ts
zoddicus: Done

src/webgpu/shader/validation/expression/access/vector.spec.ts
zoddicus: I think I might be missing something about how abstracts in WGSL work.

I thought you couldn't explicitly type abstract int and float in WGSL, i.e. there is now way to write `const r: AbstractFloat`, it can only be inferred from the assignment.

And they are always constant (or override) experessions, so I cannot do something like `let r = 0; r = V[0];` to check that V[0] is an AbstractInt, because though the initializer literal is a AI, it will be converted on the assignement to r to be an i32, due to the use of `let`.

ben-clayton: You're right that you can't explicitly write an abstract numeric, and I see why what I wrote is confusing.

`T` would be a concrete type - `i32`, `u32`, `f32` etc. I was trying to point out that `abstract-int` could be used for any of these, but `abstract-float` would only work for `f32`.

ben-clayton: Maybe it would be better done as a second statement:

```wgsl
const C : T = r;
```

That would mean you still test that `r` gets the inferred abstract type.
You also wouldn't need to change all these individual cases, just add another statement underneath the templated case one.

zoddicus: I understand now. I have filed https://github.com/gpuweb/cts/issues/3709 and will implement in a follow up CL.

ben-clayton: Currently we have nothing testing that the *inferred* type of `r` is as expected. The second assignment to a concrete type would validate that the inferred type is as expected.

PR #3715, opened 2024-04-25
Add more cases to canvas reftest with device loss

kainino0x:
Currently blocked on https://github.com/gpuweb/gpuweb/issues/4606
Expected results need careful review.

Issue: #3251

<hr>

### DIFF
--- a/src/webgpu/web_platform/reftests/canvas_display_after_device_lost.html.ts
+++ b/src/webgpu/web_platform/reftests/canvas_display_after_device_lost.html.ts
@@ -19,9 +19,14 @@ void (async () => {
     canvasId: string,
     alphaMode: GPUCanvasAlphaMode,
     {
+      unconfigureBeforeLost,
       reconfigureAfterLost,
       drawAfterLost,
-    }: { reconfigureAfterLost: boolean; drawAfterLost: boolean }
+    }: {
+      unconfigureBeforeLost: boolean;
+      reconfigureAfterLost: boolean;
+      drawAfterLost: boolean;
+    }
   ) {
     const canvas = document.getElementById(canvasId) as HTMLCanvasElement;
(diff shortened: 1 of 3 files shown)


kainino0x:
cc @perryuwang

perryuwang:
So my understanding is as follows, is it accurate?
When canvas is configured but getCurrentTexture is not called:
- Display a transparent image if configure()d as transparent.
- Display a black image if configure()d as opaque.

kainino0x:
I think so. It should be the same regardless of whether getCurrentTexture is called: both configure() and getCurrentTexture() wipe the canvas to blank (transparent or opaque depending on the configuration).

src/webgpu/web_platform/reftests/canvas_display_after_device_lost.html.ts
```diff
@@ -15,42 +15,83 @@ void (async () => {
...
+            },
+          ],
+        });
+        pass.end();
+        device.queue.submit([encoder.finish()]);
+      } catch (ex) {
+        threw = true;
+      }
+
+      assert(
```
kainino0x, line 61:
```suggestion
      // If this assert fails, takeScreenshotDelayed will never be called, and the test will time out .
      assert(
```

kainino0x:
https://github.com/gpuweb/gpuweb/issues/4606 was resolved, and this test should still be correct.

toji (approved):
A couple of nits but LGTM.

kainino0x:
This is no longer valid because the spec allows multiple behaviors. The existing test also isn't valid anymore (not sure if it was removed already or not).

It _is_ possible to test multiple behaviors:
https://web-platform-tests.org/writing-tests/reftests.html#multiple-references
So this test should be fixed to do that. But we might not be able to test so many cases in a single reftest anymore, since behavior can vary between setups.

PR #3736, opened 2024-05-07
Test variable address space restrictions for shader stages

### LINKED ISSUE #2398: address space restrictions vs. shader stage
workgroup variables are only usable in compute shaders.

storage buffer, storage textures are only usable in compute and fragment. (disallow in vertex shader)

https://github.com/gpuweb/gpuweb/pull/3947

alan-baker:
Fixes #2398

* Test the restrictions for each address space per shader stage
  * workgroup only in compute
  * RW buffer and texture not in vertex

Tint fails the RW buffer and textures by allowing them in the vertex stage.

Issue: #

<hr>

### DIFF
--- a/src/webgpu/shader/validation/decl/var.spec.ts
+++ b/src/webgpu/shader/validation/decl/var.spec.ts
@@ -776,6 +776,7 @@ g.test('shader_stage')
       .combine('stage', ['compute', 'vertex', 'fragment'] as const)
       .combine('kind', [
         'handle_ro',
+        'handle_wo',
         'handle_rw',
         'function',
         'private',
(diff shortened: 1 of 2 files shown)


src/webgpu/shader/validation/decl/var.spec.ts
```diff
@@ -768,3 +768,84 @@ g.test('var_access_mode_bad_template_delim')
...
+        fdecl = `var v : u32;`;
+        break;
+      case 'private':
+        mdecl = `var<private> v : i32;`;
+        break;
+      case 'storage_ro':
+        mdecl = `@group(0) @binding(0) var<storage> v : u32;`;
+        break;
+      case 'storage_rw':
+        mdecl = `@group(0) @binding(0) var<storage, read_write> v : u32;`;
```
dj2, line 810: For the tests with `read` and `read_write` should there also be a `write` that always fails? So `storage_w` and `handle_w` tests

alan-baker: We don't allow write for storage buffer, but it could be added for handle. In fact, this test needs to skip the handle_rw based on a language feature too!

dj2: Right, it isn't allowed, but isn't this the place we should have validation that it correctly isn't allowed?

alan-baker: That's covered in `address_space_access_mode` test suite.

dj2 (approved):

PR #3757, opened 2024-05-21
WGSL builtin textureSample w/derivatives test

greggman:
This test is different than the pattern set by the non-derivative test. The existing test renders an array of coordinates. The coordinates might come from data passed to the shader, or they might be hard coded with multiple `textureSample` calls. The results are written to a storage buffer and compared to computed expected values.

This test instead renders a quad (2 triangles) to an rgba32float texture given a simple shader. It then "software renders" the expected result and compares the 2 results.

There are a large number of permutations of parameters that affect texture sampling. The sampler itself has wrapping modes and sampling modes. `textureSample` is influenced by texture format, coordinates, derivatives and offsets. Testing all combinations X all texture formats would make the tests too slow so, these tests focus mostly on the effect of derivatives and assume the combinations in the non-derivative case give effective coverage of the other permutations.

The tolerances for whether the actual result match the expected result are rather large. They are influenced by the OpenGL dEQP test suite which has similar tolerances. The hope is these are a good starting point to get the tests written with a structure that can be adjusted later should it be decided to make these tests more strict.

There's a chromium run here: https://dawn-review.googlesource.com/c/dawn/+/188725/2?checksPatchset=2&tab=checks

Note: another possibility is I discard this PR and try to fit derivatives into the existing (rendering a list of coords) framework. I looked into that and it's certainly possible. I can use the fragment shader's `@builtin(position)` as the source of the changing value for the derivative. For any call into a texture builtin I subtract the expected frag position for that call as a constant. It has to be constant otherwise the calculation would affect the derivative. But, because it would have to be a constant, that would make every call unique which makes the shader very large.  If that's a better direction though I can switch.

The advantage to that direction is, testing other things that have to be constants gets more coverage. For example, if you check that test you'll see, for `offset`, the test parameter is `true`, `false` and if `true` then a bunch of pseudo random constants are used generating a shader with N calls to `textureSample`, one per offset. Where as in the derivatives test it just tests 2 pairs of constants and hopes that's good enough.

<hr>

### DIFF
--- a/src/webgpu/shader/execution/expression/call/builtin/textureSample.spec.ts
+++ b/src/webgpu/shader/execution/expression/call/builtin/textureSample.spec.ts
@@ -10,11 +10,12 @@ import { kTexelRepresentationInfo } from '../../../../../util/texture/texel_data
 
 import {
   vec2,
-  createRandomTexelView,
   TextureCall,
   putDataInTextureThenDrawAndCheckResults,
+  putDataInTextureThenDrawAndCheckResultsComparedToSoftwareRasterizer,
   generateSamplePoints,
   kSamplePointMethods,
+  createRandomTexelViewMipmap,
 } from './texture_utils.js';
 import { generateCoordBoundaries, generateOffsets } from './utils.js';
 
@@ -98,7 +99,7 @@ Parameters:
       size: { width: 8, height: 8 },
       usage: GPUTextureUsage.COPY_DST | GPUTextureUsage.TEXTURE_BINDING,
     };
-    const texelView = createRandomTexelView(descriptor);
+    const texelViews = createRandomTexelViewMipmap(descriptor);
     const calls: TextureCall<vec2>[] = generateSamplePoints(50, t.params.minFilter === 'nearest', {
       method: t.params.sample_points,
       textureWidth: 8,
@@ -120,7 +121,7 @@ Parameters:
     };
     const res = await putDataInTextureThenDrawAndCheckResults(
       t.device,
-      { texels: texelView, descriptor },
+      { texels: texelViews, descriptor },
       sampler,
       calls
     );
@@ -137,7 +138,65 @@ fn textureSample(t: texture_2d<f32>, s: sampler, coords: vec2<f32>, offset: vec2
 test mip level selection based on derivatives
     `
   )
-  .unimplemented();
+  .params(u =>
+    u
+      .combine('format', kEncodableTextureFormats)
+      .filter(t => {
+        const type = kTextureFormatInfo[t.format].color?.type;
+        return type === 'float' || type === 'unfilterable-float';
+      })
+      .combine('mipmapFilter', ['nearest', 'linear'] as const)
+      // note: this is the derivative we want at sample time. It is not the value
+      // passed directly to the shader. This way if we change the texture size
+      // or render target size we can compute the correct values to achieve the
+      // same results.
+      .combineWithParams([
+        { ddx: 0.5, ddy: 0.5 }, // test mag filter
+        { ddx: 1, ddy: 1 }, // test level 0
+        { ddx: 2, ddy: 1 }, // test level 1 via ddx
+        { ddx: 1, ddy: 4 }, // test level 2 via ddy
+        { ddx: 1.5, ddy: 1.5 }, // test mix between 1 and 2
+        { ddx: 6, ddy: 6 }, // test mix between 2 and 3 (there is no 3 so we should get just 2)
+        { ddx: 1.5, ddy: 1.5, offset: [7, -8] }, // test mix between 1 and 2 with offset
+        { ddx: 1.5, ddy: 1.5, offset: [3, -3] }, // test mix between 1 and 2 with offset
+        { ddx: 1.5, ddy: 1.5, ddxStart: -3, ddyStart: -4 }, // test mix between 1 and 2 with negative coords
+      ])
+  )
+  .beforeAllSubcases(t => {
+    const { format } = t.params;
+    const formatInfo = kTexelRepresentationInfo[format];
+    t.skipIfTextureFormatNotSupported(format);
+    const hasFloat32 = formatInfo.componentOrder.some(c => {
+      const info = formatInfo.componentInfo[c]!;
+      return info.dataType === 'float' && info.bitLength === 32;
+    });
+    if (hasFloat32) {
+      t.selectDeviceOrSkipTestCase('float32-filterable');
+    }
+  })
+  .fn(t => {
+    const { format, mipmapFilter, ddx, ddy, ddxStart, ddyStart, offset } = t.params;
+    const descriptor: GPUTextureDescriptor = {
+      format,
+      mipLevelCount: 3,
+      size: { width: 8, height: 8 },
+      usage: GPUTextureUsage.COPY_DST | GPUTextureUsage.TEXTURE_BINDING,
+    };
+    const texelViews = createRandomTexelViewMipmap(descriptor);
+    const sampler: GPUSamplerDescriptor = {
+      addressModeU: 'repeat',
+      addressModeV: 'repeat',
+      minFilter: 'linear',
+      magFilter: 'linear',
+      mipmapFilter,
+    };
+    putDataInTextureThenDrawAndCheckResultsComparedToSoftwareRasterizer(
+      t,
+      { texels: texelViews, descriptor },
+      sampler,
+      { ddx, ddy, ddxStart, ddyStart, offset: offset as vec2 }
+    );
+  });
 
 g.test('sampled_3d_coords')
   .specURL('https://www.w3.org/TR/WGSL/#texturesample')
--- a/src/webgpu/shader/execution/expression/call/builtin/textureSample.spec.ts
+++ b/src/webgpu/shader/execution/expression/call/builtin/textureSample.spec.ts
@@ -157,9 +157,9 @@ test mip level selection based on derivatives
         { ddx: 1, ddy: 4 }, // test level 2 via ddy
         { ddx: 1.5, ddy: 1.5 }, // test mix between 1 and 2
         { ddx: 6, ddy: 6 }, // test mix between 2 and 3 (there is no 3 so we should get just 2)
-        { ddx: 1.5, ddy: 1.5, offset: [7, -8] }, // test mix between 1 and 2 with offset
-        { ddx: 1.5, ddy: 1.5, offset: [3, -3] }, // test mix between 1 and 2 with offset
-        { ddx: 1.5, ddy: 1.5, ddxStart: -3, ddyStart: -4 }, // test mix between 1 and 2 with negative coords
+        { ddx: 1.5, ddy: 1.5, offset: [7, -8] as const }, // test mix between 1 and 2 with offset
+        { ddx: 1.5, ddy: 1.5, offset: [3, -3] as const }, // test mix between 1 and 2 with offset
+        { ddx: 1.5, ddy: 1.5, uvwStart: [-3.5, -4] as const }, // test mix between 1 and 2 with negative coords
       ])
   )
   .beforeAllSubcases(t => {
@@ -175,7 +175,7 @@ test mip level selection based on derivatives
     }
   })
   .fn(t => {
-    const { format, mipmapFilter, ddx, ddy, ddxStart, ddyStart, offset } = t.params;
+    const { format, mipmapFilter, ddx, ddy, uvwStart, offset } = t.params;
     const descriptor: GPUTextureDescriptor = {
       format,
       mipLevelCount: 3,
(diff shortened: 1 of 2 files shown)


greggman:
@kainino0x, you reviewed the other `textureSample` PR so maybe you want to review this?

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
```diff
@@ -291,6 +401,212 @@ export async function putDataInTextureThenDrawAndCheckResults<T extends Dimensio
...
+    { ddx, ddy, ddxStart, ddyStart, offset }
+  );
+
+  // Note: I'm not sure what we should do here. My assumption is, given texels
+  // have random values, the difference between 2 texels can be very large. In
+  // the current version, for a float texture they can be +/- 1000 difference.
+  // Sampling is very GPU dependent. So if one pixel gets a random value of
+  // -1000 and the neighboring pixel gets +1000 then any slight variation in how
+  // sampling is applied will generate a large difference when interpolating
+  // between -1000 and +1000.
```
kainino0x, line 554: Based on reading this comment I think what you have is fine. Would it actually help if the random numbers weren't as widely spread? I think it would be fine to constrain them more but I'm not sure it actually reduces the tolerances you need.

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
```diff
@@ -291,6 +401,212 @@ export async function putDataInTextureThenDrawAndCheckResults<T extends Dimensio
...
+  const format = 'rgba32float';
+
+  const rep = kTexelRepresentationInfo[format];
+
+  const expData = new Float32Array(width * height * 4);
+  for (let y = 0; y < height; ++y) {
+    const fragY = height - y - 1 + 0.5;
+    for (let x = 0; x < width; ++x) {
+      const fragX = x + 0.5;
+      const coords = [(fragX / width) * ddx - ddxStart, (fragY / height) * ddy - ddyStart] as T;
```
kainino0x, line 425: This bit of math is the most mysterious to me. Comment it?

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
```diff
@@ -222,6 +262,73 @@ export function expected<T extends Dimensionality>(
...
+  assert(call.ddx !== undefined);
+  assert(call.ddy !== undefined);
+  const rep = kTexelRepresentationInfo[texture.texels[0].format];
+  const texSize = reifyExtent3D(texture.descriptor.size);
+  const textureSize = [texSize.width, texSize.height];
+
+  const ddx: readonly number[] = typeof call.ddx === 'number' ? [call.ddx] : call.ddx;
+  const ddy: readonly number[] = typeof call.ddy === 'number' ? [call.ddy] : call.ddy;
+  const sDdx = ddx.map((v, i) => (v * textureSize[i]) / size[i]);
+  const sDdy = ddy.map((v, i) => (v * textureSize[i]) / size[i]);
```
kainino0x, line 285: Comment what these variables are. Maybe stick to leibniz notation, rename ddx/ddy to dPdx/dPdy (define what P is) and sDdx/sDdy to dSdx/dSdy (define what S is)

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
```diff
@@ -222,6 +262,73 @@ export function expected<T extends Dimensionality>(
...
+  const mipLevel = 0.5 * Math.log2(deltaMax);
+  const mipLevelCount = texture.texels.length;
+  const maxLevel = mipLevelCount - 1;
+
+  switch (sampler.mipmapFilter) {
+    case 'linear': {
+      const clampedMipLevel = clamp(mipLevel, { min: 0, max: maxLevel });
+      const baseMipLevel = Math.floor(clampedMipLevel);
+      const nextMipLevel = Math.ceil(clampedMipLevel);
+      const t0 = expected<T>(call, texture, sampler, baseMipLevel);
```
kainino0x, line 300: nit: perhaps rename `expected` to something more descriptive

kainino0x (commented):
LGTM I think, but I will rereview when I have a bit more time

src/webgpu/shader/execution/expression/call/builtin/textureSample.spec.ts
```diff
@@ -137,7 +138,65 @@ fn textureSample(t: texture_2d<f32>, s: sampler, coords: vec2<f32>, offset: vec2
...
+        { ddx: 6, ddy: 6 }, // test mix between 2 and 3 (there is no 3 so we should get just 2)
+        { ddx: 1.5, ddy: 1.5, offset: [7, -8] }, // test mix between 1 and 2 with offset
+        { ddx: 1.5, ddy: 1.5, offset: [3, -3] }, // test mix between 1 and 2 with offset
+        { ddx: 1.5, ddy: 1.5, ddxStart: -3, ddyStart: -4 }, // test mix between 1 and 2 with negative coords
+      ])
+  )
+  .beforeAllSubcases(t => {
+    const { format } = t.params;
+    const formatInfo = kTexelRepresentationInfo[format];
+    t.skipIfTextureFormatNotSupported(format);
```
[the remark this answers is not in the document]
greggman: These are checking different things and I'm not sure they can easily be switched?

`selectDeviceOrSkipTestCase` tests a `GPUDeviceRequestDescriptor` if a certain "feature" exists. In this case it's checking in `???32float` can be filtered. If not it skips the test. 

`skipIfTextureFormatNotSupported` tests for compat formats ('bgra8unorm-srgb') 

I'm not sure I can really combine those tests. `rgba32float` as a format always exists, the only feature we care about is if it's filterable. `bgra8unorm-srgb` doesn't exist in compat. So I'm not sure there is a single check (is this supported) that fits both case.

src/webgpu/shader/execution/expression/call/builtin/textureSample.spec.ts
```diff
@@ -137,7 +138,65 @@ fn textureSample(t: texture_2d<f32>, s: sampler, coords: vec2<f32>, offset: vec2
...
+      addressModeV: 'repeat',
+      minFilter: 'linear',
+      magFilter: 'linear',
+      mipmapFilter,
+    };
+    putDataInTextureThenDrawAndCheckResultsComparedToSoftwareRasterizer(
+      t,
+      { texels: texelViews, descriptor },
+      sampler,
+      { ddx, ddy, ddxStart, ddyStart, offset: offset as vec2 }
```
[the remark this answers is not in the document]
greggman: Fixed

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
```diff
@@ -291,6 +401,212 @@ export async function putDataInTextureThenDrawAndCheckResults<T extends Dimensio
...
+    for (let x = 0; x < width; ++x) {
+      const fragX = x + 0.5;
+      const coords = [(fragX / width) * ddx - ddxStart, (fragY / height) * ddy - ddyStart] as T;
+      const call: TextureCall<T> = {
+        builtin: 'textureSample',
+        coordType: 'f',
+        coords,
+        ddx: [ddx, 0] as T,
+        ddy: [0, ddy] as T,
+        offset: options.offset as T,
```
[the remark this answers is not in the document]
greggman: The order is

```
putDataInTextureThenDrawAndCheckResultsComparedToSoftwareRasterizer<T extends Dimensionality>
 softwareRasterize<T extends Dimensionality>
   softwareTextureRead<T extends Dimensionality>
     expected<T extends Dimensionality>
```

In the next PR, `T` will be a `vec3` and in another PR, `T` will be a `number`. Or at least that's the hope. I'll have to refactor stuff.

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
greggman: added

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
greggman: done

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
greggman: The one time I tried this I had to lower them all the way to +2/-2 for whatever tolerances I originally had. I don't know if +2/-2 is better (Because we use a lower tolerance) or just more of the same.  The dEQP is quantizing to 8 bits so I think that makes up much of the difference.

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
greggman: I changed the way this works. I changed this so `ddx` and `ddy` are the values passed to `textureSampleGrad` and for this test cases they're computed outside.

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
kainino0x: Ahh, totally missed that this was just a copy of the same math you did in the WGSL. Explains why it felt sort of arbitrary :)

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
```diff
@@ -291,6 +410,238 @@ export async function putDataInTextureThenDrawAndCheckResults<T extends Dimensio
...
+
+  const rep = kTexelRepresentationInfo[format];
+
+  const expData = new Float32Array(width * height * 4);
+  for (let y = 0; y < height; ++y) {
+    const fragY = height - y - 1 + 0.5;
+    for (let x = 0; x < width; ++x) {
+      const fragX = x + 0.5;
+      // This code calculates the same value that will be passed to
+      // `textureSample` in the fragment shader for a given frag coord. That
```
kainino0x, line 439:
```suggestion
      // `textureSample` in the fragment shader for a given frag coord (see the
      // WGSL code which uses the same formula, but using interpolation). That
```

kainino0x (approved):
Definitely clearer now, thanks! Just a few small comments but LGTM

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
```diff
@@ -412,26 +421,49 @@ export function softwareRasterize<T extends Dimensionality>(
...
+      // here we're computing the `uv` value that will be passed for a
+      // particular fragment coordinate. fragX / width, fragY / height provides
+      // the unitQuad value.
+      //
+      // ddx and ddy in this case are the derivative values we want to test. We
+      // pass those into the softwareTextureRead<T> as they would normally be
+      // derived from the change in coord.
+      const coords = [
+        (fragX / width) * screenSpaceDDX + uvwStart[0],
+        (fragY / height) * screenSpaceDDY + uvwStart[1],
```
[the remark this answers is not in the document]
greggman: It was a bug fix. They should both be added. The reason it worked before (I think) is because I was adding an integer start it didn't affect the results because `'repeat'`. I changed the test start to a non-integer

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
```diff
@@ -291,6 +410,238 @@ export async function putDataInTextureThenDrawAndCheckResults<T extends Dimensio
...
+  targetSize: [number, number],
+  options: TextureTestOptions
+) {
+  const [width, height] = targetSize;
+  const { ddx = 1, ddy = 1, uvwStart = [0, 0] } = options;
+  const format = 'rgba32float';
+
+  const textureSize = reifyExtent3D(texture.descriptor.size);
+  const screenSpaceDDX = (ddx * width) / textureSize.width;
+  const screenSpaceDDY = (ddy * height) / textureSize.height;
```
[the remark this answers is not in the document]
greggman: I renamed the variable and added a comment and a MAINTENANCE_TODO as it wasn't clear to me how to phrase things that the calculations happen to be the same. I'm going to have to modify this code to support 3D texture so I hope something will come up then.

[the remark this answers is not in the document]
kainino0x: I thought they looked the same, though I'm not positive they're actually equal. Here:
`(ddx * targetSize.width) / textureSize.width`
there:
`(ddx * renderTarget.width) / textureSize.width`
and then you use them the ~same way in the UV formulas; here:
`(fragX / width) * screenSpaceUMult + uvwStart[0], (fragY / height) * screenSpaceVMult + uvwStart[1]`
there:
`(pos * 0.5 + 0.5) * vec2f(${uMult}, ${vMult}) + vec2f(${uvwStart[0]}, ${uvwStart[1]})`

Anyway if it doesn't work out you can just leave behind a comment explaining that it's equivalent to the uMult/vMult in the shader (though nit, give them the same variable name if they're the same, and slightly clearer names if they're actually different)

PR #3777, opened 2024-06-01
Update validation tests for 3D textures with BC compression

mehmetoguzderin:
Partially fixes https://github.com/gpuweb/cts/issues/3761

Please let me know if I missed or overlooked anything. Your guidance would be appreciated. Thank you!

<hr>

**Requirements for PR author:**

- [x] All missing test coverage is tracked with "TODO" or `.unimplemented()`.
- [x] New helpers are `/** documented */` and new helper files are found in `helper_index.txt`.
- [x] Test behaves as expected in a WebGPU implementation. (If not passing, explain above.)
- [x] Test have be tested with compatibility mode validation enabled and behave as expected. (If not passing, explain above.)

**Requirements for [reviewer sign-off](https://github.com/gpuweb/cts/blob/main/docs/reviews.md):**

- [ ] Tests are properly located in the test tree.
- [ ] [Test descriptions](https://github.com/gpuweb/cts/blob/main/docs/intro/plans.md) allow a reader to "read only the test plans and evaluate coverage completeness", and accurately reflect the test code.
- [ ] Tests provide complete coverage (including validation control cases). **Missing coverage MUST be covered by TODOs.**
- [ ] Helpers and types promote readability and maintainability.

When landing this PR, be sure to make any necessary issue status updates.

PR #3786, opened 2024-06-10
Revert "Update validation tests for 3D textures with BC compression"

kainino0x:
Reverts gpuweb/cts#3777

Changing a helper affected some other tests. Those tests need to be fixed, or changed temporarily to skip the affected cases temporarily (with a TODO).


mehmetoguzderin (approved):
SGTM

PR #3792, opened 2024-06-11
Make `virtualMipSize` take a `GPUExtent3D`

greggman:
Updated the TODO for making it return a GPUExtent3DDict

Note: I'm torn on which is better, 3 numbers or the object. The object seems verbose. The object is also inconsistent with `workgroup_size`

Being able to compute the number of texels with

```js
numTexels = size.reduce((acc, v) => acc * v);
```

Seems nice though I admit it's hardly that different than

```js
numTexls = size.width * size.height * size.depthOrArrayLayers;
```

It's also nice to be able to pass the size. For example algos which take a size of texture passed to a compute shader

```js
pass.dispatchWorkgroups(...size);
```

vs

```js
pass.dispatchWorkgroups(size.width, size.height, size.depthOrArrayLayers);
```

Or putting it into a uniform/storage buffer with 

```js
someTypedArray.set(size, offset);
```

Though again it's not that big if a deal. I get that it's kind of more readable to see the names but if that's the case the same argument applies to `workgroup_size` so 🤷‍ 😅

Size as an array also matches `textureDimensions` which returns a `vec2u` or `vec3u` which can be indexed with `[0]`, `[1]`, `[2]` and which is often useful directly in texture coordinate calculations `texelCoord = uv * size`

anyway, I'm not suggesting changing more now. Mostly voicing that I'd mildly prefer it not be changed to return a `GPUExtent3DDict` in the future.

### DIFF
--- a/src/webgpu/util/texture/base.ts
+++ b/src/webgpu/util/texture/base.ts
@@ -110,22 +110,22 @@ export function physicalMipSizeFromTexture(
 /**
  * Compute the "virtual size" of a mip level of a texture (not accounting for texel block rounding).
  *
- * MAINTENANCE_TODO: Change input/output to Required<GPUExtent3DDict> for consistency.
+ * MAINTENANCE_TODO: Change output to Required<GPUExtent3DDict> for consistency.
  */
 export function virtualMipSize(
   dimension: GPUTextureDimension,
-  size: readonly [number, number, number],
+  size: GPUExtent3D,
   mipLevel: number
 ): [number, number, number] {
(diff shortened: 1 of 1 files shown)


kainino0x (approved):
I prefer the array style in general as well, I tried to remove the dict style from the spec before we shipped but people didn't want to :)

The dictionary does have somewhat better readability because it doesn't make you use [0] [1] [2], but anyway I gave up on bothering to try to make things consistent throughout the CTS as it doesn't really matter that much. Easy enough to patch things as we go, like this.

Issue #3755, opened 2024-05-20
Add coverage for validation of >=1 target in render passes


ErichDonGubler:
Following up from <https://github.com/gpuweb/cts/issues/3754>, we should add coverage that checks validation for at least one render target. See also section 10.3.1. (`Render Pipeline Creation`), which [states](https://www.w3.org/TR/webgpu/#abstract-opdef-validating-gpurenderpipelinedescriptor) in the subsection titled `validating GPURenderPipelineDescriptor(descriptor, layout, device)`:

> There must exist at least one attachment, either:
> 
> -   A non-`null` value in descriptor.[`fragment`](https://www.w3.org/TR/webgpu/#dom-gpurenderpipelinedescriptor-fragment).[`targets`](https://www.w3.org/TR/webgpu/#dom-gpufragmentstate-targets), or
>     
> -   A descriptor.[`depthStencil`](https://www.w3.org/TR/webgpu/#dom-gpurenderpipelinedescriptor-depthstencil).

PR #3808, opened 2024-06-22
Test createRenderPipeline fail without any attachment

sagudev:
Test for yourself: [https://sagudev.github.io/cts/standalone/?q=webgpu:api,validation,render_pipeline,misc:no_attachment:*](https://sagudev.github.io/cts/standalone/?q=webgpu:api,validation,render_pipeline,misc:no_attachment:*)

Only non-Async case works in Firefox (Firefox raises validation error instead of returning reject in every Async test).

Edge fails both, because it's non-conformant to the spec.

Issue: https://github.com/gpuweb/cts/issues/3755

<hr>


kainino0x (approved):
LGTM! Normally I'd say we should have a control case to make sure that it passes if there is an attachment (to make sure there are no other errors in the test that cause it to just fail unconditionally) but I think this is simple enough we don't need it.

PR #3813, opened 2024-06-24
Compat: Add a test for requestDevice

greggman:
Compat mode on Chrome in OpenGLES fails this test. Other backends it passes. Specifically, in Chrome when the backend is set to OpenGLES, requestAdapter returns an adapter when NOT requesting compatibilityMode but then returns null for a device when calling requestDevice. That behavior is not spec compliant.

<hr>


kainino0x (approved):
LGTM, I am confused how Chromium is returning null from requestAdapter because I couldn't find any code that would do that, but I'm sure it's buried in there somewhere.

src/webgpu/api/operation/adapter/requestDevice.spec.ts
```diff
@@ -433,3 +433,46 @@ g.test('limit,worse_than_default')
...
+    tested in the same browser configuration.
+  `
+  )
+  .params(u => u.combine('compatibilityMode', [false, true] as const))
+  .fn(async t => {
+    const { compatibilityMode } = t.params;
+    const gpu = getGPU(t.rec);
+    // MAINTENANCE_TODO: Remove this cast compatibilityMode is added.
+    const adapter = await gpu.requestAdapter({ compatibilityMode } as GPURequestAdapterOptions);
+    if (adapter) {
```
kainino0x, line 465: not important, but note: I don't remember if any of the other WebGPU tests pass if requestAdapter returns null. If they would fail in that case, then really this test doesn't need to pass either.

greggman: I think maybe I'm not understanding the your comment?

On a device that only supports compatibility mode, `requestAdapter()` should turn `null` whereas `requestAdapter({compatibilityMode: true})` will return a `GPUAdapter`.

kainino0x: oh yes, I see. Never mind!

Issue #1752, opened 2022-08-12
Track all the buffers and textures with TrackForCleanup()


Jiawei-Shao:
We should track all the buffers and textures with TrackForCleanup() in the test body so that they can be safely destroyed after the completion of the tests. Otherwise these resources may be destroyed at a random moment by GC, which will cause some flaky test failures due to the bugs in buffer.destroy() or texture.destroy().

kainino0x:
@Jiawei-Shao What are the bugs in buffer.destroy() and texture.destroy(), can you link to them?

Jiawei-Shao:
The related issue reports are [dawn:1628](https://bugs.chromium.org/p/dawn/issues/detail?id=1628) and [chromium:1237175](https://bugs.chromium.org/p/chromium/issues/detail?id=1237175).

PR #3816, opened 2024-06-25
Require query sets to be tracked for cleanup

kainino0x:
This results in some redundant tracking (e.g. tracking a query set immediately before destroying it), but this should not cause any problems.

I did not dry-run the entire CTS but I spot checked `webgpu:api,validation,encoding,beginComputePass:*` and `webgpu:api,validation,query_set,destroy:*` and it should be safe.

If this is successful I'll do the same for createBuffer/createTexture (a much larger change).

Issue: #1752

<hr>


kainino0x:
> 2\. [I tried making a a wrapped version](https://github.com/greggman/cts/commit/1230337c1ac1c748931744197ad69673ab0c9377). Unfortunately you run into the problem that it can't be `GPUDevice` directly because there is no way from the device back to the test.

I tried this and had the same problem. We could override the methods on GPUDevice before each test (overwriting any previous overrides each time, because one device is used for multiple tests). Or we could add an element to the prototype chain that shadows the original functions, and just hold the current fixture in an expando on the GPUDevice so we don't have to overwrite the methods every time. Or move the tracking to expandos of the GPUDevice and clear them out at the end of the test (least weird).

But not sure that's worthwhile and feels like it risks more bugs.

> 1. the rule that you can't call createXXX without tracking. Are there any buffer/texture tests that can't do that? Probably not? I guess you can disable the rule there with `/* eslint-disable xxx */`.

We'll see! But yes, `eslint-disable-next-line` works.

PR #3819, opened 2024-06-25
Set document.title to query in standalone CTS

greggman:
This means history, like long pressing on the back button, will show preivous tests vs now where they all just say 'WebGPU CTS'


kainino0x (approved):
Good idea!

PR #3835, opened 2024-07-03
Rename *.worker.js to *.as_worker.js

### LINKED ISSUE #3834: `npm run wpt` generates `*.worker.js` tests that accidentally get run by `wptrunner`
`*.worker.js` files are included in `out-wpt/` output generated by `npm run wpt`. This file extension appears to have been settled on in isolation from consideration of WPT tests, which _actually_ treat `*.worker.js` files as a form of test specification (see WPT upstream docs' section titled ["Dedicated worker tests (`.worker.js`)"](https://web-platform-tests.org/writing-tests/testharness.html#dedicated-worker-tests-worker-js).

We need to move away from this file extension for worker tests that aren't intended to be run in WPT.

kainino0x:
Avoid ".worker.js" because that conflicts with WPT test naming patterns.

Fixes #3834

I'll need to fix Dawn to handle this change though so I guess I'm going to wait a little bit until the CTS->Dawn roller is less far behind? (https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/third_party/gn/webgpu-cts/BUILD.gn;l=68;drc=ca296e18d92a21099bbf6b878de2c234207468ff)


ErichDonGubler (approved):
LGTM!

Q: Are these scripts going to be eventually used by WPT output?

kainino0x:
Verified that the dedicated worker mode does still work in WPT, by copying the output into WPT and loading up `https://web-platform.test:8443/webgpu/cts.https.html?worker=dedicated&q=webgpu:print_environment:*`

> Q: Are these scripts going to be eventually used by WPT output?

In theory yes. I just tried and `worker=shared` and `worker=service` don't work because of a path issue. They should work in principle though.

kainino0x:
> In theory yes. I just tried and `worker=shared` and `worker=service` don't work because of a path issue. It should be possible to make them work though, I think.

Never mind, it works fine! The path was correct, I just couldn't load them because the self-signed cert for `web-platform.test:8443` isn't good enough for shared/service workers in Chrome. Works fine on localhost:
<http://localhost:8000/webgpu/cts.https.html?worker=service&q=webgpu:print_environment:*>
<http://localhost:8000/webgpu/cts.https.html?worker=shared&q=webgpu:print_environment:*>

Issue #3845, opened 2024-07-08
wgsl: compound assignment for &= and |= generating invalid WGSL when types are bool or bool vector


dneto0:
```
 struct Output {
   @size(4) value : u32
 };
 @group(0) @binding(0) var<storage, read_write> outputs : array<Output, 4>;


 struct Input {
   @size(4) lhs : u32,
   @size(4) rhs : u32,
 }

 @group(0) @binding(1) var<storage, read_write> inputs : array<Input, 4>;

 @compute @workgroup_size(1)
 fn main() {
   for (var i = 0; i < 4; i++) {

         outputs[i].value = u32(inputs[i].lhs);
         outputs[i].value &= bool(inputs[i].rhs);
   }
 }
```

Fallout from #3842

PR #3846, opened 2024-07-08
Fix compound assignment storage_rw when resultType != storage(resultT…

dneto0:
…ype)

In the storage_rw case, when the result type is not the same as storage(result type) then generate the full expansion (stylistically):

     var ret = LHSType(LHS);
     ret {op}= RHSType(RHS);
     outputs.value = ret;

instead of:

     outputs.value = {storageType}(LHS);
     outputs.value {op}= ret;

This situation occurs when the value type is bool or bool-vec. In that case the result type is u32, and there are no mixed overloads for & and | such as:

   u32 & bool
   u32 | bool

Fixes: #3845

Issue: #3845

<hr>


dneto0:
Tested on Mac

PR #3874, opened 2024-07-24
Cover GPUExternalTexture with WebCam Cases

shaoboyan:
Add utils to capture frame from camera if env supported. And compared the rendering result with 2d context draw image to cover uploading webcam frames.

Issue: #

<hr>


shaoboyan:
This PR has been tested with chromium 1-copy path and passed on Windows. But when using 0-copy path, the new added cases report some pixel diffs (suspect due to pixel position diff). 
Send out the PR for reviewing now and continue investigating 0-copy path error.

P.S. Tries to add the test as reftest first. But reftest requries two html as source and ref. But the camera capture cases requires presenting the same captured frame, I don't think in reftest we can achieve this (with two different html files.)

src/webgpu/web_platform/util.ts
```diff
@@ -466,7 +466,9 @@ export async function getVideoFrameFromVideoElement(
 
   return raceWithRejectOnTimeout(
     new Promise<VideoFrame>(resolve => {
-      const videoTrack: MediaStreamVideoTrack = video.captureStream().getVideoTracks()[0];
+      const videoTrack: MediaStreamVideoTrack = video
+        .captureStream()
+        .getVideoTracks()[0] as MediaStreamVideoTrack;
```
Kangz, line 471: What else can the type be that requires a cast?

src/webgpu/web_platform/util.ts
```diff
@@ -547,3 +549,60 @@ function callbackHelper(
...
+export async function captureCameraFrame(test: GPUTest): Promise<VideoFrame> {
+  if (
+    typeof navigator.mediaDevices === 'undefined' ||
+    typeof navigator.mediaDevices.getUserMedia === 'undefined'
+  ) {
+    test.skip("Browser doesn't support capture frame from camera.");
+  }
+
+  const stream = await navigator.mediaDevices.getUserMedia({ video: true });
+  const track: MediaStreamVideoTrack = stream.getVideoTracks()[0] as MediaStreamVideoTrack;
```
Kangz, line 570: Same question. Also if we do a cast we don't need to explicitly type `track`.

src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -37,16 +39,34 @@ function createExternalTextureSamplingTestPipeline(
...
+            const pos = array(
+              vec2( 1.0,  1.0),
+              vec2( 1.0, -1.0),
+              vec2(-1.0, -1.0),
+              vec2( 1.0,  1.0),
+              vec2(-1.0, -1.0),
+              vec2(-1.0,  1.0),
+            );
+
+            const uv = array(
```
Kangz, line 57: note that the uv could be generate from `(pos + vec2(1)) / 2`

Kangz: But also why did this need to change compared to the previous approach used fragcoord?

src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -603,3 +623,99 @@ compute shader, for several combinations of video format, video color spaces and
...
+      t.skip('Captured video frame has 0 height or width.');
+    }
+
+    const frameWidth = frame.displayWidth;
+    const frameHeight = frame.displayHeight;
+
+    // Use WebGPU + GPUExternalTexture to render the captured frame.
+    const colorAttachment = t.createTextureTracked({
+      format: kFormat,
+      size: { width: frameWidth, height: frameHeight, depthOrArrayLayers: 1 },
```
Kangz, line 654:
```suggestion
      size: { width: frameWidth, height: frameHeight },
```

Kangz (approved):
LGTM

src/webgpu/util/copy_to_texture.ts
```diff
@@ -55,7 +55,7 @@ export const kCopySubrectInfo = [
   },
 ] as const;
 
-export class CopyToTextureUtils extends TextureTestMixin(GPUTest) {
```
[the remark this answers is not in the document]
shaoboyan: Ah, I think I misunderstand some concepts. My understand is that "Uploading" means transfer something to something, by copy or not. So I rename the "copyTo" to "uploading" to support "import". Seems I need a better name but I would like to revert the rename and use "copyTo" for now.

src/webgpu/web_platform/util.ts
shaoboyan: It could be "MediaStreamTrack". Unfortunately, this type conflict is introduced by "imageCapture" type file (through reference webrtc here https://www.npmjs.com/package/@types/w3c-image-capture?activeTab=code). The conflicts are:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/webrtc/MediaStream.d.ts#L86 which return type is `MediaStreamTrack`
and https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/dom-mediacapture-transform/index.d.ts#L28 which return type is `MediaStreamVideoTrack`.

src/webgpu/web_platform/external_texture/video.spec.ts
shaoboyan: In previous shader code, frag use fragcoord and calculate uv coord by setting texture size. Instead of using uniform to pass the texture size, I think maybe this way could do better.

Kangz:
Still LGTM, feel free to merge.

PR #3910, opened 2024-08-21
Reduce the number of shader modules used in texture tests

greggman:
This change reduces the number of shader modules created by 2 orders of magnitude. The issue is offsets must be constants so randomly generating them makes a new shader. This makes them less random so they are the same per test.

So for example: webgpu:shader,execution,expression,call,builtin,textureSampleLevel:sampled_2d_coords:* goes from 1197 shader modules to 4


toji (approved):
LGTM, thanks for reducing the test load!

PR #3911, opened 2024-08-21
Print more info in certain asserts

greggman:
This is useful for CQs where bots have different hardware then devs have locally.


toji (approved):
LGTM, thanks for improving the debug info we can get here.

PR #3912, opened 2024-08-21
Choose smaller texture sizes

greggman:
The old code did an lcm of width vs height where for for cubemaps because cubemaps must be square and
textures must be a multiple of block sizes.

With a format with a blockSize of 5x8 and a minSize of 32 that would end up doing lcm of 35x32 which is 1120. Then for a cube array it would end up allocating
1120x1120x24 and if the format is rgba32float that's 418meg.

The new code just gets the lcm of the blockWidth vs blockHeight and then aligning to that which will be much much smaller.


toji (approved):
LGTM

PR #3919, opened 2024-08-26
Only report on first texture builtin error

greggman:
Each texture test tests 50 calls to `textureXXX`. It then checks they all match as expected. For each one that does NOT match expected a slow binary search happens to find the samples that influenced the results.

This change makes it so only the first failure does this slow check. Ideally no tests would fail but when they do, checking all 50 sample points can cause test bots to timeout.

Note: There's another minor change to show integer results as integers.

### DIFF
--- a/src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
+++ b/src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
@@ -1308,7 +1308,7 @@ export async function checkCallResults<T extends Dimensionality>(
       ? getMaxFractionalDiffForTextureFormat(texture.descriptor.format)
       : 0;
 
-  for (let callIdx = 0; callIdx < calls.length /*&& errs.length === 0*/; callIdx++) {
+  for (let callIdx = 0; callIdx < calls.length; callIdx++) {
     const call = calls[callIdx];
     const gotRGBA = results[callIdx];
     const expectRGBA = softwareTextureReadLevel(t, call, texture, sampler, call.mipLevel ?? 0);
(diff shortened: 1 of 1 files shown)


src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
```diff
@@ -1308,7 +1308,7 @@ export async function checkCallResults<T extends Dimensionality>(
       ? getMaxFractionalDiffForTextureFormat(texture.descriptor.format)
       : 0;
 
-  for (let callIdx = 0; callIdx < calls.length /*&& errs.length === 0*/; callIdx++) {
+  for (let callIdx = 0; callIdx < calls.length && errs.length === 0; callIdx++) {
```
shrekshao, line 1311: nit: why not `break` at end of `if (bad) {}` branch?

shrekshao (approved):
LGTM with nit

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
greggman: done

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
```diff
@@ -1410,6 +1414,10 @@ export async function checkCallResults<T extends Dimensionality>(
         errs.push('  sample points:');
         errs.push(layoutTwoColumns(expectedSamplePoints, gotSamplePoints).join('\n'));
         errs.push('', '');
+
+        // This path is slow so if we took it, don't report the other errors. One is enough
+        // to fail the test.
+        break;
```
shrekshao, line 1420: Looks like this is inside `if(sampler)`?
I think it should be between line 1421 and 1422.
We can add comments to brackets to better visualize.
```
} // if (sampler)
} // if (bad)
```

greggman: Yes, because that's the slow path. Reporting all failures without samplers is fast.

greggman: I can add the comments though

shrekshao: Ok. If it is intended to only break for the sampler part not for all bad part then it looks good. (Not really need to tweak the bracket comment)

PR #3921, opened 2024-08-28
Print more info for texture builtins

greggman:


shrekshao (approved):
LGTM (I made a comment about he position of `break` in the last commit)

PR #3922, opened 2024-08-28
Fix textureGather

greggman:
The code wasn't correctly discarding edge cases for negative texture coordinates.  For example -2.5 sits in the center of a texel and for textureGather could go either left or right. We need to avoid that case.

Also, textureGather always does `nearest` filtering mipmapFilter. This was fine as it was since it would always pass 0 for mipLevel which mean the 2nd mip level always provided zero contribution but arguably it should take the `nearest` path.


shrekshao (approved):
LGTM

PR #3924, opened 2024-08-30
Execution tests for subgroupAny and subgroupAll

alan-baker:
* Compute tests with all active invocation and partially active invocations
* Fragment tests with all active invocations
* Removed unimplemented dynamically uniform subgroupBroadcast test (due to const requirement)

Issue: #

<hr>


src/webgpu/shader/execution/expression/call/builtin/subgroupAll.spec.ts
```diff
@@ -0,0 +1,401 @@
...
+    if (filter(id, size)) {
+      const subgroup_id = output[numInvs + inv];
+      const expected_v = expected.get(subgroup_id) ?? 0;
+      if (expected_v !== res) {
+        return new Error(`Invocation ${inv}:
+- expected: ${expected_v}
+-      got: ${res}`);
+      }
+    } else {
+      if (res !== 999) {
```
dneto0, line 83: This 999 is mysterious.  I see it's the default value in the output buffer, as set by runComputeTest.

The 999 should be a defined constant exported from subgroup_util.ts.

src/webgpu/shader/execution/expression/call/builtin/subgroupAll.spec.ts
```diff
@@ -0,0 +1,401 @@
...
+
+  // Record a representative subgroup id.
+  outputs[lid + ${wgThreads}] = subgroupBroadcastFirst(lid);
+
+  let res = select(0u, 1u, subgroupAll(bool(inputs[lid])));
+  outputs[lid] = res;
+}`;
+
+    const prng = new PRNG(t.params.case);
+    // Case 0 is all 0s.
```
dneto0, line 143: Yeah, thanks for putting the two special cases in. It's important!

src/webgpu/shader/execution/expression/call/builtin/subgroupAll.spec.ts
```diff
@@ -0,0 +1,401 @@
...
+    // Case 1 is all 1s.
+    // Other cases are filled with random 0s and 1s.
```
dneto0, line 145: It's good to have a few random ones.

But this allows an incorrect but simple implementation to nearly always pass this test:  Look at the first half of the bits only.

How about:
increase kNumCases to 15
a decl just before `inputData` to pick out which bit to set to 0.
   let dropout = prng.uniformInt(32); 
/// Cases 2 to 9 are many 1 bits except for a single randomly selected 0 (in position 0..31)
/// Other cases are filled with random 0s and 1s.
then inside  the iterRange:
   ...
       else if (t.params.case < 10) {
           return x === dropout ? 0 : 1; 
       } else {
           return prng.uniformInt(2); 
       }

src/webgpu/shader/execution/expression/call/builtin/subgroupAll.spec.ts
```diff
@@ -0,0 +1,401 @@
...
+        if (t.params.case === 0) {
+          return 0;
+        } else if (t.params.case === 1) {
+          return 1;
+        }
+        return prng.uniformInt(2);
+      }),
+    ]);
+
+    const uintsPerOutput = 2;
```
dneto0, line 241: uintsPerOutput unused?

src/webgpu/shader/execution/expression/call/builtin/subgroupAll.spec.ts
dneto0: Also, I recommend making a helper function to generate the input data, since the same logic is used twice and is (going to get) pretty complicated.

src/webgpu/shader/execution/expression/call/builtin/subgroupAll.spec.ts
```diff
@@ -0,0 +1,401 @@
...
+
+@group(0) @binding(0)
+var<storage, read_write> inputs : Inputs;
+
+@fragment
+fn main(
+  @builtin(position) pos : vec4f,
+) -> @location(0) vec2u {
+  var subgroup_id = 0u;
+  if subgroupElect() {
```
dneto0, line 369: I recommend avoiding divergence here.
The subgroup representative IDs don't have to be sequential.  So it's ok to delete the if condition; it's ok for every invocation to bump the counter.  The broadcast on line 372 still sends the representative ID to all the invocations in the subgroup.

dneto0 (changes requested):
This goes a long way, thanks!

Are you planning to write fragment tests with split subgroups? If so, then add an .unimplemented() test for it.

alan-baker:
> This goes a long way, thanks!
> 
> Are you planning to write fragment tests with split subgroups? If so, then add an .unimplemented() test for it.

I've tried a little, but it gets tricky quickly. Compilers collapsing conditional code can easily lead to too many active invocations as was shown in the experimental reconvergence tests. So for now I'm not planning on adding those tests and relying on the compute versions.

dneto0:
> I've tried a little, but it gets tricky quickly.

ok, sounds good.  Let's make sure this is tracked somehow: an .unimplemented() test seems best so it has a chance of getting revisited in the future.

alan-baker:
Decided to pull the fragment tests for now.

dneto0 (approved):
Thanks!

PR #3929, opened 2024-09-03
Get mip level mix weights just once.

greggman:
The issue here was, given the code to get weights is async, multiple tests would pass through this code, see that the weights are not available, and create the GPU code to get them.

With this change, the first test will get them, the other tests will wait for the same results.


shrekshao (approved):
LGTM

PR #3930, opened 2024-09-03
Same pipeline identify sample points

greggman:


src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
```diff
@@ -3530,75 +3538,88 @@ ${body}
...
-  return out;
+    return out;
+  };
+
+  const results = await run(gpuTexture);
+
+  return {
+    run,
+    results,
+    destroy() {
```
shrekshao, line 3620: Q: is this destroy called anywhere now or is meant to be called at a future commit?

shrekshao (approved):
LGTM with a question

PR #3931, opened 2024-09-03
Improve cube-array error logging

greggman:
Before:

     layer: 6 (undefined)

After:

     layer: 6, cube-layer: 1 (+x)

The issue was the layer is in 2d layers not cube layers. Cube layers = six 2d layers. So, now it prints both, and it prints the correct face name (+x, -x, +y, ...) for the face when layer >= 6


shrekshao (approved):
LGTM

PR #3935, opened 2024-09-06
More WGSL flat issues

greggman:
Fix a few more tests that should no longer fail in compat as errors have been moved from shader creation time to pipeline creation time.


shrekshao (approved):
LGTM

PR #3937, opened 2024-09-06
Skip identifying sample points unless debugging

greggman:
This will skip the texture samples doing the heavy binary search for sample locations


src/common/framework/test_config.ts
```diff
@@ -36,6 +36,11 @@ export type TestConfig = {
    * Whether to enable the `logToWebSocket` function used for out-of-band test logging.
    */
   logToWebSocket: boolean;
+
+  /**
+   * Whether to run slow failure diagnostics. In particular, computing texture sample locations.
+   */
+  verboseFailures: boolean;
```
kainino0x, line 43: I think it would make sense for this to be part of the same flag as enableDebugLogs. The usage could be something like:

```ts
if (failure) {
  if (t.rec.debugging) {
    // ...
  }
}
```

or

```ts
if (failure) {
  t.debug(() => {
    // all this code runs only if debug logs are enabled
  });
}
```

I'm OK with adding a new flag if you prefer it though.

greggman: No, it's find to use debug. Uploaded a new PR.

src/common/internal/logging/test_case_recorder.ts
```diff
@@ -44,11 +44,16 @@ export class TestCaseRecorder {
   private startTime = -1;
   private logs: LogMessageWithStack[] = [];
   private logLinesAtCurrentSeverity = 0;
-  private debugging = false;
+  private _debugging = false;
```
kainino0x, line 47: nit: I think `public readonly` would work just as well

kainino0x (approved):

src/common/internal/logging/test_case_recorder.ts
greggman: yea, it'd just be settable. I guess we don't care.

kainino0x: `readonly` would make it not settable (except in the constructor)

PR #3938, opened 2024-09-06
Avoid subnormals in expression validation for quantizing types (in normalize and refract)

petermcneeleychromium:
We want to avoid specifying behavior for subnormal values of quantized floating types (f32/f16)
We can not specify a validation outcome because that would depend on the exact expression evaluation computations and their interaction with subnormal values.

The flush to zero behavior is under-specified.
https://github.com/gpuweb/cts/blob/main/docs/fp_primer.md#subnormal-numbers

crbug.com/341282612

### DIFF
--- a/src/webgpu/shader/validation/expression/call/builtin/normalize.spec.ts
+++ b/src/webgpu/shader/validation/expression/call/builtin/normalize.spec.ts
@@ -12,7 +12,13 @@ import {
   scalarTypeOf,
   ScalarType,
 } from '../../../../../util/conversion.js';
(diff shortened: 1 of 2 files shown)


dneto0 (approved):

dneto0:
It's failing the bots due to formatting issues.  run `npm run fix` from the CTS root dir.

petermcneeleychromium:
I am having to fix the linting by hand. The --fix command is not worked no matter which variation I try

petermcneeleychromium:
Thanks I tried the command and it worked!

PR #3939, opened 2024-09-06
Add Origin Trial token for Subgroups in localhost:8080

dneto0:
Adds a localhost:8080 token to standalone/index.html for local testing.

https://crbug.com/358117283

Followup to #3895

Issue: #

<hr>


kainino0x (commented):
Previously we've served up the localhost tokens from `dev_server.ts` instead. Do you need this because you're running this through some other server on localhost?

If so, I think it's redundant to have it in both, so we could have it only here and not in `dev_server.ts`.

standalone/index.html
```diff
@@ -15,6 +15,8 @@
     <meta name="viewport" content="width=device-width" />
     <!-- Chrome "WebGPU Subgroups Features" origin trial token for https://gpuweb.github.io (see dev_server.ts for localhost tokens) -->
```
kainino0x, line 16: Also be sure to update this comment
```suggestion
    <!-- Chrome "WebGPU Subgroups Features" origin trial token for https://gpuweb.github.io -->
```

kainino0x (approved):
Realized that other servers _are_ supposed to work when building to disk (`out/` directory) so it makes sense to support localhost:8080 on those.

We can remove the token from `dev_server.ts` (though it's not critical). If you do, update the comment below

standalone/index.html
dneto0: yep, done.

kainino0x (approved):

PR #3940, opened 2024-09-09
WGSL textureSample tests

greggman:
Add all the textureSample tests. cube and cube-array tests with derivatives are currently skipped or filtered out as the software rasterizer can't handle this case correctly or at least doesn't match too many GPUs. Rather than increase the tolerances I'm hoping to find something I can measure, like the mapping between derivative and mix-weight, that will give us some way to test per GPU.

There's a big change to the soft rasterizer to compute derivatives by setting up 2x2 pixels and computing the derivatives by the differences between the directions.

<hr>

### DIFF
--- a/src/webgpu/shader/execution/expression/call/builtin/textureSample.spec.ts
+++ b/src/webgpu/shader/execution/expression/call/builtin/textureSample.spec.ts
@@ -1,14 +1,22 @@
 export const description = `
 Samples a texture.
 
+- TODO: test cube maps with derivatives. These are currently filtered out via kCoord3DViewsForDerivativeTests
+        or by t.skip below.
+
 note: uniformity validation is covered in src/webgpu/shader/validation/uniformity/uniformity.spec.ts
 `;
 
 import { makeTestGroup } from '../../../../../../common/framework/test_group.js';
 import {
   isCompressedTextureFormat,
+  isDepthTextureFormat,
+  isEncodableTextureFormat,
   kCompressedTextureFormats,
+  kDepthStencilFormats,
   kEncodableTextureFormats,
+  kTextureFormatInfo,
+  textureDimensionAndFormatCompatible,
 } from '../../../../../format_info.js';
 import { TextureTestMixin } from '../../../../../gpu_test.js';
 
@@ -29,11 +37,13 @@ import {
   chooseTextureSize,
   isPotentiallyFilterableAndFillable,
   skipIfTextureFormatNotSupportedNotAvailableOrNotFilterable,
-  getDepthOrArrayLayersForViewDimension,
   getTextureTypeForTextureViewDimension,
   WGSLTextureSampleTest,
+  kCoord3DViewsForDerivativeTests,
+  isSupportedViewFormatCombo,
+  vec1,
+  generateTextureBuiltinInputs1D,
 } from './texture_utils.js';
-import { generateCoordBoundaries, generateOffsets } from './utils.js';
 
 const kTestableColorFormats = [...kEncodableTextureFormats, ...kCompressedTextureFormats] as const;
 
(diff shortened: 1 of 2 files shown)


src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
```diff
@@ -1558,73 +1638,141 @@ export function softwareRasterize<T extends Dimensionality>(
...
+      break;
+    case 'cube-array':
+      textureWGSL = depthTexture ? 'texture_depth_cube_array' : 'texture_cube_array<f32>';
+      coordWGSL = 'v.uvw';
+      break;
+    default:
+      unreachable();
+  }
+
+  // GBR of a depth texture is undefined. The software rasterizer
```
shrekshao, line 1690: Q: what is GBR?

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
```diff
@@ -1558,73 +1638,141 @@ export function softwareRasterize<T extends Dimensionality>(
...
-    layout: 'auto',
-    vertex: { module: shaderModule },
-    fragment: {
-      module: shaderModule,
-      targets: [{ format }],
-    },
-    primitive: { topology: 'triangle-strip' },
-  });
+  const id = `${format}:${code}`;
+  const pipelines = s_deviceToDrawTexturePipelines.get(device) || new Map();
```
shrekshao, line 1756: u-nit: `||`  is used here but s_deviceToDrawTextureShaderModules uses `??`. Can use `??` for both

src/webgpu/shader/execution/expression/call/builtin/textureSample.spec.ts
```diff
@@ -336,47 +389,112 @@ Parameters:
...
+      .combine('format', kTestableColorFormats)
+      .filter(t => {
+        const type = kTextureFormatInfo[t.format].color?.type;
+        const canPotentialFilter = type === 'float' || type === 'unfilterable-float';
+        // We can't easily put random bytes into compressed textures if they are float formats
+        // since we want the range to be +/- 1000 and not +/- infinity or NaN.
+        const isFillable = !isCompressedTextureFormat(t.format) || !t.format.endsWith('float');
+        return canPotentialFilter && isFillable;
+      })
+      .combine('viewDimension', kCoord3DViewsForDerivativeTests)
```
shrekshao, line 416: Side note: One thing I notice during gardening, maybe it's better to put `viewDimension` before `format` so that suppression can be more inclusive. E.g. cube map fails on a platform and it will be `viewDiemsnions='cube';*`
When format are put first the roller is likely to add suppressions for flaky tests scattered around the expectations.txt

However I don't know if changing this would make roller suffer for the next cts roll...

shrekshao (approved):
LGTM

(I don't really do the % / math for calculating coords. If they pass local test I assume they are good -_-)

greggman:
Hey @shrekshao, could you please re-review this? I changed it to use what you suggested which is to get the min/max values of all the texels sampled and check the value returned by the GPU is within those values.

Strangely, doing that exact thing didn't quite work. The GPUs would still return values slightly outside of that range. I'm not sure what's up with that. Maybe a precision issue? I tried +/-  2 / 255 thinking it might be an 8.8 fixed point issue but that didn't help and it was still failing for even a few float32 format tests on some GPUs. But, the values only seemed slightly off. (You can see this if you modify the code so `maxFractionalDiff` = 0 or 2 / 255).

In any case, I ended up allowing +/- maxFractionalDiff. That's 0 for int formats and various other things for other formats.

It looks like it's passing everywhere except a few places for actual bugs (Intel Linux seems completely broken.  OpenGL on Linux gets 0 for depth formats). [See run](https://dawn-review.googlesource.com/c/dawn/+/207435?checksResultsFilter=builtin,t&tab=checks). I when through each run and check expected fails too. A bunch of them cleared up. Hopefully that's all legit and the test is not too lenient.

The other things I did was add more weights where I query the GPU for how it mixes stuff and use that in the software render. That seems to help. It fails on Linux Intel for actual bugs.

I didn't change the non-derivative tests to use this min/max path even though it might be better as AFAIK they already pass as is.

Anyway, enough changed that I need to ask for a re-review. Sorry about that but thank you for the suggestion on min/max.

greggman:
Hey @shrekshao, could you please re-review this. I changed it to use your min/max suggestion.

The first pass I tried `maxFractionalDiff = 0`. This doesn't work with nearest filtering because only 1 sample is checked and it needs to be an exact match. I tried +/- 1 / 255 and +/- 2 / 255 and it still failed a few places so it's using the same `maxFractionalDiff` values as the existing tests which is 0 for sint/uint textures and otherwise, various values.

Anyway, the tests seem to pass except for legit failures ([see run](https://dawn-review.googlesource.com/c/dawn/+/207435?tab=checks))

Legit failures are (1) OpenGL on Linux (compat), reading depth textures returns 0 always (2) Linux Vulkan Intel seems completely broken. Otherwise, lots of stuff that was failing is no longer failing.

Also, I added some more weight queries which was an attempt to figure out what the GPU is doing and applying those to the software renderer. That code path is used by the non-derivative tests. I didn't change the non-derviative tests to use the min/max path as they are already passing. 

Anyway, with such a big change it seemed best to ask for a new review. Thanks!

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
```diff
@@ -142,25 +157,265 @@ export function skipIfTextureFormatNotSupportedNotAvailableOrNotFilterable(
...
+
+      @fragment fn fs(v: VSOutput) -> @location(0) vec4f {
+        let mipLevel = f32(v.ndx) / ${kMipGradientSteps};
+        let size = textureDimensions(tex);
+        let d = mix(0.125, 0.25, mipLevel) * 4.;
+        let u = f32(v.pos.x) * pow(2.0, mipLevel) / f32(size.x);
+        let g = mix(0.5, 1.0, mipLevel);
+
+        let ndx = v.ndx * ${kNumWeightTypes};
+        result[ndx + 0] = textureSampleLevel(tex, smp, vec2f(0.5), mipLevel).r;
```
shrekshao, line 415: does `mipLevel` need to be `mipLevelNum` (`f32(v.ndx)`)?

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
```diff
@@ -142,25 +157,265 @@ export function skipIfTextureFormatNotSupportedNotAvailableOrNotFilterable(
...
+        return VSOutput(vec4f(p, 0, 1), iNdx);
       }
-      @fragment fn fs(@builtin(position) pos: vec4f) -> @location(0) vec4f {
-        let mipLevel = floor(pos.x) / ${kMipGradientSteps};
-        result[u32(pos.x)] = textureSampleLevel(tex, smp, vec2f(0.5), mipLevel).r;
+
+      @fragment fn fs(v: VSOutput) -> @location(0) vec4f {
+        let mipLevel = f32(v.ndx) / ${kMipGradientSteps};
+        let size = textureDimensions(tex);
+        let d = mix(0.125, 0.25, mipLevel) * 4.;
```
shrekshao, line 410: nit: `d` seems unused

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
```diff
@@ -142,25 +157,265 @@ export function skipIfTextureFormatNotSupportedNotAvailableOrNotFilterable(
...
-        return vec4f(pos[vNdx], 0, 1);
+        let p = pos[vNdx];
+        return VSOutput(vec4f(p, 0, 1), iNdx);
       }
-      @fragment fn fs(@builtin(position) pos: vec4f) -> @location(0) vec4f {
-        let mipLevel = floor(pos.x) / ${kMipGradientSteps};
-        result[u32(pos.x)] = textureSampleLevel(tex, smp, vec2f(0.5), mipLevel).r;
+
+      @fragment fn fs(v: VSOutput) -> @location(0) vec4f {
+        let mipLevel = f32(v.ndx) / ${kMipGradientSteps};
```
shrekshao, line 408: I'm confused here.

say we have
```
let mipLevelNum = f32(v.ndx);
let mipLevel = mipLevelNum / ${kMipGradientSteps};  // maybe rename to mipLevelMix?
```

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
```diff
@@ -142,25 +157,265 @@ export function skipIfTextureFormatNotSupportedNotAvailableOrNotFilterable(
...
+        let size = textureDimensions(tex);
+        let d = mix(0.125, 0.25, mipLevel) * 4.;
+        let u = f32(v.pos.x) * pow(2.0, mipLevel) / f32(size.x);
```
shrekshao, line 411: does `mipLevel` need to be `mipLevelNum` (`(f32(v.ndx)`)?

shrekshao (commented):
I don't know if I understand shaders in `queryMipGradientValuesForDevice`. I had some comments. Please take a look

shrekshao:
Am I understanding the queryMipGradientValuesForDevice shader correctly (based on my questions)?
Do you still have changes to make, or should I review the current files?

greggman:
Your understanding is correct but, given it's been a few days since I asked for this review I've made so new many changes that I guess we should close this and start a new one with the latest.

PR #3942, opened 2024-09-10
Add the largest float representable as integer (signed/unsigned) to F32 testing (execution) set

petermcneeleychromium:
This change ensures that the behavior of the u32(f32) and i32(f32) conversions are fully conformance tested.

This behavior was fixed for the tint AST
see: crbug.com/362659150


petermcneeleychromium:
Ryan has suggested

...(
   [...linearRange(kBit.f32.positive.min, kBit.f32.positive.max, counts.pos_norm),
    // Largest float as signed integer 
    0x4effffff,
    // Largest float as unsigned integer
    0x4f7fffff].sort((n1,n2) => n1 > n2))

Which i will try and have to rerun the execution test

petermcneeleychromium:
FYI the tests in question
'webgpu:shader,execution,expression,unary,*'

zoddicus (approved):
Overaly lgtm

The caches need to be regenerated, which is why the checks are failing.

Since this introduces two additional values to the default f32 range, this is going to cause additional test cases to be generated, so you will have to watch when this rolls into chromium/dawn to make sure that it doesn't cause timeouts.

petermcneeleychromium:
You want me to regen the caches? I can do so. I just didnt know if it was automagic.

zoddicus:
> You want me to regen the caches? I can do so. I just didnt know if it was automagic.

I am pretty sure you will have to, since this cause new test cases to generated, thus invalidating the cache. I think you just need to run `npx grunt run:generate-cache` to update them (it has been a while since I have added tests to the CTS)

petermcneeleychromium:
I force updated all the gen via:
npm run gen_cache --force .

petermcneeleychromium:
we need to decide what to do about pos_norm count.  The unit tests will complain if there are additional numbers.

petermcneeleychromium:
We could maybe choose to add these special numbers if pos_norm > 2 or some other number.
We might also want to reduce the number of generated pos_norms as well by 2

zoddicus:
> We could maybe choose to add these special numbers if pos_norm > 2 or some other number. We might also want to reduce the number of generated pos_norms as well by 2

I think the high road solution we want here is to only add in these special values if pos_norm is >= 4, and reduce the number of pos_nums generated by the linear range by 2 when they are added. Any related unittests will need to be updated.

My thinking on this is;
If I pass in pos_norm == 0, I expect no pos_norms
If I pass in pos_norm == 1, I expect one of the end points
If I pass in pos_norm == 2, I expect both end points.
It is only at pos_norms == 3 that the expectation is the interior values start show up
I am suggesting using pos_norms == 4 as the limit for adding these values in, since there are two of them, though you could pick one to insert at pos_norms == 3 and the other at 4, but that feels a bit baroque.

(An aside, I was wrong about sort not getting run per call, I looked at the code again and realized the const is in a function, the github view made me think it wasn't. Anyway, I think hitting the sub array with sort per call is fine, since we are already doing array building in the function).

zoddicus (approved):
:+1:

PR #3947, opened 2024-09-11
Minor fix for extra comma when specifying access mode

petermcneeleychromium:
This extra comma should be consider an error in the case where the mode is specified 
crbug.com/366000875


dj2:
According to the spec, I believe the trailing `,` should be premitted:

`'var' ( _template_args_start [expression ( ',' expression )* ',' ? _template_args_end )?`

From https://www.w3.org/TR/WGSL/#recursive-descent-syntax-variable_decl

There is no other expansion for `var` and that always permits a trailing `,`, so this seems more like it's an error in Tint?

petermcneeleychromium:
Yes it looks like you are right. We allow extra ',' here. I will change tint.

PR #3964, opened 2024-09-24
Initial commit for subgroupInclusive add/mul cts

petermcneeleychromium:
Implementation here is very similar to existing subgroup operations so the effort is minimal

crbug.com/361330160

### DIFF
--- a/src/webgpu/shader/execution/expression/call/builtin/subgroupAdd.spec.ts
+++ b/src/webgpu/shader/execution/expression/call/builtin/subgroupAdd.spec.ts
@@ -1,5 +1,5 @@
 export const description = `
-Execution tests for subgroupAdd and subgroupExclusiveAdd
+Execution tests for subgroupAdd, subgroupExclusiveAdd, and subgroupInclusiveAdd
 
 Note: There is a lack of portability for non-uniform execution so these tests
 restrict themselves to uniform control flow.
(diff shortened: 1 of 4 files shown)


src/webgpu/shader/execution/expression/call/builtin/subgroupAdd.spec.ts
```diff
@@ -269,6 +276,10 @@ function checkPredicatedAddition(
           expected += j;
         }
       }
+
+      if (operation === 'subgroupInclusiveAdd') {
```
alan-baker, line 280: This feels like it would be better placed as part of the bounds calculations at line 273.

src/webgpu/shader/execution/expression/call/builtin/subgroupMul.spec.ts
```diff
@@ -274,20 +281,27 @@ g.test('fragment').unimplemented();
...
+        return (id % 4) + 1;
+      };
       const bound = operation === 'subgroupMul' ? size : id;
       for (let j = 0; j < bound; j++) {
         if (filter(j, size)) {
-          expected *= (j % 4) + 1;
+          expected *= valueModFun(j);
         }
       }
+      if (operation === 'subgroupInclusiveMul') {
```
alan-baker, line 302: Similar here.

alan-baker (approved):
My preference is to have the bound calculation in one place, but I'll leave the choice to you.

PR #3984, opened 2024-10-07
Add textureSampleCompare/Level WGSL tests

greggman:
The tests fill depth textures with random values
except, if the comparison mode is 'equal' or 'not-equal' then we only put 0.0, 0.5, 1.0 in the texture. depthRef for all tests are 0.0, 0.5, or 1.0 so the comparisons should produce valid tests.

<hr>

### DIFF
--- a/src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
+++ b/src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
@@ -37,7 +37,7 @@ import {
   TexelComponent,
   TexelRepresentationInfo,
 } from '../../../../../util/texture/texel_data.js';
-import { TexelView } from '../../../../../util/texture/texel_view.js';
+import { PerPixelAtLevel, TexelView } from '../../../../../util/texture/texel_view.js';
 import { createTextureFromTexelViews } from '../../../../../util/texture.js';
 import { reifyExtent3D } from '../../../../../util/unions.js';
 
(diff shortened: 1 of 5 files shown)


src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
```diff
@@ -732,11 +732,60 @@ export function appendComponentTypeForFormatToTextureType(base: string, format:
...
+  const rep = kTexelRepresentationInfo[info.format as EncodableTextureFormat];
+  const size = reifyExtent3D(info.size);
+
+  const isEqualNotEqual = comparison === 'equal' || comparison === 'not-equal';
+
+  // for equal and not-equal we just want to test 0, 0.5, and 1
+  // for everything else we want 0 to 1
+  const fixedValues = [0, 0.5, 1, 1];
+  const format = isEqualNotEqual
+    ? (norm: number) => fixedValues[(norm / 3) | 0]
```
shrekshao, line 758: norm will be random number number between [0, 1], This will only return fixedValues[0]?
I think here you want something like `fixedValues[(norm / 0.3333333) | 0]`?
more formally something like  `fixedValues[Math.floor(norm * fixedValues.length)]`

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
```diff
@@ -3562,7 +3635,7 @@ export function generateSamplePointsCube(
       mipLevel,
       arrayIndex: args.arrayIndex ? makeRangeValue(args.arrayIndex, i, 2) : undefined,
       bias,
-      depthRef: args.depthRef ? makeRangeValue({ num: 1, type: 'f32' }, i, 5) : undefined,
+      depthRef: args.depthRef ? makeRandValue({ num: 3, type: 'u32' }, i, 5) / 2 : undefined,
```
shrekshao, line 3638: Q: I'm unsure about this line: we have random texel values between [0, 1]. is it we want depthRef being some random number of [0, 1.5]?

Why is type 'u32' but not 'f32' for `textureSampleCompare` and other returns f32?

shrekshao (approved):
LGTM with comments

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
greggman: `fixedValues[Math.floor(norm * fixedValues.length)]` doesn't work because `norm` goes from 0 to 1, not 0 to 0.99999999999 so when norm is 1 it will pull the last number. 

I changed it to `fixedValues[(norm * (fixedValues.length - 1)) | 0]`

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
greggman: We use `u32` so `makeRandValue` returns an integer `[0, (num -1]]` (so 0, 1, 2) then dividing by 2 gives 0, 0.5, 1. Added a comment as to why

PR #3991, opened 2024-10-10
Refactor texture-builins for compat no storage

greggman:
compat is allowed to have no storage buffers but the texture-builin tests were relying on storage buffers both for inputs and outputs.

For inputs, switching to a uniform buffer is fine. There are 50 calls with at most 5 parameters each aligned to 16 bytes so that's 4000 bytes which fits within the minimum uniform block size.

For outputs, switching to writing output to a texture via the fragment shader works but in order not to have to change how derivatives work, instead we render
1 instance at a time and use setViewport to choose which texel to write to. We were using @builtin(position) in the fragment shader and expecting it to be 0.5, 0.5 but since we're writing to different fragments now we have to subtract the instance index (v.ndx) from position.x to get it back to 0.5, 0.5.


shrekshao (approved):
LGTM

PR #3992, opened 2024-10-10
Get per stage GPU mix weights on demand

greggman:
It used to be that mix weights for all stages were queried at initialization time. The problem is, on Intel Mac, the compute stage always fails. That means the majority of texture tests just would not run on Intel Mac, at least on Chrome and Firefox at the moment.

Changing the code to load them per stage on demand means the texture builtin tests can run for compute and fragment stages on Intel Mac.

You could argue this doesn't need to be fixed since Chrome and Firefox should fix their texture code on Intel Mac but I think at the moment there are a bunch more high priority things being worked on so it's best not to lose coverage for compute and fragment shaders.


shrekshao (approved):
LGTM

PR #3995, opened 2024-10-11
Include depth textures in more tests

greggman:
Depth textures were excluded originally because there was no way to fill one via copyBufferToTexture. We added a way to fill them later which was used for the textureLoad tests. Now going back and enabling them for other tests.

Note that might be better to move minFilter to a case instead of a subcase but that would break all expectations in CQs so it seemed best to filter these in the test itself.


shrekshao (approved):
LGTM

PR #4010, opened 2024-10-21
Texture builtins: move addressMode? to case

greggman:
addressMode is broken on many GPUs. Moving this from a subcase to a case lets us separate out those failures.

I opted to put the parameters in minFilter, addressMode, offset order which means for some builtins I moved minFilter from subcase to case.

This turned into a massive change because the query strings got too long. so had to change

* addressModeX -> modeX
* filter to filt
* clamp-to-edge -> c
* repeat -> r
* mirror-repeat -> m
* viewDimension -> dim
* fragment -> f
* compute -> c
* vertex -> v

Other option would be to move all the tests from
`webgpu:shader,execution,expression,call,builtin` to `webgpu:texture` though I'm not sure that would be enough on its own

Note: this takes `webgpu:shader,execution,expression,call,builtin,*` from 21738 cases to 113546 cases

### DIFF
--- a/src/webgpu/shader/execution/expression/call/builtin/textureGather.spec.ts
+++ b/src/webgpu/shader/execution/expression/call/builtin/textureGather.spec.ts
@@ -32,7 +32,6 @@ import {
   kDepthStencilFormats,
   kEncodableTextureFormats,
 } from '../../../../../format_info.js';
-import { kShaderStages } from '../../../../validation/decl/util.js';
 
 import {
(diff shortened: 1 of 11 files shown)


kainino0x (approved):
lgtm! Sorry for the dumb restriction, it originally comes from when we were using WPT to run these tests in Chromium, which we aren't anymore, but I'm pretty sure that at least WebKit runs into similar issues for the same reason (though file path limits on Mac are surely not as bad as Windows)

PR #4011, opened 2024-10-21
index.html: Remove CSS attribute that did nothing and now breaks layout

kainino0x:
It starts breaking with the current Chrome Canary.

See https://crbug.com/374115881


greggman (approved):
lgtm

PR #4014, opened 2024-10-28
Fix textureSample:sampled_2d_coords offset

greggman:
textureSample:sampled_2d_coords had offset always on


shrekshao (approved):
LGTM (Sorry I must slip this away initial review)

PR #4016, opened 2024-10-29
Test fixes for alignment validation landing

petermcneeleychromium:
Issue:
crbug.com/375467276

Change (correct validation) that introduced this issue
https://dawn-review.googlesource.com/c/dawn/+/212315


petermcneeleychromium:
You mentioned this being part of a spec discussion for the align(1)

src/webgpu/shader/validation/shader_io/align.spec.ts
```diff
@@ -13,7 +13,10 @@ const kTests = {
   },
   one: {
     src: '@align(1)',
-    pass: true,
+    pass: false,
+    // EXCEPTION: Error: Unexpected validation error occurred:
+    // Error while parsing WGSL: :6:10 error: alignment must be a
+    // multiple of '4' bytes for the 'uniform' address space @align(1) a: i32,
```
dneto0, line 19: yes, this LGTM.

src/webgpu/shader/validation/shader_io/align.spec.ts
```diff
@@ -45,6 +48,14 @@ const kTests = {
   },
   const_expr: {
     src: '@align(i_val + 4 - 6)',
+    pass: false,
+    // EXCEPTION: Error: Unexpected validation error occurred:
```
dneto0, line 52: note to self:  i_val is a const value of 4.  So this computes to 2, and should be invalid.
LGTM.

src/webgpu/shader/validation/shader_io/align.spec.ts
```diff
@@ -45,6 +48,14 @@ const kTests = {
...
+    // Error while parsing WGSL: :6:10 error: alignment must be a
+    // multiple of '4' bytes for the 'uniform' address space
+    // @align(i_val + 4 - 6) a: i32
+  },
+  const_expr_2: {
+    src: '@align(i_val + 8 - 4)',
```
dneto0, line 58: This computes to 4 + 8 - 4 = 8, which is valid.
LGTM.

src/webgpu/shader/validation/shader_io/align.spec.ts
```diff
@@ -185,7 +196,7 @@ g.test('required_alignment')
         { name: 'mat3x4<f16>', storage: 8, uniform: 8 },
         { name: 'mat4x4<f16>', storage: 8, uniform: 8 },
         { name: 'array<vec2<i32>, 2>', storage: 8, uniform: 16 },
-        { name: 'array<vec4<i32>, 2>', storage: 8, uniform: 16 },
+        { name: 'array<vec4<i32>, 2>', storage: 16, uniform: 16 },
```
dneto0, line 199: Right.  `alignof(vec4<i32>)` is 16.

dneto0 (approved):
Thanks!

src/webgpu/shader/validation/shader_io/align.spec.ts
dneto0: whoops, no this is wrong.

The real check is whether the offset if incorrect.

petermcneeleychromium: Based on our discussion it sounds like this change is correct.
Shall i remove the extraneous comment in the source

petermcneeleychromium:
Some of these tests only consider the case of storage and uniform and not workgroup
 .combine('address_space', ['storage', 'uniform'])

src/webgpu/shader/validation/shader_io/align.spec.ts
dneto0: ok, after a bunch of discussion, we agreed to change Tint to be more in line with Naga and WebKit.

crbug.com/379760170  is the Tint change.

But also clarify the spec: https://github.com/gpuweb/gpuweb/pull/4978

dneto0: LGTM.  

Given the proposed spec change, additional tests will be needed.  For followup.

dneto0 (approved):

PR #4020, opened 2024-10-30
Refactor to not render to rgba32float

greggman:
Compat will not have rgba32float as a render target support except as an optional feature so refactor to not use it.


src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
```diff
@@ -4510,23 +4510,33 @@ function createTextureCallsRunner<T extends Dimensionality>(
...
-@fragment fn fsVertex(v: VOut) -> @location(0) ${returnType} {
-  return v.result;
+@fragment fn fsVertex(v: VOut) -> @location(0) vec4u {
+  return bitcast<vec4u>(v.result);
 }
 `
       : stage === 'fragment'
       ? `
 // --------------------------- fragment stage shaders --------------------------------
+fn convertToVec4u(v: ${returnType}) -> vec4u {
```
shrekshao, line 4520: Is this called? is it used to convert rgba32sint to rgba32uint?

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
```diff
@@ -4470,7 +4470,7 @@ function createTextureCallsRunner<T extends Dimensionality>(
   const samplerType = isCompare ? 'sampler_comparison' : 'sampler';
 
   const renderTarget = t.createTextureTracked({
-    format: resultFormat,
+    format: 'rgba32uint',
```
shrekshao, line 4473: I think result format can be either `rgba32sint` and `rgba32uint`?

shrekshao (commented):
Questions listed. (My mind is unclear)

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
greggman: thanks. That's not needed. Deleted

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
greggman: I just made it always write to `rgba32uint` so the 3 paths are

* `rgba32uint` -> `rgba32uint`
* `rgba32sint` -> `rgba32uint`
* `rgba32float` -> `rgba32uint`

shrekshao (approved):

PR #4023, opened 2024-10-31
Add tests for GPUDevice.adapterInfo

beaufortfrancois:
This PR adds simple tests for the adapterInfo GPUDevice member. It can be tested in a Chromium build patched with https://chromium-review.googlesource.com/c/chromium/src/+/5980752. It is available behind the "Unsafe WebGPU support" flag at `chrome://flag#enable-unsafe-webgpu` as the spec PR https://github.com/gpuweb/gpuweb/pull/4809 has not landed yet.

image


src/webgpu/api/operation/device/info.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+    const adapter = await gpu.requestAdapter();
+    assert(adapter !== null);
+
+    const device = await t.requestDeviceTracked(adapter);
+    assert(device !== null);
+
+    assert(device.adapterInfo instanceof GPUAdapterInfo);
+    assert(adapter.info instanceof GPUAdapterInfo);
+
+    for (const k of Object.keys(GPUAdapterInfo.prototype)) {
```
kainino0x, line 28: I think we should actually require `device.adapterInfo === adapter.info`. I've proposed it here:
https://github.com/gpuweb/gpuweb/pull/4809#discussion_r1826388788

src/webgpu/api/operation/device/info.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+    const gpu = getGPU(t.rec);
+    const adapter = await gpu.requestAdapter();
+    assert(adapter !== null);
+
+    const device = await t.requestDeviceTracked(adapter);
+    assert(device !== null);
+
+    assert(device.adapterInfo instanceof GPUAdapterInfo);
+    assert(adapter.info instanceof GPUAdapterInfo);
+
```
kainino0x, line 27: Should test the `[SameObject]` property, i.e.:

`t.expect(device.adapterInfo === device.adapterInfo);`

This should be tested for `adapter.info` too if it is not already.

src/webgpu/api/operation/device/info.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+`;
+
+import { Fixture } from '../../../../common/framework/fixture.js';
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { getGPU } from '../../../../common/util/navigator_gpu.js';
+import { assert, objectEquals } from '../../../../common/util/util.js';
+
+export const g = makeTestGroup(Fixture);
+
+g.test('device_adapter_info')
```
kainino0x, line 12: nit: test name is redundant with filename; maybe `'matches_adapter'`

or if merged into `adapter,info`, `device_matches_adapter`

src/webgpu/api/operation/device/info.spec.ts
kainino0x, line 1: Logically it would make sense to group this with `operation/adapter/info.spec.ts`. To avoid a test rename how about just adding these tests there? Even though they're not strictly on the adapter, it's still testing adapter info, so IMO it's fine.

src/webgpu/api/operation/device/info.spec.ts
beaufortfrancois: Done

src/webgpu/api/operation/device/info.spec.ts
beaufortfrancois: I'll be waiting for https://github.com/gpuweb/gpuweb/pull/4809#pullrequestreview-2411155432 to be resolved.

src/webgpu/api/operation/device/info.spec.ts
kainino0x: This can already be tested at least for adapter.info, as the spec already says `[SameObject]`.

beaufortfrancois: You're right!

src/webgpu/api/operation/adapter/info.spec.ts
```diff
@@ -39,3 +39,42 @@ g.test('adapter_info')
...
+    const adapter = await gpu.requestAdapter();
+    assert(adapter !== null);
+
+    const device = await t.requestDeviceTracked(adapter);
+    assert(device !== null);
+
+    assert(device.adapterInfo instanceof GPUAdapterInfo);
+    assert(adapter.info instanceof GPUAdapterInfo);
+
+    for (const k of Object.keys(GPUAdapterInfo.prototype)) {
```
kainino0x, line 74: Landed the spec; didn't make adapter and device use the same object, so we should test that `adapter.info !== device.adapterInfo`.

src/webgpu/api/operation/adapter/info.spec.ts
```diff
@@ -39,3 +39,42 @@ g.test('adapter_info')
...
+  )
+  .fn(async t => {
+    const gpu = getGPU(t.rec);
+    const adapter = await gpu.requestAdapter();
+    assert(adapter !== null);
+
+    const adapterInfo1 = adapter.info;
+    const adapterInfo2 = adapter.info;
+    t.expect(adapterInfo1 === adapterInfo2);
+  });
```
kainino0x, line 56: Landed the spec, we should now have this test for device.adapterInfo too.

src/webgpu/api/operation/adapter/info.spec.ts
```diff
@@ -39,3 +39,42 @@ g.test('adapter_info')
...
+
+    const device = await t.requestDeviceTracked(adapter);
+    assert(device !== null);
+
+    assert(device.adapterInfo instanceof GPUAdapterInfo);
+    assert(adapter.info instanceof GPUAdapterInfo);
+
+    for (const k of Object.keys(GPUAdapterInfo.prototype)) {
+      t.expect(
+        objectEquals(device.adapterInfo[k], adapter.info[k]),
```
kainino0x, line 76: Since browsers are allowed to make last-moment decisions about the contents of GPUAdapterInfo, I'd like to have a few variants on this test which access things in the following order:

- device.adapterInfo, adapter.info, device.adapterInfo[k1], adapter.info[k1], device.adapterInfo[k2], adapter.info[k2] (this is what you have now)
- same but flip adapter and device
- device.adapterInfo, device.adapterInfo[k1], device.adapterInfo[k2], adapter.info, adapter.info[k1], adapter.info[k2]
- same but flip adapter and device

(Each variant needs to get its own adapter and device, not reuse them.)

Just to be paranoid.

beaufortfrancois: @kainino0x `device.adapterInfo, device.adapterInfo[k1], device.adapterInfo[k2], adapter.info, adapter.info[k1], adapter.info[k2]` is a bit messy when coding it. Are we sure we want to be that paranoid for such as simple thing?

src/webgpu/api/operation/adapter/info.spec.ts
beaufortfrancois: The spec doesn't say they have to be different object, so I wouldn't test that `adapter.info !== device.adapterInfo`. Note that Chromium experimental implementation actually returns the same object: https://chromium-review.googlesource.com/c/chromium/src/+/5980752/3/third_party/blink/renderer/modules/webgpu/gpu_device.cc

src/webgpu/api/operation/adapter/info.spec.ts
beaufortfrancois: Will do

src/webgpu/api/operation/adapter/info.spec.ts
kainino0x: It does implicitly require them to be different objects, because a new one is created the first time each of them is accessed, there's no opportunity to return the same one.

src/webgpu/api/operation/adapter/info.spec.ts
kainino0x: It will be a little messy to write once but I think it is worth it. Just in case any browser ever does fingerprinting mitigations at this level, it could be very weird if they got this wrong.

To do that one I think you can basically do:

```
const deviceInfo: unknown[] = [];
for (const k of Object.keys(GPUAdapterInfo.prototype)) deviceInfo.push(device.adapterInfo[k]);
const adapterInfo: unknown[] = [];
for (const k of Object.keys(GPUAdapterInfo.prototype)) deviceInfo.push(adapter.info[k]);
//objectEquals(deviceInfo, adapterInfo)
```

kainino0x: The potential complexity of this could be an argument in favor of having both return the same object... but I don't think it's really worth changing the spec further.

src/webgpu/api/operation/adapter/info.spec.ts
beaufortfrancois: I've updated this PR.
Do you know by any chance how to return a copy of `adapter_->info()` in Chromium?

src/webgpu/api/operation/adapter/info.spec.ts
beaufortfrancois: Done

src/webgpu/api/operation/adapter/info.spec.ts
kainino0x: We should construct the `adapterInfo` when we construct `GPUDevice`, the same way we do for `info` when we construct `GPUAdapter`. The initialization will be the same so we should take this code, move it into a GPUAdapter method called something like `CreateAdapterInfoForAdapter()`, have it return the GPUAdapterInfo object, and then call it from both GPUAdapter::GPUAdapter and GPUDevice::GPUDevice.
https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/webgpu/gpu_adapter.cc;l=198-210;drc=0afc9ac9afcaab79fc54299039f4d27abf3a086d

src/webgpu/api/operation/adapter/info.spec.ts
```diff
@@ -39,3 +39,90 @@ g.test('adapter_info')
       `adapterInfo.device should be a normalized identifier. But it's '${adapterInfo.device}'`
     );
   });
+
+g.test('same_object')
+  .desc(
+    `
+GPUAdapter.info and GPUDevice.adapterInfo provide the same object each time it's accessed`
```
kainino0x, line 46:
```suggestion
GPUAdapter.info and GPUDevice.adapterInfo provide the same object each time they're accessed,
but different objects from one another.`
```

src/webgpu/api/operation/adapter/info.spec.ts
```diff
@@ -39,3 +39,90 @@ g.test('adapter_info')
...
+    const deviceAdapterInfo2 = device.adapterInfo;
+    t.expect(deviceAdapterInfo1 === deviceAdapterInfo2);
+
+    t.expect(adapter.info !== device.adapterInfo);
+  });
+
+g.test('device_matches_adapter')
+  .desc(
+    `
+Test that GPUDevice.adapterInfo matches GPUAdapter.info`
```
kainino0x, line 70:
```suggestion
Test that GPUDevice.adapterInfo matches GPUAdapter.info. Cases access the members in
different orders to make sure that they are consistent regardless of the access order.`
```

src/webgpu/api/operation/adapter/info.spec.ts
```diff
@@ -39,3 +39,90 @@ g.test('adapter_info')
...
+    assert(adapter !== null);
+
+    const device = await t.requestDeviceTracked(adapter);
+    assert(device !== null);
+
+    const deviceInfo: unknown[] = [];
+    const adapterInfo: unknown[] = [];
+
+    type GPUAdapterInfoKey = 'vendor' | 'architecture' | 'device' | 'description';
+    const GPUAdapterInfoKeys = Object.keys(GPUAdapterInfo.prototype) as GPUAdapterInfoKey[];
```
kainino0x, line 89: We have a helper to make this better, I'll push a commit to use it.

kainino0x (approved):

kainino0x:
Tested my changes in Canary with --enable-unsafe-webgpu, still works as expected.

kainino0x (approved):

beaufortfrancois:
`import { keysOf } from '../../../../common/util/data_tables.js';` is useful. Thanks for improving the PR!

PR #4024, opened 2024-10-31
Fix fragment shader subgroup builtin io test

jzm-intel:
This PR fix the expectation of fragment shader `subgroup_invocation_id`, which can be assigned to inactivate invocations between active ones and therefore go larger than active subgroup invocations number but still smaller than subgroup size. This PR also fix the draw call for fragment subgroup tests. With this PR, tests can passed on Intel devices.

### DIFF
--- a/src/webgpu/shader/execution/shader_io/fragment_builtins.spec.ts
+++ b/src/webgpu/shader/execution/shader_io/fragment_builtins.spec.ts
@@ -1537,7 +1537,7 @@ function checkSubgroupSizeConsistency(
  * @param height The framebuffer height
  * @param checker A functor to check the framebuffer values
  */
- async function runSubgroupTest(
+async function runSubgroupTest(
   t: FragmentBuiltinTest,
   format: GPUTextureFormat,
   fsShader: string,
@@ -1818,7 +1818,7 @@ fn fsMain(
   var repId = atomicAdd(&counter, 1);
   repId = subgroupBroadcast(repId, 0);
 
-  return vec4u(id, ballotSize, repId, 0);
+  return vec4u(id, sg_size, ballotSize, repId);
 }`;
 
     await runSubgroupTest(
--- a/src/webgpu/shader/execution/shader_io/fragment_builtins.spec.ts
+++ b/src/webgpu/shader/execution/shader_io/fragment_builtins.spec.ts
@@ -1454,6 +1454,76 @@ function popcount(input: number): number {
 }
 
 /**
+ * Checks subgroup_size builtin value consistency.
+ *
+ * The builtin subgroup_size is not assumed to be uniform in fragment shaders.
+ * Therefore, this function checks the value is a power of two within the device
+ * limits and that the ballot size is less than the stated size.
+ * @param data An array of vec4u that contains (per texel):
+ *             * builtin value
+ *             * ballot size
+ *             * comparison to other invocations
+ *             * 0
+ * @param format The texture format for data
+ * @param min The minimum subgroup size from the device
+ * @param max The maximum subgroup size from the device
+ * @param width The width of the framebuffer
+ * @param height The height of the framebuffer
+ */
+function checkSubgroupSizeConsistency(
+  data: Uint32Array,
+  format: GPUTextureFormat,
+  min: number,
+  max: number,
+  width: number,
+  height: number
+): Error | undefined {
+  const { blockWidth, blockHeight, bytesPerBlock } = kTextureFormatInfo[format];
+  const blocksPerRow = width / blockWidth;
+  // Image copies require bytesPerRow to be a multiple of 256.
+  const bytesPerRow = align(blocksPerRow * (bytesPerBlock ?? 1), 256);
+  const uintsPerRow = bytesPerRow / 4;
+  const uintsPerTexel = (bytesPerBlock ?? 1) / blockWidth / blockHeight / 4;
+
+  for (let row = 0; row < height; row++) {
+    for (let col = 0; col < width; col++) {
+      const offset = uintsPerRow * row + col * uintsPerTexel;
+      const builtinSize = data[offset];
+      const ballotSize = data[offset + 1];
+      const comparison = data[offset + 2];
+      if (builtinSize === 0) {
+        continue;
+      }
+
+      if (popcount(builtinSize) !== 1) {
+        return new Error(`Subgroup size '${builtinSize}' is not a power of two`);
+      }
+
+      if (builtinSize < min) {
+        return new Error(`Subgroup size '${builtinSize}' is less than minimum '${min}'`);
+      }
+      if (max < builtinSize) {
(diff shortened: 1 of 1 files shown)


src/webgpu/shader/execution/shader_io/fragment_builtins.spec.ts
```diff
@@ -1726,31 +1728,57 @@ function checkSubgroupInvocationIdConsistency(
...
+  >();
   for (let row = 0; row < height; row++) {
     for (let col = 0; col < width; col++) {
       const offset = uintsPerRow * row + col * uintsPerTexel;
       const id = data[offset];
-      const size = data[offset + 1];
-      const repId = data[offset + 2];
+      const sgSize = data[offset + 1];
+      const ballotSize = data[offset + 2];
+      const repId = data[offset + 3];
```
alan-baker, line 1742: This seems incorrect for the shader that is run. The fragment shader returns `vec4u(id, ballotSize, repId, 0)`. So:
* offset + 0: id (this matches)
* offset + 1: ballotSize (this doesn't match)
* offset + 2: repId (this doesn't match)

Because this grabs `repId` from `offset + 3` it seems that all values will be zero, which will hit the continue statement and not perform the other checks. Is there a missing shader modification here?

alan-baker (changes requested):
The changes to the checker code look sound, but they don't seem to match the code that is run. If we can resolve that this definitely looks more robust.

src/webgpu/shader/execution/shader_io/fragment_builtins.spec.ts
jzm-intel: Oops... Sorry the shader change is missed. Fixed it and the test pass on Intel UHD630.

jzm-intel:
Tried to assert all outputs are zero for fragment that `repId === 0`, I found it was not true. Investigated the output of framebuffer size 15x5 and 16x16, the `repId` of some activate fragments could be duplicated, `0` or even `4294967295`. We can also see this `repId === 0` cases in the screenshot provided above, like for fragment (12, 1) and its neighbors.

I thought this is due to currently we get the `repId` from `subgroupBroadcast` from invocation 0, which can be actually inactive and the broadcasted result could be undefined.

I think I need to change the way we get `repId` for subgroups, and I will do it tomorrow.

jzm-intel:
Things could be tricky since we don't know which invocation will be active in compile time, and `subgroupBroadcast` requires a constant subgroup invocation id to read from. A subgroup function that broadcast from the lowest active invocation could solve the issue, but I thought we don't have that for now?

alan-baker:
> Things could be tricky since we don't know which invocation will be active in compile time, and `subgroupBroadcast` requires a constant subgroup invocation id to read from. A subgroup function that broadcast from the lowest active invocation could solve the issue, but I thought we don't have that for now?

There is `subgroupBroadcastFirst`, but unfortunately APIs aren't consistent about whether helpers participate in subgroup operations. Some say yes, some say no. For the built-in function tests I tried making the fragment tests draw a single full screen triangle and skipping the last row and column as potential helpers. That might help here, but it may still assume the upper left pixels are assigned lower ids. I'll think on this a bit more.

jzm-intel:
A terrible but workable (should it?) solution is:
* Fragment shader directly store the output to storage buffer, using the built-in input `position` to get the fragment position `p: u32 = rowOfFragment * framebufferColsPerRow + colOfFragment` and compute the offset `o = p * strideOfOutput` for output buffer. Such position `p` is unique for each fragment.
* To know which fragments are of the same subgroup, we make a large storage buffer of size `[fragmentNumber] * [maximium subgroup size == 128] * sizeof(u32)`, i.e. we have a storage buffer of array `broadcastedP: array<array<maxSubgroupSize, u32>, [fragmentNumber]>` , and each fragment store the subgroup-broadcasted `p` from all subgroup invocations (oops still `subgroupBroadcast` from inactive invocation... but we can use ballot result to filter the results from inactive later):
```
if (sgSize >= 4) {
    broadcastedP[p][0] = subgroupBroadcast(p, 0);
    broadcastedP[p][1] = subgroupBroadcast(p, 1);
    broadcastedP[p][2] = subgroupBroadcast(p, 2);
    broadcastedP[p][3] = subgroupBroadcast(p, 3);
}
if (sgSize >= 8) {
    broadcastedP[p][4] = subgroupBroadcast(p, 4);
    ...
}
...
```
And also store the subgroup ballot result of each fragment to indicate which results comes from inactive invocation and should be ignored.
* Then on the CPU side we can use the filtered `broadcastedP` to tell which fragments are in the same subgroup of fragment `p`, i.e. those of `broadcastedP[p][0..sgSize-1]`.

jzm-intel:
> oops still `subgroupBroadcast` from inactive invocation... but we can use ballot result to filter the results from inactive later

This still rely on subgroup ballot to filter the result, and it would be wrong if helper lane also take part in the ballot (will it?) but don't compute the correct `p`, so it seems no better than `subgroupBroadcastFirst(repId)`?

alan-baker:
Here's an alternative maybe:

```
@fragment
fn fsMain(
  @builtin(position) pos : vec4f,
  @builtin(subgroup_invocation_id) id : u32,
  @builtin(subgroup_size) sgSize : u32
) -> vec4u {
  var error = 0;
  for (var i = 0; i < sgSize; i++) {
    let idBallot = subgroupBallot(id == i);
    let count = countOneBits(idBallot);
    let sum = count.x + count.y + count.z + count.w;
   error += select(1, 0, sum == 1);
  }
  return vec4u(id, sgSize, error, 0);
}
```

Then we can check that for all pixels, `id < sgSize` and `error  === 0`. I think we could even hardcode the loop bound to 128 in case we're worried sgSize is implemented incorrectly. 

What do you think?

jzm-intel:
This would check that each `id < sgSize` is used once and only once for all active invocations (if only the active ones take part in the ballot), but some id might be used by inactive invocations. Use `error += select(1, 0, sum <= 1);` might check each `id` used at most once (no duplication) instead?

And we can also check `(sum == 0) || (id < sgSize)` to ensure no id is larger than subgroup size, if we use `128` as loop boundary.

alan-baker:
> This would check that each `id < sgSize` is used once and only once for all active invocations (if only the active ones take part in the ballot), but some id might be used by inactive invocations. Use `error += select(1, 0, sum <= 1);` might check each `id` used at most once (no duplication) instead?
> 
> And we can also check `(sum == 0) || (id < sgSize)` to ensure no id is larger than subgroup size, if we use `128` as loop boundary.

I suppose if the invocation is inactive it doesn't really matter which id it is assigned, but I agree the select condition is better formulated as an inequality. The secondary check makes sense too.

jzm-intel:
Please take a look, thanks! `repId` is removed, and do more validation within shader.

alan-baker (approved):
Thanks for this fix! I also tested it on my Macbook M1.

jzm-intel:
Thanks for reviewing!

PR #4026, opened 2024-11-01
Re-enable linear with comparisons.

greggman:
It's a little confusing but depth textures used for "sampling" are not filterable but used for "comparisons" they are.

|                               | texture_?d<f32> with depth texture | texture_depth_?? |
| ----------------------------- | ---------------------------------- | ---------------- |
| sampler `'linear'`            |                bad                 |      bad         |
| sampler_comparison `'linear'` |                N/A                 |      OK          |

### DIFF
--- a/src/webgpu/format_info.ts
+++ b/src/webgpu/format_info.ts
@@ -1793,9 +1793,21 @@ export function canUseAsRenderTarget(format: GPUTextureFormat) {
   return kTextureFormatInfo[format].colorRender || isDepthOrStencilTextureFormat(format);
 }
 
-export function isFilterable(format: GPUTextureFormat) {
+export function is32Float(format: GPUTextureFormat) {
+  return format === 'r32float' || format === 'rg32float' || format === 'rgba32float';
+}
+
+/**
+ * Returns true if texture is filterable as `texture_xxx<f32>`
+ *
+ * examples:
+ * * 'rgba8unorm' -> true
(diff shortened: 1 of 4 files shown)


src/webgpu/format_info.ts
```diff
@@ -1793,9 +1793,21 @@ export function canUseAsRenderTarget(format: GPUTextureFormat) {
...
+
+/**
+ * Returns true if texture is filterable as `texture_xxx<f32>`
+ *
+ * examples:
+ * * 'rgba8unorm' -> true
+ * * 'depth16unorm' -> false
+ * * 'rgba32float' -> true (you need to enable feature 'float32-filterable')
+ */
+export function isFilterableAsTextureF32(format: GPUTextureFormat) {
```
shrekshao, line 1808: Question: are r16float, rg16float, rgba16float filterable as TextureF32?

greggman: yes, all normal color formats are filtering as textureF32 (r8unorm, r8snorm, rgba8norm). Effectively this is can you use the texture with `texture_2d<f32>`, `texture_2d_array<f32>`, `texture_cube<f32>`, etc...

shrekshao: Oh for r16float  and etc, their `info.color?.type` is float already.

shrekshao (approved):
LGTM

PR #4028, opened 2024-11-02
Don't test texture_depth_xxx with linear

greggman:
texture_depth_xxx can only be used with non-filtering samplers or comparison samples, not with filtering samplers.

see table in #4026

Also added a bunch of depth formats. Sorry it's so large but both issues were editing the same lines.


shrekshao (approved):
LGTM

Recap: This PR does ( use textureGather(Compare) to represent all texture related builtins in this table):
| | texture_?d (with depth texture)|	texture_depth_?? |
|---|---|---|
|textureGather | not touched | remove 'filt' from params (don't test linear) ; MAINTENANCE_TODO|
| textureGatherCompare	| N/A | MAINTENANCE_TODO|

PR #4031, opened 2024-11-05
Test non-filterable textures can't be used with filtering sampler.

greggman:
IIUC these tests are missing from the CTS

Note that if `u32`/`i32` are removed from `textureGather` then there is no other texture builtin that uses `u32`/`i32` except `textureLoad`

Testing locally on Chrome with [this patch check does this validation](https://dawn-review.googlesource.com/c/dawn/+/213176/4/src/dawn/native/ShaderModule.cpp) they pass. 
With [this patch that does this validation but allows depth with a deprecation warning](https://dawn-review.googlesource.com/c/dawn/+/213176/5/src/dawn/native/ShaderModule.cpp) they all pass except for depth. On Chrome Canary with neither patch they all fail except for the `f32` tests which are supposed to work.

<hr>

### DIFF
--- a/src/webgpu/api/validation/non_filterable_texture.spec.ts
+++ b/src/webgpu/api/validation/non_filterable_texture.spec.ts
@@ -1,0 +1,126 @@
+export const description = `
+Tests that non-filterable textures used with filtering samplers generate a validation error.
+`;
+
+import { makeTestGroup } from '../../../common/framework/test_group.js';
+import { keysOf } from '../../../common/util/data_tables.js';
+
+import { ValidationTest } from './validation_test.js';
+
+const kNonFilterableCaseInfo: Record<GPUTextureSampleType, { type: string; component: string }> = {
+  sint: { type: 'i32', component: '0,' },
+  uint: { type: 'u32', component: '0,' },
+  float: { type: 'f32', component: '0,' }, // no error for f32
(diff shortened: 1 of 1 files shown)


src/webgpu/api/validation/non_filterable_texture.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+import { keysOf } from '../../../common/util/data_tables.js';
+
+import { ValidationTest } from './validation_test.js';
+
+const kNonFilterableCaseInfo: {
+  [key: string]: { sampleType: GPUTextureSampleType; component: string };
+} = {
+  i32: { sampleType: 'sint', component: '0,' },
+  u32: { sampleType: 'uint', component: '0,' },
+  f32: { sampleType: 'float', component: '0,' }, // no error for f32
```
kainino0x, line 15: what about `'unfilterable-float'`?
If intentionally skipping it, just leave a short comment about why (maybe because it's already tested elsewhere, though probably no harm in testing it here too)

kainino0x (approved):

src/webgpu/api/validation/non_filterable_texture.spec.ts
greggman: added

PR #4034, opened 2024-11-06
Assert GPUAdapterInfo instances

beaufortfrancois:
Without this fix, Chrome Stable passes https://gpuweb.github.io/cts/standalone/?q=webgpu:api,operation,adapter,info:same_object:* even if GPUDevice.adapterInfo is not implemented and returns `undefined`.

@kainino0x Please review.


kainino0x (approved):
I sorta noticed this was passing in stable when I was editing your last PR, I figured it didn't _really_ matter because the other test would fail. But this makes the results make more sense! :)

PR #4035, opened 2024-11-06
Fragment subgroup_size test: Don't use broadcasting

jzm-intel:
This CL modify the shader for fragment `subgroup_size` built-in test to prevent using broadcasting (from invocation 0), because any invocation id can be inactivate for fragment shader. This CL use ballot instead and do more validation within the shader.

### DIFF
--- a/src/webgpu/shader/execution/shader_io/fragment_builtins.spec.ts
+++ b/src/webgpu/shader/execution/shader_io/fragment_builtins.spec.ts
@@ -1622,13 +1622,6 @@ function checkSubgroupSizeConsistency(
 -  countActive: ${countActive}`);
       }
 
-      if (subgroupSize < countActive) {
-        return new Error(`Unexpected active invocations number larger than subgroup size
--       icoord: (${row}, ${col})
-- subgroupSize: ${subgroupSize}
--  countActive: ${countActive}`);
-      }
-
       if (subgroupSize !== ballotedSubgroupSize) {
         return new Error(`Inconsistent subgroup size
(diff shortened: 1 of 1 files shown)


jzm-intel:
Note that during test we found on Intel UHD770 the first run of fragment shader could be in `subgroup_size == 8` while the D3D12 API report `WaveLaneCountMin == 16` and thus WebGPU report `adapter.limits.minSubgroupSize == 16` in Chromium. The compiled shader seems get replaced in the second and following run, and then it would have a `subgroup_size` of 16, which means that a retry can make the CTS pass, and the try-bot might be not easy to catch such kind of flaky issues.

We are going to report this issue as a driver bug, and make a workaround to make WebGPU loosing reported `minSubgroupSize` as 8 for certain devices.

alan-baker:
> Note that during test we found on Intel UHD770 the first run of fragment shader could be in `subgroup_size == 8` while the D3D12 API report `WaveLaneCountMin == 16` and thus WebGPU report `adapter.limits.minSubgroupSize == 16` in Chromium. The compiled shader seems get replaced in the second and following run, and then it would have a `subgroup_size` of 16, which means that a retry can make the CTS pass, and the try-bot might be not easy to catch such kind of flaky issues.
> 
> We are going to report this issue as a driver bug, and make a workaround to make WebGPU relaxing reported `minSubgroupSize` as 8 for certain devices.

Thanks! Not entirely unexpected as Vulkan originally had this problem too.

alan-baker (approved):

jzm-intel:
Thanks for reviewing!

PR #4048, opened 2024-11-19
Fix makeRandomDepthComparisonTexelGenerator

greggman:
This bug broke `textureXXCompare` tests with unencodable depth formats which were added recently.

https://github.com/gpuweb/cts/pull/4012


kainino0x (approved):
LGTM

PR #4049, opened 2024-11-19
Fix readTextureToTextureViews for multisampled textures

greggman:
I'm not sure when this bug was introduced. The tests this broke used to pass (`textureLoad:multisampled`). Fixed the bug and added a test to hopefully find it quicker if it breaks in the future.

Also added a test for `createTextureWithRandomDataAndGetTexels` with a generator. This was fixed in a previous PR without a a specific test as the fact that it failed other tests showed it needed fixing but I decided it's probably best to keep these tests too.

### DIFF
--- a/src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
+++ b/src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
@@ -3031,11 +3031,14 @@ export async function readTextureToTexelViews(
         }
       `,
     });
-    const sampleType = isDepthTextureFormat(texture.format)
-      ? 'unfilterable-float'
-      : isStencilTextureFormat(texture.format)
+    const info = kTextureFormatInfo[texture.format];
+    const sampleType = info.depth
+      ? 'unfilterable-float' // depth only supports unfilterable-float if not a comparison.
+      : info.stencil
       ? 'uint'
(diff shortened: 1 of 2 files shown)


src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
```diff
@@ -3031,10 +3031,13 @@ export async function readTextureToTexelViews(
...
       `,
     });
+    const colorType = kTextureFormatInfo[texture.format].color?.type;
     const sampleType = isDepthTextureFormat(texture.format)
       ? 'unfilterable-float'
       : isStencilTextureFormat(texture.format)
       ? 'uint'
+      : colorType === 'float'
+      ? 'unfilterable-float'
       : kTextureFormatInfo[texture.format].color?.type ?? 'unfilterable-float';
```
kainino0x, line 3041: nit: there's a lot of logic happening in this ternary chain, I was going to suggest pulling it out into a function but I realized it can be simplified by taking advantage of type narrowing:
```suggestion
    const info = kTextureFormatInfo[texture.format];
    const sampleType = info.depth
      ? 'unfilterable-float' // depth or depth+stencil (TODO: should comment something about why we pick this sample type in this case)
      : info.stencil
      ? 'uint'
      : info.color.type === 'float'
      ? 'unfilterable-float'
      : info.color.type;
```

kainino0x (approved):

PR #4050, opened 2024-11-20
Do not query weights for `textureSampleBaseClampToEdge`

greggman:
`textureSampleBaseClampToEdge` does not need the GPU weights.

Skip loading them. This will let the tests run on Mac Intel where the weights are bad.

### DIFF
--- a/src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
+++ b/src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
@@ -191,7 +191,10 @@ export function skipIfTextureFormatNotSupportedNotAvailableOrNotFilterable(
 }
 
 const builtinNeedsMipGradientValues = (builtin: TextureBuiltin) =>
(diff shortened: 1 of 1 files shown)


src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
```diff
@@ -191,7 +191,10 @@ export function skipIfTextureFormatNotSupportedNotAvailableOrNotFilterable(
 }
 
 const builtinNeedsMipGradientValues = (builtin: TextureBuiltin) =>
```
kainino0x, line 193: If I'm understanding this variable name correctly, shouldn't `textureSampleGrad`, `textureSampleLevel`, and `textureSampleCompareLevel` also be listed?

LGTM regardless though.

kainino0x (approved):

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
greggman: renamed to something more appropriate

PR #4053, opened 2024-11-21
Add warning in checkCallResults for compressed textures

greggman:
We sample a compressed texture on the GPU and compare that to expected. If they don't match we try to identify which texels were sampled. This is done by doing a binary search, creating textures with some pixels white and some black and then sampling the texture to see if those white pixels contributed to the sample values.

We can't do this for compressed textures because we can not easily create a compressed texture with white and black pixels exactly where we want them to appear.

So, for compressed textures we just use rgba8unorm when identifying sample points.

That can be useful to help find major bugs either in the CTS itself or in the GPU that are true for all texture formats. Unfortunately, if the bug is specifically with compressed textures this doesn't help.

So, print out a warning that the sample points might not match.

Also, change the diagram to show blocks. Example when blockWidth = 2, blockHeight = 2

```
   +===+===+===+===+
   # a |   #   |   #
   +---+---+---+---+
   #   |   #   |   #
   +===+===+===+===+
   #   |   #   |   #
   +---+---+---+---+
   #   |   #   | b #
   +===+===+===+===+
```

really wish we could switch back to unicode 😢

### DIFF
--- a/src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
+++ b/src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
@@ -2582,6 +2582,21 @@ export async function checkCallResults<T extends Dimensionality>(
 
           const callForSamplePoints = checkInfo.calls[callIdx];
 
+          // We're going to create textures with black and white texels
+          // but if it's a compressed texture we use an encodable texture.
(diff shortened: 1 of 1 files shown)


shrekshao (approved):
LGTM

PR #4054, opened 2024-11-21
Compat: Don't test depth textures with non-comparison samplers

greggman:
https://github.com/gpuweb/gpuweb/issues/4970

### DIFF
--- a/src/webgpu/gpu_test.ts
+++ b/src/webgpu/gpu_test.ts
@@ -300,6 +300,14 @@ export class GPUTestSubcaseBatchState extends SubcaseBatchState {
     }
   }
 
+  /** Skips this test case if the a depth texture can not be used with a non-comparison sampler. */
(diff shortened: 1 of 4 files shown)


src/webgpu/gpu_test.ts
```diff
@@ -300,6 +300,14 @@ export class GPUTestSubcaseBatchState extends SubcaseBatchState {
     }
   }
 
+  /** Skips this test case if the a depth texture can not be used with a non-comparison sampler. */
```
shrekshao, line 303: nit: "if the ~~a~~ depth texture"

shrekshao (approved):
LGTM w nit

src/webgpu/gpu_test.ts
greggman, line 303:
```suggestion
  /** Skips this test case if a depth texture can not be used with a non-comparison sampler. */
```

PR #4055, opened 2024-11-22
Compat: Validate non-comparsion sampler + depth texture fails.

greggman:
Using a depth texture with a non-comparison sampler should generate a validation error in compat but not in core.

Note: this PR should probably wait for https://github.com/gpuweb/gpuweb/pull/4988 to be approved in case this is not where the validation error should happen.

<hr>

### DIFF
--- a/src/webgpu/compat/api/validation/pipeline_creation.spec.ts
+++ b/src/webgpu/compat/api/validation/pipeline_creation.spec.ts
@@ -1,0 +1,146 @@
+export const description = `
+Tests that createComputePipeline(async), and createRenderPipeline(async)
+reject pipelines that are invalid in compat mode
+
+- test that depth textures can not be used with non-comparison samplers
+
+TODO:
+- test that a shader that has more than min(maxSamplersPerShaderStage, maxSampledTexturesPerShaderStage)
+  texture+sampler combinations generates a validation error.
+`;
+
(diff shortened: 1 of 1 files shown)


src/webgpu/compat/api/validation/pipeline_creation.spec.ts
```diff
@@ -0,0 +1,146 @@
...
+            return vec4f(0);
+        }
+
+        @compute @workgroup_size(1) fn cs() {
+            ${stage === 'compute' ? usageWGSL : ''};
+        }
+      `,
+    });
+
+    const success = !t.isCompatibility || textureType === 'texture_2d<f32>';
```
shrekshao, line 122: nit: this is under `compat/` so I assume `!t.isCompatibility` is not needed?

shrekshao (approved):
LGTM

src/webgpu/compat/api/validation/pipeline_creation.spec.ts
greggman: It was decided all of the compat tests should run in core and pass in the case where compat fails, as a validation of the test itself. In other words, the test is testing that something that's allowed in core is not allowed in compat. Checking it passes in core checks the test is correct.

PR #4058, opened 2024-11-22
Use minSubgroupSize, maxSubgroupSize from GPUAdapterInfo

dneto0:
https://github.com/gpuweb/gpuweb/pull/4963  defines how to get minSubgroupSize, maxSubgroupSize.
It's now from GPUAdapterInfo.


dneto0:
I tested this locally with dawn.node

alan-baker (approved):

alan-baker:
Note: the spec now specifies default values so there should be an API side test for when the feature isn't supported.

PR #4059, opened 2024-11-23
Add MaxLimitsTest that sets all limits to the adapter's

greggman:
This is what I came up with. Without a major refactor the issue is the adapter is requested at an extremely low level but a test needs access to an adapter to select limits.

So, made `selectDeviceOrSkipTestCase` pass an optional callback all the way down that gets passed the adapter and a `CanonicalDeviceDescriptor`, giving it a chance to modify it down at the level where an adapter is available.

More of the CTS tests need to be testing with max limits.

Issue: #3363

### DIFF
--- a/src/webgpu/gpu_test.ts
+++ b/src/webgpu/gpu_test.ts
@@ -41,7 +41,13 @@ import {
 import { checkElementsEqual, checkElementsBetween } from './util/check_contents.js';
 import { CommandBufferMaker, EncoderType } from './util/command_buffer_maker.js';
 import { ScalarType } from './util/conversion.js';
-import { DevicePool, DeviceProvider, UncanonicalizedDeviceDescriptor } from './util/device_pool.js';
+import {
+  CanonicalDeviceDescriptor,
+  DescriptorModifierFn,
+  DevicePool,
+  DeviceProvider,
+  UncanonicalizedDeviceDescriptor,
+} from './util/device_pool.js';
 import { align, roundDown } from './util/math.js';
 import { physicalMipSizeFromTexture, virtualMipSize } from './util/texture/base.js';
 import {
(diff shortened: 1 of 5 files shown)


greggman:
Note: This test fails in Chrome because of the origin trial for subgroups.

It fails with

> WebGPU device failed to initialize with OperationError "Failed to execute 'requestDevice' on 'GPUAdapter': The limit "minSubgroupSize" is not recognized."

I don't think the test should change and I think it should be a high-priority for Chrome not to fail if you request all the limits by copying them

src/webgpu/gpu_test.ts
```diff
@@ -1276,6 +1287,82 @@ export class GPUTest extends GPUTestBase {
...
+}
+
+function setAllLimitsToAdapterLimits(
+  adapter: GPUAdapter,
+  desc: CanonicalDeviceDescriptor | undefined
+) {
+  const descWithMaxLimits: CanonicalDeviceDescriptor = {
+    requiredFeatures: [],
+    defaultQueue: {},
+    ...(desc ?? desc),
```
kainino0x, line 1309: nit: is this different from just `...desc`, or just a leftover?

src/webgpu/gpu_test.ts
```diff
@@ -1276,6 +1287,82 @@ export class GPUTest extends GPUTestBase {
...
+      return setAllLimitsToAdapterLimits(adapter, desc);
+    };
+    super.selectDeviceOrSkipTestCase(initUncanonicalizedDeviceDescriptor(descriptor), wrapper);
+  }
+}
+
+/**
+ * A Test that requests all the max limits from the adapter on the device.
+ */
+export class MaxLimitsTest extends GPUTestBase {
```
kainino0x, line 1334: nit: Does this need to extend GPUTestBase instead of GPUTest? It doesn't duplicate that much code so no big deal, but seems like it might not need to?

I also wonder if this could be a mixin but I'm pretty sure that's not necessary.

kainino0x (approved):
Nice, this is probably a simpler way of adding support for this than anything I'd thought of! 

> I don't think the test should change and I think it should be a high-priority for Chrome not to fail if you request all the limits by copying them

+1

src/webgpu/gpu_test.ts
greggman: changed to inherit from `GPUTest`

PR #4067, opened 2024-11-27
Compat: stop using storage buffer when reading texture mix weights

greggman:
storage buffers in both vertex and fragment shaders are optional


shrekshao (approved):
LGTM

PR #4072, opened 2024-12-03
Move towards deprecating subgroups_f16 enable (feature

petermcneeleychromium:
We are deprecating the 'subgroups_f16' enable
One will only need 'f16' and 'subgroups' enables to use f16 subgroups operations.

Going forward the 'subgroups_f16' enable will have no effect. Eventually we want to remove this enable.

https://github.com/gpuweb/gpuweb/blob/main/proposals/subgroups.md

crbug.com/380862306


petermcneeleychromium:
Conclusion from meeting. Simply delete the tests that check for the requirements of subgroup_f16
(this is because we dont want to force this deprecation behavior on all implementers of webgpu)

jrprice (approved):
LGTM.

I assume there will be more followups to remove `enable subgroups_f16` from other tests that are still using it.

PR #4081, opened 2024-12-06
Set `featureLevel: "compatibility"` when compatibility=1 is enabled

beaufortfrancois:
As suggested in https://g-issues.chromium.org/issues/366151404#comment9, this PR adds a featureLevel CTS option.

Note that I'm not 100% sure I cover all changes required for the CTS.

<hr>


kainino0x:
It's not immediately apparent that we will need to be able to run the CTS under customizable featureLevel values. It would be much simpler and avoid a lot of migration work if the CTS just keeps `?compatibility=1` as the option. We can add `?featureLevel=compatibility` later if we find it's needed.

The thing that we _need_ to do is have CTS request adapters using `featureLevel: "compatibility"` so that chromium can stop obeying `compatibilityMode: true` (but of CTS should keep `compatibilityMode: true` for now as well during migration).

beaufortfrancois:
@kainino0x This makes this PR much simpler ;)

src/common/util/navigator_gpu.ts
```diff
@@ -43,6 +43,9 @@ export function setDefaultRequestAdapterOptions(options: GPURequestAdapterOption
     throw new Error('must call setDefaultRequestAdapterOptions before getGPU');
   }
   defaultRequestAdapterOptions = { ...options };
+  // MAINTENANCE_TODO: remove this once compatibilityMode is removed.
```
kainino0x, line 46: I think this should be updated at the call sites outside `setDefaultRequestAdapterOptions`, rather than here. It's confusing if this function does anything other than just set results directly into `defaultRequestAdapterOptions`

beaufortfrancois: Done

src/resources/cache/webgpu/shader/execution/bitcast.bin
kainino0x, line 1: This change should be undone; this file gets rewritten constantly back and forth because it generates differently on Linux vs Mac (or something like that). It doesn't actually need to change.

kainino0x (approved):

src/resources/cache/webgpu/shader/execution/bitcast.bin
kainino0x: Done

greggman:
I think this should be changed so it doesn't request `'core'`.

Instead of

```
    featureLevel: globalTestConfig.compatibility ? 'compatibility' : 'core',
```

It should be

```
    globalTestConfig.compatibility && { featureLevel: 'compatibility' },
```

wdyt?

beaufortfrancois:
`'core'` is the default value for `featureLevel` so I think it's fine. See https://gpuweb.github.io/gpuweb/#dom-gpurequestadapteroptions-featurelevel

greggman:
The problem is it makes the CTS never test that you get core if you don't request any feature level

beaufortfrancois:
I believe this will be tested once `featureLevel` GPUAdapter attribute is added to the spec in https://github.com/gpuweb/gpuweb/pull/5012?

PR #4083, opened 2024-12-09
Remove subgroups f16 enable from cts  ("subgroups_f16")

petermcneeleychromium:
Remove subgroups f16 enable from cts  ("subgroups_f16")

This completes the CTS change corresponding to:
https://dawn-review.googlesource.com/c/dawn/+/217294

crbug.com/380244620

### DIFF
--- a/src/webgpu/shader/validation/parse/enable.spec.ts
+++ b/src/webgpu/shader/validation/parse/enable.spec.ts
@@ -70,24 +70,6 @@ enable f16;`,
     code: `enable subgroups;`,
     pass: true,
   },
-  subgroups_f16_fail: {
-    code: `enable subgroups_f16;`,
-    pass: false,
-  },
-  subgroups_f16_pass1: {
-    code: `
-    enable subgroups_f16;
-    enable subgroups;
-    enable f16;`,
-    pass: true,
-  },
-  subgroups_f16_pass2: {
-    code: `
-    enable f16;
-    enable subgroups;
-    enable subgroups_f16;`,
-    pass: true,
-  },
 };
 
 g.test('enable')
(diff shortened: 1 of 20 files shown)


src/webgpu/shader/validation/parse/enable.spec.ts
```diff
@@ -70,36 +70,14 @@ enable f16;`,
     code: `enable subgroups;`,
     pass: true,
   },
-  subgroups_f16_fail: {
-    code: `enable subgroups_f16;`,
-    pass: false,
-  },
-  subgroups_f16_pass1: {
```
petermcneeleychromium, line 77: Maybe I should leave one of these?

petermcneeleychromium:
Tested against latest local dawn.

src/webgpu/shader/validation/parse/enable.spec.ts
alan-baker: Yes, maybe we should leave one and have a case to enable both features.

src/webgpu/shader/validation/parse/enable.spec.ts
```diff
@@ -95,11 +89,7 @@ g.test('enable')
   .beforeAllSubcases(t => {
     const features: GPUFeatureName[] = [];
     const name: string = t.params.case as string;
-    if (name.includes('subgroups_f16')) {
-      features.push('subgroups' as GPUFeatureName);
-      features.push('subgroups-f16' as GPUFeatureName);
-      features.push('shader-f16');
-    } else if (name.includes('subgroups')) {
+    if (name.includes('subgroups')) {
```
alan-baker, line 92: This logic seems like it will fail the `subgroups_f16_pass1` and `subgroups_f16_pass2` tests because only the subgroups feature will be enabled.

petermcneeleychromium: Yes fixed.

alan-baker (approved):

PR #4085, opened 2024-12-10
Move MaxLimitsTest to MaxLimitTestMixin

greggman:
I wanted to use `MaxLimitsTest` with `resource_compatibility.spec.ts` but that file's tests are based on `CreateRenderPipelineValidationTest` which itself is based on `ValidationTest` and so would force making `ValidationTest` be based on `MaxLimitsTest`. I actually think thats a good idea. The default should be max limits. But, it breaks too many things.

So, max MaxLimitsTest a mixin.

### DIFF
--- a/src/webgpu/gpu_test.ts
+++ b/src/webgpu/gpu_test.ts
@@ -1333,16 +1333,27 @@ export class MaxLimitsGPUTestSubcaseBatchState extends GPUTestSubcaseBatchState 
   }
 }
 
-/**
- * A Test that requests all the max limits from the adapter on the device.
- */
-export class MaxLimitsTest extends GPUTest {
-  public static override MakeSharedState(
-    recorder: TestCaseRecorder,
-    params: TestParams
-  ): GPUTestSubcaseBatchState {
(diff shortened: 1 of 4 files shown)


src/webgpu/gpu_test.ts
```diff
@@ -1333,10 +1333,14 @@ export class MaxLimitsGPUTestSubcaseBatchState extends GPUTestSubcaseBatchState
...
 
+export type MaxLimitsTestMixinType = {
+  // placeholder. Change to an interface if we need MaxLimits specific methods.
+};
+
 /**
  * A Test that requests all the max limits from the adapter on the device.
  */
-export class MaxLimitsTest extends GPUTest {
+export class MaxLimitsTest2 extends GPUTest {
```
kainino0x, line 1343: nit: Why `2`? Is this used at all?

kainino0x (approved):
LGTM regardless of comment

src/webgpu/gpu_test.ts
greggman: Doh, sorry, Just forgot to delete it.

PR #4093, opened 2024-12-16
Refactor discard test for no storage buffers

greggman:


alan-baker (commented):
Still reviewing, but what do you you think instead about making the choice between framebuffer and storage buffer a subcase to each test and skip the storage buffer version on compat implementations that don't support them? To me, the benefit would be that the same tests are run everywhere so and it's not the case that the test functions differently depending on the hardware. It also ensures we cover discard functionality more thoroughly.

greggman:
sgtm. i"ll update

greggman:
updated. please take a look.

greggman:
ah, I made them cases, would be prefer subcases? There appears to be only one expectation in dawn related to this and it should be fine is I think.

alan-baker (approved):

PR #4101, opened 2024-12-17
Compat: skip memory_sync buffer tests when no support

greggman:
At a glance I doesn't seem like there is anything to refactor here. The test that are actually using storage buffers need storage buffers. So, skip them if there are none.


shrekshao (approved):
LGTM

shrekshao:
FYI I will be ooo starting tomorrow till new year and won't be available too much on github review

PR #4103, opened 2024-12-18
Fix DeviceModifier

greggman:
The issue is the key is selected before the modification happens which means the devices in the pool will not match the key.

I can't think of a way to fix this using the adaptor so maybe this hacky solution works by letting you modify the key.

Generally you just append a string but I made it function because MaxLimitsTestMixin wants to be able to chain to other modifiers.

Maybe it should all be re-designed

This came up because I notice a test passing that shouldn't have passed. It passsed because a `MaxLimitsTestMixin` test ran first and it's key was `''` which meant that other tests got a max limits device from the pool.

### DIFF
--- a/src/webgpu/gpu_test.ts
+++ b/src/webgpu/gpu_test.ts
@@ -43,7 +43,7 @@ import { CommandBufferMaker, EncoderType } from './util/command_buffer_maker.js'
 import { ScalarType } from './util/conversion.js';
 import {
   CanonicalDeviceDescriptor,
-  DescriptorModifierFn,
+  DescriptorModifier,
   DevicePool,
   DeviceProvider,
   UncanonicalizedDeviceDescriptor,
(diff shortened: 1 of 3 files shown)


src/webgpu/gpu_test.ts
```diff
@@ -1466,7 +1473,7 @@ export interface TextureTestMixinType {
    * Effectively it's a Uint8Array to Uint8Array copy that
    * does the same thing as `writeTexture` but because the
    * destination is a buffer you have to provide the parameters
-   * of the destination buffer similarly to how you'd provide them
+   * of the destination buffer similarly to how you'esc provide them
```
kainino0x, line 1476: typo

kainino0x (approved):
The whole adapter/device init system could definitely use rework someday but for now I'm totally fine with piling in minor hacks, this is not that bad at all.

src/webgpu/gpu_test.ts
greggman:
```suggestion
   * of the destination buffer similarly to how you'd provide them
```

PR #4104, opened 2024-12-18
Compat: Skip texture sync tests that use storage textures if 0

greggman:
It's not clear to me if writing to a storage texture is important for the render-pass-encoder and render-bundle-encoder or if there should be more cases

Right now `render-pass-encoder` and `render-bundle-encoder` write directly to a storage texture, another case could instead render to that texture instead as a colorAttachment. I guess that's covered by the `attachment-store` and `attachment-resolve` cases though.

In any case, this PR makes the test skip using storage textures if the device doesn't have any.


shrekshao (approved):
LGTM

PR #4109, opened 2024-12-19
Vertex state correctness

greggman:
Refactor these tests so they don't use storage buffers.

The tests were using one uniform or storage buffer per
vertex attribute. Switched to using just a single uniform
buffer for all attributes.

Note: There was a note about increasing the vertex count
by +1 in vertex_buffer_used_multiple_times_interleaved.
I don't know if that's old because removing the +1 there
and a few lines below does not break the test nor generate
any validation errors.

The reason I needed to remove it was the code that packs
the data into a single buffer needed the size of each buffer
to match kVertexCount but that particular test was passing
+1 in some places and not in others.


kainino0x (approved):
lgtm, that was a pleasantly simpler review than I thought given how complex I remember this test being :)

Issue #4108, opened 2024-12-19
Test CopyEI2T with Image has rotation metadata


shaoboyan091:
The lack of such cases causes isseus like https://issues.chromium.org/issues/384858956. We need to generate pictures with rotation metadata and have a case to cover.

PR #4111, opened 2024-12-19
CopyEI2T: Cases To Cover Rotated Images

shaoboyan091:
This PR created images with rotation metadata. It tests using CopyEI2T to upload image and imageBitmap with 'from-image' orientation flag.

Issue: #4108

<hr>


shaoboyan091:
Seems I cannot add reviewers in right side bar. Explicit @kainino0x @beaufortfrancois  for reviewing.

src/resources/README.md
```diff
@@ -91,3 +91,16 @@ ffmpeg -display_vflip -i temp.mp4 -c copy four-colors-vp9-bt601-vflip.mp4
...
+// Generate jpg picture with 90 cw rotation metadata
+exiftool -Orientation#=6 four-colors.jpg -o .\four-colors-rotate-90-cw.jpg
+
+// Generate jpg picture with 180 cw rotation metadata
+exiftool -Orientation#=3 four-colors.jpg -o .\four-colors-rotate-180-cw.jpg
+
+// Generate jpg picture with 270 cw rotation metadata
+exiftool -Orientation#=8 four-colors.jpg -o .\four-colors-rotate-270-cw.jpg
+
+```
```
kainino0x, line 106: To simplify the tests I think it would be preferable for all four images (and perhaps the png) to have the same arrangement of colors, _post-rotation_. That way the expected colors will be the same for all four.

You could for example use imagemagick to generate four hard-rotated jpgs from `four-colors.png`, then use exiftool to soft-rotate them back.

src/webgpu/web_platform/copyToTexture/image.spec.ts
```diff
@@ -339,3 +346,113 @@ g.test('copy_subrect_from_2D_Canvas')
...
+    const dstTexture = t.createTextureTracked({
+      size: { width, height },
+      format: kColorFormat,
+      usage:
+        GPUTextureUsage.COPY_DST | GPUTextureUsage.COPY_SRC | GPUTextureUsage.RENDER_ATTACHMENT,
+    });
+
+    t.device.queue.copyExternalImageToTexture(
+      {
+        source: image,
```
kainino0x, line 396: It may simplify the tests to just have a parameter for how we upload the image, instead of having two very similar tests - because the rest of the test (the checking code) will be the same. (There's also at least one more code path that it would be good to test, which is ImageBitmap from blob.)

(If you do that then maybe it should go in a new file like `copyToTexture/image_file.spec.ts` since it's not divided up the same way as the other tests.)

To track writing more tests like that I've filed #4114. If you think of any other upload methods please log them on that issue. If you don't cover all of those in this PR, feel free to leave a TODO for now.

kainino0x (commented):
Test code lgtm, but a few comments

src/resources/README.md
shaoboyan091: Done.

src/webgpu/web_platform/copyToTexture/image.spec.ts
shaoboyan091: Done.

src/webgpu/web_platform/util.ts
```diff
@@ -606,3 +616,124 @@ export async function captureCameraFrame(test: GPUTest): Promise<VideoFrame> {
...
+}
+
+// Use fetch() to load image file as blob
+function loadXHR(url: string): Promise<Blob> {
+  return new Promise<Blob>((resolve, reject) => {
+    const xhr = new XMLHttpRequest();
+    xhr.open('GET', url);
+    xhr.responseType = 'blob';
+    xhr.onerror = function () {
+      reject(new Error('Network error.'));
```
shaoboyan091, line 728: @kainino0x  I'm not sure what's the best way to handle this reject case. Do you have any suggestions?

src/resources/README.md
```diff
@@ -91,3 +91,25 @@ ffmpeg -display_vflip -i temp.mp4 -c copy four-colors-vp9-bt601-vflip.mp4
...
+
+// Generate jpg picture with 180 cw rotation metadata
+Use a image tool (e.g. "Paint" app on Windows) to create four-colors-hard-rotate-180-ccw.jpg and check with exiftool to ensure no orientation metadata been set.
+exiftool -Orientation#=3 four-colors.jpg -o .\four-colors-rotate-180-cw.jpg
+rm four-clors-hard-rotate-180-ccw.jpg
+
+// Generate jpg picture with 270 cw rotation metadata
+Use a image tool (e.g. "Paint" app on Windows) to create four-colors-hard-rotate-270-ccw.jpg and check with exiftool to ensure no orientation metadata been set.
+exiftool -Orientation#=8 four-colors.jpg -o .\four-colors-rotate-270-cw.jpg
+rm four-colors-hard-rotate-270-ccw.jpg
```
kainino0x, line 113:
```suggestion
Use a image tool (e.g. "Paint" app on Windows) to create four-colors-hard-rotate-90-ccw.jpg and check with exiftool to ensure no orientation metadata been set.
exiftool -Orientation#=6 four-colors-hard-rotate-90-ccw.jpg -o four-colors-rotate-90-cw.jpg
rm four-colors-hard-rotate-90-ccw.jpg

// Generate jpg picture with 180 cw rotation metadata
Use a image tool (e.g. "Paint" app on Windows) to create four-colors-hard-rotate-180-ccw.jpg and check with exiftool to ensure no orientation metadata been set.
exiftool -Orientation#=3 four-colors-hard-rotate-180-ccw.jpg -o four-colors-rotate-180-cw.jpg
rm four-colors-hard-rotate-180-ccw.jpg

// Generate jpg picture with 270 cw rotation metadata
Use a image tool (e.g. "Paint" app on Windows) to create four-colors-hard-rotate-270-ccw.jpg and check with exiftool to ensure no orientation metadata been set.
exiftool -Orientation#=8 four-colors-hard-rotate-270-ccw.jpg -o four-colors-rotate-270-cw.jpg
rm four-colors-hard-rotate-270-ccw.jpg
```

src/webgpu/web_platform/util.ts
kainino0x: This looks good, we want any error case to fail the test.

However I don't think XHR should ever be need, rather than `fetch()` + `.blob()`:
https://developer.mozilla.org/en-US/docs/Web/API/Response/blob

(I'm surprised to see we don't already have any code that uses `fetch()`, but it seems the only place we use a still-image resource uses `HTMLImageElement`.)

src/webgpu/web_platform/util.ts
```diff
@@ -606,3 +616,124 @@ export async function captureCameraFrame(test: GPUTest): Promise<VideoFrame> {
...
+      },
+      display: {
+        topLeftColor: 'yellow',
+        topRightColor: 'red',
+        bottomLeftColor: 'blue',
+        bottomRightColor: 'green',
+      },
+    },
+  },
+} as const);
```
kainino0x, line 679: Are the `coded` values needed? If we need them in the future, we can always add them later. For now we can simplify:

```suggestion
const kFourColorsInfo = {
  display: {
    topLeftColor: 'yellow',
    topRightColor: 'red',
    bottomLeftColor: 'blue',
    bottomRightColor: 'green',
  },
} as const;

export const kImageInfo = makeTable({
  table: {
    'four-colors.jpg': kFourColorsInfo,
    'four-colors-rotate-90-cw.jpg': kFourColorsInfo,
    'four-colors-rotate-180-cw.jpg': kFourColorsInfo,
    'four-colors-rotate-270-cw.jpg': kFourColorsInfo,
  },
} as const);
```

kainino0x, line 679: Also just a quick question, what does makeTable do? At least in this case it seems like this is equivalent, but I'm not sure:

```ts
export const kImageInfo = {
  'four-colors.jpg': kFourColorsInfo,
  'four-colors-rotate-90-cw.jpg': kFourColorsInfo,
  'four-colors-rotate-180-cw.jpg': kFourColorsInfo,
  'four-colors-rotate-270-cw.jpg': kFourColorsInfo,
} as const;
```

src/webgpu/web_platform/util.ts
```diff
@@ -606,3 +616,124 @@ export async function captureCameraFrame(test: GPUTest): Promise<VideoFrame> {
...
+  objectTypeFromFile: ObjectTypeFromFile
+): Promise<ImageBitmap | HTMLImageElement> {
+  const image = new Image();
+  const imageUrl = getResourcePath(imageName);
+  let imageBitmap: ImageBitmap;
+
+  if (objectTypeFromFile === 'blob') {
+    // Load image file through fetch.
+    const blob = await loadXHR(imageUrl);
+    imageBitmap = await createImageBitmap(blob, { imageOrientation: 'from-image' });
```
kainino0x, line 706: `imageOrientation: 'from-image'` is the default and could be removed. I would probably remove it because otherwise as a reader I feel that I need to go look up what it does.

src/webgpu/web_platform/util.ts
```diff
@@ -606,3 +616,124 @@ export async function captureCameraFrame(test: GPUTest): Promise<VideoFrame> {
...
+      },
+    },
+  },
+} as const);
+
+type ImageName = keyof typeof kImageInfo;
+export const kImageNames: readonly ImageName[] = keysOf(kImageInfo);
+
+type ObjectTypeFromFile = (typeof kObjectTypeFromFiles)[number];
+export const kObjectTypeFromFiles = ['imageBitmap', 'image', 'blob'] as const;
```
kainino0x, line 685: Two of these return ImageBitmaps but it's hard to tell. How about:
```
ImageBitmap-from-Image
ImageBitmap-from-Blob
Image
```
or something like that?

(Update the function comment too, or just don't document the `@param`s so they can't get out of date)

src/webgpu/web_platform/util.ts
```diff
@@ -606,3 +616,124 @@ export async function captureCameraFrame(test: GPUTest): Promise<VideoFrame> {
...
+ * @param objectTypeFromFile: The object to load image file, e.g. ImageBitmap, blob
+ *
+ */
+export async function GetSourceFromImageFile(
+  imageName: ImageName,
+  objectTypeFromFile: ObjectTypeFromFile
+): Promise<ImageBitmap | HTMLImageElement> {
+  const image = new Image();
+  const imageUrl = getResourcePath(imageName);
+  let imageBitmap: ImageBitmap;
```
kainino0x, line 701: Some of these variables should be moved down into the blocks where they are used

src/webgpu/web_platform/util.ts
```diff
@@ -606,3 +616,124 @@ export async function captureCameraFrame(test: GPUTest): Promise<VideoFrame> {
...
+
+  if (objectTypeFromFile === 'blob') {
+    // Load image file through fetch.
+    const blob = await loadXHR(imageUrl);
+    imageBitmap = await createImageBitmap(blob, { imageOrientation: 'from-image' });
+    return imageBitmap;
+  } else {
+    // Load image file through HTMLImageElement.
+    image.src = imageUrl;
+    await raceWithRejectOnTimeout(image.decode(), 5000, 'decode image timeout');
```
kainino0x, line 711: `image.decode()` should not be called before `createImageBitmap()`, as that's redundant (both of them asynchronously decode the image, but there's no guarantee that `image.decode()` will decode it correctly for `createImageBitmap()`)

src/webgpu/web_platform/util.ts
```diff
@@ -606,3 +616,124 @@ export async function captureCameraFrame(test: GPUTest): Promise<VideoFrame> {
...
+    xhr.onload = function () {
+      if (xhr.status === 200) {
+        resolve(xhr.response);
+      } else {
+        reject(new Error('Loading error:' + xhr.statusText));
+      }
+    };
+    xhr.send();
+  });
+}
```
kainino0x, line 739: Suggestion incorporating all the comments above:

```suggestion
type ObjectTypeFromFile = (typeof kObjectTypeFromFiles)[number];
export const kObjectTypeFromFiles = [
  'ImageBitmap-from-Blob',
  'ImageBitmap-from-Image',
  'Image',
] as const;

/**
 * Load image file(e.g. *.jpg) from ImageBitmap, blob or HTMLImageElement. And
 * convert the result to valid source that GPUCopyExternalImageSource supported.
 */
export async function GetSourceFromImageFile(
  imageName: ImageName,
  objectTypeFromFile: ObjectTypeFromFile
): Promise<ImageBitmap | HTMLImageElement> {
  const imageUrl = getResourcePath(imageName);

  switch (objectTypeFromFile) {
    case 'ImageBitmap-from-Blob': {
      // Load image file through fetch.
      const response = await fetch(imageUrl);
      return createImageBitmap(await response.blob());
    }
    case 'ImageBitmap-from-Image':
    case 'Image': {
      // Load image file through HTMLImageElement.
      const image = new Image();
      image.src = imageUrl;
      if (objectTypeFromFile === 'Image') {
        await raceWithRejectOnTimeout(image.decode(), 5000, 'decode image timeout');
        return image;
      }

      return createImageBitmap(image);
    }
  }
}
```

src/webgpu/web_platform/copyToTexture/image_file.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+    - Image with rotation metadata
+    - Valid 'flipY' config in 'GPUCopyExternalImageSourceInfo' (named 'srcDoFlipYDuringCopy' in cases)
+    - TODO: partial copy tests should be added
+    - TODO: all valid dstColorFormat tests should be added.
+  `
+  )
+  .params(u =>
+    u //
+      .combine('imageName', kImageNames)
+      .combine('objectTypeFromFile', kObjectTypeFromFiles)
```
kainino0x, line 41: Please also make sure this test works as expected in various worker types, including skipping when the case is not possible in that worker type. `Image` (aka `HTMLImageElement`) doesn't exist in workers and I'm not sure if `createImageBitmap` works in all workers or not.

kainino0x (approved):
LGTM with a pile of small things

src/webgpu/web_platform/copyToTexture/image_file.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+} from '../util.js';
+
+export const g = makeTestGroup(TextureUploadingUtils);
+
+g.test('from_orientation_metadata_file')
+  .desc(
+    `
+    Test HTMLImageElements with rotation metadata can be copied to WebGPU texture correctly.
+
+    It creates Images with images under Resource folder.
```
kainino0x, line 23:
```suggestion
    It creates an ImageBitmap or HTMLImageElement using images in the 'resources' folder.
```

src/webgpu/web_platform/copyToTexture/image_file.spec.ts
kainino0x: Even `ImageBitmap-from-Blob` seems to fail on a service worker, I'm not sure why but probably due to some origin security thing? Needs investigating.

`http://localhost:8080/standalone/?worker=service&runnow=1&q=webgpu:web_platform,copyToTexture,image_file:from_orientation_metadata_file:imageName=%22four-colors.jpg%22;*`

src/webgpu/web_platform/util.ts
shaoboyan091: Yes. `coded ` are not needed right now. I'm Ok to remove them.

src/webgpu/web_platform/copyToTexture/image_file.spec.ts
shaoboyan091: Ah, thanks for reminding. Worker is not covered in my local try.

src/webgpu/web_platform/util.ts
shaoboyan091: The reason is that `createImageBitmap()` requires `HTMLImageElement` to be fully loaded. Or we might get error `Failed to execute 'createImageBitmap' on 'Window': The source image width is 0`. We can wait for `onload()` or wait for `image.decode()`. Waiting for `image.decode()` unify both uploading cases.

src/webgpu/web_platform/copyToTexture/image_file.spec.ts
shaoboyan091: The reason is that `serviceWorker` tries to load resource from `/out/webgpu/webworker/web_platform/copyToTexture/resources/four-colors.jpg` and it seems there is no folder in this place?

Maybe we should set `baseResourcePath`(https://github.com/gpuweb/cts/blob/main/src/common/framework/resources.ts#L93) correctly for service worker.

Left a maintainance_todo here and skip this test when launching tests in serviceWorker.

src/webgpu/web_platform/util.ts
```diff
@@ -650,24 +650,38 @@ export const kObjectTypeFromFiles = [
...
 export async function GetSourceFromImageFile(
+  test: GPUTest,
   imageName: ImageName,
   objectTypeFromFile: ObjectTypeFromFile
 ): Promise<ImageBitmap | HTMLImageElement> {
   const imageUrl = getResourcePath(imageName);
 
   switch (objectTypeFromFile) {
     case 'ImageBitmap-from-Blob': {
+      // MAINTENANCE_TODO: resource folder path when using service worker is not correct. Return
```
kainino0x, line 661: Since this TODO affects coverage, please make sure it also has a corresponding non-MAINTENANCE `TODO` in a test file description or test description.

src/webgpu/web_platform/util.ts
kainino0x: Ahh, I forgot about that. `onload` seems technically more appropriate, but the difference is subtle. I would almost suggest testing both, but that's probably out of scope of WebGPU testing.

The difference is so minimal that just using `decode` is fine and less confusing than trying to wait for `onload`. (Using `onload` is messy, but normally I would tell people they shouldn't use the HTMLImageElement-to-ImageBitmap path anyway.)

kainino0x (approved):

src/webgpu/web_platform/copyToTexture/image_file.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+    of dst texture, and read one pixel out to compare with the manually documented expected color.
+
+    If 'flipY' in 'GPUCopyExternalImageSourceInfo' is set to 'true', copy will ensure the result
+    is flipped.
+
+    The tests covers:
+    - Image with rotation metadata
+    - Valid 'flipY' config in 'GPUCopyExternalImageSourceInfo' (named 'srcDoFlipYDuringCopy' in cases)
+    - TODO: partial copy tests should be added
+    - TODO: all valid dstColorFormat tests should be added.
```
kainino0x, line 35:
```suggestion
    - TODO: all valid dstColorFormat tests should be added.
    - TODO(#4108): Make this work in service workers (see GetSourceFromImageFile)
```

kainino0x:
FYI I sent your new account an invite to gpuweb (https://github.com/orgs/gpuweb/people) so you should be able to land changes once you accept it.

shaoboyan091:
@kainino0x  Thanks!

>FYI I sent your new account an invite to gpuweb (https://github.com/orgs/gpuweb/people) so you should be able to land changes once you accept it.

Where can I find the invitation so that I could accept it?

kainino0x:
@shaoboyan091 it should be sent to the primary email address on your github account. I think this link _should_ work:
https://github.com/orgs/gpuweb/invitation (I can't verify because there's no open invitation for me to join the org)

PR #4112, opened 2024-12-19
Compat: refactor state_tracking test for 0 frag buffers.

greggman:
This is a first attempt. Feel free to push back and/or give ideas.

The original tests use 2 read-only-storage buffers and 1 read-write storage buffer. Each has a single i32 in it and generally they subtract the 2nd from the first `out = a - b`

Storage buffers in the fragment stage might not exist on some compat devices so the question is how to work around that and still test.

This solution is to add subcases, `storage` and `uniform`. The `storage` case is unchanged. The compute pass case will run in compat always. The render pass and render bundle cases only run in compat if the device supports storage buffers in the fragment stage.

The uniform cases use 2 uniform buffers and render to a single pixel r32sint texture. They then copy that texture to the `out` buffer that the original test was checking. This path needs no storage buffers in the fragment shader and so always runs.

This works but it's effectively only checking 2 bindings, not 3. So, the question is, should I add 3rd buffer and change the algo to `out =  a - b - c` etc.... so that we can shuffle more bindings? Or is this good enough? Or should I do something completely different. Maybe 3rd should be a 1x1 px `texture_2d<i32>` so it's a different type?

Also note: the last test 'compatible_pipelines' is unchanged and so only runs the comput pass unless the device supports storage buffers in fragment shaders.

I didn't update it yet because for it to work requires either (a) two render passes to render to 2 different render targets. Or it needs some viewport settings to render to 2 different pixels in the same target. Or something..., all of which seem like the might require some big refactors. In the `createEncoder` infra in gpu_test.ts or else they'd just have to do their own thing entirely.

Maybe that change doesn't need to happen in this PR but ideas are welcome.

### DIFF
--- a/src/webgpu/api/operation/command_buffer/programmable/state_tracking.spec.ts
+++ b/src/webgpu/api/operation/command_buffer/programmable/state_tracking.spec.ts
@@ -5,13 +5,18 @@ times in different orders) for setBindGroup and setPipeline.
 
 import { makeTestGroup } from '../../../../../common/framework/test_group.js';
 import { GPUConst } from '../../../../constants.js';
+import { MaxLimitsTestMixin } from '../../../../gpu_test.js';
 import { kProgrammableEncoderTypes } from '../../../../util/command_buffer_maker.js';
 
 import { ProgrammableStateTest } from './programmable_state_test.js';
 
-export const g = makeTestGroup(ProgrammableStateTest);
+export const g = makeTestGroup(MaxLimitsTestMixin(ProgrammableStateTest));
 
-const kBufferUsage = GPUConst.BufferUsage.COPY_SRC | GPUConst.BufferUsage.STORAGE;
+const kBufferUsage =
+  GPUConst.BufferUsage.COPY_SRC |
+  GPUConst.BufferUsage.COPY_DST |
+  GPUConst.BufferUsage.STORAGE |
+  GPUConst.BufferUsage.UNIFORM;
 
 g.test('bind_group_indices')
   .desc(
@@ -24,6 +29,7 @@ g.test('bind_group_indices')
     u //
       .combine('encoderType', kProgrammableEncoderTypes)
       .beginSubcases()
+      .combine('type', ['storage', 'uniform'] as const)
       .combine('groupIndices', [
         { a: 0, b: 1, out: 2 },
         { a: 1, b: 2, out: 0 },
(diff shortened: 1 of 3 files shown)


shrekshao:
> This works but it's effectively only checking 2 bindings, not 3. So, the question is, should I add 3rd buffer and change the algo to out =  a - b - c etc...

It's only testing 2 bindings because the uniform path changes the output buffer to render target? Am I understanding correctly?

But this only applies to the newly added uniform path. The storage path still checks 3 bindings. Test coverage is not decreasing. I think it's ... fine?

> the last test 'compatible_pipelines' is unchanged and so only runs the compute pass unless the device supports storage buffers in fragment shaders.

If we really want the coverage for no storage in fragment shaders devices, is having a seperate `g.test('compatible_pipelines,uniform')` test serving the purpose and easier to implement? (just use uniform buffer as input, render to target as output)

(I'm good with not adding one)

src/webgpu/api/operation/command_buffer/programmable/state_tracking.spec.ts
```diff
@@ -265,10 +338,13 @@ g.test('compatible_pipelines')
   )
   .fn(t => {
     const { encoderType } = t.params;
-    const pipelineA = t.createBindingStatePipeline(encoderType, { a: 0, b: 1, out: 2 });
+    t.skipIfNeedsStorageBuffersInFragmentStageAndHaveNone('storage', encoderType);
+
+    const pipelineA = t.createBindingStatePipeline(encoderType, { a: 0, b: 1, out: 2 }, 'storage');
```
shrekshao, line 343: nit: (although it's existing code) prefer having explicit 'a.value - b.value' passed in as algorithm, since it's needed context for the `t.expectGPUBufferValuesEqual(outA, new Int32Array([1]));`)

shrekshao (approved):
LGTM

src/webgpu/api/operation/command_buffer/programmable/state_tracking.spec.ts
greggman: done

greggman:
> > This works but it's effectively only checking 2 bindings, not 3. So, the question is, should I add 3rd buffer and change the algo to out =  a - b - c etc...
> 
> It's only testing 2 bindings because the uniform path changes the output buffer to render target? Am I understanding correctly?

correct


> But this only applies to the newly added uniform path. The storage path still checks 3 bindings. Test coverage is not decreasing. I think it's ... fine?

yea, no coverage is lost
 
> > the last test 'compatible_pipelines' is unchanged and so only runs the compute pass unless the device supports storage buffers in fragment shaders.
> 
> If we really want the coverage for no storage in fragment shaders devices, is having a seperate `g.test('compatible_pipelines,uniform')` test serving the purpose and easier to implement? (just use uniform buffer as input, render to target as output)
> 
> (I'm good with not adding one)

agreed, we can add one later if needed.

PR #4117, opened 2024-12-23
Create bind group

greggman:
Note: Several tests just switched from FRAGMENT
to COMPUTE as nothing about the test seemed dependent
on using the FRAGMENT stage. We could add a stage parameter
if we want to test in all stages. For now I just switched
those tests that didn't need FRAGMENT to use COMPUTE.

For those that were testing all stages I added a parameter
to test all vs only compute with the appropriate skipIf
if the resources are not available.


toji (approved):
LGTM

PR #4121, opened 2024-12-27
Compat: createBGL, createPipelineLayout, etc...

greggman:
Refactor createBindGroupLayout, createPipelineLayout, vertex_state,correctness, and requestDevice tests for 0 storage buffers/textures in vertex/fragment stage.

Note: I needed to add `maxStorage(Buffers/Textures)In(Fragment/Vertex)Stage` to `capability_info.ts`. That had cascading effects which is why so many files are changed.

For one, since these new limits are marked as optional, any place that uses limits has to check it's not undefined or least add a `!` to tell TS to stop complaining. I'm kind of wondering if we should change them to be required. They'll be required in the spec eventually, or at least that's the plan. In any case, this adds them.

Another issue was the existing structure of `kPerStageBindingLimits` in `capability_info.ts`. I refactored that to have per stage limits and added `getBindingLimitsForBindingType` where you pass in the type you're using and the visibility and returns the limit for those stages.

I also moved `getDefaultLimit/s` from GPUTestBase to LimitTestImpl as it seems like only the limits test should care about the default. All other tests should use what's on the device.

### DIFF
--- a/src/webgpu/api/operation/adapter/requestDevice.spec.ts
+++ b/src/webgpu/api/operation/adapter/requestDevice.spec.ts
@@ -274,7 +274,7 @@ g.test('limits,supported')
         result = value;
         break;
       case 'adapter':
-        value = adapter.limits[limit];
+        value = adapter.limits[limit]!;
         result = value;
         break;
       case 'undefined':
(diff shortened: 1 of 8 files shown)


shrekshao (approved):
LGTM

Issue #4120, opened 2024-12-27
Add options to force MaxLimitsTestMixin to default


greggman:
As I refactor the tests so they run/skip if there are no storage buffers or no storage textures in vertex/fragment shaders I first make sure the test runs with none, but then I switch the test to use `MaxLimitsTestMixin` so that tests of functionality are not skipped on devices that do actually support the,

Unfortunately, that means it's easy to write a test that doesn't work if no storage buffers or no storage textures are supported in the current device. Such devices are not tested as far as I know. 

New tests, especially of based on existing test, might use `MaxLimitsTestMixin` to make sure they are testing as much as possible but those same tests might fail if the limit is 0. So, maybe we should add an option to the CTS run the tests where `MaxLimitsTestMixin` doesn't ask for the max, or maybe even lower-level, we could hack `requestAdapter` so it returns the defaults. That way, there would be an easy way to check the tests run when these values are at their minimums.

PR #4124, opened 2024-12-30
Add "enforceDefaultLimits" flag

greggman:
I'm not sure what to do about this but dealing with dependent limits is kind of a PITA.

Tests that deal with storage buffers and storage textures need to take into account that they might have 0. But, they also need to be tested with the maxiumum number of storage buffers/textures.

So, if you set the test to use max storage buffers/textures then, unless you have a device that supports 0 you have no easy way to test that the test functions when the limit is 0.

While refactoring the tests I start without requesting the limit so on compat I get zero. Once that works I add `MaxLimitsTestMixin` or similar to request the maximum limits, otherwise a compat device would get no coverage.

But, now I have the issue that if I'm modifying the test I need to remember to test with 0 so I have to go manually comment out the code that's requesting max limits.

So, I thought I'd add this option.

Unfortunately, the default limits are defined in src/webgpu/capability_info.ts which is listed as off limits to navigator_gpu.ts. I tried moving the limits table to src/common/limits.ts but then src/webgpu/capabilities_info.ts complained that src/common/limits.ts is off limits. So, I'm not sure where I can put a shared limits.ts file that both navigator_gpu.ts and capabilities_info.ts can access.

In the webgpu-dev-extension, I implemented this by requesting 2 adapters and a temp device.

```js
// pseudo code
tempAdapter = await requestAdapter(options);
tempDevice = await requestDevice(...)
const defaultLimits = tempDevice.limits;
tempDevice.destroy();
adapter = await requestAdapter(options)
adapter.limits = defaultLimits;
```

But I wasn't sure that was a good solution in the CTS vs using the limits table.

Also, in the webgpu-dev-extension I don't really care about actually enforcing the limits but in the cts, the limits tests in src/webgpu/api/validation/capability_checks/limits will fail if the limits are not enforced and I needed this flag to be able to debug them and make sure they work when a limit is 0.

Issue: #4120


kainino0x:
> Unfortunately, the default limits are defined in src/webgpu/capability_info.ts which is listed as off limits to navigator_gpu.ts. I tried moving the limits table to src/common/limits.ts but then src/webgpu/capabilities_info.ts complained that src/common/limits.ts is off limits. So, I'm not sure where I can put a shared limits.ts file that both navigator_gpu.ts and capabilities_info.ts can access.

I think `src/common/util/limits.ts` would be allowed. I don't remember the rule syntax, but the message seems to imply that:

```
              {
                "target": "./src/webgpu",
                "from": "./src/common",
                "except": ["./framework", "./util"],
                "message": "Non-framework common/ code imported from webgpu/ suite"
              },
```

src/common/util/navigator_gpu.ts
```diff
@@ -60,6 +65,64 @@ export function getGPU(recorder: TestCaseRecorder | null): GPU {
...
+        Object.defineProperty(adapter, 'limits', {
+          get() {
+            return limits;
+          },
+        });
+      }
+      return adapter;
+    };
+
+    if (!s_wrappedForEnforceDefaultLimits) {
```
kainino0x, line 90: Not sure this is necessary? We already have `if (impl) { return impl; }` at the top that should prevent this from running multiple times.

kainino0x (approved):

src/common/util/navigator_gpu.ts
greggman: You're right, I missed that. I'll remove this check

kainino0x (approved):

greggman:
So I wrote this change before the update to the limits to add `maxXXXInYYYStage`. With those added this code broke because it would request limits that don't exist.

So, I made it so it makes a temp device. This might be fine but, tests that test `low-power` vs `high-performance` may have issues. I'd either have to always request a temp adapter and temp device OR, I'd have to turn the requestAdapter descriptor into a key to cache the limits. Can that wait or later or should I do something now?

src/common/runtime/helper/options.ts
```diff
@@ -100,6 +102,7 @@ export const kCTSOptionsInfo: OptionsInfos<CTSOptions> = {
   debug: { description: 'show more info' },
   compatibility: { description: 'run in compatibility mode' },
   forceFallbackAdapter: { description: 'pass forceFallbackAdapter: true to requestAdapter' },
+  enforceDefaultLimits: { description: 'force the adapter limits to the default limits' },
```
kainino0x, line 105: > So, I made it so it makes a temp device. This might be fine but, tests that test `low-power` vs `high-performance` may have issues.

That seems fine since I don't think this option isn't necessarily meant to work with every test. Perhaps make a note of that here though.

kainino0x (commented):
This is fine, though I don't quite understand why it's needed. In requestAdapter, after having already called origRequestAdapterFn, we already have the list of supported limits.

greggman:
> This is fine, though I don't quite understand why it's needed. In requestAdapter, after having already called origRequestAdapterFn, we already have the list of supported limits.

I'm not sure what you're saying is not needed. An actual adapter will always have max limits. We create a device with no `requiredLimits` to get the default limits. We then create an adapter and set all `adapter.limits` to the defaults so down stream it won't see the max limits. There are tests that check things throw if we request higher than max. These tests will fail to throw if we don't manually enforce that the max is the default (not the actual max). That's why we need the defaults.

src/common/runtime/helper/options.ts
greggman: added

kainino0x:
> I'm not sure what you're saying is not needed. An actual adapter will always have max limits. We create a device with no `requiredLimits` to get the default limits. We then create an adapter and set all `adapter.limits` to the defaults so down stream it won't see the max limits. There are tests that check things throw if we request higher than max. These tests will fail to throw if we don't manually enforce that the max is the default (not the actual max). That's why we need the defaults.

Sorry, I mean I'm not sure why a temp device is needed. We know the default value for each limit from our info tables (as long as we know which `featureLevel` to use). From any adapter we can see which limits are supported by the browser (by whether they exist in `adapter.limits`). So in the shim for `adapter.limits` we can return only the limit names that the original adapter had, with the values reduced to the defaults.

Though I'm not sure if what I'm saying makes sense because I haven't tried it.

greggman:
That makes sense. I can put up a new PR to use the limits table. 

another thing we could do is just set all the max limits to 0 since the impl is supposed to set them to the default. For min limits though, some have restrictions like power-of-2 so I guess we'd still need the table.

PR #4127, opened 2025-01-02
Add texture_builtin_notes.md

greggman:
I'm not sure this belongs here. Maybe it should be moved to `/docs` or maybe it shouldn't exist in the repo at all. I thought about just making an issue in gpuweb/gpuweb/issues and posting each failure as a reply under that issue.

Thoughts?

### DIFF
--- a/src/webgpu/shader/execution/expression/call/builtin/texture_builtin_notes.md
+++ b/src/webgpu/shader/execution/expression/call/builtin/texture_builtin_notes.md
@@ -1,12 +1,29 @@
-# Texture Builtin Notes
+# Texture Builtins Issues
 
-These are known failures for various GPUs. If you find more, please add to this list so developers
-can be aware what does and does not work.
+Various GPUs and drivers have issues (bugs) related sampling textures. Below is a list of known issues. 
+If you'd like to check your own GPU you can run
+[all of the WGSL builtin function tests](https://gpuweb.github.io/cts/standalone/?debug=1&runnow=1&q=webgpu:shader,execution,expression,call,builtin,*)
+or you can run the individual texture builtin function tests
+
+* [textureLoad](https://gpuweb.github.io/cts/standalone/?debug=1&runnow=1&q=webgpu:shader,execution,expression,call,builtin,textureLoad:*)
+* [textureGather](https://gpuweb.github.io/cts/standalone/?debug=1&runnow=1&q=webgpu:shader,execution,expression,call,builtin,textureGather:*)
+* [textureGatherCompare](https://gpuweb.github.io/cts/standalone/?debug=1&runnow=1&q=webgpu:shader,execution,expression,call,builtin,textureGatherCompare:*)
+* [textureSample](https://gpuweb.github.io/cts/standalone/?debug=1&runnow=1&q=webgpu:shader,execution,expression,call,builtin,textureSample:*)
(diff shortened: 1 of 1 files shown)


greggman:
@kainino0x @Kangz @toji @kdashg @mwyrzykowski

toji:
I think this is a valuable thing to have in the repo. I'll let others weigh in on whether it belongs here or in `/docs`, though I'm leaning towards keeping it where you've got it (close to the tests that it's relevant to.)

src/webgpu/shader/execution/expression/call/builtin/texture_builtin_notes.md
```diff
@@ -0,0 +1,918 @@
+# Texture Builtin Notes
+
+These are known failures for various GPUs. If you find more, please add to this list so developers
+can be aware what does and does not work.
+
+------------------------------------------------------------------------------------------------------------
+## gpu:Intel Comet Lake S UHD Graphics 630 (8086:9bc5-31.0.101.2127)
+## os:Windows-10
+## affected: `textureGather` with cube and cube-arrays.
```
kainino0x, line 9: nit: confusing in the rendered version because it looks like there are two empty sections. Maybe use
```md
## A<br>B<br>C
```

or just put the details inside the section body

src/webgpu/shader/execution/expression/call/builtin/texture_builtin_notes.md
```diff
@@ -0,0 +1,918 @@
+# Texture Builtin Notes
+
+These are known failures for various GPUs. If you find more, please add to this list so developers
+can be aware what does and does not work.
```
kainino0x, line 4: If we want this to be discoverable by developers then maybe it should be on some github wiki or something. I'm not sure where is a good place for it.

Otherwise having it here is fine, though I'd add a mention of this file in one of the READMEs or test description files so it's not "unreferenced" from the test suite.

greggman: wiki works for me.

greggman: I started a wiki here

https://github.com/gpuweb/gpuweb/wiki/Texture-Builtins-Issues

I thought about making one page per issue and linking them from the that page. Or I can just past them all in that one page.

kainino0x: Personally I think one page is better, github wikis don't really handle having many pages that well.

As for this PR, it'll still useful to add a link to that wiki page in one of the readmes/descriptions.

greggman: I posted the issue is the wiki but unfortunately it's not really readable as github gives the wiki a maximum width. Maybe the wiki should just link to this and/or, maybe we could make it available at `https://gpuweb.github.io/cts/standalone/src/webgpu/shader/execution/expression/call/builtin/texture_builtin_notes.md`

kainino0x: Ah, shame.

Linking is fine, we could have a gpuweb wiki page for errata, and link directly to the .md file in this repo. Don't think there's any need to make it work on gpuweb.github.io.

kainino0x (approved):

PR #4137, opened 2025-01-10
Refactor maxStorage(Buffers/Textures)PerShaderStage tests

greggman:
These tests can't just set `maxStorage(Buffers/Textures)In(Fragment/Vertex)Stage to the maximum limit because those can't be set above their `..PerShaderStage` counterpart.

Further, these were trying to use writable storage in the vertex stage which is not allowed.

### DIFF
--- a/src/webgpu/api/validation/capability_checks/limits/limit_utils.ts
+++ b/src/webgpu/api/validation/capability_checks/limits/limit_utils.ts
@@ -343,7 +343,7 @@ export const kMinimumLimitBaseParams = kUnitCaseParamsBuilder
  * Adds a maximum limit upto a dependent limit.
  *
  * Example:
- *   You want to test `maxStorageBuffersPerShaderStage` in fragment stagee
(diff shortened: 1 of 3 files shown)


src/webgpu/api/validation/capability_checks/limits/limit_utils.ts
```diff
@@ -338,6 +339,53 @@ export const kMinimumLimitBaseParams = kUnitCaseParamsBuilder
   .combine('limitTest', kMinimumLimitValueTests)
   .combine('testValueName', kMinimumTestValues);
 
+/**
+ * Adds a maximum limit upto a dependent limit.
+ *
+ * Example:
+ *   You want to test `maxStorageBuffersPerShaderStage` in fragment stagee
```
shrekshao, line 346: nit: typo 'stagee' -> 'stage'

shrekshao (approved):
LGTM

PR #4141, opened 2025-01-10
Adjust print_environment for node

greggman:
node doesn't have a global `navigator` variable.


kainino0x (approved):

greggman:
fyi: I was using this test to verify that `--enforce-default-limits` was working ([cl](https://dawn-review.googlesource.com/c/dawn/+/221754)) when I noticed dawn.node was not making these enumerable. Fixed [here](https://dawn-review.googlesource.com/c/dawn/+/221755)

PR #4149, opened 2025-01-16
Compat: Refactor draw test for 0 storage buffers

greggman:
Refactored one test to only optionally use a storage buffer. Refactored another test to be skipped if no storage buffers in fragment shaders.

It's possible the 2nd test is not needed as it seems similar things are tested in src/webgpu/api/operation/vertex_state/correctness.spec.ts.

It's also possible the first test's case of seeing that the fragment shader is not called for certain cases is also tested elsewhere.


shrekshao (approved):

shrekshao:
Yeah it is actually coverred elsewhere. If it fails, vast cts tests should have failed anyway

PR #4150, opened 2025-01-16
Refactor selectDeviceOrSkipTest system

greggman:
The existing system selected the device immediately which means code like this

    t.selectDeviceOrSkipTest('timestamp-query');
    t.selectDeviceOrSkipTesT('float32-renderable'); // fail!

would fail on the 2nd line because a device had already been requested on the first line.

Refactored so that the various requirements are merged and only at the end is a device requested.

Tested here:
https://github.com/greggman/cts/commit/d4abf411a78ba4d79f712fa2ad3d82a324b99719


kainino0x (approved):
lgtm!

PR #4155, opened 2025-01-22
Compat: Fix pipeline_bind_group_compat.spec.ts for 0 storage bufs/texs

greggman:


src/webgpu/api/validation/encoding/programmable/pipeline_bind_group_compat.spec.ts
```diff
@@ -777,7 +777,6 @@ g.test('bgl_resource_type_mismatch')
   )
   .params(
     kCompatTestParams
-      .beginSubcases()
```
shrekshao, line 780: our skips are inside `fn` so you don't have to remove the `beginSubcases()`, right?

greggman: thanks for catching that. I deleted that by mistake as I had commented it out to run just one subcase. Will fix.

greggman: fixed

shrekshao (approved):

PR #4156, opened 2025-01-22
Skip web_platform tests if API does not exist

greggman:
Skip tests if `navigator`, `OffscreenCanvas`, `ImageData`, etc... do not exist.

This is mostly for dawn.node and similar environments.

If these tests are supposed to fail then that's fine. We can discard this PR

### DIFF
--- a/src/webgpu/web_platform/canvas/getPreferredCanvasFormat.spec.ts
+++ b/src/webgpu/web_platform/canvas/getPreferredCanvasFormat.spec.ts
@@ -2,7 +2,7 @@ export const description = `
 Tests for navigator.gpu.getPreferredCanvasFormat.
 `;
 
-import { Fixture } from '../../../common/framework/fixture.js';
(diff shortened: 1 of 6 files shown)


src/webgpu/web_platform/canvas/getPreferredCanvasFormat.spec.ts
```diff
@@ -13,6 +13,11 @@ g.test('value')
     Ensure getPreferredCanvasFormat returns one of the valid values.
     `
   )
+  .beforeAllSubcases(t => {
+    if (typeof navigator === 'undefined') {
+      throw new SkipTestCase('navigator does not exist in this environment');
```
greggman, line 18: maybe `skipIf` should be moved to `SubcaseBatchState` from `GPUTestSubcaseBatchState` in another PR.

kainino0x: SGTM

kainino0x (approved):

PR #4159, opened 2025-01-23
Compat: fix more float16(32)-renderable tests

shrekshao:


src/webgpu/api/operation/texture_view/write.spec.ts
```diff
@@ -351,6 +351,8 @@ TODO: Test rgb10a2uint when TexelRepresentation.numericRange is made per-compone
         t.skipIfTextureFormatNotUsableAsStorageTexture(format);
         break;
     }
+
+    t.selectDeviceForRenderableColorFormatOrSkipTestCase(format);
```
greggman, line 355: Isn't this only if the method is `render-pass-resolve`, `render-pass-store`?

shrekshao: Fixed. Also fix multisample issue

greggman (approved):

PR #4163, opened 2025-01-28
Compat: fix float16(32)-renderable tests (more in validation, shader,…

shrekshao:


src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -281,12 +281,17 @@ g.test('sampleCount,various_sampleCount_with_all_formats')
...
     const info = kTextureFormatInfo[format];
     t.skipIfTextureFormatNotSupported(format);
+    t.selectDeviceForRenderableColorFormatOrSkipTestCase(format);
     t.selectDeviceOrSkipTestCase(info.feature);
   })
   .fn(t => {
     const { dimension, sampleCount, format } = t.params;
     const info = kTextureFormatInfo[format];
 
+    if (sampleCount > 1) {
```
greggman, line 291: should this instead just move to the `success` calculation?

src/webgpu/api/validation/texture/destroy.spec.ts
```diff
@@ -43,6 +43,8 @@ g.test('invalid_texture')
     invalidTexture.destroy();
   });
 
+const kColorTextureFormat: GPUTextureFormat = 'rgba32float';
```
greggman, line 46: does this need to be `rgba32float`? Will `rgba8unorm` work just as well for this test?

src/webgpu/shader/execution/expression/call/builtin/textureLoad.spec.ts
```diff
@@ -376,6 +376,7 @@ Parameters:
     t.skipIfTextureFormatNotSupported(format);
     t.skipIfTextureLoadNotSupportedForTextureType(texture_type);
     t.skipIfMultisampleNotSupportedForFormat(format);
+    t.selectDeviceForRenderableColorFormatOrSkipTestCase(format);
```
greggman, line 379: This check seems, at least by name, strange. This test isn't supposed to be rendering to  the format. If it is, there's a bug or a workaround that needs to be added. Otherwise it won't test that `textureLoad` works with this format.

greggman (changes requested):

src/webgpu/api/validation/createTexture.spec.ts
shrekshao: You are right. Fixed

src/webgpu/api/validation/texture/destroy.spec.ts
shrekshao: It also uses storage "read-write" usage. So rgba8unorm won't work without test modification.

src/webgpu/shader/execution/expression/call/builtin/textureLoad.spec.ts
shrekshao: It's that multisampled texture requires `GPUTextureUsage.RENDER_ATTACHMENT`. Validation fails texture with such usage when texture in not color renderable.

The usage is added: https://github.com/gpuweb/cts/blob/main/src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts#L3152

src/webgpu/api/validation/texture/destroy.spec.ts
greggman: I'm not seeing this "read-write" usage. Am I missing it? To put it another way, I'm not seeing what breaks if this changes to `rgba8unorm`. All it tries to do is render to a destroyed texture. Any renderable texture format should cover that right?

src/webgpu/shader/execution/expression/call/builtin/textureLoad.spec.ts
greggman: This still seems like the wrong check. `t.skiptIfMultisampleNotSupportForFormat` should be enough because there's no such thing as a non-renderable multisample texture since the only way to use them is to render to them. It seems like `skipIfMultisampleNotSupported` should check this condition. If you can't render to it then multisample is not supported.

greggman (changes requested):

src/webgpu/shader/execution/expression/call/builtin/textureLoad.spec.ts
shrekshao: You are right.

Fixed by:
* rename `skipIfMultisampleNotSupportedForFormat` -> `skipIfMultisampleNotSupportedForFormatOrSelectDevice`
  * select device or skip to ensure the format is colorRenderable
* Fix callsites: Remove `selectDeviceForRenderableColorFormatOrSkipTestCase` calls where `skipIfMultisampleNotSupportedForFormatOrSelectDevice` is already called
* Fixes for previous commits
  * remove redundant `GPUTestSubcaseBatchState.skipIfColorRenderableNotSupportedForFormat` and `GPUTestBase.skipIfMultisampleNotSupportedForFormatOrSelectDevice`
  * Fix `skipIfColorRenderableNotSupportedForFormat` by checking device.features

src/webgpu/api/validation/resource_usages/texture/in_render_common.spec.ts
```diff
@@ -115,6 +115,11 @@ g.test('subresources,color_attachments')
       .combine('inSamePass', [true, false])
       .unless(t => t.inSamePass && t.level0 !== t.level1)
   )
+  .beforeAllSubcases(t => {
+    if (t.isCompatibility) {
```
greggman, line 119: This feels like the `if (t.isCompatibility)` check should be moved to `selectDeviceForRendrableColorFormatOrSkipTestCase`. It's on the test, it knows if it's in compat mode or not.

greggman (approved):
thanks for all the changes. one nit. lgtm!

src/webgpu/api/validation/resource_usages/texture/in_render_common.spec.ts
shrekshao: Yes this is redundant. removed

PR #4164, opened 2025-01-28
Check more JavaScript features.

greggman:
getter_replacement fails on dawn.node. If I understand correctly it's because the getters are placed on the instances, not on the class prototype.

### DIFF
--- a/src/webgpu/idl/javascript.spec.ts
+++ b/src/webgpu/idl/javascript.spec.ts
@@ -385,7 +385,7 @@ g.test('getter_replacement')
       // check replacing the getter on constructor works.
       const ctorPrototype = obj.constructor.prototype;
       const origProperties = Object.getOwnPropertyDescriptor(ctorPrototype, getter);
-      t.expect(
+      assert(
         !!origProperties,
         `Object.getOwnPropertyDescriptor(${type}, '${getter}') !== undefined`
       );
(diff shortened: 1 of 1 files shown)


src/webgpu/idl/javascript.spec.ts
```diff
@@ -338,3 +355,110 @@ g.test('limits')
...
+        const value = (obj as unknown as Record<string, string>)[getter];
+        t.expect(
+          value === 'testGetterValue',
+          `replacing getter: '${getter}' on ${type} returns test value`
+        );
+      } finally {
+        Object.defineProperty(ctorPrototype, getter, origProperties!);
+      }
+
+      // Check it turns the same value after restoring as before restoring.
```
kainino0x, line 407: Is it possible to (also/instead) check that the getter function itself has been restored, using `getOwnPropertyDescriptor(getter).get === oldGetter`?

Technically a getter can return different things each time it's called (though I think none of the getters we have now can do that).

src/webgpu/idl/javascript.spec.ts
```diff
@@ -338,3 +355,110 @@ g.test('limits')
...
+            return 'testGetterValue';
+          },
+        });
+        const value = (obj as unknown as Record<string, string>)[getter];
+        t.expect(
+          value === 'testGetterValue',
+          `replacing getter: '${getter}' on ${type} returns test value`
+        );
+      } finally {
+        Object.defineProperty(ctorPrototype, getter, origProperties!);
```
kainino0x, line 404: Line 388 should `assert` instead of `expect`, because it doesn't really make sense to continue the test if `getOwnPropertyDescriptor` didn't work.

This restoration code will fail if `origProperties` is undefined, leaving the page in a different state than it started.
This should also remove the need for `!` on `origProperties!`.

src/webgpu/idl/javascript.spec.ts
```diff
@@ -338,3 +355,110 @@ g.test('limits')
...
+    const { requiredKeys, getters } = kResourceInfo[type];
+    const gettersSet = new Set<string>(getters);
+    const methods = requiredKeys.filter(k => !gettersSet.has(k));
+
+    const obj = createResource(t, type);
+    for (const method of methods) {
+      const ctorPrototype = obj.constructor.prototype;
+      const origFunc = ctorPrototype[method];
+
+      t.expect(typeof origFunc === 'function', `${type}.prototype.${method} is a function`);
```
kainino0x, line 436: nit: probably also `assert` here though it doesn't matter very much in this case

kainino0x (approved):

src/webgpu/idl/javascript.spec.ts
greggman: done

src/webgpu/idl/javascript.spec.ts
greggman: done

PR #4165, opened 2025-01-28
Test 1 layer 2d-array textures.

greggman:
1 layer 2d-array textures were not tested. A backend might have issues where it incorrectly uses "layers > 1" to distinguish 2d-arrays. So test.


shrekshao (approved):
LGTM

PR #4169, opened 2025-01-31
Test properties that are supposed to be readonly are.

greggman:
In JavaScript these properties are supposed to be readonly and generate a "`TypeError`, no setter". This test tests that.

Interestingly, in the Chrome dev console, it doesn't assign them but it doesn't generate an error, which mis-lead me into thinking the behavior was different between dawn.node and the browser. Fortunately the test passes in both.

### DIFF
--- a/src/webgpu/idl/javascript.spec.ts
+++ b/src/webgpu/idl/javascript.spec.ts
@@ -377,6 +377,7 @@ g.test('readonly_properties')
   )
   .params(u => u.combine('type', kResources))
   .fn(t => {
+    'use strict';
     const { type } = t.params;
     const { getters, settable } = kResourceInfo[type];
     const obj = createResource(t, type);
--- a/src/webgpu/idl/javascript.spec.ts
+++ b/src/webgpu/idl/javascript.spec.ts
@@ -84,7 +84,13 @@ class AllFeaturesTest extends GPUTest {
   }
 }
 
-const kResourceInfo = {
+type ResourceInfo = {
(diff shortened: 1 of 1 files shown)


src/webgpu/idl/javascript.spec.ts
```diff
@@ -356,6 +369,38 @@ g.test('limits')
...
   });
 
+g.test('readonly_properties')
+  .desc(
+    `
+    Test that setting a property with no setter throws.
+    `
+  )
+  .params(u => u.combine('type', kResources))
+  .fn(t => {
```
kainino0x, line 379: nit: Our build settings enable `'use strict'`, but best to do it here explicitly, just in case someone were to build it with different options.

If there is a compiler warning or something (since the compiler knows this won't actually do anything in our build), you can suppress it if possible, otherwise never mind.

```suggestion
  .fn(t => {
    'use strict'; // This makes setting a readonly property produce a TypeError.
```

kainino0x (approved):
> Interestingly, in the Chrome dev console, it doesn't assign them but it doesn't generate an error, which mis-lead me into thinking the behavior was different between dawn.node and the browser. Fortunately the test passes in both.

The difference is whether you are in a `'use strict';` JS context or not. If I do this in the dev tools console I get a TypeError:

`(() => { 'use strict'; buf.mapState = 2; })()`

src/webgpu/idl/javascript.spec.ts
greggman: added

Issue #4139, opened 2025-01-10
Add wgslLanguageFeatures test


sagudev:
test that returned object is same object (is this done as part of idl tests?)

greggman:
added. ok to close?

sagudev:
> added. [#4175](https://github.com/gpuweb/cts/pull/4175)
> 
> Ok to close?

Yes, I just tested.

BTW there is [webgpu:api,operation,device,lost:same_object:*](https://gpuweb.github.io/cts/standalone/?q=webgpu:api,operation,device,lost:same_object:*) and [webgpu:api,operation,adapter,info:same_object:*](https://gpuweb.github.io/cts/standalone/?q=webgpu:api,operation,adapter,info:same_object:*), but only last one is also done in [webgpu:idl,javascript:sameObject:*](https://gpuweb.github.io/cts/standalone/?q=webgpu:idl,javascript:sameObject:*)

PR #4175, opened 2025-02-03
Test SameObject properties

greggman:
Issue: #4139


shrekshao (approved):

kainino0x (commented):
I think we had some of these tests already as one-offs; maybe it makes sense to remove the redundant ones? (Not a big deal though, they should be extremely fast)

PR #4184, opened 2025-02-10
Add `AllFeaturesMaxLimitsGPUTest`

greggman:
This is a test that requests all features and maximum limits. This should be the default test for the majority of tests, otherwise optional features will not be tested. The exceptions are only tests that explicitly test the absence of a feature or specific limits such as the tests under validation/capability_checks.

As a concrete example to demonstrate the issue, texture format `rg11b10ufloat` is optionally renderable and can optionally be used multisampled. Any test that tests texture formats should test this format, skipping only if the feature is missing. So, the default should be that the test tests `kAllTextureFormats` with the appropriate filters from format_info.ts or the various helpers. This way, `rg11b10ufloat` will included in the test and fail if not appropriately filtered. If instead you were to use GPUTest then `rg11b10ufloat` would just be skipped as its never enabled. You could enable it manually but that spreads enabling to every test instead of being centralized in one place, here.

Honestly, I'd prefer to rename `GPUTest` to `SpecialGPUTest` or possibly even `DeprecatedGPUTest` it clear not to use it and where it needs to be fixed and then name this `GPUTest`. But, we can do that later. For now, I just effectively replaced `MaxLimitsTestMixin` with `AllFeaturesMaxLimitsGPUTest` in the places where it could easily be replaced.

The remaining places bring up an issue. All of them are based on `ValidationTest`. I think the same rules apply. For example, if we're validating that textures fail validation for certain reasons we should be checking formats that are optional also pass the same validation rules. The simplest way to do that would be to make `ValidationTest` inherit from `AllFeaturesMaxLimitsGPUTest`.

Unfortunately, the capability_checks/* are all based off `ValidationTest` so they need to be moved to something else before we can make that change.

Issue: #4178

<hr>

### DIFF
--- a/src/webgpu/api/operation/command_buffer/programmable/programmable_state_test.ts
+++ b/src/webgpu/api/operation/command_buffer/programmable/programmable_state_test.ts
@@ -1,5 +1,5 @@
 import { unreachable } from '../../../../../common/util/util.js';
-import { GPUTest, GPUTestBase } from '../../../../gpu_test.js';
+import { AllFeaturesMaxLimitsGPUTest, GPUTestBase } from '../../../../gpu_test.js';
 import { EncoderType } from '../../../../util/command_buffer_maker.js';
 
 interface BindGroupIndices {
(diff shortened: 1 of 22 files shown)


shrekshao:
> Honestly, I'd prefer to rename GPUTest to SpecialGPUTest or possibly even DeprecatedGPUTest it clear not to use it and where it needs to be fixed and then name this GPUTest. But, we can do that later. For now, I just effectively replaced MaxLimitsTestMixin with AllFeaturesMaxLimitsGPUTest in the places where it could easily be replaced.

Sounds good

shrekshao (approved):
LGTM

PR #4185, opened 2025-02-11
Skip color-space-conversion test if no ImageData

greggman:


src/webgpu/util/texture/color_space_conversions.spec.ts
```diff
@@ -34,6 +34,7 @@ g.test('util_matches_2d_canvas')
   )
   .fn(t => {
     const { srcColorSpace, dstColorSpace } = t.params;
+    t.skipIf(typeof ImageData === 'undefined', `ImageData does not exist in this environment`);
```
shrekshao, line 37: Is it because `ImageData` is not available in Web worker?

Then maybe skip it in 
```
.beforeAllSubcases(t => {
    t.skipIf(typeof ImageData === 'undefined', `ImageData does not exist in this environment`);
  })
```

shrekshao (approved):
LGTM with comment

src/webgpu/util/texture/color_space_conversions.spec.ts
greggman: It's because it's not available in node.

switched to `beforeAllSubcases`

Issue #4181, opened 2025-02-07
Refactor format_info.ts and related texture functions


greggman:
The CTS has the issue that there's a global table of texture data `kAllTextureFormatInfo` and it's often used to filter/select formats. Example

```js
      .combine('colorFormat', kEncodableTextureFormats)
      // Filter out any non-renderable formats
      .filter(({ colorFormat }) => !!kTextureFormatInfo[colorFormat].colorRender)
```

But this is wrong. `rg11b10ufloat` is conditionally renderable so as defined it will never be tested here.

Similarly there is code like this

```js
export function isMultisampledTextureFormat(
  format: GPUTextureFormat,
  isCompatibilityMode: boolean
): boolean {
  if (isCompatibilityMode) {
    if (kCompatModeUnsupportedMultisampledTextureFormats.indexOf(format) >= 0) {
      return false;
    }
  }
  return kAllTextureFormatInfo[format].multisample;
}
```

This is also wrong. `rg11b10ufloat` is marked as `multisample: false` but enabling `rg11b10ufloat-renderable` is supposed to make it support multisampling which means, every test using this filter will not be testing this format so we'd have no idea if it actually works.

It feels like this should be solved by making `kAllTextureFormatInfo` private and refactoring any of these queries about formats to require enough info to be useful, and/or, be renamed to reflect what they're returning. So for example, something that wants to know all renderable formats can not use

```js
export function canUseAsRenderTarget(format: GPUTextureFormat) {
  return !!kTextureFormatInfo[format].colorRender || isDepthOrStencilTextureFormat(format);
}

   .combine('format', kAllTextureFormats)
   .filter(t => canUseAsRenderTarget(t.format))
```

because, you can't know if a format is renderable until you know if you have the feature. That particular function could be renamed to `canPotentiallyUseAsRenderTarget` in which case it should return true for `rg11b10ufloat`. Similarly `isMultisampledTextureFormat` would need to be `isPotentiallyMultisampledTextureFormat`.

But ultimately, with all of these special exceptions, it doesn't feel like tests should be rolling their own checks by accessing `kAllTextureFormatInfo` directly. They should all be gated on the data needed to give the correct answer, forcing you to get the larger list and then use `skipIf...` and/or `selectDeviceForXOrSkipTest`

I think I'd also suggest that most texture format filtering in general should be moved to some helpers that tests all share the same filtering where appropriate.

@shrekshao, @kainino0x

greggman:
[...]

PR #4192, opened 2025-02-20
Rename deprecated texture info functions and add new ones.

greggman:
This is a step to fixing the tests so they test all optional texture formats.

Note: I couldn't figure out how to share the code in validation_test.ts without bigger changes throught the code base. Maybe that can come in another PR.

The issue is, the rest of the code base expects that ValidationTest is a class which means you do this

    function foo(t: ValidationTest) { ... }
    foo(new ValidationTest());

Things like the Mixin pattern don't support this as the thing returned is at best a constructor but not a type. Hopefully this can be solved later.

Note that only `format_info.ts`, `gpu_test.ts`, `test_config.ts`, `validation_test.ts`
were edited directly. The rest are changed via VSCode's rename refactor functions.

Issue: #4181


greggman:
ran with dawn.node just the `webgpu:api,*` branch. Same pass/fail/skip numbers before and after this change.

kainino0x (approved):

kainino0x:
> Things like the Mixin pattern don't support this as the thing returned is at best a constructor but not a type. Hopefully this can be solved later.

Might be able to do `typeof new ValidationTest` / `typeof new (SomeMixin(GPUTest))` or something but I'm not sure if that works

PR #4217, opened 2025-02-26
Refactor src/webpgu/api/validation/encoding and a few extra

greggman:
Issue #4181
Issue #4178

Note: For copyTextureToTexture:texture_format_compatibility it was not at all clear to me why it was choosing textures with features. AFAICT that was testing almost nothing. Refactored it.


greggman:
this needs some work before review

PR #4218, opened 2025-02-26
Add coverage for overrides in nested arrays and overrides in array members of structures

petermcneeleychromium:
crbug.com/399151683


src/webgpu/shader/validation/expression/access/structure.spec.ts
```diff
@@ -114,3 +115,112 @@ g.test('result_type_runtime_array')
...
+
+interface OutOfBoundsCase {
+  code: string;
+  result: boolean;
+  pipeline?: boolean;
+  value?: number;
+}
+
+const kOutOfBoundsCases: Record<string, OutOfBoundsCase> = {
+  runtime_structure_array_override_oob_neg: {
```
jrprice, line 127: Nit: These tests should probably just go in the `array.spec.ts` file with others, since it's the array access that is OOB that we are trying to test.

jrprice (approved):

src/webgpu/shader/validation/expression/access/structure.spec.ts
petermcneeleychromium: Yes because from wgsl we cannot generate OOB in a structure with override unless it is an array. (It would only be an internal compiler error is this was to happpen.)

I might move it since it is cleaner.

petermcneeleychromium:
retested after minor refactor

greggman:
These test seem invalid? The test tries to call `y()` but the case snippet defines `fn foo()`

https://gpuweb.github.io/cts/standalone/?debug=1&runnow=1&q=webgpu:shader,validation,expression,access,array:early_eval_errors:*

petermcneeleychromium:
The tests were all green locally but yes we should change the reference from y to foo.

PR #4255, opened 2025-03-05
Make ShaderValidationTest auto skip if required features missing.

greggman:
With this change, if the WGSL passed to testing functions in ShaderValidationTest uses 'enable X' the test is skipped if the corresponding feature for 'X' does not exist on the device.

Tests that want to check that you can not use 'enable X' when feature 'X' does not exist should use different functions or these functions can be modified to take a 'do not auto skip' option.

Issue: #4178

### DIFF
--- a/src/webgpu/shader/validation/shader_validation_test.ts
+++ b/src/webgpu/shader/validation/shader_validation_test.ts
@@ -1,6 +1,45 @@
 import { keysOf } from '../../../common/util/data_tables.js';
 import { ErrorWithExtra } from '../../../common/util/util.js';
-import { AllFeaturesMaxLimitsGPUTest, UniqueFeaturesOrLimitsGPUTest } from '../../gpu_test.js';
+import {
+  AllFeaturesMaxLimitsGPUTest,
+  GPUTest,
+  UniqueFeaturesOrLimitsGPUTest,
+} from '../../gpu_test.js';
+
+const kEnables: Record<string, GPUFeatureName> = {
+  f16: 'shader-f16',
+  subgroups: 'subgroups' as GPUFeatureName,
+  clip_distances: 'clip-distances' as GPUFeatureName,
+};
+
+/**
+ * Note: These regular expressions are not meant to be perfect. This is not production code expecting
+ * to work with any WGSL passed by a user. It's only test code for working with WGSL written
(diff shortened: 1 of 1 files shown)


src/webgpu/shader/validation/shader_validation_test.ts
```diff
@@ -1,13 +1,39 @@
...
+
+const kEnables: Record<string, GPUFeatureName> = {
+  f16: 'shader-f16',
+  subgroups: 'subgroups' as GPUFeatureName,
+  clip_distances: 'clip-distances' as GPUFeatureName,
+};
+
+const kEnableREs = Object.entries(kEnables).map(([enableName, feature]) => {
+  return {
+    re: new RegExp(`\\benable\\s+${enableName}\\s*;`),
```
shrekshao, line 17: Don't think this works for cases like enable f16, subgroups;?
there seems to be these test cases in src/webgpu/shader/validation/parse/enable.spec.ts

I made the new regex and the following unitests
```
g.test('test')
  .fn(t => {
    const kEnableFeatureRegexCases = [
      { code: 'enable f16;', match: true },
      { code: 'enable f16, f16;', match: true },
      { code: 'enable f16, subgroups;', match: true },
      { code: 'enable subgroups, f16;', match: true },
      { code: 'enable clip_distances, f16;', match: true },
      { code: 'enable f16, clip_distances;', match: true },
      { code: 'enable test, another, f16;', match: true },
      { code: 'enable f16, test, another;', match: true },
      { code: 'enable test, f16, another;', match: true },
    
      { code: 'enable test, another;', match: false },
      { code: 'enable f16', match: false },
      { code: 'enable subgroups;', match: false },
    ];

    const enableName = 'f16';
    const re = new RegExp(`^enable\\s+(?:\\s*\\w+\\s*,)*\\s*${enableName}\\s*(?:,\\s*\\w+)*\\s*;$`);

    for (const {code, match} of kEnableFeatureRegexCases) {
      t.expect(re.test(code) === match, code);
    }
  });
```

greggman: Good catch

Uploaded a patch using your regexp

shrekshao (approved):
LGTM

PR #4260, opened 2025-03-06
Fix enable regular expression.

greggman:
This was missing a case in enable.spec.ts which used `/* ... */enable f16;`

I must have changed the expression after testing this path.


shrekshao (approved):
(Ah ... didn't think about the case for comments...)

PR #4262, opened 2025-03-06
Add comment case to WGSL enable test.

greggman:
The test is auto-skipped by expectCompileResult if shader-f16 doesn't exist. If shader-f16 does exist then this should fail because the enable appears in a comment.

I'm not sure there is a point to this test since `enable` is a language keyword and not a preprocessor directive like it was in GLSL but it can't hurt?

### DIFF
--- a/src/webgpu/shader/validation/parse/enable.spec.ts
+++ b/src/webgpu/shader/validation/parse/enable.spec.ts
@@ -82,6 +82,13 @@ enable f16;`,
     enable f16;`,
     pass: true,
   },
+  in_comment_f16: {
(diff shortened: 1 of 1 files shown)


shrekshao (approved):
(For the regex we add to detect enable feature, I think this one will match, and will be skipped on devices without shader-f16.)
Does this ^ looks correct?

But it's unharmful anyway. so I'm okay with it.

greggman:
yes, our regex will match. It's fine though. We really just want to test that a comment does not enable the WGSL feature if the device feature is available.

PR #4264, opened 2025-03-06
Return mismatched device holder to the mismatched device pool

greggman:
I didn't dig into why this was working but coming changes broke without this.

Issue: #4178


kainino0x (approved):
huh. I'm a bit surprised we didn't hit asserts when trying to put two devices back in the same pool.

PR #4270, opened 2025-03-06
Make selectDeviceXXX unreachable on AllFeaturesMaxLimitsGPUTest

greggman:
Ideally I think I'd like to move these to UniqueFeaturesAndLimitsGPUTest so they just wouldn't exist on AllFeaturesMaxLimitsGPUTest and you'd get a compiler error but that's big change so for now effectively assert if they are used.

Note: a few other PRs need to land to make sure this isn't being called before this lands.

Issue: #4178


shrekshao (approved):
LGTM (Land last)

PR #4274, opened 2025-03-07
Bump tsconfig.json lib to es2023

greggman:
Currently it's on ES2020 which does not define `WeakRef`. Given no other browser has shipped yet maybe it's ok to set it to ES2023?


kainino0x (approved):
This should be fine. If we introduce anything that doesn't work on some browser, it should hopefully be limited to just a few tests. (Any changes to the core infra we should be more careful to test on other browsers though, or at least check caniuse. WeakRef seems to be fine everywhere.

PR #4300, opened 2025-03-21
Tests that pipeline constant (override) errors only trigger on entry point usage.

petermcneeleychromium:
crbug.com/404819910


dneto0 (approved):
thanks!

Issue #2356, opened 2023-03-06
Check GPUAdapterInfo.isFallbackAdapter is present and a boolean


Kangz:
This is not checked by the CTS at the moment.

Also maybe we need to check that if a fallback adapter is requested, any returned adapter is indeed fallback (does the spec require this?)

beaufortfrancois:
Spec currently says:

> forceFallbackAdapter, of type boolean, defaulting to false
>
> When set to true indicates that only a fallback adapter may be returned. If the user agent does not support a fallback adapter, will cause requestAdapter() to resolve to null.

beaufortfrancois:
I believe we can close this issue now that I've merged https://github.com/gpuweb/cts/pull/4303.

If there's more to be done, feel free to re-open this issue

PR #4302, opened 2025-03-22
Roll types 0.1.52 -> 0.1.59

kainino0x:
This version adds `subgroupMinSize`/`subgroupMaxSize`, `textureBindingViewDimension`, `GPUAdapterInfo.isFallbackAdapter`, and tweaks some type definitions.

`idl_test.ts` needed to change because bitmask types changed from object types to interface types, which ran into this subtle difference between the two: https://github.com/microsoft/TypeScript/issues/15300

There shouldn't be any significant test behavior changes but since most of these edits are in common files I wasn't able to verify.

Issue: None

<hr>

**Requirements for PR author:**

- [n/a] All missing test coverage is tracked with "TODO" or `.unimplemented()`.
- [x] New helpers are `/** documented */` and new helper files are found in `helper_index.txt`.
- [ ] Test behaves as expected in a WebGPU implementation. (If not passing, explain above.)
- [ ] Test have be tested with compatibility mode validation enabled and behave as expected. (If not passing, explain above.)

**Requirements for [reviewer sign-off](https://github.com/gpuweb/cts/blob/main/docs/reviews.md):**

- [n/a] Tests are properly located in the test tree.
- [n/a] [Test descriptions](https://github.com/gpuweb/cts/blob/main/docs/intro/plans.md) allow a reader to "read only the test plans and evaluate coverage completeness", and accurately reflect the test code.
- [n/a] Tests provide complete coverage (including validation control cases). **Missing coverage MUST be covered by TODOs.**
- [ ] Helpers and types promote readability and maintainability.

When landing this PR, be sure to make any necessary issue status updates.

PR #4303, opened 2025-03-24
Add tests for GPUAdapterInfo.isFallbackAdapter

beaufortfrancois:
Issue: https://github.com/gpuweb/cts/issues/2356

image

<hr>


beaufortfrancois:
I think I need to wait for https://github.com/gpuweb/cts/pull/4302 to land to make CI bots happy and I'll rebase.

src/webgpu/api/operation/adapter/requestAdapter.spec.ts
```diff
@@ -114,10 +114,15 @@ g.test('requestAdapter')
       ...(forceFallbackAdapter !== undefined && { forceFallbackAdapter }),
     });
 
-    // failing to create an adapter when forceFallbackAdapter is true is ok.
-    if (forceFallbackAdapter && !adapter) {
-      t.skip('No adapter available');
-      return;
+    assert(typeof adapter?.info.isFallbackAdapter === 'boolean');
```
kainino0x: This assert will fail if `adapter` is null.

I adjusted the logic a bit and also switched to `t.expect`. PTAL

beaufortfrancois: Thanks for catching! LGTM

FWIW, I would have followed the following pattern.

```js
    if (!adapter) {
      // Failing to create an adapter is only OK when forceFallbackAdapter is true.
      t.expect(forceFallbackAdapter === true);

      // Mark the test as skipped (as long as nothing else failed before this point).
      t.skip('No fallback adapter available');
      return;
    }
    
    t.expect(adapter.info.isFallbackAdapter === Boolean(forceFallbackAdapter));
    await testAdapter(t, adapter);
```

kainino0x: That makes sense! Feel free to apply that and land

kainino0x (approved):

PR #4311, opened 2025-03-28
Remove lod_clamp.spec.ts

greggman:
I think this is effectively covered by the texture builtin tests? Though it depends on what you want to test

textureStore, textureLoad:storage, and textureSample:sampled_2d_coords check a 3 mip level texture both with and without a sub range and expect clamping.

We could add sub ranges to any other tests as the texture builtin helpers support sub ranges of layers and levels

* Do we need more (like should we be making maxTextureDimensions2Dx1) for every format to test all possible mip levels. How many sub ranges do we want to check?

* Should they be tested in the texture builtins? All builtins? All formats? Or just a select few? Or should they be tested somewhere else ... could still use the texture builtin utils for this. Just copy a texture builtin test and change the parameters to use less permutations.

* One thing to test would be multiple views of the same texture in the same shader.

One thing, IIRC, the texture builtin tests require the smallest mip level to be 2x2 or larger as they try to avoid edge cases and need some space to do so. (could be fixed if important)


kainino0x (dismissed):
The two most important things are:
- That lodMinClamp/lodMaxClamp are not ignored
- Corner cases where they might be applied incorrectly (like texture view with baseMipLevel > 0)

That seems like enough to make sure the browser implementation is correct. Beyond that, I think it's just testing the driver (unless there are any builtins that treat clamping differently from other builtins). You're in a better position to judge that than I am, but it sounds like a lot of work and a lot of test time and I'm OK with not going to that degree of detail.

greggman:
Doh! I think I thought `lodMaxClamp` and `lodMaxClamp` were synonyms for `baseMipLevel` and `mipLevelCount`. Now I see they are not, they're sampler parameters. They aren't tested anywhere in the CTS at the moment.

It would be easy to make the texture_utils.ts take them into account. I can then either add some cases to the texture builtins or put a test here so maybe I won't delete this right now.

Just to be clear, and glancing at the spec,

```
texture:
  mipLevelCount: 10,

view:
  baseMipLevel: 3,
  mipLevelCount: 4,

sampler:
  lodClampMin: 1,
  lodClampMax: 2,
```

Means 

1. The texture has 10 mip levels
2. From the POV of builtins, the texture only has 4 mip levels, physical levels 3,4,5,6. virtual levels 0,1,2,3
3. All calculations are done with the virtual levels (for example derivative mip selection will select between levels 0,1,2,3
4. Before actually sampling the selected mip level is clamped to virtual levels 1,2

Seems important to get some tests in for this. I'll put it on the fix it.

PR #4313, opened 2025-03-28
Add block-all-features flag

greggman:
This is useful to testing tests that switch on whether or not a feature is available. You can also use the webgpu-dev-extension for this but you can't use that extension in dawn-node so adding a flag to the CTS is useful. Plus the webgpu-dev-extension is not currently available on all browsers.


greggman:
note: the switch to `Object.defineProperty` was to make `webgpu:idl,javascript,*` tests pass.

kainino0x (approved):
LGTM!

We could have options to expose only bc or only etc2+astc, but it probably won't matter for any other tests, so no need.

Issue #4320, opened 2025-04-01
readbackFromWebGPUCanvas: Relax cross color space tolerance for Qualcomm srgb


jiangzhaoming:
Currently in readbackFromWebGPUCanvas WPT tests the ULP tolerance of different color space canvas is 2, making Qualcomm devices fail the `webgpu:web_platform,canvas,readbackFromWebGPUCanvas:offscreenCanvas,snapshot:alphaMode="premultiplied";colorSpace="srgb";*` and `webgpu:web_platform,canvas,readbackFromWebGPUCanvas:onscreenCanvas,snapshot:alphaMode="premultiplied";colorSpace="srgb";*`.

It turns out that Qualcomm devices (tested Adreno X1-85 GPU) have a maximum error of 3 ULP. Relaxing the tolerance of cross color space from 2 ULP to 3 ULP can make all these tests pass on Qualcomm devices.

PR #4321, opened 2025-04-01
Relax tolerance of readbackFromWebGPUCanvas WPT test to 3 ULP

jiangzhaoming:
This PR relax the cross color space error tolerance from 2 ULP to 3 ULP, making srgb testcases pass on Qualcomm devices.

Issue: #4320
<hr>

- [ ] ~New helpers are `/** documented */` and new helper files are found in `helper_index.txt`.~

### DIFF
--- a/src/webgpu/web_platform/canvas/readbackFromWebGPUCanvas.spec.ts
+++ b/src/webgpu/web_platform/canvas/readbackFromWebGPUCanvas.spec.ts
@@ -69,6 +69,9 @@ const expect = {
   ]),
 };
 
+// ULP tolerance fo cross color space readback
+const kMaxDiffULPsForNormFormatWithDifferentColorSpaceCanvas = 3;
+
 /**
  * Given 4 pixels in rgba8unorm format, puts them into an ImageData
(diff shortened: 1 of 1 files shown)


greggman (approved):

shrekshao (approved):

kainino0x (commented):
If that's 3 ULP of a normalized format (i.e. 3 / 255) that's quite a large difference. It's probably fine though, I'm guessing it will still catch if the conversion is completely wrong or doesn't happen at all. (Since you've looked more closely at the test, would you agree about that?)

jiangzhaoming:
Thanks for reviewing!

> If that's 3 ULP of a normalized format (i.e. 3 / 255) that's quite a large difference. It's probably fine though, I'm guessing it will still catch if the conversion is completely wrong or doesn't happen at all. (Since you've looked more closely at the test, would you agree about that?)

Yes I think so, the expect vs. actual diff seems to be 250 vs. 253. I think the conversion should definitely happened, as the diff between expectation of same and different color space seems much much larger than 3 ULP.

Issue #4325, opened 2025-04-02
cmdline runner not exiting


amaiorano:
This is most obvious when running `dawn.node` with `-isolate`, which runs the cmdline runner. After running the input tests, it just hangs until it eventually times out.

PR #4326, opened 2025-04-02
Fix cmdline runner not exiting

amaiorano:
At some point in the last few months, the cmdline runner started hanging after running the input tests. We're relying on node to shut down when there are no tasks left, so maybe we introduced something async that's not ending. In any case, it's safe to force exit after running the last test successfully.

Issue: #4325


kainino0x (approved):
Thanks!

PR #4327, opened 2025-04-03
Destroy devices when cleaning the device pool

kainino0x:
DevicePool won't hold more than 5 devices. When it drops the least-recently-used device, destroy it.

Clean up obsolete tricks to try do destroy GPUDevice before we had a destroy() method. Minor refactor where the destroy() call is made for devices that fail cleanup.

Spot tested locally, but haven't verified the whole CTS runs the same.
Manually verified it has the intended effect (see PR comments).

Issue: Maybe hopefully this will help with some of problems Chromium has been seeing with tests that create a lot of devices (though many of them won't be affected by this change)

<hr>


greggman (approved):

greggman:
FYI, I've run various parts of the cts with "count active devices" on in the [webgpu-dev-extension](https://chromewebstore.google.com/detail/webgpu-dev-extension/gkeaijopdhfempknmaggbjbedicopjgm). The CTS seems to work as is. I haven't seen extra devices. I probably only checked (1) the texture builtins (2) the limits tests, I don't remmeber if I tested the `requestDevice` tests

kainino0x:
Thanks, that's a useful feature!

I tried <https://gpuweb.github.io/cts/standalone/?q=webgpu:api,operation,adapter,requestDevice:*> and it got up to 9. But that test also makes its own devices, so it won't be affected by this.

The limits tests got as high as 10. This case for example (which is skipped on my machine) gets up to 9: <https://gpuweb.github.io/cts/standalone/?q=webgpu:api,validation,capability_checks,limits,maxBindGroupsPlusVertexBuffers:draw,at_over:limitTest=%22atDefault%22;testValueName=%22atLimit%22;encoderType=%22render%22>
This is a GPUTest but it seems to actually also be making its own devices. I think what's happening is just that subcases are running in parallel with one another?

Anyway, I don't know of any tests that stress this, so I just made one:

```
g.test('stress_device_pool')
  .params(u =>
    u //
      .combine('x', [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16])
  )
  .beforeAllSubcases(s => {
    s.selectDeviceOrSkipTestCase({ requiredLimits: { maxBufferSize: s.params.x } });
  })
  .fn(t => {
  });
```

This confirms it works. Without the patch, it reaches 17; with the patch, it reaches 6.

PR #4328, opened 2025-04-03
Fix assertReject

kainino0x:
assertReject() tried to throw an error if the promise didn't reject, but then accidentally caught that and treated it as thrown from the promise. And then it wasn't actually checking the expectedName of the exception. So, it did nothing, and would always pass. Whoops.

This was used in two places. Both are verified to be working.

- Popping an error scope after the test to make sure the test didn't leak any error scopes. This wasn't working, now it does.

  Hopefully there aren't any tests actually hitting this; I didn't run them all to find out.

- `api,operation,adapter,requestDevice:stale:awaitInitialError=true;*` as a workaround for the fact that shouldReject isn't awaitable. This still passes.

Issue: None, I just noticed this while testing #4327.

<hr>


greggman:
this somehow broke dawn.node. I'm guessing that's a bug in dawn.node but not sure.

kainino0x:
Since this isn't having trouble in chromium I think it's unlikely we have tests that are leaking error scopes. Maybe dawn.node is just doing the wrong thing when we pop an empty error scope stack?

kainino0x:
Actually that would cause all tests to fail. Hm.

PR #4329, opened 2025-04-03
Try to destroy all devices on page/worker teardown

kainino0x:
Rather than rely on garbage collection, use 'beforeunload' wherever possible to explicitly destroy GPUDevices.

For workers, 'beforeunload' sends a message to the worker telling it to unload, which destroys any devices on the worker and may (depending on the worker type) have the worker then close itself.

Mostly this is just to avoid relying on garbage collection, but service workers in particular can live beyond the lifetime of the page, so it's plausible they could beactually holding onto GPUDevices after the page is supposed to be done (and, say, the harness is trying to run another test in a new page load).

Issue: Might help with https://crbug.com/407088849 and others

<hr>


src/common/runtime/helper/test_worker.ts
```diff
@@ -10,22 +10,27 @@ import { WorkerTestRunRequest } from './utils_worker.js';
...
-  window.addEventListener('beforeunload', () => {
-    unregisterAllServiceWorkers();
-  });
+// NOTE: This code runs on startup for any runtime with worker support. Here, we use that chance
+// to delete any leaked service workers (in case they weren't cleaned up by the 'beforeonload'
+// handler below).
+unregisterAllServiceWorkers();
+
+/**
+ * Add a 'beforeunload' handler. (We use this here instead of registerShutdownTask so that we don't
```
beaufortfrancois, line 28: I wonder if we should use `pagehide` instead as `beforeunload` has issues: https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event
See https://developer.chrome.com/docs/web-platform/page-lifecycle-api#the_beforeunload_event as well

kainino0x: It sounds like the main reason not to use beforeunload is because it disables bfcache, but we don't care about that.

But maybe we should register both, if it trigger more reliably?

kainino0x:
PTAL!

greggman (approved):

beaufortfrancois (approved):

PR #4333, opened 2025-04-07
Use featureLevel instead of compatibilityMode

beaufortfrancois:


src/webgpu/util/device_pool.ts
```diff
@@ -261,11 +261,8 @@ function canonicalizeDescriptor(
   /** Canonicalized version of the requested limits: in canonical order, with only values which are
    * specified _and_ non-default. */
   const limitsCanonicalized: Record<string, number> = {};
-  // MAINTENANCE_TODO: Remove cast when @webgpu/types includes compatibilityMode
-  const adapterOptions = getDefaultRequestAdapterOptions() as unknown as {
-    compatibilityMode?: boolean;
-  };
-  const featureLevel = adapterOptions?.compatibilityMode ? 'compatibility' : 'core';
+  const adapterOptions = getDefaultRequestAdapterOptions();
+  const featureLevel = adapterOptions?.featureLevel === 'compatibility' ? 'compatibility' : 'core';
```
kainino0x: We should not assume that compatibility and core are the only two possible feature levels, or this is going to be hard to track down when that changes. I'll push a suggestion.

kainino0x (approved):

greggman (approved):

PR #4334, opened 2025-04-07
Save cts results as json

petermcneeleychromium:
Copying large json to the clipboard does not work well on non desktop machines
Save to a file works better and works better for filing bugs.

### DIFF
--- a/src/common/runtime/standalone.ts
+++ b/src/common/runtime/standalone.ts
@@ -701,6 +701,15 @@ void (async () => {
     void navigator.clipboard.writeText(logger.asJSON(2));
   });
 
+  document.getElementById('saveResultsJSON')!.addEventListener('click', () => {
+    const text = JSON.stringify(logger.asJSON(2));
+    const blob = new Blob([text], { type: 'text/plain' });
+    const link = document.createElement('a');
+    link.download = 'results-webgpu-cts.json';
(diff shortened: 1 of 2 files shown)


src/common/runtime/standalone.ts
```diff
@@ -701,6 +701,15 @@ void (async () => {
     void navigator.clipboard.writeText(logger.asJSON(2));
   });
 
+  document.getElementById('saveResultsJSON')!.addEventListener('click', () => {
+    const text = JSON.stringify(logger.asJSON(2));
```
greggman, line 705: logger.asJSON returns JSON (a string). 

```suggestion
    const text = logger.asJSON(2);
```

greggman (changes requested):

src/common/runtime/standalone.ts
petermcneeleychromium: Ah sounds good. Its nice to have a real js programmer here.

greggman (approved):

kainino0x (approved):
Honestly I didn't think anyone was using that functionality at all. I originally built the JSON export functionality thinking we would submit official conformance results like we did in WebGL, but we don't.

If we have a usecase for it we could even consider making it a bit more specialized for that, like producing human readable output instead of json if that would be better.

PR #4335, opened 2025-04-08
Unbounded precision intervals for addition and subtraction

petermcneeleychromium:
In most computations doubles (number in js) are equivalent to correctly rounded intervals. However in some cases (addition/subtraction) doubles do not provide enough numeric precision to fulfill WGSLs unbounded precision specification.

https://www.w3.org/TR/WGSL/#floating-point-accuracy 

These cases are whenever a number, x (e.g. 1.0) is added (or subtracted) from a much smaller number, y (e.g. smallest positive normal), such that the difference between them is smaller then ULP(x). When working in JS numbers the result will, incorrectly, simply be x, since it will get rounded. We must detect these cases where double precision does not represent infinitely accurate computations accurately then we must manually create the interval.

crbug.com/407987898


src/unittests/floating_point.spec.ts
```diff
@@ -6062,8 +6080,8 @@ const kDotIntervalCases = {
     // positive and negative large number cancel each other first, the result would be
```
zoddicus, line 6080: I think this comment is out of date

src/webgpu/util/floating_point.ts
```diff
@@ -708,6 +709,39 @@ export abstract class FPTraits {
     return new FPInterval(this.kind, n, n);
   }
 
+  /**
+   * WGSL specifies unbounded precision:
+   *   https://www.w3.org/TR/WGSL/#floating-point-accuracy
+   * In most computations doubles (number in js) can provide enough
+   * precision to give the correctly rounded interval.
```
zoddicus, line 716: nit: Maybe "In most computations doubles (number in js) are equivalent to for correctly rounded intervals."

src/webgpu/util/floating_point.ts
```diff
@@ -708,6 +709,39 @@ export abstract class FPTraits {
...
+   * However in some cases (addition/subtraction) doubles clearly do
```
zoddicus, line 717: nit: remove 'clearly'

src/webgpu/util/floating_point.ts
```diff
@@ -708,6 +709,39 @@ export abstract class FPTraits {
...
+   * not provide enough numeric precision.
+   * These cases are whenever a large number (eg 1.0) is added or
```
zoddicus, line 719: nit/suggested rewrite:

These cases are whenever a number, `x` (e.g. 1.0) is added (or subtracted) from a much smaller number, `y` (e.g. smallest positive normal), such that the difference between them is significantly smaller then `ULP(x)`. When working in JS `number`s the result will, incorrectly, simply be `x`, since it will get rounded. We must detect these cases where double precision does not represent infinitely accurate computations accurately then we must manually create the interval.

zoddicus (commented):
Overall looks good, just some nits/documentation comments. I would suggest adding David to the review.

Also could you add a bit more detail to the PR description. Just outlining what the issue is/what this PR is doing to fix them, so if in the future if someone is trying to investigate an issue with the floating point interval for example, the git log will tell them whether or not this sounds related.

dneto0:
Making Peter's example a little more concrete helps me understand this better:

* x = 1.0;   has biased exponent 127 in f32, and biased exponent 1023 in f64
* y = smallest normal in f32 = 1.17e-38,  has biased exponent 1 in f32, biased exponent 897 in f64
* Then x - y, when computed in f64 (number in JS), subtracts a number with binary exponent 897 from a number with binary exponent 1023, so their exponents are 126 apart.  
* So the significant digits of `y` are 126 positions to the right of the significant bits of `x`.  But f64 only has 52 explicit bits of mantissa, or 53 if you count the hidden 1 for normal numbers.   So in doing the subtraction, all the significant bits of `y` are lost.  The result in JS uses round-to-even, and we get `x - y = x`
* In the current state of the JS framework, the only allowed results are identically equal to `x`.
* But WGSL implementations are allowed to round down or up, so even in f32 `x-y` can be `x - 1 ULP(x)`.  This change accommodates that possibility.

src/webgpu/util/floating_point.ts
```diff
@@ -708,6 +709,41 @@ export abstract class FPTraits {
...
+   *   https://www.w3.org/TR/WGSL/#floating-point-accuracy
+   * In most computations doubles (number in js) are equivalent
+   * to correctly rounded intervals. However in some cases
+   * (addition/subtraction) doubles do not provide enough numeric precision.
+   *
+   * These cases are whenever a number, x (e.g. 1.0) is added (or subtracted)
+   * from a much smaller number, y (e.g. smallest positive normal), such that
+   * the difference between them is smaller then ULP(x). When working in JS
+   * numbers the result will, incorrectly, simply be x, since it will get
+   * rounded. We must detect these cases where double precision does not
```
dneto0, line 723: I think this also has to say that WGSL allows rounding either up or down, so it may not compute the same rounded result as in JS.

JS uses round-ties-to-even.  https://tc39.es/ecma262/#sec-ecmascript-language-types-number-type

src/webgpu/util/floating_point.ts
```diff
@@ -708,6 +709,41 @@ export abstract class FPTraits {
...
 
+  /**
+   * WGSL specifies unbounded precision:
+   *   https://www.w3.org/TR/WGSL/#floating-point-accuracy
+   * In most computations doubles (number in js) are equivalent
+   * to correctly rounded intervals. However in some cases
+   * (addition/subtraction) doubles do not provide enough numeric precision.
+   *
+   * These cases are whenever a number, x (e.g. 1.0) is added (or subtracted)
+   * from a much smaller number, y (e.g. smallest positive normal), such that
```
dneto0, line 720: I would reorder this to say that y is added or subtracted from x.
Y has the small magnitude, and X has the relatively large magnitude.

src/webgpu/util/floating_point.ts
```diff
@@ -4317,7 +4356,10 @@ export abstract class FPTraits {
 
   private readonly SubtractionIntervalOp: ScalarPairToIntervalOp = {
     impl: (x: number, y: number): FPInterval => {
-      return this.correctlyRoundedInterval(x - y);
+      const difference: number = x - y;
```
dneto0, line 4359: This is the addition case    but using -y for the second argument.
It was easier to understand when I read it that way.  May be worth a comment.

src/webgpu/util/floating_point.ts
```diff
@@ -708,6 +709,41 @@ export abstract class FPTraits {
...
+   * from a much smaller number, y (e.g. smallest positive normal), such that
+   * the difference between them is smaller then ULP(x). When working in JS
+   * numbers the result will, incorrectly, simply be x, since it will get
+   * rounded. We must detect these cases where double precision does not
+   * represent infinitely accurate computations accurately then we must
+   * manually create the interval.
+   */
+  /** @returns an interval containing the correctly rounded val with respect to unbounded precision */
+  public correctlyRoundedIntervalWithUnboundedPrecision(
+    val: number,
```
dneto0, line 729: I recommend documenting the parameters:

@param sum the result of adding large_val + small_val, using Number
@param large_val: the summand with larger magnitude
@param small_val: the summand with smaller magnitude

src/unittests/floating_point.spec.ts
```diff
@@ -4048,6 +4048,14 @@ const kAdditionInterval64BitsNormalCases = {
     { input: [0.1, -0.1], expected: [reinterpretU32AsF32(0x3dcccccc)+reinterpretU32AsF32(0xbdcccccd), reinterpretU32AsF32(0x3dcccccd)+reinterpretU32AsF32(0xbdcccccc)] },  // ~0.0
     // -0.1+0.1 expect f32 interval [0xBDCCCCCD+0x3DCCCCCC, 0xBDCCCCCC+0x3DCCCCCD]
     { input: [-0.1, 0.1], expected: [reinterpretU32AsF32(0xbdcccccd)+reinterpretU32AsF32(0x3dcccccc), reinterpretU32AsF32(0xbdcccccc)+reinterpretU32AsF32(0x3dcccccd)] },  // ~0.0
+    { input: [1, kValue.f32.positive.min], expected:  [ 1.0, reinterpretU32AsF32(0x3f800001)]  },
```
dneto0, line 4051: Please add some cases where the small magnitude value is the first summand.

src/unittests/floating_point.spec.ts
```diff
@@ -5017,6 +5025,14 @@ const kSubtractionInterval64BitsNormalCases = {
     { input: [0.1, -0.1], expected: [reinterpretU32AsF32(0x3dcccccc)-reinterpretU32AsF32(0xbdcccccc), reinterpretU32AsF32(0x3dcccccd)-reinterpretU32AsF32(0xbdcccccd)] },
     // Expect f32 interval [0xBDCCCCCD-0x3DCCCCCD, 0xBDCCCCCC-0x3DCCCCCC]
     { input: [-0.1, 0.1], expected: [reinterpretU32AsF32(0xbdcccccd)-reinterpretU32AsF32(0x3dcccccd), reinterpretU32AsF32(0xbdcccccc)-reinterpretU32AsF32(0x3dcccccc)] },
+    { input: [1, kValue.f32.positive.min], expected: [reinterpretU32AsF32(0x3f7fffff), 1.0] },
```
dneto0, line 5028: Please add cases where the small-magnitude summand is the first operand.

dneto0 (changes requested):

src/webgpu/util/floating_point.ts
```diff
@@ -708,6 +709,41 @@ export abstract class FPTraits {
...
+   * These cases are whenever a number, x (e.g. 1.0) is added (or subtracted)
+   * from a much smaller number, y (e.g. smallest positive normal), such that
+   * the difference between them is smaller then ULP(x). When working in JS
+   * numbers the result will, incorrectly, simply be x, since it will get
+   * rounded. We must detect these cases where double precision does not
+   * represent infinitely accurate computations accurately then we must
+   * manually create the interval.
+   */
+  /** @returns an interval containing the correctly rounded val with respect to unbounded precision */
+  public correctlyRoundedIntervalWithUnboundedPrecision(
```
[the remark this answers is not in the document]
dneto0: Recommend `correctlyRoundedSumIntervalWithUnboundedPrecision`  or use Add instead of Sum

petermcneeleychromium:
I might be able to simply have subtraction call addition with the flipped sign.
Since in all floating point (f64,f32,f16) this is a symmetric operation this should be fine.

Shall I do this cleanup?

dneto0 (approved):
Thanks!

PR #4336, opened 2025-04-09
Add tests for consumed adapter in requestDevice

beaufortfrancois:
According to https://gpuweb.github.io/gpuweb/#ref-for-dom-adapter-state-consumed%E2%91%A1, the adapter must not be "consumed" when requesting a device, otherwise it should return a rejected promise with OperationError. This was not tested and not even working in Chromium, Safari, and Firefox when testing locally. 

This makes me wonder if the spec should be updated or if we need to update all browsers implementations.
image

<hr>


src/webgpu/api/operation/adapter/requestDevice.spec.ts
```diff
@@ -157,6 +157,12 @@ g.test('stale')
           );
         }
         break;
+      case 'ConsumedAdapter':
+        // Cause an operation error by requesting a device on a consumed adapter.
+        if (!awaitInitialError) {
```
kainino0x, line 162: `awaitInitialError` was just intended to control whether we `await` or not on the first requestDevice call before continuing the test. Why is it different in this case?

beaufortfrancois: My understanding is that it doesn't make sense to wait for requestDevice to fail as the adapter is not consumed at that point.

kainino0x: The test is set up so that if `initialError !== undefined`, there's always two requestDevice calls, one in this switch, and one later. Now this parameter is controlling whether there's one or two.

To make this make more sense, the parameters should be renamed something like:

```
    u
      .combine('request1', ['None', 'Valid', 'TypeError', 'OperationError'])
      .expand('awaitRequest1', p => p.firstRequest === 'None' ? [false] : [true, false])
      .combine('awaitRequest2', [true, false])
```

beaufortfrancois: I've renamed parameters and adjusted code. PTAL

src/webgpu/api/operation/adapter/requestDevice.spec.ts
```diff
@@ -104,26 +104,27 @@ g.test('stale')
...
 
-    switch (initialError) {
-      case undefined:
+    switch (request1) {
+      case 'None':
+        // Cause an operation error by requesting a device on a consumed adapter.
+        t.shouldReject('OperationError', t.requestDeviceTracked(adapter));
+        break;
+      case 'Valid':
         break;
```
kainino0x, line 124: There haven't been any requestDevice calls yet, why would it be consumed?

What I was thinking here was:
```suggestion
      case 'None':
        break;
      case 'Valid':
        if (awaitRequest1) {
          await t.requestDeviceTracked(adapter);
        } else {
          t.requestDeviceTracked(adapter);
        }
        break;
```

And then later on, request2 is where you check for an OperationError for a consumed adapter.

kainino0x: No request2 parameter, but there is a second requestDevice call. That's the one that we're testing. The first requestDevice is just setup for the second one.

beaufortfrancois: Gotcha! I've updated code.

src/webgpu/api/operation/adapter/requestDevice.spec.ts
```diff
@@ -161,16 +166,24 @@ g.test('stale')
...
+        t.shouldResolve(
+          (async () => {
+            const device = await promise;
+            device.destroy();
+          })()
+        );
+      }
     }
 
     const kTimeoutMS = 1000;
```
beaufortfrancois, line 189: My understanding is that the the following promise `const lostDevice = await t.requestDeviceTracked(adapter);` will always fail now because we already have called `const promise = t.requestDeviceTracked(adapter);` earlier and  only one device can be requested from an adapter according to the spec. Shall we remove this bit?

kainino0x: Wait, so sorry, I'm very confused. I didn't realize we already had a check down here that tests that the adapter is consumed. If we're already testing this, what test coverage are you trying to add with this CL?

beaufortfrancois: According to https://gpuweb.github.io/gpuweb/#ref-for-dom-adapter-state-consumed%E2%91%A1, the adapter must not be "consumed" when requesting a device, otherwise it should return a rejected promise with OperationError, not resolving with a lost device. The following assumes we'll always get a lost device. This is not true according to the spec.

FYI I'm working on https://dawn-review.googlesource.com/c/dawn/+/237274 to fix this behavior in Chromium.

kainino0x: Ah right, that's because we changed the spec. But isn't this bit at the end of the test trying to specifically test the "consume"/"expire" behavior? Why not just update it so it expects a rejection instead of a device loss? (Won't this be necessary anyway, to make the test passable?)

beaufortfrancois: I've changed it to reject instead of device loss as advised in https://github.com/gpuweb/cts/pull/4336#discussion_r2053195855.
It works locally with https://dawn-review.googlesource.com/c/dawn/+/237274

image

kainino0x (approved):
LGTM! Sorry for the roundabout path to get here.

PR #4340, opened 2025-04-10
Fix shader,execution,expression,call,builtin,subgroupShuffle:compute,split

haoxli:
This case should be designed to use local_invocation_index for getting input values.
On most Intel/Nivida GPUs, the subgroup_invocation_id is incremented by rows, and same to local_invocation_index in a workgroup when the workgroup size is same to subgroup size, even if the test uses subgroup id, it also can pass. But on Intel Gen12 integrated GPUs, the subgroup_invocation_id is incremented by columns, the test cannot run as expected.

Issue: [chromium:401558014](https://issues.chromium.org/issues/401558014)

<hr>


haoxli:
At first, when I saw the shader code, I thought it was using `subgroup_invocation_id` to read the inputs. After I changed the `checkCompute`  function and `compute,all_active` test, I just realized that I should use `local_invocation_index` here. And the `checkCompute` also depends on the order of `local_invocation_index`. I'm not sure if can we optimize the expected values to depend only on the inputs and `subgroup_invocation_id`? We have recorded the `subgroup_invocation_id` corresponding to `local_invocation_index`.

haoxli:
@alan-baker PTAL, thanks!

alan-baker (approved):

PR #4344, opened 2025-04-10
Fix block/enforce/compat option interaction

greggman:
If you turned on either blockAllFeatures or enforceDefaultLimits and then you also turned on compatibility mode you'd get an error that `impl.requestAdapter` is read-only.

This is the fix.

### DIFF
--- a/src/common/util/navigator_gpu.ts
+++ b/src/common/util/navigator_gpu.ts
@@ -197,22 +197,26 @@ export function getGPU(recorder: TestCaseRecorder | null): GPU {
 
   if (defaultRequestAdapterOptions) {
     // eslint-disable-next-line @typescript-eslint/unbound-method
-    const oldFn = impl.requestAdapter;
-    impl.requestAdapter = function (
-      options?: GPURequestAdapterOptions
-    ): Promise<GPUAdapter | null> {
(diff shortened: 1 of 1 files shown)


src/common/util/navigator_gpu.ts
```diff
@@ -197,22 +197,26 @@ export function getGPU(recorder: TestCaseRecorder | null): GPU {
...
+        const adapter = await origRequestAdapterFn.call(this, {
+          ...defaultRequestAdapterOptions,
+          ...options,
+        });
+        if (recorder && adapter) {
           if (adapter) {
             const adapterInfo = adapter.info;
             const infoString = `Adapter: ${adapterInfo.vendor} / ${adapterInfo.architecture} / ${adapterInfo.device}`;
             recorder.debug(new ErrorWithExtra(infoString, () => ({ adapterInfo })));
           }
```
kainino0x, line 215: Redundant if()?

```suggestion
          const adapterInfo = adapter.info;
          const infoString = `Adapter: ${adapterInfo.vendor} / ${adapterInfo.architecture} / ${adapterInfo.device}`;
          recorder.debug(new ErrorWithExtra(infoString, () => ({ adapterInfo })));
```

kainino0x (approved):

PR #4350, opened 2025-04-11
Refactor validation tests

greggman:
Sorry this is so large, it's mostly search and replace tho.

This pattern of making test classes and putting methods on it
is problematic because it leads to diamond pattern issues.
In this case, this became a problem because some tests want
AllFeaturesMaxLimits and others want UniqueFeaturesOrLimits.

Moving all these function to just standalone methods removes
the need for class ValidationTest. It's a big rename though

AllFeaturesMaxLimistValidationTest -> AllFeaturesMaxLimitsGPUTest
UniqueFeaturesOrLimitsValidationTest -> UniqueFeaturesOrLimitsGPUTest

Then, any file that wants to use these helpers can either import
them individually or use `import * as vtu from './validation_test_utils.js'.

The calls go from

```js
t.someUtil(args);
```

to

```js
vts.someUtil(t, args);
```

Note that there were several helper methods only called by 1 or 2
tests at most which suggests maybe they shouldn't here but just
implemented in those individual test files but we can leave that
for another time.


src/webgpu/shader/validation/shader_io/workgroup_size.spec.ts
```diff
@@ -116,7 +116,7 @@ const kWorkgroupSizeTests = {
   },
 
   trailing_comma_x: {
-    src: `@workgroup_size(8, )`,
+    src: `@workgroup_size(8)`,
```
shrekshao, line 119:
```suggestion
    src: `@workgroup_size(8, )`,
```

shrekshao (approved):

PR #4351, opened 2025-04-14
Update fp primer unbounded precision

petermcneeleychromium:
Minor additional documentation for when double precision is not enough to represent the results of a computation.
crbug.com/407987898

### DIFF
--- a/docs/fp_primer.md
+++ b/docs/fp_primer.md
@@ -261,6 +261,8 @@ since its native number format is higher precision (double-precision/f64), so
 all f64, f32, and f16 values can be represented in it. Where this breaks down
 will be discussed in the section on compile time vs runtime execution.
 
+Here?
+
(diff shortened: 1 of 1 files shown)


docs/fp_primer.md
```diff
@@ -258,8 +258,16 @@ digits of precision.
 
 For the CTS it is often sufficient to calculate the true value using TypeScript,
 since its native number format is higher precision (double-precision/f64), so
-all f64, f32, and f16 values can be represented in it. Where this breaks down
-will be discussed in the section on compile time vs runtime execution.
+all f64, f32, and f16 values can be represented in it. Unfortunately, the results of some operations cannot be exactly
```
zoddicus, line 261: nit: Can you line wrap these to match the rest of the body text? (I think it is 80 column width)

zoddicus (approved):
lgtm other then a nit about line wrapping

PR #4358, opened 2025-04-16
Expression in workgroup size

petermcneeleychromium:
crbug.com/408233377

### DIFF
--- a/src/webgpu/shader/validation/shader_io/workgroup_size.spec.ts
+++ b/src/webgpu/shader/validation/shader_io/workgroup_size.spec.ts
@@ -131,13 +131,13 @@ const kWorkgroupSizeTests = {
     src: `@id(42) override block_width = 12u;
 @workgroup_size(block_width)`,
     pass: true,
-    pipeline: false,
+    pipeline: true,
   },
   override_no_default: {
     src: `override block_width: i32;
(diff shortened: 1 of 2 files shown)


jrprice (commented):
I think this is OK, although I was kinda expecting to see this test coverage added to our execution tests instead so that we make sure the actual value is correct, e.g. a new test in [this file](https://github.com/gpuweb/cts/blob/main/src/webgpu/shader/execution/shader_io/workgroup_size.spec.ts).

If we were to do that I think we don't really need to make the changes to the `expectPipelineResult` helper here, which is maybe nicer?

petermcneeleychromium:
I sorta like this this is a validation failure. And it belongs as a validation failure. There are many other pipeline validation failures

jrprice:
The case we care about (from the Tint bug) is not a failure case though, so we really should have execution test coverage anyway. As in, we should have a test that makes sure a complex override expression in the workgroup size attribute actually computes the correct workgroup size.

I'm fine with keeping the coverage you've already added too though.

petermcneeleychromium:
Ok sounds good. I will add an additional execution test as well to make sure the expressions are correctly evaluated

jrprice (approved):

PR #4376, opened 2025-04-28
Fix usages of textureDimensionAndFormatCompatible

greggman:
Sorry this is so big. Most of it is relatively search and replace

Fix usages of `textureDimensionAndFormatCompatible`.

All of these usages were wrong as they need the device to
know if a dimension and format are compatible since
extensions enable 3d dimension with compressed textures.

So, rename `textureDimensionAndFormatCompatible` to
`textureFormatAndDimensionAndPossiblyCompatible` to
let a test filter out formats that can't possibly
be used with certain dimensions. Then, for each test,
call `t.skipIfTextureFormatAndDimensionNotCompatible`
which checks the extensions.

Further, a few tests needed fixing.

copyTextureToTexture validation tests needed to handle
3d compress texture size limits correctly.

createTexture needed to not skip 3d compressed textures

layout_related needed to pass dimension to createAlignedTexture
and also, several tests had dimension as a test parameter but
didn't seem to be using it anywhere so added it into at least
texture creation on those tests.

expectTextureToMatchByRendering needed to handle 3d textures

texture_zero tests were mostly indirectly affected by
changes to texture_zero_init_test.ts. Nothing interesting
changed in this PR but the params there are extremely
convoluted and I didn't bother to decode exactly what
is going on in detail. I did glance over them and I think
they aren't skipping compressed textures.

### DIFF
--- a/src/webgpu/api/operation/command_buffer/copyTextureToTexture.spec.ts
+++ b/src/webgpu/api/operation/command_buffer/copyTextureToTexture.spec.ts
@@ -21,7 +21,7 @@ import {
   kDepthStencilFormats,
   kRegularTextureFormats,
   RegularTextureFormat,
-  textureDimensionAndFormatCompatible,
+  textureFormatAndDimensionPossiblyCompatible,
   textureFormatsAreViewCompatible,
 } from '../../../format_info.js';
 import { AllFeaturesMaxLimitsGPUTest } from '../../../gpu_test.js';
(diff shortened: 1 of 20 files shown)


shrekshao (approved):
LGTM, thanks for the detailed comments on each test modification

beaufortfrancois:
Thanks @greggman!

PR #4384, opened 2025-05-05
Pipeline error for out of bounds array with dynamic and offset

petermcneeleychromium:
Const indices are validated against array types for bounds checking even if there is a dynamic index dependency on the array expression.

For example 
```
const q = array(array(14,2,3),array(1,7,3));

const o = 17;
fn compute() -> f32 {
  var a = 20;
  return f32(q[a][o]);
}
```
Will generate an out of bounds error (compile)
whereas currently we expect (but do not get) a pipeline error for 
```

const q = array(array(14,2,3),array(1,7,3));

const o = 17;
fn compute() -> f32 {
  var a = 20;
  return f32(q[a][o]);
}
```

crbug.com/415298444

### DIFF
--- a/src/webgpu/shader/validation/expression/access/array.spec.ts
+++ b/src/webgpu/shader/validation/expression/access/array.spec.ts
@@ -378,7 +378,7 @@ const kOutOfBoundsCases: Record<string, OutOfBoundsCase> = {
     pipeline: true,
     value: 1,
   },
-  override_array_dynamic_tested_oob_pos: {
+  override_array_dynamic_type_checked_oob_pos: {
     code: `@group(0) @binding(0) var<storage> v : array<array<array<u32, 3>, 4>, 5>;
     override x : i32;
     override w = 0u;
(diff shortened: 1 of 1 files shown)


src/webgpu/shader/validation/expression/access/array.spec.ts
```diff
@@ -378,6 +378,45 @@ const kOutOfBoundsCases: Record<string, OutOfBoundsCase> = {
     pipeline: true,
     value: 1,
   },
+  override_array_dynamic_tested_oob_pos: {
```
jrprice, line 381: What does "tested" signify in these case names?

jrprice (approved):

src/webgpu/shader/validation/expression/access/array.spec.ts
petermcneeleychromium: tested is pretty vague. It should be type checked. As it is the type itself that has bounds.

PR #4388, opened 2025-05-12
Add some JavaScript inheritance tests.

greggman:
This is mostly for non-browser implementations to pass. dawn.node passes none of these. Working on getting it pass all of them. This is related to it correctly handling events.

### DIFF
--- a/src/webgpu/idl/javascript.spec.ts
+++ b/src/webgpu/idl/javascript.spec.ts
@@ -538,3 +538,27 @@ g.test('sameObject')
       assert(origValue2.foo === undefined);
     }
   });
+
+const kClassInheritanceTests: { [key: string]: () => boolean } = {
+  GPUDevice: () => GPUDevice.prototype instanceof EventTarget,
+  GPUPipelineError: () => GPUPipelineError.prototype instanceof DOMException,
+  GPUValidationError: () => GPUValidationError.prototype instanceof GPUError,
+  GPUOutOfMemoryError: () => GPUOutOfMemoryError.prototype instanceof GPUError,
+  GPUInternalError: () => GPUInternalError.prototype instanceof GPUError,
+  GPUUncapturedErrorEvent: () => GPUUncapturedErrorEvent.prototype instanceof Event,
+} as const;
+
+g.test('inheritance')
+  .desc(
+    `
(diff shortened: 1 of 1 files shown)


src/webgpu/idl/javascript.spec.ts
```diff
@@ -538,3 +538,48 @@ g.test('sameObject')
       assert(origValue2.foo === undefined);
     }
   });
+
+const kClassInheritanceTests: { [key: string]: (device: GPUDevice) => [boolean, string] } = {
+  GPUDevice(device: GPUDevice) {
+    return [device instanceof EventTarget, 'GPDevice extends EventTarget'];
```
kainino0x, line 544: optional nit: Could be `GPUDevice.prototype instanceof EventTarget`, then you don't need to pass a device to test this.

Actually if you did this for all of them then you wouldn't even need custom constructor code for each one, since you wouldn't need to construct them.

I'm not sure if this would result in less complete test coverage though.

kainino0x (approved):

src/webgpu/idl/javascript.spec.ts
greggman: I'm not sure we want to require that the next parent be specific. We just want to require that a specific parent is somewhere higher in the prototype chain. I could walk the chain but that's what `instanceof` does.

kainino0x: Wouldn't this still walk the chain? It still uses `instanceof`. It just uses the prototype object instead of real one so you don't need to have a real one.

```
class A {}
class B extends A {}
class C extends B {}
> C.prototype instanceof A
true
```

kainino0x: Also I think it actually would be fine to require the parent be specific. That's what the spec says, and inserting an extra thing in the chain can break code.

greggman: I switched it to `Derived.prototype instead of Base` as you suggested. I didn't require direct parent inheritance though we can add that if important.

This is appearing to be problematic for node plugins. Also, while here, there is no `DOMException` in node which means we have to make one but that may or may not match anything the user creates (like if they use the jsdom library)

PR #4389, opened 2025-05-15
Add some tests related to GPUDevice extending EventTarget

greggman:
* Test you can pass Event and CustomEvent through GPUDevice.
* Test if you override a method on EventTarget it's called when you call the corresponding method on GPUDevice.
* Test if you can pass GPUUncapturedErrorEvent through an EventTarget that is not GPUDevice

These pass on Chrome, Firefox, and Safari. dawn.node fails these tests though I have a patch that makes it work.

Note: I made Fixture.expect take a function. If function was passed and no msg was provided then it uses the source of the function as the string. This makes it nice for small tests to see what passed/failed. Without this all you get is OK/FAIL and you need to provide a message. 

Example:

Screenshot 2025-05-15 at 16 07 56

wdyt?


kainino0x (approved):
> Note: I made Fixture.expect take a function. If function was passed and no msg was provided then it uses the source of the function as the string. This makes it nice for small tests to see what passed/failed. Without this all you get is OK/FAIL and you need to provide a message.

doh, that's a good idea! when I was originally bringing up the CTS I really wanted something that would show nicer results for asserts (like https://github.com/power-assert-js/power-assert but that was too hard to integrate). Never thought of this for some reason.

lgtm!

PR #4392, opened 2025-05-21
Expand acceptable range of tanh

petermcneeleychromium:
Assuming this pr goes through
https://github.com/gpuweb/gpuweb/pull/5199

crbug.com/390221422


zoddicus (commented):
Looks fine to me, though you might need to regenerate the test cache, since the checks are failing.

I am not going to approve right now, since this is dependent on an upstream spec change, but adding @dneto0 incase that lands while I am away, so he can approve.

zoddicus:
The command to re-generate the cache is `npx grunt run:generate-cache` I believe, it has been a while since I have done it.

dneto0 (approved):

petermcneeleychromium:
We have approval. I will  resolve conflicts and reupload
https://github.com/gpuweb/gpuweb/pull/5199

src/resources/cache/hashes.json
ErichDonGubler, line 1: question: It's rather unfortunate that so many hashes got changed here for such a small change to `TanhIntervalOp`. I wonder if it's desirable to break up `util/floating_point.ts` so that fewer hashes get invalidated in the future? I'm not sure what the consequences of churn in `hashes.json` are, though.

kainino0x: The hashes hash all of the files in the dependency tree that produces a .bin file. They aren't used at test runtime, only at build time to verify that the .bin files are up to date. So churn in .bin matters a bit (it changes the test), but in the hashes it doesn't.

PR #4395, opened 2025-05-27
Follow common JavaScript style

greggman:
This is probably my OCD but, JavaScript style is almost universally, constructors are Capitalized and everything else is not.

Personally I wish we'd turn on the "new-cap" rule for eslint but unfortunately there are some slightly more difficult usages in the shader tests.

So, instead, thought I'd just try to refactor most code using C++ style to use JavaScript style so there are less examples to copy from.

Feel free to reject this PR if you think this is too intrusive.


kainino0x (approved):
I think this is good. Thank you! Lint would be great if you manage to get it passing.

PR #4396, opened 2025-05-27
Fix out-of-order subcase logs

kainino0x:
- First commit fixes a bug where we were capturing a variable too late which prevented the subcase logs from actually being re-sorted the way they were supposed to be
- Second commit simplifies the code a ~bunch~ bit and moves the prefix to the beginning of the log line

image

It still works for DOMException without needing all the fallback cases:

image

Issue: None filed

<hr>

- [n/a] Test have be tested with compatibility mode validation enabled and behave as expected. (If not passing, explain above.)


src/common/internal/logging/test_case_recorder.ts
```diff
@@ -181,4 +181,26 @@ export class TestCaseRecorder {
...
+  makeDeferredSubRecorder(prefix: string, deferUntilPromise: Promise<void>) {
+    return new Proxy(this, {
+      get: (target, k) => {
+        /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
+        if (k !== 'logImpl') return (target as any)[k];
+
+        globalTestConfig.testHeartbeatCallback();
+        return function (level: LogSeverity, name: string, baseException: unknown) {
+          void deferUntilPromise.then(() => {
+            TestCaseRecorder.prototype.logImpl.call(target, level, prefix + name, baseException);
```
greggman, line 200: should this be `target.logImpl.call`?

greggman (approved):

src/common/internal/logging/test_case_recorder.ts
kainino0x: yes, suppose so. I thought that would cause infinite recursion, but of course the proxy object is a different object than `target`.

PR #4397, opened 2025-05-28
Test onuncapturederror order vs addEventListener

greggman:
The Web spec says setting onuncaptrederror adds a listener via addEventListener that calls the callback. Changing onuncapturederror changes the callback to the existing listener. Setting onuncapturederror to null removes the listener.

https://html.spec.whatwg.org/multipage/webappapis.html#activate-an-event-handler

I ran into this testing how onuncapturederror is supposed to work while trying to add support to dawn.node

Maybe the WebGPU spec should make it clear it follow the same rules if that's not already true.

### DIFF
--- a/src/webgpu/api/operation/uncapturederror.spec.ts
+++ b/src/webgpu/api/operation/uncapturederror.spec.ts
@@ -45,7 +45,7 @@ g.test('onuncapturederror_order_wrt_addEventListener')
     `
 Test that onuncapturederror and addEventListener work in the correct order.
 
-The spec says setting onuncaptrederror adds a listener via addEventListener that
+The spec says setting onuncapturederror adds a listener via addEventListener that
 calls the callback. Changing onuncapturederror changes the callback to the existing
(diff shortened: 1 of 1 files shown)


src/webgpu/api/operation/uncapturederror.spec.ts
```diff
@@ -38,3 +39,90 @@ g.test('uncapturederror_from_non_originating_thread')
 (since deserialized ones don't have EventTarget/onuncapturederror).`
   )
   .unimplemented();
+
+g.test('onuncapturederror_order_wrt_addEventListener')
+  .desc(
+    `
+Test that onuncapturederror and addEventListener work in the correct order.
+
+The spec says setting onuncaptrederror adds a listener via addEventListener that
```
kainino0x, line 48:
```suggestion
The spec says setting onuncapturederror adds a listener via addEventListener that
```

kainino0x (approved):
literally never understood how this worked before now. TIL!

PR #4400, opened 2025-06-03
Roll @webgpu/types

greggman:
This pulls in new formats for feature 'texture-formats-tier1' which required updates to format_info.ts, capability_info.ts and several tests.

Note: Since no implementation actually supports 'texture-formats-tier1' yet it's very possible some tests will need fixes when the feature is implemented.


greggman:
Ran locally in dawn.node and fixed what I saw. Running here https://dawn-review.googlesource.com/c/dawn/+/245034 to test on browser.

greggman:
Also, it's not clear from the spec I think, either WGSL is supposed to allow these formats by default.  For example, WGSL allows `texture_storage_2d<bgra8norm, ...>` even if `bgra8norm-storage` is not available.

* Should WGSL allow `texture_storage_2d<rgba16unorm>` (a format that does not exist without texture-formats-tier1)
* Should WGSL allow `texture_storage_2d<r8unorm>` (a format that does exist but doesn't support storage without `texture-formats-tier1`

As it is, a few tests here and there currently fail with WGSL compilation errors for these formats. I feel like WGSL should allow these, same as it allows `bgra8unorm`

kainino0x:
Agree it should work the same as `bgra8unorm`. AFAICT neither API nor WGSL spec calls out `bgra8unorm` as special, so seems like these new ones should behave the same as it.

src/webgpu/format_info.ts
```diff
@@ -288,6 +288,34 @@ const kRegularTextureFormatInfo = formatTableWithDefaults({
...
+      color: {
+        type: 'float',
+        copySrc: true,
+        copyDst: true,
+        storage: true,
+        readWriteStorage: false,
+        bytes: 2,
+      },
+      colorRender: { blend: true, resolve: true, byteCost: 2, alignment: 2 },
+      multisample: false,
```
kainino0x: spec says this should be true

src/webgpu/format_info.ts
```diff
@@ -328,6 +356,34 @@ const kRegularTextureFormatInfo = formatTableWithDefaults({
...
+      color: {
+        type: 'float',
+        copySrc: true,
+        copyDst: true,
+        storage: true,
+        readWriteStorage: false,
+        bytes: 4,
+      },
+      colorRender: { blend: true, resolve: true, byteCost: 4, alignment: 2 },
+      multisample: false,
```
kainino0x: same

src/webgpu/format_info.ts
```diff
@@ -368,6 +424,34 @@ const kRegularTextureFormatInfo = formatTableWithDefaults({
...
+      color: {
+        type: 'float',
+        copySrc: true,
+        copyDst: true,
+        storage: true,
+        readWriteStorage: false,
+        bytes: 8,
+      },
+      colorRender: { blend: true, resolve: true, byteCost: 8, alignment: 2 },
+      multisample: false,
```
kainino0x: same

kainino0x (approved):

src/webgpu/format_info.ts
greggman: done

src/webgpu/format_info.ts
greggman: done

src/webgpu/format_info.ts
greggman: done

PR #4401, opened 2025-06-03
Additional test cases for encoder states

andyleiserson:
Additional tests related to encoder states.

* In the `encoder_open_state` tests, add the case of attempting to record commands onto a finished pass while the encoder is still open. Previously, the test only covered the case where both the pass and the encoder are finished.
* In the `call_after_successful_finish` test, add the case where the call made after the successful finish is another call to `finish`. This is motivated by a [bug in Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1972949).
* Test that it is possible to open passes on an invalid encoder.

The following tests are failing:
* In Chrome, the `encoder_open_state` tests for both render and compute passes. I noted this issue in https://github.com/gpuweb/gpuweb/issues/5207#issuecomment-2932996352. The issue description is a more convoluted case that is covered by an existing test, but this case is pretty much the same issue so I noted it in a comment. I also added a TODO on the other, existing test.
* In Safari, the `encoder_open_state:render_pass_commands` tests for the debug commands. Unlike the rest of the render pass commands, these cases were previously encoding the command onto the parent encoder rather than the render pass. This seemed unintentional, so I changed it, but maybe there's a special case here I'm not aware of.

I am not sure how to turn on "compatibility mode validation". I did try the tests in compatibility mode, if that's all this is.

<hr>

### DIFF
--- a/src/webgpu/api/validation/encoding/encoder_state.spec.ts
+++ b/src/webgpu/api/validation/encoding/encoder_state.spec.ts
@@ -98,7 +98,13 @@ g.test('call_after_successful_finish')
   .desc(`Test that encoding command after a successful finish generates a validation error.`)
   .params(u =>
     u
-      .combine('callCmd', ['beginComputePass', 'beginRenderPass', 'insertDebugMarker'])
+      .combine('callCmd', [
+        'beginComputePass',
+        'beginRenderPass',
+        'finishAndSubmitFirst',
+        'finishAndSubmitSecond',
+        'insertDebugMarker',
+      ])
       .beginSubcases()
       .combine('prePassType', ['compute', 'render', 'no-op'])
       .combine('IsEncoderFinished', [false, true])
@@ -115,8 +121,9 @@ g.test('call_after_successful_finish')
       pass.end();
     }
 
+    let buffer;
     if (IsEncoderFinished) {
-      encoder.finish();
+      buffer = encoder.finish();
     }
 
     switch (callCmd) {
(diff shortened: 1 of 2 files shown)


src/webgpu/api/validation/encoding/encoder_state.spec.ts
```diff
@@ -247,3 +286,85 @@ g.test('pass_end_twice,render_pass_invalid')
...
+  .params(u =>
+    u
+      .combine('pass0Type', ['compute', 'render'])
+      .combine('pass1Type', ['compute', 'render'])
+      .beginSubcases()
+      .combine('firstPassInvalid', [false, true])
+  )
+  .beforeAllSubcases(t => t.usesMismatchedDevice())
+  .fn(t => {
+    t.skipIfDeviceDoesNotSupportQueryType('timestamp');
```
greggman: why use a timestamp query here given that they are optional? I'm guessing because using a buffer is more work but it is a little strange to test something that is supposed to validate everywhere but if that implementation doesn't support timestamp-query then this path isn't test.

If I understand correctly, the test is trying to check begin works on an invalid encoder so you don't have to make the encode invalid by using beginXXXPass right?
Would an empty bindGroup work? setBindGroup(0, emptyBindGroupFromMismatchedDevice)

andyleiserson: I agree it would be better to use something else. I may have been looking at a `timestampWrites` test for whatever reason, and then used it as a basis for a new test without thinking it through carefully.

I'll plan to update the test.

greggman (approved):

PR #4404, opened 2025-06-11
Refactor depth_clamp_and_clip test

greggman:
The test had the comment that using `depthCompare: 'not-equal'` might not work so, switch the test to just reading the depth texture directly and comparing in the shader. This makes 4 cases pass on M1 Mac. Tested Intel Mac, AMD Mac, AMD Linux, GL Linux. For those there was no change. They were not failing the same 4 test cases that are fixed on M1 with this change.

Note: the hasStorageBuffers was removed because
maxStorageBuffersInFragmentShader can no longer be 0, even in compat.

Screenshot 2025-06-11 at 11 56 41

Interestingly, Safari TP doesn't fail the 4 cases that Chrome is failing without these changes.

Screenshot 2025-06-11 at 12 52 43


Kangz (approved):
LGTM

PR #4406, opened 2025-06-17
Add texture_1d support to textureSampleLevel

beaufortfrancois:
As discussed in https://github.com/gpuweb/gpuweb/issues/5001#issuecomment-2515450660, this PR adds tests in the CTS for texture_1d support to textureSampleLevel. This is to make sure this actually works as intended as we're implementing in Tint with https://issues.chromium.org/382514673

image

<hr>

- [x] Test behaves as expected in a WebGPU implementation. (If not passing, explain above.) It works with Chromium patched with https://dawn-review.googlesource.com/c/dawn/+/247034


src/webgpu/shader/execution/expression/call/builtin/textureSampleLevel.spec.ts
```diff
@@ -33,12 +34,100 @@ import {
...
+  .specURL('https://www.w3.org/TR/WGSL/#texturesamplelevel')
+  .desc(
+    `
+fn textureSampleLevel(t: texture_1d<f32>, s: sampler, coords: f32, level: f32) -> vec4<f32>
+
+Parameters:
+ * t  The sampled or depth texture to sample.
+ * s  The sampler type.
+ * coords The texture coordinates used for sampling.
+ * level The mip level, clamped to 0.
```
beaufortfrancois: Will `clamped to 0` be part of the spec? Do I need to do something about it in the CTS and Tint implementation?

beaufortfrancois: FYI Here's my naive thoughts regarding the spec: https://github.com/gpuweb/gpuweb/compare/main...beaufortfrancois:gpuweb:textureSampleLevel1d?expand=1

jrprice: I think "clamped to zero" would fall out of the generic mip level selection rules that apply to all textures. That said, I'm struggling to find exactly where that is written down in either the WGSL or API specs.

As for Tint, I don't think you need to do anything special, as I believe the backend APIs already do The Right Thing (i.e. clamp).

Ideally CTS would test out-of-bounds mip levels to ensure this is the case, but if we're not already doing this for other textures then that would be a separate issue.

src/webgpu/shader/execution/expression/call/builtin/textureSampleLevel.spec.ts
```diff
@@ -33,12 +34,105 @@ import {
...
+      .combine('modeV', kShortAddressModes)
+      .beginSubcases()
+      .combine('samplePoints', kSamplePointMethods)
+  )
+  .fn(async t => {
+    const { format, stage, samplePoints, modeU, modeV, filt: minFilter } = t.params;
+    t.skipIfTextureFormatAndDimensionNotCompatible(format, '1d');
+    skipIfTextureFormatNotSupportedOrNeedsFilteringAndIsUnfilterable(t, minFilter, format);
+
+    // We want at least 4 blocks or something wide enough for 3 mip levels.
```
jrprice, line 78: This comment seems inaccurate since we're only testing a single level here?

greggman: maybe leave it? If 1d textures ever support mip levels nothing needs to change here. We'd just change `chooseTextureSize`. It's not technically inaccurate. We're choosing a size that could support at least 3 mip levels. We're not actually making 3 mip levels.

src/webgpu/shader/execution/expression/call/builtin/textureSampleLevel.spec.ts
greggman: This test tests out-of-bounds mip levels. I'm not sure the comment is needed. Whatever number of mips the texture is, the test will test out of bounds mip levels.

src/webgpu/shader/execution/expression/call/builtin/textureSampleLevel.spec.ts
```diff
@@ -33,12 +34,105 @@ import {
...
+    const { texels, texture } = await createTextureWithRandomDataAndGetTexels(t, descriptor);
+    const sampler: GPUSamplerDescriptor = {
+      addressModeU: kShortAddressModeToAddressMode[modeU],
+      addressModeV: kShortAddressModeToAddressMode[modeV],
+      minFilter,
+      magFilter: minFilter,
+      mipmapFilter: minFilter,
+    };
+
+    const calls: TextureCall<vec1>[] = generateTextureBuiltinInputs1D(50, {
```
greggman, line 103: I think this code needs the mipLevel code to match the 2d code. Both in the call to genearteTextureBuiltinInputs1D and in the return code. That's what makes it test out of bounds mip levels. It doesn't exist in `textureSample` test because texture sample test chooses mip levels by derivatives.

src/webgpu/shader/execution/expression/call/builtin/textureSampleLevel.spec.ts
beaufortfrancois: I'll leave it then.

src/webgpu/shader/execution/expression/call/builtin/textureSampleLevel.spec.ts
beaufortfrancois: Done. Thanks for catching this!

src/webgpu/shader/execution/expression/call/builtin/textureSampleLevel.spec.ts
beaufortfrancois: I've updated it and used same prose as the other tests for `level`

PR #4408, opened 2025-06-24
Fix textureStore test

greggman:
It needs all storage readable texture formats
but it was filtering on read-write texture formats.

### DIFF
--- a/src/webgpu/shader/execution/expression/call/builtin/textureStore.spec.ts
+++ b/src/webgpu/shader/execution/expression/call/builtin/textureStore.spec.ts
@@ -100,7 +100,7 @@ g.test('texel_formats')
   )
   .fn(t => {
     const { format, stage, access, viewDimension, mipLevel } = t.params;
-    t.skipIfTextureFormatNotUsableAsReadWriteStorageTexture(format);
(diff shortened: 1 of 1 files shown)


src/webgpu/format_info.ts
```diff
@@ -1680,7 +1680,7 @@ type TextureFormatInfo_TypeCheck = {
  * * isTextureFormatResolvable
  * * isTextureFormatBlendable
  * * isTextureFormatMultisampled
- * * isTextureFormatUsableAsStorageFormat
```
shrekshao: Is this change intended?

greggman: yes, the comment was mis-spelled. Corrected that spelling.

greggman: actually you're right. Sigh. I'll remove that here and add an other PR

On GPUTest is `AsStorageTexture`. here it's `AsStorageFormat`. I think it should be renamed to `AsStorageTexture` but I see there are a few more

greggman: fixed

shrekshao (approved):

PR #4411, opened 2025-06-25
Refactor that not all storage texture support read-only

greggman:
The code assumed all textures marked as 'storage: true' could both do `read-only` and `write-only` but `bgra8norm` can never do `read-only`.

Refactor be able to separately check for read-only and write-only usage.

Issue:
* https://github.com/gpuweb/gpuweb/issues/5232
* https://github.com/gpuweb/cts/issues/4412

### DIFF
--- a/src/webgpu/api/operation/storage_texture/read_only.spec.ts
+++ b/src/webgpu/api/operation/storage_texture/read_only.spec.ts
@@ -553,7 +553,7 @@ g.test('basic')
   .fn(t => {
     const { format, shaderStage, dimension, depthOrArrayLayers } = t.params;
     t.skipIfTextureFormatNotSupported(format);
-    t.skipIfTextureFormatNotUsableAsStorageTexture(format);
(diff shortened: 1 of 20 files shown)


greggman:
Sorry this is big. Hopefully it's straight forward.

Two things I was wondering. 

1. Should I get rid of `xxxUsableAs(ReadOnly|WriteOnly|ReadWrite)StorageTexture` and just have `xxxUsableWithStorageAccessMode`. The later was written after the former. It seems like maybe it's a good idea to have one way to do things? wdyt?

2. I've wondered if all the `skipIfTetxureFormatDoesNotDoX` should internally call `skipIfTextureFormatNoSupported`. As it you need to call the first one manually. wdyt?

I thing I'd also like to rename `isTextureFormatPossiblyStorageReadable` but this PR was already too big.

shrekshao (approved):
The refactor makes sense and LGTM

PR #4416, opened 2025-06-28
tier1/2 add pipeline auto-layout tests.

greggman:
Note: these tests will fail on most browsers until tier1 formats are added to their WGSL code. WGSL is supposed to accept these formats regardless of whether or not the they are supported by the device. pipeline / bindGroupLayout creation is what prevents them from being used.

### DIFF
--- a/src/webgpu/api/validation/capability_checks/features/texture_formats_tier2.spec.ts
+++ b/src/webgpu/api/validation/capability_checks/features/texture_formats_tier2.spec.ts
@@ -69,7 +69,7 @@ g.test('pipeline_auto_layout,storage_texture')
   .desc(
     `
   Test creating a pipeline with auto layout with a storage texture binding format enabled by
-  'texture-formats-tier1' fails if the feature is not enabled.
+  'texture-formats-tier2' fails if the feature is not enabled.
   `
   )
   .params(u =>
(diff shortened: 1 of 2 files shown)


src/webgpu/api/validation/capability_checks/features/texture_formats_tier2.spec.ts
```diff
@@ -63,3 +64,69 @@ g.test('bind_group_layout,storage_binding_read_write_access')
       });
     }, !enable_feature);
   });
+
+g.test('pipeline_auto_layout,storage_texture')
+  .desc(
+    `
+  Test creating a pipeline with auto layout with a storage texture binding format enabled by
+  'texture-formats-tier1' fails if the feature is not enabled.
```
shrekshao:
```suggestion
  'texture-formats-tier2' fails if the feature is not enabled.
```

src/webgpu/api/validation/capability_checks/features/texture_formats_tier2.spec.ts
```diff
@@ -63,3 +64,69 @@ g.test('bind_group_layout,storage_binding_read_write_access')
...
+      .combine('format', kTextureFormatsTier2EnablesStorageReadWrite)
+      .combine('enable_feature', [true, false])
+      .beginSubcases()
+      .combine('isAsync', [false, true] as const)
+      .combine('type', ['compute', 'render'] as const)
+  )
+  .beforeAllSubcases(t => {
+    const { enable_feature } = t.params;
+    if (enable_feature) {
+      t.selectDeviceOrSkipTestCase('texture-formats-tier1');
```
shrekshao:
```suggestion
      t.selectDeviceOrSkipTestCase('texture-formats-tier2');
```

PR #4422, opened 2025-07-11
Fix subgroup shuffle (xor) access

petermcneeleychromium:
The subgroup shuffle split tests appear to perform bounds array access in the shader.
This will produce indeterminate value so these tests might produce different error for different devices.
(depending on robustness implementations)

The out of bounds access was when the input array is index (0-127) via the thread idx (0-255)

I have added additional loggin to this test in case we have more issues or we really do have a platform specific bug (and not a bug in the CTS).

crbug.com/417977426

### DIFF
--- a/src/webgpu/shader/execution/expression/call/builtin/subgroupShuffle.spec.ts
+++ b/src/webgpu/shader/execution/expression/call/builtin/subgroupShuffle.spec.ts
@@ -439,47 +439,59 @@ function checkCompute(
   numInvs: number,
   filter: (id: number, size: number) => boolean
 ): Error | undefined {
-  const mapping = new Map<number, number[]>();
+  const sub_unique_id_to_inv_idx = new Map<number, number[]>();
   const empty = [...iterRange(128, x => -1)];
-  for (let i = 0; i < numInvs; i++) {
-    const id = metadata[i];
-    const subgroup_id = metadata[i + numInvs];
-    const v = mapping.get(subgroup_id) ?? Array.from(empty);
-    v[id] = i;
-    mapping.set(subgroup_id, v);
+  for (let inv = 0; inv < numInvs; inv++) {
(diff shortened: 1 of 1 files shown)


src/webgpu/shader/execution/expression/call/builtin/subgroupShuffle.spec.ts
```diff
@@ -578,17 +590,17 @@ g.test('compute,split')
     const testcase = kPredicateCases[t.params.predicate];
     const wgThreads = t.params.wgSize[0] * t.params.wgSize[1] * t.params.wgSize[2];
 
-    let value = `input[lid]`;
+    let value = `input[lidx]`;
     if (t.params.op !== 'subgroupShuffle') {
-      value = `subgroupBroadcastFirst(input[lid])`;
+      value = `subgroupBroadcastFirst(input[lidx])`;
```
petermcneeleychromium, line 595: This is the suspected out of bounds access.

src/webgpu/shader/execution/expression/call/builtin/subgroupShuffle.spec.ts
```diff
@@ -608,24 +620,26 @@ var<storage, read_write> metadata : Metadata;
...
   if ${testcase.cond} {
-    output.res[lid] = ${t.params.op}(lid, value);
+    output.res[lidx] = ${t.params.op}(lidx, value);
   } else {
     return;
   }
 }`;
 
-    const inputArray = new Uint32Array([...iterRange(128, x => x)]);
+    const inputArray = new Uint32Array([...iterRange(wgThreads, x => x % 128)]);
```
petermcneeleychromium, line 642: buffer size now is the size of the workgroup to avoid the bounds access.

greggman (approved):
lgtm

Issue #4421, opened 2025-07-11
Texture component swizzle 🥤


beaufortfrancois:
This issue tracks CTS tests for [Texture component swizzle proposal](https://github.com/gpuweb/gpuweb/blob/main/proposals/texture-component-swizzle.md):
- [ ] Works with all textures and is validated out.
- [ ] Swizzle for storage textures.
- [ ] Depth textures used as texture_depth
- [ ] Depth textures used as texture_2d<f32>
- [ ] Stencil textures used as texture_2d<u32>
- [ ] Multisampled color textures.
- [ ] Multisampled depth textures.
- [ ] Also the various ways to use the view: Load/Fetch, Sampling, Gather, maybe the compare versions etc.

beaufortfrancois:
@greggman As I'm adding [support in dawn.node](https://dawn-review.googlesource.com/c/dawn/+/253154), would you be willing to write some CTS tests for the proposed texture component swizzle feature?

greggman:
Yes, I'm happy to write some tests

beaufortfrancois:
Thank you very much!

greggman:
Do we know that swizzle works with depth textures?

What I see on Mac M1 in Chrome Canary using `texture-component-swizzle` it doesn't seem to make any sense.

options:

1. GPU gathers four, 4 channel (vec4f) texels, swizzles each, takes the R value from each, compares

   this would make setting any swizzle to 'g', 'b', 'a' implementation defined
   since only `r` is defined.

2. GPU gathers four, 1 channel depth values, swizzles, compares

   this would mean the comparison happens on the swizzled value so passing in `one` would
   fail `ref(0.5) < value` and `pass ref(0.5) > value`

3. GPU gathers four, 1 channel depth values, compares, swizzles

   this would mean a swizzle of `one` or `zero` overrides the comparison

So here's M1 via Canary

```
texels: 0.2, 0.4, 0.6, 0.8
gather: a    b    g    r     

compare: less , ref: 0.5
1,1,0,0 swizzle: {}
1,1,1,1 swizzle: {r:'one',b:'one',g:'one',a:'one'}
0,0,0,0 swizzle: {r:'zero',b:'zero',g:'zero',a:'zero'}
1,1,1,1 swizzle: {r:'a',b:'g',g:'b',a:'r'}
1,1,0,0 swizzle: {r:'r',b:'r',g:'r',a:'r'}
[...]

PR #4427, opened 2025-07-17
Add texture-component-swizzle tests

greggman:
This is NOT ready to be merged. The extension first needs to be approved and @webgpu/types updated. Then the PR updated to removed the type hacks.

Note: Should swizzle in any form be disallowed without this extension or should it be allowed but only generate a validation error if it's not the identity?

* Issue: #4421


src/webgpu/api/validation/capability_checks/features/texture_component_swizzle.spec.ts
```diff
@@ -0,0 +1,490 @@
...
+  `
+  )
+  .fn(t => {
+    const texture = t.createTextureTracked({
+      format: 'rgba8unorm',
+      size: [1],
+      usage: GPUTextureUsage.COPY_DST | GPUTextureUsage.TEXTURE_BINDING,
+    });
+    t.expectValidationError(() => {
+      texture.createView({ swizzle: {} });
```
beaufortfrancois: Depending on https://github.com/gpuweb/gpuweb/pull/5252, we should also try a default swizzle.

beaufortfrancois:
> > > * I only tested in a compute shader. I did not test in a vertex shader nor a fragment shader. Should I?
> > 
> > 
> > Yes. I initially also tested it in a compute shader but @Kangz strongly suggested [1] in https://dawn-review.googlesource.com/c/dawn/+/249434/comment/10136679_ddbd080c/ to not rely on the accuracy of the division,
> 
> I don't know what that means. There is no division in my tests.

Since you use `textureStore` it may be fine actually. @Kangz will know more.

> 
> > This is part of the ["Open Questions" section](https://github.com/gpuweb/gpuweb/blob/main/proposals/texture-component-swizzle.md#open-questions). Without CTS tests to see how this would react, I would allow it for now and then see how it works in practice. We could also raise a validation error if G, B, or A swizzles are non default for depth texture binding.
> 
> > > * Issue: [Texture component swizzle 🥤 #4421](https://github.com/gpuweb/cts/issues/4421)
> 
> I think maybe I didn't get my point across
> 
> ```wgsl
> @group(0) @binding(0) var t1: texture_2d<f32>;  // this as float or unfilterered-float texture
> @group(0) @binding(1) var t2: texture_depth;    // this is a depth texture
> ```
> 
> * `t1` above works and is tested already with depth formats.
> * `t2` above only returns `f32`, not `vec4f`. There is nothing to swizzle. It doesn't return an R or G or B or A so swizzling makes no sense. So I asked, ignore the swizzle or generate a validation error if you try to provide one? I'd think generating a validation error is more inline with the rest of the API.

I assumed you meant `texture_depth_2d` in your WGSL example, not `texture_depth`. My understanding was that the first and unique value would be the R channel so that's why I would check only G, B, A swizzles. We could also generate a validation error if a swizzle is provided at all for `texture_depth_*` in the spec and relax this if developers actually use swizzle R for those depth textures.

greggman:
> The division I mentioned @beaufortfrancois was the one done by the fragment shader postfix to turn the `vec4f` into an rgba8unorm. This would still happen using `textureStore` I believe?

I'm using textureStore to a rgba32float, rgba32uint, or rgba32sint as appropriate. I'm not sure what division we are talking about here.

> We should test the swizzle with the various builtins to make sure that it works in the various APIs. Right now the investigation was on what the APIs expose but they don't exhaustively say what happens with the whole zoo of texture features, so we need to validate that it is consistent across backends before exposing it.

Ok, but how far do we want to go? The current texture tests take ~1/3 of the CTS time. If I add  swizzle tests to all possible combination of stuff that will go up by 2x to 3x

My current test already takes 8 minutes on an M1 Mac because of all the combinations.

* all texture formats ⓧ
* all builtins ⓧ
* 3 stages ⓧ
* 24 swizzles (undefined, identity, all red, all green, all blue, all alpha, all zero, all one, abgr, rot1, rot2, rot3, one in each slot (4), zero in each slot (4), undefined in each slot (4) ⓧ
* 3 alternates (this is to test that different swizzles on the same texture used in the same stage get different results)

I suppose we could test the 3 alternates as part of the 24 swizzles (so 27 total) instead of 24x3 which it is now. So that would lower it from 8 minutes to ~3 but then we need to add in a few more things that aren't being testing which will likely bring it back up to at least 5-6 mins.

It's currently only testing with 2d textures. It is not testing 1d, 3d, 2d-array, cube, cube-array (so adding those would be ~6x the combos). It's also not testing if mip settings affect it, nor layer settings, nor other sampler settings, nor re-intepretation (view is '-srgb'), and it's also not testing `texture_depth_2d` except for comparisons (that will ~double the tests)

greggman:
I think this PR is ready for at least the current version of the spec proposal. It's only slow if `debug=1` is passed. It's fast enough otherwise (~25 seconds for all tests combined on M1)

src/webgpu/api/validation/capability_checks/features/texture_component_swizzle.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+    });
+  });
```
beaufortfrancois: According to https://github.com/gpuweb/gpuweb/issues/5264#issuecomment-3128939986, we're going this way:
```suggestion
g.test('no_default_swizzle')
  .desc(
    `
  Test that if texture-component-swizzle is not enabled, having a non-default swizzle property generates a validation error.
  `
  )
  .params(u =>
    u
      .beginSubcases()
      .combine('swizzleSpec', kSwizzleTests)
  )
  .fn(t => {
    const { swizzleSpec } = t.params;
    const swizzle = swizzleSpecToGPUTextureComponentSwizzle(swizzleSpec);
    const texture = t.createTextureTracked({
      format: 'rgba8unorm',
      size: [1],
      usage: GPUTextureUsage.COPY_DST | GPUTextureUsage.TEXTURE_BINDING,
    });
    const shouldError = !isIdentitySwizzle(swizzle);
    t.expectValidationError(() => {
      texture.createView({ swizzle });
    }, shouldError);
  });
```

FYI Here's https://dawn-review.googlesource.com/c/dawn/+/253694

src/webgpu/api/validation/capability_checks/features/texture_component_swizzle_operation.spec.ts
```diff
@@ -0,0 +1,675 @@
+export const description = `
+Operational tests for the 'texture-component-swizzle' feature.
```
beaufortfrancois, line 2: Should it live in `api/operation` instead of `api/validation`?

greggman: yea, maybe. Preference for location?

* `src/webgpu/api/operation/features/texture_component_swizzle.spec.ts` ?
* `src/webgpu/api/operation/texture/texture_component_swizzle.spec.ts` ?
* other?

beaufortfrancois: I don't have strong preferences to be honest. Both suggestions look good to me.

beaufortfrancois:
@greggman FYI https://chromium-review.googlesource.com/c/chromium/src/+/6818591 has just been merged in Chromium. It should be in next Chrome Canary.

greggman:
updated the tests for compat

beaufortfrancois:
@greggman Shall we land these tests to see how different backends react in Chromium or are there more you want to add first?

greggman:
Throwing out ideas, maybe the tests/spec should have 3 states

1. `adapter.features` does **not** have `texture-component-swizzle`

   in this case, you can ignore the `swizzle` property. This is what browsers do today

2. `adapter.features` has `texture-component-swizzle` but you don't request it

   in this case, you must validate the `swizzle` property is the identity if set

3. `adapter.features` has `texture-component-swizzle` and you request it

   use the feature

I don't know if state 1 is desirable or if we're just going to update the spec to require swizzle validation even if the feature doesn't exist.

greggman:
FYI, checking D3D12 NVidia, swizzle doesn't appear to work with stencil textures. Like `{r: 'zero', g: 'zero', b: 'zero', a: 'zero' }` doesn't return 0

From WG meeting: We should make the tests skip "must be identity" validation when the feature does not exist until the spec is approved. But we can check in the tests for when the feature exists. So I'll update the tests.

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
```diff
@@ -3197,14 +3224,36 @@ function createTextureFromTexelViewsLocal(
...
   if (isDepthOrStencilTextureFormat(desc.format) || desc.sampleCount! > 1) {
-    modifiedDescriptor.usage = desc.usage | GPUTextureUsage.RENDER_ATTACHMENT;
+    modifiedDescriptor.usage =
+      desc.usage | GPUTextureUsage.RENDER_ATTACHMENT | GPUTextureUsage.COPY_SRC;
+  }
+  let texture = createTextureFromTexelViews(t, texelViews, modifiedDescriptor);
+  if ((texture.usage & ~GPUTextureUsage.COPY_DST) !== (desc.usage & ~GPUTextureUsage.COPY_DST)) {
+    const copy = t.createTextureTracked({
+      ...desc,
+      usage: desc.usage | GPUTextureUsage.COPY_DST,
```
beaufortfrancois: This should help with errors I get locally in http://127.0.0.1:8080/standalone/?runnow=1&q=webgpu:shader,execution,expression,call,builtin,textureLoad:multisampled:stage=%22c%22;texture_type=%22texture_depth_multisampled_2d%22;format=%22depth16unorm%22 for instance.

```suggestion
      ...modifiedDescriptor,
      usage: modifiedDescriptor.usage | GPUTextureUsage.COPY_DST,
```
as in https://github.com/gpuweb/cts/commit/1a4dae0e7b306375623a2d7abe2fec779e9763c4

src/webgpu/api/operation/texture_view/texture_component_swizzle.spec.ts
```diff
@@ -0,0 +1,681 @@
...
+  gatherChannel: number
+) {
+  if (!isDepthOrStencilTextureFormat(format)) {
+    return {
+      validMask: [1, 1, 1, 1],
+      altResult: [0, 0, 0, 0],
+    };
+  }
+  const swizzleByChannel = [swizzle.r ?? 'r', swizzle.g ?? 'g', swizzle.b ?? 'b', swizzle.a ?? 'a'];
+  const rgbaDepthMask = {
```
beaufortfrancois: I think there's an issue with this code as I get the following errors locally for cases like webgpu:api,operation,texture_view,texture_component_swizzle:read_swizzle:format="stencil8";func="textureGather"
```
EXCEPTION: Error: Unexpected validation error occurred: The texture view's component swizzle r (ComponentSwizzle::G) must be ComponentSwizzle::R when usage includes TextureUsage::RenderAttachment or TextureUsage::StorageBinding.
```
We should not test swizzles with `g` in the `r` channel for formats like "stencil8"

greggman: ~~That sounds like you're no running with all the changes from this PR. That error should happen only if the patches to texture_utils.ts, texture.ts, etc have not been applied. I don't get those errors locally on 3 different machines (M1 Mac, Windows NVidia, Linux AMD Vulkan)~~

The other change you suggest is causing this error. Remove that other change and this error should go away

src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts
greggman: The modifiedDesc is used to be able to render to a texture. We then make a copy that is not renderable (doesn't have RENDER_ATTACHMENT). Otherwise we start getting errors that the texture can't be used with swizzle.

greggman: As to that specific error, the simplest fix would be to add `RENDER_ATTACHMENT` at `textureLoad.spec.ts` line 372. That said, AFAICT that's not a requirement and so would be a bug in Chrome. I'll do some checking

greggman: Ok checked. Yes, the proper fix is to add RENDER_ATTACHMENT in textureLoad.spec.ts. Added

greggman (commented):
.

beaufortfrancois:
After applying your fix, I get the following errors in https://ci.chromium.org/ui/p/chromium/builders/try/mac-arm64-dawn-rel/19727/test-results:

```
webgpu_cts_integration_test_base.py(419): AssertionError: webgpu:shader,execution,expression,call,builtin,textureLoad:multisampled: failed
  EXCEPTION: Error: Unexpected validation error occurred: [Texture (unlabeled 8x8 px, TextureFormat::Depth24PlusStencil8)] sample count (4) is not 1 when copying to or from a buffer.
```

for those tests:

```
shader,execution,expression,call,builtin,texture_utils:readTextureToTexelViews:srcFormat="depth24plus-stencil8";texelViewFormat="rgba32float";viewDimension="2d";sampleCount=4
shader,execution,expression,call,builtin,texture_utils:readTextureToTexelViews:srcFormat="depth24plus-stencil8";texelViewFormat="rgba32float";viewDimension="2d";sampleCount=4
shader,execution,expression,call,builtin,texture_utils:readTextureToTexelViews:srcFormat="depth24plus";texelViewFormat="rgba32float";viewDimension="2d";sampleCount=4
shader,execution,expression,call,builtin,texture_utils:readTextureToTexelViews:srcFormat="depth24plus";texelViewFormat="rgba32float";viewDimension="2d";sampleCount=4
shader,execution,expression,call,builtin,texture_utils:readTextureToTexelViews:srcFormat="r8sint";texelViewFormat="rgba32sint";viewDimension="2d";sampleCount=4
shader,execution,expression,call,builtin,texture_utils:readTextureToTexelViews:srcFormat="r8sint";texelViewFormat="rgba32sint";viewDimension="2d";sampleCount=4
shader,execution,expression,call,builtin,texture_utils:readTextureToTexelViews:srcFormat="r8uint";texelViewFormat="rgba32uint";viewDimension="2d";sampleCount=4
shader,execution,expression,call,builtin,texture_utils:readTextureToTexelViews:srcFormat="r8uint";texelViewFormat="rgba32uint";viewDimension="2d";sampleCount=4
shader,execution,expression,call,builtin,texture_utils:readTextureToTexelViews:srcFormat="r8unorm";texelViewFormat="rgba32float";viewDimension="2d";sampleCount=4
shader,execution,expression,call,builtin,texture_utils:readTextureToTexelViews:srcFormat="r8unorm";texelViewFormat="rgba32float";viewDimension="2d";sampleCount=4
shader,execution,expression,call,builtin,texture_utils:readTextureToTexelViews:srcFormat="stencil8";texelViewFormat="stencil8";viewDimension="2d";sampleCount=4
shader,execution,expression,call,builtin,texture_utils:readTextureToTexelViews:srcFormat="stencil8";texelViewFormat="stencil8";viewDimension="2d";sampleCount=4
shader,execution,expression,call,builtin,textureLoad:multisampled:stage="c";texture_type="texture_depth_multisampled_2d";format="depth24plus-stencil8"
shader,execution,expression,call,builtin,textureLoad:multisampled:stage="c";texture_type="texture_depth_multisampled_2d";format="depth24plus-stencil8"
shader,execution,expression,call,builtin,textureLoad:multisampled:stage="c";texture_type="texture_depth_multisampled_2d";format="depth32float-stencil8"
shader,execution,expression,call,builtin,textureLoad:multisampled:stage="c";texture_type="texture_depth_multisampled_2d";format="depth32float-stencil8"
shader,execution,expression,call,builtin,textureLoad:multisampled:stage="f";texture_type="texture_depth_multisampled_2d";format="depth24plus-stencil8"
shader,execution,expression,call,builtin,textureLoad:multisampled:stage="f";texture_type="texture_depth_multisampled_2d";format="depth24plus-stencil8"
shader,execution,expression,call,builtin,textureLoad:multisampled:stage="f";texture_type="texture_depth_multisampled_2d";format="depth32float-stencil8"
shader,execution,expression,call,builtin,textureLoad:multisampled:stage="f";texture_type="texture_depth_multisampled_2d";format="depth32float-stencil8"
shader,execution,expression,call,builtin,textureLoad:multisampled:stage="v";texture_type="texture_depth_multisampled_2d";format="depth24plus-stencil8"
shader,execution,expression,call,builtin,textureLoad:multisampled:stage="v";texture_type="texture_depth_multisampled_2d";format="depth24plus-stencil8"
shader,execution,expression,call,builtin,textureLoad:multisampled:stage="v";texture_type="texture_depth_multisampled_2d";format="depth32float-stencil8"
shader,execution,expression,call,builtin,textureLoad:multisampled:stage="v";texture_type="texture_depth_multisampled_2d";format="depth32float-stencil8"
```

beaufortfrancois:
And I'm getting more on Windows (https://ci.chromium.org/ui/p/chromium/builders/try/win-dawn-rel/49419/test-results):

```
webgpu_cts_integration_test_base.py(419): AssertionError: webgpu:shader,execution,expression,call,builtin,texture_utils:readTextureToTexelViews: failed
  EXCEPTION: Error: Unexpected validation error occurred: The usage (TextureUsage::(CopyDst|TextureBinding)) of a multisampled texture doesn't include (TextureUsage::RenderAttachment).
```

for the following tests:

```
gpu_tests.webgpu_cts_integration_test.WebGpuCtsIntegrationTest.webgpu:shader,execution,expression,call,builtin,texture_utils:readTextureToTexelViews:srcFormat="depth24plus-stencil8";texelViewFormat="rgba32float";viewDimension="2d";sampleCount=4
gpu_tests.webgpu_cts_integration_test.WebGpuCtsIntegrationTest.webgpu:shader,execution,expression,call,builtin,texture_utils:readTextureToTexelViews:srcFormat="depth24plus";texelViewFormat="rgba32float";viewDimension="2d";sampleCount=4
gpu_tests.webgpu_cts_integration_test.WebGpuCtsIntegrationTest.webgpu:shader,execution,expression,call,builtin,texture_utils:readTextureToTexelViews:srcFormat="r8sint";texelViewFormat="rgba32sint";viewDimension="2d";sampleCount=4
gpu_tests.webgpu_cts_integration_test.WebGpuCtsIntegrationTest.webgpu:shader,execution,expression,call,builtin,texture_utils:readTextureToTexelViews:srcFormat="r8uint";texelViewFormat="rgba32uint";viewDimension="2d";sampleCount=4
gpu_tests.webgpu_cts_integration_test.WebGpuCtsIntegrationTest.webgpu:shader,execution,expression,call,builtin,texture_utils:readTextureToTexelViews:srcFormat="r8unorm";texelViewFormat="rgba32float";viewDimension="2d";sampleCount=4
gpu_tests.webgpu_cts_integration_test.WebGpuCtsIntegrationTest.webgpu:shader,execution,expression,call,builtin,texture_utils:readTextureToTexelViews:srcFormat="stencil8";texelViewFormat="stencil8";viewDimension="2d";sampleCount=4
gpu_tests.webgpu_cts_integration_test.WebGpuCtsIntegrationTest.webgpu:shader,execution,expression,call,builtin,textureLoad:multisampled:stage="c";texture_type="texture_depth_multisampled_2d";format="depth24plus-stencil8"

**Comment by beaufortfrancois** [comment]
When applying the following fix (as in https://github.com/gpuweb/cts/commit/c805a27dc697c80c15db31037ae57553f4ab7948), I get rid of errors happening specifically in compatibility mode for depth textures. See results at https://dawn-review.googlesource.com/c/dawn/+/257115?checksResultsFilter=swizzle&tab=checks

```js
    if (t.isCompatibility && isDepthTextureFormat(format)) {
      t.skipIf(
        func === 'textureLoad',
        'textureLoad can not be used with depth textures in compatibility mode'
      );
      t.skipIf(
        func === 'textureGather' || func === 'textureSample' || func === 'textureSampleLevel',
        'texture_depth_xx can not be used with non-comparison samplers in compatibility mode'
      );
    }
```

src/webgpu/api/operation/texture_view/texture_component_swizzle.spec.ts
```diff
@@ -0,0 +1,681 @@
+export const description = `
+Operational tests for the 'texture-component-swizzle' feature.
+
+Test that:
+* when the feature is on, swizzling is applied correctly.
+
+TODO:
+* test stencil aspect of depth-stencil formats
+* test texture_depth_xxx with textureSample
```
beaufortfrancois: I believe texture_depth_xx are actually tested with textureSample in this PR. Am I missing something?

greggman:
fyi, `textureGather` for all the uint/sint formats including stencil8 fail on AMD Mac. I think this is a known issue

beaufortfrancois:
@greggman I'm OOO today but did you try running src/webgpu/api/operation/texture_view/texture_component_swizzle.spec.ts tests in compatibility mode? If it still fails, you may need this fix: https://github.com/gpuweb/cts/pull/4427#issuecomment-3178743377

src/webgpu/api/operation/texture_view/texture_component_swizzle.spec.ts
beaufortfrancois: @greggman gentle ping in case it was missed.

beaufortfrancois:
Thank you @greggman for pursuing this!

Based on the [following results](https://dawn-review.googlesource.com/c/dawn/+/257855), shall we do something about it like disabling texture compoment swizzle feature entirely on some platforms or for combinations of formats & platforms?

```
crbug.com/0000 [ android-pixel-4 ] webgpu:api,operation,texture_view,texture_component_swizzle:read_swizzle:format="depth16unorm";func="textureSampleCompare" [ Failure ]
crbug.com/0000 [ android-pixel-4 ] webgpu:api,operation,texture_view,texture_component_swizzle:read_swizzle:format="depth16unorm";func="textureSampleCompareLevel" [ Failure ]
crbug.com/0000 [ android-pixel-4 ] webgpu:api,operation,texture_view,texture_component_swizzle:read_swizzle:format="depth24plus";func="textureSampleCompare" [ Failure ]
crbug.com/0000 [ android-pixel-4 ] webgpu:api,operation,texture_view,texture_component_swizzle:read_swizzle:format="depth24plus";func="textureSampleCompareLevel" [ Failure ]
crbug.com/0000 [ android-pixel-4 ] webgpu:api,operation,texture_view,texture_component_swizzle:read_swizzle:format="depth24plus-stencil8";func="textureSampleCompare" [ Failure ]
crbug.com/0000 [ android-pixel-4 ] webgpu:api,operation,texture_view,texture_component_swizzle:read_swizzle:format="depth24plus-stencil8";func="textureSampleCompareLevel" [ Failure ]
crbug.com/0000 [ android-pixel-4 ] webgpu:api,operation,texture_view,texture_component_swizzle:read_swizzle:format="depth32float";func="textureSampleCompare" [ Failure ]
crbug.com/0000 [ android-pixel-4 ] webgpu:api,operation,texture_view,texture_component_swizzle:read_swizzle:format="depth32float";func="textureSampleCompareLevel" [ Failure ]
crbug.com/0000 [ android-pixel-4 ] webgpu:api,operation,texture_view,texture_component_swizzle:read_swizzle:format="depth32float-stencil8";func="textureSampleCompare" [ Failure ]
crbug.com/0000 [ android-pixel-4 ] webgpu:api,operation,texture_view,texture_component_swizzle:read_swizzle:format="depth32float-stencil8";func="textureSampleCompareLevel" [ Failure ]
crbug.com/0000 [ amd mac ] webgpu:api,operation,texture_view,texture_component_swizzle:read_swizzle:format="r16sint";func="textureGather" [ Failure ]
crbug.com/0000 [ amd mac ] webgpu:api,operation,texture_view,texture_component_swizzle:read_swizzle:format="r32uint";func="textureGather" [ Failure ]
crbug.com/0000 [ amd mac ] webgpu:api,operation,texture_view,texture_component_swizzle:read_swizzle:format="r8sint";func="textureGather" [ Failure ]
crbug.com/0000 [ amd mac ] webgpu:api,operation,texture_view,texture_component_swizzle:read_swizzle:format="r8uint";func="textureGather" [ Failure ]
crbug.com/0000 [ amd mac ] webgpu:api,operation,texture_view,texture_component_swizzle:read_swizzle:format="rg16sint";func="textureGather" [ Failure ]
crbug.com/0000 [ amd mac ] webgpu:api,operation,texture_view,texture_component_swizzle:read_swizzle:format="rg16uint";func="textureGather" [ Failure ]
crbug.com/0000 [ amd mac ] webgpu:api,operation,texture_view,texture_component_swizzle:read_swizzle:format="rg32sint";func="textureGather" [ Failure ]
crbug.com/0000 [ amd mac ] webgpu:api,operation,texture_view,texture_component_swizzle:read_swizzle:format="rg8sint";func="textureGather" [ Failure ]
crbug.com/0000 [ amd mac ] webgpu:api,operation,texture_view,texture_component_swizzle:read_swizzle:format="rg8uint";func="textureGather" [ Failure ]
crbug.com/0000 [ amd mac ] webgpu:api,operation,texture_view,texture_component_swizzle:read_swizzle:format="rgb10a2uint";func="textureGather" [ Failure ]
crbug.com/0000 [ amd mac ] webgpu:api,operation,texture_view,texture_component_swizzle:read_swizzle:format="rgba16sint";func="textureGather" [ Failure ]
crbug.com/0000 [ amd mac ] webgpu:api,operation,texture_view,texture_component_swizzle:read_swizzle:format="rgba16uint";func="textureGather" [ Failure ]
crbug.com/0000 [ amd mac ] webgpu:api,operation,texture_view,texture_component_swizzle:read_swizzle:format="rgba32sint";func="textureGather" [ Failure ]
crbug.com/0000 [ amd mac ] webgpu:api,operation,texture_view,texture_component_swizzle:read_swizzle:format="rgba32uint";func="textureGather" [ Failure ]
crbug.com/0000 [ amd mac ] webgpu:api,operation,texture_view,texture_component_swizzle:read_swizzle:format="rgba8sint";func="textureGather" [ Failure ]
crbug.com/0000 [ amd mac ] webgpu:api,operation,texture_view,texture_component_swizzle:read_swizzle:format="rgba8uint";func="textureGather" [ Failure ]
crbug.com/0000 [ amd mac ] webgpu:api,operation,texture_view,texture_component_swizzle:read_swizzle:format="stencil8";func="textureGather" [ Failure ]
```

beaufortfrancois:
@greggman As noted in https://github.com/gpuweb/gpuweb/blob/main/proposals/texture-component-swizzle.md#open-questions, shall we test whether texture component swizzle requires new validation rules if the view is multisampled?

beaufortfrancois:
> > @greggman As noted in https://github.com/gpuweb/gpuweb/blob/main/proposals/texture-component-swizzle.md#open-questions, shall we test whether texture component swizzle requires new validation rules if the view is multisampled?
> 
> You can not swizzle a multisample texture because a multisample texture requires `RENDER_ATTACHMENT` usage: (https://www.w3.org/TR/webgpu/#abstract-opdef-validating-gputexturedescriptor)

Oh! Good catch! Then, we can remove it from the proposal then. @Kangz FYI

greggman:
> For swizzle, what matters is the usage of the view, not the usage of the texture. They are not necessarily the same because the view's usage can be a subset if [`GPUTextureViewDescriptor.usage`](https://www.w3.org/TR/webgpu/#dom-gputextureviewdescriptor-usage) is used.

Oh! I think this is the first case I'm aware of where the `view.usage` is actually used?

I updated the tests.

In other news though, my M1 Mac is having issues similar to when I tested the texture builtins. Everything is going fine and then suddenly some tests take forever. I had one take 18 seconds. But, if I re-run the test it passes. As it is, it gets into this state where a few tests timeout. It could be because it's a chromium build?

greggman:
checked, the speed issue is chromium. Canary doesn't have it.

greggman:
I have a local change for expected depth to always be `D,0,0,1`. If you, @beaufortfrancois  have an impl I can try it out.

greggman:
Uploaded a version that expects `D, 0, 0, 1` instead of `D, ?, ?, ?` when `texture-component-swizzle` is enabled.

beaufortfrancois:
> @greggman And most importantly, does it work on all backends with this PR or do we need to update Dawn?

FYI I've started https://dawn-review.googlesource.com/c/dawn/+/260516?tab=checks but it's not finished yet as I'm writing this.

greggman:
> > @greggman And most importantly, does it work on all backends with this PR or do we need to update Dawn?
> 
> FYI I've started https://dawn-review.googlesource.com/c/dawn/+/260516?tab=checks but it's not finished yet as I'm writing this.

These are failing on Mac but if download [your CL](https://dawn-review.googlesource.com/c/dawn/+/258594) and run locally everything passes. Are you sure [https://dawn-review.googlesource.com/c/dawn/+/260516](https://dawn-review.googlesource.com/c/dawn/+/260516) incorporates your changes?

greggman:
Oh, I guess they are only supposed to pass on M1 Mac, which they are. Intel and AMD Mac are buggy and this has not yet been implemented on other backends.

beaufortfrancois:
> These are failing on Mac but if download [your CL](https://dawn-review.googlesource.com/c/dawn/+/258594) and run locally everything passes. Are you sure https://dawn-review.googlesource.com/c/dawn/+/260516 incorporates your changes? 

I'm unable to make this CTS roll with my CL changes sadly ;( Is this something you can try to see if Macs behave better?

beaufortfrancois:
@greggman I've uploaded https://chromium-review.googlesource.com/c/chromium/src/+/6958569 for review that switches to the WGSL syntax proposed in https://github.com/gpuweb/gpuweb/pull/5319 so that we can write `createView({swizzle: 'rgba'})`. I've tried it locally and it works fine.

We'd need to update these tests to use the new syntax (there's no `"u"` anymore) and add a new test to check invalid swizzles (`"rgbA"`, `"rgb"`, `""`, `null`, etc.). Let me know if you need help.

src/webgpu/api/validation/capability_checks/features/texture_component_swizzle.spec.ts
```diff
@@ -0,0 +1,231 @@
...
+export const g = makeTestGroup(UniqueFeaturesOrLimitsGPUTest);
+
+g.test('invalid_swizzle')
+  .desc(
+    `
+  Test that setting an invalid swizzle value on a texture view throws an exception.
+  `
+  )
+  .params(u =>
+    u.beginSubcases().combine('invalidSwizzle', ['rgbA', 'rgb', 'rgba01', '', 1234, null])
```
beaufortfrancois: As highlighted in https://chromium-review.googlesource.com/c/chromium/src/+/6958569/comment/47daa61a_23127924/, we should also test potential lossy cast cases such as `ɲ` (0x0272) and `r` (0x72)

beaufortfrancois (changes requested):

src/webgpu/api/validation/capability_checks/features/texture_component_swizzle.spec.ts
```diff
@@ -0,0 +1,250 @@
...
+      '𝓇ℊ𝒷𝒶',
+      'ⓡⓖⓑⓐ',
+      'ｒｇｂａ',
+      'ʳᵍᵇᵃ',
+      'rgba',
+      1234,
+      0x72676261, // big endian rgba
+      0x71726772, // little endian rgba
+      null,
+    ])
```
kainino0x, line 54:
```
'',
'rgba\0',
'rgb\0',
'\0\0\0\0',
```

src/webgpu/api/validation/capability_checks/features/texture_component_swizzle.spec.ts
```diff
@@ -0,0 +1,250 @@
...
+      'ɲgba', // r with 0x200 added to each code point to make sure values are not truncated.
+      'ɲɧɢɡ', // rgba with 0x200 added to each code point to make sure values are not truncated.
+      '𝐫𝐠𝐛𝐚', // various unicode values that normalize to rgba
+      '𝑟𝑔𝑏𝑎',
+      '𝗋𝗀𝖻𝖺',
+      '𝓇ℊ𝒷𝒶',
+      'ⓡⓖⓑⓐ',
+      'ｒｇｂａ',
+      'ʳᵍᵇᵃ',
+      'rgba',
```
kainino0x: This one looks like ascii, and copies like it too. Whatever it's intended to be, might need to write it with escape codes

src/webgpu/api/validation/capability_checks/features/texture_component_swizzle.spec.ts
```diff
@@ -0,0 +1,250 @@
...
+
+g.test('only_identity_swizzle')
+  .desc(
+    `
+  Test that if texture-component-swizzle is not enabled, having a non-default swizzle property generates a validation error.
+  `
+  )
+  .params(u => u.beginSubcases().combine('swizzle', kSwizzleTests))
+  .fn(t => {
+    // MAINTENANCE_TODO: Remove this check if the spec is updated to say that all implementations must validate this.
```
kainino0x, line 82: I think the spec will require this, even though this test will fail on browsers that haven't implemented swizzle at all yet. We'll just have to call that noncompliant even though normally we say browsers don't have to implement optional features at all[1]. If we didn't do this, then we would just ignore the swizzle option when the feature is not enabled, which would be too confusing (code looks like it does something different from what it does). Only other option would be to only enforce it as a warning, which isn't what we've done in the past though technically an option if we want [1].

See `unclippedDepth` for an analogue.

src/webgpu/api/validation/capability_checks/features/texture_component_swizzle.spec.ts
greggman: removed

src/webgpu/api/validation/capability_checks/features/texture_component_swizzle.spec.ts
greggman: added

beaufortfrancois:
@greggman With https://dawn-review.googlesource.com/c/dawn/+/261634 and https://dawn-review.googlesource.com/c/dawn/+/258594 now merged in Dawn, can you run a roll similar to https://dawn-review.googlesource.com/c/dawn/+/257855?tab=checks to see if we're good?

beaufortfrancois:
Thank you very much @greggman for https://dawn-review.googlesource.com/c/dawn/+/264016!

It looks like `r8snorm`, `rg8snorm`, and `rgba8snorm` texture formats are missing in [`kLoadValueFromStorageInfo`](https://github.com/gpuweb/cts/blob/b62c3fc76c9c644a8597399b0d23f7032cabb366/src/webgpu/util/texture.ts#L19) causing mac-arm64-dawn-rel to fail. 

Moreover, I think the commit to remove ascii `rgba` was missing as it [is still there](https://github.com/gpuweb/cts/pull/4427/commits/5c185cc05d047820dc08bad8983696ebeafe3544#r2371028446) and causing this issue:
```
--> (in subcase: invalidSwizzle="rgba") EXPECTATION FAILED: unexpectedly did not throw
        at UniqueFeaturesOrLimitsGPUTest.shouldThrow (http://127.0.0.1:49489/third_party/webgpu-cts/src/common/framework/fixture.js:281:44)
        at RunCaseSpecific.fn (http://127.0.0.1:49489/third_party/webgpu-cts/src/webgpu/api/validation/capability_checks/features/texture_component_swizzle.spec.js:49:7)
```

There are other tests failing but I suspect this is not because of swizzle. Can you confirm?

beaufortfrancois:
> Updated. [run](https://dawn-review.googlesource.com/c/dawn/+/264354?tab=checks)

Thanks! There are still some tests failing that are not related to the CTS. Do you know what is happening?

I also noticed some errors like this one in https://dawn-review.googlesource.com/c/dawn/+/264354?filter=mac&checksRunsSelected=mac-arm64-dawn-rel&tab=checks:
```
--> EXCEPTION: Error: Unexpected validation error occurred: Invalid destination gpu texture format.
        at APIInjectError (../../third_party/dawn/src/dawn/native/Device.cpp:1923)
    
        at DeviceHolder.attemptEndTestScope (http://127.0.0.1:51919/third_party/webgpu-cts/src/webgpu/util/device_pool.js:332:19)
        at async DevicePool.release (http://127.0.0.1:51919/third_party/webgpu-cts/src/webgpu/util/device_pool.js:48:13)
        at async Promise.allSettled (index 0)
        at async AllFeaturesMaxLimitsGPUTestSubcaseBatchState.finalize (http://127.0.0.1:51919/third_party/webgpu-cts/src/webgpu/gpu_test.js:69:25)
        at async RunCaseSpecific.run (http://127.0.0.1:51919/third_party/webgpu-cts/src/common/internal/test_group.js:396:17)
        at async handleOnMessage (http://127.0.0.1:51919/third_party/webgpu-cts/src/common/runtime/helper/test_worker-worker.js:22:5)
Unexpected exception while running webgpu:web_platform,copyToTexture,ImageBitmap:from_ImageData:alpha="none";orientation="flipY";colorSpaceConversion="default";srcFlipYInCopy=false;dstFormat="r16snorm";dstPremultiplied=false
```

beaufortfrancois:
> That one is related to tier1/tier2 textures and the requirement that some of those new formats are supposed to work with `copyExternalImageToTexture` but Chrome/Dawn still rejects them.
> 
> That PR has been committed but not rolled into Chrome so testing shows the error. You can see the list of commits in the CL above

1. https://chromium-review.googlesource.com/c/chromium/src/+/6969496 seems to suggest this PR has been rolled into Chrome 2 weeks ago. I may misunderstand what is happening though 🤷 

2. Do you think we could have a "clean" run so that we can inspect failures specific to swizzle?

src/webgpu/api/validation/capability_checks/features/texture_component_swizzle.spec.ts
```diff
@@ -98,33 +97,30 @@ g.test('only_identity_swizzle')
...
   )
   .params(u =>
     u
-      .combine('usage', [
-        GPUConst.TextureUsage.COPY_SRC,
-        GPUConst.TextureUsage.COPY_DST,
-        GPUConst.TextureUsage.TEXTURE_BINDING,
-        GPUConst.TextureUsage.RENDER_ATTACHMENT,
-        GPUConst.TextureUsage.STORAGE_BINDING,
+      .combine('useCase', [
```
beaufortfrancois, line 109: I believe we should test depth stencil attachment as well. See https://dawn-review.googlesource.com/c/dawn/+/264854/6/src/dawn/tests/unittests/validation/TextureViewValidationTests.cpp

Issue #4426, opened 2025-07-17
Test the view used is the default view if GPUTexture is used instead of GPUTextureView


beaufortfrancois:
Despite https://github.com/gpuweb/cts/pull/4407#issuecomment-3047575316, we need to test that the texture view used is the default view when a GPUTexture is used instead of a GPUTextureView with https://github.com/gpuweb/gpuweb/pull/5228.

@greggman Do you know where I could plug myself in to test this properly for all of these cases?
- [ ] GPUBindingResource
- [ ] GPURenderPassColorAttachment view
- [ ] GPURenderPassColorAttachment resolveTarget
- [ ] GPURenderPassDepthStencilAttachment view

PR #4429, opened 2025-07-18
Test view is default if texture is used instead of view

beaufortfrancois:
Issue: https://github.com/gpuweb/cts/issues/4426

<hr>


Kangz:
Not really, do we have tests that check that the default descriptor used in `texture.createView()` is correctly the full descriptor (so with full mips / levels, etc)? This is what we should duplicate for end2end tests.

greggman:
We could also change the `textureNumLayers` and `textureNumLevels` tests to check this. They current have parameters `full` and `partial` for views but pass in all parameters. We could make them pass in nothing (so the default) for `full`.

PR #4445, opened 2025-08-22
Expectations in createBindGroup.spec.ts are invalid

mwyrzykowski:
The specification says:
https://www.w3.org/TR/webgpu/#adapter-capability-guarantees

  minUniformBufferOffsetAlignment and minStorageBufferOffsetAlignment must both be ≥ 32 bytes.

but the test expected 128 to be unaligned and hence invalid. 128 mod 32 = 0.

Issue: #

<hr>

### DIFF
--- a/src/webgpu/api/validation/createBindGroup.spec.ts
+++ b/src/webgpu/api/validation/createBindGroup.spec.ts
@@ -463,7 +463,7 @@ g.test('buffer_offset_and_size_for_bind_groups_match')
     // Unaligned buffer offset is invalid
     { offset: 1, size: 256, _success: false },
     { offset: 1, size: undefined, _success: false },
-    { offset: 128, size: 256, _success: false },
+    { offset: 127, size: 256, _success: false },
     { offset: 255, size: 256, _success: false },
 
     // Out-of-bounds


src/webgpu/api/validation/createBindGroup.spec.ts
```diff
@@ -463,7 +463,7 @@ g.test('buffer_offset_and_size_for_bind_groups_match')
     // Unaligned buffer offset is invalid
     { offset: 1, size: 256, _success: false },
     { offset: 1, size: undefined, _success: false },
-    { offset: 128, size: 256, _success: false },
```
mwyrzykowski, line 466: 128 passes in Chrome since Chrome reports 256 byte alignment which is the default guaranteed by the specification. 

But as noted in the commit message, an alignment as low as 32 is allowed. The alignment must be a power of two (see https://www.w3.org/TR/webgpu/#limit-class-alignment). So 128 is not a valid value to expect a validation error, as 128 is aligned to a 32 byte alignment

greggman: good catch! This test broke when I made the default use the best limits but didn't have a an impl that had better limits.

greggman (approved):

PR #4407, opened 2025-06-24
Allow GPUTexture where GPUTextureView is used

beaufortfrancois:
Spec PR: https://github.com/gpuweb/gpuweb/pull/5228

<hr>

**Requirements for PR author:**

- [x] All missing test coverage is tracked with "TODO" or `.unimplemented()`.
- [x] New helpers are `/** documented */` and new helper files are found in `helper_index.txt`.
- [x] Test behaves as expected in a WebGPU implementation. (If not passing, explain above.) They pass in a Chromium build patched with https://chromium-review.googlesource.com/c/chromium/src/+/6656722
- [x] Test have be tested with compatibility mode validation enabled and behave as expected. (If not passing, explain above.)

**Requirements for [reviewer sign-off](https://github.com/gpuweb/cts/blob/main/docs/reviews.md):**

- [ ] Tests are properly located in the test tree.
- [ ] [Test descriptions](https://github.com/gpuweb/cts/blob/main/docs/intro/plans.md) allow a reader to "read only the test plans and evaluate coverage completeness", and accurately reflect the test code.
- [ ] Tests provide complete coverage (including validation control cases). **Missing coverage MUST be covered by TODOs.**
- [ ] Helpers and types promote readability and maintainability.

When landing this PR, be sure to make any necessary issue status updates.

PR #4447, opened 2025-08-23
fix: properly apply `bindTextureResource` in `render_pass_descriptor:one_*_attachment` tests

ErichDonGubler:
This is a bug in #4407, where this coverage was originally intended to be introduced.


greggman (approved):
good catch!

PR #4452, opened 2025-09-05
Support git version from environment

bsheedy-work:
Adds support in version.ts for retrieving the version from the WEBGPU_CTS_GIT_VERSION environment variable instead of from calling "git describe" directly. This is so that the tests can be run in contexts where the actual git checkout is not available, such as in certain cases within Chromium.

Issue: #

<hr>

### DIFF
--- a/src/common/tools/version.ts
+++ b/src/common/tools/version.ts
@@ -1,5 +1,11 @@
 /* eslint-disable-next-line n/no-restricted-require */
-export const version = require('child_process')
-  .execSync('git describe --always --abbrev=0 --dirty')
-  .toString()
-  .trim();
+// Providing the version via an environment variable is supported so that these
+// tests can be run without the actual git checkout available.
+let possible_version = process.env.WEBGPU_CTS_GIT_VERSION;
+if (!possible_version) {
+  possible_version = require('child_process')
+    .execSync('git describe --always --abbrev=0 --dirty')
+    .toString()
+    .trim();
+}
+export const version = possible_version;


kainino0x (commented):
This is fine, though I think there are two other simpler possible solutions:

- Non-node builds [replace](https://github.com/gpuweb/cts/blob/42c0504c66193d231d31f1b64c1213c639281bc9/Gruntfile.js#L180-L193) this file with a [generated](https://github.com/gpuweb/cts/blob/main/tools/gen_version) file containing the hash.
  If we are building the project first (e.g. into `out-node`) then that would work (actually it would be more correct because it would tell you the commit hash of the built code).
  But IIRC you said we were not building the project first which means we're using something like `tools/run_node`? In which case we can't do that.
  IIRC you said we were not building the project first (into `out-node`) but using something like `tools/run_node`? If so then I don't think that would work 
- I don't think the version _actually_ needed though (it only happens if you try to serialize the test results as JSON) so we could try replacing `version` with `version()` so it's computed on demand.

kainino0x:
Ah that makes sense, I thought building the CTS turned out not to be necessary but that was just in the GN files, Go is still invoking the compilation. (It looks like `run-cts` does _not_ use `tools/run_node`, it just runs various other tools through ts-in-node.)

Let me see about just removing this from the dependency chain for `npx grunt node` since it's not used.

kainino0x:
OK, that's kind of a pain. What about this? #4455

PR #4459, opened 2025-09-11
Fix usage of bgra8unorm when not-write only in createBindGroupLayout tests

dj2:
In the `createBindGroupLayout` `storage_texture,formats` tests we attempt to use bgra8unorm as one of the testing formats, but that usage was not part of the spec. This CL changes the test to skip for bgra8unorm when the access is not write.

<hr>

### DIFF
--- a/src/webgpu/api/validation/createBindGroupLayout.spec.ts
+++ b/src/webgpu/api/validation/createBindGroupLayout.spec.ts
@@ -510,6 +510,10 @@ g.test('storage_texture,formats')
     const { format, access } = t.params;
     t.skipIfTextureFormatNotSupported(format);
 
+    if (format === 'bgra8unorm' && access !== 'write-only') {
+      t.skip(`bgra8unorm only supports write access mode`);
+    }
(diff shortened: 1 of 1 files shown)


greggman (commented):
As far as I can tell the test is valid as is

The only error locally is on Canary is

> VALIDATION FAILED: Validation succeeded unexpectedly.

For the case 

webgpu:api,validation,createBindGroupLayout:storage_texture,formats:format="bgra8unorm";access="read-only"

Which is correct. Chrome has only deprecated this. It's not yet generating a validation error like it is supposed to. [We marked it as Oct 1st to switch to a validation error](https://issues.chromium.org/issues/430635197)

Note: No test should ever check for specific formats as that leads to having to go through all the tests to find all the edge cases when something breaks.

The check that rejects `bgra8urnom` it line just below your PR

>    const success = isTextureFormatUsableWithStorageAccessMode(t.device, format, access);

Which works for all formats and adjusts based on the features on the device.

PR #4460, opened 2025-09-17
Update copyExternalImageToTexture test formats

greggman:
Update these tests so they can automatically handle new formats as they are enabled by features.

* Issue: [#5289](https://github.com/gpuweb/gpuweb/issues/5289)

<hr>

### DIFF
--- a/src/webgpu/format_info.ts
+++ b/src/webgpu/format_info.ts
@@ -1829,6 +1829,24 @@ export const kOptionalTextureFormats = kAllTextureFormats.filter(
   t => kTextureFormatInfo[t].feature !== undefined
 );
 
+/** Formats added from 'texture-formats-tier1' to be usable with `copyExternalImageToTexture`.
+ * DO NOT EXPORT. Use kPossibleValidTextureFormatsForCopyE2T and
+ * filter with `isTextureFormatUsableWithCopyExternalImageToTexture`
+ * or `GPUTest.skipIfTextureFormatNotUsableWithCopyExternalImageToTexture`
+ */
+const kValidTextureFormatsForCopyE2TTier1 = [
+  'r16unorm',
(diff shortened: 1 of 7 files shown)


src/webgpu/format_info.ts
```diff
@@ -1827,8 +1829,26 @@ export const kOptionalTextureFormats = kAllTextureFormats.filter(
...
+  'r16snorm',
+  'rg16unorm',
+  'rg16snorm',
+  'rgba16unorm',
+  'rgba16snorm',
+  'r8snorm',
+  'rg8snorm',
+  'rgba8snorm',
+  'rg11b10ufloat',
+] as const;
```
beaufortfrancois, line 1848: Nit: Why are those not `as GPUTextureFormat` instead?

beaufortfrancois (approved):
LGTM with nit

src/webgpu/format_info.ts
greggman: I'm not sure I understand your nit.

`as const` makes them assignable to `GPUTextureFormat` because the compiler can see that all entries are a valid `GPUTextureFormat`. That pattern seems common in the CTS. Are you suggesting something more, like it would be better to cast the array to `GPUTextureFormat[]`?

beaufortfrancois: Thanks for the explanation. I thought the compiler didn't know those were `GPUTextureFormat`. We're good then!

kainino0x: FYI we use `as const` because it's type-safe, whereas `as GPUTextureFormat` allows conversions even when we can statically tell they're wrong (e.g. `'not-a-real-format' as GPUTextureFormat` or `'x' as 'y'`)

Issue #4462, opened 2025-09-24
Several limits related tests are broken


greggman:
For example: [https://gpuweb.github.io/cts/standalone/?debug=1&runnow=1&q=webgpu:api,operation,adapter,requestDevice:*](https://gpuweb.github.io/cts/standalone/?debug=1&runnow=1&q=webgpu:api,operation,adapter,requestDevice:*)

PR #4463, opened 2025-09-24
Fix issues wrt to compat only limits

greggman:
A bunch of tests were failing

1. Because they require limits that don't exist in core

2. Because a few compat added limits were missing from capability_info.ts

There doens't seem to be a great way to support optional limits. Limits that are planned to be added but are not yet officially added and we want both cases. We want to test the limits exist but we also want current spec compliant implementations to pass.

For now, we only check these limits in compatibility mode. We may need to figure something else out if more limits are added in the future.

Fixes: #4462
Issue: #4462

https://dawn-review.googlesource.com/c/dawn/+/263414

<hr>


src/webgpu/capability_info.ts
```diff
@@ -727,7 +727,7 @@ assertTypeTrue<TypeEqual<GPUIndexFormat, (typeof kIndexFormat)[number]>>();
 const [kLimitInfoKeys, kLimitInfoDefaults, kLimitInfoData] =
   /* prettier-ignore */ [
                                                [    'class',    'core', 'compatibility',            'maximumValue'] as const,
-                                               [  'maximum',          ,                ,     kMaxUnsignedLongValue] as const, {
+                                               [  'maximum',          ,                ,     kMaxUnsignedLongValue, false] as const, {
```
kainino0x: does this do anything? seems like it's not associated with a key

kainino0x (approved):

src/webgpu/capability_info.ts
greggman: yea, sorry. removed. it was left over from a first attempt and I missed undoing it.

Issue #4476, opened 2025-10-15
wgsl uniformity: test uniformity for loop where body has behaviour {Return}


dneto0:
See spec fix PR https://github.com/gpuweb/gpuweb/pull/5392/

PR #4477, opened 2025-10-15
uniformity: test uniformity when a loop body only returns

dneto0:
Consider: loop { s1 continuing { s2 } }
    
    If the statement behaviour of s1 is exactly {Return}, then the statement
    behaviour of the entire loop is {Return}. In particular, nonuniformity
    effects do not leak out from s2 to the context after or outside of the
    whole loop.
    
    See the WGSL spec fix for https://github.com/gpuweb/gpuweb/issues/5100
    
    Test three new statement scenarios, each where s1 starts with `return;`,
    but varying where a collective operation is placed:
     - immediately after the return statement
     - in the continuing block of the same loop
     - immediately after the loop, where the continuing block in the loop
       has a break-if. This case requires the analysis to avoid using a
       Next behaviour from the overall loop.
     - immediately after the loop
    
    Also test similar variants where we use `break` instead of `return`.
    
    Also test loop constructs:
     - for with a condition: desugars to loop with initial `if !(cond) { break;}`
     - for without a condition: desugars to loop without that initial
       conditional break.
     - while loop

Fixed: #4476

Issue: #4476

<hr>

### DIFF
--- a/src/webgpu/shader/validation/uniformity/uniformity.spec.ts
+++ b/src/webgpu/shader/validation/uniformity/uniformity.spec.ts
@@ -185,60 +185,296 @@ function generateOp(op: string): string {
   }
 }
 
-const kStatementKinds = ['if', 'for', 'while', 'switch', 'break-if'] as const;
+const kStatementKinds = [
+  'if',
+  'for',
+  'while',
+  'switch',
+  'break-if',
+  'loop-always-break-op-inside',
+  'loop-always-return-op-inside',
+  'loop-always-break-op-continuing',
+  'loop-always-return-op-continuing',
+  'loop-always-break-op-after',
+  'loop-always-return-op-after',
+  'for-with-cond-always-break-op-inside',
+  'for-with-cond-always-return-op-inside',
+  'for-with-cond-always-break-op-after',
+  'for-with-cond-always-return-op-after',
+  'for-without-cond-always-break-op-inside',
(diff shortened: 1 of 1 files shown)


jrprice:
This case fails with Tint locally:
```
@compute @workgroup_size(16,1,1)
fn main(@builtin(local_invocation_index) lid: u32) {
	loop {
		return;
		continuing {
			break if true;
		}
	}
	if (lid > 0) {
		workgroupBarrier();
	}
}
```

Looks like the tests here do not cover the case where the is a `break if` in the continuing block. Is the corresponding spec PR supposed to cover that or is that a separate issue?

dneto0:
> Looks like the tests here do not cover the case where the is a break if in the continuing block. Is the corresponding spec PR supposed to cover that or is that a separate issue?

Good catch.  It should be covered by this rule.    I've added your case locally and it fails. So that's good.  I'll file a Tint issue to fix it.

dneto0:
I want to add more cases, as found from testing the Tint fixes.  

See the DeadLoopTests in https://dawn-review.googlesource.com/c/dawn/+/267474 that cover `loop`, `for-with-cond`, `for-without-cond`, and `while`.
It's tricky enough that it needs coverage.

dneto0:
I have a followup which greatly expands coverage of loops.  I'll land this now as it is still valid and does fix the #4476 issue

PR #4489, opened 2025-10-28
Increase the OOM allocation size to max precise value

### LINKED ISSUE #4488: OOM allocation in `src/webgpu/api/operation/buffers/map_oom.spec.ts` too small
Currently the CTS uses an allocation size of 128GB for tests where the allocation is supposed to fail with OOM.

If you have a substantially overprovisioned machine that can satisfy this allocation (e.g. 256GB RAM + 256GB swap) this causes significant issues, since the kernel will actually map the pages.

Which causes a significant slow down in the system as it does this and potentially causes crashes/panics as random things getting killed due to low memory if multiple subcases try to do this at once.

Also the case will fail unexpectedly, because the expectations are not met, since the allocation actually succeeds.

This value should increased to avoid this.

Unfortunately the theorical max amount of RAM for a 64-bit machine, 16 exabytes, is not precisely representable as a Number in TS, since it is a float64 under the hood. Number.MAX_SAFE_INTEGER is the equivalent of ~9000TB, so hopefully will be a sufficient value to avoid issues for the forseeable future. (There exist available niche server systems that theoretically support 64TB already, so just bumping up a couple of powers of two may not be sufficient).

zoddicus:
The current value causes issues with machines that have more than 128GB of RAM. This increase takes the allocation to the highest precisely representable value, which is ~8,192TB.

I am aware of no commercially available machine supports anything near this much RAM, so this should be good for the foreseeable future. If there is a device in the future with this much RAM, hopefully Number will have been re-specified to something like a float128 by then.

Fixes #4488

<hr>

### DIFF
--- a/src/webgpu/api/operation/buffers/map_oom.spec.ts
+++ b/src/webgpu/api/operation/buffers/map_oom.spec.ts
@@ -10,12 +10,7 @@ import { kMaxSafeMultipleOf8 } from '../../../util/math.js';
 const oomAndSizeParams = kUnitCaseParamsBuilder
   .combine('oom', [false, true])
   .expand('size', ({ oom }) => {
-    return oom
-      ? [
-          kMaxSafeMultipleOf8,
-          0x20_0000_0000, // 128 GB
-        ]
-      : [16];
+    return oom ? [kMaxSafeMultipleOf8] : [16];
   });
 
 export const g = makeTestGroup(AllFeaturesMaxLimitsGPUTest);


src/webgpu/api/operation/buffers/map_oom.spec.ts
```diff
@@ -13,7 +13,7 @@ const oomAndSizeParams = kUnitCaseParamsBuilder
     return oom
       ? [
           kMaxSafeMultipleOf8,
-          0x20_0000_0000, // 128 GB
+          Number.MAX_SAFE_INTEGER, // ~8,192 TB
```
jrprice: The `kMaxSafeMultipleOf8` constant used above is defined as `Number.MAX_SAFE_INTEGER - 7`, so do we really need this second value at all?

zoddicus: I don't think so. I should have check the value :upside_down_face:

zoddicus (commented):
PTAAL

src/webgpu/api/operation/buffers/map_oom.spec.ts
```diff
@@ -10,12 +10,7 @@ import { kMaxSafeMultipleOf8 } from '../../../util/math.js';
 const oomAndSizeParams = kUnitCaseParamsBuilder
   .combine('oom', [false, true])
   .expand('size', ({ oom }) => {
-    return oom
-      ? [
-          kMaxSafeMultipleOf8,
-          0x20_0000_0000, // 128 GB
```
kainino0x, line 16: Just FYI, I think the reason to test both `kMaxSafeMultipleOf8` and 128 GB was that we expected `kMaxSafeMultipleOf8` to hit some hardcoded limit and 128 GB to hit an actual OOM. Unfortunately we did expect that to be possibly unreliable, so it was one of those test cases I call "aspirational".

zoddicus: Do you think it would be useful to introduce a new version of the 'actual OOM' limit, maybe 4TB or some other value, with a note about it potentially needing to upped in the future, or is it likely we are already getting into the space where it will be hard to distinguish between a hardcoded limit and a true OOM?

----

Did a brief investigation on what current operation system RAM limits look like, and Windows 11 supports between 128GB and 6TB of RAM depending on the license/version. And I believe the current memory page scheme of the Linux kernel has a theoretical ceiling of 4 PB of RAM.

So we are kinda in a world already where I can build a workstation with more that the maximum amount of RAM for a 'common' OS. So I am not sure how much having a seperate test value would give us, and I really don't like the idea of trying to do some sort of OS inspection to figure out an appropriate value.

kainino0x: We should probably just leave it up to individual implementations to test this if they want to, since there's no way to know the hardcoded OOM threshold. Also that way the test could have some backdoor to see whether an OOM is real (came from the underlying API) or hardcoded.

PR #4494, opened 2025-11-07
Add a test for many timestamp query sets

greggman:
This test is because Metal has a limit of 32 timestamp query sets. Implementations are supposed to workaround this limit by allocating larger metal sets and having WebGPU set be subsets in those larger sets.

This is especially important as the limit is 32 per process so a few pages making a few queries would easily hit the limit and prevent pages from running.

In issue [5261](https://github.com/gpuweb/gpuweb/issues/5261) 65k queries was mentioned so this test creates 65536 (not sure if that's the same 65k Kai was referring to)

I think that maybe the issue should be reopened and the spec should state the minimum or it should defer it to the CTS. Except for a few explicit OOM tests, we don't allow implementations to return `out-of-memory` for all buffers and textures and claim to pass the CTS and I think we should do the same here and require implementations to pass this test too. Is 65536 good for this?

<hr>

### DIFF
--- a/src/webgpu/api/operation/command_buffer/queries/timestampQuery.spec.ts
+++ b/src/webgpu/api/operation/command_buffer/queries/timestampQuery.spec.ts
@@ -1,0 +1,154 @@
+export const description = `
+API operations tests for timestamp queries.
+
+Given the values returned are implementation defined
+there is not much we can test except that there are no errors.
+
+- test query with
+  - compute pass
+  - render pass
+  - 64k query objects
+`;
+
+import { makeTestGroup } from '../../../../../common/framework/test_group.js';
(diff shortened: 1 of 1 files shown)


kainino0x:
> In issue [5261](https://github.com/gpuweb/gpuweb/issues/5261) 65k queries was mentioned so this test creates 65536 (not sure if that's the same 65k Kai was referring to)

That was based on Mike's comment:
> With a global limit of 32 sample buffers, 65,536 timestamp queries should be able to be in flight before we run into this limitation.

It may be good to (also) test more, but knowing that some implementations may choose not to deal with that because it's probably vanishingly unlikely to happen in practice.

kainino0x (approved):

PR #4501, opened 2025-11-12
Detect invalid end() in "pass_end_invalid_order"

Kangz:
FYI @andyleiserson

Calling end() on an already ended or invalid pass should generate a device error. Detect these cases in pass_end_invalid_order and expect a validation error in these cases.

Passing locally after many other fixes for encoder error reporting in Dawn.

Issue: #

<hr>

### DIFF
--- a/src/webgpu/api/operation/command_buffer/queries/timestampQuery.spec.ts
+++ b/src/webgpu/api/operation/command_buffer/queries/timestampQuery.spec.ts
@@ -1,0 +1,154 @@
+export const description = `
+API operations tests for timestamp queries.
+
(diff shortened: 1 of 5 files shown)


kainino0x (approved):

andyleiserson:
I'm still having trouble reconciling the expected behavior with the spec.

> Calling end() on an already ended or invalid pass should generate a device error

The spec says "_this_.[[state]] must be _open_", which does imply the pass has not ended, but is not the same thing as the pass being valid. But I don't think that requiring that it be valid is what we want either, because invalidity can be due to other types of errors that are reported through `finish`.

PR #4510, opened 2025-11-25
Minor change in sample mask behavior for the case of sample shading

petermcneeleychromium:
This is to make us aligned with existing spec and vulkan behavior.
Webgpu issue
https://github.com/gpuweb/gpuweb/issues/5457


alan-baker (approved):

mwyrzykowski (changes requested):
I think this is incorrect. The current definition of sample_mask:
> Sample coverage mask for the current fragment. It contains a bitmask indicating which samples in this fragment are covered by the primitive being rendered.

There is nothing in the definition indicating `sample_mask` is per-sample, one would it expect it to be constant per-fragment.

kainino0x:
Isn't one fragment = the output of one invocation of the fragment shader? So may be a single sample or not, depending on which mode the fragment shader is running in.

mwyrzykowski:
> Isn't one fragment = the output of one invocation of the fragment shader? So may be a single sample or not, depending on which mode the fragment shader is running in.

We discussed briefly in the WGSL meeting, there is certainly confusion regarding this.

In Metal, one fragment corresponds to a single texel in the render target. I.e., there may be up to [sample-count] fragment function invocations per fragment. This also aligns with the definition [here](https://en.wikipedia.org/wiki/Fragment_(computer_graphics)), not say that's a definitive source either.
> In [computer graphics](https://en.wikipedia.org/wiki/Computer_graphics), a fragment is the data necessary to generate a single [pixel](https://en.wikipedia.org/wiki/Pixel)'s worth of a drawing primitive in the [frame buffer](https://en.wikipedia.org/wiki/Frame_buffer).

petermcneeleychromium:
My [reading of opengl ](https://wikis.khronos.org/opengl/Fragment#:~:text=A%20Fragment%20is%20a%20collection,next%20stage%20of%20the%20pipeline.)   is that there are multiple fragments per pixel.  

However clearly, in the case we were discussing (sample shading) it is as you say multiple invocations **per** fragment
>If [sample shading](https://vkdoc.net/chapters/primsrast#primsrast-sampleshading) is enabled and multiple invocations per fragment are required, additional invocations must be executed as specified.

from [vkdoc](https://vkdoc.net/chapters/fragops#fragops-samplemask#:~:text=if%20sample%20shading)

However we are still stuck in vulkan with 
> If sample shading is enabled, fragment shaders will only see values of 1 for samples being shaded - other bits will be 0.

Which is why I think we get the vulkan behavior we see. (this is possibly motivated due to writing to sample mask as well)

kainino0x:
> My [reading of opengl ](https://wikis.khronos.org/opengl/Fragment#:~:text=A%20Fragment%20is%20a%20collection,next%20stage%20of%20the%20pipeline.) is that there are multiple fragments per pixel.
> 
> > but rasterization can produce multiple fragments from the same triangle per-pixel, depending on various multisampling parameters and OpenGL state. There will be at least one fragment produced for every pixel area covered by the primitive being rasterized.

Oops forgot to say - I agree this text from the OpenGL wiki does read the way you interpreted it, but I think that's inconsistent with the (latest) OpenGL specs.

petermcneeleychromium:
Alan send me this
https://github.com/KhronosGroup/SPIRV-Registry/pull/398

Which simply regurgitates existing behavior for vulkan. 

```
The existing input decoration *SampleMask* is intended to provide the set of samples that contribute
to a fragment. However, when a fragment shader executes in a per-sample shading mode (e.g. when
a fragment shader input is decorated with *SampleId* or *SamplePosition*), the *SampleMask* built-in
is masked to only include the bits corresponding to the specific sample(s) currently being processed
by that invocation. In the case of full per-sample shading (one invocation per sample), this results
in only a single bit being set. When a sample shading rate is specified (e.g. via Vulkan's
*minSampleShading* parameter), an invocation may process a subset of samples, and *SampleMask* will
contain the corresponding subset of bits.
```

As discussed we simply need to polyfill this behavior on other platforms (D3D and metal)

mwyrzykowski:
> As discussed we simply need to polyfill this behavior on other platforms (D3D and metal)

Don't we still have the concern of `sample_mask` having different behavior based on whether or not sample shading is used? I suppose its better than divergent behavior across backends, but it seems a bit confusing that `sample_mask` would give different results if a developer adds a variable backed by `sample_index`

petermcneeleychromium:
Sorry for the confusion. This is the pr for cts.
The issue is here https://github.com/gpuweb/gpuweb/issues/5457 and probably where the discussion should be. I did see your comment above and I will try to address it.

mwyrzykowski (commented):
Instead of using the wording:
"When Sample Shading ..."

I think we should define this on:
"When sample_index is used", perhaps?

Sample shading is not guaranteed per the MSL spec, but we could change the behavior of sample_mask manually when sample_index is emitted in the shader

petermcneeleychromium:
**Sample Shading** is of course Vulkan speak for Metal's **per-sample** execution.  

There are other ways to potentially trigger Sample Shading one of which is using the 'sample' interpolation attribute. 
This applies to metal (MSL) as well. So just saying "When sample_index is used" I suspect is not enough or unclear even in the limited context of WGSL.

From MSL 5.5
```
The sampling attribute identifies if
fragment input interpolation is per-sample or per-fragment. Similarly, the [[sample_id]]
attribute identifies the current sample index
```

Issue #4297, opened 2025-03-17
Remove Float16Array polyfill


kainino0x:
According to caniuse, `Float16Array` has shipped in Firefox and Safari, and is shipping in [Chrome 134](https://chromiumdash.appspot.com/schedule): https://caniuse.com/mdn-javascript_builtins_float16array

We should try removing our `Float16Array` polyfill and see if it works!

kainino0x:
Corrected chrome version above.

We could do this now, but it's convenient for the CTS to run on stable. We should probably wait until a bit after Apr 15 when it starts shipping on ChromeOS.

kainino0x:
When we do this we also need to make sure we remove any code that was skipping `Float16Array` because it wasn't a builtin. (There are at least examples of this in writeBuffer and setImmediates.)

PR #4515, opened 2025-12-03
Validate setImmediates API in different encoder types

shaoboyan091:
This PR adding validation tests to cover setImmediates API in  different encoder types (compute pass, render pass, render bundle) by covering: 
* Interpretation:
  - Passing a TypedArray, the data offset and size are given in elements (not bytes).
* Alignment:
  - rangeOffset is not a multiple of 4 bytes.
  - content size, converted to bytes, is not a multiple of 4 bytes.
* Arithmetic overflow
  - rangeOffset + contentSize is overflow
* Bounds:
  - dataOffset + contentSize (in bytes) exceeds the content data size.
  - rangeOffset + contentSize (in bytes) exceeds the maxImmediateSize.

Issue: #

<hr>


src/webgpu/api/validation/encoding/cmds/setImmediates.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+  - Passing a TypedArray, the data offset and size are given in elements (not bytes).
+* Alignment:
+  - rangeOffset is not a multiple of 4 bytes.
+  - content size, converted to bytes, is not a multiple of 4 bytes.
+* Arithmetic overflow
+  - rangeOffset + contentSize is overflow
+* Bounds:
+  - dataOffset + contentSize (in bytes) exceeds the content data size.
+  - rangeOffset + contentSize (in bytes) exceeds the maxImmediateSize.
+`;
```
kainino0x, line 4: Each of these seems to describe a specific `g.test`, so please associate the descriptions with the tests instead of putting them at the top of the file

src/webgpu/api/validation/encoding/cmds/setImmediates.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+      const data = new arrayBufferType(maxImmediateSize);
+
+      encoder.setImmediates(/* rangeOffset */ 0, data, /* dataOffset */ validOffset, validDataSize);
+
+      validateFinish(success || elementSize === 1);
+    }
+
+    for (const arrayType of kTypedArrayBufferViewConstructors) {
+      if (arrayType === Float16Array) {
+        // Skip Float16Array since it is supplied by an external module, so there isn't an overload for it.
```
kainino0x: Just a note: Float16Array is now supported by browsers natively, and we need to remove the external module, and remove any skips like this. Noted in #4297

src/webgpu/api/validation/encoding/cmds/setImmediates.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+
+      const { encoder, validateFinish } = t.createEncoder(encoderType);
+      const data = new arrayBufferType(maxImmediateSize);
+
+      encoder.setImmediates(/* rangeOffset */ 0, data, /* dataOffset */ validOffset, validDataSize);
+
+      validateFinish(success || elementSize === 1);
+    }
+
+    for (const arrayType of kTypedArrayBufferViewConstructors) {
```
kainino0x: Normally we try to use subcases instead of for-loops. I see writeBuffer already has a test like this, but it also has another test that uses subcases. Not sure why that's true, but if it's not too much effort, could you refactor this to use subcases?

src/webgpu/api/validation/encoding/cmds/setImmediates.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+    function runTest(arrayBufferType: TypedArrayBufferViewConstructor) {
+      const kMinAlignmentBytes = 4;
+      const elementSize = arrayBufferType.BYTES_PER_ELEMENT;
+      const maxImmediateSize = t.device.limits.maxImmediateSize; // using this limit for tests
+      const validDataSize = success
+        ? Math.floor((maxImmediateSize - kMinAlignmentBytes) / elementSize)
+        : maxImmediateSize - kMinAlignmentBytes;
+      const validOffset = success
+        ? Math.ceil(kMinAlignmentBytes / elementSize)
+        : kMinAlignmentBytes;
```
kainino0x: I think there are too many combinations of things that can go wrong here, so it's hard to be sure that any combination of implementation bugs would be caught by this test. Also, this only tests validation, and won't verify that an implementation doesn't interpret the sizes wrong when actually doing the operation.

It seems to me that this should be an operation test, then we can look at the actual values written to memory which makes the test much more sensitive to possible implementation bugs.

src/webgpu/api/validation/encoding/cmds/setImmediates.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+        { offset: 4, contentByteSize: 4, _offsetValid: true, _contentValid: true },
+        // offset is not aligned to 4 bytes
+        { offset: 1, contentByteSize: 4, _offsetValid: false, _contentValid: true },
+        // contentSize is not aligned to 4 bytes
+        { offset: 4, contentByteSize: 5, _offsetValid: true, _contentValid: false },
+      ] as const)
+  )
+  .fn(t => {
+    const { encoderType, offset, contentByteSize, _offsetValid, _contentValid } = t.params;
+    const data = new Uint8Array(contentByteSize);
```
kainino0x: This should be tested for other array types too.

src/webgpu/api/validation/encoding/cmds/setImmediates.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+
+    const doSetImmediates = () => {
+      encoder.setImmediates(offset, data, 0, contentByteSize);
+    };
+
+    if (_contentValid) {
+      doSetImmediates();
+    } else {
+      t.shouldThrow('RangeError', doSetImmediates);
+    }
```
kainino0x:
```suggestion
    t.shouldThrow(_contentValid ? false : 'RangeError', () => {
      encoder.setImmediates(offset, data, 0, contentByteSize);
    });
```

src/webgpu/api/validation/encoding/cmds/setImmediates.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+
+    validateFinish(_offsetValid);
+  });
+
+g.test('overflow')
+  .desc('Tests that rangeOffset + contentSize exceeds Number.MAX_SAFE_INTEGER.')
```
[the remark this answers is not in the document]
kainino0x:
```suggestion
  .desc('Tests that rangeOffset + contentSize is handled correctly if it exceeds Number.MAX_SAFE_INTEGER.')
```

src/webgpu/api/validation/encoding/cmds/setImmediates.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+
+    const doSetImmediates = () => {
+      encoder.setImmediates(rangeOffset, data, dataOffset, size);
+    };
+
+    if (_contentValid) {
+      doSetImmediates();
+    } else {
+      t.shouldThrow('RangeError', doSetImmediates);
+    }
```
kainino0x: same

src/webgpu/api/validation/encoding/cmds/setImmediates.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+  .paramsSubcasesOnly(u =>
+    u //
+      .combine('encoderType', kProgrammableEncoderTypes)
+      .combineWithParams([
+        // control case
+        { offset: 4, contentByteSize: 4, _rangeValid: true, _contentValid: true },
+        // rangeOffset + contentSize overflows
+        {
+          offset: 8,
+          contentByteSize: kMaxSafeMultipleOf8,
```
kainino0x: I think this should test `2**31 - 8` and `2**32 - 8`. That's closer to "reasonable" so it's more likely to find bugs before being caught by some hardcoded OOM limit.

I can't think of a bug that would be caught by `2**53 - 8` but not by `2**32 - 8`. And of course we can't test `2**64 - 8` since it's a `double`.

src/webgpu/api/validation/encoding/cmds/setImmediates.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+
+    const doSetImmediates = () => {
+      encoder.setImmediates(offset, data, 0, contentByteSize);
+    };
+
+    if (_contentValid) {
+      doSetImmediates();
+    } else {
+      t.shouldThrow('RangeError', doSetImmediates);
+    }
```
kainino0x, line 142: same

src/webgpu/api/validation/encoding/cmds/setImmediates.spec.ts
kainino0x: same

src/webgpu/api/validation/encoding/cmds/setImmediates.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+      ] as const)
+  )
+  .fn(t => {
+    const { encoderType, offset, contentByteSize, _rangeValid, _contentValid } = t.params;
+    const data = new Uint8Array(t.device.limits.maxImmediateSize);
+
+    const { encoder, validateFinish } = t.createEncoder(encoderType);
+
+    const doSetImmediates = () => {
+      encoder.setImmediates(offset, data, 0, contentByteSize);
```
kainino0x: What about the case where `dataOffset + size` could overflow? `contentByteSize` is going to be too large for `data` regardless, so this would probably fail in Blink before Dawn has looked at `offset` at all. Perhaps that's what you intended to test here?

Also, `out_of_range` looks more general - maybe this can be folded into the cases of that test and just explained in a comment?

src/webgpu/api/validation/encoding/cmds/setImmediates.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+  });
+
+g.test('overflow')
+  .desc('Tests that rangeOffset + contentSize exceeds Number.MAX_SAFE_INTEGER.')
+  .paramsSubcasesOnly(u =>
+    u //
+      .combine('encoderType', kProgrammableEncoderTypes)
+      .combineWithParams([
+        // control case
+        { offset: 4, contentByteSize: 4, _rangeValid: true, _contentValid: true },
```
kainino0x: Also need a control case where contentByteSize is 0, because that's what would happen if there were an overflow.

src/webgpu/api/validation/encoding/cmds/setImmediates.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+          offset: 8,
+          contentByteSize: kMaxSafeMultipleOf8,
+          _rangeValid: true,
+          _contentValid: false,
+        },
+      ] as const)
+  )
+  .fn(t => {
+    const { encoderType, offset, contentByteSize, _rangeValid, _contentValid } = t.params;
+    const data = new Uint8Array(t.device.limits.maxImmediateSize);
```
kainino0x: This should be tested for other array types too.

src/webgpu/api/validation/encoding/cmds/setImmediates.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+      rangeRemainSpace,
+      dataByteSize,
+      dataOffset,
+      size,
+      _rangeValid,
+      _contentValid,
+    } = t.params;
+    const maxImmediates = t.device.limits.maxImmediateSize;
+    const rangeOffset = maxImmediates - rangeRemainSpace;
+    const data = new Uint8Array(dataByteSize);
```
kainino0x: This should be tested for other array types too.

src/webgpu/api/validation/encoding/cmds/setImmediates.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+        ? Math.floor((maxImmediateSize - kMinAlignmentBytes) / elementSize)
+        : maxImmediateSize - kMinAlignmentBytes;
+      const validOffset = success
+        ? Math.ceil(kMinAlignmentBytes / elementSize)
+        : kMinAlignmentBytes;
+
+      const { encoder, validateFinish } = t.createEncoder(encoderType);
+      const data = new arrayBufferType(maxImmediateSize);
+
+      encoder.setImmediates(/* rangeOffset */ 0, data, /* dataOffset */ validOffset, validDataSize);
```
kainino0x: Another thing that needs to be tested is that `rangeOffset` is in bytes, NOT in elements.

src/webgpu/api/validation/encoding/cmds/setImmediates.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+      .combine('encoderType', kProgrammableEncoderTypes)
+      .combineWithParams([
+        // control case
+        {
+          rangeRemainSpace: 4,
+          dataByteSize: 32,
+          dataOffset: 0,
+          size: 4,
+          _rangeValid: true,
+          _contentValid: true,
```
kainino0x: nit: `_*Valid` should be easily computable inside the test, that makes it easier to add more cases, and easier to review the correctness.

src/webgpu/api/validation/encoding/cmds/setImmediates.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+  .desc('Tests that rangeOffset + contentSize exceeds Number.MAX_SAFE_INTEGER.')
+  .paramsSubcasesOnly(u =>
+    u //
+      .combine('encoderType', kProgrammableEncoderTypes)
+      .combineWithParams([
+        // control case
+        { offset: 4, contentByteSize: 4, _rangeValid: true, _contentValid: true },
+        // rangeOffset + contentSize overflows
+        {
+          offset: 8,
```
kainino0x: nit: for clarity, stick to the naming in the spec, so call this `rangeOffset`.

src/webgpu/api/validation/encoding/cmds/setImmediates.spec.ts
shaoboyan091: Yes. Dawn doesn't see any raw offset and size, but just a handled span bytes instead. So I expect this could fail in blink side.

src/webgpu/api/validation/encoding/cmds/setImmediates.spec.ts
shaoboyan091: I hit an error when using Float16Array . It says " Type 'Float16Array' is not assignable to type 'GPUAllowSharedBufferSource'." So I need to use "as any" to bypass this check. But running it in canary reports:
"TypeError: Failed to execute 'setImmediates' on 'GPUComputePassEncoder': Overload resolution failed."

So I think this might be a real type error.

kainino0x: Skipping it is correct for now, because we are still using the Float16Array polyfill. _When_ we remove the polyfill we'll need to remove this skip. But you don't need to do that now.

src/webgpu/api/validation/encoding/cmds/setImmediates.spec.ts
kainino0x: Right, `dataOffset` gets processed by Blink. But `rangeOffset` gets processed by Dawn, which won't happen at all because Blink fails first, so I'm just a bit confused what the intent of the test is - why pass a `rangeOffset` in this test? Is there a possible bug it's intended to catch?

src/webgpu/api/validation/encoding/cmds/setImmediates.spec.ts
shaoboyan091: Agree. I'd like to remove this case and use operation test to cover it.

src/webgpu/api/validation/encoding/cmds/setImmediates.spec.ts
shaoboyan091: Yes. I tend to create a scenario that `rangeOffset + size` is overflow in dawn side to test the check about "rangeOffset + size <= maxImmediateSize".

src/webgpu/api/validation/encoding/cmds/setImmediates.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+class SetImmediatesTest extends AllFeaturesMaxLimitsGPUTest {
+  override async init() {
+    await super.init();
+    if (
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      !('setImmediates' in (GPURenderPassEncoder.prototype as any)) ||
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      !('setImmediates' in (GPUComputePassEncoder.prototype as any)) ||
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      !('setImmediates' in (GPURenderBundleEncoder.prototype as any))
```
kainino0x: Will fully review tomorrow but one comment.

This will run the tests only if all three are available. I would like to run all of the tests if ANY of these, OR the WGSL feature OR the new limit, are available.

Also these checks need to be packaged up into a "skipIf" helper at some point, so they can also be used in other tests that don't use `SetImmediatesTest`.

shaoboyan091: >This will run the tests only if all three are available. I would like to run all of the tests if ANY of these, OR the WGSL feature OR the new limit, are available.

A bit concern about this. I think it faill the scenario that implementation is in middle. For example, when implementation added wgslLanguageFeature but not do API implementation, the case failed.

So my thoughts is that the test should check what it tends to check. For example, for this test case, we want to check all encoder API validation, then we need to check the API is available.(Maybe we can do more detail checking for each encoder). And for other case which test immediate state, it requires both API and wgslFeature.

WDYT?

kainino0x: I think it's preferable if tests fail on a partial implementation. Otherwise some applications might break because they feature detect one thing, and then try to use all the other things.

Of course if the whole feature is behind a browser flag, this is not a big deal. But test failures also shouldn't be a big deal, because they will be handled by browser test expectation files.

src/webgpu/api/validation/encoding/cmds/setImmediates.spec.ts
```diff
@@ -1,209 +1,194 @@
...
-        { offset: 4, contentByteSize: 4, _offsetValid: true, _contentValid: true },
-        // offset is not aligned to 4 bytes
-        { offset: 1, contentByteSize: 4, _offsetValid: false, _contentValid: true },
-        // contentSize is not aligned to 4 bytes
-        { offset: 4, contentByteSize: 5, _offsetValid: true, _contentValid: false },
-      ] as const)
+        // control case: rangeOffset 4 is aligned. contentByteSize 8 is aligned.
+        { rangeOffset: 4, contentByteSize: 8 },
+        // rangeOffset 5 is unaligned (5 % 4 !== 0).
+        { rangeOffset: 5, contentByteSize: 8 },
```
kainino0x: nit: Use 6 since it's "more aligned" than 5 but still invalid.

src/webgpu/api/validation/encoding/cmds/setImmediates.spec.ts
```diff
@@ -1,209 +1,194 @@
...
         // control case
-        { offset: 4, contentByteSize: 4, _rangeValid: true, _contentValid: true },
+        { rangeOffset: 0, dataOffset: 0, elementCount: 4, _expectedError: null },
         // rangeOffset + contentSize overflows
         {
-          offset: 8,
-          contentByteSize: kMaxSafeMultipleOf8,
-          _rangeValid: true,
-          _contentValid: false,
+          rangeOffset: Math.pow(2, 31) - 8,
```
kainino0x: µNit: can use `2 ** 31`

src/webgpu/api/validation/encoding/cmds/setImmediates.spec.ts
```diff
@@ -1,209 +1,194 @@
 export const description = `
 setImmediates validation tests.
```
kainino0x, line 2: Add `TODO(#4297): enable Float16Array` to this description since the test is not fully "complete" until we do that.

(Or add it to each test individually. But I think one TODO at the top is fine.)

src/webgpu/api/validation/encoding/cmds/setImmediates.spec.ts
```diff
@@ -1,209 +1,194 @@
...
-    } = t.params;
-    const maxImmediates = t.device.limits.maxImmediateSize;
-    const rangeOffset = maxImmediates - rangeRemainSpace;
-    const data = new Uint8Array(dataByteSize);
+    const { encoderType, arrayType, rangeOffsetDelta, dataLengthDelta } = t.params;
+    const arrayBufferType = kTypedArrayBufferViews[arrayType];
+    const elementSize = arrayBufferType.BYTES_PER_ELEMENT;
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const maxImmediateSize = (t.device.limits as any).maxImmediateSize;
```
kainino0x, line 173: nit: MAINTENANCE_TODO to remove this cast when the types are updated.

kainino0x (approved):

src/webgpu/api/validation/encoding/cmds/setImmediates.spec.ts
shaoboyan091: Good point about the application part. I'll address this.

kainino0x (approved):

Issue #4509, opened 2025-11-25
Transient Attachments


beaufortfrancois:
This issue tracks CTS tests for [Transient Attachments](https://github.com/gpuweb/gpuweb/pull/5450):
- [x] Validation tests that check the texture creation usage rules and the constraint on load/store ops.
- [x] Operation tests to check implementations don't crash

PR #4518, opened 2025-12-04
Fix transient attachment usage in validation tests

beaufortfrancois:
This PR fixes issues related to transient attachment usage that were discovered in https://dawn-review.googlesource.com/c/dawn/+/278375 after https://github.com/gpuweb/cts/pull/4508 was merged

Issue: https://github.com/gpuweb/cts/pull/4509

<hr>

### DIFF
--- a/src/webgpu/capability_info.ts
+++ b/src/webgpu/capability_info.ts
@@ -776,6 +776,7 @@ const [kLimitInfoKeys, kLimitInfoDefaults, kLimitInfoData] =
   'maxComputeWorkgroupSizeY':                  [           ,       256,             128,                          ],
   'maxComputeWorkgroupSizeZ':                  [           ,        64,              64,                          ],
   'maxComputeWorkgroupsPerDimension':          [           ,     65535,           65535,                          ],
+  'maxImmediateSize':                          [           ,        64,              64,                          ],
 } as const];
 
 // MAINTENANCE_TODO: Remove when the compat spec is merged.
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,7 +24,7 @@
         "@types/w3c-image-capture": "^1.0.10",
         "@typescript-eslint/eslint-plugin": "^6.9.1",
         "@typescript-eslint/parser": "^6.9.1",
-        "@webgpu/types": "^0.1.66",
+        "@webgpu/types": "^0.1.67",
         "ansi-colors": "4.1.3",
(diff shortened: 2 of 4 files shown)


src/webgpu/capability_info.ts
```diff
@@ -215,6 +215,17 @@ export const kTextureUsageInfo: {
 /** List of all GPUTextureUsage values. */
 export const kTextureUsages = numericKeysOf<GPUTextureUsageFlags>(kTextureUsageInfo);
 
+/** Check if `usage0` is TRANSIENT_ATTACHMENT and `usage1` is not RENDER_ATTACHMENT. */
```
greggman: This seems like a very specific situation of 2 usages. How about 3? Or 1?

wdyt about just one usage?

```js
export function IsValidTransientAttachmentUsage(
  usage: GPUTextureUsageFlags,
): boolean {
  return (
    usage === GPUConst.TextureUsage.TRANSIENT_ATTACHMENT |
              GPUConst.TextureUsage.RENDER_ATTACHMENT
  );
}
```

And then call it with 2 usages `isValidTransientUsageAttachment(usage0 | usage1)`

Also, wdyt about making it a positive assertion?

beaufortfrancois: How would you use `IsValidTransientAttachmentUsage(usage0 | usage1)` in https://github.com/gpuweb/cts/pull/4518/files#diff-e0bc33c614ca9f7cdbbf7fdec58243d4af5681749e8d56739c95e970e521d15fR1019 for instance?

beaufortfrancois: My understanding is that we would have to do something like that... 
```
        const usage = usage0 | usage1;
        
        // If the usage does not contain TRANSIENT, we don't need to filter it here.
        if ((usage & GPUConst.TextureUsage.TRANSIENT_ATTACHMENT) === 0) {
          return false;
        }

        // If it *does* contain TRANSIENT, exclude it unless it matches the valid combination.
        return !IsValidTransientAttachmentUsage(usage);
```

In that case, should I bundle the `usage & GPUConst.TextureUsage.TRANSIENT_ATTACHMENT` part inside `IsValidTransientAttachmentUsage` as well?

beaufortfrancois: I've updated this PR without bundling. Let me know if that looks good to you.

greggman (approved):
lgtm

PR #4533, opened 2025-12-16
Require depthOrArrayLayers=1 and mipLevelCount=1 for transient textures

beaufortfrancois:
As raised in https://github.com/gpuweb/gpuweb/pull/5450#discussion_r2621731509, this PR updates validation tests to require `depthOrArrayLayers=1` and `mipLevelCount=1` when creating transient textures.

Issue: https://github.com/gpuweb/cts/pull/4509

<hr>


src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -1023,17 +1025,23 @@ g.test('texture_usage')
       .unless(({ usage0, usage1 }) => {
         return !IsValidTextureUsageCombination(usage0 | usage1);
       })
+      .combine('depthOrArrayLayers', [1, 2])
+      .unless(({ dimension, depthOrArrayLayers }) => {
+        return dimension === '1d' && depthOrArrayLayers !== 1;
+      })
+      .combine('mipLevelCount', [1, 2])
```
kainino0x: This makes the whole test roughly 4x as long. Technically you could add these extra cases only when usage has TRANSIENT_ATTACHMENT, but that's hard to do.

I think it would be better to add a separate test that just these validation rules. They shouldn't interact with most of the parameters so I don't think they need to be tested in combination.

(note to self, definitely related to why we have too many tests. Both doing conditional expansions and adding new tests is more annoying than just adding combinatorics.)

beaufortfrancois: I've added a separate test as requested.

mwyrzykowski (approved):

src/webgpu/api/validation/createTexture.spec.ts
```diff
@@ -1067,15 +1061,51 @@ g.test('texture_usage')
...
     t.expectValidationError(() => {
       t.createTextureTracked(descriptor);
     }, !success);
   });
 
+g.test('depthOrArrayLayers_and_mipLevelCount_for_transient_attachments')
+  .desc(`Test depthOrArrayLayers and mipLevelCount must be 1 for transient attachments`)
+  .params(u =>
+    u
+      .combine('format', kAllTextureFormats)
```
kainino0x: nit: I would probably not even test all formats. Maybe just rgba8unorm and depth24plus.

(Most tests test all formats - I started that trend when I wrote early test plans. But it takes a lot of time and I think we should try to be more intentional about when we want to do it - lots of operation tests need it, but most validation tests probably don't.)

kainino0x (approved):

src/webgpu/api/validation/createTexture.spec.ts
beaufortfrancois: I've updated test to just check rgba8unorm and depth24plus formats as suggested.

beaufortfrancois: If that looks good to you, please merge.

kainino0x: LGTM, if anyone has an issue with that we can always expand it afterward

PR #4534, opened 2025-12-17
Bump @webgpu/types to 0.1.68

greggman:
Note: There's a bunch of issues here that need to be fixed in other PRs

1. Should `maxImmediateSize` and `setImmediates` have been added to `@webgpu/types`? Add them there makes things here fail. If they should be removed we need a PR upstream

2. src/webgpu/api/validation/encoding/encoder_open_state.spec.ts is broken on all implementations. https://github.com/gpuweb/gpuweb/issues/5207

### DIFF
--- a/src/webgpu/api/validation/capability_checks/limits/limit_utils.ts
+++ b/src/webgpu/api/validation/capability_checks/limits/limit_utils.ts
@@ -404,9 +404,11 @@ export class LimitTestsImpl extends GPUTestBase {
     this._adapter = await gpu.requestAdapter();
     const limit = this.limit;
     // MAINTENANCE_TODO: consider removing this skip if the spec has no optional limits.
+    // Note: The cast below is required because an optional limit has no entry
+    // in capability_info.ts kLimitInfoKeys, kLimitInfoDefaults, kLimitInfoData
     this.skipIf(
       (this._adapter?.limits[limit] === undefined && !!this.limitTestParams.limitOptional) ||
-        getDefaultLimitsForCTS()[limit] === undefined,
+        !(limit in getDefaultLimitsForCTS()),
       `${limit} is missing but optional for now`
     );
     this.defaultLimit = getDefaultLimitForAdapter(this.adapter, limit);
--- a/src/webgpu/api/validation/encoding/encoder_open_state.spec.ts
+++ b/src/webgpu/api/validation/encoding/encoder_open_state.spec.ts
@@ -98,6 +98,7 @@ const kRenderPassEncoderCommandInfo: {
   setScissorRect: {},
   setBlendConstant: {},
   setStencilReference: {},
+  setImmediates: {},
   beginOcclusionQuery: {},
   endOcclusionQuery: {},
   executeBundles: {},
@@ -122,6 +123,7 @@ const kRenderBundleEncoderCommandInfo: {
   setBindGroup: {},
   setIndexBuffer: {},
   setVertexBuffer: {},
+  setImmediates: {},
   pushDebugGroup: {},
   popDebugGroup: {},
   insertDebugMarker: {},
@@ -139,6 +141,7 @@ const kComputePassEncoderCommandInfo: {
 } = {
   setBindGroup: {},
   setPipeline: {},
+  setImmediates: {},
   dispatchWorkgroups: {},
   dispatchWorkgroupsIndirect: {},
   pushDebugGroup: {},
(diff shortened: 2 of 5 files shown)


src/webgpu/api/validation/capability_checks/limits/limit_utils.ts
```diff
@@ -404,9 +404,11 @@ export class LimitTestsImpl extends GPUTestBase {
     this._adapter = await gpu.requestAdapter();
     const limit = this.limit;
     // MAINTENANCE_TODO: consider removing this skip if the spec has no optional limits.
+    // Note: The cast below is required because an optional limit has no entry
+    // in capability_info.ts kLimitInfoKeys, kLimitInfoDefaults, kLimitInfoData
     this.skipIf(
       (this._adapter?.limits[limit] === undefined && !!this.limitTestParams.limitOptional) ||
-        getDefaultLimitsForCTS()[limit] === undefined,
+        (getDefaultLimitsForCTS() as Record<string, object>)[limit] === undefined,
```
kainino0x: nit: maybe `!(limit in getDefaultLimitsForCTS())`

or have `capability_info` export a (static) array of limit names?

src/webgpu/api/validation/encoding/encoder_open_state.spec.ts
```diff
@@ -298,6 +301,11 @@ g.test('render_pass_commands')
     const encoder = t.device.createCommandEncoder();
     const renderPass = beginRenderPassWithQuerySet(t, encoder, querySet);
 
+    t.skipIf(
+      command === 'setImmediates' && !('setImmediates' in renderPass),
+      'setImmediates not supported'
+    );
```
kainino0x: nit: prefer to do this in `beforeAllSubcases` - just as a matter of style, it's not a big deal for this test since it only has 3 subcases

src/webgpu/api/validation/capability_checks/limits/limit_utils.ts
```diff
@@ -404,9 +404,11 @@ export class LimitTestsImpl extends GPUTestBase {
     this._adapter = await gpu.requestAdapter();
     const limit = this.limit;
     // MAINTENANCE_TODO: consider removing this skip if the spec has no optional limits.
+    // Note: The cast below is required because an optional limit has no entry
+    // in capability_info.ts kLimitInfoKeys, kLimitInfoDefaults, kLimitInfoData
     this.skipIf(
       (this._adapter?.limits[limit] === undefined && !!this.limitTestParams.limitOptional) ||
```
kainino0x, line 410: [1]

src/webgpu/capability_info.ts
```diff
@@ -817,6 +777,7 @@ const [kLimitInfoKeys, kLimitInfoDefaults, kLimitInfoData] =
   'maxComputeWorkgroupSizeY':                  [           ,       256,             128,                          ],
   'maxComputeWorkgroupSizeZ':                  [           ,        64,              64,                          ],
   'maxComputeWorkgroupsPerDimension':          [           ,     65535,           65535,                          ],
+  // DO NOT ADD OPTIONAL limits here for proposals to the spec. Any limit here is immediately required by all implementations.
```
kainino0x: Ah sorry I thought the issue was `setImmediates` rather than `maxImmediateSize`. What test requires all the keys in this table? I couldn't find it by code-hopping but we should look at how we could make it understand "optional"/"temporarily optional" limits.

kainino0x: I guess this works because we don't have a static-assertion right now that this table has all the limits from the types? But we probably _should_ have that, so this table doesn't get outdated.

kainino0x (commented):
> * Should `maxImmediateSize` and `setImmediates` have been added to `@webgpu/types`? Add them there makes things here fail. If they should be removed we need a PR upstream

Sorry I didn't try the CTS side of this change before landing that. ~I _think_ what we need to do here is just add non-null assertion operator `!` in a lot of places - once the test does feature detection, it can assume this stuff exists (as it was already doing).~ [EDIT: oops ignore that it's irrelevant and I forgot to delete before posting]

src/webgpu/api/validation/capability_checks/limits/limit_utils.ts
greggman: done

src/webgpu/api/validation/encoding/encoder_open_state.spec.ts
greggman: done

src/webgpu/capability_info.ts
greggman: * `src/webgpu/api/operation/adapter/requestDevice.spec.ts`
* `src/webgpu/idl/javascript.spec.ts`

greggman: those can be adjusted in other PRs?

kainino0x: Got it sounds good. Yeah let's make sure those are set up so that they don't test these limits somehow. Probably a table of `limit name` -> `{ nonStandard }`

kainino0x: Could you file a bug and put a `MAINTENANCE_TODO` here (even though technically this is a testing `TODO`. there's no good place for a `TODO`)

greggman: Added a TODO and referenced https://github.com/gpuweb/cts/issues/4535. Let's discuss what to do over there.

PR #4537, opened 2025-12-20
Test reflection of GPUTexture.textureBindingViewDimension.

greggman:
In core it should always be undefined.
In compat it should be what was passed in or the default.

This passes in a build of chromium that exposes this property

In compatibility mode

Screenshot 2025-12-19 at 17 09 14

When not in compatibility mode

Screenshot 2025-12-19 at 17 09 07

Also checked Firefox stable (as chrome stable has trials on)

Screenshot 2025-12-19 at 17 26 15

### DIFF
--- a/src/webgpu/api/operation/command_buffer/copyTextureToTexture.spec.ts
+++ b/src/webgpu/api/operation/command_buffer/copyTextureToTexture.spec.ts
@@ -33,8 +33,23 @@ import { DataArrayGenerator } from '../../../util/texture/data_generation.js';
 import { kBytesPerRowAlignment, dataBytesForCopyOrFail } from '../../../util/texture/layout.js';
 import { TexelView } from '../../../util/texture/texel_view.js';
 import { findFailedPixels } from '../../../util/texture/texture_ok.js';
+import { reifyExtent3D } from '../../../util/unions.js';
 
 const dataGenerator = new DataArrayGenerator();
+
+// If a texture could be textureBindingViewDimension: 'cube' then set it to 'cube'
+function applyTextureBindingViewDimensionForTest(descriptor: GPUTextureDescriptor) {
+  const size = reifyExtent3D(descriptor.size);
+  if (
+    descriptor.textureBindingViewDimension === undefined &&
+    descriptor.dimension === '2d' &&
+    size.width === size.height &&
+    size.depthOrArrayLayers === 6
+  ) {
+    descriptor.textureBindingViewDimension = 'cube';
+  }
+  return descriptor;
+}
 
 class F extends AllFeaturesMaxLimitsGPUTest {
   getInitialDataPerMipLevel(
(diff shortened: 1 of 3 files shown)


src/webgpu/api/operation/reflection.spec.ts
```diff
@@ -172,6 +237,10 @@ g.test('texture_reflection_attributes')
       t.expect(texture.dimension === (descriptor.dimension || '2d'));
       t.expect(texture.mipLevelCount === (descriptor.mipLevelCount || 1));
       t.expect(texture.sampleCount === (descriptor.sampleCount || 1));
+      t.expect(
+        texture.textureBindingViewDimension ===
```
kainino0x, line 241: aside: We should also have a separate (`idl`) test for `'textureBindingViewDimension' in GPUTexture.prototype`. It's a really minor interoperability thing, but I think we do want all browsers to implement it

src/webgpu/api/operation/reflection.spec.ts
```diff
@@ -115,18 +117,59 @@ const kTextureSubcases: readonly {
...
     usage: GPUConst.TextureUsage.TEXTURE_BINDING,
     dimension: '3d',
+    textureBindingViewDimension: '3d',
+  },
+  {
+    size: [16, 16, 6],
+    format: 'rgba8unorm',
+    usage: GPUConst.TextureUsage.TEXTURE_BINDING,
+    dimension: '2d',
+    textureBindingViewDimension: 'cube',
```
kainino0x, line 158: It just occurred to me that if you request `'cube-array'` here, it looks like the spec will allow you to create the GPUTexture, just not allow you to create a GPUTextureView (but you can still use it for copies, I guess). This seems like an oversight in the spec. Can you verify?

src/webgpu/api/operation/reflection.spec.ts
greggman: Do we? At the moment the only thing a core implementation needs to do is add the 4 new limits. They do not need to add this `textureBindingViewDimension` and have it be undefined. It's undefined by the fact that it doesn't exist. I feel like that's fine. I also feel like apple would push back. From their POV it would be a wart and things would arguably work without adding it. (vs the limits which were required)

src/webgpu/api/operation/reflection.spec.ts
greggman: sigh.... Filed an issue: https://github.com/gpuweb/gpuweb/issues/5496

I checked and, copying is not tested with `textureBindingViewDimesion`. I hacked it to check and it passed on Linux OpenGL compat which I was surprised at. I checked and it's because it's OpenGL and [Dawn maps cube-array to `GL_TEXTURE_CUBE_ARRAY`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/src/dawn/native/opengl/TextureGL.cpp;l=66) so it happens to work there but should fail on actual GLES 3.1 which doesn't have `GL_TEXTURE_CUBE_ARRAY`

greggman: I'm going to mark this as resolved if that's ok as this is an operation test, not a validation test. Added the issue above as well as [this PR](https://github.com/gpuweb/cts/pull/4543) in anticipation of validating out `cube-array`.

src/webgpu/api/operation/reflection.spec.ts
greggman: Going to mark this as resolved of that's ok. If we decide the getter is required then we can add an idl test at that time.

kainino0x: Ack. Filed https://github.com/gpuweb/gpuweb/issues/5498

kainino0x (approved):

Issue #4538, opened 2025-12-20
`maxInterStageShaderVariables`  should be decremented for _each_ fragment input built-in, not just once


ErichDonGubler:
In this following inter-stage interface validation routine in the WebGPU spec. ([link](https://www.w3.org/TR/webgpu/#abstract-opdef-validating-inter-stage-interfaces), emphasis mine):

> validating inter-stage interfaces(device, descriptor)
>
> …
>
> [Device timeline](https://www.w3.org/TR/webgpu/#device-timeline) steps:
>
> …
>
>
> 7.  If <code>descriptor.<a href="https://www.w3.org/TR/webgpu/#dom-gpurenderpipelinedescriptor-fragment">fragment</a></code> [is provided](https://infra.spec.whatwg.org/#map-exists):
>
>     1.  Let maxFragmentShaderInputVariables be <code>device.limits.<a href="https://www.w3.org/TR/webgpu/#dom-supported-limits-maxinterstageshadervariables">maxInterStageShaderVariables</a></code>.
>
>     2.  **For each of the [Inter-Stage Builtins](https://www.w3.org/TR/webgpu/#inter-stage-builtins) that are an input of <code>descriptor.<a href="https://www.w3.org/TR/webgpu/#dom-gpurenderpipelinedescriptor-fragment">fragment</a></code>:**
>
>         1.  **Decrement maxFragmentShaderInputVariables by 1.**

I interpret this to mean: for each built-in, decrement once _per built-in found_. However, the [`webgpu:api,validation,capability_checks,limits,maxInterStageShaderVariables:createRenderPipeline,at_over:*` test path in CTS](https://gpuweb.github.io/cts/standalone/?q=webgpu:api,validation,capability_checks,limits,maxInterStageShaderVariables:createRenderPipeline,at_over:*) only decrements the maximum _once_ if _any_ of them are set:

https://github.com/gpuweb/cts/blob/d5a131f3afc77be7f76bbf956db054e799f2e79d/src/webgpu/api/validation/capability_checks/limits/maxInterStageShaderVariables.spec.ts#L19-L21

~~Which interpretation is right here?~~ EDIT: Confirmed, it should be a deduction per built-in.

PR #4546, opened 2025-12-27
Fix fragment input deductions in `max_variables_count,input`

ErichDonGubler:
Issue: #4538

See also <https://github.com/gpuweb/cts/pull/4539>.

<hr>

### DIFF
--- a/src/webgpu/api/validation/render_pipeline/inter_stage.spec.ts
+++ b/src/webgpu/api/validation/render_pipeline/inter_stage.spec.ts
@@ -315,15 +315,15 @@ g.test('max_variables_count,input')
     u.combine('isAsync', [false, true]).combineWithParams([
       // Number of user-defined output variables in test shader =
       //     device.limits.maxInterStageShaderVariables + numVariablesDelta
-      { numVariablesDelta: 0, useExtraBuiltinInputs: false },
-      { numVariablesDelta: 1, useExtraBuiltinInputs: false },
-      { numVariablesDelta: -1, useExtraBuiltinInputs: true },
-      { numVariablesDelta: -2, useExtraBuiltinInputs: true },
-      { numVariablesDelta: -3, useExtraBuiltinInputs: true },
+      { numVariablesDelta: 0, useExtraBuiltinInputs: false, _success: true },
+      { numVariablesDelta: 1, useExtraBuiltinInputs: false, _success: false },
+      { numVariablesDelta: -1, useExtraBuiltinInputs: true, _success: true },
+      { numVariablesDelta: -2, useExtraBuiltinInputs: true, _success: false },
+      { numVariablesDelta: -3, useExtraBuiltinInputs: true, _success: true },
     ] as const)
   )
   .fn(t => {
-    const { isAsync, numVariablesDelta, useExtraBuiltinInputs } = t.params;
+    const { isAsync, numVariablesDelta, useExtraBuiltinInputs, _success: success } = t.params;
 
     const numVec4 = t.device.limits.maxInterStageShaderVariables + numVariablesDelta;
 
@@ -340,7 +340,6 @@ g.test('max_variables_count,input')
         );
       }
     }
-    const success = inputs.length <= t.device.limits.maxInterStageShaderVariables;
 
     const descriptor = t.getDescriptorWithStates(
       t.getVertexStateWithOutputs(outputs),


src/webgpu/api/validation/render_pipeline/inter_stage.spec.ts
```diff
@@ -319,19 +319,18 @@ g.test('max_variables_count,input')
       { numVariablesDelta: 1, useExtraBuiltinInputs: false },
       { numVariablesDelta: 0, useExtraBuiltinInputs: true },
       { numVariablesDelta: -1, useExtraBuiltinInputs: true },
+      { numVariablesDelta: -3, useExtraBuiltinInputs: true },
```
kainino0x: nit: Test -2 and -3 because those are the cases on either side of the boundary. (Helps catch off-by-one errors or if an implementation tries to merge two of the variables but not the third.)
```suggestion
      { numVariablesDelta: -2, useExtraBuiltinInputs: true },
      { numVariablesDelta: -3, useExtraBuiltinInputs: true },
```

pnpm-lock.yaml
kainino0x, line 1: Accidental extra file.

ErichDonGubler: Uuuuuugh, sorry! Fixed with the latest force-push to 7c3fdf75.

Maybe we should add this to `.gitignore`? 😅 Or maybe I should just be more disciplined with my review.

src/webgpu/api/validation/render_pipeline/inter_stage.spec.ts
ErichDonGubler: Added the `-2` case with the latest force-push to 7c3fdf75. I want to clarify, though: your suggestion block would strip the `-1` and `0` cases, and I'm guessing that's not intentional. Is that right?

pnpm-lock.yaml
kainino0x: gitignore seems good! I don't know pnpm vs npm stuff.

src/webgpu/api/validation/render_pipeline/inter_stage.spec.ts
kainino0x: IIUC (please let me know if I'm misunderstanding the test):

It used to test -1 and 0 because those were on either side of the boundary (-1 would give space for the builtins, 0 would not). Now the boundary is between -3 (max valid) and -2 (min invalid). I think it should be pretty safe to assume that if the implementation gets -2 correct, then it will get -1 and 0 correct as well.

To be fair this is not clear because of the structure of the test since it computes `success` in the test body instead of passing it along with each case. It would probably be a good idea to do a little refactor here and make `success` a parameter (`_success: true` / `_success: false`) so we can actually tell that the test is testing the boundaries it intends to.

ErichDonGubler: Makes sense. Changed cases (but not the `_success` refactor) as suggested with latest force-push to f9fd2cc09693.

ErichDonGubler: Ah, but wait: If this is compat, then `-1` is still an interesting boundary case to test, isn't it? It seems like that's still interesting to keep. That also means that we can't naively add `_success` to the case parameters, because whether it should succeed or not would actually depend on how many built-ins are being added, which is variable.

ErichDonGubler: Force-pushed to e0157af96f44 to include `-1`.

kainino0x: very good point... I should have actually looked at the code again before suggesting things.

I pulled it locally and experimented with another fix idea. Couldn't push to your branch so I made a new PR here that includes your commits and mine. PTAL: #4555

src/webgpu/api/validation/render_pipeline/inter_stage.spec.ts
```diff
@@ -315,23 +315,22 @@ g.test('max_variables_count,input')
...
       //     device.limits.maxInterStageShaderVariables + numVariablesDelta
-      { numVariablesDelta: 0, useExtraBuiltinInputs: false },
-      { numVariablesDelta: 1, useExtraBuiltinInputs: false },
-      { numVariablesDelta: 0, useExtraBuiltinInputs: true },
-      { numVariablesDelta: -1, useExtraBuiltinInputs: true },
+      { numVariablesDelta: 0, useExtraBuiltinInputs: false, _success: true },
+      { numVariablesDelta: 1, useExtraBuiltinInputs: false, _success: false },
+      { numVariablesDelta: -1, useExtraBuiltinInputs: true, _success: true },
+      { numVariablesDelta: -2, useExtraBuiltinInputs: true, _success: false },
+      { numVariablesDelta: -3, useExtraBuiltinInputs: true, _success: true },
```
greggman, line 322: AFAICT `numVariablesDelta` is how many to add to `maxInterStageShaderVariables`. I don't think you can really decide where how many to subtract and statically determine success without changing the rest of the test. If the test is in compatibility mode then only 1 extra builtin will be added. If it's in core, 3 will be added.

kainino0x: There is a thread about this above https://github.com/gpuweb/cts/pull/4546#discussion_r2674851861
and that's how we ended up with #4555. Actually I should close this PR (at least for now) in favor of that one.

PR #4547, opened 2025-12-31
Validation: Add immediate data pipeline validation tests

shaoboyan091:
Adds `pipeline_immediate.spec.ts` to validate immediate data usage in RenderPassEncoder, ComputePassEncoder, and RenderBundleEncoder.

Tests cover:
- Required immediate data slots are set.
- Unused variables do not require slots.
- Pipeline creation fails if shader immediate size exceeds layout limit.
- RenderBundle execution invalidates pipeline and immediate data state.

Issue: #

<hr>

### DIFF
--- a/src/webgpu/api/operation/command_buffer/queries/timestampQuery.spec.ts
+++ b/src/webgpu/api/operation/command_buffer/queries/timestampQuery.spec.ts
@@ -35,55 +35,64 @@ and prevent pages from running.
       .combine('numQuerySets', [8, 16, 32, 64, 256, 65536] as const)
       .combine('stage', ['compute', 'render'] as const)
   )
-  .fn(t => {
+  .fn(async t => {
     const { stage, numQuerySets } = t.params;
 
     t.skipIfDeviceDoesNotHaveFeature('timestamp-query');
 
-    const view = t
-      .createTextureTracked({
-        size: [1, 1, 1],
-        format: 'rgba8unorm',
-        usage: GPUTextureUsage.RENDER_ATTACHMENT,
-      })
-      .createView();
-    const encoder = t.device.createCommandEncoder();
+    // At large numQuerySets, this test incurs a lot of validation, which can take several seconds.
+    // Explicitly wrap the test in its own error scope to avoid triggering timeouts in test-cleanup.
+    t.device.pushErrorScope('validation');
+    try {
+      const view = t
+        .createTextureTracked({
+          size: [1, 1, 1],
+          format: 'rgba8unorm',
+          usage: GPUTextureUsage.RENDER_ATTACHMENT,
+        })
+        .createView();
+      const encoder = t.device.createCommandEncoder();
 
-    for (let i = 0; i < numQuerySets; ++i) {
-      const querySet = t.createQuerySetTracked({
-        type: 'timestamp',
-        count: 2,
-      });
+      for (let i = 0; i < numQuerySets; ++i) {
+        const querySet = t.createQuerySetTracked({
+          type: 'timestamp',
+          count: 2,
+        });
 
-      switch (stage) {
-        case 'compute': {
-          const pass = encoder.beginComputePass({
-            timestampWrites: {
-              querySet,
-              beginningOfPassWriteIndex: 0,
-              endOfPassWriteIndex: 1,
-            },
-          });
-          pass.end();
-          break;
-        }
-        case 'render': {
-          const pass = encoder.beginRenderPass({
-            colorAttachments: [{ view, loadOp: 'load', storeOp: 'store' }],
-            timestampWrites: {
-              querySet,
-              beginningOfPassWriteIndex: 0,
-              endOfPassWriteIndex: 1,
-            },
-          });
-          pass.end();
-          break;
+        switch (stage) {
+          case 'compute': {
+            const pass = encoder.beginComputePass({
+              timestampWrites: {
+                querySet,
+                beginningOfPassWriteIndex: 0,
+                endOfPassWriteIndex: 1,
+              },
+            });
+            pass.end();
+            break;
+          }
+          case 'render': {
+            const pass = encoder.beginRenderPass({
+              colorAttachments: [{ view, loadOp: 'load', storeOp: 'store' }],
+              timestampWrites: {
+                querySet,
+                beginningOfPassWriteIndex: 0,
+                endOfPassWriteIndex: 1,
+              },
+            });
+            pass.end();
+            break;
+          }
         }
       }
-    }
 
-    const shouldError = false; // just expect no error
-    t.expectValidationError(() => t.device.queue.submit([encoder.finish()]), shouldError);
+      const shouldError = false; // just expect no error
+      t.expectValidationError(() => t.device.queue.submit([encoder.finish()]), shouldError);
+    } finally {
+      const error = await t.device.popErrorScope();
+      // Make sure there weren't any unexpected validation errors caught by the scope.
+      t.expect(error === null, error?.message);
+    }
   });
 
 g.test('many_slots')
--- a/src/webgpu/api/operation/render_pass/transient_attachment.spec.ts
+++ b/src/webgpu/api/operation/render_pass/transient_attachment.spec.ts
@@ -1,0 +1,94 @@
+export const description = `API Operation Tests for transient attachment in render passes.`;
+
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { AllFeaturesMaxLimitsGPUTest } from '../../../gpu_test.js';
+
+const kSize = 4;
+
+export const g = makeTestGroup(AllFeaturesMaxLimitsGPUTest);
+
+g.test('increasing_attachments_count')
+  .desc(
+    `
+    Use multiple render passes with increasing amounts of transient attachments.
+  `
+  )
+  .fn(t => {
+    // MAINTENANCE_TODO(#4509): Remove this when TRANSIENT_ATTACHMENT is added to the WebGPU spec.
+    t.skipIfTransientAttachmentNotSupported();
+
+    const maxAttachments = t.device.limits.maxColorAttachments;
+    const encoder = t.device.createCommandEncoder();
+
+    for (let count = 1; count <= maxAttachments; count++) {
+      const colorAttachments: GPURenderPassColorAttachment[] = [];
+
+      for (let i = 0; i < count; i++) {
+        const texture = t.createTextureTracked({
+          size: [kSize, kSize, 1],
+          format: 'rgba8unorm',
+          usage: GPUTextureUsage.RENDER_ATTACHMENT | GPUTextureUsage.TRANSIENT_ATTACHMENT,
+        });
+
+        colorAttachments.push({
+          view: texture.createView(),
+          clearValue: { r: 0.0, g: 0.0, b: 0.0, a: 0.0 },
+          loadOp: 'clear',
+          storeOp: 'discard',
+        });
+      }
+
+      const pass = encoder.beginRenderPass({ colorAttachments });
+      pass.end();
+    }
+    t.device.queue.submit([encoder.finish()]);
+  });
+
+g.test('overlapping_transient_attachments')
+  .desc(
+    `
+    Use multiple render passes with transient attachments in a circular overlap pattern:
+    Pass 1: (T1, T2)
+    Pass 2: (T2, T3)
+    Pass 3: (T3, T1)
+
+    This stresses the driver's transient memory allocator. If the driver naively reuses
+    T1's memory for T3 during Pass 2 (because T1 isn't active in Pass 2),
+    Pass 3 will fail or corrupt because T1 and T3 are both needed simultaneously again.
+  `
+  )
+  .fn(t => {
+    // MAINTENANCE_TODO(#4509): Remove this when TRANSIENT_ATTACHMENT is added to the WebGPU spec.
+    t.skipIfTransientAttachmentNotSupported();
+
+    const encoder = t.device.createCommandEncoder();
+
+    const desc = {
+      size: [kSize, kSize, 1],
+      format: 'rgba8unorm',
+      usage: GPUTextureUsage.RENDER_ATTACHMENT | GPUTextureUsage.TRANSIENT_ATTACHMENT,
+    } as const;
+    const t1 = t.createTextureTracked(desc);
+    const t2 = t.createTextureTracked(desc);
+    const t3 = t.createTextureTracked(desc);
+
+    const passes = [
+      [t1, t2], // Pass 1
+      [t2, t3], // Pass 2
+      [t3, t1], // Pass 3
+    ];
+
+    for (const attachments of passes) {
+      const colorAttachments: GPURenderPassColorAttachment[] = attachments.map(texture => ({
+        view: texture.createView(),
+        clearValue: { r: 0.0, g: 0.0, b: 0.0, a: 0.0 },
+        loadOp: 'clear',
+        storeOp: 'discard',
+      }));
+
+      const pass = encoder.beginRenderPass({ colorAttachments });
+      pass.end();
+    }
+
+    t.device.queue.submit([encoder.finish()]);
+  });
--- a/src/webgpu/shader/execution/expression/call/builtin/extractBits.spec.ts
+++ b/src/webgpu/shader/execution/expression/call/builtin/extractBits.spec.ts
@@ -179,6 +179,11 @@ g.test('u32')
               0b00000000000000000000000010101010
             ),
           },
+          // Difficult edge case involving a MAX_UINT 'count'.
+          {
+            input: [V(5), u32(1), u32(4294967295)],
+            expected: V(2),
+          },
         ]
       );
     }
@@ -318,6 +323,11 @@ g.test('i32')
               0b00000000000000000000000101010101,
               0b00000000000000000000000010101010
             ),
+          },
+          // Difficult edge case involving a MAX_UINT 'count'.
+          {
+            input: [V(5), u32(1), u32(4294967295)],
+            expected: V(2),
           },
         ]
       );
--- a/src/webgpu/shader/execution/expression/call/builtin/insertBits.spec.ts
+++ b/src/webgpu/shader/execution/expression/call/builtin/insertBits.spec.ts
@@ -337,7 +337,6 @@ g.test('integer')
           0b01010101010101010101010101010101
         ),
       },
-
       // Zero count
       { input: [pattern, all_1, u32(0), u32(0)], expected: pattern },
       { input: [pattern, all_1, u32(1), u32(0)], expected: pattern },
@@ -367,6 +366,11 @@ g.test('integer')
               0b10101010101010101010101010101010,
               0b01010101010101010101010101010101
             ),
+          },
+          // Difficult edge case involving a MAX_UINT 'count'.
+          {
+            input: [V(5), V(1), u32(1), u32(4294967295)],
+            expected: V(3),
           },
         ]
       );
--- a/src/webgpu/shader/validation/decl/assignment_statement.spec.ts
+++ b/src/webgpu/shader/validation/decl/assignment_statement.spec.ts
@@ -1,0 +1,82 @@
+export const description = `
+Validation tests for assignment statements.
+`;
+
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { keysOf } from '../../../../common/util/data_tables.js';
+import { ShaderValidationTest } from '../shader_validation_test.js';
+
+export const g = makeTestGroup(ShaderValidationTest);
+
+g.test('scalar_assignment')
+  .desc('Test simple scalar assignments.')
+  .fn(t => {
+    const code = `
+      @fragment
+      fn main() {
+        var a: i32 = 0;
+        a = 1;
+        let b: f32 = 0.0;
+        var c: f32;
+        c = b;
+      }
+    `;
+    t.expectCompileResult(true, code);
+  });
+
+g.test('vector_full_assignment')
+  .desc('Test full vector assignments.')
+  .fn(t => {
+    const code = `
+      @fragment
+      fn main() {
+        var v1: vec3<f32> = vec3(0.0, 0.0, 0.0);
+        var v2: vec3<f32>;
+        v2 = v1;
+        v2 = vec3(1.0, 2.0, 3.0);
+      }
+    `;
+    t.expectCompileResult(true, code);
+  });
+
+g.test('vector_indexed_assignment')
+  .desc('Test vector indexed assignments.')
+  .fn(t => {
+    const code = `
+      @fragment
+      fn main() {
+        var v: vec3<i32> = vec3(0, 0, 0);
+        v[0] = 1;
+        v[2] = 5;
+      }
+    `;
+    t.expectCompileResult(true, code);
+  });
+
+const kSwizzleTests = {
+  single: {
+    src: 'v.x = 1.0',
+    pass: true,
+  },
+  multi: {
+    src: 'v.xy = vec2(1.0, 2.0)',
+    pass: false,
+  },
+  swizzleswizzle: {
+    src: 'v.xy.x = 1.0',
+    pass: false,
+  },
+};
+
+g.test('vector_swizzle_assignment')
+  .desc('Test vector swizzle assignments.')
+  .params(u => u.combine('case', keysOf(kSwizzleTests)))
+  .fn(t => {
+    const wgsl = `
+      @fragment
+      fn main() {
+        var v: vec4<f32> = vec4(0.0, 0.0, 0.0, 0.0);
+        ${kSwizzleTests[t.params.case].src};
+      }`;
+    t.expectCompileResult(kSwizzleTests[t.params.case].pass, wgsl);
+  });
--- a/src/webgpu/shader/validation/expression/access/vector.spec.ts
+++ b/src/webgpu/shader/validation/expression/access/vector.spec.ts
@@ -150,6 +150,10 @@ const kConcreteCases = {
   xyxz_rbg_yx: { wgsl: 'let r : vec2<T> = v.xyxz.rbg.yx;', ok: (width: number) => width > 2 },
   wxyz_bga_xy: { wgsl: 'let r : vec2<T> = v.wxyz.bga.xy;', ok: (width: number) => width > 3 },
 
+  // mixed swizzle and indexing
+  xy_0: { wgsl: 'let r = v.xy[0];', result_width: 1, ok: true },
+  xy_3: { wgsl: 'let r = v.xy[3];', ok: false },
+
   // error: invalid convenience letterings
   xq: { wgsl: 'let r : vec2<T> = v.xq;', ok: false },
   py: { wgsl: 'let r : vec2<T> = v.py;', ok: false },
@@ -357,6 +361,10 @@ const kAbstractCases = {
     result_width: 2,
     ok: (width: number) => width > 3,
   },
+
+  // mixed swizzle and indexing
+  xy_0: { wgsl: 'const r = V.xy[0];', result_width: 1, ok: true },
+  xy_3: { wgsl: 'const r = V.xy[3];', ok: false },
 
   // error: invalid convenience letterings
   xq: { wgsl: 'const r = V.xq;', ok: false },


src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
```diff
@@ -0,0 +1,505 @@
...
+    `
+    Validate that creating a pipeline fails if the shader uses immediate data
+    larger than the immediateSize specified in the pipeline layout.
+    `
+  )
+  .params(u => u.combine('encoderType', kProgrammableEncoderTypes))
+  .fn(t => {
+    const { encoderType } = t.params;
+    const kLayoutSize = 16;
+    const kShaderSize = 32; // Larger than layout
```
Kangz: Could these be parameters with a few cases, including just 4 bytes more and equal values?

src/webgpu/api/validation/createPipelineLayout.spec.ts
```diff
@@ -512,3 +512,45 @@ g.test('bind_group_layouts,set_pipeline_with_null_bind_group_layouts')
       }
     }
   });
+
+g.test('immediate_data_size')
```
Kangz: Do we have the same test for the defaulted pipeline layouts?

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
```diff
@@ -0,0 +1,505 @@
...
+      pass.setPipeline(pipeline);
+      pass.draw(3);
+      validateFinishAndSubmit(false, true);
+    } else if (check === 'pipeline_and_immediates_reset') {
+      // Re-set pipeline and immediates. Should succeed.
+      pass.setPipeline(pipeline);
+      (pass as unknown as EncoderWithImmediates).setImmediates(0, immediateData, 0, 16);
+      pass.draw(3);
+      validateFinishAndSubmit(true, true);
+    }
```
Kangz, line 437: Could we also have tests for the validation of setImmediates itself?

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
```diff
@@ -0,0 +1,505 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    t.runPass(encoderType, encoder as any, pipeline);
+
+    validateFinishAndSubmit(true, true);
+  });
+
+g.test('overprovisioned_immediate_data')
+  .desc(
+    `
+    Validate that setting more immediate data than used by the shader (but within layout limits) is valid.
```
Kangz: Setting more immediates than the layout is also valid. What we should test it the validation of setImmediates.

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
```diff
@@ -0,0 +1,505 @@
...
+      }
+    });
+  });
+
+g.test('render_bundle_execution_state_invalidation')
+  .desc(
+    `
+    Validate that executeBundles invalidates the current pipeline and immediate data state
+    in the RenderPassEncoder.
+    - Pipeline must be re-set after executeBundles.
```
Kangz: Isn't this tested somewhere else already?

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
```diff
@@ -0,0 +1,505 @@
...
+          fn use_data() { _ = data; }
+          @compute @workgroup_size(1) fn main_compute() { use_data(); }
+          @vertex fn main_vertex() -> @builtin(position) vec4<f32> { use_data(); return vec4<f32>(0.0, 0.0, 0.0, 1.0); }
+          @fragment fn main_fragment() -> @location(0) vec4<f32> { use_data(); return vec4<f32>(0.0, 1.0, 0.0, 1.0); }
+        `;
+        break;
+      case 'function_unused':
+        kRequiredSize = 16;
+        code = `
+          var<immediate> data: vec4<u32>;
```
Kangz: It seems that we could use the switch to decide how to declare `data` and `use_data` but that the rest of the code can be the same between all cases. It would reduce some noise.

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
```diff
@@ -0,0 +1,505 @@
...
+          @compute @workgroup_size(1) fn main_compute() { _ = data; }
+          @vertex fn main_vertex() -> @builtin(position) vec4<f32> { _ = data; return vec4<f32>(0.0, 0.0, 0.0, 1.0); }
+          @fragment fn main_fragment() -> @location(0) vec4<f32> { _ = data; return vec4<f32>(0.0, 1.0, 0.0, 1.0); }
+        `;
+        break;
+      case 'struct_padding':
+        kRequiredSize = 64;
```
Kangz:
```suggestion
        kRequiredSize = 40;
```

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
```diff
@@ -0,0 +1,505 @@
...
+    const { encoder, validateFinishAndSubmit } = t.createEncoder(encoderType);
+
+    const setImmediates = (offset: number, size: number) => {
+      const data = new Uint8Array(size);
+      (encoder as unknown as EncoderWithImmediates).setImmediates(offset, data, 0, size);
+    };
+
+    if (scenario === 'struct_padding') {
+      // Actual data ends at byte 40. A.v: 0-12. Padding: 12-32 (includes A padding 12-16 and inter-member padding 16-32). B.v: 32-40. Total struct size: 64.
+      if (usage === 'full') {
```
Kangz: It seems that partial, full and overprovision could all be handled the same way for all cases (we'd need a to make split handle the exact coverage for `struct_padding` in that case).

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
kainino0x: There are tests in `src/webgpu/api/validation/encoding/cmds/setImmediates.spec.ts`, that probably cover it?

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
```diff
@@ -0,0 +1,505 @@
...
+
+    if (
+      !supportsRenderPass &&
+      !supportsComputePass &&
+      !supportsRenderBundle &&
+      !supportsLimit &&
+      !supportsWgslFeature
+    ) {
+      this.skip('setImmediates not supported');
+    }
```
kainino0x: nit: Share this logic with `src/webgpu/api/validation/encoding/cmds/setImmediates.spec.ts`

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
```diff
@@ -0,0 +1,505 @@
...
+import { getGPU } from '../../../../../common/util/navigator_gpu.js';
+import { AllFeaturesMaxLimitsGPUTest } from '../../../../gpu_test.js';
+import {
+  kProgrammableEncoderTypes,
+  ProgrammableEncoderType,
+} from '../../../../util/command_buffer_maker.js';
+
+interface EncoderWithImmediates {
+  setImmediates(offset: number, data: Uint8Array, srcOffset: number, size: number): void;
+}
```
kainino0x: We already have this in the `@webgpu/types` dependency, I think you should be able to use it with
`encoder.setImmediates!(offset, data, srcOffset, number)`

(Could you also clean this up in `src/webgpu/api/validation/encoding/cmds/setImmediates.spec.ts`?)

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
```diff
@@ -0,0 +1,505 @@
...
+      });
+    }
+  }
+
+  runPass(
+    encoderType: ProgrammableEncoderType,
+    encoder: GPUComputePassEncoder | GPURenderPassEncoder | GPURenderBundleEncoder,
+    pipeline: GPURenderPipeline | GPUComputePipeline
+  ) {
+    if (encoderType === 'compute pass') {
```
kainino0x: nit: If you branch on `encoder instanceof GPUComputePassEncoder`, it'll reduce the number of casts. I think the pass/bundle can also be handled as one case (polymorphically because they have the same interface).

(_Possibly_ you could even roll `createPipeline` into this function, and avoid the casts altogether, but don't try this until after addressing my other comments about vertex/fragment stages because that will change things.)

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
```diff
@@ -0,0 +1,505 @@
...
+      bindGroupLayouts: [],
+      immediateSize,
+    });
+
+    if (encoderType === 'compute pass') {
+      return this.device.createComputePipeline({
+        layout,
+        compute: {
+          module: this.device.createShaderModule({ code }),
+          entryPoint: 'main_compute',
```
kainino0x: nit: I think there is only one entry point of each type, so you can just omit this.

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
```diff
@@ -0,0 +1,505 @@
...
+    - Scenarios:
+      - scalar: Simple u32 usage.
+      - vector: Simple vec4<u32> usage.
+      - struct_padding: Struct with padding. Padding bytes do not need to be set.
+        When a struct variable is statically used, all non-padding bytes of the entire struct must be set,
+        even if only one member is accessed. In this test, data.b.v is accessed, but both data.a and data.b
+        must be set (excluding padding).
+      - functions: Immediate data used in a function called by entry point.
+      - dynamic_indexing: Array with dynamic indexing.
+    - Usage:
```
kainino0x, line 71: `overprovision` not explained here

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
```diff
@@ -0,0 +1,505 @@
...
+    - For each immediate data variable statically used by the pipeline:
+      - All accessible slots (4-byte words) must be set via setImmediates.
+    - Scenarios:
+      - scalar: Simple u32 usage.
+      - vector: Simple vec4<u32> usage.
+      - struct_padding: Struct with padding. Padding bytes do not need to be set.
+        When a struct variable is statically used, all non-padding bytes of the entire struct must be set,
+        even if only one member is accessed. In this test, data.b.v is accessed, but both data.a and data.b
+        must be set (excluding padding).
+      - functions: Immediate data used in a function called by entry point.
```
kainino0x: `function_unused` not explained here

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
```diff
@@ -0,0 +1,505 @@
...
+        'struct_padding',
+        'functions',
+        'function_unused',
+        'dynamic_indexing',
+      ] as const)
+      .combine('usage', ['full', 'partial', 'split', 'overprovision'] as const)
+      .filter(t => {
+        if (t.scenario === 'scalar' && t.usage === 'split') return false;
+        return true;
+      })
```
kainino0x: µNit: `t` is normally the "test", `p` is normally the "params". Also this can be shortened to read more naturally
```suggestion
      .unless(p => p.scenario === 'scalar' && p.usage === 'split')
```

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
```diff
@@ -0,0 +1,505 @@
...
+    let kRequiredSize = 0;
+
+    switch (scenario) {
+      case 'scalar':
+        kRequiredSize = 4;
+        code = `
+          var<immediate> data: u32;
+          @compute @workgroup_size(1) fn main_compute() { _ = data; }
+          @vertex fn main_vertex() -> @builtin(position) vec4<f32> { _ = data; return vec4<f32>(0.0, 0.0, 0.0, 1.0); }
+          @fragment fn main_fragment() -> @location(0) vec4<f32> { _ = data; return vec4<f32>(0.0, 1.0, 0.0, 1.0); }
```
kainino0x: The immediate will always be used in both stages of a render pipeline. Probably they should be tested separately because we want to make sure implementations are looking at both stages.

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
```diff
@@ -0,0 +1,505 @@
...
+        `;
+        break;
+      case 'vector':
+        kRequiredSize = 16;
+        code = `
+          var<immediate> data: vec4<u32>;
```
kainino0x: IIRC we didn't put anything in the spec requiring the vertex and fragment stages of a pipeline to both use matching var<immediate> blocks. What happens if they have different types? I assume they just alias the memory with different views (like would happen if you rebind the pipeline without changing the immediates, and dispatch again).

If so, that means that the required set of words is the union of words used in each stage. And we need to have a test that exercises this, with different types, ensuring both stages' accessible words are required to be initialized.

Combined with the suggestion above, this would 3x the number of render-pipeline cases:

- Vertex uses immediate, fragment doesn't
- Fragment uses immediate, vertex doesn't
- Vertex and fragment both use immediates but with different types

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
```diff
@@ -0,0 +1,505 @@
...
+  )
+  .params(u =>
+    u
+      .combine('encoderType', kProgrammableEncoderTypes)
+      .combine('scenario', [
+        'scalar',
+        'vector',
+        'struct_padding',
+        'functions',
+        'function_unused',
```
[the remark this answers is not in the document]
kainino0x: I started writing a comment about this, then I realized Copilot was onto the same thing. I think this case should be in a separate test, instead of part of this big parameterization, as it's testing  specific thing that probably doesn't benefit from being tested with multiple `usage`s.

It looks like `referenced_in_unused_function` already covers it.

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
kainino0x: `functions` seems orthogonal to the other scenarios which are about different patterns of words being used in the immediate block.

TBH I don't think it matters that much to specifically test usages in regular functions vs the entry point function - the "static use" concept is already well exercised by other tests. I would suggest just making _all_ of the tests reference the immediate block through a function call, and then remove this case.

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
```diff
@@ -0,0 +1,505 @@
...
+      // Size 4.
+      if (usage === 'full') {
+        setImmediates(0, 4);
+      } else if (usage === 'partial') {
+        // Set nothing
+      } else if (usage === 'overprovision') {
+        setImmediates(0, kRequiredSize + 4);
+      }
+    } else {
+      // Vector (size 16), functions (size 16), dynamic_indexing (size 16)
```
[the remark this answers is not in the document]
kainino0x: nit: Prefer to use `else if ... else { unreachable(); }` so it's clear all cases are considered. (or a `switch` with `default: unreachable()`)

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
```diff
@@ -0,0 +1,505 @@
...
+        setImmediates(0, 8);
+        setImmediates(8, 8);
+      } else if (usage === 'partial') {
+        setImmediates(0, 8); // Missing last 8 bytes
+      } else if (usage === 'overprovision') {
+        setImmediates(0, kRequiredSize + 4);
+      }
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
kainino0x: IIRC this cast shouldn't be needed.

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
```diff
@@ -0,0 +1,505 @@
...
+  .desc(
+    `
+    Validate that if an immediate data variable is declared but not statically used,
+    it does not require slots to be set.
+    `
+  )
+  .params(u =>
+    u
+      .combine('encoderType', kProgrammableEncoderTypes)
+      .combine('usage', ['none', 'full', 'partial_start'] as const)
```
kainino0x: I see how `none` vs `partial_start` could catch different bugs in an implementation, but `full` doesn't seem like it matters. I think that case can be removed.

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
```diff
@@ -0,0 +1,505 @@
...
+
+    if (usage === 'full') {
+      const data = new Uint8Array(16);
+      (encoder as unknown as EncoderWithImmediates).setImmediates(0, data, 0, 16);
+    } else if (usage === 'partial_start') {
+      const data = new Uint8Array(8);
+      (encoder as unknown as EncoderWithImmediates).setImmediates(0, data, 0, 8);
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
kainino0x: same

src/webgpu/api/validation/createPipelineLayout.spec.ts
```diff
@@ -512,3 +512,45 @@ g.test('bind_group_layouts,set_pipeline_with_null_bind_group_layouts')
...
+    - immediateSize must be <= device.limits.maxImmediateSize.
+  `
+  )
+  .params(u => u.combine('immediateSize', [0, 4, 'max', 'max+4', 1, 2, 3, 5] as const))
+  .fn(t => {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const maxImmediateSize = (t.device.limits as any).maxImmediateSize;
+
+    if (maxImmediateSize === undefined) {
+      t.skip('maxImmediateSize limit is not supported on this device.');
```
kainino0x: Use the common helper here too

src/webgpu/api/validation/createPipelineLayout.spec.ts
```diff
@@ -512,3 +512,45 @@ g.test('bind_group_layouts,set_pipeline_with_null_bind_group_layouts')
...
+    }
+
+    const { immediateSize: sizeVariant } = t.params;
+    let size: number;
+    if (sizeVariant === 'max') {
+      size = maxImmediateSize;
+    } else if (sizeVariant === 'max+4') {
+      size = maxImmediateSize + 4;
+    } else {
+      size = sizeVariant as number;
```
kainino0x: nit: I _think_ this cast shouldn't be needed?

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
kainino0x: +1, validation tests should have control cases, and should have invalid cases that are as close as possible to being valid.

Also this would be a good place to have a simple test that it's allowed for the shader to use _less_ than the pipeline layout declares (if not already tested elsewhere).

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
```diff
@@ -0,0 +1,505 @@
...
+    const data = new Uint8Array(32);
+    (encoder as unknown as EncoderWithImmediates).setImmediates(0, data, 0, 32);
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    t.runPass(encoderType, encoder as any, pipeline);
+
+    validateFinishAndSubmit(true, true);
+  });
+
+g.test('pipeline_creation_immediate_size_mismatch')
```
kainino0x: This is the encoding folder but this is not an encoding test. Should be categorized as a create-pipeline validation test somewhere.

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
kainino0x: +1. I think this test is basically fine as-is except it should have the setImmediate size > the layout size.

Even though it's a little weird that it has no invalid cases - it's looking for a very specific possible bug so it's good to keep it separate from other tests.

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
```diff
@@ -0,0 +1,505 @@
...
+    - Immediate data must be re-set after executeBundles.
+    - setImmediates in bundle does not leak to pass.
+    `
+  )
+  .params(u =>
+    u.combine('check', [
+      'pipeline_invalidated',
+      'immediates_invalidated',
+      'bundle_no_leak',
+      'pipeline_and_immediates_reset',
```
kainino0x: This feels like potentially a list of separate tests rather than cases of one test.

I would perhaps suggest splitting this test out into a separate PR because this one is large and this test is pretty distinct.

src/webgpu/api/validation/createPipelineLayout.spec.ts
kainino0x: I would also suggest making this a separate PR

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
shaoboyan091: > What happens if they have different types? I assume they just alias the memory with different views (like would happen if you rebind the pipeline without changing the immediates, and dispatch again).

Yes, that's exactly what I expected.

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
shaoboyan091: +1, `src/webgpu/api/validation/encoding/cmds/setImmediates.spec.ts` covers it.

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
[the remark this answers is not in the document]
shaoboyan091: Yes. `referenced_in_unused_function` covers it. Remove this scenario here.

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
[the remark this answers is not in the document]
shaoboyan091: Done.

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
shaoboyan091: Move this cases to another PR. And I'll address this comments there.

src/webgpu/api/validation/createPipelineLayout.spec.ts
shaoboyan091: (Seperate this to another PR)
But I don't think we could have a validation test for this easily. Because `setImmediates` allows overprovision, all we could do for default pipeline layouts is a control case.

How about cover this in operation test or you prefer a control case for it.

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
shaoboyan091: Done.

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
shaoboyan091: Yes. Removed this. Done.

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
shaoboyan091: Done. And it indeed caught a bug in implementation! Tint calculated whole immediate block struct size as the immediate block size. And it doesn't get rid of tail padding bytes so it triggers validation report wrong fail due to it. (report used more bytes than layout required)

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
shaoboyan091: Done.

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
shaoboyan091: Done.

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
shaoboyan091: Done.

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
shaoboyan091: Done.

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
shaoboyan091: Done.

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
shaoboyan091: Done.

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
shaoboyan091: Removed it. Done.

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
shaoboyan091: Done.

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
shaoboyan091: Done.

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
shaoboyan091: Done.

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
shaoboyan091: Done.

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
shaoboyan091: Correct. Remove `full` usage.

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
shaoboyan091: Done.

src/webgpu/api/validation/createPipelineLayout.spec.ts
shaoboyan091: Move this to another PR. Done.

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
shaoboyan091: Move this to another PR. Will address this comment there.

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
shaoboyan091: Removed this but has some thoughts.
Maybe we don't need `'pipeline_and_immediates_reset',`. I think for bundle we have pipeline invalidate and we have immediate states invalidate here. So this scenario is covered.

src/webgpu/api/validation/createPipelineLayout.spec.ts
shaoboyan091: Done.

Kangz: I meant, can we have a test that the default pipeline layout causes an error to be produced when compiling pipeline that use too many immediates?

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
kainino0x: I noticed this is still open, are you planning to still address it in this PR?

I did see you added `both_different` which uses one `var<immediate>` with each stage accessing a separate part, but even in that case both stages require both `data.a` and `data.b` to be initialized. So I think still need a case where each stage uses a different `var<immediate>`.

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
kainino0x: SGTM.

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
shaoboyan091: Thanks for pointing it out. I've updated the test cases.

src/webgpu/api/validation/createPipelineLayout.spec.ts
shaoboyan091: Resolved in another PR.

src/webgpu/api/validation/encoding/programmable/pipeline_immediate.spec.ts
```diff
@@ -216,6 +230,13 @@ g.test('required_slots_set')
         // Total size: 32. Layout size 32.
         setImmediates(0, 4);
         setImmediates(16, 16);
+      } else if (scenario === 'multiple_variables') {
+        // v1: struct S {u32} (size 4, align 4) -> offset 0
+        // v2: vec2<u32> (size 8, align 8) -> offset 8
```
kainino0x: I thought the two immediate blocks were treated totally independently and so v2 would be at offset 0, and whatever data is in the immediates would get aliased by the two stages with different types.

I was thinking it would work like:
```wgsl
struct S1 { a: u32, x: u32 };       // slots used: 1100 0000
struct S2 { a: f32, y: vec4<f32> }; // slots used: 1000 1111
// so the slots required to be set are:            1100 1111
```

If this is wrong then I'll need to understand a bit better.

shaoboyan091: First, I think the code comment here is incorrect. You understand is correct.  They should be treated independently.
And I think the proposed case is better than what I want to do(padding bytes in fragment overlap with required bytes in vertex). Thanks!

PR #4554, opened 2026-01-08
Add new builtins to maxInterStageShaderVariables test

greggman:
primitive_index, subgroup_id, subgroup_size were missing from the test.

note: Chromium doesn't currently count these. Bugs filed. Also, chromium doesn't currently count correctly for the pre-existing ones. Fixes in progress.

<hr>

### DIFF
--- a/src/webgpu/api/validation/capability_checks/limits/maxInterStageShaderVariables.spec.ts
+++ b/src/webgpu/api/validation/capability_checks/limits/maxInterStageShaderVariables.spec.ts
@@ -21,34 +21,35 @@ const kExtraItems = [
   'sample_mask_out', // special - see below
 ] as const;
 
-function getCombinations<T>(arr: readonly T[], sizes: number[]): T[][] {
-  const combos: T[][] = [];
-
-  function backtrack(start: number, path: T[], size: number) {
-    if (path.length === size) {
-      combos.push(path.slice());
-      return;
-    }
-
-    for (let i = start; i < arr.length; i++) {
-      path.push(arr[i]);
-      backtrack(i + 1, path, size);
-      path.pop();
-    }
+/**
+ * Generates every combination of size elements from array
+ * combinations([a, b, c], 2) generates [a, b], [a, c], [b, c]
+ */
+function* combinations<T>(
+  arr: readonly T[],
+  size: number,
+  start = 0,
+  path: T[] = []
+): Generator<T[]> {
+  if (path.length === size) {
+    yield [...path];
+    return;
   }
 
-  sizes.forEach(size => {
-    backtrack(0, [], size);
-  });
-
-  return combos;
+  for (let i = start; i < arr.length; i++) {
+    path.push(arr[i]);
+    yield* combinations(arr, size, i + 1, path);
+    path.pop();
+  }
 }
 
 const kTestItems = [...kItemsThatCountAgainstLimit, ...kExtraItems] as const;
 const kTestItemCombinations = [
   [], // no builtins case
-  ...getCombinations(kTestItems, [1, 2, 3]),
-  kTestItems, // all case
+  ...combinations(kTestItems, 1), // one builtin
+  ...combinations(kTestItems, 2), // 2 builtins
+  ...combinations(kTestItems, 3), // 3 builtins
+  kTestItems, // all builtins case
 ] as const;
 
 const requiresSubgroupsFeature = (items: Set<(typeof kTestItems)[number]>) =>


greggman:
Well yea, kind of. Your other PR fixed the test so it counted 1 for each builtin. That pointed out that chrome was only counting one. When I went into the spec and saw 3 others were missing I started adding them but with 6 it felt like it needed a bigger refactor. I guess I missed your newer one. Sorry about that. Which one do you prefer?

greggman:
Hi Erich, I just wanted to apologize. I totally missed that you had more PRs up for this issue. Feel free to close this PR.

src/webgpu/api/validation/render_pipeline/inter_stage.spec.ts
kainino0x: I think we still need the restructure in #4555 otherwise this test isn't hitting both the control case and the invalid case, because `numVariablesDelta` is the wrong way to parameterize.

It's good to add the extra builtins though. I think I should try to pull that into #4555.

src/webgpu/api/validation/capability_checks/limits/maxInterStageShaderVariables.spec.ts
```diff
@@ -1,75 +1,145 @@
...
+
+const kFragmentInputs = keysOf(kFragmentInputTypes);
+
+const kItemsThatCountAgainstLimit = ['point-list', ...kFragmentInputs] as const;
+
+const kExtraItems = [
+  'sample_mask_out', // special - see below
+] as const;
+
+function getCombinations<T>(arr: readonly T[], sizes: number[]): T[][] {
```
kainino0x: nit: This is neat but I found it really hard to understand. WDYT of something un-generalized but structurally the same, I think it's more understandable because it's more "typical" logic but I'm not sure:
```ts
const kTestItemCombinations = [
  ...(function* () {
    yield []; // 0 items
    for (let i1 = 0; i1 < kTestItems.length; ++i1) {
      yield [kTestItems[i1]]; // 1 item
      for (let i2 = i1 + 1; i2 < kTestItems.length; ++i2) {
        yield [kTestItems[i1], kTestItems[i2]]; // 2 items
        for (let i3 = i2 + 1; i3 < kTestItems.length; ++i3) {
          yield [kTestItems[i1], kTestItems[i2], kTestItems[i3]]; // 3 items
        }
      }
    }
  })(),
  kTestItems, // all items
];
```
(or instead of wrapping a generator, just build an array with push)

src/webgpu/api/validation/capability_checks/limits/maxInterStageShaderVariables.spec.ts
```diff
@@ -92,51 +162,59 @@ const limit = 'maxInterStageShaderVariables';
 export const { g, description } = makeLimitTestGroup(limit);
 
 g.test('createRenderPipeline,at_over')
```
kainino0x, line 165: Also now that I've read through this test, does it make the one in https://github.com/gpuweb/cts/pull/4555/changes totally redundant? 
The structure is pretty similar (it tests at/over the limit) but it tests multiple devices with different limit values.

src/webgpu/api/validation/capability_checks/limits/maxInterStageShaderVariables.spec.ts
greggman: I changed it to a simpler generator. wdyt?

src/webgpu/api/validation/capability_checks/limits/maxInterStageShaderVariables.spec.ts
greggman: yea, there are several of these redundant tests.

src/webgpu/api/validation/capability_checks/limits/maxInterStageShaderVariables.spec.ts
kainino0x: It's hard for me to say since I did already spend the time to figure out how the old one worked, but it does seem simpler, so LGTM

src/webgpu/api/validation/capability_checks/limits/maxInterStageShaderVariables.spec.ts
kainino0x: Thanks, I guess I remember talking about that but forgot. Perhaps before we've forgotten what's going on and are still confident it's redundant, we should start by removing that one?

kainino0x (approved):

src/webgpu/api/validation/capability_checks/limits/maxInterStageShaderVariables.spec.ts
greggman: yea, good idea.

kainino0x: I looked at this a bit. It's not as straightforward as I thought, inter_stage covers more stuff (inputs vs outputs). I can keep looking at it if you aren't yet.

kainino0x: Notes:
- Would add a vertex inputs test to `capability_checks`. Merger of:
  - `webgpu:api,validation,render_pipeline,inter_stage:max_variables_count,input`
  - `webgpu:api,validation,capability_checks,features,clip_distances:createRenderPipeline,at_over`
- Would expand the fragment outputs test in `capability_checks` from:
  - `webgpu:api,validation,render_pipeline,inter_stage:max_variables_count,output`
- Would migrate `webgpu:api,validation,render_pipeline,inter_stage:max_shader_variable_location` to `capability_checks`.
- I don't think we currently have test coverage of a case where both vertex and fragment are separately under their limits, but _would_ be over limit if all vertex outputs were counted as fragment inputs even if the fragment shader doesn't declare them. I think this is valid because the pipeline should elide the unused outputs?
- These tests probably _should_ be kept separate because one is making sure that the requested limit takes effect, and the other is making sure that the accounting is correct in lots of corner cases. It seems unlikely that an implementation would get things wrong in a way that requires them to be tested in combination. Actually kind of thinking the test should test _fewer_ combinations than it does now since it gets multiplied with so many `MaximumLimitValueTest` cases. This leads to some general thoughts about capability_checks,limits tests:
  - `atDefault` vs `underDefault` is testing for a very specific thing (correct defaulting) and shouldn't need to be combined with much of anything else.
  - Same for `overMaximum`, isn't it just testing that you don't get a device? So none of the cases do anything.
  - `betweenDefaultAndMaximum` and `atMaximum` could end up being the same as `atDefault` if the browser only exposes the default limit. If that happens do we skip those cases? Doesn't seem like we do but we should.

kainino0x: I just filed #4564, so between that an #4561 I think everything important here is covered. The dedup is not worth the effort I think.

PR #4558, opened 2026-01-12
Operation tests for setImmediates

shaoboyan091:
Implement operation tests for setImmediates in ComputePassEncoder, RenderPassEncoder, and RenderBundleEncoder.

- Add basic execution tests for scalar, vector, and struct types.
- Add tests for partial updates and multiple updates (using range verification).
- Add tests for pipeline switching (compatible and incompatible).
- Add tests for large data (maxImmediateSize) with range verification.
- Add tests for TypedArray arguments with offsets.
- Add tests for mixing render pass and bundle execution.

Issue: #

<hr>


src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -0,0 +1,902 @@
...
+    }
+
+    const pipeline = createPipeline(t, encoderType, wgslDecl, wgslUsage, immediateSize);
+
+    runAndCheck(
+      t,
+      encoderType,
+      pipeline,
+      encoder => {
+        encoder.setImmediates!(0, inputData.buffer, 0, inputData.buffer.byteLength);
```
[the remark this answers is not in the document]
shaoboyan091: SetImmediates accept both.

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -0,0 +1,979 @@
...
+        } else if (sType === 'i32') {
+          const val = -1000 - i;
+          inputData[i] = new Uint32Array(new Int32Array([val]).buffer)[0];
+          expected.push(inputData[i]);
+        } else if (sType === 'f32') {
+          const val = 1.5 + i;
+          inputData[i] = new Uint32Array(new Float32Array([val]).buffer)[0];
+          expected.push(inputData[i]);
+        }
+      }
```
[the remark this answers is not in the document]
shaoboyan091: SkipIf ensure this won't happens. But I think I don't add unreachable for it. Done.

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -0,0 +1,974 @@
...
+              // Map vIdx 0..3 to pixel centers.
+              // Uses a 4x1 texture.
+              // vIdx 0 -> pixel 0 -> x in [-1, -0.5]
+              // vIdx 1 -> pixel 1 -> x in [-0.5, 0]
+              let x = (f32(vIdx) + 0.5) / 2.0 - 1.0;
+              return vec4f(x, 0.0, 0.0, 1.0);
+            }
+          `,
+        }),
+        entryPoint: 'vs_main',
```
Kangz: entryPoint names are optional and can be defaulted here.

Kangz: And CS / FS

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -0,0 +1,974 @@
...
+        targets: [{ format: 'r32uint' }],
+      },
+      primitive: {
+        topology: 'point-list',
+      },
+    });
+  }
+}
+
+function executePass(
```
Kangz: Name is a bit weird, maybe `encodeForPassType`? "execute" makes it sound like there is a queue.submit in here.

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -0,0 +1,974 @@
...
+      pass.setPipeline(pipeline as GPUComputePipeline);
+      pass.setBindGroup(0, bindGroup, [0]); // Index 0
+      setImmediatesFn(pass);
+      pass.dispatchWorkgroups(1);
+    } else if (encoderType === 'render pass') {
+      const pass = encoder as GPURenderPassEncoder;
+      pass.setPipeline(pipeline as GPURenderPipeline);
+      pass.setBindGroup(0, bindGroup, [0]);
+      setImmediatesFn(pass);
+      pass.draw(1, 1, 0, 0); // Vertex 0 -> Pixel 0
```
Kangz:
```suggestion
      pass.draw(1); // Vertex 0 -> Pixel 0
```

Kangz: What does Vertex 0 -> Pixel 0 means here?

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -0,0 +1,974 @@
...
+}
+
+function runAndCheck(
+  t: AllFeaturesMaxLimitsGPUTest,
+  encoderType: ProgrammableEncoderType,
+  pipeline: GPURenderPipeline | GPUComputePipeline,
+  setImmediatesFn: (
+    encoder: GPUComputePassEncoder | GPURenderPassEncoder | GPURenderBundleEncoder
+  ) => void,
+  expectedValues: number[]
```
Kangz, line 233: Can we require that expectedValues is of size 0 here? Or should the size of outputBuffer be variable?

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -0,0 +1,974 @@
...
+  const outputBuffer = t.createBufferTracked({
+    size: 4 * 4, // 4 u32s
+    usage: GPUBufferUsage.STORAGE | GPUBufferUsage.COPY_SRC,
+  });
+  const indexUniformBuffer = t.createBufferTracked({
+    size: 256 * 4, // Enough for dynamic offsets
+    usage: GPUBufferUsage.UNIFORM,
+    mappedAtCreation: true,
+  });
+  // Initialize indices 0, 1, 2, 3 at offsets 0, 256, 512, 768
```
Kangz: Why do we have a dynamic offset uniform buffer that's only used with offset 0?

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -0,0 +1,974 @@
...
+      pass.setPipeline(pipeline as GPURenderPipeline);
+      pass.setBindGroup(0, bindGroup, [0]);
+      setImmediatesFn(pass);
+      pass.draw(1, 1, 0, 0); // Vertex 0 -> Pixel 0
+    } else if (encoderType === 'render bundle') {
+      const bundleEncoder = encoder as GPURenderBundleEncoder;
+      bundleEncoder.setPipeline(pipeline as GPURenderPipeline);
+      bundleEncoder.setBindGroup(0, bindGroup, [0]);
+      setImmediatesFn(bundleEncoder);
+      bundleEncoder.draw(1, 1, 0, 0);
```
Kangz: ditto

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -0,0 +1,974 @@
...
+      const vSize = vectorSize as number;
+
+      immediateSize = vSize * 4;
+      wgslDecl = `var<immediate> data: ${dataType};`;
+
+      let readCode = '';
+      for (let i = 0; i < vSize; i++) {
+        let valExpr = vSize === 1 ? 'data' : `data[${i}]`;
+        if (sType === 'i32' || sType === 'f32') {
+          valExpr = `bitcast<u32>(${valExpr})`;
```
Kangz: Can we do `bitcast<u32>(a u32)`? then we could bitcast unconditionally.

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -0,0 +1,974 @@
...
+        }
+        readCode += `output[${i}] = ${valExpr};`;
+      }
+
+      wgslUsage = `
+        @compute @workgroup_size(1) fn cs_main() {
```
Kangz: Shouldn't the `@compute` and `@fragment` parts be handled automatically in `createPipeline` such that only `readCode` needs to be passed to `createPipeline`? (also that code copies so maybe `copyCode`?)

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -0,0 +1,974 @@
...
+g.test('update_data')
+  .desc('Verify setImmediates updates data correctly within a pass, including partial updates.')
+  .params(u => u.combine('encoderType', kProgrammableEncoderTypes))
+  .fn(t => {
+    const { encoderType } = t.params;
+    t.skipIfStorageBuffersInFragmentStageNotAvailable(encoderType);
+    const immediateSize = 16;
+    const wgslDecl = 'var<immediate> data: vec4<u32>;';
+    const wgslUsage = `
+      @compute @workgroup_size(1) fn cs_main() {
```
Kangz: Ah I understand why there are firstVertex mentions earlier, did you mean to use it as an `outIndex`? Why did you choose not to do that, is it because of the emulation of baseVertex in D3D12?

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -0,0 +1,974 @@
...
+    `;
+
+    const pipeline = createPipeline(t, encoderType, wgslDecl, wgslUsage, immediateSize);
+
+    const outputBuffer = t.createBufferTracked({
+      size: 4 * 4 * 3, // 3 steps, 4 u32s each
+      usage: GPUBufferUsage.STORAGE | GPUBufferUsage.COPY_SRC,
+    });
+
+    const indexUniformBuffer = t.createBufferTracked({
```
Kangz: Maybe comment what this is used for as it was very unclear.

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -0,0 +1,974 @@
...
+      data: Uint32Array,
+      dstOffset: number = 0
+    ) => {
+      pass.setBindGroup(0, bindGroup, [stepIndex * 256]);
+      pass.setImmediates!(dstOffset, data, 0, data.length);
+
+      if (encoderType === 'compute pass') {
+        (pass as GPUComputePassEncoder).dispatchWorkgroups(1);
+      } else if (encoderType === 'render pass') {
+        (pass as GPURenderPassEncoder).draw(1, 1, 0, 0);
```
Kangz: ditto defaults here and everywhere else.

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -0,0 +1,974 @@
...
+  .desc(
+    `Verify immediate data is correctly set after switching pipelines.
+    - compatible: Both pipelines use the same immediateSize (same layout).
+    - incompatible: Pipelines use different immediateSize values (different layouts).
+    In both cases, immediates must be reset correctly between draws/dispatches.`
+  )
+  .params(u =>
+    u
+      .combine('encoderType', ['render pass', 'compute pass'] as const)
+      .combine('compatibility', ['compatible', 'incompatible'] as const)
```
Kangz: nit: `sameImmediateSize` maybe?

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -0,0 +1,974 @@
...
+    }
+
+    const pipelineA = createPipeline(t, encoderType, wgslDeclA, wgslUsageA, 16);
+    const pipelineB = createPipeline(t, encoderType, wgslDeclB, wgslUsageB, immediateSizeB);
+
+    const outputBuffer = t.createBufferTracked({
+      size: 32,
+      usage: GPUBufferUsage.STORAGE | GPUBufferUsage.COPY_SRC,
+    });
+    const indexUniformBuffer = t.createBufferTracked({
```
Kangz: Maybe we could split the creation of these "index" uniform buffers to a `createDynamicUniformForOutputIndex` or something like that?

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -0,0 +1,974 @@
...
+      usage: GPUBufferUsage.UNIFORM,
+      mappedAtCreation: true,
+    });
+    const indexData = new Uint32Array(indexUniformBuffer.getMappedRange());
+    indexData[0] = 0;
+    indexData[256 / 4] = 1;
+    indexUniformBuffer.unmap();
+
+    const bindGroupA = t.device.createBindGroup({
+      layout: pipelineA.getBindGroupLayout(0),
```
Kangz: It would be ideal to use the same layout, so that the bindgroup can be set once and then not changed, this way we can catch issues where Vulkan doesn't immediate VkDesrciptorSets that were bound prior. Right now it's not sure that it does that because the bindgroup is set again.

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -0,0 +1,974 @@
...
+        { binding: 0, resource: { buffer: outputBuffer } },
+        { binding: 1, resource: { buffer: indexUniformBuffer, size: 4 } },
+      ],
+    });
+
+    const commandEncoder = t.device.createCommandEncoder();
+
+    if (compatibility === 'compatible') {
+      executePass(t, encoderType, commandEncoder, enc => {
+        if (encoderType === 'compute pass') {
```
Kangz: There is no better way to do the polymorphism? :(

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -0,0 +1,974 @@
...
+      layout: pipelineB.getBindGroupLayout(0),
+      entries: [
+        { binding: 0, resource: { buffer: outputBuffer } },
+        { binding: 1, resource: { buffer: indexUniformBuffer, size: 4 } },
+      ],
+    });
+
+    const commandEncoder = t.device.createCommandEncoder();
+
+    if (compatibility === 'compatible') {
```
Kangz: I don't think that we need to change the immdiates between the pipelines for this test to be useful, so we could remove this branch here, always set only [1, 2, 3, 4] and at the end when generating the expected array, check `compatible`. It would cut down on the code a lot.

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -0,0 +1,974 @@
...
+      }
+    `;
+
+    const pipeline = createPipeline(t, encoderType, wgslDecl, wgslUsage, maxImmediateSize);
+
+    const outputBuffer = t.createBufferTracked({
+      size: 8,
+      usage: GPUBufferUsage.STORAGE | GPUBufferUsage.COPY_SRC,
+    });
+    const indexUniformBuffer = t.makeBufferWithContents(
```
Kangz: The indexUniformBuffer isn't actually needed here.

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -0,0 +1,974 @@
...
+    t.device.queue.submit([commandEncoder.finish()]);
+    t.expectGPUBufferValuesEqual(outputBuffer, new Uint32Array([0xdeadbeef, 0xcafebabe]));
+  });
+
+g.test('typed_array_arguments')
+  .desc('Verify srcOffset and srcSize are in elements for TypedArrays.')
+  .params(u =>
+    u
+      .combine('typedArray', kTypedArrayBufferViewKeys)
+      .combine('encoderType', kProgrammableEncoderTypes)
```
Kangz, line 596: This test could use a parameter for the offset (including undefined) and for the size (including undefined) because there is some complexity as you saw when you made the Blink bindings so we need good coverage.

I suggest doing something like having known `data` that's [1, 2, 3, 4 ...] `setImmediate` that, then `setImmediate` some other data with the provided offset + size, and programmatically inserting that in the original `data` so we can check that what the shader read is the updated `data`.

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -204,28 +223,17 @@ function runAndCheck(
   });
 
   const commandEncoder = t.device.createCommandEncoder();
-  executePass(t, encoderType, commandEncoder, encoder => {
+  encodeForPassType(t, encoderType, commandEncoder, encoder => {
     if (encoderType === 'compute pass') {
```
Kangz: nit: we could also have a helper for setting the pipeline with the correct encoder type.

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -115,7 +116,35 @@ function createPipeline(
   }
 }
 
-function executePass(
+/** Dispatch or draw based on encoder type. firstVertex selects the output pixel for render passes. */
+function dispatchOrDraw(
+  encoderType: ProgrammableEncoderType,
+  encoder: GPUComputePassEncoder | GPURenderPassEncoder | GPURenderBundleEncoder,
+  firstVertex: number = 0
```
Kangz: Why do we need firstVertex? The only place we use it as non-0, the output index uniform buffer is changed as well, so we could use that output index for the pixel position as well no?

Kangz: It would change [1] as well.

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -92,20 +97,16 @@ function createPipeline(
         module: t.device.createShaderModule({
           code: `
             @vertex fn vs_main(@builtin(vertex_index) vIdx: u32) -> @builtin(position) vec4f {
-              // Map vIdx 0..3 to pixel centers.
-              // Uses a 4x1 texture.
-              // vIdx 0 -> pixel 0 -> x in [-1, -0.5]
-              // vIdx 1 -> pixel 1 -> x in [-0.5, 0]
+              // Map vIdx 0..3 to pixel centers in a 4x1 render target.
```
Kangz: [1]

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -290,24 +290,13 @@ g.test('basic_execution')
...
-      wgslUsage = `
-        @compute @workgroup_size(1) fn cs_main() {
-          ${readCode}
-        }
-        @fragment fn fs_main() -> @location(0) vec4u {
-          ${readCode}
-          return vec4u(0);
-        }
-      `;
+      copyCode = readCode;
```
Kangz: Why not append to `copyCode` directly instead of having an intermediate variable?

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
Kangz: And use it in other places as well.

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -115,7 +116,35 @@ function createPipeline(
...
+  firstVertex: number = 0
+) {
+  if (encoderType === 'compute pass') {
+    (encoder as GPUComputePassEncoder).dispatchWorkgroups(1);
+  } else {
+    (encoder as GPURenderPassEncoder | GPURenderBundleEncoder).draw(1, 1, firstVertex, 0);
+  }
+}
+
+/** Create a uniform buffer with output indices at 256-byte aligned offsets for dynamic binding. */
```
Kangz: Can we explain why it is used instead of other methods (works in all stages, doesn't use immediates which is what is under testing) it will make it easier for future readers.

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
Kangz: The way it is done now with `getBindGroupLayout` shouldn't work if the pipelines are different because bindgroups created with `getBindGroupLayout` are only allowed to be used with the pipeline the came from. We need to explicitly create the BindGroupLayout.

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -0,0 +1,880 @@
...
+      .beginSubcases()
+      .combine('dataOffset', [undefined, 0, 2])
+      .combine('dataSize', [undefined, 2])
+      .filter(t => {
+        // WebGPU requires the byte size of the provided data to `setImmediates` to be a multiple of 4.
+        const elementSize = kTypedArrayBufferViews[t.typedArray].BYTES_PER_ELEMENT;
+        const actualDataOffset = t.dataOffset ?? 0;
+
+        // Let's test a larger size to accommodate BigUint64Array/BigInt64Array matrices more easily
+        const maxElementsIn64Bytes = 64 / elementSize;
```
Kangz: nit: maybe have a named constant and explain that the test will use 64bytes of immediates.

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -0,0 +1,880 @@
...
+      .combine('encoderType', kProgrammableEncoderTypes)
+      .beginSubcases()
+      .combine('dataOffset', [undefined, 0, 2])
+      .combine('dataSize', [undefined, 2])
+      .filter(t => {
+        // WebGPU requires the byte size of the provided data to `setImmediates` to be a multiple of 4.
+        const elementSize = kTypedArrayBufferViews[t.typedArray].BYTES_PER_ELEMENT;
+        const actualDataOffset = t.dataOffset ?? 0;
+
+        // Let's test a larger size to accommodate BigUint64Array/BigInt64Array matrices more easily
```
Kangz:
```suggestion
        // Let's test a large size to accommodate BigUint64Array/BigInt64Array
```

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -0,0 +1,880 @@
...
+  });
+
+g.test('typed_array_arguments')
+  .desc('Verify dataOffset and dataSize arguments work correctly for all TypedArray types.')
+  .params(u =>
+    u
+      .combine('typedArray', kTypedArrayBufferViewKeys)
+      .combine('encoderType', kProgrammableEncoderTypes)
+      .beginSubcases()
+      .combine('dataOffset', [undefined, 0, 2])
```
Kangz: With these params, I don't think that the filter below is ever false?

Also why 2 and not other offsets/sizes? For uint8_t it will be an issue as a size of 2 is not a multiple. Instead can we make the dataSize depend on the BYTES_PER_ELEMENT to be either 1 element or 4 bytes (whichever is smaller)?

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
Kangz: But also this filter seems like it might not be needed, instead we could have an ASSERT in the test body.

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -0,0 +1,880 @@
...
+      output[13] = data.m3.y;
+      output[14] = data.m3.z;
+      output[15] = data.m3.w;
+    `;
+    const pipeline = createPipeline(t, encoderType, wgslDecl, copyCode, immediateSize);
+
+    const actualDataOffset = dataOffset ?? 0;
+    // We want to write `dataSize` elements. If dataSize is undefined, it defaults to:
+    // array.length - dataOffset
+    const maxElementsIn64Bytes = 64 / elementSize;
```
Kangz: nit: name the constant `64` here and maybe put it next to the WGSL code as it needs to match it.

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -0,0 +1,880 @@
...
+        { binding: 0, resource: { buffer: outputBuffer } },
+        { binding: 1, resource: { buffer: indexUniformBuffer, size: 4 } },
+      ],
+    });
+
+    encodeForPassType(t, encoderType, commandEncoder, enc => {
+      setPipeline(encoderType, enc, pipeline);
+      enc.setBindGroup(0, bindGroup, [0]);
+
+      // Initialize pipeline storage to a clear pattern mapping to up to 64 bytes
```
Kangz:
```suggestion
      // Initialize immediates to a clear pattern mapping to up to 64 bytes
```

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -0,0 +1,880 @@
...
+        }
+      } else {
+        enc.setImmediates!(0, arr, dataOffset, dataSize);
+      }
+
+      dispatchOrDraw(encoderType, enc);
+    });
+    t.device.queue.submit([commandEncoder.finish()]);
+
+    const expected = new Uint32Array(16);
```
Kangz: Let's just reuse clearsData?

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -0,0 +1,880 @@
...
+    encodeForPassType(t, encoderType, commandEncoder, enc => {
+      setPipeline(encoderType, enc, pipeline);
+      enc.setBindGroup(0, bindGroup, [0]);
+
+      // Initialize pipeline storage to a clear pattern mapping to up to 64 bytes
+      const clearData = new Uint32Array(16);
+      for (let i = 0; i < 16; i++) clearData[i] = 0xaaaaaaaa + i * 0x11111111;
+      enc.setImmediates!(0, clearData);
+
+      if (dataSize === undefined) {
```
Kangz: I thought that passing undefined to the argument would cause the default values to be used, is that not the case and we need to make calls with different numbers of parameters?

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -0,0 +1,880 @@
...
+
+      dispatchOrDraw(encoderType, enc);
+    });
+    t.device.queue.submit([commandEncoder.finish()]);
+
+    const expected = new Uint32Array(16);
+    for (let i = 0; i < 16; i++) expected[i] = 0xaaaaaaaa + i * 0x11111111;
+    const expectedView = new Uint8Array(expected.buffer);
+
+    for (let i = 0; i < actualDataSize; i++) {
```
Kangz: Instead of replicating the conversion, can we copy from `view` into the correct spot in expected ?

Kangz: There is a memcpy utility somewhere in the CTS IIRC

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
Kangz: It will need to be hosted out of the `encodeForPassType` but that's ok to avoid copying the same filler pattern code twice.

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -0,0 +1,880 @@
...
+    const pipeline = createPipeline(t, encoderType, wgslDecl, copyCode, immediateSize);
+
+    const actualDataOffset = dataOffset ?? 0;
+    // We want to write `dataSize` elements. If dataSize is undefined, it defaults to:
+    // array.length - dataOffset
+    const maxElementsIn64Bytes = 64 / elementSize;
+    const actualDataSize = dataSize ?? maxElementsIn64Bytes - actualDataOffset;
+
+    // Create a typed array buffer. If dataSize is undefined, we shouldn't add padding
+    // because setImmediates uses the rest of the array, which would exceed our intended size or break the % 4 rule.
```
Kangz: The padding of 0 or 4 doesn't improve the % 4 rule because both of these numbers are multiples of 4. We need to make the actualDataSize smaller when BYTES_PER_ELEMENT is no a multiple of 4, so that the `dataSize == undefined` case passes validation. And then we can add paddingElements on top of that.

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
shaoboyan091: BYTES_PER_ELEMENT  is a good idea, I think I could try param generator.

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
shaoboyan091: Yes. That's the main reason. It uses internal immediates and mix things up.

shaoboyan091:
I think I've resolved comments. PTAL, thanks !

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
Kangz: Did your AI start rewriting this test for some reason?

In any case, `base` is no longer used here.

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -641,31 +646,39 @@ g.test('typed_array_arguments')
...
+    );
+
+    // When dataSize is explicit, add padding elements to verify setImmediates
+    // respects the dataSize boundary and doesn't read beyond it.
+    // When dataSize is undefined, no padding since setImmediates reads to the end of the array.
+    const paddingElements = dataSize !== undefined ? 4 : 0;
     const arr = new Ctor(actualDataOffset + actualDataSize + paddingElements);
     const view = new DataView(arr.buffer);
 
+    // Fill the data region with a recognizable byte pattern.
```
Kangz, line 667: This could be a single loop over `actualDataSize + elementSize`, or better over the byte size of `view`

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
shaoboyan091: By checking the commit message, I think it do this due to a new rename and no return? However, I think we don't need such base because outIndex is always 0.

Kangz: After you get some LLM to do code changes, can you make sure to do a simplification pass? Models tend to do  more code than necessary, and any code is an ongoing cost so we need to minimize it.

Kangz (approved):

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
shaoboyan091: Yes, I think that's the truth.

src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
```diff
@@ -0,0 +1,885 @@
...
+      this.skip('Immediate data not supported');
+      return;
+    }
+  }
+
+  skipIfStorageBuffersInFragmentStageNotAvailable(encoderType: ProgrammableEncoderType) {
+    if (!this.isCompatibility) {
+      return;
+    }
+    const needsStorageBuffersInFragmentStage =
```
greggman, line 36: If I understand correctly, this test is skipped on devices that don't support storage buffers in fragment shaders. But, those device do support immediates meaning this test fails to test immediates in fragment shaders on those device.

Should the test be refactored not to use storage buffers in fragment shaders? Many other tests were refactored to do this.

shaoboyan091: Fair point. I think we need to refactor this to cover more devices. Is there anyone working a uniform refactor for all test cases or we could do it seperately.

greggman: I think I refactored all exsiting tests to not require storage buffers in fragment shaders (except for the tests that are specifically testing storage buffers in fragment shaders 😛) If there are any left I'm happy to fix them.

PR #4560, opened 2026-01-14
Allow tests to extend endTestScope timeout

haoxli:
Added a per-test endTestScope timeout function so heavy cases can extend the time without slowing down the rest of the suite, and used it to give the 64K timestamp-query stress test run extra headroom, avoiding flakes while leaving the default 5000ms limit untouched elsewhere.

Issue: #

<hr>

### DIFF
--- a/src/webgpu/api/operation/render_pass/transient_attachment.spec.ts
+++ b/src/webgpu/api/operation/render_pass/transient_attachment.spec.ts
@@ -1,0 +1,94 @@
+export const description = `API Operation Tests for transient attachment in render passes.`;
+
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { AllFeaturesMaxLimitsGPUTest } from '../../../gpu_test.js';
+
+const kSize = 4;
+
+export const g = makeTestGroup(AllFeaturesMaxLimitsGPUTest);
+
+g.test('increasing_attachments_count')
+  .desc(
+    `
+    Use multiple render passes with increasing amounts of transient attachments.
+  `
+  )
+  .fn(t => {
+    // MAINTENANCE_TODO(#4509): Remove this when TRANSIENT_ATTACHMENT is added to the WebGPU spec.
+    t.skipIfTransientAttachmentNotSupported();
+
+    const maxAttachments = t.device.limits.maxColorAttachments;
+    const encoder = t.device.createCommandEncoder();
+
+    for (let count = 1; count <= maxAttachments; count++) {
+      const colorAttachments: GPURenderPassColorAttachment[] = [];
+
+      for (let i = 0; i < count; i++) {
+        const texture = t.createTextureTracked({
+          size: [kSize, kSize, 1],
+          format: 'rgba8unorm',
+          usage: GPUTextureUsage.RENDER_ATTACHMENT | GPUTextureUsage.TRANSIENT_ATTACHMENT,
+        });
+
+        colorAttachments.push({
+          view: texture.createView(),
+          clearValue: { r: 0.0, g: 0.0, b: 0.0, a: 0.0 },
+          loadOp: 'clear',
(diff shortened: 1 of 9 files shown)


src/webgpu/api/operation/command_buffer/queries/timestampQuery.spec.ts
```diff
@@ -40,6 +40,11 @@ and prevent pages from running.
 
     t.skipIfDeviceDoesNotHaveFeature('timestamp-query');
 
+    if (numQuerySets === 65536) {
+      // Allow extra time for massive timestamp query allocation/cleanup churn.
+      t.setEndTestScopeTimeout(10000);
```
kainino0x: I'm not sure if I want to make this part of the "official" interface of the test fixture. It seems a bit fragile or easy to misuse because it's very specific to the test why this might need to be done.

Would it work to wait inside the test by just doing `await t.queue.onSubmittedWorkDone()` at the end before returning from the test function?

haoxli: I try it not work for this timeout case. This failure was not because the test ended before the commands were completed,  but rather because it didn't have enough time to finish the tests  (includes waiting submitted commands done + error scop popping + celanup). I think what we need here is a mechanism to increase the execution time of cases that require long execution times.

To avoid exposing it as a public interface and thus causing misuse, how about make the framework to auto-detect slow tests  and extend timeouts automatically, if the default timeout in endTestScope is hit, run the case again with twice the default timeout. However, this approach increases the overall runtime of the test suite if a network-related timeout occurs.

Another idea is whether we can move this case into the stress test  `src/stress/queries/timestamps.spec.ts`. I see there are some stress tests for render/compute pass are unimplemented, It would be reasonable to set a separate timeout for them.

kainino0x: Ah... it occurs to me that the test fixture already does `onSubmittedWorkDone` at the end of tests, so I should have realized that wouldn't work.

I was thinking that if the `popErrorScope` timeout worked, `onSubmittedWorkDone` would also work. But maybe not anymore because of work in Dawn to make error scopes asynchronous?

Anyway, what about just wrapping the test body in `t.device.pushErrorScope('validation')`/`await t.device.popErrorScope()`? If error scopes are able to capture the slowness at the test-fixture level, they should work at the test level too.

> To avoid exposing it as a public interface and thus causing misuse, how about make the framework to auto-detect slow tests and extend timeouts automatically, if the default timeout in endTestScope is hit, run the case again with twice the default timeout.

IMO that would add too much complexity to the CTS _and_ would probably interact poorly with the browsers' test harnesses which have their own timeout systems. My objection to putting it in the public interface is not _that_ strong :)

> Another idea is whether we can move this case into the stress test `src/stress/queries/timestamps.spec.ts`. I see there are some stress tests for render/compute pass are unimplemented, It would be reasonable to set a separate timeout for them.

I thought about suggesting moving this to `stress`, but right now I don't think anyone is running those tests so I don't want to lose the coverage. If we can make it run reliably then I'd like to keep it in the conformance section.

haoxli: I'm OOO this week and will update it next week.

haoxli: > Anyway, what about just wrapping the test body in t.device.pushErrorScope('validation')/await t.device.popErrorScope()? If error scopes are able to capture the slowness at the test-fixture level, they should work at the test level too

This works. The test level `pushErrorScope` and `popErrorScope` can complete immediately, it won’t trigger that timeout.

kainino0x: Great!

PR #4562, opened 2026-01-15
Skip invalid subgroupBroadcast tests

### LINKED ISSUE #4532: SubgroupBroadcast test bug for subgroup sizes < 16
Test case `webgpu:shader,execution,expression,call,builtin,subgroupBroadcast:compute,all_active:wgSize=[3,3,3]` produces a thread group with 3*3*3=27 threads. Test subcase 3 attempts to subgroupBroadcast from subgroup lane 3. When subgroup size is 4 or 8, the last subgroup to execute will not have a thread executing at subgroup index 3.

Subgroup size 4 will have:
* 0, 1, 2, 3
* 4, 5, 6, 7
* 8, 9, 10, 11
* 12, 13, 14, 15
* 16, 17, 18, 19
* 20, 21, 22, 23
* 24, 25, 26

Subgroup size 8 will have:
* 0 - 7
* 8 - 15
* 16 - 23
* 24, 25, 26

It's not until subgroup size 16 where all active subgroups include an active lane at index 3:
* 0 - 15
* 16 - 26

alan-baker:
Fixes #4532

* Skip tests in `compute,active_all` and `compute,split` if the broadcasted ID is not guaranteed to appear in all subgroups
  * For subgroup size 4 this skips 4 subtests (ID=3 for [3,3,3], [15,3,3], [3,15,3], and [3,3,15] sizes)
  * For subgroup size 8 this skips 1 subtest (ID=3 for [3,3,3] size)

Issue: #

<hr>

### DIFF
--- a/src/webgpu/shader/execution/expression/call/builtin/subgroupBroadcast.spec.ts
+++ b/src/webgpu/shader/execution/expression/call/builtin/subgroupBroadcast.spec.ts
@@ -273,6 +273,8 @@ fn main(@builtin(subgroup_invocation_id) id : u32,
     t.expectGPUBufferValuesEqual(outputBuffer, new Uint32Array(expect));
   });
 
+const kUnattemptedBroadcast = 555 << 16;
+
 /**
  * Checks the results of broadcast in compute shaders.
  *
@@ -280,6 +282,8 @@ fn main(@builtin(subgroup_invocation_id) id : u32,
  *                 * first half is subgroup_invocation_id
  *                 * second half is subgroup_size
  * @param output An array uint32s containing the broadcast results
+ *               If the broadcast was not attempted due to generating a dynamic error,
+ *               the value will be kUnattemptedWrite | ballot-based subgroup size
  * @param numInvs The number of invocations
  * @param broadcast The broadcast invocation (or 'first' to indicate the lowest active)
  * @param filter A functor indicating whether the invocation participates in the broadcast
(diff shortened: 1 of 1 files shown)


src/webgpu/shader/execution/expression/call/builtin/subgroupBroadcast.spec.ts
```diff
@@ -358,6 +358,16 @@ g.test('compute,all_active')
     t.skipIfDeviceDoesNotHaveFeature('subgroups' as GPUFeatureName);
     const wgThreads = t.params.wgSize[0] * t.params.wgSize[1] * t.params.wgSize[2];
 
+    interface SubgroupProperties extends GPUAdapterInfo {
```
dneto0: This solution works if we assume a somewhat greedy packing of invocations into subgroups.
It excludes the case where a greedy packing will produce at least one non-full subgroup.

But there is no reliable mapping between local invocation ID and subgroup IDs.  So this still allows bad cases.

I think a more correct fix would be to have the shader itself count the number of active IDs, and then only do the broadcast when the selected ID is present.  Alternately, use a ballot (in case we're concerned about holes in the subgroup_ids.)  Of course the output has to encode the fact of whether the broadcast was attempted or not.

dneto0 (changes requested):
I don't think this quite fixes it. suggestion below.

src/webgpu/shader/execution/expression/call/builtin/subgroupBroadcast.spec.ts
dneto0: With the suggested changes, we'd be able to expand coverage of tested IDs, e.g. trying the upper ranges of subgroup ids: 7, 15, 31, 63, 127, if the runtime is still reasonable.

dneto0: I don't think the spec says that if subgroup_invocation_id space has no holes.
That is, if subgroup_invocation_id k is active then all lower-numbered IDs also exist (at the start of the shader at least).   It's not quite derivable from the assertions on the density of subgroup_invocation_id and subgroup_id.

alan-baker: We confirmed in compute shaders that the subgroup_invocation_id is densely packed. Changed to use a ballot-based size in the shader instead of coarse skips. Verifies the unattempted broadcast results.

alan-baker: Also tests more ids for larger subgroups since they can be skipped appropriately.

PR #4567, opened 2026-01-17
Consider tier1 formats language feature in RO/RW storage texture test

andyleiserson:
Changes the `readonly_and_readwrite_storage_textures` test to consider the tier1 formats language feature.

Issue: none

<hr>

### DIFF
--- a/src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
+++ b/src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts
@@ -1,0 +1,885 @@
+export const description = `
+Operation tests for immediate data usage in RenderPassEncoder, ComputePassEncoder, and RenderBundleEncoder.
+`;
+
+import { makeTestGroup } from '../../../../../common/framework/test_group.js';
+import { getGPU } from '../../../../../common/util/navigator_gpu.js';
+import {
+  assert,
+  kTypedArrayBufferViews,
+  kTypedArrayBufferViewKeys,
+  memcpy,
+  supportsImmediateData,
+  unreachable,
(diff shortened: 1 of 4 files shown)


src/webgpu/shader/validation/extension/readonly_and_readwrite_storage_textures.spec.ts
```diff
@@ -28,6 +31,11 @@ g.test('var_decl')
   .fn(t => {
     const { type, format, access } = t.params;
 
+    const requiredFeature = getRequiredFeatureForTextureFormat(format);
+    if (requiredFeature) {
+      t.skipIfDeviceDoesNotHaveFeature(requiredFeature);
```
kainino0x: Hm........ in theory, the correct thing to do here would be to check `wgslLanguageFeatures` and not the device features. Because:
1. It's not just the new formats that should be skipped when the browser doesn't have support, it's also the existing formats that were newly-added to WGSL.
1. We should be running this test if the browser (compiler) supports the format, even if the device doesn't.

But we didn't add a wgsl language feature! Filed https://github.com/gpuweb/gpuweb/issues/5524

src/webgpu/shader/validation/extension/readonly_and_readwrite_storage_textures.spec.ts
kainino0x:
```suggestion
    `Checks that the read and read_write access modes are only allowed with the language feature present

    TODO(https://github.com/gpuweb/gpuweb/issues/5524): Fix this to check wgslLanguageFeatures
    of the device features.
    `
```

PR #4570, opened 2026-01-20
Add mixed swizzle and indexing tests

sudonatalie:
Check that a valid mix of swizzles and index works and that an invalid OOB case fails.

<hr>


src/webgpu/shader/validation/expression/access/vector.spec.ts
```diff
@@ -358,6 +358,10 @@ const kAbstractCases = {
     ok: (width: number) => width > 3,
   },
 
+  // mixed swizzle and indexing
+  xy_0: { wgsl: 'const r = V.xy[0];', result_width: 1, ok: true },
+  xy_3: { wgsl: 'const r = V.xy[3];', ok: false },
```
jrprice, line 367: I think this passes because `V` is not a memory view, so it's not hitting the SwizzleView case.

sudonatalie: Thanks, added 2 more tests that are on variables, and confirmed that that test fails on my old patchset and passes on my new one.

jrprice (approved):

PR #4582, opened 2026-02-03
Modify video import tests

alan-baker:
Bug: https://crbug.com/dawn/481071812

* Use the videoFrame's visible rectangle for sub-rectangle
* Change the sample locations slightly

Issue: #

<hr>

### DIFF
--- a/src/webgpu/web_platform/external_texture/video.spec.ts
+++ b/src/webgpu/web_platform/external_texture/video.spec.ts
@@ -405,9 +405,10 @@ parameters are present.
       // All tested videos are derived from an image showing yellow, red, blue or green in each
       // quadrant. In this test we crop the video to each quadrant and check that desired color
       // is sampled from each corner of the cropped image.
-      // visible rect clip applies on raw decoded frame, which defines based on video frame coded size.
-      const srcVideoHeight = source.codedHeight;
-      const srcVideoWidth = source.codedWidth;
+      // visible rect clip applies on raw decoded frame, which defines based on video frame visible size.
+      const visibleRect = source.visibleRect ?? new DOMRectReadOnly();
+      const srcVideoHeight = visibleRect.height;
+      const srcVideoWidth = visibleRect.width;
 
       const srcColorSpace = kVideoInfo[videoName].colorSpace;
       const presentColors = kVideoExpectedColors[srcColorSpace][dstColorSpace];
@@ -419,14 +420,19 @@ parameters are present.
       const cropParams = [
         // Top left
         {
-          subRect: { x: 0, y: 0, width: srcVideoWidth / 2, height: srcVideoHeight / 2 },
+          subRect: {
+            x: visibleRect.x,
+            y: visibleRect.y,
+            width: srcVideoWidth / 2,
+            height: srcVideoHeight / 2,
+          },
           color: convertToUnorm8(presentColors[expect.topLeftColor]),
         },
         // Top right
         {
           subRect: {
-            x: srcVideoWidth / 2,
-            y: 0,
+            x: visibleRect.x + srcVideoWidth / 2,
+            y: visibleRect.y,
             width: srcVideoWidth / 2,
             height: srcVideoHeight / 2,
           },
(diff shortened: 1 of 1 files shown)


src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -405,9 +405,10 @@ parameters are present.
       // All tested videos are derived from an image showing yellow, red, blue or green in each
       // quadrant. In this test we crop the video to each quadrant and check that desired color
       // is sampled from each corner of the cropped image.
-      // visible rect clip applies on raw decoded frame, which defines based on video frame coded size.
-      const srcVideoHeight = source.codedHeight;
-      const srcVideoWidth = source.codedWidth;
+      // visible rect clip applies on raw decoded frame, which defines based on video frame visible size.
+      const visibleRect = source.visibleRect ?? new DOMRectReadOnly();
```
dalecurtis, line 409: I think visibleRect should always be there, but using the coded size when missing should be correct.

src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -492,10 +498,10 @@ parameters are present.
         // For validation, we sample a few pixels away from the edges to avoid compression
         // artifacts.
         ttu.expectSinglePixelComparisonsAreOkInTexture(t, { texture: colorAttachment }, [
-          { coord: { x: kWidth * 0.1, y: kHeight * 0.1 }, exp: cropParam.color },
-          { coord: { x: kWidth * 0.9, y: kHeight * 0.1 }, exp: cropParam.color },
-          { coord: { x: kWidth * 0.1, y: kHeight * 0.9 }, exp: cropParam.color },
-          { coord: { x: kWidth * 0.9, y: kHeight * 0.9 }, exp: cropParam.color },
+          { coord: { x: kWidth * 0.2, y: kHeight * 0.2 }, exp: cropParam.color },
```
dalecurtis, line 501: Did you separately decide to tweak these values or was it necessary somehow?

src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -405,9 +405,10 @@ parameters are present.
       // All tested videos are derived from an image showing yellow, red, blue or green in each
       // quadrant. In this test we crop the video to each quadrant and check that desired color
       // is sampled from each corner of the cropped image.
-      // visible rect clip applies on raw decoded frame, which defines based on video frame coded size.
-      const srcVideoHeight = source.codedHeight;
-      const srcVideoWidth = source.codedWidth;
+      // visible rect clip applies on raw decoded frame, which defines based on video frame visible size.
+      const visibleRect = source.visibleRect ?? new DOMRectReadOnly();
+      const srcVideoHeight = visibleRect.height;
```
dalecurtis, line 410: FWIW, I think it's more legible to inline these, but up to you.

src/webgpu/web_platform/external_texture/video.spec.ts
alan-baker: I had one case where there ended being 2 ULP error in the color. It was definitely still the right quadrant's color, but I wanted to avoid the problem. It happened on one sample (bottom right) of one subrectangle (top right) of one video (the rotated 90 bt601). So I just tried to make it a bit simpler to match.

src/webgpu/web_platform/external_texture/video.spec.ts
alan-baker: I'm brand new to most of these APIs so I was just trying to satisfy the typescript compile error. I figured it's always there which is why it uses a default value, but I'll look into using coded size instead.

kainino0x: If we think it'll always be available, we can just do `source.visibleRect!` to tell typescript to assume it's non-nullish, and let it crash if it's not there for some reason.

dalecurtis: I'm not sure how other browsers handle it, but if you can get away with expecting it always it'd be simpler.

src/webgpu/web_platform/external_texture/video.spec.ts
dalecurtis: Ack. Given the sizes used, I wasn't expecting much a change here, I could see the colors being slightly off relative to where the coded size was being sampled previously.

PR #4589, opened 2026-02-10
Add importExternalTexture test for video frame display size diff with coded size

haoxli:
Validation error was thrown if the video frame's dispaly size was different with its coded size in importExternalTexture, because it used the visible size for copySize. This issue had been fixed by updating the copySize to reflect the coded size rather than the visible size, and add a test for this case.

Issue: https://crbug.com/471021591

<hr>

### DIFF
--- a/src/webgpu/api/validation/pipeline/immediates.spec.ts
+++ b/src/webgpu/api/validation/pipeline/immediates.spec.ts
@@ -1,0 +1,149 @@
+export const description = `
+Pipeline creation validation tests for immediate data size mismatches.
+
+Validates that creating a pipeline fails if the shader uses immediate data
+larger than the immediateSize specified in the pipeline layout, or larger than
+maxImmediateSize if layout is 'auto'.
+`;
+
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { getGPU } from '../../../../common/util/navigator_gpu.js';
+import { assert, range, supportsImmediateData } from '../../../../common/util/util.js';
+import { AllFeaturesMaxLimitsGPUTest } from '../../../gpu_test.js';
+import * as vtu from '../validation_test_utils.js';
+
+export const g = makeTestGroup(AllFeaturesMaxLimitsGPUTest);
+
+/**
+ * Generate shader code for a given stage with the specified immediate data size.
+ * If size is 0, the shader has no immediate data.
+ */
+function makeShaderCode(size: number, stage: 'compute' | 'vertex' | 'fragment'): string {
+  if (size === 0) {
+    switch (stage) {
+      case 'compute':
+        return `@compute @workgroup_size(1) fn main_compute() {}`;
+      case 'vertex':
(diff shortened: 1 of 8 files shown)


src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -381,6 +381,142 @@ Tests that we can import an VideoFrame with non-YUV pixel format into a GPUExter
...
+      default:
+        unreachable();
+    }
+
+    const frameInit: VideoFrameBufferInit = {
+      format: 'RGBA',
+      codedWidth: kWidth,
+      codedHeight: kHeight,
+      displayWidth,
+      displayHeight,
```
kainino0x: We should also add a test that `textureDimensions` returns the correct value for textures like this. It might even make sense to make it part of the same test (or at least sharing much of the same code).

I think I'd like to go ahead and add this now, but if not, please add a `TODO(#bugnumber)`.

src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -381,6 +381,142 @@ Tests that we can import an VideoFrame with non-YUV pixel format into a GPUExter
...
+      timestamp: 0,
+    };
+
+    const frame = new VideoFrame(imageData.data.buffer, frameInit);
```
kainino0x: It would be ideal if we additionally had a version of this test where the frame comes from a video file instead of a VideoFrame. Mostly in case there are differences in the YUV codepaths, but also just to have a better chance of the test running on all browsers.

If this won't be easy (I don't know how hard it is) then just add a `TODO(#bugnumber)`.

haoxli: This test is not too hard, but need to find a suitable video. I'll first try the existing videos, if it's not suitable for detecting display size, I'll need to create a new one. Add a TODO firstly.

src/webgpu/web_platform/external_texture/video.spec.ts
haoxli: Add a new test for checking `textureDimensions` and reuse much of the same code.

src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -515,6 +525,91 @@ sampling works without validation errors.
...
+    pass.end();
+    encoder.copyBufferToBuffer(storageBuffer, 0, readBuffer, 0, 8);
+    t.device.queue.submit([encoder.finish()]);
+
+    await readBuffer.mapAsync(GPUMapMode.READ);
+    const data = new Uint32Array(readBuffer.getMappedRange().slice(0));
+    readBuffer.unmap();
+
+    t.expect(
+      data[0] === displayWidth && data[1] === displayHeight,
```
kainino0x: Should be able to use `t.expectGPUBufferValuesEqual(storageBuffer, new Uint32Array([displayWidth, displayHeight]));` and remove `readBuffer`

haoxli: Got it. I will avoid rebases during reviewing.

PR #4597, opened 2026-02-17
Fix race condition in shader validation tests

diegodelatoba:
**Changes:**

  - Modified both `expectCompileResult` implementations to use a single
    `eventualAsyncExpectation` that executes GPU validation and compilation
    checks sequentially
  - Changed `let shaderModule` to `const shaderModule` for better code quality
  - Removed unnecessary non-null assertion operators

  **Testing:**

  Verified in WebKit with 500 consecutive test runs with no failures (previously
  failed around iteration 15-304).

  **Note:** This is a test infrastructure bug fix, not related to a spec change.
  No CTS project tracker issue is required.

  <hr>

    - N/A - This is a test infrastructure fix, not adding new test coverage
    - N/A - No new helpers added
    - Yes - Tested extensively in WebKit with 500+ consecutive successful runs
    - Yes - All affected tests pass in WebKit's test suite

### DIFF
--- a/src/webgpu/shader/validation/shader_validation_test.ts
+++ b/src/webgpu/shader/validation/shader_validation_test.ts
@@ -62,18 +62,22 @@ export class ShaderValidationTest extends AllFeaturesMaxLimitsGPUTest {
     if (options?.autoSkipIfFeatureNotAvailable !== false) {
       skipIfCodeNeedsFeatureAndDeviceDoesNotHaveFeature(this, code);
     }
-    let shaderModule: GPUShaderModule;
-    this.expectGPUError(
-      'validation',
-      () => {
-        shaderModule = this.device.createShaderModule({ code });
-      },
-      expectedResult !== true
-    );
 
-    const error = new Error();
-    this.eventualAsyncExpectation(async () => {
-      const compilationInfo = await shaderModule!.getCompilationInfo();
+    this.eventualAsyncExpectation(async (niceStack: Error) => {
+      if (expectedResult !== true) {
+        this.device.pushErrorScope('validation');
+      }
+      const shaderModule = this.device.createShaderModule({ code });
+
+      if (expectedResult !== true) {
+        const gpuError = await this.device.popErrorScope();
+        if (!(gpuError instanceof GPUValidationError)) {
+          niceStack.message = `Expected validation error`;
+          this.rec.expectationFailed(niceStack);
+        }
+      }
+
+      const compilationInfo = await shaderModule.getCompilationInfo();
 
       // MAINTENANCE_TODO: Pretty-print error messages with source context.
       const messagesLog =
@@ -83,6 +87,7 @@ export class ShaderValidationTest extends AllFeaturesMaxLimitsGPUTest {
         '\n\n---- shader ----\n' +
         code;
 
+      const error = new Error();
       if (compilationInfo.messages.some(m => m.type === 'error')) {
         if (expectedResult) {
           error.message = `Unexpected compilationInfo 'error' message.\n` + messagesLog;
@@ -124,7 +129,7 @@ export class ShaderValidationTest extends AllFeaturesMaxLimitsGPUTest {
 
     const error = new Error();
     this.eventualAsyncExpectation(async () => {
-      const compilationInfo = await shaderModule!.getCompilationInfo();
+      const compilationInfo = await shaderModule.getCompilationInfo();
 
       // MAINTENANCE_TODO: Pretty-print error messages with source context.
       const messagesLog = compilationInfo.messages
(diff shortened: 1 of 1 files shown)


diegodelatoba:
> I'm not seeing the race but I'm terrible at seeing races 😅
> 
> The first check just expects that eventually the compilation validation succeeds or fails. Regardless, it will immediately and synchronously return a shader module.
> 
> The 2nd check is that you can call `shaderModule.getComplationInfo` and eventually get a result, and, if you expected an error at least one message is an error message.
> 
> I don't see a race about which order those are expected happen. The test seems like it allows them happening in any order
> 
> @diegodelatoba where do you see the race?

@greggman  From my understanding the race isn't about which check happens first since those can happen in any order. Perhaps "race" was not the correct wording here but the issue here is that currently there is non-deterministic error message printing in the test output.

What I found is that `expectGPUError` and `eventualAsyncExpectation` both write to the test recorder (this.rec.debug() / this.rec.validationFailed()). When they run in parallel, their output messages can interleave non-deterministically. This causes issues in WebKit's test suite which relies on deterministic error logging.

greggman:
> > I'm not seeing the race but I'm terrible at seeing races 😅
> > The first check just expects that eventually the compilation validation succeeds or fails. Regardless, it will immediately and synchronously return a shader module.
> > The 2nd check is that you can call `shaderModule.getComplationInfo` and eventually get a result, and, if you expected an error at least one message is an error message.
> > I don't see a race about which order those are expected happen. The test seems like it allows them happening in any order
> > @diegodelatoba where do you see the race?
> 
> @greggman From my understanding the race isn't about which check happens first since those can happen in any order. Perhaps "race" was not the correct wording here but the issue here is that currently there is non-deterministic error message printing in the test output.
> 
> What I found is that `expectGPUError` and `eventualAsyncExpectation` both write to the test recorder (this.rec.debug() / this.rec.validationFailed()). When they run in parallel, their output messages can interleave non-deterministically. This causes issues in WebKit's test suite which relies on deterministic error logging.

ok, that makes more sense to me. I wouldn't expect to see logs get interleaved, there's still only one JS thread. But, the order individual tests get logged might be different.

I wonder if it would be better to order the outputs at a lower level. As it is, it looks like the code calls `this.rec.validationFailed` and similar things and those can get called in any order. Maybe we can do something there to handle this in a generic way everywhere? I don't know how many other places this might come up.

throwing out ideas

1. remove `this.rec` and require calling `t.getRec` which you'd have to call outside a callback

   ```js
   // OLD
   this.eventualAsyncExpectation(async () => {
     ...
     this.rec.validationFailed(error);
   );
   ```

   ```js
   // new
   const rec = this.getRec();
   this.eventualAsyncExpectation(async () => {
     ...
     rec.validationFailed(error);
   );
   ```

   This would allow `getRec` to return things that output in order. 

   Or along the same lines, make `eventualAsyncExpectation` pass in a `rec`

   ```js
   // new
   this.eventualAsyncExpectation(rec, async () => {
     ...
     rec.validationFailed(error);
   );
   ```

2. change `eventualAsyncExpectation` so it just adds functions to a list of things to do later and then calls them in order and awaits each one. Currently it calls the function immediately and puts the promise in an array to wait on. But maybe it could be changed put the function on an array then call each one in order.

I'm leaning toward something more like 1. There aren't that many calls to `this.rec` so maybe it would be a relatively small change to add internal order ids.

Or maybe we can just make this change for this test. I'd be nice to get @kainino0x's feedback.

kainino0x:
The ordering of test logs is definitely not meant to be deterministic, because it can always depend on things in the browser that have nondeterministic orders. We can certainly make some common code more deterministic, but it'll always be possible to write a test with nondeterministic logs.

Let me make sure I'm understanding the issue correctly.
- WebKit uses the WPT "runtime" for the CTS?
- The WPT runtime [only prints logs on failing tests](https://github.com/gpuweb/cts/blob/d483bf3bc9cb339181b269fbbb44e0cdfc74ea3f/src/common/runtime/wpt.ts#L79)

In which case you could only be running into log ordering on failing tests. So is the issue that tests that are expected to fail in WebKit have nondeterministic error logs, so your test infra can't verify that they are failing in the expected way? (If my memory is still accurate about how WebKit's failure expectations work)

mwyrzykowski:
> > > * WebKit uses the WPT "runtime" for the CTS?
> > 
> > 
> > At least not for the WebGPU CTS or many other WebKit tests
> 
> I didn't realize this, I thought you ran the [WPT-compatible build](https://github.com/gpuweb/cts/tree/gh-pages/out-wpt) using your WPT infra. Is the code open source where I could take a look to understand a bit better? It'll be useful for me in the future.

I think we use wpt test runner for wpt tests, but we use WebKitTestRunner for other tests, chatgpt does a pretty good job explaining it:

> 2. Layout Tests (Traditional WebKit Tests)
However, for layout tests — which are WebKit’s own long-standing format of rendering and behavior verification tests — WebKit does not use the WPT test runner.
Instead, it uses its own runner, called the WebKit Test Runner (WebKitTestRunner).
WebKitTestRunner is a part of WebKit’s internal tools.
It’s a small browser shell specifically built to run layout tests.
You typically invoke it through the top-level command:
bash

> run-webkit-tests
It supports pixel tests, text comparisons, dumping render trees, etc.

WebGPU CTS is run as part of our layout tests



> Ah, if it's just about the ordering of the "subcase ran" messages (as those seem to be the only logs in this test) then I think we can fix that generally. 

Yes this is the issue, if we can make that deterministic it would be great.

PR #4599, opened 2026-02-19
Explicitly specify rootDir in tsconfig.json to satisfy upcoming TypeScript v6.

freshp86:
See [1] and [2] where the "rootDir" needs to be explicitly specified, otherwise it will default to "." unlike v5 which was defaulted to the common root folder of all input files.

This fixes a build error in the downstream Chromium repository, see https://issues.chromium.org/issues/485931998.

[1] https://devblogs.microsoft.com/typescript/announcing-typescript-6-0-beta/#up-front-adjustments
[2] https://devblogs.microsoft.com/typescript/announcing-typescript-6-0-beta/#rootdir-now-defaults-to-.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -3,6 +3,7 @@
   "compilerOptions": {
     "lib": ["dom", "ES2023"],
     "module": "esnext",
+    "rootDir": "./src",
     // Use types only from these packages (not everything in package.json, which is the default).
     // Note that TypeScript's `/// <reference />` is the same as adding an entry here - it's global
     // to the whole project.


greggman:
this change doesn't seem to break anything, but, upgrading to typescript@6.0.0-beta break many things. Is this sufficient? If you're using typescript 6+, it seems like just this fix will not be enough?

freshp86:
> but, upgrading to [typescript@6.0.0-beta](mailto:typescript@6.0.0-beta) break many things.

The only thing I verified is that with this change I could build Chrome successfully (I can't run the bots without fixing the build error first, which requires this change). 

> Is this sufficient? If you're using typescript 6+, it seems like just this fix will not be enough?

There are probably more things that need to be fixed, but I don't think this change needs to be blocked, as it simply explicitly specifies a default value since it is no longer default in v6.

Moreover, landing this change and rolling it into Chromium, will help me identify what more needs to be fixed so that I can file bugs as needed.

greggman (approved):

PR #4608, opened 2026-02-26
Test swizzle assignment eval order

sudonatalie:
Test that the lhs swizzle's vector pointer is evaluated before the rhs. For a regular assignment, the vector value can be loaded after rhs evaluation (because rhs side effects on the target components will be overwritten anyways). However for a compound assignment, the lhs value of the swizzle must be loaded before rhs evaluation, and the vector should be re-loaded after rhs evaluation for use as the "old components" in the store of the whole result vector, because side effects from the rhs on non-swizzled components should persist.

<hr>

### DIFF
--- a/src/webgpu/shader/execution/statement/swizzle_assignment.spec.ts
+++ b/src/webgpu/shader/execution/statement/swizzle_assignment.spec.ts
@@ -7,6 +7,7 @@ import { keysOf } from '../../../../common/util/data_tables.js';
 import { TypedArrayBufferView } from '../../../../common/util/util.js';
 import { Float16Array } from '../../../../external/petamoriken/float16/float16.js';
 import { AllFeaturesMaxLimitsGPUTest, GPUTest } from '../../../gpu_test.js';
+import { runFlowControlTest } from '../flow_control/harness.js';
 
 export const g = makeTestGroup(AllFeaturesMaxLimitsGPUTest);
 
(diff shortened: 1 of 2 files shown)


src/webgpu/shader/execution/statement/swizzle_assignment.spec.ts
```diff
@@ -420,3 +421,76 @@ fn main() {
...
+fn foo() -> u32 {
+  ${f.expect_order(1)}
+  arr[0].x = 6;       // modifies x before add
+  arr[0].z = 7;       // overwritten by bar()
+  arr[0].w = 8;       // persists
+  return 0;
+}
+fn bar() -> vec2u {
+  ${f.expect_order(2)}
+  arr[0].x = 2;       // no visible effect
```
sudonatalie: @dneto0  It's how I've written the test here, but I'm not really sure about the 2 loads for a compound assignment. If rhs modifies both swizzle components and non-swizzle components, it feels like its modifications to the swizzle components should be no-ops (because lhs was already evaluated), but the modifications to the non-swizzle components should persist. I'm more confident the latter is necessary than the former. If we're already going to be enumerating in the spec that the vec pointer on the lhs is evaluated before the rhs, but not loaded until after rhs, then I can see an argument for desugaring a compound assignment resulting in rhs influencing the lhs before it's operated on. And since hopefully no one would ever ever do these side effects, it does seem better for it to only be 1 load.

Issue #4605, opened 2026-02-25
Test `importExternalTexture` with video frame display sizes and coded sizes from video file.


haoxli:
If a video frame has different display sizes and coded sizes, `importExternalTexture` may throw validation errors and the external texture cannot be sampled correctly. This bug had been fixed in https://crbug.com/471021591, but the tests was missing.

The [PR](https://github.com/gpuweb/cts/pull/4589) added the tests for a video frame from `VideoFrame` element. We would also prefer a version of these tests where the frame comes from a video file instead of a `VideoFrame`.

PR #4614, opened 2026-03-17
Update display size scale tests with video frame from video file

haoxli:
For display‑size scale tests, a VideoFrame created from a video file is the right source to validate importExternalTexture because it exercises the real decode, color conversion, and display‑size metadata path. A canvas ImageData frame bypasses the video decode path, so it’s removed.

Added a focused video list to validate scaling across codec and color‑space variation, covering both codecs (VP9, H.264) and both color spaces (bt.601, bt.709).

Issue: #4605

<hr>

### DIFF
--- a/src/webgpu/web_platform/external_texture/video.spec.ts
+++ b/src/webgpu/web_platform/external_texture/video.spec.ts
@@ -7,8 +7,6 @@ Tests for external textures from HTMLVideoElement (and other video-type sources?
 TODO: consider whether external_texture and copyToTexture video tests should be in the same file
 TODO(#3193): Test video in BT.2020 color space
 TODO(#4364): Test camera capture with copyExternalImageToTexture (not necessarily in this file)
-TODO(#4605): Test importExternalTexture with video frame display size different with coded size from
-  a video file
 `;
 
 import { makeTestGroup } from '../../../common/framework/test_group.js';
@@ -33,6 +31,14 @@ import {
 const kHeight = 16;
 const kWidth = 16;
 const kFormat = 'rgba8unorm';
+
+const kDisplayScaleVideoNames = [
+  'four-colors-h264-bt601.mp4',
+  'four-colors-vp9-bt601.webm',
+  'four-colors-vp9-bt709.webm',
+] as const;
+type DisplayScale = 'smaller' | 'same' | 'larger';
+const kDisplayScales: DisplayScale[] = ['smaller', 'same', 'larger'];
 
 export const g = makeTestGroup(TextureUploadingUtils);
 
@@ -164,67 +170,55 @@ function checkNonStandardIsZeroCopyIfAvailable(): { checkNonStandardIsZeroCopy?:
   }
 }
 
-function createVideoFrameWithDisplayScale(
+async function createVideoFrameWithDisplayScale(
   t: GPUTest,
-  displayScale: 'smaller' | 'same' | 'larger'
-): { frame: VideoFrame; displayWidth: number; displayHeight: number } {
-  const canvas = createCanvas(t, 'onscreen', kWidth, kHeight);
-  const canvasContext = canvas.getContext('2d');
+  videoName: (typeof kDisplayScaleVideoNames)[number],
+  displayScale: DisplayScale
+): Promise<VideoFrame> {
+  let sourceFrame: VideoFrame | undefined;
+  let codedWidth: number | undefined;
+  let codedHeight: number | undefined;
 
-  if (canvasContext === null) {
-    t.skip(' onscreen canvas 2d context not available');
+  const videoElement = getVideoElement(t, videoName);
+
+  await startPlayingAndWaitForVideo(videoElement, async () => {
+    const source = await getVideoFrameFromVideoElement(t, videoElement);
+    sourceFrame = source;
+    codedWidth = source.codedWidth;
+    codedHeight = source.codedHeight;
+  });
+
+  if (sourceFrame === undefined || codedWidth === undefined || codedHeight === undefined) {
+    unreachable();
   }
 
-  const ctx = canvasContext;
-
-  const rectWidth = Math.floor(kWidth / 2);
-  const rectHeight = Math.floor(kHeight / 2);
-
-  // Red
-  ctx.fillStyle = `rgba(255, 0, 0, 1.0)`;
-  ctx.fillRect(0, 0, rectWidth, rectHeight);
-  // Lime
-  ctx.fillStyle = `rgba(0, 255, 0, 1.0)`;
-  ctx.fillRect(rectWidth, 0, kWidth - rectWidth, rectHeight);
-  // Blue
-  ctx.fillStyle = `rgba(0, 0, 255, 1.0)`;
-  ctx.fillRect(0, rectHeight, rectWidth, kHeight - rectHeight);
-  // Fuchsia
-  ctx.fillStyle = `rgba(255, 0, 255, 1.0)`;
-  ctx.fillRect(rectWidth, rectHeight, kWidth - rectWidth, kHeight - rectHeight);
-
-  const imageData = ctx.getImageData(0, 0, kWidth, kHeight);
-
-  let displayWidth = kWidth;
-  let displayHeight = kHeight;
+  let displayWidth = codedWidth;
+  let displayHeight = codedHeight;
   switch (displayScale) {
     case 'smaller':
-      displayWidth = Math.floor(kWidth / 2);
-      displayHeight = Math.floor(kHeight / 2);
+      displayWidth = Math.floor(codedWidth / 2);
+      displayHeight = Math.floor(codedHeight / 2);
       break;
     case 'same':
-      displayWidth = kWidth;
-      displayHeight = kHeight;
+      displayWidth = codedWidth;
+      displayHeight = codedHeight;
       break;
     case 'larger':
-      displayWidth = kWidth * 2;
-      displayHeight = kHeight * 2;
+      displayWidth = codedWidth * 2;
+      displayHeight = codedHeight * 2;
       break;
     default:
       unreachable();
   }
 
-  const frameInit: VideoFrameBufferInit = {
-    format: 'RGBA',
-    codedWidth: kWidth,
-    codedHeight: kHeight,
+  const frame = new VideoFrame(sourceFrame, {
     displayWidth,
     displayHeight,
-    timestamp: 0,
-  };
+    timestamp: sourceFrame.timestamp,
+  });
+  sourceFrame.close();
 
-  const frame = new VideoFrame(imageData.data.buffer, frameInit);
-  return { frame, displayWidth, displayHeight };
+  return frame;
 }
 
 g.test('importExternalTexture,sample')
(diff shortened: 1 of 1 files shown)


src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -496,11 +496,30 @@ sampling works without validation errors.
...
     t.device.queue.submit([commandEncoder.finish()]);
 
+    // Build expected sampled colors by drawing the same source frame to a 2D canvas.
+    // This makes the check robust across codecs/container metadata while still validating
+    // that importExternalTexture sampling matches browser video rendering behavior.
+    const canvas = createCanvas(t, 'onscreen', kWidth, kHeight);
+    const canvasContext = canvas.getContext('2d', { colorSpace: 'srgb' });
+    if (canvasContext === null) {
+      frame.close();
+      t.skip(' onscreen canvas 2d context not available');
```
shaoboyan091: leading space, pls remove.

src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -164,67 +170,55 @@ function checkNonStandardIsZeroCopyIfAvailable(): { checkNonStandardIsZeroCopy?:
   }
 }
 
-function createVideoFrameWithDisplayScale(
+async function createVideoFrameWithDisplayScale(
```
shaoboyan091, line 177: Would you mind to add comments similar to your commit message to simple describe we need real video file instead of canvas created ones?

src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -164,67 +170,55 @@ function checkNonStandardIsZeroCopyIfAvailable(): { checkNonStandardIsZeroCopy?:
...
+  await startPlayingAndWaitForVideo(videoElement, async () => {
+    const source = await getVideoFrameFromVideoElement(t, videoElement);
+    sourceFrame = source;
+    codedWidth = source.codedWidth;
+    codedHeight = source.codedHeight;
+  });
 
-  if (canvasContext === null) {
-    t.skip(' onscreen canvas 2d context not available');
+  if (sourceFrame === undefined || codedWidth === undefined || codedHeight === undefined) {
```
shaoboyan091: Using `assert`? Or `skip` with correct info

src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -164,67 +170,55 @@ function checkNonStandardIsZeroCopyIfAvailable(): { checkNonStandardIsZeroCopy?:
...
-  ctx.fillRect(0, rectHeight, rectWidth, kHeight - rectHeight);
-  // Fuchsia
-  ctx.fillStyle = `rgba(255, 0, 255, 1.0)`;
-  ctx.fillRect(rectWidth, rectHeight, kWidth - rectWidth, kHeight - rectHeight);
-
-  const imageData = ctx.getImageData(0, 0, kWidth, kHeight);
-
-  let displayWidth = kWidth;
-  let displayHeight = kHeight;
+  let displayWidth = codedWidth;
```
shaoboyan091, line 243: Or just assign 0 instead of `codedWidth`.

src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -34,6 +32,14 @@ const kHeight = 16;
 const kWidth = 16;
 const kFormat = 'rgba8unorm';
 
+const kDisplayScaleVideoNames = [
+  'four-colors-h264-bt601.mp4',
+  'four-colors-vp9-bt601.webm',
+  'four-colors-vp9-bt709.webm',
+] as const;
+type DisplayScale = 'smaller' | 'same' | 'larger';
+const kDisplayScales: DisplayScale[] = ['smaller', 'same', 'larger'];
```
shaoboyan091: If thit won't expand in future then I slightly lean to keep the origin format (Using `'smaller' | 'same' | 'larger'`  in all three places).

src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -496,11 +496,30 @@ sampling works without validation errors.
     passEncoder.end();
     t.device.queue.submit([commandEncoder.finish()]);
 
+    // Build expected sampled colors by drawing the same source frame to a 2D canvas.
+    // This makes the check robust across codecs/container metadata while still validating
+    // that importExternalTexture sampling matches browser video rendering behavior.
+    const canvas = createCanvas(t, 'onscreen', kWidth, kHeight);
```
shaoboyan091: Just double check, are we using `kWidth`, `kHeight` instead of using `codedWidth`, `codedHeight` for some purpose? 

The origin part use `kWidth` and `kHeight` due to the generated video frame is from a `kWidth` and `kHeight` (which mapping to `codedWidth` and `codedHeight` here) canvas.

shaoboyan091 (commented):
BTW, this http://127.0.0.1:8080/standalone/?q=webgpu:web_platform,external_texture,video:importExternalTexture,video_frame_display_size_scale:videoName=%22four-colors-h264-bt601.mp4%22;displayScale=%22same%22 fail in canary.

src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -164,67 +170,55 @@ function checkNonStandardIsZeroCopyIfAvailable(): { checkNonStandardIsZeroCopy?:
...
       break;
     default:
       unreachable();
   }
 
-  const frameInit: VideoFrameBufferInit = {
-    format: 'RGBA',
-    codedWidth: kWidth,
-    codedHeight: kHeight,
+  const frame = new VideoFrame(sourceFrame, {
```
kainino0x, line 258: Is there a way to encode this data in a video file rather than going through VideoFrame? Doing it with VideoFrame seems kind of contrived - the way people would actually run into issues in the wild is through video files that have different dimensions encoded in them.

src/webgpu/web_platform/external_texture/video.spec.ts
kainino0x: There are too many different widths and heights. Can you rename `kWidth` and `kHeight` to something more specific? (`kCanvasWidth` `kCanvasHeight`)

haoxli: That's not entirely accurate. I only used `kWidth` and `kHeight` to avoid having to define an additional size variable here. We also can use  `displayWidth` and `displayHeight` in `VideoFrame` for that. In the origin PR, I used `canvas` becuase I found it also can reporduce the issue and I can control the sizes easily.

haoxli: Ok, I will return `displayWidth` and `displayHegith` from `createVideoFrameWithDisplayScale`, and use them in the cases stead of `kWidth` and `kHeight`.

src/webgpu/web_platform/external_texture/video.spec.ts
haoxli: It doesn't seem easy to create a video file with display size scale. I see we can use `setsar=2/1` to make video display size becomes 2 times.

`ffmpeg.exe -loop 1 -i .\four-colors.png -c:v libx264 -pix_fmt yuv420p  -frames 50 -colorspace smpte170m -color_primaries smpte170m -color_trc smpte170m -color_range tv  -vf "setsar=2/1" four-colors-h264-bt601-sar2.mp4`

The display height becomes two times of coded width, but the width does not,  and the four colored squares are still displayed according to the coded size.
Image

If I force `DAR` using `setsar=2/1,setdar=4/3`, the `SAR` will be 1:1.
`ffmpeg.exe -loop 1 -i .\four-colors.png -c:v libx264 -pix_fmt yuv420p  -frames 50 -colorspace smpte170m -color_primaries smpte170m -color_trc smpte170m -color_range tv  -vf "setsar=2/1,setdar=4/3" four-colors-h264-bt601-sar2.mp4`

Verify:
`ffprobe.exe  -v error -select_streams v:0 -show_entries stream=width,height,coded_width,coded_height,sample_aspect_ratio,display_aspect_ratio -of default=noprint_wrappers=1 .\four-colors-h264-bt601-sar2.mp4`

```
width=320
height=240
coded_width=320
coded_height=240
sample_aspect_ratio=1:1
display_aspect_ratio=4:3
```


I will further investigate how to make it works.

If it works, I think we need to add 6 new video files at least (3 different codes/color spaces videos x 2 display scales (`larger`/`smaller`).

kainino0x: That should be fine as long as they're not too large. Thank you for investigating!

haoxli: I can create the mp4 video which display size (160x120) is less than its coded size(320x240), but this video file cannot trigger the validation error like `cropSize [Extent2D width:16, height:16] exceeds the texture size, defined by Plane0 size (8, 8).` if directly passing a video frame from `MediaStreamTrackProcessor` to `importExternalTexture()`.
```
ffmpeg.exe -loop 1 -i .\four-colors.png -c:v libx264 -pix_fmt yuv420p -frames 50 -colorspace smpte170m -color_primaries smpte170m -color_trc smpte170m -color_range tv -bsf:v "h264_metadata=crop_left=80:crop_right=80:crop_top=60:crop_bottom=60" four-colors-h264-bt601-cropped.mp4
```

# Further investigation 
In the `importExternalTexture()` bug, the 1-copy path sets `cropSize` from `visible_rect` at [external_texture_helper.cc:254](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/webgpu/external_texture_helper.cc;l=254) but allocates the texture at `natural_size`. When `visible_rect` > `natural_size`, the crop rectangle exceeds the texture bounds, causing a Dawn validation error.

Cropped video (SAR=1:1, coded 320×240, visible 160×120)
| Property | Value | Example |
| -- | -- | -- |
| `coded_size` | Full encoded frame (includes padding)  | 320×240  |
| `visible_rect`  | Cropped content region  | 160 × 120 |
| `natural_size`  | Display size = `visible_rect` × SAR  | 160 × 120 (SAR=1:1) |

```
cropSize = visible_rect (320×240)
textureSize = natural_size (160×120)
cropSize > textureSize → VALIDATION ERROR
```

Our fix sets `cropSize = natural_size` in the 1-copy path so it matches the texture size. So actually the test requires the video frame which `visible_rect > natural_size` . But for video files, `natural_size` is computed, not stored. The media pipeline computes `natural_size = GetNaturalSize(visible_rect, SAR)`, which always scales up, guaranteeing `natural_size >= visible_rect`. No SAR value can produce `natural_size < visible_rect`. So **the cropped videos don't help, we need to test it with `new VideoFrame(data, {codedWidth, codedHeight, displayWidth, displayHeight})`**.

```
new VideoFrame(buffer, {
  codedWidth: 320,    // → coded_size = 320×240
  codedHeight: 240,
  timestamp: 0,
  format: 'RGBA',
  displayWidth: 160,  // → natural_size = 160×120  (SET DIRECTLY)
  displayHeight: 120,
});
```

```
coded_size    = 320×240  (from codedWidth/Height)
visible_rect  = (0, 0, 320, 240)  (defaults to full coded_size)
natural_size  = 160×120  (from displayWidth/Height - NO computation, just stored as-is)
```

And the video-decoded frames take the 0-copy path anyway at [external_texture_helper.cc:274](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/webgpu/external_texture_helper.cc;l=274). Decoded frames have a GPU `SharedImage (HasSharedImage() = true)`, so they may enter the 0-copy path, completely bypassing the buggy 1-copy code.
```
  const bool zero_copy =
      (media_video_frame->HasSharedImage() &&
       (media_video_frame->format() == media::PIXEL_FORMAT_NV12) &&
       device_support_zero_copy &&
       media_video_frame->metadata().is_webgpu_compatible &&   // ← usually false for video decode
       DstColorSpaceSupportedByZeroCopy(dst_predefined_color_space));
```

In practice, `is_webgpu_compatible` is typically false for video-decoded frames, so it will fall through to 1-copy. But this is not guaranteed by the test itself, it's an implementation detail that varies by platform/config. To guarantee the 1-copy path, **the test should create a VideoFrame from raw pixel buffer data** (not from a decoded video). A buffer-backed VideoFrame will never have `HasSharedImage()`, so it always hits 1-copy. 

I will keep tests with both video-decoded frames and buffer-backed frames, and add the comments to the tests.

src/webgpu/web_platform/external_texture/video.spec.ts
haoxli: Added both video files and canvas elements as VideoFrame sources for separate paths.

src/webgpu/web_platform/external_texture/video.spec.ts
```diff
@@ -496,58 +561,124 @@ sampling works without validation errors.
...
+        const xi = Math.floor(x);
+        const yi = Math.floor(y);
+        const i = (yi * displayWidth + xi) * 4;
+        return new Uint8Array([bytes[i + 0], bytes[i + 1], bytes[i + 2], bytes[i + 3]]);
+      };
+      expected = {
+        topLeft: sample(displayWidth * 0.25, displayHeight * 0.25),
+        topRight: sample(displayWidth * 0.75, displayHeight * 0.25),
+        bottomLeft: sample(displayWidth * 0.25, displayHeight * 0.75),
+        bottomRight: sample(displayWidth * 0.75, displayHeight * 0.75),
```
kainino0x, line 615: Shouldn't these four sample points still correspond to the centers of the four solid colors? It's hard to understand what this `sample` logic is doing which makes it hard to verify the test is correct, compared with just expecting 4 different colors.

PR #4633, opened 2026-04-27
Correct MAINTENANCE_TODO about TRANSIENT_ATTACHMENT

kainino0x:
This has been added to the spec but we can't remove it yet because it would cause tests to fail on Firefox/Safari. Update the TODO to indicate this.

Issue: #4509


beaufortfrancois (approved):
LGTM

Issue #4640, opened 2026-05-14
Add tests on the feature`subgroup-size-control`


Jiawei-Shao:
This issue tracks the efforts on the tests against the WebGPU feature `subgroup-size-control` based on the WebGPU SPEC PR (https://github.com/gpuweb/gpuweb/pull/5578).

API Validations:
- Enabling `subgroup-size-control` on the creation of device should also enable `subgroups`.

Shader Validations: 
- Enabling `subgroup_size_control` must also enable `subgroups`.
- The WGSL attribute `@subgroup_size` cannot be used when `subgroup_size_control` is not enabled.
- `@subgroup_size` can only be used in the compute stage
- The value of `@subgroup_size` must be a constant expression or an override expression that resolves to an `i32` or an `u32`.
- The value of `@subgroup_size` must be a power of 2.
- The x-dimension of the entry point's `workgroup_size` must be a multiple of the `@subgroup_size` value.

Shader Execution Tests:
- The value of  the built-in variable `subgroup_size` must equal the value of the `@subgroup_size` attribute.
- At least one value in the range of `[GPUAdapterInfo.subgroupMinSize, GPUAdapterInfo.subgroupMaxSize]` can be used as `@subgroup_size` attribute in a simple compute pipeline.

kainino0x:
> * The WGSL attribute `@subgroup_size` must be used when `subgroup_size_control` is enabled.

This seems wrong, do you mean the reverse? (The feature must be enabled to use the attribute)

Jiawei-Shao:
> > * The WGSL attribute `@subgroup_size` must be used when `subgroup_size_control` is enabled.
> 
> This seems wrong, do you mean the reverse? (The feature must be enabled to use the attribute)

Yes it was my fault. I've corrected it.

PR #4641, opened 2026-05-14
Add shader validation tests on `subgroup-size-control` - Part I

Jiawei-Shao:
This patch adds the first part of the shader validation tests on the extension `subgroup-size-control`:
- Enabling `subgroup_size_control` must also enable `subgroups`.
- The WGSL attribute `@subgroup_size` cannot used when `subgroup_size_control` is not enabled.
- `@subgroup_size` can only be used in the compute stage

Issue: #4640

<hr>

- [*] Test have be tested with compatibility mode validation enabled and behave as expected. (If not passing, explain above.)


Jiawei-Shao:
PTAL, thanks!

src/webgpu/shader/validation/extension/subgroup_size_control.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+import { makeTestGroup } from '../../../../common/framework/test_group.js';
+import { UniqueFeaturesAndLimitsShaderValidationTest } from '../shader_validation_test.js';
+
+export const g = makeTestGroup(UniqueFeaturesAndLimitsShaderValidationTest);
+
+/**
+ * Returns a subgroup size value that is valid for use in the @subgroup_size
+ * attribute on the current adapter.
+ *
+ * On Intel gen-12lp, subgroupMinSize may be 8 in fragment stages, which is below the allowed range
```
kainino0x: You're saying adapterInfo.subgroupMinSize is 8 on these devices, but 8 is not actually a valid subgroup size in compute? That's kind of weird, it requires applications to do weird stuff like this to get the right subgroup size. I would have thought subgroupMinSize/subgroupMaxSize are just for compute, especially now that we're adding subgroup-size-control which only applies to compute.

I understand we're saying not all values between subgroupMinSize and subgroupMaxSize are valid to create all pipelines with, but surely they should all work for a trivial compute pipeline? Maybe we can't technically specify that, but I think we should try to test it unless there are known counterexamples.

src/webgpu/shader/validation/extension/subgroup_size_control.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+    });
+  })
+  .fn(t => {
+    const { enableExtension } = t.params;
+
+    const subgroupSize = getValidSubgroupSizeForSubgroupSizeAttribute(t.device.adapterInfo);
+    t.expectCompileResult(
+      enableExtension,
+      `
+        ${enableExtension ? 'enable subgroups; enable subgroup_size_control;' : ''}
```
kainino0x: Tighten this so we're changing as little as possible between the true and false cases.
```suggestion
        enable subgroups;
        ${enableExtension ? 'enable subgroup_size_control;' : ''}
```

src/webgpu/shader/validation/extension/subgroup_size_control.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+  .params(u => u.combine('enableExtension', [false, true] as const))
+  .beforeAllSubcases(t => {
+    t.selectDeviceOrSkipTestCase({
+      requiredFeatures: ['subgroup-size-control' as GPUFeatureName],
+    });
+  })
+  .fn(t => {
+    const { enableExtension } = t.params;
+
+    const subgroupSize = getValidSubgroupSizeForSubgroupSizeAttribute(t.device.adapterInfo);
```
kainino0x: Does compiling a shader really require the subgroup size to be valid for the device? It seems like it should work for any size, and only be validated at pipeline creation.

src/webgpu/shader/validation/extension/subgroup_size_control.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+      enableExtension,
+      `
+        ${enableExtension ? 'enable subgroups; enable subgroup_size_control;' : ''}
+        @compute @workgroup_size(${subgroupSize}) @subgroup_size(${subgroupSize})
+        fn main() {}
+      `
+    );
+  });
+
+const kStageShaders = {
```
kainino0x: nit: If these are only going to be used by one test, put them inside the test.

src/webgpu/shader/validation/extension/subgroup_size_control.spec.ts
jrprice: I believe it is true that on these devices, the subgroup size is 8 for fragment shaders but that is never a valid subgroup size for a compute shader. This is the reason we decided to say that it would be a spurious failure if you requested a subgroup size that is not supported (otherwise we have to add a second set of limits as previously proposed, which we decided against).

So no, unfortunately not all sizes between min and max will work even for trivial compute pipelines.

src/webgpu/shader/validation/extension/subgroup_size_control.spec.ts
jrprice: Agreed, I don't see anything in the spec PR that suggests that the devices constraints would cause a shader-creation error, just the power-of-two and multiple-of-workgroup-x-dim things which are not dependent on device. So in theory you could just pick something like 32 here for the purposes of these tests.

src/webgpu/shader/validation/extension/subgroup_size_control.spec.ts
Jiawei-Shao: > You're saying adapterInfo.subgroupMinSize is 8 on these devices, but 8 is not actually a valid subgroup size in compute?

Yes that's the case on Intel D3D12 Gen12 GPUs. D3D12 only requires the value used for `[WaveSize]` must be between `[WaveLaneCountMin, WaveLaneCountMax]`, while it never means `[WaveLaneCountMin, WaveLaneCountMax]` has the same concept as `minSubgroupSize` and `maxSubgroupSize` in Vulkan. 

That's why in the initial proposal there were another two members in `adapterInfo` (`minComputeExplicitSubgroupSize` and `maxComputeExplicitSubgroupSize`) that show the actual range we can use for `@subgroup_size`.

> Maybe we can't technically specify that, but I think we should try to test it unless there are known counterexamples.

I am afraid we can only test it per vendor and architecture, which makes the test doesn't look so good, but I have to do so as we have dropped the query on the range of the subgroup sizes that can be used for `@subgroup_size`.

src/webgpu/shader/validation/extension/subgroup_size_control.spec.ts
Jiawei-Shao: Done
Now I always use `4` as `@subgroup_size`, which actually cannot be supported on a variety of GPUs.

src/webgpu/shader/validation/extension/subgroup_size_control.spec.ts
kainino0x: Continuing this thread in https://github.com/gpuweb/gpuweb/pull/5578/files#r3268652728

kainino0x (approved):

PR #4643, opened 2026-05-18
Add shader execution tests on the feature `subgroup-size-control`

Jiawei-Shao:
This patch adds the shader execution tests on the feature `subgroup-size-control`:
- The value of the built-in variable `subgroup_size` must equal the value of the `@subgroup_size` attribute.
- At least one value in the range of `[subgroupMinSize, subgroupMaxSize]` can be used as `@subgroup_size`
  attribute in a simple compute pipeline.

Issue: #4640

<hr>

- [*] Test have be tested with compatibility mode validation enabled and behave as expected. (If not passing, explain above.)


Jiawei-Shao:
@kainino0x @jrprice PTAL, thanks!

src/webgpu/shader/execution/shader_io/compute_builtins.spec.ts
```diff
@@ -1168,3 +1168,203 @@ fn main(@builtin(local_invocation_id) local_id : vec3u,
 
     t.expectOK(checkNumSubgroupsConsistency(countData, outputData, wgThreads, t.params.numWGs));
   });
+
+g.test('subgroup_size_attribute')
+  .desc(
+    'Tests that the value of the subgroup_size builtin must equal the value of the @subgroup_size attribute.'
```
jrprice: This test feels like just a subset of the test below, can we fold them together? i.e. for any pipeline that is successfully created in the below test, the value of the `subgroup_size` builtin should match the subgroup size attribute, and there must be at least one that succeeds.

That way we get better coverage of the fact that subgroup_size does actually change the subgroup size.

Jiawei-Shao: Done

src/webgpu/shader/execution/shader_io/compute_builtins.spec.ts
```diff
@@ -1168,3 +1168,132 @@ fn main(@builtin(local_invocation_id) local_id : vec3u,
...
+  .fn(async t => {
+    t.skipIfDeviceDoesNotHaveFeature('subgroup-size-control' as GPUFeatureName);
+
+    const { numWorkGroups, numSubgroups } = t.params;
+
+    const subgroupSizes = await getValidSubgroupSizes(t.device);
+    t.expect(
+      subgroupSizes.length > 0,
+      `No valid @subgroup_size value found in [subgroupMinSize, subgroupMaxSize]`
+    );
```
jrprice: Do we need to do this as a separate step like this?
I think it'd be simpler to just trap the potential pipeline creation error in the loop below and skip over it.
Then just have a flag to track whether or not there has been at least one pipeline that succeeded.

Otherwise this change LGTM now.

Jiawei-Shao: Done

jrprice (approved):
LGTM

PR #4652, opened 2026-05-30
buffer_view CTS

alan-baker:
buffer_view validation and execution CTS tests.

Issue: #

<hr>


mwyrzykowski (approved):

src/webgpu/api/validation/dispatch.spec.ts
```diff
@@ -141,3 +143,338 @@ fn main(@builtin(${t.params.builtin}) input : u32,
...
+  )
+  .fn(t => {
+    const testcase = kRequiredSizeCases[t.params.case];
+    const features = testcase.requires ?? [];
+    features.forEach(f => {
+      t.skipIfLanguageFeatureNotSupported(f);
+    });
+
+    const buffer = t.createBufferTracked({
+      size: t.params.valid ? testcase.size : testcase.size - 4,
```
dneto0, line 504: I had to look up why this uses "- 4".  
It comes from the fact that the buffer binding size in createBindGroup must be a multiple of 4.  That's just a familiarity thing for me. 

No request for change.

src/webgpu/api/validation/dispatch.spec.ts
```diff
@@ -141,3 +143,338 @@ fn main(@builtin(${t.params.builtin}) input : u32,
...
+fn main() {
+  let o = 16;
+  let s = 32;
+  let p = bufferArrayView<array<S>>(&v, o, s);
+}`,
+    size: 32,
+    binding_type: 'read-only-storage',
+    requires: ['buffer_view'],
+  },
+};
```
dneto0, line 485: This needs test cases for:
- offset is an override expression
    -  with initializer
    -  with pipeline-time-supplied initializer value
- size parameter is an override expression.
    - if so, then use that size.
    - otherwise use MinTypeSize.  (See my question / comment at https://github.com/gpuweb/gpuweb/pull/6291/changes#r3469837734  )

src/webgpu/api/validation/dispatch.spec.ts
```diff
@@ -141,3 +143,338 @@ fn main(@builtin(${t.params.builtin}) input : u32,
...
+
```
dneto0, line 486: Also, the bufferArrayView cases less rich than the bufferView cases.

Interesting ones missing for bufferArrayView
- compute max value across multiple uses in a shader
- multiple uses in the shader, but only pay attention to the ones statically used by the shader in the pipeline

src/webgpu/shader/execution/expression/call/builtin/buffer_view_utils.ts
```diff
@@ -0,0 +1,957 @@
...
+
+export function isValidArrayLengthCase(
+  type: ArrayLengthType,
+  offset: number,
+  size: number,
+  bufferSize: number
+): boolean {
+  if (size === 0) {
+    // bufferView case
+    return offset + (type.arrayOffset ?? 0) + type.stride < bufferSize;
```
dneto0: Should this be less-than-or-equal?
Also on line 45.

src/webgpu/shader/execution/expression/call/builtin/buffer_view_utils.ts
```diff
@@ -0,0 +1,957 @@
...
+  } else {
+    // bufferArrayView case
+    if ((type.arrayOffset ?? 0) + type.stride > size) {
+      return false;
+    }
+    return offset + size < bufferSize;
+  }
+}
+
+export function calculateArrayLength(
```
dneto0, line 67: Worth saying this returns the element count. That's a bit less ambiguous

src/webgpu/shader/validation/decl/let.spec.ts
```diff
@@ -109,6 +109,36 @@ const kTypeCases: Record<string, Case> = {
     valid: 'texture_and_sampler_let',
     decls: `@group(0) @binding(0) var samp_comp : sampler_comparison;`,
   },
+  buffer: {
+    code: `let x : buffer = b;`,
+    valid: false,
+    decls: `@group(0) @binding(0) var<storage> b : buffer;`,
+  },
+  buffer_sized: {
+    code: `let x = buffer<128> = b;`,
```
dneto0: typo:  first `=` should be `:`

src/webgpu/shader/validation/decl/let.spec.ts
```diff
@@ -178,6 +210,16 @@ const kInitCases: Record<string, Case> = {
...
+  buffer_mismatch: {
+    code: `let x : ptr<uniform, buffer> = &b;`,
+    valid: false,
+    decls: `@group(0) @binding(0) var<uniform> b : buffer<128>;`,
+  },
+  buffer_smaller_size: {
+    code: `let x : ptr<uniform, buffer<64>> = &b;`,
+    valid: false,
+    decls: `@group(0) @binding(0) var<uniform> b : buffer<128>;`,
+  },
```
dneto0, line 222: another case:

- workgroup sized buffer with override size, assign its address to ptr-workgr-sized-buffer but mismatched override identifier.

dneto0 (changes requested):
only partway reviewed

src/webgpu/api/validation/dispatch.spec.ts
dneto0: uh, I think my request is nonsense because the size param has to be a const-expression. (!)

src/webgpu/shader/validation/types/buffer.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+  },
+  const_negative: {
+    code: `const x = -1; alias T = buffer<x>;`,
+    valid: false,
+  },
+  const_zero: {
+    code: `const x = 0; alias T = buffer<x>;`,
+    valid: false,
+  },
+  const_functio: {
```
dneto0: nit: typo in key name

src/webgpu/shader/validation/types/buffer.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+}`;
+
+    let expected = t.params.aspace !== 'function' && t.params.aspace !== 'private';
+    if (t.params.case === 'unsized') {
+      expected = t.params.aspace === 'storage';
+    } else if (t.params.case.includes('override')) {
+      expected = t.params.aspace === 'workgroup';
+    }
+    t.expectCompileResult(expected, wgsl);
+  });
```
dneto0, line 188: Could add a case testing pointer types.


> alias PT = ptr<address_space, T>;

(with shenanigans for read-only/read_write access modes)

src/webgpu/api/validation/dispatch.spec.ts
dneto0: Alan reminded me that the case here is using an override expression for the function arguments; not for the declarations of the buffers

src/webgpu/shader/validation/types/buffer.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+      .combine('aspace', ['function', 'private', 'storage', 'uniform', 'workgroup'] as const)
+  )
+  .fn(t => {
+    t.skipIfLanguageFeatureNotSupported('buffer_view');
+    const testcase = kParseCases[t.params.case];
+
+    let mvar = '';
+    let fvar = '';
+    switch (t.params.aspace) {
+      case 'function':
```
dneto0, line 156: now that immedates have landed, please add that case.  Should always fail, so no need to skip if immediates are not supported. :-)

src/webgpu/shader/validation/functions/restrictions.spec.ts
```diff
@@ -519,6 +586,81 @@ const kFunctionParamValueCases: Record<string, ParamValueCase> = {
     matches: ['ptrWorkgroupOverrideExpr'],
     needsUnrestrictedPointerParameters: true,
   },
+
```
dneto0, line 589: this is a glorious set of tests.  I love how explicit these are.

src/webgpu/shader/validation/expression/call/builtin/bufferView.spec.ts
```diff
@@ -0,0 +1,435 @@
...
+      .beginSubcases()
+      .combine('ptr', [false, true] as const)
+  )
+  .fn(t => {
+    t.skipIfLanguageFeatureNotSupported('buffer_view');
+    const template_type = kTypes[t.params.template_type];
+    const return_type = kTypes[t.params.return_type];
+    let ret_type = return_type.toString();
+    if (t.params.ptr) {
+      ret_type = `ptr<storage, ${ret_type}>`;
```
dneto0, line 59: So this test only checks the `storage` address space.   

This test is not exactly excercising all return types, but rather that the store type of the return type matches the template type. 

Suggest renaming the test to 'returned_store_type'

src/webgpu/shader/validation/expression/call/builtin/bufferView.spec.ts
```diff
@@ -0,0 +1,435 @@
...
+@group(0) @binding(2) var<storage> sized_ro_storage : buffer<128>;
+@group(0) @binding(3) var<storage, read_write> sized_storage : buffer<128>;
+@group(0) @binding(4) var<uniform> sized_uniform : buffer<128>;
+var<workgroup> sized_workgroup : buffer<128>;
+override o : u32;
+var<workgroup> override_workgroup : buffer<o>;
+
+@compute @workgroup_size(1)
+fn main() {
+  let p = bufferView<u32>(${t.params.ptr ? '&' : ''}${t.params.type}, 0);
```
dneto0: I'd be happier if `p` was given an explicit type here.

Then we'd be additionally checking the address spaces on the return type.

src/webgpu/shader/validation/expression/call/builtin/bufferView.spec.ts
```diff
@@ -0,0 +1,435 @@
...
+fn bar(p : ptr<uniform, buffer>) {
+  let q = bufferView<vec4u>(p, 16);
+}
+fn foo() {
+  bar(&v);
+}`,
+    valid: false,
+    ptr_param: true,
+  },
+  uniform_buffer_too_small_through_unsized_function_override: {
```
dneto0, line 263: very nice test!

src/webgpu/shader/validation/expression/call/builtin/bufferView.spec.ts
```diff
@@ -0,0 +1,435 @@
...
+fn bar(p : ptr<workgroup, buffer<16>) {
+  let q = bufferView<array<vec2u, 3>(p, 0);
+}
+fn foo() {
+  bar(&v);
+}`,
+    valid: false,
+    ptr_param: true,
+  },
+  ro_storage_unsized_buffer_to_smaller_param: {
```
dneto0: the problem is not that it's smaller, it's that unsized buffer doesn't convert to sized buffer.

src/webgpu/shader/validation/expression/call/builtin/bufferView.spec.ts
```diff
@@ -0,0 +1,435 @@
...
+}`,
+  },
+  u32_array_ptr_composite: {
+    code: `
+@group(0) @binding(0) var<storage, read_write> v : buffer;
+@compute @workgroup_size(1)
+fn main() {
+  bufferView<array<u32, 4>>(&v, 0)[0] = 123u;
+}`,
+    requires: ['pointer_composite_access'],
```
dneto0, line 408: good one!

src/webgpu/shader/validation/expression/call/builtin/bufferView.spec.ts
```diff
@@ -0,0 +1,435 @@
...
+    requires: ['swizzle_assignment'],
+  },
+  vec4u_ptr_composite_swizzle_assign: {
+    code: `
+@group(0) @binding(0) var<storage, read_write> v : buffer;
+@compute @workgroup_size(1)
+fn main() {
+  bufferView<vec4u>(&v, 0).zxy = vec3u(42);
+}`,
+    requires: ['pointer_composite_access', 'swizzle_assignment'],
```
dneto0, line 426: Nice!

src/webgpu/shader/validation/expression/call/builtin/bufferArrayView.spec.ts
```diff
@@ -0,0 +1,524 @@
...
+fn main() {
+  ${t.params.must_use ? 'let p = ' : ''}bufferArrayView<array<u32>>(&v, 0, 64);
+}`;
+
+    t.expectCompileResult(t.params.must_use, wgsl);
+  });
+
+const kTypes = objectsToRecord(kAllScalarsAndVectors);
+
+g.test('return_type')
```
dneto0: similar to `bufferView`, I suggest renaming this test to `returned_store_type`.  (See bufferView comment for rationale)

src/webgpu/shader/validation/expression/call/builtin/bufferArrayView.spec.ts
```diff
@@ -0,0 +1,524 @@
...
+@group(0) @binding(2) var<storage> sized_ro_storage : buffer<128>;
+@group(0) @binding(3) var<storage, read_write> sized_storage : buffer<128>;
+@group(0) @binding(4) var<uniform> sized_uniform : buffer<128>;
+var<workgroup> sized_workgroup : buffer<128>;
+override o : u32;
+var<workgroup> override_workgroup : buffer<o>;
+
+@compute @workgroup_size(1)
+fn main() {
+  let p = bufferArrayView<array<u32>>(${t.params.ptr ? '&' : ''}${t.params.type}, 0, 128);
```
dneto0: please make p's type explicit, so we ensure the address space is correct.

src/webgpu/shader/validation/expression/call/builtin/bufferLength.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+@group(0) @binding(0) var<storage> v : buffer;
+@compute @workgroup_size(1)
+fn main() {
+  let res : ${type.toString()} = bufferLength(&v);
+}`;
+
+    t.expectCompileResult(type === Type.u32, wgsl);
+  });
+
+g.test('data_type')
```
dneto0, line 56: Please add a tests:
 - calling bufferLength on other types: e .g. scalar, array, ptr-to-scalar, ptr-to-array.  (don't have to do too much here)
 - wrong number of parameters: 0, and 2.

src/webgpu/shader/execution/expression/call/builtin/bufferView.spec.ts
```diff
@@ -0,0 +1,578 @@
...
+
+fn sizedLength(p : ptr<${t.params.aspace}, buffer<${t.params.bufferSize / 2}>>, gidx : u32) -> u32 {
+  return arrayLength(${addrOf}bufferView<${testcase.type}>(p, offsets[gidx])${access});
+}
+
+fn sized_indirectLength(p : ptr<${t.params.aspace}, buffer<${
+      t.params.bufferSize / 2
+    }>>, gidx : u32) -> u32 {
+  return unsizedLength(p, gidx);
+}
```
dneto0, line 208: This is a great test.

Suggest another scenario: 
-  call to sized-N param that calls into sized-N/2 param  --> yields N/2

src/webgpu/shader/execution/expression/call/builtin/bufferView.spec.ts
```diff
@@ -0,0 +1,578 @@
...
+@group(0) @binding(1) var<storage, read_write> out : buffer<256 * 4>;
+
+
+var<workgroup> wg_var : buffer<256 * 4>;
+
+@compute @workgroup_size(1)
+fn main() {
+  let val = ${testcase.f16 ? 'f16' : testcase.f32 ? 'f32' : 'u32'}(in);
+  let ptr = ${call_expr};
+  (*${t.params.assign === 'call' ? call_expr : 'ptr'})${testcase.access} = val;
```
dneto0, line 363: Really great to see both 'let' and 'call' variations here!

src/webgpu/shader/execution/expression/call/builtin/bufferView.spec.ts
```diff
@@ -0,0 +1,578 @@
...
+@compute @workgroup_size(1)
+fn main() {
+  let ptr = ${call_expr};
+  (*${t.params.assign === 'call' ? call_expr : 'ptr'})${swizzle} = in;
+
+  ${post_assign}
+}
+`;
+
+    runReadWriteTest(false, t, wgsl, ele_ty, ty, t.params.aspace, t.params.offset, bufferSize);
```
dneto0, line 583: Very impressed and happy with the creation and use of `readWriteTest` to clarify  so much of the mechanics!

src/webgpu/shader/execution/expression/call/builtin/bufferArrayView.spec.ts
```diff
@@ -0,0 +1,612 @@
...
+  return arrayLength(${addrOf}bufferArrayView<${
+    testcase.type
+  }>(p, offsets[gidx], sizes[gidx])${access});
+}
+
+fn sized_indirectLength(p : ptr<${t.params.aspace}, buffer<${
+      t.params.bufferSize / 2
+    }>>, gidx : u32) -> u32 {
+  return unsizedLength(p, gidx);
+}
```
dneto0, line 230: Similar to arrayView, suggest another scenario:

- call to sized-N param that calls into sized-N/2 param --> yields N/2

dneto0 (changes requested):
Finished reviewing.  This is a very impressive suite. Thanks for creating it!

I also suggest adding additional validation test cases for `arrayLength`, i.e. in `src/webgpu/shader/validation/expression/call/arrayLength`.  Sized and unsized buffer types should be tested.

src/webgpu/api/validation/dispatch.spec.ts
alan-baker: Updated the spec after to discussion to only use minTypeSize. So updated the tests to reflect that. So offset and/or size are not considered now.

src/webgpu/shader/validation/expression/call/builtin/bufferLength.spec.ts
dneto0: i see the new test. thanks

src/webgpu/shader/execution/expression/call/builtin/bufferView.spec.ts
dneto0: I see it: the new  `sized_sized_indirect` case

src/webgpu/api/validation/dispatch.spec.ts
dneto0: Thanks. I see that.

dneto0 (changes requested):
You've addressed all my feedback.

Please add a test case (that you mentioned offline) for dispatch validation:

- in a render pipeline, when vertex and fragment use the same unsized buffer, the min size is the larger of the two uses. 

I looked for it in `webgpu:api,validation,dispatch:shader_required_buffer_size`  but it's fine for it to be in a new test, to avoid complicating that one.

alan-baker:
> You've addressed all my feedback.
> 
> Please add a test case (that you mentioned offline) for dispatch validation:
> 
> * in a render pipeline, when vertex and fragment use the same unsized buffer, the min size is the larger of the two uses.
> 
> I looked for it in `webgpu:api,validation,dispatch:shader_required_buffer_size` but it's fine for it to be in a new test, to avoid complicating that one.

I've added some simple tests for this. I moved them all to a new file. There's no point checking uniform since it must be sized and that will take precedence.

src/webgpu/api/validation/min_buffer_binding_size.spec.ts
```diff
@@ -0,0 +1,622 @@
...
+    binding_type: 'read-only-storage',
+    requires: ['buffer_view'],
+  },
+  fragment_larger_buffer_array_view_ro_storage: {
+    code: `
+@group(0) @binding(0) var<storage> buf : buffer;
+
+@vertex
+fn vs() -> @builtin(position) vec4f {
+  _ = bufferArrayView<array<u32>>(&buf, 0, 16);
```
dneto0: nit: It would be better to use a 'size' param of something like 8 here, to avoid thinking that this 16 drives the ultimate size of 16 (line 494)

dneto0 (approved):
LGTM! Thank you.
(with one nit)

src/webgpu/api/validation/min_buffer_binding_size.spec.ts
alan-baker: done

dneto0 (approved):

Issue #4650, opened 2026-05-29
Undefined behavior tested by quadBroadcast/quadSwap split test cases


jenatali:
Running the WebGPU CTS against WARP, I see failures in:
* webgpu:shader,execution,expression,call,builtin,quadSwap:compute,split:predicate="lower_half";wgSize=[4,1,1]
* webgpu:shader,execution,expression,call,builtin,quadSwap:compute,split:predicate="lower_half";wgSize=[1,4,1]
* webgpu:shader,execution,expression,call,builtin,quadSwap:compute,split:predicate="lower_half";wgSize=[1,1,4]
* webgpu:shader,execution,expression,call,builtin,quadSwap:compute,split:predicate="upper_half";wgSize=[4,1,1]
* webgpu:shader,execution,expression,call,builtin,quadSwap:compute,split:predicate="upper_half";wgSize=[1,4,1]
* webgpu:shader,execution,expression,call,builtin,quadSwap:compute,split:predicate="upper_half";wgSize=[1,1,4]
* webgpu:shader,execution,expression,call,builtin,quadBroadcast:compute,split:predicate="lower_half";wgSize=[4,1,1]
* webgpu:shader,execution,expression,call,builtin,quadBroadcast:compute,split:predicate="lower_half";wgSize=[1,4,1]
* webgpu:shader,execution,expression,call,builtin,quadBroadcast:compute,split:predicate="lower_half";wgSize=[1,1,4]
* webgpu:shader,execution,expression,call,builtin,quadBroadcast:compute,split:predicate="upper_half";wgSize=[4,1,1]
* webgpu:shader,execution,expression,call,builtin,quadBroadcast:compute,split:predicate="upper_half";wgSize=[1,4,1]
* webgpu:shader,execution,expression,call,builtin,quadBroadcast:compute,split:predicate="upper_half";wgSize=[1,1,4]

As far as I can tell, these test cases split the work group in half, with half of the lanes active, and half inactive. The description of the quad tests includes the wording:
> Quad operations require a fully active quad to operate correctly so several of the predication filters are skipped.

But the test cases that WARP fails clearly fall into a not-fully-active quad, since it's a group size of 4 (one quad) which is half active. I believe these are test bugs and these cases should be skipped.

jenatali:
[...]

PR #4656, opened 2026-06-02
Skip quadBroadcast/quadSwap split tests when subgroup_size < 8

gyagp:
The compute,split tests use the predicate `id < subgroupSize / 2`, which deactivates the upper half of each subgroup. When subgroupSize is 8 or greater the boundary lands on a 4-lane multiple, so every quad stays fully active and the test exercises legitimate predicated quad operations. When subgroupSize is 4, the only quad gets bisected, leaving no fully active quad — calling quadBroadcast/quadSwap there is undefined behavior and the result is not meaningful to validate.

GPUAdapterInfo.subgroupMinSize is not a sufficient pre-check: the size the implementation actually selects depends on the compiled shader (workgroup shape, register pressure, vendor heuristics), not just the adapter's minimum. Two implementations from the same vendor can report identical {min,max} and pick different runtime sizes for the same shader. Reading @builtin(subgroup_size) from inside the test shader is the only reliable signal.

Issue: #4650

<hr>

- [ ] ~New helpers are `/** documented */` and new helper files are found in `helper_index.txt`.~
- [ ] ~Test have be tested with compatibility mode validation enabled and behave as expected. (If not passing, explain above.)~


src/webgpu/shader/execution/expression/call/builtin/quadBroadcast.spec.ts
```diff
@@ -326,6 +326,17 @@ predication filters are skipped.
     const wgThreads = t.params.wgSize[0] * t.params.wgSize[1] * t.params.wgSize[2];
     const testcase = kPredicateCases[t.params.predicate];
 
+    // Skip if the implementation might choose a subgroup size of 4 for a
+    // 4-thread workgroup (as WARP does). With subgroup size 4 and a
+    // split predicate, only 2 lanes remain active in the only quad,
+    // leaving no fully active quad (undefined behavior).
+    if (wgThreads === 4 && t.device.adapterInfo.subgroupMinSize <= 4) {
```
jenatali: Worth also skipping unconditionally if max subgroup size is 4, since that would split every quad as well? Though, WARP still only chooses native wave sizes without explicit requests to go wider, so on arm64 systems, I think it would pick subgroup size 4 since it only supports NEON (and even with SVE there's no real systems that support vector length > 128 bit).

So I think this might need to be a skip from inspecting the subgroup size as reported by running the shader itself, rather than a pre-emptive skip.

alan-baker: That's simple enough as we already capture that data in the shader.

gyagp:
I changed to use the reported subgroup size instead. Below are the test results on various situations.
  | @workgroup_size | NVIDIA RTX 5080 | WARP 1.0.20 | WARP 10.0.26100.8246 (Win11 24H2 built-in) |
  |---|---:|---:|---:|
  | [4,1,1]    | 32 | 4  | 4 |
  | [8,1,1]    | 32 | 8  | 4 |
  | [16,1,1]   | 32 | 16 | 4 |
  | [32,1,1]   | 32 | 16 | 4 |
  | [64,1,1]   | 32 | 16 | 4 |
  | [256,1,1]  | 32 | 16 | 4 |
  | [4,4,4]    | 32 | 16 | 4 |

  All three report adapter.subgroupMinSize=4 (WARP) or 32 (NVIDIA), but the runtime size for the same shader differs across:
    * NVIDIA: always 32 — no skip needed.
    * WARP 1.0.20: min(wgThreads, 16) — skip only the subgroup size is 4 (6 of the 46 compute,split cases).
    * WARP system Windows 11 24H2: always 4 — every case skips.

jenatali (approved):

alan-baker (approved):

PR #4658, opened 2026-06-05
Add shared support for WGSL immediate validation

shaoboyan091:
This is the base PR for follow-up WGSL `immediate` address space
validation coverage.

It adds the shared CTS support needed by later test-suite PRs:

- adds `immediate_address_space` to the known WGSL language features
- models `immediate` in the shared address space metadata
- emits required WGSL feature headers from declaration helpers
- skips immediate validation tests when immediate data support is not exposed
- removes redundant address-space type assertions made unnecessary by the
  stricter metadata typing

Follow-up PRs will add the declaration, type, expression, function, and
uniformity test coverage one commit at a time after this base PR lands.

Testing:

- `npm.cmd run typecheck`
- `npm.cmd run lint`

### DIFF
--- a/src/webgpu/shader/validation/decl/var.spec.ts
+++ b/src/webgpu/shader/validation/decl/var.spec.ts
@@ -4,7 +4,7 @@ Validation tests for host-shareable types.
 
 import { makeTestGroup } from '../../../../common/framework/test_group.js';
 import { keysOf } from '../../../../common/util/data_tables.js';
-import { AddressSpace, kAccessModeInfo, kAddressSpaceInfo } from '../../types.js';
+import { kAccessModeInfo, kAddressSpaceInfo } from '../../types.js';
 import { ShaderValidationTest } from '../shader_validation_test.js';
 
 import {
@@ -582,9 +582,7 @@ g.test('address_space_access_mode')
   });
 
 // Address spaces that can hold an i32 variable.
-const kNonHandleAddressSpaces = keysOf(kAddressSpaceInfo).filter(
-  as => as !== 'handle'
-) as AddressSpace[];
+const kNonHandleAddressSpaces = keysOf(kAddressSpaceInfo).filter(as => as !== 'handle');
 
 g.test('explicit_access_mode')
   .desc('Validate uses of an explicit access mode on a var declaration')
--- a/src/webgpu/shader/validation/decl/var.spec.ts
+++ b/src/webgpu/shader/validation/decl/var.spec.ts
@@ -582,7 +582,7 @@ g.test('address_space_access_mode')
   });
 
 // Address spaces that can hold an i32 variable.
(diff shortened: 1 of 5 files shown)


shaoboyan091:
@jrprice ping softly for reviewing.

src/webgpu/shader/validation/decl/var.spec.ts
```diff
@@ -582,9 +582,7 @@ g.test('address_space_access_mode')
   });
 
 // Address spaces that can hold an i32 variable.
-const kNonHandleAddressSpaces = keysOf(kAddressSpaceInfo).filter(
-  as => as !== 'handle'
-) as AddressSpace[];
+const kNonHandleAddressSpaces = keysOf(kAddressSpaceInfo).filter(as => as !== 'handle' && as !== 'immediate');
```
jrprice: Why did you have to exclude `immediate` here? Seems like these tests should still be able to run since they pull the info from `kAddressSpaceInfo`?

src/webgpu/shader/validation/types/pointer.spec.ts
```diff
@@ -140,9 +140,7 @@ g.test('type')
   });
 
 // Address spaces that can hold an i32 variable.
-const kNonHandleAddressSpaces = keysOf(kAddressSpaceInfo).filter(
-  as => as !== 'handle'
-) as AddressSpace[];
+const kNonHandleAddressSpaces = keysOf(kAddressSpaceInfo).filter(as => as !== 'handle' && as !== 'immediate');
```
jrprice: As above, why did you have to exclude `immediate` here?

src/webgpu/shader/validation/decl/var.spec.ts
shaoboyan091: This is the review comments from copilot. It said, if I don't have the concrete immediate test cases yet, we should exclude it. I think it is fair. Maybe I need a TODO for this.

jrprice: Did you try running those particular tests without excluding immediates? AFAICT the tests should already do the right thing with immediates because you defined the immediate config in `kAddressSpaceInfo`.

shaoboyan091: Ah, good point. I think I ran them before this commit and seems good. But let me double check.

shaoboyan091: Yes, the local running shows no need to do this but let me check the logic in the test file to confirm it.

shaoboyan091: And the logic proves that current exist cases handles immediates address space good from my local try result. So maybe we could include `immediate`?

jrprice (approved):

PR #4659, opened 2026-06-05
Add declaration tests for WGSL immediate variables

shaoboyan091:
This PR adds the first WGSL `immediate` address space validation
coverage.

It includes the shared CTS plumbing required by the tests:

- adds `immediate_address_space` to the known WGSL language features
- models `immediate` in the shared address space metadata
- emits required WGSL feature headers from declaration helpers
- skips immediate validation tests when immediate data support is not exposed

It then adds declaration validation coverage for `var<immediate>`:

- accepted scalar, vector, matrix, and struct store types
- rejected store types, including arrays and f16 cases
- module-scope-only declarations
- rejection of binding attributes
- rejection of explicit access modes
- one statically used immediate variable per entry point
- basic `ptr<immediate>` declaration behavior

Follow-up PRs will add type, expression, function, and uniformity
coverage one suite at a time after this lands.

Testing:

- `npm.cmd run typecheck`
- `npm.cmd run lint`

### DIFF
--- a/src/webgpu/capability_info.ts
+++ b/src/webgpu/capability_info.ts
@@ -987,6 +987,7 @@ export const kKnownWGSLLanguageFeatures = [
   'swizzle_assignment',
   'linear_indexing',
   'texture_formats_tier1',
+  'immediate_address_space',
 ] as const;
 
 export type WGSLLanguageFeature = (typeof kKnownWGSLLanguageFeatures)[number];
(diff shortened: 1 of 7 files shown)


shaoboyan091:
Closing this combined PR. We are going back to the separated PR flow: #4658 will carry only the shared support, and the declaration CTS cases will be sent in a follow-up PR after that lands.

PR #4661, opened 2026-06-11
Add declaration tests for WGSL immediate variables

shaoboyan091:
Adds a focused declaration suite for `var<immediate>`.

The new tests cover accepted scalar, vector, matrix, and struct store types, rejected store types, module-scope-only declarations, rejection of binding attributes, rejection of explicit access modes, and the one-immediate-variable-per-entry-point rule.

It also threads `immediate` through the existing declaration tests so the generic address space coverage checks the same declaration access rules as storage, uniform, private, workgroup, and function variables. The context-dependent resolution table now recognizes `immediate_address_space` as a language feature name.

Notes for reviewers:
- Tests skip on the `immediate_address_space` WGSL language feature, so they run only where the `requires immediate_address_space;` shaders can compile (not merely where the immediate API surface is exposed).
- `f16`/`vec3h` are treated as valid immediate store types (host-shareable and constructible, gated on `shader-f16`), consistent with how `uniform` accepts them.

Issue: #

<hr>

**Requirements for reviewer sign-off:**

- [ ] Test descriptions are accurate and complete.
- [ ] Tests avoid over-parameterization (see case count report).


shaoboyan091:
@jrprice  PTAL, catch a potential "issue" in chrome canary, it seems canary supported fixed array in immediate address space but the spec said it shouldn't.

Fixed array could be supported in implementation but it doesn't be discussed in community yet.

shaoboyan091:
@jrprice  softly ping for reviewing, thanks!

src/webgpu/shader/validation/decl/immediate.spec.ts
```diff
@@ -0,0 +1,302 @@
...
+export const g = makeTestGroup(ShaderValidationTest);
+
+const kImmediateFeature = 'immediate_address_space' as const;
+const kImmediateHeader = `requires ${kImmediateFeature};`;
+
+const kValidStoreTypes = {
+  u32: { enable: ``, prelude: ``, type: `u32` },
+  i32: { enable: ``, prelude: ``, type: `i32` },
+  f32: { enable: ``, prelude: ``, type: `f32` },
+  f16: { enable: `enable f16;`, prelude: ``, type: `f16` },
```
dj2, line 20: Why is `f16` here if it isn't supported yet?

src/webgpu/shader/validation/decl/immediate.spec.ts
```diff
@@ -0,0 +1,302 @@
...
+${testcase.prelude}
+var<immediate> data : ${testcase.type};
+@compute @workgroup_size(1)
+fn main() {
+  _ = data;
+}`;
+    t.expectCompileResult(false, wgsl);
+  });
+
+g.test('scope')
```
dj2, line 86: Given the test is mostly predicated on the `module` vs `function` can we just make 2 tests?

src/webgpu/shader/validation/decl/immediate.spec.ts
jrprice: The spec text we landed seems to allow f16, but I can't find any discussion that says that this was intentional. That said it's probably easy enough to support since in Tint at least we just decompose the buffers anyway. I've asked internally whether this is expected to be supported or not.

@shaoboyan091 do we already have CTS execution/operation tests that cover f16 in immediate, to make sure they actually work?

src/webgpu/shader/validation/decl/immediate.spec.ts
jrprice: Yeah the code paths are too different for module-scope to be acting as a control-case here.

Suggest dropping module-scope since it's covered by other tests, and just having function scope where the control case uses the `function` address space.

src/webgpu/shader/validation/decl/immediate.spec.ts
shaoboyan091: TBH, This is from this https://github.com/gpuweb/cts/blob/main/src/webgpu/api/operation/command_buffer/programmable/immediate.spec.ts#L410
But I suspect this is due to API limit https://github.com/gpuweb/cts/issues/4297 and shouldn't affect decl test here. Let me change the code and try.

shaoboyan091: After adding f16 support, local try report " error: using 'f16' in 'immediate' address space is not implemented yet" And I find in dawn part. https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/src/tint/lang/wgsl/resolver/address_space_validation_test.cc;drc=874b76badc54f1ece2369347db39ee5406678021;l=756. That's a bit surprise me.

shaoboyan091: I noticed the spec PR https://github.com/gpuweb/gpuweb/pull/6297 about restrict f16 support, is this because we don't have implementation to verify the ability?

dj2:
Discussing internally, we think `f16` should be supportable, so it isn't an oversight, but just a bug in Tint. So, we should set the CTS to have it as available, remove the skip and then file a bug to fix the Tint code.

shaoboyan091:
> Discussing internally, we think `f16` should be supportable, so it isn't an oversight, but just a bug in Tint. So, we should set the CTS to have it as available, remove the skip and then file a bug to fix the Tint code.

Good to know this. F16 is important.

shaoboyan091:
@dj2 and @jrprice  pls take a look at new commits which should address above comments.

shaoboyan091:
@dj2 and @jrprice  softly ping for another review

shaoboyan091:
Hi, @kainino0x  can you help on this?

jrprice (approved):
LGTM, sorry for the delay.

shaoboyan091:
@jrprice  Thanks!

Issue #4592, opened 2026-02-12
Is time-based auto-chunking useful?


kainino0x:
In #2770 we added a mechanism to automatically subdivide WPT tests into chunks that don't individually take too long. This was to make the tests runnable under the WPT harness-adapter in runtimes that have per-test limits (like Chromium's 5-second default limit, though Chromium does not use the WPT adapter).

@mwyrzykowski @ErichDonGubler do you know if you're relying on this mechanism at all? It takes some work to maintain `listing_meta.json` and we have not been keeping it up to date. So I'd like to remove it, if it has turned out not to be useful.

(Apologies if I asked this before. I can't find it.)

mwyrzykowski:
We are not using it @kainino0x , though we have been running into timeouts for some of the slower tests, though I think that is a separate issue.

jimblandy:
Some of Firefox's runtimes are catastrophic; see the bugzilla link if you are curious about the details. Incorporating that data into any dataset would make it pretty weird. It will get much better soon.

PR #4680, opened 2026-07-08
Remove code using the listing_meta.json file.

dj2:
This CL removes the infrastructure which validates and uses the listing_meta.json file. The listing_meta.json file itself will be removed in a followup CL.

Issue: #4592


dj2:
I tested locally by running some cases with dawn.node and by running some cases through Chrome using the `npm serve` runner. Is there anything else I need to test?

dj2:
Hm, it seems we use the case count for reporting on CI. So I may need to try to put that bit back.

kainino0x:
Thanks for the cleanup, LGTM other than that

dj2:
Ok, test case counts are restored.

kainino0x (approved):
LGTM! We could probably also remove the code that sends test times to a websocket, but that's not really bothering anyone so it's probably good to leave.

PR #4706, opened 2026-09-18
Test alphaToCoverageEnabled constraints

Kangz:
That a fragment stage and target 0 must exist and that its format must be blendable and have an alpha channel.

Add componentCount to color formats in format_info.ts as there was no way to know which formats have an alpha channel.

Add an isColorTextureFormatWithAlpha helper.

Fix isTextureFormatBlendable to take `texture-formats-tier1` into account.

Issue: #

<hr>


src/webgpu/api/validation/render_pipeline/multisample_state.spec.ts
```diff
@@ -47,6 +52,90 @@ g.test('alpha_to_coverage,count')
...
+
+g.test('alpha_to_coverage,first_format_blendable_and_has_alpha')
+  .desc(
+    `If multisample.alphaToCoverageEnabled is true, a fragment.targets[0].format must be blendable.`
+  )
+  .params(u =>
+    u
+      .combine('isAsync', [false, true])
+      .combine('alphaToCoverageEnabled', [false, true])
+      //.beginSubcases()
```
kainino0x, line 120: uncomment or remove

src/webgpu/format_info.ts
```diff
@@ -778,6 +821,7 @@ const kBCTextureFormatInfo = formatTableWithDefaults({
     'bc1-rgba-unorm': {
       color: {
         type: 'float',
+        componentCount: 4,
```
kainino0x, line 824: nit: For non-renderable formats does a componentCount really mean anything? Sampling any format will give you 4 channels even if it doesn't have 4 channels. The number of components probably only matters for render targets.

I think it would make sense to omit these, but I don't feel too strongly.

src/webgpu/format_info.ts
```diff
@@ -2328,6 +2426,10 @@ export function isColorTextureFormat(format: GPUTextureFormat) {
   return !!kTextureFormatInfo[format].color;
 }
 
+export function isColorTextureFormatWithAlpha(format: GPUTextureFormat) {
+  return isColorTextureFormat(format) && !!(kTextureFormatInfo[format].color?.componentCount === 4);
```
kainino0x, line 2430: `!!` appears to be a no-op here? (it's already a bool)

```suggestion
  return isColorTextureFormat(format) && kTextureFormatInfo[format].color?.componentCount === 4;
```

kainino0x (approved):

src/webgpu/api/validation/render_pipeline/multisample_state.spec.ts
```diff
@@ -47,6 +52,90 @@ g.test('alpha_to_coverage,count')
...
+      .combine('format', kColorTextureFormats)
```
kainino0x, line 121: Use `kPossibleColorRenderableTextureFormats`

src/webgpu/format_info.ts
Kangz: It's just a bit hard to omit, or it would have to be moved in `colorRender`. But also it's additional info about the format that's not wrong, just unclear if it is totally necessary in the future. Tentatively keeping.

